PR #197, opened 2016-05-19
Chartbeat: remove hardcoded `http` protocol from the API URLs

### LINKED ISSUE #196: Don't use hardcoded http protocol in Chartbeat's JS
The Chartbeat's JS ( https://github.com/Automattic/vip-go-mu-plugins/blob/master/shared-plugins/chartbeat/media/cbdashboard.compiled.js ) contains some hardcoded API `http` only calls.

The latest version on the WP.org contains them as well. Since we have already manually hacked the plugin on the WordPress.com, we might want to do the same there as well.

This is affecting a production site hosted on VIP Go.

david-binda:
since it's causing Mixed content warnings on wp-admin dashboard

Since the latest version of the plugin hosted on the WordPress.org repository is also hardcoding the http protocol for certain dashboard specific API requests, we can't simply update the plugin and have to modify the minified file manually.

Fixes #196


david-binda:
@nickdaugherty or @joshbetz may I ask you for a review? It's affecting one of the clients on the VIP Go which is about to launch soon.

david-binda:
I perhaps also could ask @emrikol for a review :)

nickdaugherty:
@david-binda The change looks ok, but we should probably have this client put the plugin in their `plugins` directory instead, as shared plugins will be removed at some point.

david-binda:
thanks for the review @nickdaugherty ! Talking to @simonwheatley I'll go ahead and merge the PR, get it fixed on the platform for all the clients and will ask this particular one to copy the plugin to their repository in preparation for the shared plugins deprecation.

PR #199, opened 2016-05-21
Update hooks used to check filename uniqueness and filetype validity

ethitter:
WordPress 4.5 introduced a new filter, `wp_unique_filename`, which we can use instead of relying on a variable hook to modify the ultimate filename. Since themes and plugins can change the action that is used in the variable hook, it's possible for our filters to be bypassed. Switching to this new filter ensures that regardless of how files are uploaded to Go sites, they'll receive unique filenames.


ethitter:
@darthhexx, I'd really appreciate your input on this, given your ownership of the bits I'm poking at in these changes.

darthhexx:
Looks good to me @ethitter

PR #200, opened 2016-05-26
Add feed caching from WPCOM to Go

sboisvert:
The get_lastpostmodified() and get_lastpostdate() are SUPER slow, on .com we circumvent these functions to avoid the 3s+ queries. This should fix all the slowness in */feed/* on Go sites

(totally untested)

### DIFF
--- a/wpcom-feed-cache.php
+++ b/wpcom-feed-cache.php
@@ -1,0 +1,166 @@
+<?php
+
+/* Various functions dealing with how we cache feeds on WordPress.com
+ * Feeds are cached upstream from PHP via nginx, so we need to make sure
+ * that we invalidate the cache when something happens
+ *
+ * We also want to make sure we only cache feeds we can invalidate - this means
+ * no query strings
+ * */
+
+
+function wpcom_invalidate_post_data( $post_id ) {
+	global $wpdb;
+	
+	// Let's try to skip this when importing - it can queue tons of jobs that are un-necessary
+	if ( defined( 'WP_IMPORTING' ) && constant('WP_IMPORTING') )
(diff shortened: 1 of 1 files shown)


emrikol:
Just doing a cursory look, I think we'll need to do a bit of work and testing on this to make it work with the single site installs.  Probably just needs some `is_multisite()` love

emrikol:
It looks like this isn't adding any caching, but only invalidating the nginx cache used for feeds on WordPress.com.  It's also heavily relying on the internal WordPress.com jobs system.

It looks to me like both of these functions are already cached in core (per post type) but please correct me if I'm wrong.

nickdaugherty:
Feeds get cached in Varnish on Go, and our existing cache invalidation logic clears them out when needed, such as on post update - https://github.com/Automattic/vip-go-mu-plugins/blob/8cee7130362c8c17a6c6f211a01cf39b071be339/cache-manager.php#L220-L228

sboisvert:
get_lastpostmodified() and get_lastpostdate() are still slow for sites like DenverPost so something is happening that isn't on .com. I can try to look into it some more.

joshbetz:
@nickdaugherty We should consider lowering the ttl for feeds to something like 5 minutes and not busting the cache every time a post is published.

sboisvert:
What was the decision on this?

joshbetz:
@sboisvert I think the solution will be outside mu-plugins. As Nick mentioned, feeds are already cached in Varnish, so extra Nginx caching should be unnecessary. We may still need to adjust the cache TTL and/or PURGE logic though.

PR #202, opened 2016-05-30
Add URI to New Relic

sboisvert:
We want to know the exact URL that was hit so we can reproduce what's happening.


sboisvert:
This will also help us track down 404s

nickdaugherty:
@sboisvert Do you envision us using this on every site, or just enabling it ad-hoc?

sboisvert:
For all sites

sboisvert:
Now with 100% more hooking!

nickdaugherty:
Looks good, merging.

PR #203, opened 2016-05-31
Sync WP.com performance tweaks

sboisvert:
Add performance tweaks that have been added to .com

Many of these should be enabled by default but they are not on GO they are enabled in .com inside /plugins/vip-helper-wpcom.php
Where should we add functions from that file ^^?

### DIFF
--- a/vip-helpers/vip-caching.php
+++ b/vip-helpers/vip-caching.php
@@ -129,11 +129,11 @@ function wp_flush_term_exists( $term, $tt_id, $taxonomy, $deleted_term ){
  * Returns permalink for a taxonomy term archive, or a WP_Error object if the term does not exist.
  *
  * @param int|string|object $term The term object / term ID / term slug whose link will be retrieved.
- * @param string $taxonomy The taxonomy slug. NOT required if you pass the term object in the first parameter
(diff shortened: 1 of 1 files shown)


simonwheatley:
I'd say add them in https://github.com/Automattic/vip-go-mu-plugins/blob/master/000-vip-init.php

nickdaugherty:
I think they should go into an appropriate file in the `vip-helpers` directory - `000-vip-init.php` should really only 'bootstrap' by requiring other files.

nickdaugherty:
All the stampedeless cache stuff here won't work b/c `require_lib` doesn't exist, and neither does the class - can we remove it? It's really easy to use it wrong and cause worse performance, so probably doesn't need to be kept.

joshbetz:
Is there an opportunity to get any of this in core?

joshbetz:
@sboisvert What's the status of this?

sboisvert:
Where is the totally haven't looked at it emoji?

joshbetz:
@sboisvert No worries. If you can push to a branch on this repo, I can help clean this up and get it deployed. Would be good to keep wp.com and Go in sync.

PR #204, opened 2016-06-01
Add a filter to allow purged urls to be modified by client code

nickdaugherty:
Clients have custom endpoints, such as feeds, etc, that also need
purged to avoid a 30m TTL.

This makes the list of purged urls filterable by client code, passing
the list and post id as arguments

### DIFF
--- a/cache-manager.php
+++ b/cache-manager.php
@@ -229,6 +229,8 @@ class WPCOM_VIP_Cache_Manager {
 
 		foreach ( $feeds as $feed )
 			$this->purge_urls[] = $feed;
+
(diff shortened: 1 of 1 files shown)


ethitter:
Looks pretty good. Since we're now giving clients the ability to modify the list, should we validate the final result, filter out empties or duplicates, etc?

nickdaugherty:
Duplicates get filtered out here https://github.com/Automattic/vip-go-mu-plugins/blob/8cee7130362c8c17a6c6f211a01cf39b071be339/cache-manager.php#L152

Is that good enough for now?

ethitter:
Yup, works for me!

PR #182, opened 2016-04-29
Prevent concatenation outside VIP Go

simonwheatley:
Closes #181

PR #205, opened 2016-06-01
Only concatenate on the correct environments

### LINKED ISSUE #181: Check for WPCOM_IS_VIP_ENV before concatenating
This will prevent the concatenation from breaking environments that can't deal with it (i.e. environments that aren't either VIP Go or VIP Go Quickstart).

scottsweb:
Closes #181


scottsweb:
Already started https://github.com/Automattic/vip-go-mu-plugins/pull/182

PR #207, opened 2016-06-02
Use WP-Cron Control to, well, control cron execution from an external service

ethitter:
The WordPress.com jobs system will be used trigger cron tasks, with the aid of the WP-Cron Control plugin. This is, essentially, how cron works on WordPress.com VIP.

### DIFF
--- a/wp-cron-control.php
+++ b/wp-cron-control.php
@@ -8,7 +8,7 @@
  Text Domain: wp-cron-control
  */
 
-class WPCom_VIP_Cron_Control {
+class WPCOM_VIP_Cron_Control {
 	/**
 	 * Register hooks
 	 */
@@ -57,4 +57,4 @@ class WPCom_VIP_Cron_Control {
 	}
 }
 
-new WPCom_VIP_Cron_Control;
+new WPCOM_VIP_Cron_Control;
--- a/wp-cron-control.php
+++ b/wp-cron-control.php
@@ -8,12 +8,9 @@
  Text Domain: wp-cron-control
  */
 
-// TODO:
-// Global WP_CRON_CONTROL_SECRET
-
 class WPCom_VIP_Cron_Control {
 	/**
-	 *
+	 * Register hooks
 	 */
 	public function __construct() {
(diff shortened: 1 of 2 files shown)


ethitter:
For the record, the branch number is because I conflated this repo with Automattic/vipd: https://github.com/Automattic/vipd/issues/220.

ethitter:
Depends on an async job to trigger the crons (Phabricator forthcoming), and a value for `WP_CRON_CONTROL_SECRET` in a shared location that isn't this public repo. ;)

ethitter:
See also https://code.a8c.com/D1935.

ethitter:
Since the shared secret won't differ across sites (doing so would require storing it in sitemeta, to make it available via the API), it'll be added to the `wp-config.php` stubs in our software stacks. The key will also be defined on WordPress.com, to make it accessible from the jobs system. 

Accordingly, the Phabricator differential noted above must be deployed first, followed by the shared secret's addition to `wp-config.php`. After both are deployed, the updates from this PR can be deployed. If this PR isn't the last, cron will be completely blocked until this PR is deployed.

simonwheatley:
Noting that we need to enhance this functionality (or perhaps the Cron Control plugin?) to accommodate triggering cron on all the sites in a multisite instance.

wp-cron-control.php
```diff
@@ -0,0 +1,60 @@
...
+/*
+ Plugin Name: WP-Cron Control
+ Plugin URI: https://wordpress.org/plugins/wp-cron-control/
+ Description: Take control of wp-cron execution.
+ Author: Thorsten Ott, Erick Hitter, Automattic
+ Version: 0.7.1
+ Text Domain: wp-cron-control
+ */
+
+class WPCom_VIP_Cron_Control {
```
nickdaugherty: Can we change the casing here to `WPCOM`?

nickdaugherty:
@ethitter left a note about `WPcom_VIP`vs `WPCOM_VIP`, otherwise looks good 👍

ethitter:
Good catch. Took care of that, now just waiting for Travis to finish...

nickdaugherty:
We need to fix Travis for this repo...

PR #212, opened 2016-06-04
Sync syndication cache

david-binda:
In order to properly setup the cache class for the Syndication plugin as it was implemented by @sboisvert on WordPress.com platform, we have to load the class early during the WP init

This code has already been deployed to the VIP Go platform


david-binda:
Thanks @nickdaugherty !

PR #213, opened 2016-06-06
Add debug bar back

nickdaugherty:
Adds basic debug bar


ethitter:
We can't use `is_automattician()`, not unless we can move from `after_setup_theme` to `init`. The user isn't available until right between those two actions. For reference, we dealt with this in Query Monitor by using a capability: https://github.com/Automattic/vip-go-mu-plugins/blob/c9c05beb3ab5a39d5c222dd3a7689f5e5ce5a07c/query-monitor.php.

The `is_automattician()` check "works" on WordPress.com because WordPress Core is forgiving, and because we built the platform around early (ab)use of the user, but we can't expect the same on Go.

nickdaugherty:
@ethitter - This is actually how we had it working before on Go, and I just tested it out again successfully.

`is_automattician()` internally uses `wp_get_current_user()`, which is populated by the time this hook runs.

We can move it to `init`, though it'll need to be at a low priority as Query Monitor conditionally loads a `Debug_Bar` class if one doesn't exist yet on `init`, which causes duplicate class errors.

ethitter:
It works only because WordPress is forgiving. We're technically `_doing_it_wrong()`, though, as shown in `wp-settings.php`: https://github.com/WordPress/WordPress/blob/4.5.2/wp-settings.php#L379. Core internally calls the same user-setup function that we we do (https://github.com/WordPress/WordPress/blob/4.5.2/wp-includes/class-wp.php#L587), and our "preemptive" call comes very close to where Core expects it, so it may not break anything. Core has tried to be less forgiving in the past, though (https://core.trac.wordpress.org/ticket/25690), so just something to bear in mind. We can probably leave it alone for now.

joshbetz:
Should we also try to get object cache stats in Query Monitor or just depend on Debug Bar?

nickdaugherty:
Query Monitor should have full object cache stats too.

nickdaugherty:
@ethitter I fixed the hook - now to add the custom panels (some need tweaked).

nickdaugherty:
I'm considering this mergeable now that we have the custom db queries panel. Will keep working on the additional panel enhancements.

Needs plenty of independent testing before merge. cc @ethitter

ethitter:
Sorry for the delay. I'm digging into this tonight.

vip-helpers/vip-utils.php
```diff
@@ -1130,3 +1130,38 @@ function wpcom_vip_add_URI_to_newrelic(){
 	}
 }
 add_action( 'muplugins_loaded', 'wpcom_vip_add_URI_to_newrelic' );
+
+function wpcom_vip_save_query_callback( $query, $elapsed, $backtrace, &$wpdb ) {
+	global $current_user;
```
ethitter: This doesn't get used anywhere that I can tell. Can we remove it?

ethitter:
This looks good to me; I haven't encountered any issues with this on my sandbox.

I left one note about what appears to be an unused global, but otherwise, this is ready to merge.

vip-helpers/vip-utils.php
nickdaugherty: Good catch, removed.

ethitter:
👍

PR #216, opened 2016-06-09
Add extra caching for global comment counts.

emrikol:
When global comment counts get in the millions, the DB requests can take 5+ seconds on each page load.


emrikol:
This will also bring parity with WordPress.com and the global comment caching that it uses.

emrikol:
Closing this in favor of a future more full-fledged performance plugin

PR #220, opened 2016-06-16
Remove the cloning to vip-go-mu-plugins-public

nickdaugherty:
This repo is now public and replaces `vip-go-mu-plugins-public` (which
has been deleted), so we should stop trying to push code there.


ethitter:
👍

PR #226, opened 2016-06-28
Introduce REST API endpoint to list domains for current site

### LINKED ISSUE #225: Add REST endpoint to list subsites of a multisite installation
Following #207, we need a way to list the subsites on a multisite network so we can make requests to their cron endpoints. Using the REST API framework introduced in WP 4.4 seems the way to approach this. The VIP Go API should probably be updated to indicate multisite in the inflated result (issue on that repo forthcoming), which the cron-job trigger will use as an indicator to request a multisite's subsites from this new REST endpoint.

ethitter:
To ensure we can access a list of subsites on a multisite network, we need a REST API endpoint on the multisite network that will list all sites (main and its subsites) on the network. The VIP Go API is only aware of the main site, and in lieu of synching a list of subsites to the API, we'll rely on the WordPress network to report on itself.

For consistency, the response matches that from the VIP Go API. I've also added a response for single-site installs.

The domain name isn't simply pulled from the `wp_blogs` table, as that would bypass any filtering VIP or clients may apply to the domain name. While `switch_to_blog()` doesn't load the full blog context for each subsite, retrieving the `home_url()` while switched to each child blog will, at least, encounter some of the filtering we may add.

WordPress 4.6 should simplify the retrieval of sites in a multisite network thanks to its introduction of `get_sites()` and the `WP_Site_Query` class: https://core.trac.wordpress.org/ticket/35791. In the meantime, we fall back to a DB query that's cached for 30 minutes. Invalidation isn't warranted at this point, given the low number of multisite networks we host, and the extreme infrequency with which they've added new subsites. Should that change before `get_sites()` is in Core, we may consider hooks for invalidations.

Currently, the endpoint has no authentication, and is discoverable through GET requests to the JSON API's endpoints. As the endpoint only lists a domain name, though, I'm not sure if there's a need for any authentication.

Fixes #225.

### DIFF
--- a/vip-rest-api.php
+++ b/vip-rest-api.php
@@ -119,19 +119,22 @@ class WPCOM_VIP_REST_API_Endpoints {
 	 * Consistency simplifies incorporating data from the VIP Go API with this API
 	 */
 	private function format_response( $data ) {
-		$response = array();
-
 		if ( is_wp_error( $data ) ) {
-			$response['status'] = 'error';
-			$response['data']   = false;
+			$response = array(
+				'status' => 'error',
+				'data'   => false,
+			);
 		} else {
-			$response['status'] = 'success';
-			$response['page']      = 1;
-			$response['pagesize']  = 1000;
-			$response['totalrecs'] = count( $data );
-			$response['result']    = $response['totalrecs'];
+			$response = array(
+				'status'    => 'success',
+				'page'      => 1,
+				'pagesize'  => 1000,
+				'totalrecs' => count( $data ),
+				'result'    => null, // placeholder to maintain array order
+				'data'      => $data,
+			);
 
-			$response['data'] = $data;
+			$response['result'] = $response['totalrecs'];
 		}
 
 		return $response;
--- a/vip-rest-api.php
+++ b/vip-rest-api.php
@@ -58,7 +58,33 @@ class WPCOM_VIP_REST_API_Endpoints {
 		$sites = array();
 
 		if ( is_multisite() ) {
-			//
+			if ( function_exists( 'get_sites' ) ) {
+				$_sites = get_sites( array(
+					'public'   => 1,
+					'archived' => 0,
+					'spam'     => 0,
+					'deleted'  => 0,
+					'fields'   => 'ids',
+				) );
+			} else {
+				// Add support for 4.4 and 4.5, as `get_sites()` wasn't introduced until 4.6
+				$_sites = array();
+			}
+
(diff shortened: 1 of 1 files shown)


nickdaugherty:
How much or little sense would it make to make this an endpoint in the Jetpack JSON API module, or WP API?

Obviously those would need to be required on every site, while this is standalone and bypasses that requirement.

Conceptually I like the idea of plugging into existing infrastructure, but I also dislike forcing those plugins to be enabled on every Go site.

ethitter:
This uses the REST API infrastructure that came into Core in 4.4 (https://make.wordpress.org/core/2015/10/28/rest-api-welcome-the-infrastructure-to-core/), so its only dependency is on Core. From 4.4 on, the WP-API plugin only provides endpoints, and there isn't one that provides the data we need anyway, so there's no concern about introducing a new dependency. Since Core provides the REST API framework for us, I don't see any reason to explore using the Jetpack API for this, particularly given that Jetpack isn't always connected.

nickdaugherty:
Thanks for the explanation, I only skimmed the code earlier and missed that it was building on the built-in REST api. Perfect!

vip-rest-api.php
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	/**
+	 * PLUGIN SETUP
+	 */
+
+	/**
+	 * Class properties
+	 */
+	private $namespace = 'wpcom-vip/v1';
```
nickdaugherty: Using `vip` would be a little cleaner IMO.

vip-rest-api.php
```diff
@@ -0,0 +1,148 @@
...
+			);
+		} else {
+			$response = array(
+				'status'    => 'success',
+				'page'      => 1,
+				'pagesize'  => 500,
+				'totalrecs' => count( $data ),
+				'result'    => null, // placeholder to maintain array order
+				'data'      => $data,
+			);
```
nickdaugherty: I think we can forego matching the Go API response format here. We'll be getting rid of that in the future (in a v2), and only the API itself will be interacting with this endpoint (not end clients like a CLI tool or admin console) so it would be better to use a response format that more closely aligns with the rest of the WP REST api.

ethitter: Sure, seems fine. I'll switch it to an array of objects. Site IDs aren't useful to us, and single-sites don't have them, so I'll leave them out.

nickdaugherty:
Looks great.

PR #229, opened 2016-06-30
Purge term archives

simonwheatley:
Comprehensive cache clearance for archives of associated terms
* Get all the taxonomies which apply to a post and loop over them
* Filter to allow devs to change the max number of pages purged from the term archive
* Also purge associated feed URLs

I'm a little uncertain about the filter introduced by https://github.com/Automattic/vip-go-mu-plugins/commit/28389b0f78bd5a45c9926df92bc51bdbebe42f87; I wonder if this is necessary as the `$GLOBALS['wp_rewrite']->pagination_base` is a public property, and I think it's designed for this purpose.

### DIFF
--- a/cache-manager.php
+++ b/cache-manager.php
@@ -202,15 +202,18 @@ class WPCOM_VIP_Cache_Manager {
 		$taxonomies = get_object_taxonomies( $post, 'object' );
 
 		/**
-		 * Allows you to customise the URL suffix used to specify a page for paged term archives
+		 * Allows you to customise the URL suffix used to specify a page for
+		 * paged term archives.
 		 *
-		 * Developers should hook this filter to provide a different page endpoint if they have
-		 * custom or translated rewrite rules for paging in term archives:
+		 * Developers should hook this filter to provide a different page
+		 * endpoint if they have custom or translated rewrite rules for
+		 * paging in term archives:
 		 *
 		 * Standard:     example.com/category/news/page/2
 		 * Non-standard: example.com/category/news/p/2
 		 *
-		 * The string should be formatted as for sprintf, with a `%d` in place of the page number.
+		 * The string should be formatted as for `sprintf`, with a `%d` in place
+		 * of the page number.
 		 *
 		 * @param string sprintf formatted string, including `%d`
 		 * }
--- a/cache-manager.php
+++ b/cache-manager.php
@@ -236,6 +236,9 @@ class WPCOM_VIP_Cache_Manager {
 		 */
 		$max_pages = apply_filters( 'wpcom_vip_cache_purge_urls_max_pages', 5 );
 
+		// Set some limits on max and min values for pages
+		$max_pages = max( 1, min( 20, $max_pages ) );
+
 		foreach ( $taxonomies as $taxonomy ) {
 			if ( true !== $taxonomy->public ) {
 				continue;
--- a/cache-manager.php
+++ b/cache-manager.php
@@ -201,6 +201,21 @@ class WPCOM_VIP_Cache_Manager {
 
 		$taxonomies = get_object_taxonomies( $post, 'object' );
 
+		/**
+		 * Allows you to customise the URL suffix used to specify a page for paged term archives
+		 *
+		 * Developers should hook this filter to provide a different page endpoint if they have
+		 * custom or translated rewrite rules for paging in term archives:
+		 *
+		 * Standard:     example.com/category/news/page/2
+		 * Non-standard: example.com/category/news/p/2
+		 *
+		 * The string should be formatted as for sprintf, with a `%d` in place of the page number.
+		 *
+		 * @param string sprintf formatted string, including `%d`
+		 * }
+		 */
+		$paging_endpoint = apply_filters( 'wpcom_vip_cache_purge_urls_paging_endpoint', $GLOBALS['wp_rewrite']->pagination_base . '/%d/' );
 
 		/**
 		 * The maximum page to purge from each term archive when a post associated with that term is published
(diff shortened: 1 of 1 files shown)


cache-manager.php
```diff
@@ -199,22 +199,68 @@ function queue_post_purge( $post_id ) {
...
+		 * }
+		 */
+		$paging_endpoint = apply_filters( 'wpcom_vip_cache_purge_urls_paging_endpoint', $GLOBALS['wp_rewrite']->pagination_base . '/%d/' );
 
-			foreach ( $tags as $tag )
-				$this->purge_urls[] = home_url( $tag_base . $tag->slug . '/' );
+		/**
+		 * The maximum page to purge from each term archive when a post associated with that term is published
+		 *
+		 * Developers should hook this filter to provide a different page endpoint if they have
```
nickdaugherty: This description has some copy/pasta from the above filter - can omit the stuff about custom rewrite rules as that is NA for the pagination limi

nickdaugherty:
This looks pretty good, and seems to work as expected in my testing. I'd like to get another set of eyes on it before merging, cc @ethitter + @joshbetz.

cache-manager.php
```diff
@@ -199,22 +199,68 @@ function queue_post_purge( $post_id ) {
...
+		 * the following pagination URLs will be purged for the news category archive:
+		 *
+		 * example.com/category/news/
+		 * example.com/category/news/page/2
+		 * example.com/category/news/page/3
+		 *
+		 * @param int The maximum page to purge from each term archive
+		 * }
+		 */
+		$max_pages = apply_filters( 'wpcom_vip_cache_purge_urls_max_pages', 5 );
```
ethitter: We should have a max on this, just in case someone wants to purge 100 pages per taxonomy or something extreme like that. Wrapping the filtered result in a `max( 1, min( 20, apply_filters(...) )` will ensure that the result is valid (positive) and within a range we control.

cache-manager.php
```diff
@@ -199,22 +199,68 @@ function queue_post_purge( $post_id ) {
...
+			if ( true !== $taxonomy->public ) {
+				continue;
+			}
+			$taxonomy_name = $taxonomy->name;
+			$terms = get_the_terms( $post_id, $taxonomy_name );
+			if ( false === $terms ) {
+				continue;
+			}
+			foreach ( $terms as $term ) {
+				if ( ! term_exists( $term->slug, $taxonomy_name ) ) {
```
ethitter: This is unnecessary. If, somehow, `get_the_terms()` returns a term that doesn't exist, we've bigger issues.

cache-manager.php
```diff
@@ -199,22 +199,68 @@ function queue_post_purge( $post_id ) {
...
+		 * }
+		 */
+		$max_pages = apply_filters( 'wpcom_vip_cache_purge_urls_max_pages', 5 );
+
+		foreach ( $taxonomies as $taxonomy ) {
+			if ( true !== $taxonomy->public ) {
+				continue;
+			}
+			$taxonomy_name = $taxonomy->name;
+			$terms = get_the_terms( $post_id, $taxonomy_name );
```
ethitter: One thing we don't have a good way of dealing with here are terms that a post was removed from. We may want to hook into `clean_term_cache` as well, to catch removals.

cache-manager.php
```diff
@@ -199,22 +199,68 @@ function queue_post_purge( $post_id ) {
...
+					continue;
+				}
+				$maybe_purge_url = get_term_link( $term, $taxonomy_name );
+				if ( is_wp_error( $maybe_purge_url ) ) {
+					continue;
+				}
+				if ( $maybe_purge_url && is_string( $maybe_purge_url ) ) {
+					$this->purge_urls[] = $maybe_purge_url;
+					for( $i = 2; $i <= $max_pages; $i++ ) {
+						$this->purge_urls[] = sprintf( trailingslashit( $maybe_purge_url ) . $paging_endpoint, $i );
```
ethitter: We should use `user_trailingslashit()` instead, since we allow sites on Go to use permalink structures that lack trailing slashes (ie, MultilingualPress, and the existence of a permalinks UI). We should also abuse `path_join()` here, rather than adding all of the "is this slashed?!?" logic to our URL building.

ethitter: Clarifying the above, the full URL should be passed through `user_trailingslashit()` after it's joined using `path_join()`, to ensure that the final result matches what Core expects for trailing slashes.

ethitter:
Really solid start. I left a few notes inline.

cache-manager.php
simonwheatley: Tracking that in https://github.com/Automattic/vip-go-mu-plugins/issues/231

cache-manager.php
simonwheatley: @ethitter I haven't used `path_join()` because as you point out, this is not quite its intended purpose, which makes me nervous about unintended consequences in any changes down the line, and the code is clean enough without it, IMO.

``` php
$maybe_purge_url_page = rtrim( $maybe_purge_url, '/' ) . '/' . ltrim( $paging_endpoint, '/' );
```

cache-manager.php
simonwheatley: Addressed in https://github.com/Automattic/vip-go-mu-plugins/pull/229/commits/5bf7a5d41040dee687c135fdf172f099dbcab6cf

cache-manager.php
simonwheatley: Addressed in https://github.com/Automattic/vip-go-mu-plugins/pull/229/commits/954635f153ed9fd0581d1cd1448293fb0f454a84

ethitter:
Looks good!

Issue #231, opened 2016-07-01
When terms are removed from a post, purge the associated term archives


simonwheatley:
> One thing we don't have a good way of dealing with here are terms that a post was removed from. We may want to hook into `clean_term_cache` as well, to catch removals.
> – @ethitter ([source](https://github.com/Automattic/vip-go-mu-plugins/pull/229#discussion_r69234575))

PR #232, opened 2016-07-01
Purge removed term archives

simonwheatley:
**Should be merged after #231**

Hook `clean_term_cache` and purge URLs associated with the terms.

### DIFF
--- a/cache-manager.php
+++ b/cache-manager.php
@@ -28,6 +28,7 @@ class WPCOM_VIP_Cache_Manager {
 		}
 
 		add_action( 'clean_post_cache', array( $this, 'queue_post_purge' ) );
+		add_action( 'clean_term_cache', array( $this, 'queue_term_purge' ), 10, 3 );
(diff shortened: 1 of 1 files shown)


cache-manager.php
```diff
@@ -252,6 +255,98 @@ function queue_post_purge( $post_id ) {
...
+		 * news category archive:
+		 *
+		 * example.com/category/news/
+		 * example.com/category/news/page/2
+		 * example.com/category/news/page/3
+		 *
+		 * @param int The maximum page to purge from each term archive
+		 * }
+		 */
+		$max_pages = apply_filters( 'wpcom_vip_cache_purge_urls_max_pages', 5 );
```
ethitter: This should receive the same range restriction we apply to post-type archives.

simonwheatley: Yes, yes it should. Bad copy/pasta. Addressed in https://github.com/Automattic/vip-go-mu-plugins/pull/232/commits/973c32b67dbe6677e8641a6417037204e47fbefe

Thanks!

PR #234, opened 2016-07-02
Return full result of `home_url()` in REST API endpoint, as path is important

ethitter:
As we only support subdirectory multisite at the moment, we need the full URL that `home_url()` returns.

Currently, endpoint will simply return multiple of the same entry for whatever the main site is, omitting the subdirectory we should be checking.


nickdaugherty:
Another approach would be to return the path separately, so return a `domain_name` and `path` property. This would keep `domain_name` strictly meaning just that - the domain, and allow us to differentiate on the path. Thoughts?

Alternatively, an `url` property for the entire url.

ethitter:
@nickdaugherty, I'm not opposed to making those changes, but for now, this change fixes the immediate issue without modifying the jobs code that triggers cron. As we're discussing changes to that code to deal with concurrency and long-running requests, I'd prefer to make this change now and hold off on more significant changes until it's clear what's needed.

PR #240, opened 2016-07-10
Purge caches on changed meta

### LINKED ISSUE #239: Purge caches when term or post meta changes
If post meta is changed outside a post being changed, the related caches are not being purged. We should purge all caches for a post, related term archive URLs, etc, when post meta is changed.

When term meta is changed, we should change all term archive URLs.

simonwheatley:
Fixes #239


simonwheatley:
@ethitter Would you please review this?

simonwheatley:
Closing in favour of #247

PR #244, opened 2016-07-13
Updating wp-parsely to 1.9

vaurdan:
Last week `wp-parsely` was updated in WordPress.com to version 1.9. 

I'm now merging that update with VIP Go.


ethitter:
👍

PR #257, opened 2016-07-27
Update/qm and debug bar query saver

nickdaugherty:
Moving the custom query collector to a lower level (db-config.php), and only using the custom panels if the collector is present, so that the debug plugins work in dev environments

Adds custom collector for Query Monitor db queries, to support the VIP style saved queries


ethitter:
Looks good

PR #269, opened 2016-08-15
Default to no pingbacks

simonwheatley:
Hooks `pre_ping` to stop any pinging from happening, unless `WPCOM_VIP_DO_PINGS` is set to `true`

### DIFF
--- a/misc.php
+++ b/misc.php
@@ -79,3 +79,19 @@ add_action( 'update_option', function( $option ) {
 		}
 	}
 }, 10, 1 );
+
+/**
+ * Hooks pre_ping to stop any pinging from happening,
+ * unless `WPCOM_VIP_DO_PINGS` is set to `true`.
+ *
+ * @param array $post_links The URLs to be pinged
+ */
+function wpcom_vip_pre_ping( $post_links ) {
+	if ( ! defined( 'WPCOM_VIP_DO_PINGS' ) || true !== WPCOM_VIP_DO_PINGS ) {
+		return;
+	}
+	// Clear our the post links array, so we ping nothing
+	$post_links = array();
+}
+add_action( 'pre_ping', 'wpcom_vip_pre_ping' );
+
--- a/misc.php
+++ b/misc.php
@@ -82,7 +82,7 @@ add_action( 'update_option', function( $option ) {
 
 /**
  * Hooks pre_ping to stop any pinging from happening,
- * unless `WPCOM_VIP_DO_PINGS` is set to `true`.
+ * unless `WPCOM_VIP_DO_PINGS` is set to `true` (boolean).
  *
  * @param array $post_links The URLs to be pinged
  */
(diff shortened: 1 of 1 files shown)


misc.php
```diff
@@ -79,3 +79,19 @@ function action_wpcom_vip_verify_string() {
...
+
+/**
+ * Hooks pre_ping to stop any pinging from happening,
+ * unless `WPCOM_VIP_DO_PINGS` is set to `true`.
+ *
+ * @param array $post_links The URLs to be pinged
+ */
+function wpcom_vip_pre_ping( $post_links ) {
+	if ( ! defined( 'WPCOM_VIP_DO_PINGS' ) || true !== WPCOM_VIP_DO_PINGS ) {
+		return;
```
nickdaugherty: We want to empty out the array here, when `WPCOM_VIP_DO_PINGS` is _not_ `true`. Then below, return `$post_links`. 

Currently, this will not work since we're returning void here, and then clearing the array and returning nothing down below, so it effectively disables pinging even when the constant is set to `true`.

nickdaugherty:
@simonwheatley the logic needs some work, left a note.

misc.php
nickdaugherty: Scratch what I said about returning void, I looked at the actual call to this action and it's not expecting a return value, :facepalm:.

The logic is still backwards - we want to return without modifying the `$post_links` if the constant _is_ true, not if it's not true.

simonwheatley: Addressed in https://github.com/Automattic/vip-go-mu-plugins/pull/269/commits/0f66d853a3546c79118b56fa27f4bdbd330f459a

simonwheatley:
Thanks, @nickdaugherty. I think I've clarified and fixed that logic.

misc.php
```diff
@@ -79,3 +79,20 @@ function action_wpcom_vip_verify_string() {
...
+
+/**
+ * Hooks pre_ping to stop any pinging from happening,
+ * unless `WPCOM_VIP_DO_PINGS` is set to `true` (boolean).
+ *
+ * @param array $post_links The URLs to be pinged
+ */
+function wpcom_vip_pre_ping( $post_links ) {
+	$do_pings = ( defined( 'VIP_DO_PINGS' ) && true === WPCOM_VIP_DO_PINGS );
+	if ( ! $do_pings ) {
```
nickdaugherty: This conditional is still backwards. If we _are_ going to do pings - if `$do_pings` is `true`, then we want to `return` without modifying anything.

If we _aren't_ going to run pings, then we want to clear out the array, _not_ `return`.

nickdaugherty:
@simonwheatley left some more feedback, and looks like there is a conflict now, d'oh.

simonwheatley:
Thanks, @nickdaugherty. Merge conflict resolved, and messed up logic straightened.

nickdaugherty:
Looks good!

PR #277, opened 2016-08-17
Use wp_mail_from filter, instead of setting $phpmailer->Sender directly

nickdaugherty:
WP 4.6 adds some improvements to how it handles the 'from' address, so
by using the filter we take advantage of that and use the 'native' way

### DIFF
--- a/vip-mail.php
+++ b/vip-mail.php
@@ -12,13 +12,14 @@ class VIP_SMTP {
 	function init() {
 		add_action( 'phpmailer_init',    array( $this, 'phpmailer_init' ) );
 		add_action( 'bp_phpmailer_init', array( $this, 'phpmailer_init' ) );
+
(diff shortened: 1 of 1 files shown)


vip-mail.php
```diff
@@ -12,13 +12,14 @@ class VIP_SMTP {
 	function init() {
 		add_action( 'phpmailer_init',    array( $this, 'phpmailer_init' ) );
 		add_action( 'bp_phpmailer_init', array( $this, 'phpmailer_init' ) );
+
+		add_filter( 'wp_mail_from', array( $this, 'filter_wp_mail_from' ) );
```
kraftbj: What about setting this to priority 1 so it hits earlier, so other code can still override without worrying about priorities?

nickdaugherty: Very good idea, will implement that.

PR #278, opened 2016-08-18
Set the wp_mail_from priority to 1

nickdaugherty:
Will better allow other code to override our filter without changing
priority

props @kraftbj


ethitter:
👍

PR #280, opened 2016-08-19
On add|update|delete of autoloaded option, clear alloptions cache

nickdaugherty:
The DB gets updated properly, but the cache gets fubar'd because old
data can overwrite new in concurrent environments

See https://core.trac.wordpress.org/ticket/31245

And specifically https://core.trac.wordpress.org/ticket/31245#comment:19 and https://core.trac.wordpress.org/ticket/31245#comment:26 (fix comes from the latter)


ethitter:
👍

PR #284, opened 2016-08-23
After add|update|delete_option, also conditionally clear notoptions cache

nickdaugherty:
notoptions is subject to the same race condition that alloptions is, so
we need to do the same cache clearing here.

This is slightly different in that we copy the 'notoptions' array very
early in script execution, and use that instead of the value in cache,
because the value in cache gets updated when add|update_option runs,
which makes it impossible to then check `isset( $notoptions[ $option ]
)` for conditional clearing.

If we didn't store a copy of the notoptions variable, we'd have to
clear that cache on every add|update, which is wasteful

WIP - still needs testing

### DIFF
--- a/misc.php
+++ b/misc.php
@@ -65,6 +65,22 @@ add_action( 'template_redirect', 'action_wpcom_vip_verify_string' );
 add_action( 'pre_amp_render_post', 'wpcom_vip_disable_new_relic_js' );
 
 /**
+ * Store a copy of the 'notoptions' cache value, so we know which ones we need
+ * to clear the cache for after add|update_option
+ *
+ * This is because add|update_option removes the entry from the notoptions local
+ * and memcached array, which means it's no longer available for testing via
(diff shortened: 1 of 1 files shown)


nickdaugherty:
This PR doesn't actually work, going to go a different route.

GaryJones:
@mjangda @nickdaugherty Just had a client caught by this (correct value in DB, missing value in object cache) - did an implementation of this `notoptions` get added somewhere? It's labelled as deployed, but the only reference in code appears to be [this](https://github.com/Automattic/vip-go-mu-plugins/commit/e70da5ee38cf72579d8195f0937ee7b5d327bd7b) which clears `notoptions` after Jetpack is connected.

nickdaugherty:
`notoptions` hasn't yet been addressed, it's a harder problem (also one that hasn't really come up lately).

GaryJones:
Removing the label, as that is incorrect.

PR #285, opened 2016-08-24
Rewrite Rules Inspector: prevent resetting permalink structure

david-binda:
…to default in case the site is not setting permalink structure via code, but rather via admin screen

Calling the `wpcom_vip_refresh_wp_rewrite` function in case none of the helper functions for setting permalinks via theme's code, eg.: `wpcom_vip_load_permastruct`, is used leads to resetting the permalink structure to the default WordPress one.

All of the helper functions for setting a custom permalink structure set a constant which can be used for the necessary check inside the Rewrite Rules Inspector's custom hook which is calling the `wpcom_vip_refresh_wp_rewrite`.

This commit adds the conditional, taking advantage of mentioned constans, and only runs the `wpcom_vip_refresh_wp_rewrite` if the permalink structure is being set via code.


ethitter:
Good catch!

PR #293, opened 2016-09-02
Add Jetpack Force 2FA

joshbetz:
In https://github.com/Automattic/jetpack-force-2fa/pull/11, we only load
the plugin if the necessary module is active. This means we can load it
as an mu-plugin without worrying about breaking things.

This also disables the standard admin notice since it's mostly
unnecessary on VIP Go. In the future we will enable Jetpack SSO in code.
If for some reason it's not enabled, it will be an explicit choice by
the site owner.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -34,3 +34,6 @@
 [submodule "jetpack-beta"]
 	path = jetpack-beta
 	url = git@github.com:Automattic/jetpack.git
+[submodule "jetpack-force-2fa"]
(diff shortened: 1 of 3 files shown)


joshbetz:
@nickdaugherty @ethitter Thoughts on this? It forces admins to use 2fa if Jetpack SSO is in use.

nickdaugherty:
Looks good to me, :shipit: 

Can you fix the conflict?

PR #294, opened 2016-09-06
Add 000-local-env

joshbetz:


000-local-env.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+if ( defined( 'VIP_GO_ENV' ) && VIP_GO_ENV ) {
```
nickdaugherty: I think you meant to use the `WPCOM_IS_VIP_ENV` constant? `VIP_GO_ENV` tracks the environment name of the site - production/staging/development. That could (and probably should) exist on local environments

joshbetz: @nickdaugherty https://github.com/Automattic/vip-go-mu-plugins/blob/master/000-vip-init.php#L6-L8

Which I'm now realizing will load after this...

nickdaugherty: Let's move this file to `vip-helpers/vip-local.php` and conditionally include it from `000-vip-init.php`.

I think we should still use `WPCOM_IS_VIP_ENV` for this check, and let `VIP_GO_ENV` track the 'environment name', rather than use it to determine where the code is running. 

Currently `VIP_GO_ENV` is set to `false` in dev, but IMO it should be set to something like `local-development`. The environment name constant specifies which 'version' of the site the code is running on, vs. _where_ the code is physically running (Go vs local).

joshbetz: > I think we should still use WPCOM_IS_VIP_ENV for this check

We'd also have to fix `WPCOM_IS_VIP_ENV` because it currently always returns true. It's flagged as deprecated and only defined for back-compat, so I think this was on purpose.

> Currently VIP_GO_ENV is set to false in dev, but IMO it should be set to something like local-development.

Hmm... I think this is right. The idea is that it returns false if it's not a VIP Go Environment.

> The environment name constant specifies which 'version' of the site the code is running on, vs. where the code is physically running (Go vs local).

It really just specifies the environment name or `false` if it's not a VIP Go environment.

nickdaugherty: > Hmm... I think this is right. The idea is that it returns false if it's not a VIP Go Environment.

In a perfect world, there is little difference between running code on Go production and on N local environments. Those local environments should also be able to specify what their 'environmen name' is - like 'QA-12345' or 'nicks-local-docker-install'.

If we use this constant as specifying if it's running on Go hardware or not, we can't distinguish different local environments from one-another, and we already have a constant that was made to tell you if you are running on production hardware or not, which is `WPCOM_IS_VIP_ENV`.

We use that exact name for backwards compatibility with dotcom code, but it's still important to specify the _where_ (Go hardware vs other) and the _what_ (what 'version' of the site this is) separately.

PR #297, opened 2016-09-10
CLI command to convert `utf8` tables to `utf8mb4`

ethitter:
Modeled on Core's `upgrade_430()` and `upgrade_network()` functions, and leveraging a clone of its `maybe_convert_table_to_utf8mb4()` function, this CLI command will convert an installation's eligible `utf8` tables to `utf8mb4`.

### DIFF
--- a/wp-cli.php
+++ b/wp-cli.php
@@ -5,3 +5,4 @@ if ( ! defined( 'WP_CLI' ) || ! WP_CLI ) {
 }
 
 require __DIR__ . '/wp-cli/vip-fixers.php';
+require __DIR__ . '/wp-cli/vip-utf8-to-utf8mb4.php';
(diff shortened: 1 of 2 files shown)


nickdaugherty:
This looks good to me, though I'd like a second opinion from someone more familiar with collation, if possible.

ethitter:
Something to note for a future date: BuddyPress isn't upgraded yet, as the plugin hasn't finalized its conversion process: https://buddypress.trac.wordpress.org/ticket/6346.

ethitter:
I started writing this comment about bbPress, before remembering that bbPress 2 uses CPTs and CTs, rather than custom tables, so there's nothing to worry about. 😄

ethitter:
Merging now that I've run this against a half-dozen sites without incident.

PR #301, opened 2016-09-21
New Device Notifications on Go incorrectly exclude super admins

cfg:
By default [NewDeviceNotifications](https://github.com/Automattic/vip-go-mu-plugins/blob/master/shared-plugins/new-device-notification/new-device-notification.php#L53
) excludes super admins from NDN notifications.

`$is_privileged_user = ! is_super_admin() && current_user_can( 'edit_posts' );`

This was presumably copied wholesale from WPCOM where it excluded A8C employees, and on VIP Go it will exclude NDN for all admin users which isn't the original intent, and undoes much of the benefit of NDN.

### DIFF
--- a/shared-plugins/new-device-notification/new-device-notification.php
+++ b/shared-plugins/new-device-notification/new-device-notification.php
@@ -48,11 +48,11 @@ class New_Device_Notification {
 		get_currentuserinfo();
 
 		// By default, users to skip:
-		// * Super admins (Automattic employees visiting your site)
 		// * Users who don't have /wp-admin/ access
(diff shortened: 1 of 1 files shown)


joshbetz:
We probably still need a way to exclude Automatticians or we'll end up with a lot of false positives.

cfg:
How do you currently identify Automatticians on Go? By more than just IP? Can you use the VIP Support user role? Either way the current config is incorrect; admin users on all Go sites are exposed by default.

joshbetz:
Closing in favor of #306

PR #302, opened 2016-09-22
Support multisite in wp imports

nickdaugherty:
Needs the `/sites/ . get_current_blog_id()` code from
`save_image_file()` to properly generate urls when downloading the
import file

props @emrikol - he did all the work, I just copied his 'screen shot' :)


ethitter:
👍

PR #304, opened 2016-09-26
Cache Purging: More sensible purge limits

joshbetz:
Trying to purge too many URLs at once slows down post saving and will eventually cause post save actions to fail. We can be a bit more conservative with the number of pages that we purge. Luckily, this method of manually sending purge requests to each cache node is temporary and the new method should be much more robust.


nickdaugherty:
Going to merge this - I don't think we have anyone using a higher number than this right now, and it's sensible.

PR #305, opened 2016-09-27
Conditionally load debug-bar.php plugin

david-binda:
Since client sites may include Debug Bar in their repository and load it as a plugin, we should not attempt to load it again if it's already present.

This commit fixes fatal errors for Automatticians in case Debug Bar plugin is present and being loaded from client repository by checking if `class_exists( 'Debug Bar' )` before requiring the plugin.

It also still loads the WordPress.com VIP custom panels.

### DIFF
--- a/debug-bar.php
+++ b/debug-bar.php
@@ -21,7 +21,9 @@ add_action( 'init', function() {
 		return;
 	}
 
-	require_once( __DIR__ . '/debug-bar/debug-bar.php' );
+	if ( ! class_exists( 'Debug_Bar' ) ) {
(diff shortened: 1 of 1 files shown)


nickdaugherty:
The only real downside to letting the client control Debug Bar is versioning. It's possible we'll run into situations where they are loading an old version of Debug Bar that doesn't support some feature one of our custom panels does, and we don't find out until a critical moment when the site is down.

nickdaugherty:
The better way to approach this is have us control the Debug Bar plugin, and make it available to clients.

david-binda:
Thanks for the input @nickdaugherty ! Let's close this PR then.

PR #306, opened 2016-09-28
Make NDN compatible with VIP Go

emrikol:
There are three main changes that _need_ to be done for VIP Go compatibility:

1. `is_super_admin()` needs replaced to better work with single site installations/non-WordPress.com multisite
2. The cookie domain needs set to the primary domain
3. Until `ip2location()` is available on VIP Go, it shouldn't be used

Other things that should be done specific for VIP Go, but can be done somewhere else in mu-plugins

1. The email message should be reworded
2. The email "from" addresses seems to be hardcoded to "noreply@wordpress.com" on VIP Go

----

**Update Friday 30 September 2016**

Checklist for merge:

* [x] `is_super_admin()` needs replaced to better work with single site installations/non-WordPress.com multisite
* [x] The cookie domain needs set to the primary domain
* [x] Until `ip2location()` is available on VIP Go, it shouldn't be used
* [x] The email message should be reworded
* [x] The email "from" addresses seems to be hardcoded to "noreply@wordpress.com" on VIP Go
* [ ] WordPress.com functionality needs testing before it's synced across


joshbetz:
Before we can ship this, we need to add a proxy check back to `is_automattician()`, but otherwise looks good.

shared-plugins/new-device-notification/new-device-notification.php
```diff
@@ -168,30 +168,32 @@ public function ip_to_city( $ip ) {
 		if ( ! function_exists( 'ip2location' ) ) {
 			$location = new stdClass();
```
cfg: Can you use `VIP_Go_Geo_Uniques::get_country_code()` for `$location->country_short`? Also the location object is passed to the `ndn_send_email` and `ndn_notify_of_new_device` hooks, so for better compatibility it would be good to set all the same properties as [wpcom-geo-uniques](https://vip-svn.wordpress.com/plugins/wpcom-geo-uniques/wpcom-geo-uniques.php) does, e.g.:

```
$location                = new stdClass;
$location->latitude      = '43.6525';
$location->longitude     = '-79.381667';
$location->country_short = 'CA';
$location->country_long  = 'CANADA';
$location->region        = 'ONTARIO';
$location->city          = 'TORONTO';

```

joshbetz: We could set the country code, but we currently don't have any of the other details available on VIP Go.

simonwheatley:
Added a checklist for merging this, and getting it shipped.

cfg:
With the last few commits it seems like the version of the plugin on Go is going to be completely different than the version on WPCOM? What about adding filters to the WPCOM version, and hooking them on Go? All that's missing would be new filters for the email headers, and the value returned from `ip2city()`.

simonwheatley:
> With the last few commits it seems like the version of the plugin on Go is going to be completely different than the version on WPCOM?

Are you listening to our P2 conversations, @cfg? We're just discussing exactly that 😁 

The changes are relatively minor right now, so currently not too bad to keep in sync for a plugin of this size, but I'm now looking to abstract away from the platform(s) a bit.

simonwheatley:
Needs a check through for more VIP Go or WordPress.com specific stuff, e.g. @cfg's [comment here](https://github.com/Automattic/vip-go-mu-plugins/pull/306#discussion_r81810185)…

shared-plugins/new-device-notification/new-device-notification.php
```diff
@@ -105,12 +105,17 @@ public function set_cookie() {
 			$parts = parse_url( home_url() );
 			setcookie( $this->cookie_name, $this->cookie_hash, $tenyrsfromnow, COOKIEPATH, $parts['host'], false, true );
 		}
+
+		// Is this a VIP Go mapped domain?
+		if ( defined( 'VIP_GO_ENV' ) ) {
```
simonwheatley: The value of `VIP_GO_ENV` [can be false](https://github.com/Automattic/vip-go-mu-plugins/blob/16a271f3a518e6a95df7d730ce844acc3b627781/000-vip-init.php#L3-L8), so the check just for it being defined is adequate. Thanks for checking.

shared-plugins/new-device-notification/vipgo-helper.php
```diff
@@ -0,0 +1,317 @@
...
+			'EH' => 'Western Sahara',
+			'YE' => 'Yemen',
+			'ZM' => 'Zambia',
+			'ZW' => 'Zimbabwe',
+		);
+		if ( isset( $longs[ $code ] ) ) {
+			return $longs[ $code ];
+		}
+
+		return '';
```
joshbetz: `return '¯\_(ツ)_/¯'`? :)

joshbetz:
@simonwheatley @emrikol Looks good to me.

shared-plugins/new-device-notification/vipgo-helper.php
simonwheatley: LOL

PR #310, opened 2016-10-04
Purge attachment pages and files when attachments are updated

ethitter:
Since attachments use the `inherit` post status, their updates do not currently trigger Varnish purges because we ignore that status. We'd also left out the file URL itself, so even if purging had applied to attachments, only the attachment pages would've been cleared.


nickdaugherty:
Looks great.

PR #312, opened 2016-10-10
Sync do_pings hack from wp.com

joshbetz:
In some cases, we use this to disable do_pings as it can be quite slow. This can be enabled on a per-site basis by setting `VIP_CUSTOM_PINGS`. By default trackbacks are disabled since they're the slowest, but there are filters to enable/disable the other actions as well.


ethitter:
👍

PR #313, opened 2016-10-10
Add .gitignore'd directory for defining secrets.

joshbetz:


000-vip-init.php
```diff
@@ -1,5 +1,9 @@
 <?php
 
+if ( file_exists( __DIR__ . '/.secrets/vip-go-secrets.php' ) ) {
+	require __DIR__ . '/.secrets/vip-go-secrets.php';
```
nickdaugherty: Small tweak, can we drop the `go-` from this so it's just `vip-secrets.php`? Just trying to be a little more generic and forward looking.

PR #315, opened 2016-10-11
Add is_proxied_automattician()

### LINKED ISSUE #314: Update is_automattican requirements
We have a proxy check now. I think we should remove the email verification step and either:

* Add a proxy check to `is_automattican()`; or
* Add an `is_proxied_automattician()` function

I'm leaning toward adding a proxy check directly to `is_automattician()`, but could maybe be convinced otherwise.

joshbetz:
Now that we can make a proxy check, we should add a function which
checks if the current user is proxied. Instead of adding this directly
to is_automattician() like I originally thought, we're introducing a new
function because is_automattician() can also check other users.

Fixes #314

### DIFF
--- a/000-vip-init.php
+++ b/000-vip-init.php
@@ -4,6 +4,10 @@ if ( file_exists( __DIR__ . '/.secrets/vip-secrets.php' ) ) {
 	require __DIR__ . '/.secrets/vip-secrets.php';
 }
 
+if ( ! defined( 'A8C_PROXIED_REQUEST' ) ) {
+	define( 'A8C_PROXIED_REQUEST', false );
+}
+
 /**
  * @constant VIP_GO_ENV The name of the current VIP Go environment. Falls back to `false`.
  */
--- a/vip-helpers/vip-utils.php
+++ b/vip-helpers/vip-utils.php
@@ -1079,6 +1079,8 @@ function _wpcom_vip_include_plugin( $file ) {
  * It's possible to fake that data (it's just meta and user_email), so don't use this
  * for protecting sensitive info or performing sensitive tasks
  *
+ * NOTE: This does not guarantee the current user is proxied. Use is_proxied_automattician() for that
+ *
  * @param int The WP User id
  * @return bool Bool indicating if user is an Automattician
  */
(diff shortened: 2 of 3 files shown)


000-vip-init.php
```diff
@@ -4,6 +4,10 @@
 	require __DIR__ . '/.secrets/vip-secrets.php';
 }
 
+if ( ! defined( 'A8C_PROXIED_REQUEST' ) ) {
+	define( 'A8C_PROXIED_REQUEST', false );
```
nickdaugherty: I'm worried that it's technically possible to define this constant in the user space.

What prevents a user from adding `define( 'A8C_PROXIED_REQUEST', true )` in their `vip-config.php`?

vip-helpers/vip-utils.php
```diff
@@ -1098,15 +1098,22 @@ function is_automattician( $user_id = false ) {
 		return false;
 	}
 
-	// $vip_support = WPCOM_VIP_Support_User::init();
-
-	if ( WPCOM_VIP_Support_User::is_verified_automattician( $user->ID ) ) {
-		return true;
+	if ( A8C_PROXIED_REQUEST && WPCOM_VIP_Support_User::is_a8c_email( $user->user_email ) ||
```
nickdaugherty: Mixing an 'automattician' check and a proxied 'bypass' here seems a little messy. What's the benefit to `is_automattician()` returning true if you're not 'verified', but are proxied?

I think if we want to take the proxy status into account, we should be using `is_proxied_automattician()`, and `is_automattician()` strictly checks if this user is really an Automattician. As is, this introduces some variability into what actually constitutes an Automattician to the system, which is harder to grok. IMO they should be separate concerns unless we require you to be proxied for `is_automattician()` to return `true` in all cases, which I don't think we want.

000-vip-init.php
joshbetz: @nickdaugherty Gooood question. In production, it is always defined in `vip-secrets.php`, so I think you'd generate a fatal if you try to set it in a client repo.

joshbetz:
@nickdaugherty fixed

PR #319, opened 2016-10-12
Updating wpcom_vip_wp_oembed_helper()

emrikol:
`wpcom_vip_wp_oembed_helper()` was recently updated on WordPress.com to allow the cache to expire, otherwise old and/or unwanted data might stay in it forever


mjangda:
r85569-deploy

PR #322, opened 2016-10-14
Update VIP Dashboard to 2.0.4

scottsweb:
This should give us the email sending fixes, update the welcome widget and remove the maintenance mode plugin from the plugins ui. Also includes lots of code fixes and refactoring.


nickdaugherty:
Turns out that vip-dashboard still needs some work, closing.

PR #325, opened 2016-10-17
Add Jetpack Force 2FA

joshbetz:
Forces 2FA for admins if Jetpack SSO is enabled.

See https://github.com/Automattic/vip-go-mu-plugins/pull/293


mjangda:
Filter deployed in r85554-deploy.

The submodule still needs to be deployed but needs to be moved out of `shared-plugins` first.

PR #329, opened 2016-10-20
Cron Parallelization

ethitter:
Rather than executing scheduled cron tasks in series, multiple tasks can run simultaneously, preventing long-running tasks from blocking other events. 

**How**
1. An endpoint provides a list of tasks to run in the next 60 seconds. Enough information about each is provided to identify it, but its specific arguments are not exposed.
2. A second endpoint accepts the JSON object for an individual task. After confirming that it isn't scheduled more than 15 seconds ahead of the current time, the endpoint runs that job, mimicking the environment Core provides.
3. The existing runners on WordPress.com are updated to use these two endpoints. For each cron task to run, an individual job will be queued on WordPress.com's jobs system.

_All_ attempts to execute Core's native cron methods (`wp-cron.php`, `wp_cron()`, `spawn_cron()`) are blocked to prevent rogue execution.

**Remaining issues:**
- [x] Confirm all appropriate constants and PHP runtime calls are made
- [x] Scheduled-post validation
- [x] Force-publishing for missed schedules
- [ ] Request validation via IP
- [x] Limit concurrency
- [ ] Proper endpoint authentication

**Considerations**
1. A task could run up to 15 seconds early; this seems better than it running late. Returning the next 60-seconds' worth of jobs, but only running those scheduled for the next 15 seconds, is designed to compensate for delays in the process of requesting tasks and scheduling their runners.
2. The name of each job's action is exposed in the endpoint. Is this a concern? If so, the names can be hashed server-side.
3. If the plugin creates jobs, these shouldn't diminish the portion of `$batch_size` that are dedicated to client jobs.


ethitter:
This sets us up for other things like the second endpoint triggering a job via a host action and CLI command, too.

ethitter:
> The name of each job's action is exposed in the endpoint. Is this a concern? If so, the names can be hashed server-side.

Resolved in 213747e. I was really uncomfortable passing around the unhashed action names.

joshbetz:
Is there a way to set the max number of parallel jobs? Similarly, does this create the possibility to use all available PHP processes for long-running cron?

ethitter:
> Is there a way to set the max number of parallel jobs?

Not directly, but it's needed. It currently limits how many jobs it'll try to run in a given execution, but that does nothing to deal with long-running jobs. I'll work on limiting concurrency as well.

> Similarly, does this create the possibility to use all available PHP processes for long-running cron?

Presently, yes. Is that a problem? 👿 

I kid, I kid.

ethitter:
IP-validation presents an interesting challenge thanks to Docker. The approach used on WordPress.com relies on a command that accesses something that's only available on the VIP Go hosts, so we have to deal somehow with bridging the gap between host and container.

ethitter:
Corresponding jobs updates: https://code.a8c.com/D3227.

ethitter:
I'm considering punting "Proper endpoint authentication" to a future issue. Whatever we implement needs to be a shared approach, akin to what we have now, as we don't want to be in a situation of needing to authenticate each new site to use the new cron approach.

ethitter:
Hooks are in place to ensure parity in alerting and stats for force-publishing, but the relevant callbacks still need to be written. I don't really want to muddy the plugin's code with those a8c-specific things, and `misc.php` doesn't seem appropriate, so perhaps a `stats.php` is warranted.

cron.php
```diff
@@ -0,0 +1,464 @@
...
+			define( 'DOING_CRON', true );
+
+			// Remove the event, and reschedule if desired
+			// Follows pattern Core uses in wp-cron.php
+			if ( false !== $event['schedule'] ) {
+				$reschedule_args = array( $event['timestamp'], $event['schedule'], $event['action'], $event['args'] );
+				call_user_func_array( 'wp_reschedule_event', $reschedule_args );
+			}
+
+			wp_unschedule_event( $event['timestamp'], $event['action'], $event['args'] );
```
mjangda: I wonder if there's a possible race condition here: two jobs trigger at the exact same time and which means that `wp_unschedule_event` for both is triggered at the same time for both as well. One of them will win and likely overwrite the other (since all cron is stored in a single option entry)?

cron.php
```diff
@@ -0,0 +1,464 @@
...
+ Description: Take control of wp-cron execution.
+ Author: Erick Hitter, Automattic
+ Version: 0.1
+ Text Domain: wp-cron-control-revisited
+ */
+
+// Enable in production only for specific sites
+// Otherwise, class WP-Cron Control is loaded
+$whitelisted_sites = array();
+if ( VIP_GO_ENV && ! in_array( FILES_CLIENT_SITE_ID, $whitelisted_sites ) ) {
```
ethitter: We check and define `VIP_GO_ENV` in `000-init.php` (https://github.com/Automattic/vip-go-mu-plugins/blob/cc326452d8745063ffef8032ab46de9b6ef7cc8e/000-vip-init.php#L14), so we're safe. 

As for the whitelist, that's meant to be temporary, but also can't be moved without moving the new plugin into a subdirectory. I'd prefer to do that only if I also break the plugin down into smaller pieces as you've suggested; I'm still pondering that.

cron.php
```diff
@@ -0,0 +1,464 @@
...
+	 * Class properties
+	 */
+	private $namespace = 'wp-cron-control-revisited/v1';
+	private $secret    = null;
+
+	private $job_queue_size                  = 10;
+	private $job_queue_window_in_seconds     = 60;
+	private $job_execution_buffer_in_seconds = 15;
+	private $job_timeout_in_minutes          = 10;
+	private $job_concurrency_limit           = 10;
```
ethitter: Initially, I'd thought about making these filterable, or something that could be controlled via the API requests, but at this point, I'm not sure there's a value in doing so. I'm going to leave this for now in case testing shows that we need to vary these settings based on site size/traffic/whatever the 🎱  might show.

cron.php
```diff
@@ -0,0 +1,464 @@
...
+
+	/**
+	 * Execute a specific event
+	 */
+	public function run_event( $request ) {
+		// Parse request
+		$event     = $request->get_json_params();
+		$timestamp = isset( $event['timestamp'] ) ? absint( $event['timestamp'] ) : null;
+		$action    = isset( $event['action'] ) ? trim( sanitize_text_field( $event['action'] ) ) : null;
+		$instance  = isset( $event['instance'] ) ? trim( sanitize_text_field( $event['instance'] ) ) : null;
```
ethitter: `$index` will work. To differentiate jobs with the same timestamp and action, WP passes the event's arguments through `md5` and uses that to index the cron array, which is how I arrived at `$instance`.

cron.php
```diff
@@ -0,0 +1,464 @@
...
+		}
+
+		// Limit batch size to avoid resource exhaustion
+		if ( count( $current_events ) > $this->job_queue_size ) {
+			$current_events = array_slice( $current_events, 0, $this->job_queue_size );
+		}
+
+		return rest_ensure_response( array(
+			'events'          => $current_events,
+			'internal_events' => $internal_events,
```
ethitter: I can't decide: 2a0588a. 😂 

At one point, I thought it would be useful for the WordPress.com jobs to be able to differentiate between the two event types, but now that I've updated that side of things, it's proven unnecessary. I'll nix it as it also simplifies everything.

cron.php
```diff
@@ -0,0 +1,464 @@
...
+					// There are some jobs we never care to run
+					if ( $this->is_blocked_event( $action ) ) {
+						wp_unschedule_event( $timestamp, $action, $instance_args['args'] );
+						continue;
+					}
+
+					// Queue internal events separately to avoid them being blocked
+					$queue = $this->is_internal_event( $action ) ? 'internal_events' : 'current_events';
+
+					array_push( $$queue, array(
```
ethitter: Fair enough. Sometimes I can be a little too "clever" in the name of concision.

cron.php
```diff
@@ -0,0 +1,464 @@
...
+
+	/**
+	 * Published scheduled posts that miss their schedule
+	 */
+	public function force_publish_missed_schedules() {
+		global $wpdb;
+
+		$missed_posts = $wpdb->get_col( $wpdb->prepare( "SELECT ID FROM {$wpdb->posts} WHERE post_status = 'future' AND post_date <= %s LIMIT 100;", current_time( 'mysql', false ) ) );
+
+		if ( is_array( $missed_posts ) && ! empty( $missed_posts ) ) {
```
ethitter: Indeed, `$wpdb->get_col()` will always return an array. I should've checked the source.

cron.php
```diff
@@ -0,0 +1,464 @@
...
+
+	/**
+	 * Ensure scheduled posts have a corresponding cron job to publish them
+	 */
+	public function confirm_scheduled_posts() {
+		global $wpdb;
+
+		$future_posts = $wpdb->get_results( $wpdb->prepare( "SELECT ID, post_date FROM {$wpdb->posts} WHERE post_status = 'future' AND post_date > %s LIMIT 100;", current_time( 'mysql', false ) ) );
+
+		if ( is_array( $future_posts ) && ! empty( $future_posts ) ) {
```
ethitter: True, the `empty()` check will handle situations where `$wpdb->get_results()` returns `null`.

cron.php
ethitter: Yeah...

Let me ponder this a bit. The craziest solution I've devised thus far is to intercept Core's cron storage and move it all to a CPT. I suspect there's a more-reasonable way to deal with this, though.

cron.php
```diff
@@ -0,0 +1,464 @@
...
+
+		return in_array( $action, $blocked_hooks );
+	}
+
+	/**
+	 * Set a lock and limit how many concurrent jobs are permitted
+	 */
+	private function check_lock() {
+		// Prevent deadlock
+		if ( (int) wp_cache_get( $this->cache_key_lock_timestamp, null, true ) < time() - $this->job_timeout_in_minutes * MINUTE_IN_SECONDS ) {
```
ethitter: 👍

cron.php
```diff
@@ -0,0 +1,464 @@
...
+			wp_cache_set( $this->cache_key_lock, 0 );
+			wp_cache_set( $this->cache_key_lock_timestamp, time() );
+			return true;
+		}
+
+		// Check if process can run
+		if ( (int) wp_cache_get( $this->cache_key_lock, null, true ) >= $this->job_concurrency_limit ) {
+			return false;
+		} else {
```
ethitter: I know, there's just something about the symmetry and explicitness that I'm fond of.

cron.php
ethitter: I'm having second thoughts. Particularly when paired with the context the events endpoint provides, `instance` seems better than `index`. The hash represents an instance of a scheduled event, which also happens to be its index in an array.

cron.php
```diff
@@ -0,0 +1,464 @@
...
+			return new WP_Error( 'no-secret', __( 'Secret must be specified with all requests', 'wp-cron-control-revisited' ) );
+		}
+
+		return true;
+	}
+
+	/**
+	 * Events that are always run, regardless of how many jobs are queued
+	 */
+	private function is_internal_event( $action ) {
```
ethitter: `priority` doesn't quite fit, because these jobs appear last in the list the endpoint returns, so their runners are scheduled after those for non-internal events.

cron.php
```diff
@@ -0,0 +1,464 @@
...
+	 */
+	public function register_internal_events_schedules( $schedules ) {
+		return array_merge( $schedules, $this->internal_jobs_schedules );
+	}
+
+	/**
+	 * Schedule internal jobs
+	 */
+	public function schedule_internal_events() {
+		$when = strtotime( sprintf( '+%d seconds', 2 * $this->job_queue_window_in_seconds ) );
```
ethitter: I honestly have no idea; I might've been using one of Core's time constants before switching to the class property. Regardless, I'll remove it; shifting by the job queue window just avoids scheduling something that'll immediately miss its schedule.

cron.php
```diff
@@ -0,0 +1,464 @@
...
+			return;
+		}
+
+		// Load plugin functionality, when conditions are met
+		if ( defined( 'WP_CRON_CONTROL_SECRET' ) ) {
+			// Block normal cron execution
+			define( 'DISABLE_WP_CRON', true );
+			define( 'ALTERNATE_WP_CRON', false );
+
+			add_action( 'muplugins_loaded', array( $this, 'block_direct_cron' ) );
```
ethitter: I'll leave this for now, as `muplugins_loaded` is the first action that we can use to block rogue cron requests, and using it avoids wasting time loading any client code.

cron.php
```diff
@@ -0,0 +1,464 @@
...
+			remove_action( 'init', 'wp_cron' );
+
+			add_filter( 'cron_request', array( $this, 'block_spawn_cron' ) );
+
+			// Core plugin functionality
+			$this->prepare();
+			add_action( 'rest_api_init', array( $this, 'rest_api_init' ) );
+			add_filter( 'cron_schedules', array( $this, 'register_internal_events_schedules' ) );
+			add_action( 'muplugins_loaded', array( $this, 'schedule_internal_events' ), 11 );
```
ethitter: My choice of action, and use of priority, was pretty silly in this case; there's no need for it, so I'll find a more-appropriate hook.

cron.php
mjangda: Cool, that works. Maybe we can add some inline docs to avoid confusion?

ethitter:
Thanks for the feedback, Mo! I'm thinking about the race-condition issue and how to address that with a CPT; everything else should be addressed.

As for the structure, I almost started this as its own plugin, then thought about putting it in a subdirectory, before finally writing it as a single class. I was undecided as to whether or not it would be large enough to warrant anything more complex. It might make sense to restructure it as a plugin, particularly if a CPT enters the mix.

ethitter:
Storing the `cron` option's constituents as CPT entries will be fairly straightforward thanks to the `pre_` filters in Core's options functions. All of Core's cron functions rely on two helpers to interact with the `cron` option, rather than using the options functions. Thanks to this, `_get_cron_array()` and `_set_cron_array()` enforce the consistency we need to intercept uses of the `cron` option. When executing events, the plugin can modify the CPT entries directly, which eliminates the race condition updating the `cron` option.

ethitter:
Note to self, should this become its own plugin: https://help.github.com/articles/splitting-a-subfolder-out-into-a-new-repository/.

ethitter:
There's surely more de-duplication/simplification to be done now that I've reorganized things.

I also need to deal with the case of events that are unscheduled without being run, and some missing events from plugins like Jetpack and VaultPress.

ethitter:
I should've done this in the first place: https://github.com/Automattic/WP-Cron-Control-Revisited/pull/1.

ethitter:
@mjangda This is ready for another review over in https://github.com/Automattic/WP-Cron-Control-Revisited/pull/1.

mjangda:
Sweet, will take a look.

001-cron.php
```diff
@@ -56,4 +56,14 @@ public function block_admin_page() {
 	}
 }
 
-new WPCOM_VIP_Cron_Control;
+// Allow testing of new approach to cron execution
+$whitelisted_sites = array();
+if ( VIP_GO_ENV && in_array( FILES_CLIENT_SITE_ID, $whitelisted_sites ) ) {
+	add_filter( 'wpcom_vip_go_enable_new_cron_control', '__return_true' );
```
mjangda: Do we need the filter here? We could just do the `require` here and add an `else` clause as a fallback.

mjangda:
Note: We should probably "Squash and Merge" (most of the commits are not necessary to merge into master since the relevant code was removed).

001-cron.php
ethitter: The filter's there to allow for local testing, as we set `VIP_GO_ENV` to false in https://github.com/Automattic/vip-go-mu-plugins/blob/cc326452d8745063ffef8032ab46de9b6ef7cc8e/000-vip-init.php#L14.

001-cron.php
```diff
@@ -56,4 +56,14 @@ public function block_admin_page() {
 	}
 }
 
-new WPCOM_VIP_Cron_Control;
+// Allow testing of new approach to cron execution
+$whitelisted_sites = array();
```
mjangda: Do we need to `unset` this after use to avoid persisting the global?

mjangda:
One small note about the `$whitelisted_sites` var but otherwise 🚢

001-cron.php
ethitter: 👍

ethitter:
Thanks!

PR #247, opened 2016-07-14
Cache purging API

simonwheatley:
Fixes #246

PR #334, opened 2016-10-24
Purge entire site cache check

simonwheatley:
**Branches off https://github.com/Automattic/vip-go-mu-plugins/pull/247 Please merge that PR first.**

From @ethitter and @joshbetz's [comments here](https://github.com/Automattic/vip-go-mu-plugins/pull/247#discussion_r84173878):

> _ethitter 5 days ago_
> For single-site WordPress, is_super_admin() uses the delete_users capability (let's set aside that this function even existing in single-site is just silly). manage_options strikes me as a more-appropriate single-site capability to check. Thoughts?
> 
> _joshbetz 5 days ago Automattic member_
> For single sites, I think manage_options makes sense. It's probably also OK for sub-domain multisites. It may make sense to require super_admin on sub-directory multisites, at least on the main site, as that will clear cache for the entire network.

Instead I've added an additional capability, which I've succinctly named `wpcom_vip_cache_manager_purge_entire_site`. 

This diff may help reviewing changes, prior to https://github.com/Automattic/vip-go-mu-plugins/pull/247 being merged: https://github.com/Automattic/vip-go-mu-plugins/compare/246-cache-purging-api...246-address-is_super_admin-check

Feedback welcome 😄

### DIFF
--- a/vip-cache-manager/vip-cache-manager.php
+++ b/vip-cache-manager/vip-cache-manager.php
@@ -43,7 +43,7 @@ class WPCOM_VIP_Cache_Manager {
 	}
 
 	public function init() {
-		if ( is_super_admin() && isset( $_GET['cm_purge_all'] ) && check_admin_referer( 'manual_purge' ) ) {
+		if ( isset( $_GET['cm_purge_all'] ) && current_user_can( 'wpcom_vip_cache_manager_purge_entire_site' ) && check_admin_referer( 'manual_purge' ) ) {
 			$this->purge_site_cache();
 			add_action( 'admin_notices' , array( $this, 'manual_purge_message' ) );
 		}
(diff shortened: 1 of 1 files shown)


ethitter:
Makes sense, though we should use `wpcom_vip_merge_role_caps()` to add the capability: https://vip.wordpress.com/functions/wpcom_vip_merge_role_caps/.

joshbetz:
@simonwheatley Note: I force pushed to this branch after rebasing to clean up the diff.

joshbetz:
The version check seems unnecessary. If we define the cap in code like Erick mentioned we can probably avoid it. Otherwise I'd say we can use manage_options everywhere except site_id = 1 in subdirectory multisites?

ethitter:
There is some danger in making this a unique capability, since it could be applied to any user/user role. That said, code review is one way to catch that, and I appreciate the explicitness of the capability. `map_meta_cap` could be used to accomplish the same, but it's not as clear. The capability's also useful for sites that implement custom roles that might also need the ability to purge, or that might be more-appropriate places for that capability.

joshbetz:
Superseded by #353

PR #338, opened 2016-10-25
Only PURGE the_permalink on clean_post_cache

joshbetz:
For normal post updates (publish->publish), we don't need to clear all related URLs because the post still exists. Only clear all the related URLs when the status transitions (i.e. draft->publish or publish->draft).

### DIFF
--- a/cache-manager.php
+++ b/cache-manager.php
@@ -35,6 +35,7 @@ class WPCOM_VIP_Cache_Manager {
 		add_action( 'clean_post_cache', array( $this, 'queue_post_purge' ) );
 		add_action( 'clean_term_cache', array( $this, 'queue_term_purge' ), 10, 3 );
 		add_action( 'transition_post_status', array( $this, 'transition_post_status' ), 10, 3 );
+		add_action( 'post_updated', array( $this, 'post_updated' ), 10, 3 );
 		add_action( 'switch_theme', array( $this, 'purge_site_cache' ) );
 
 		add_action( 'activity_box_end', array( $this, 'get_manual_purge_link' ), 100 );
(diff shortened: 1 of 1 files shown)


nickdaugherty:
My thoughts:

This is going to leave stale data in places like feeds and archive pages - for example, if a post title has a typo and it's noticed after the post is live, editorial will go to fix it, but the update won't be reflected in feeds, archives, etc, for up to 30m and folks are going to wonder why.

joshbetz:
I think the alternative is to lower the TTL on feeds to 5 minutes and disable purges, which will lead to more stale data.

nickdaugherty:
IMO the right way to handle this is to aggregate purge requests across a given timeframe, so that we can't spam an arbitrary number of purges against the caches, which is what happens currently with no throttling.

cache-manager.php
```diff
@@ -246,11 +248,65 @@ function queue_post_purge( $post_id ) {
...
+			return;
+		}
+
+		// Only send PURGE requests for public post types
+		if ( ! is_post_type_viewable( $post->post_type ) ) {
+			return;
+		}
+
+		if ( 'publish' !== get_post_status( $post_id ) ) {
+			return
```
nickdaugherty: Missing semi-colon

cache-manager.php
```diff
@@ -246,11 +248,65 @@ function queue_post_purge( $post_id ) {
...
+			return;
+		}
+		
+		// Only send PURGE requests for public post types
+		if ( ! is_post_type_viewable( $post->post_type ) ) {
+			return;
+		}
+		
+		// Don't PURGE if we're not transitioning to or from publish
+		if ( $new === $old || ( $new != 'publish' && $old != 'publish' ) ) {
```
nickdaugherty: Do other post statuses need to be considered here? Mostly thinking about EditFlow.

cache-manager.php
```diff
@@ -246,11 +248,65 @@ function queue_post_purge( $post_id ) {
...
+		// PURGE all feeds if one of the whitelisted fields is updated
+		foreach( $fields as $field ) {
+			if ( $post->{$field} != $old_post->{$field} ) {
+				$this->clear_feed_urls( $post );
+				return;
+			}
+		}
+	}
+
+	function clear_feed_urls( $post ) {
```
nickdaugherty: We may want to name this function differently, since it purges more than just feeds.

cache-manager.php
joshbetz: I thought about that... I think publish is always publish. Edit Flow adds intermediate steps before publish, but (afaik) the publish state is still the same.

nickdaugherty: Cool - that's what I was thinking too. If we find there are edge cases, we can revisit.

joshbetz:
Regardless of additional measures, this should only eliminate unnecessary cache purges.

PR #345, opened 2016-10-28
Add monitoring

simonwheatley:


joshbetz:
`WPCOM_VIP_Query_Log` seems like a lot of overhead for what comes down to essentially 1 error_log statement on shutdown. Getting rid of the singleton would reduce it to ~10 lines of code.

simonwheatley:
@joshbetz Fair point, I thought I was going to have to store the queries somewhere when I started… but then I remembered `WPDB::queries`.

misc.php
```diff
@@ -140,3 +140,21 @@ function _vip_filter_rest_url_for_ssl( $url ) {
 
 	return $url;
 }
+
+
+function wpcom_vip_query_log() {
+	if ( '/cache-healthcheck?' === $_SERVER['REQUEST_URI'] ) {
+		return;
+	}
+	$num_queries = count( $GLOBALS['wpdb']->queries );
```
joshbetz: I think `$wpdb->num_rows` will do the same thing?

joshbetz:
Looks good to me.

simonwheatley:
I'll probably remove the Jetpack `updating_jetpack_version` before deploying it to production…

simonwheatley:
I've added some conditional checks for constants, so we can carefully control which sites get instrumented and we don't end up running lots of logging everywhere unexpectedly.

misc.php
nickdaugherty: This will be empty unless SAVEQUERIES has been set, right?

simonwheatley: Fair point, I'll conditionally set `SAVEQUERIES`.

simonwheatley: Fair point, I'll conditionally set `SAVEQUERIES`.

PR #349, opened 2016-10-31
VIP: Feed Cache Plugin

emrikol:
This should work as a temporary fix to a caching problem with SimplePie (and `fetch_feed()`).

### DIFF
--- a/vip-feed-cache.php
+++ b/vip-feed-cache.php
@@ -1,97 +1,15 @@
 <?php
-// Require core feed class.
-require_once( ABSPATH . WPINC . '/class-feed.php' );
-
-class VIPGO_Feed_Cache extends SimplePie_Cache {
-	/**
-	 * Creates a new SimplePie_Cache object.
-	 *
-	 * @access public
-	 *
-	 * @param string $location  URL location (scheme is used to determine handler).
-	 * @param string $filename  Unique identifier for cache object.
-	 * @param string $extension 'spi' or 'spc'.
-	 * @return VIPGO_Feed_Cache_Transient Feed cache handler object that uses transients and normalizes SimplePie Build number.
-	 */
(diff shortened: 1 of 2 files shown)


vip-feed-cache.php
```diff
@@ -0,0 +1,99 @@
+<?php
+// Require core feed class.
+require_once( ABSPATH . WPINC . '/class-feed.php' );
```
mjangda: Rather than loading `class-feed.php` (and as a result, SimplePie) on every single request, we should move this require + our custom classes to a separate file, which is then `require`-ed inside `vipgo_feed_options`

vip-feed-cache.php
```diff
@@ -0,0 +1,99 @@
...
+	 * @return mixed Transient value.
+	 */
+	public function load() {
+		$transient = get_transient( $this->name );
+
+		// If we don't have the required data, bail.
+		if ( ! defined( 'SIMPLEPIE_BUILD' ) || ! isset( $transient['build'] ) ) {
+			return $transient;
+		}
+		$transient['build'] = SIMPLEPIE_BUILD;
```
joshbetz: Can we use the WordPress version string as the build number?

emrikol: Unfortunately not.  Because of how SimplePie checks the `SIMLPEPIE_BUILD` it is outside of our _non-core-hack_ control.  The best we can do is ignore whatever it's set to and fool SimplePie into thinking it's correct.  As long as we use object caching for transients and the cache is cleared on core WordPress updates, there shouldn't be any issue with incompatible cache data.

mjangda:
Added one small change (use full path when `require`-ing).

:ship:

mjangda:
@emrikol rebased this and tests pass now. Good to merge?

mjangda:
r85413-deploy

PR #350, opened 2016-11-02
Bringing vip-caching up to date from wpcom

emrikol:
SVN commit messages:

r266855-wpcom-vip: `put a limit to how long we cache when something is found to not be present. Also add cache busting to term_existsso that when a new term is created we purge any not found results in cache`

r266857-wpcom-vip: `skip a costly meta_value lookup in wp_old_slug_redirect when we're pretty sure this is a mistake. with help from @jmdodd`


mjangda:
👍

PR #351, opened 2016-11-14
Improve perfomance of _WP_Editors::wp_link_query

david-binda:
The WordPress core is currently [not setting `no_found_rows`](https://github.com/WordPress/WordPress/blob/493f76a3d2ef8c030ab5dcd4333f9a401208f534/wp-includes/class-wp-editor.php#L1351,L1358) inside the `_WP_Editors::wp_link_query` and [is setting `suppress_filters` to `true`](https://github.com/WordPress/WordPress/blob/493f76a3d2ef8c030ab5dcd4333f9a401208f534/wp-includes/class-wp-editor.php#L1353).

Since the `_WP_Editors::wp_link_query` method [is not using the `found_posts` nor `max_num_pages` properties](https://github.com/WordPress/WordPress/blob/493f76a3d2ef8c030ab5dcd4333f9a401208f534/wp-includes/class-wp-editor.php#L1383,L1385) of `WP_Query` class, the `SQL_CALC_FOUND_ROWS` in produced SQL query is extra and useless.

Setting `suppress_filters` to `false`, on the other hand, is blocking the Advanced Posts Cache plugin and thus all those queries are not being cached, despite being the same across multiple post edit screens for all logged in users

### DIFF
--- a/misc.php
+++ b/misc.php
@@ -163,3 +163,30 @@ if ( defined( 'WPCOM_VIP_QUERY_LOG' ) && WPCOM_VIP_QUERY_LOG ) {
 	$GLOBALS['wpdb']->save_queries = SAVEQUERIES;
 	add_action( 'shutdown', 'wpcom_vip_query_log' );
 }
+
+/**
+ * Improve perfomance of the `_WP_Editors::wp_link_query` method
+ *
+ * The WordPress core is currently not setting `no_found_rows` inside the `_WP_Editors::wp_link_query`
+ * and is setting `suppress_filters` to `true`.
+ * 
+ * Since the `_WP_Editors::wp_link_query` method is not using the `found_posts` nor `max_num_pages`
+ * properties of `WP_Query` class, the `SQL_CALC_FOUND_ROWS` in produced SQL query is extra and
+ * useless.
+ *
+ * Setting `suppress_filters` to `false`, on the other hand, is blocking the Advanced Posts Cache
+ * plugin and thus all those queries are not being cached, despite being the same across multiple
+ * post edit screens for all logged in users
+ */
+function wpcom_vip_wp_link_query_args( $query ) {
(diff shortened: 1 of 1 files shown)


mjangda:
Setting `no_found_rows` feels like a good core improvement; we should open a ticket + patch if there isn't one already.

As for `suppress_filters`, I think it's debatable if it's useful. Was this a result of an issue we've seen on a site? Or just an extra uncached query we spotted? Outside of the recent posts query (which is shown by default when you are adding a link), it's unlikely that authors are searching for the same terms (so the cache is probably not going to be used that often). Plus, in an admin context, the cache is probably being flushed quite a bit as well (especially in high activity environments).

philipjohn:
Core ticket is https://core.trac.wordpress.org/ticket/38784

david-binda:
@mjangda Thanks for a nice code review!

> Setting no_found_rows feels like a good core improvement; we should open a ticket + patch if there isn't one already.

Yup, [I have opened a core ticket + supplied a patch](https://core.trac.wordpress.org/ticket/38784)

> As for suppress_filters, I think it's debatable if it's useful. Was this a result of an issue we've seen on a site? Or just an extra uncached query we spotted?

This is a result of an issue we've seen on a site. The site was reporting slowness related to this feature.

> Outside of the recent posts query (which is shown by default when you are adding a link), it's unlikely that authors are searching for the same terms (so the cache is probably not going to be used that often). Plus, in an admin context, the cache is probably being flushed quite a bit as well (especially in high activity environments).

This seems legit. My motivation for adding that was primarily the "recent posts query", not the search queries. While I still believe that it could have some positive impact, just not that huge as one could expect, it might be better if that query arg was managed by a specific plugin - eg.: Advanced Post Cache or es-wp-query. Let me remove that one from this PR.

mjangda:
Rebased against master to fix the broken tests; this is 👍

mjangda:
Deployed in r85408-deploy

PR #352, opened 2016-11-14
Upgrading the Chartbeat plugin to v2.0.3.1

david-binda:
* Added 'chartbeat_config_path' filter to return a 'path' to Chartbeat configuration.
* Extra chaging for the dashboard widget

Syncing with the WordPress.com VIP


mjangda:
Deployed in r85522-deploy

PR #353, opened 2016-11-15
Remove cache purge button for non-a12s

joshbetz:


vip-cache-manager/vip-cache-manager.php
```diff
@@ -52,20 +52,18 @@ public function init() {
 	}
 
 	public function get_manual_purge_link() {
+		if ( ! function_exists( 'is_proxied_automattician' ) || ! is_proxied_automattician() ) {
```
mjangda: Probably should add this same check in `init` above.

vip-cache-manager/vip-cache-manager.php
```diff
@@ -37,7 +37,7 @@ public function __construct() {
 	}
 
 	public function init() {
-		if ( is_super_admin() && isset( $_GET['cm_purge_all'] ) && check_admin_referer( 'manual_purge' ) ) {
+		if ( ! function_exists( 'is_proxied_automattician' ) || ! is_proxied_automattician() && isset( $_GET['cm_purge_all'] ) && check_admin_referer( 'manual_purge' ) ) {
```
mjangda: This probably needs to be a positive check?, i.e. `true === is_proxied_automattician()`

Might be good to to move to a helper method in the class:

```
private function can_purge() {
    return function_exists( 'is_proxied_automattician' ) && is_proxied_automattician();
}
```

mjangda:
@joshbetz DRY-ied the code a bit. Can you take a look?

One related issue is that `WPCOM_VIP_Cache_Manager` has some public `purge_*` and `queue_*` methods which sites could abuse. For completeness, it would be good to make those `protected` or `private` but not a big deal.

mjangda:
r84995-deploy

PR #355, opened 2016-11-17
Adding filter for disabling term cache updates

david-binda:
Following code can be used from within a theme for disabling term cache updates

```
add_filter( 'wpcom_vip_disable_term_cache_invalidation', '__return_true' );
```


mjangda:
Deployed earlier in r83731-deploy

Issue #358, opened 2016-11-24
Run the cache health check as early as possible


simonwheatley:
When `cache-manager.php` was renamed to `vip-cache-manager.php`, the time at which the caching health check was run was moved later. Now Jetpack, VaultPresss, and many other plugins are loaded before the health check runs. If we moved the check to the top of `000-vip-init.php` we'd run the health check and bail before the following loaded, which has *got* to have some performance advantage 😄 

```
wp-content/mu-plugins/vip-helpers/vip-utils.php
wp-content/mu-plugins/vip-helpers/vip-caching.php
wp-content/mu-plugins/vip-helpers/vip-roles.php
wp-content/mu-plugins/vip-helpers/vip-permastructs.php
wp-content/mu-plugins/vip-helpers/vip-mods.php
wp-content/mu-plugins/vip-helpers/vip-media.php
wp-content/mu-plugins/vip-helpers/vip-elasticsearch.php
wp-content/mu-plugins/vip-helpers/vip-stats.php
wp-content/mu-plugins/vip-helpers/vip-deprecated.php
wp-content/mu-plugins/vip-helpers/vip-syndication-cache.php
wp-content/mu-plugins/a8c-files.php
wp-content/mu-plugins/advanced-post-cache/advanced-post-cache.php
wp-content/mu-plugins/akismet.php
wp-content/mu-plugins/akismet/akismet.php
wp-content/mu-plugins/akismet/class.akismet.php
wp-content/mu-plugins/akismet/class.akismet-widget.php
wp-content/mu-plugins/akismet/wrapper.php
wp-content/mu-plugins/debug-bar.php
wp-content/mu-plugins/jetpack.php
wp-content/mu-plugins/jetpack/jetpack.php
wp-content/mu-plugins/jetpack/class.jetpack.php
wp-content/mu-plugins/jetpack/class.jetpack-network.php
wp-content/mu-plugins/jetpack/class.jetpack-client.php
wp-content/mu-plugins/jetpack/class.jetpack-data.php
wp-content/mu-plugins/jetpack/class.jetpack-client-server.php
wp-content/mu-plugins/jetpack/sync/class.jetpack-sync-actions.php
wp-content/mu-plugins/jetpack/class.jetpack-options.php
wp-content/mu-plugins/jetpack/class.jetpack-user-agent.php
wp-content/mu-plugins/jetpack/class.jetpack-post-images.php
wp-content/mu-plugins/jetpack/class.jetpack-error.php
[...]

PR #360, opened 2016-11-24
Fix travis checks

simonwheatley:
Do not merge


simonwheatley:
I think this is failing because our submodules use URLs in the format `git@github.com:x/y.git`, which seems to require key authentication even if the repository is public. We try to get around this by converting the submodules with `sed -i -e "s|git@\([^:]*\):|https://\1/|" .gitmodules` in `ci/prepare.sh`, but this doesn't catch a submodule within a submodule 😐

simonwheatley:
Moving this fix into #358

PR #361, opened 2016-11-24
vip-cache-manager: adding cache purging logic for changed permalinks

david-binda:
In case a permalink is changed during the post update, the old permalink is never purged as `get_permalink` call inside the `queue_post_purge` is being called with `$post_id` and thus returns the current/new URL.

This commit is adding a new `WPCOM_VIP_Cache_Manager::queue_old_permalink_purge` method hooked to `post_udapted` action and obtains the old permalink by calling `get_permalink` while passing it the `$post_before` WP_Post object.

The cache purge is scheduled only in case the permalink got changed and if the post's status before the update was 'publish'.

### DIFF
--- a/vip-cache-manager/vip-cache-manager.php
+++ b/vip-cache-manager/vip-cache-manager.php
@@ -519,7 +519,7 @@ class WPCOM_VIP_Cache_Manager {
 		 * }
 		 * @param type  $term_id The ID of the term which is the primary reason for the purge
 		 */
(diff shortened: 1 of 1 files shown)


david-binda:
gah, I run into some minor issues due to the discrepancy in between GH and SVN version. It's ready for a review @simonwheatley !

david-binda:
Merging since it has already been deployed to production

simonwheatley:
Added the deployed label

Issue #365, opened 2016-12-02
Require for WP_Feed_Cache fails fatally on 4.7


simonwheatley:
See: https://github.com/Automattic/vip-go-mu-plugins/blob/568ba2fe26c2ad176a509782b459a6b379ff5792/vip-helpers/vip-syndication-cache.php#L4-L6

The files have been moved/split up, see: https://github.com/WordPress/WordPress/blob/493f76a3d2ef8c030ab5dcd4333f9a401208f534/wp-includes/feed.php#L672-L679

fklein-lu:
Reported to Core in https://core.trac.wordpress.org/ticket/39027 because the file should have been deprecated instead of removed.

fklein-lu:
This was fixed in Core in https://core.trac.wordpress.org/changeset/39459

PR #366, opened 2016-12-02
Require the required class files separately, if necessary

simonwheatley:
Cribs from the core approach here:
https://github.com/WordPress/WordPress/blob/493f76a3d2ef8c030ab5dcd4333f9a401208f534/wp-includes/feed.php#L672-L679

Addresses #365

### DIFF
--- a/ci/prepare.sh
+++ b/ci/prepare.sh
@@ -7,10 +7,14 @@
 set -ex
 
 cd $TRAVIS_BUILD_DIR
-if [ -w .gitmodules ]; then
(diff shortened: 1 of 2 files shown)


simonwheatley:
There is a core ticket which proposes moving the `class-feed.php` file back, and adding a deprecated notice. We should reverse the order of our checks, so we check for the 4.7 files first and only after than check the `class-feed.php` file in 4.6 and under.

simonwheatley:
That core ticket: https://core.trac.wordpress.org/ticket/39027

PR #372, opened 2016-12-05
Load Debug Bar Cron if it isn't already present

### LINKED ISSUE #371: Include Debug Bar Cron
Since we're already including the Debug Bar plugin, [Debug Bar Cron](https://wordpress.org/plugins/debug-bar-cron/) would be a helpful addition, particularly to monitor the new cron plugin (#329).

ethitter:
Since we're already including the Debug Bar plugin, Debug Bar Cron would be a helpful addition, particularly to monitor the new cron plugin (#329).

Fixes #371

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -37,3 +37,6 @@
 [submodule "cron-control"]
 	path = cron-control
 	url = git@github.com:Automattic/Cron-Control.git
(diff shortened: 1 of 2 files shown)


ethitter:
FWIW, we have a version of this on dotcom already.

mjangda:
🚢

PR #376, opened 2016-12-06
a8c-files.php: fix media upload after switch_to_blog

david-binda:
In case code swiches from a site with a directory style URL (eg. /my-awesome-site) to a main site using `switch_to_blog`, the upload URL path does not get properly set, since the `upload_url_path` option inside the _wp_upload_dir() function call is empty. The `WP_CONTENT_DIR` constant gets used, but it is already set to previous site's url path.

This commit is filtering the `upload_url_path` option and always, when `is_multisite` is true, sets it to `site_url() . '/wp-content/uploads'` which resolves the issues.

### DIFF
--- a/a8c-files.php
+++ b/a8c-files.php
@@ -546,8 +546,8 @@ class A8C_Files {
 	}
 
 	public function upload_url_path( $upload_url_path, $option ) {
-		// No modifications needed outside multisite or in case the option is not empty
-		if ( false === is_multisite() || false === empty( $upload_url_path ) ) {
+		// No modifications needed outside multisite
+		if ( false === is_multisite() ) {
 			return $upload_url_path;
 		}
 		// Change the upload url path to site's URL + wp-content/uploads without trailing slash
--- a/a8c-files.php
+++ b/a8c-files.php
@@ -60,7 +60,7 @@ class A8C_Files {
 		add_filter( 'pre_option_uploads_use_yearmonth_folders', function( $arg ) { return '1'; } );
 
 		// ensure the correct upload URL is used even after switch_to_blog is called
-		add_filter( 'option_upload_url_path', array( &$this, 'upload_url_path' ), 10, 2 );
+		add_filter( 'option_upload_url_path', array( $this, 'upload_url_path' ), 10, 2 );
 	}
 
 	function check_to_upload_file( $data, $postarr ) {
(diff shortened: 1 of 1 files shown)


david-binda:
I haven't looked properly into it, but I have a feeling that @darthhexx 's https://github.com/Automattic/vip-go-mu-plugins/commit/9ce63bf155356377bd9345fbe0ff194f6c3e6f59 could be reverted when this is in place. Also, filtering the option seems to be more bullet proof solution than using regexes to fix the issue post-mortem.

a8c-files.php
```diff
@@ -58,6 +58,9 @@ function __construct() {
 
 		// ensure we always upload with year month folder layouts
 		add_filter( 'pre_option_uploads_use_yearmonth_folders', function( $arg ) { return '1'; } );
+
+		// ensure the correct upload URL is used even after switch_to_blog is called
+		add_filter( 'pre_option_upload_url_path', array( &$this, 'upload_url_path' ), 10, 1 );
```
ethitter: It's probably best that the mu-plugin always has the final say, but is there any risk or downside to using `option_upload_url_path` instead? The `pre_` version does skip the DB call, which is a nice side effect.

a8c-files.php
```diff
@@ -542,6 +545,18 @@ function get_files_service_hostname() {
...
 	}
 
+	public function upload_url_path( $upload_url_path ) {
+		// No modifications needed outside multisite.
+		if ( false === is_multisite() ) {
+			return $upload_url_path;
+		}
+		// Change the upload url path to site's URL + wp-content/uploads without trailing slash
+		// Related core code: https://core.trac.wordpress.org/browser/tags/4.6.1/src/wp-includes/functions.php#L1929
+		$upload_url_path = trailingslashit( get_site_url() ) . 'wp-content/uploads';
```
ethitter: An easier approach is to pass `wp-content/uploads` as the second argument to `get_site_url()`, leaving the first argument `null` to let Core sort out the blog ID. This likely eliminates the need for `trailingslashit()`, though replacing it with `untrailingslashit()` may be necessary/advantageous.

david-binda: Sure. That looks like a better / cleaner solution.

david-binda:
Merging since it has been deployed. Thanks @ethitter and @simonwheatley !

PR #378, opened 2016-12-07
No Core Update Nags

mjangda:
We don't want to show core update nags since VIP currently maintains the update. Two related changes here:

- Removes `update_core` cap from all users (including super admins).
- Disables the displays of the core update nags (since those are still shown to users that cannot update core).


ethitter:
👍

mjangda:
r84964-deploy

PR #382, opened 2016-12-08
alloptions helpers

nickdaugherty:
Adds monitoring/alerting of alloptions size and a WP CLI command for identifying large options.


simonwheatley:
This may be controversial, but should we change the error message to something more anodyne?

nickdaugherty:
> This may be controversial, but should we change the error message to something more anodyne?

Possibly - @darthhexx what HTTP response code would you like sites to use when they are blocked from loading due to their `alloptions` size being > 1MB?

For reference, we don't do anything on dotcom (so you get a `200`) - I believe this is mostly due to Batcache.

nickdaugherty:
> This may be controversial, but should we change the error message to something more anodyne?

Oh the message text itself - yes I'm all for changing it to something else.

nickdaugherty:
Reworked the error page in 2bcc0cb to use design by @keoshi, with more anodyne copy (and no VIP logo :D )

service-unavailable

nickdaugherty:
This is ready for review.

Notes: 

I went with a 503 for blocked sites. This will cause Varnish to continue in 'grace' mode while preventing caching and triggering nagios alerts. dotcom currently uses a 200.

We hook the alert / blocking on `plugins_loaded`, which is not as ideal as `muplugins_loaded`. This is because `wp_mail()` is not yet available at `muplugins_loaded` and is needed for sending alerts. We can either omit the email or get `mail()` working if this is a concern.

We also may need to tweak some stuff to get IRC messages flowing on prod. Getting connection timeouts (from sandbox) during testing. cc @darthhexx 

There are 3 constants that must be defined in `wp-config.php` to receive warnings: `VIP_IRC_HOSTNAME`, `VIP_IRC_PORT`, and `VIP_ALLOPTIONS_NOTIFY_EMAIL`.

The current approach will still cause at least one db query for `alloptions`, as we don't attempt to block before `alloptions` is loaded. To do so would require code to be placed in `object-cache.php`. This is the same behavior as dotcom.

alloptions-limit.php
```diff
@@ -0,0 +1,118 @@
...
+
+		if ( 0 !== $wpdb->blogid ) {
+			$subject .= ", blog ID {$wpdb->blogid}";
+		}
+
+		$subject .= ') options is up to %s';
+
+		$subject = sprintf(
+			$subject,
+			esc_url( $_SERVER['HTTP_HOST'] ),
```
ethitter: Is there an advantage to this over `home_url()` or some other Core function?

ethitter: Actually, I think this has to be switched to `home_url()` or `site_url()`, to account for subdirectory multisites. Right now, wouldn't this mis-identify which site is blocked if it's not the main site?

nickdaugherty: No advantage that I can think of - that line came across from WPCOM so I'm unsure why it was that way and not the other.

You're totally right about the subsites - I'll get it changed.

ethitter: > ...that line came across from WPCOM so I'm unsure why it was that way and not the other.

Probably for the same reason Nagios couldn't check subdirectories--there wasn't a need until Go came about. 😄

alloptions-limit.php
```diff
@@ -0,0 +1,118 @@
...
+function wpcom_vip_sanity_check_alloptions() {
+	if ( defined( 'WP_CLI' ) && WP_CLI ) {
+		return;
+	}
+
+	// Warn should *always* be =< die
+	$alloptions_size_warn  =  800000;
+	$alloptions_size_die   = 1000000; // 1000000 ~ 1MB, too big for memcache
+
+	static $alloptions_size = null; // Avoids repeated cache requests
```
mjangda: I'm not sure we need the static here since this function would only be called once on `plugins_loaded`.

errors/alloptions-limit.html
```diff
@@ -0,0 +1,37 @@
...
+			<div class="col-left">
+				<video class="video" width="100%" poster="wp-content/mu-plugins/errors/img/wpcomvip-error-poster.png" autoplay loop onclick="this.paused?this.play():this.pause();">
+					<source src="wp-content/mu-plugins/errors/video/wpcomvip-error-video.mov" type="video/mp4" />
+					<img src="wp-content/mu-plugins/errors/img/wpcomvip-error-poster.png" />
+				</video>
+			</div>
+
+			<div class="col-right">
+				<h1>Service Unavailable</h1>
+				<p>This service is temporarily unavailable, please try again in a few moments.</p>
```
mjangda: Should we change "service" to "site"?

mjangda:
@nickdaugherty Unless you have other changes pending, I'll work on deploying this tomorrow.

nickdaugherty:
@mjangda no more changes planned, I've been doing some last minute testing of this and it seems fine.

mjangda:
Broke up the deploy into smaller pieces to allow for easier reverts if needed:

wp-cli: r87176-deploy
helpers: r87177-deploy
error page: r87178-deploy
alloptions: r87186-deploy

PR #386, opened 2016-12-08
Enhance the healthchecks

simonwheatley:
WORK IN PROGRESS

Enhance the healthchecks to verify the DB and object cache

### DIFF
--- a/000-vip-init.php
+++ b/000-vip-init.php
@@ -18,19 +18,23 @@ function wpcomvip_healthcheck_fail( $msg ) {
  */
 function wpcomvip_run_healthcheck() {
         // Exercise the database by checking for a non-autoloaded option
-        global $wpdb;
-        $sql = "SELECT option_value FROM $wpdb->options WHERE option_name = 'wpcomvip_db_healthcheck'";
-        $db_check = $wpdb->get_var( $sql );
-        // The write only happens when the DB doesn't have the entry, which
-        // should just be the first time this check runs on a given database
-        if ( is_null( $db_check ) ) {
-                add_option( 'wpcomvip_db_healthcheck', true, null, 'no' );
-                $db_check = $wpdb->get_var( $sql );
-        }
+        $db_check = $GLOBALS['wpdb']->get_var( 'SELECT VERSION()' );
+
         // If this check fails, throw an exception so we can see it in the logs
         if ( is_null( $db_check ) ) {
                 wpcomvip_healthcheck_fail( 'VIP Go: DB check failed' );
+				return;
         }
+
+		// Check for memcached status; we fail if more than 50% of cache containers are unable to connect.
+		$memcached_server_count = count( $GLOBALS['memcached_servers']['default'] );
+		$memcached_error_count = count( $GLOBALS['wp_object_cache']['connection_errors'] );
+		$memcached_ratio = ceil( $memcached_error_count / $memcached_server_count * 100 );
+
+		if ( $memcached_ratio > 50 ) {
+			wpcomvip_healthcheck_fail( 'VIP Go: memcached check failed' );
+			return;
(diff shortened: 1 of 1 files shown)


simonwheatley:
The code so far checks the DB in a fairly efficient fashion, I hope.

I'd love to check the object cache, but we really only want to know that a Memcached backed object cache is present… we don't care about the in-PHP-memory object cache that WordPress will fall back to. I can't work out how best to do this.

Here's some super rough mostly not really working code:

``` php
        // Do we have at least one Memcached node in our 
        // pool which seems good?
        $got_memcached = false;
        foreach ( $GLOBALS['memcached_servers']['default'] as $server ) {
                list ( $node, $port ) = explode( ':', $server );
                $connected = $wp_object_cache->mc['default']->getServerStatus( $node, $port );
                if ( true === $connected ) {
                        $got_memcached = true;
                        break;
                }
        }
        if ( false === $got_connections ) {
                wpcomvip_healthcheck_fail( 'VIP Go: Memcached connections unavailable' );
        }
```

I'm very open to suggestions to do this in an improved fashion… I also think we might merge just the DB check for the time being.

/cc @darthhexx whose enquiry started me off here

000-vip-init.php
```diff
@@ -1,8 +1,46 @@
...
+        // Exercise the database by checking for a non-autoloaded option
+        global $wpdb;
+        $sql = "SELECT option_value FROM $wpdb->options WHERE option_name = 'wpcomvip_db_healthcheck'";
+        $db_check = $wpdb->get_var( $sql );
+        // The write only happens when the DB doesn't have the entry, which
+        // should just be the first time this check runs on a given database
+        if ( is_null( $db_check ) ) {
+                add_option( 'wpcomvip_db_healthcheck', true, null, 'no' );
+                $db_check = $wpdb->get_var( $sql );
+        }
```
mjangda: If we just care about verifying that the database connection is there, can we just run `$wpdb->get_var( 'SELECT VERSION()' )` and watch for a non-`null` value?

Alternatively, we could query for a non-existent value like `$wpdb->get_var( $wpdb->prepare( "SELECT option_id FROM $wpdb->options WHERE option_name = %s", 'vip_healthcheck_' . rand() ) )` and verify that `$wpdb->last_error` is empty.

This way, we can avoid having to add the option.

mjangda:
> I'd love to check the object cache, but we really only want to know that a Memcached backed object cache is present

`Memcache::addServer` has a failure callback, which we [currently hook into but do nothing with](https://github.com/Automattic/wp-memcached/blob/master/object-cache.php#L482). We could store any connection failures and look for them in the healthcheck.

mjangda:
Needs https://github.com/Automattic/wp-memcached/pull/10

simonwheatley:
> I'd love to check the object cache, but we really only want to know that a Memcached backed object cache is present… we don't care about the in-PHP-memory object cache that WordPress will fall back to. I can't work out how best to do this.

@mjangda Do we want to fail a healthcheck on an object cache failure? For many sites, not having memcached would degrade performance but not actually take the site down. Would a better alternative be to generate an alert, so someone can look into the issue?

mjangda:
> Do we want to fail a healthcheck on an object cache failure?

This might be a better question for @darthhexx. Feels like we should fail if > 50% of memcached servers return an error on connect.

darthhexx:
We would not want the health check to fail completely for memcached failures. The best would be to log the failure to `php-errors`. The error logs are going to be scanned in real time in the future and alerts sent for anomalous spikes, which would pick up a slew of Memcached connection failures. However, in the meantime, sending an alert from the Web Container is a good way to go.

simonwheatley:
Closing this as won't fix

PR #388, opened 2016-12-08
Revert "Move the health check early"

mjangda:
Reverts Automattic/vip-go-mu-plugins#359

@darthhexx says:

> the check is happening before any DB or Memcached queries so the container could be broken and yet still be kept in the Varnish cache "Healthy" list.

We need to run the check at a point where we're confident that db and memcache connections are active and working.


mjangda:
r84995-deploy

PR #392, opened 2016-12-12
a8c-files: image_resize should return false if image does not exists

david-binda:
In case the `wp_get_attachment_url` returns `false`, we should return false from the `image_resize` in order to not interfere with core's `image_downsize` functionality.

See related code:
* https://developer.wordpress.org/reference/functions/wp_get_attachment_url/
* https://developer.wordpress.org/reference/functions/image_downsize/
* https://developer.wordpress.org/reference/functions/wp_get_attachment_image_url/

### DIFF
--- a/a8c-files.php
+++ b/a8c-files.php
@@ -645,6 +645,10 @@ class A8C_Files {
 		$resized = false;
 		$img_url = wp_get_attachment_url( $id );
 
+		if ( false === $img_url ) {
+			return false;
+		}
+
 		/**
 		 * Filter the original image Photon-compatible parameters before changes are
 		 *
--- a/a8c-files.php
+++ b/a8c-files.php
@@ -585,6 +585,13 @@ class A8C_Files {
 	function image_resize( $ignore, $id, $size ) {
 		global $_wp_additional_image_sizes, $post;
 
+		// Don't bother resizing non-image (and non-existent) attachment.
+		// We fallthrough to core's image_downsize but that bails as well.
+		$is_img = wp_attachment_is_image( $id );
+		if ( ! $is_img ) {
+			return false;
+		}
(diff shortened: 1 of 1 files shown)


joshbetz:
* Core checks `wp_attachment_is_image( $id )`, should we do the same?
* Returning false could change functionality because the `image_downsize()` function will now continue past the filter.

philipjohn:
Crazy idea: What about this at line 712:

```
		// Don't try to modify a non-existent URL.
		if ( false === $img_url ) {
			add_filter( 'wp_get_attachment_image_src', function( $image ) {
				if ( true === $image ) {
					return false;
				}
			} );
			return true;
		}
```
That stops `image_downsize()` doing it's thing because [the conditional at line 201](https://core.trac.wordpress.org/browser/tags/4.7/src/wp-includes/media.php#L201) will evaluate to true We then catch that within `wp_get_attachment_image_src()` with [the filter](https://core.trac.wordpress.org/browser/tags/4.7/src/wp-includes/media.php#L851).

It feels hacky, and I'm not sure of other side effects yet.

mjangda:
Without patch:

```
wp> wp_get_attachment_image_src( 9999, 'large' )
array(4) {
  [0] =>
  string(8) "/?w=1024"
  [1] =>
  string(4) "1024"
  [2] =>
  string(4) "1024"
  [3] =>
  bool(true)
}
```

With patch:

```
wp> wp_get_attachment_image_src( 9999, 'large' )
bool(false)
```

mjangda:
Updated to use `wp_attachment_is_image` to match core, moved the check much earlier and just passthrough to core (which also just bails).

mjangda:
r91676-deploy

PR #401, opened 2016-12-15
Performance: Add plugin that stores lastpostmodified async

mjangda:
Instead of calculating on-the-fly, which can be slow on large sites (or problematic when publishing large volumes of content), store the modified date of the last post save on publish/update and use that value instead.

### DIFF
--- a/performance.php
+++ b/performance.php
@@ -1,0 +1,3 @@
+<?php
+
+require_once( __DIR__ . '/performance/lastpostmodified.php' );
--- a/performance/lastpostmodified.php
+++ b/performance/lastpostmodified.php
@@ -77,5 +77,7 @@ class Last_Post_Modified {
 	}
 }
 
-// TODO: make this opt-in to start
-add_action( 'init', [ 'Automattic\VIP\Performance\Last_Post_Modified', 'init' ] );
+// Temporarily behind a constant so we can test it out on some sites first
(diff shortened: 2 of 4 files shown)


joshbetz:
I wonder if we should use a transient for this instead? I'm wondering if there's a race condition here where two post updates fire at about the same time and `bump_lastpostmodified()` completes in the wrong order. I think updating memcached would also be faster in cases where publish volume is high and this value changes a lot.

mjangda:
> I think updating memcached would also be faster in cases where publish volume is high and this value changes a lot.

Cache is definitely faster, but the downside is that we could still run into scenarios where the value is flushed/missing, which would trigger the slow queries when generating feeds.

To get around the issue of too many updates, we could add a cache-based lock that only allows an update every `n` seconds?

> I'm wondering if there's a race condition here where two post updates fire at about the same time and bump_lastpostmodified() completes in the wrong order.

The true value of `lastpostmodified` is not actually that interesting/useful since it's primarily used to generate `last-modified`/`etag` headers and `lastBuildDate` for feeds (especially if, with the race condition, the difference is seconds apart).

joshbetz:
Good points.

> To get around the issue of too many updates, we could add a cache-based lock that only allows an update every n seconds?

Yeah, that sounds good.

mjangda:
Lock (+ tests) added and restricted behind a constant for now.

mjangda:
Deployed in r85396-deploy

PR #404, opened 2016-12-16
Adding the 'From' header to the wpcom-helper file

emrikol:
It seems that WordPress.com doesn't set the From header automatically,
and it shows as the emails are from "unknown sender"


mjangda:
r85521-deploy

Issue #402, opened 2016-12-15
PHP7 Deprecation Notices


mjangda:
Flagged on Travis:

```
Deprecated:  Methods with the same name as their class will not be constructors in a future version of PHP; DisqusAPI has a deprecated constructor in ./shared-plugins/disqus/lib/api/disqus/disqus.php on line 59

Deprecated: Methods with the same name as their class will not be constructors in a future version of PHP; DisqusAPI has a deprecated constructor in ./shared-plugins/disqus/lib/api/disqus/disqus.php on line 59

Deprecated:  Methods with the same name as their class will not be constructors in a future version of PHP; JSON has a deprecated constructor in ./shared-plugins/disqus/lib/api/disqus/json.php on line 64

Deprecated: Methods with the same name as their class will not be constructors in a future version of PHP; JSON has a deprecated constructor in ./shared-plugins/disqus/lib/api/disqus/json.php on line 64

Deprecated:  Methods with the same name as their class will not be constructors in a future version of PHP; DisqusWordPressAPI has a deprecated constructor in ./shared-plugins/disqus/lib/wpapi.php on line 32

Deprecated: Methods with the same name as their class will not be constructors in a future version of PHP; DisqusWordPressAPI has a deprecated constructor in ./shared-plugins/disqus/lib/wpapi.php on line 32

Deprecated:  Methods with the same name as their class will not be constructors in a future version of PHP; MediaPass has a deprecated constructor in ./shared-plugins/mediapass/mediapass_api.php on line 35

Deprecated: Methods with the same name as their class will not be constructors in a future version of PHP; MediaPass has a deprecated constructor in ./shared-plugins/mediapass/mediapass_api.php on line 35

Deprecated:  Methods with the same name as their class will not be constructors in a future version of PHP; WP_Import has a deprecated constructor in ./wordpress-importer/wordpress-importer.php on line 38

Deprecated: Methods with the same name as their class will not be constructors in a future version of PHP; WP_Import has a deprecated constructor in ./wordpress-importer/wordpress-importer.php on line 38

```

PR #412, opened 2016-12-21
Update wordpress-importer to v0.6.3

mjangda:
[Changelog](https://wordpress.org/plugins/wordpress-importer/changelog/) for 0.6.2 + 0.6.3:

- Add support for import term metadata.
- Fix bug that caused slashes to be stripped from imported content.
- Fix bug that caused characters to be stripped inside of CDATA in some cases.
- Fix PHP notices.
- Add wp_import_existing_post filter. See: https://core.trac.wordpress.org/ticket/33721


mjangda:
Fixes #402

mjangda:
r85568-deploy

PR #416, opened 2017-01-02
Upgrade VaultPress to 1.8.5

mjangda:
* Delete plugin option when plugin is deleted via admin area.
* Fix horizontal scroll bar on the fresh installation settings page at high resolutions.


mjangda:
r86198-deploy

PR #418, opened 2017-01-03
Only force https REST API URLs on Go environments

mjangda:
It's awkward to disable locally since you need local modifications to mu-plugins, which makes updates a pain. Only add the filter when we are in a Go-server environment (i.e. VIP_GO_ENV !== false) to make this easier.

### DIFF
--- a/misc.php
+++ b/misc.php
@@ -129,11 +129,13 @@ if ( defined( 'VIP_CUSTOM_PINGS' ) && true === VIP_CUSTOM_PINGS ) {
  *
  * Filter `rest_url` to always return the https:// version
  *
- * If this must be disabled for local development, the filter
- * can be removed, but be aware that HTTPS is enforced at the web server
- * level in production, meaning non-HTTPS API calls will result in a 406 error.
+ * This is disabled for local development, but be aware
+ * that HTTPS is enforced at the web server level in production,
+ * meaning non-HTTPS API calls will result in a 406 error.
  */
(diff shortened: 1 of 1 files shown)


misc.php
```diff
@@ -129,11 +129,13 @@ function _wpcom_vip_maybe_clear_alloptions_cache( $option ) {
...
  *
- * If this must be disabled for local development, the filter
- * can be removed, but be aware that HTTPS is enforced at the web server
- * level in production, meaning non-HTTPS API calls will result in a 406 error.
+ * This is disabled for local development, but be aware
+ * that HTTPS is enforced at the web server level in production,
+ * meaning non-HTTPS API calls will result in a 406 error.
  */
-add_filter( 'rest_url', '_vip_filter_rest_url_for_ssl' );
+if ( defined( 'VIP_GO_ENV' ) && false !== VIP_GO_ENV ) {
```
nickdaugherty: The better constant to use would be `WPCOM_IS_VIP_ENV` - `VIP_GO_ENV` is intended to identify the environment name of the site, such as `production`, `preprod`, `develop`, etc...rather than indicate if the site is in a VIP environment or not.

Currently `VIP_GO_ENV` is being used as a replacement for `WPCOM_IS_VIP_ENV` in a few spots (and gets set to `false` in local environments) but I'd like to see us use it purely for environment identification rather than to identify _where_ the code is running.

mjangda: That sounds reasonable. Updated.

mjangda:
r86386-deploy

PR #422, opened 2017-01-06
Simplify Lock

mjangda:
Instead of `cache.get => cache.add` for locking, just rely on `cache.add` since it's atomic and bails if the lock is already set. This also reduces races due to https://en.wikipedia.org/wiki/Time_of_check_to_time_of_use


mjangda:
r86902-deploy

PR #424, opened 2017-01-11
Updated all submodule references from git to https

### LINKED ISSUE #298: Consistent use of `git` or `https` protocol for submodules
Right now, our submodules use a random mix of `git` and `https` protocols. This isn't breaking anything that I'm aware of, but it'd still be nice to be consistent.

https://github.com/Automattic/vip-go-mu-plugins/blob/f9e87113dea10fc77f5eaa1d5bacc3490c24280e/.gitmodules

mdrovdahl:
Fixes #298

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,42 +1,44 @@
 [submodule "vip-dashboard"]
-	path = vip-dashboard
-	url = git@github.com:Automattic/vip-dashboard.git
+    path = vip-dashboard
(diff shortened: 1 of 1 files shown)


simonwheatley:
I don't think we should do this, see [my comment here](https://github.com/Automattic/vip-go-mu-plugins/issues/298#issuecomment-272146679).

PR #426, opened 2017-01-13
Swap bool uses of VIP_GO_ENV to WPCOM_IS_VIP_ENV

### LINKED ISSUE #354: VIP_GO_ENV is not a boolean
In a few spots, we're treating the `VIP_GO_ENV` constant as a boolean, but this is the environment name.

For clarity, it should be set to something like `local` in dev environments. Anywhere that is checking this constant should be updated to not treat it as a boolean.

mjangda:
VIP_GO_ENV isn't really meant to be used as a boolean value; More as a
way to differentiate Go-environments.

WPCOM_IS_VIP_ENV is a better choice here.

Fixes #354

### DIFF
--- a/001-cron.php
+++ b/001-cron.php
@@ -58,7 +58,7 @@ class WPCOM_VIP_Cron_Control {
 
 // Allow testing of new approach to cron execution
 $whitelisted_sites = array();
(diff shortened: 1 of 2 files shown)


mjangda:
r87239-deploy

PR #429, opened 2017-01-19
Bump msm-sitemap to 1d0c1d0

mjangda:
Includes:
- handling non-existent dates
- better escaping of URLs
- query optimization for when getting a list of modified posts


mjangda:
r88446-deploy

PR #430, opened 2017-01-19
Adding `vip-go-users` WP-CLI file

emrikol:
This file currently only has one command:

`wp vip-go-users change-username`

This will allow username changes using a similar approach that we use on
WordPress.com

### DIFF
--- a/wp-cli/vip-users.php
+++ b/wp-cli/vip-users.php
@@ -1,0 +1,69 @@
+<?php
+class VIP_Go_Users_Command extends WPCOM_VIP_CLI_Command {
+	/**
+	 * Changes the WordPress username (user_login)
+	 *
+	 * Replicates functionality from WordPress.com's NameChange::rename_user()
+	 *
+	 * ## OPTIONS
+	 *
+	 * <username>
+	 * : The existing username of the user to rename.
+	 *
+	 * <newname>
+	 * : The new username to use.
+	 *
+	 * eg.: `wp vip-go-users change-username jsmith johns`
+	 *
+	 * @subcommand change-username
+	 */
+	public function change_username( $args, $assoc_args ) {
+		list( $username, $newname ) = $args;
+
+		$current_user = get_user_by( 'login', $username );
+
+		if ( false === $current_user ) {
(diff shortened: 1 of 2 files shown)


mjangda:
As mentioned in Slack, something like `username_changer_process` in https://plugins.svn.wordpress.org/username-changer/trunk/includes/functions.php might be a better approach to take here.

wp-cli/vip-users.php
```diff
@@ -0,0 +1,155 @@
...
+				// Update username!
+				$query_user_login = $wpdb->prepare( "UPDATE $wpdb->users SET user_login = %s WHERE user_login = %s", $new_username, $old_username );
+
+				if ( false !== $wpdb->query( $query_user_login ) ) {
+					// Update user_nicename
+					$query_user_nicename = $wpdb->prepare( "UPDATE $wpdb->users SET user_nicename = %s WHERE user_login = %s AND user_nicename = %s", $new_username, $new_username, $old_username );
+					$wpdb->query( $query_user_nicename );
+
+					// Update display_name
+					$query_display_name = $wpdb->prepare( "UPDATE $wpdb->users SET display_name = %s WHERE user_login = %s AND display_name = %s", $new_username, $new_username, $old_username );
```
philipjohn: We should probably only change display name if it also matches `$old_username`, just like the nickname code does below.

wp-cli/vip-users.php
```diff
@@ -0,0 +1,155 @@
...
+					$wpdb->query( $query_display_name );
+
+					// Update nickname
+					$nickname = get_user_meta( $user_id, 'nickname', true );
+					if ( $nickname === $old_username ) {
+						update_user_meta( $user_id, 'nickname', $new_username );
+					}
+
+					// If the user is a Super Admin, update their permissions
+					if ( is_multisite() && is_super_admin( $user_id ) ) {
```
philipjohn: Do we also need to remove SA rights from the old username? Otherwise could someone sign up with the old username and get SA caps?

mjangda:
Going to close this out since there isn't a pressing need for this functionality.

PR #431, opened 2017-01-20
Allow Core's cron behaviour to be restored

ethitter:
Depending on site's needs, we offer two approaches to improving Core's cron reliability. In some cases, such as local development environments or private multisite subsites, we aren't able to execute either alternative approach, so Core's offering should be used.

### DIFF
--- a/001-cron.php
+++ b/001-cron.php
@@ -56,6 +56,16 @@ class WPCOM_VIP_Cron_Control {
 	}
 }
 
+// Non-public multisites can't be reached by our cron runners, so should use Core's native approach
+if ( is_multisite() ) {
+	$details = get_blog_details( get_current_blog_id(), false );
+
+	// get_blog_details() uses numeric strings for backcompat
+	if ( '1' !== $details->public || in_array( '1', array( $details->archived, $details->spam, $details->deleted ), true ) ) {
+		add_filter( 'wpcom_vip_go_use_core_cron', '__return_true' );
+	}
+}
+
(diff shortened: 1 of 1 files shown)


001-cron.php
```diff
@@ -56,6 +56,21 @@ public function block_admin_page() {
...
 
+// Non-public multisites can't be reached by our cron runners, so should use Core's native approach
+if ( is_multisite() ) {
+	$details = get_blog_details( get_current_blog_id(), false );
+
+	// get_blog_details() uses numeric strings for backcompat
+	if ( '1' !== $details->public || in_array( '1', array( $details->archived, $details->spam, $details->deleted ), true ) ) {
+		add_filter( 'wpcom_vip_go_use_core_cron', '__return_true' );
+	}
+}
```
mjangda: Can we just move this whole block into a function instead of the filter-based approach?

ethitter: Specifically for disabling our cron overrides in this situation, or for the disabling in general?

mjangda: > Specifically for disabling our cron overrides in this situation

Just our cron overrides. Something like this:

```
function wpcom_vip_can_use_alternate_cron() {
    $details = get_blog_details( get_current_blog_id(), false );

	// get_blog_details() uses numeric strings for backcompat
	return '1' === $details->public && ! in_array( '1', array( $details->archived, $details->spam, $details->deleted ), true );
}

if ( ! wpcom_vip_can_use_alternate_cron() ) {
    return;
}
```

PR #433, opened 2017-01-20
CLI scripts to automate Jetpack connections

mjangda:
Can help easily set up VaultPress, Akismet, and Jetpack connections if we have the valid keys.

### DIFF
--- a/000-vip-init.php
+++ b/000-vip-init.php
@@ -34,7 +34,7 @@ if ( ! defined( 'WPCOM_IS_VIP_ENV' ) ) {
 
 define( 'WPCOM_VIP_MACHINE_USER_LOGIN', 'wpcomvip' );
 define( 'WPCOM_VIP_MACHINE_USER_NAME', 'WordPress.com VIP' );
-define( 'WPCOM_VIP_MACHINE_USER_EMAIL', 'vip-support@wordpress.com' );
+define( 'WPCOM_VIP_MACHINE_USER_EMAIL', 'vip@matticspace.com' );
 define( 'WPCOM_VIP_MACHINE_USER_ROLE', 'administrator' );
 
 $hostname = gethostname();
--- a/000-vip-init.php
+++ b/000-vip-init.php
@@ -29,8 +29,13 @@ if ( ! defined( 'VIP_GO_ENV' ) ) {
 
 // For backwards compatibility - always true.
 if ( ! defined( 'WPCOM_IS_VIP_ENV' ) ) {
-    define( 'WPCOM_IS_VIP_ENV', false );
+	define( 'WPCOM_IS_VIP_ENV', false );
 }
+
+define( 'WPCOM_VIP_MACHINE_USER_LOGIN', 'wpcomvip' );
+define( 'WPCOM_VIP_MACHINE_USER_NAME', 'WordPress.com VIP' );
+define( 'WPCOM_VIP_MACHINE_USER_EMAIL', 'vip-support@wordpress.com' );
+define( 'WPCOM_VIP_MACHINE_USER_ROLE', 'administrator' );
 
 $hostname = gethostname();
 define( 'WPCOM_SANDBOXED', false !== strpos( $hostname, '_web_dev_' ) );
--- a/jetpack-start/jetpack-start.php
+++ b/jetpack-start/jetpack-start.php
@@ -1,0 +1,12 @@
+<?php
+
+/**
+ * Plugin Name: VIP Jetpack Start
+ * Description: Jetpack connections made easy; just put your key in the ignition and go! 
+ * Version: 0.1
+ * Author: Automattic
+ */
+
+if ( defined( 'WP_CLI' ) && WP_CLI ) {
+	require_once( __DIR__ . '/wp-cli-keys.php' );
+}
--- a/jetpack-start/jetpack-start.php
+++ b/jetpack-start/jetpack-start.php
@@ -2,11 +2,13 @@
 
 /**
  * Plugin Name: VIP Jetpack Start
- * Description: Jetpack connections made easy; just put your key in the ignition and go! 
+ * Description: Jetpack connections made easy; just put your key in the ignition and go!
  * Version: 0.1
  * Author: Automattic
  */
 
-if ( defined( 'WP_CLI' ) && WP_CLI ) {
+// Jetpack Start can triggered via A8C servers only
+if ( defined( 'WP_CLI' ) && WP_CLI && true == A8C_PROXIED_REQUEST ) {
 	require_once( __DIR__ . '/wp-cli-keys.php' );
+	require_once( __DIR__ . '/wp-cli-api.php' );
 }
--- a/jetpack-start/jetpack-start.php
+++ b/jetpack-start/jetpack-start.php
@@ -7,7 +7,6 @@
  * Author: Automattic
  */
 
-// Jetpack Start can triggered via A8C servers only
 if ( defined( 'WP_CLI' ) && WP_CLI ) {
 	require_once( __DIR__ . '/wp-cli-keys.php' );
 	require_once( __DIR__ . '/wp-cli-api.php' );
--- a/jetpack-start/jetpack-start.php
+++ b/jetpack-start/jetpack-start.php
@@ -8,7 +8,7 @@
  */
 
 // Jetpack Start can triggered via A8C servers only
-if ( defined( 'WP_CLI' ) && WP_CLI && true == A8C_PROXIED_REQUEST ) {
+if ( defined( 'WP_CLI' ) && WP_CLI ) {
 	require_once( __DIR__ . '/wp-cli-keys.php' );
 	require_once( __DIR__ . '/wp-cli-api.php' );
 }
--- a/jetpack-start/wp-cli-api.php
+++ b/jetpack-start/wp-cli-api.php
@@ -76,9 +76,10 @@ class Jetpack_Start_API_CLI_Command extends WP_CLI_Command {
 		$data = $this->api_fetch_keys( $user );
 		if ( is_wp_error( $data ) ) {
 			$message = $data->get_error_message();
-			if ( self::API_ERROR_EXISTING_SUBSCRIPTION === $message ) {
+			// Use strpos because our API method appends stuff to the error message
+			if ( false !== strpos( $message, self::API_ERROR_EXISTING_SUBSCRIPTION ) ) {
 				$message = 'There is an existing Jetpack Start subcription for this site. Please disconnect using the `cancel` subcommand and try again.';
-			} elseif ( self::API_ERROR_USER_PERMISSIONS === $message ) {
+			} elseif ( false !== strpos( $message, self::API_ERROR_USER_PERMISSIONS ) ) {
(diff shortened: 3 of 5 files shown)


jetpack-start/wp-cli-keys.php
```diff
@@ -0,0 +1,218 @@
+<?php
```
mjangda: This script may eventually end up in Jetpack itself.

simonwheatley:
Should `--skip_is_active` be `--force` or `--force-reconnection`?

I just ran this on a blog within a multisite. It seems like it worked:

screen shot 2017-03-23 at 14 24 14

…but I got some odd output:

[output.txt](https://github.com/Automattic/vip-go-mu-plugins/files/865003/output.txt)

Running it on another site, I got:

``` bash
$ wp jetpack-start api connect
-- Verifying VIP machine user exists (or creating one, if not)
-- Fetching keys from Jetpack Start API
Error: Failed to fetch keys from Jetpack Start: got non-200 response from jetpack start api (code: 400 | body: {"error":"invalid_input","message":"Failed to provision WPCOM store subscription"})
```

jetpack-start/wp-cli-api.php
```diff
@@ -0,0 +1,221 @@
...
+		}
+
+		WP_CLI::line( '-- Fetching keys from Jetpack Start API' );
+		$data = $this->api_fetch_keys( $user );
+		if ( is_wp_error( $data ) ) {
+			$message = $data->get_error_message();
+			if ( self::API_ERROR_EXISTING_SUBSCRIPTION === $message ) {
+				$message = 'There is an existing Jetpack Start subcription for this site. Please disconnect using the `cancel` subcommand and try again.';
+			} elseif ( self::API_ERROR_USER_PERMISSIONS === $message ) {
+				$message = 'This site already has an existing Jetpack shadow site but the VIP Machine User does not have the necessary caps. Please `add_user_to_blog` using your WP.co sandbox to continue.';
```
ethitter: Can we note which user I should add?

Also, `s/WP.co/WP.com/gi`

simonwheatley:
Should `--force` cancel a preexisting subscription? (Naming parameters is hard)

jetpack-start/wp-cli-api.php
mjangda: Yep, good idea.

jetpack-start/wp-cli-api.php
```diff
@@ -0,0 +1,221 @@
...
+		update_option( 'vip_jetpack_start_connected_on', time(), false );
+
+		WP_CLI::success( 'All done! Welcome to Jetpack! ✈️️✈️️✈️️' );
+	}
+
+	private function maybe_create_user() {
+		$user = get_user_by( 'login', WPCOM_VIP_MACHINE_USER_LOGIN );
+		if ( ! $user ) {
+			$cmd = sprintf(
+				'user create --url=%s %s %s --role=%s --display_name=%s --porcelain',
```
nickdaugherty: Can we put the named arguments before the positional arguments, then add a `--` before the positional args?

`user --url=%s --role=%s --display_name=%s --porcelain -- create %s %s`

It's not a huge deal because the positional args are constants in this case, but `--` tells the CLI to not interpret anything after it as an option. It helps prevent values like `--some_other_option=true` from injecting options, which `escapeshellarg()` does not actually prevent against.

jetpack-start/wp-cli-api.php
```diff
@@ -0,0 +1,221 @@
...
+
+		$user_id = $user->ID;
+		if ( is_multisite() ) {
+			add_user_to_blog( get_current_blog_id(), $user_id, WPCOM_VIP_MACHINE_USER_ROLE );
+
+			if ( ! is_super_admin( $user_id ) ) {
+				grant_super_admin( $user_id );
+			}
+		} else {
+			if ( ! $user->has_cap( WPCOM_VIP_MACHINE_USER_ROLE ) ) {
```
nickdaugherty: I think it would be better to check `$user->roles` than `has_cap()`, see https://core.trac.wordpress.org/ticket/22624#comment:10

jetpack-start/wp-cli-api.php
```diff
@@ -0,0 +1,221 @@
...
+				$user->set_role( WPCOM_VIP_MACHINE_USER_ROLE );
+			}
+		}
+
+		return $user;
+	}
+
+	private function api_request( $url, $args ) {
+		$headers = array_merge( [], $this->get_api_auth_header() );
```
nickdaugherty: I'm not sure what the `array_merge()` is doing here, might be worth an inline comment to clarify.

jetpack-start/wp-cli-api.php
```diff
@@ -0,0 +1,221 @@
...
+		$jetpack_id = $data['jetpack_id'] ?? '';
+		$jetpack_secret = $data['jetpack_secret'] ?? '';
+		$jetpack_access_token = $data['jetpack_access_token'] ?? '';
+		$jetpack_result = WP_CLI::runcommand( sprintf(
+			'jetpack-start keys jetpack --jetpack_id=%s --jetpack_secret=%s --jetpack_access_token=%s --user=%s',
+			escapeshellarg( $jetpack_id ),
+			escapeshellarg( $jetpack_secret ),
+			escapeshellarg( $jetpack_access_token ),
+			escapeshellarg( WPCOM_VIP_MACHINE_USER_LOGIN )
+		), $runcommand_args );
```
nickdaugherty: Do we need to be handling errors from these CLI commands, or will the whole process exit if they fail?

jetpack-start/wp-cli-keys.php
```diff
@@ -0,0 +1,196 @@
...
+			WP_CLI::error( '- Failed to configure Akismet: ' . $result->get_error_message() );
+		}
+
+		WP_CLI::success( 'Akismet configured successfully.' );
+	}
+
+	private function register_akismet( $key ) {
+		update_option( 'wordpress_api_key', $key );
+
+		$verification_response = Akismet::verify_key( $key );
```
nickdaugherty: Should we be verifying _before_ updating the option?

jetpack-start/wp-cli-keys.php
```diff
@@ -0,0 +1,196 @@
...
+		// TODO: set up WP=>VaultPress SSH access
+
+		WP_CLI::success( 'VaultPress configured successfully.' );
+	}
+
+	private function register_vaultpress( $api_key ) {
+		return VaultPress::register( $api_key );
+	}
+
+	private function install_vaultpress_ssh_keys( $ssh_key ) {
```
nickdaugherty: How is this planned to be used in practice? Obviously we need to take great care with it. Maybe we can leave it off on VIP, since it won't work anyway?

mjangda:
> Should --force cancel a preexisting subscription? (Naming parameters is hard)

That's probably something we should do at the API level since running the `cancel` needs Jetpack active or you have to provide the site ID.

jetpack-start/wp-cli-api.php
```diff
@@ -0,0 +1,221 @@
...
+		$user = $this->maybe_create_user();
+		if ( is_wp_error( $user ) ) {
+			WP_CLI::error( $user->get_error_message() );
+		}
+
+		WP_CLI::line( '-- Fetching keys from Jetpack Start API' );
+		$data = $this->api_fetch_keys( $user );
+		if ( is_wp_error( $data ) ) {
+			$message = $data->get_error_message();
+			if ( self::API_ERROR_EXISTING_SUBSCRIPTION === $message ) {
```
mjangda: Unfortunately not. The error code returned from the API are pretty generic. I'll see if we can update them on the API side.

jetpack-start/wp-cli-api.php
mjangda: They run as independent processes and output errors on screen on failure. Probably something we'll need to deal with when we wire up to run automatically via the API but okay for now.

jetpack-start/wp-cli-keys.php
mjangda: I left it in because the plan is to upstream this file to Jetpack.

jetpack-start/jetpack-start.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+/**
+ * Plugin Name: VIP Jetpack Start
+ * Description: Jetpack connections made easy; just put your key in the ignition and go!
+ * Version: 0.1
+ * Author: Automattic
+ */
+
+// Jetpack Start can triggered via A8C servers only
```
mjangda: Removed comment.

jetpack-start/wp-cli-keys.php
mjangda: Good catch.

000-vip-init.php
```diff
@@ -29,9 +29,14 @@
 
 // For backwards compatibility - always true.
 if ( ! defined( 'WPCOM_IS_VIP_ENV' ) ) {
-    define( 'WPCOM_IS_VIP_ENV', false );
+	define( 'WPCOM_IS_VIP_ENV', false );
 }
 
+define( 'WPCOM_VIP_MACHINE_USER_LOGIN', 'wpcomvip' );
+define( 'WPCOM_VIP_MACHINE_USER_NAME', 'WordPress.com VIP' );
+define( 'WPCOM_VIP_MACHINE_USER_EMAIL', 'vip@matticspace.com' );
```
mjangda: 🤷‍♂️

jetpack-start/wp-cli-api.php
```diff
@@ -0,0 +1,221 @@
...
+			'Authorization' => 'Bearer ' . WPCOM_VIP_JP_START_API_TOKEN,
+		];
+	}
+
+	private function get_api_url( $endpoint ) {
+		return WPCOM_VIP_JP_START_API_BASE . $endpoint;
+	}
+}
+
+WP_CLI::add_command( 'jetpack-start api', 'Jetpack_Start_API_CLI_Command' );
```
mjangda: Yep, done.

jetpack-start/wp-cli-api.php
mjangda: Done.

jetpack-start/wp-cli-api.php
mjangda: Done.

jetpack-start/wp-cli-api.php
mjangda: TIL; done.

jetpack-start/wp-cli-keys.php
```diff
@@ -0,0 +1,196 @@
...
+		WP_CLI::success( 'Jetpack configured successfully.' );
+	}
+
+	private function register_jetpack( $user_id, $args ) {
+		$jetpack_id = (int) $args['jetpack_id'];
+		$jetpack_secret = (string) $args['jetpack_secret'];
+		$jetpack_access_token = (string) $args['jetpack_access_token'];
+		$jetpack_is_private = (bool) WP_CLI\Utils\get_flag_value( $args, 'jetpack_is_private', false );
+
+		Jetpack_Options::update_options(
```
seanosh: One potential pitfall I've run into with `Jetpack_Options::update_options` is that it essentially calls `update_option` on each provided key/value pair, but then masks the `true`/`false` boolean returned from each. So one or more options can fail updating (for whatever reason) and the command output (as implemented) wouldn't have any output or indication of failure.

mjangda: Good point. It would be nicer if Jetpack had a `register` method (similar to `VaultPress::register`) that would accept these args instead of us manually updating the options.

mjangda:
r90563-deploy

PR #436, opened 2017-01-25
Disallow flv files from being uploaded

### LINKED ISSUE #435: Block FLV uploads
Flash isn't allowed on VIP. We need to block `.flv` files from being uploaded.

joshbetz:
Fixes #435

### DIFF
--- a/misc.php
+++ b/misc.php
@@ -8,7 +8,7 @@ License: GPL version 2 or later - http://www.gnu.org/licenses/old-licenses/gpl-2
 */
 
 add_filter( 'upload_mimes', function( $mimes ) {
-	unset( $mimes['flv'] ;
+	unset( $mimes['flv'] );
 	return $mimes;
 });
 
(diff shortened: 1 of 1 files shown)


misc.php
```diff
@@ -7,6 +7,11 @@
 License: GPL version 2 or later - http://www.gnu.org/licenses/old-licenses/gpl-2.0.html
 */
 
+add_filter( 'upload_mimes', function( $mimes ) {
+	unset( $mimes['flv'] );
+	return $mimes;
+});
```
nickdaugherty: Should we set the priority really high, so that other client code can't come along and re-enable it?

joshbetz: I guess that wouldn't hurt, but obviously malicious code could just set the priority higher than we do.

joshbetz: Or we could set the priority to `PHP_INT_MAX` ;)

nickdaugherty: Yeah they could always 1-up us - though us setting a high priority is also a signal to devs that this is meant to be the final say and not overridden, and makes an override more deliberate and perhaps easier to detect.

nickdaugherty:
🎉

joshbetz:
r87840-deploy

PR #437, opened 2017-01-25
Whitelisted Proxy IP Forwarding

mjangda:
This allows more securely forwarding the origin IP address. It requires a trusted whitelist of proxy IP addresses (both IPv4 -- exact and CIDR -- and IPv6 are supported).

Assumes that the local proxy passes through a server var with comma-separated IPs in the form of `user_ip, remote_proxy_ip`.

In `vip-config/vip-config`:

```
$proxy_lib = ABSPATH . '/wp-content/mu-plugins/lib/proxy/ip-forward.php';
if ( file_exists( $proxy_lib ) && ! empty( $_SERVER['HTTP_X_SECRET_TRAIL_KEY'] ) ) {
    require_once( ABSPATH . '/wp-content/mu-plugins/lib/proxy/ip-forward.php' );

    $ip_whitelist = [ '1.2.3.4', '5.6.7/8' ];
    Automattic\VIP\Proxy\fix_remote_address_from_ip_trail( $_SERVER['HTTP_X_SECRET_TRAIL_KEY'], $ip_whitelist );
    unset( $ip_whitelist );
}
```

### DIFF
--- a/lib/proxy/ip-forward.php
+++ b/lib/proxy/ip-forward.php
@@ -1,0 +1,53 @@
+<?php
+
+namespace Automattic\VIP\Proxy;
+
+/**
+ * This allows more securely forwarding the origin IP address when there are multiple proxies in play.
+ *
+ * Example setup:
+ * User => Remote Proxy (e.g. Cloudflare) => Local Proxy (Varnish) => Application (PHP/WP)
+ *
+ * Without this, the Application will see the Remote Proxy's IP address as the REMOTE_ADDR.
+ * With this, if Remote Proxy's IP address matches a known whitelist, the Application will see the User's real IP address as REMOTE_ADDR.
+ *
+ * Only two levels of proxies are supported.
+ *
+ * @param (string) $ip_trail Comma-separated list of IPs (something like `user_ip, proxy_ip`)
+ * @param (string|array) $proxy_ip_whitelist Whitelisted IP addresses for the remote proxy. Supports IPv4 and IPv6, including CIDR format.
+ *
+ * @return (bool) true, if REMOTE_ADDR updated; false, if not.
+ */
+function fix_remote_address_from_ip_trail( $ip_trail, $proxy_ip_whitelist ) {
+	// If X-Forwarded-For is not set, we're not dealing with a remote proxy or something in the proxy configs is doing it wrong.
+	if ( ! isset( $_SERVER['HTTP_X_FORWARDED_FOR'] ) ) {
+		return false;
+	}
+
+	// Verify that the IP trail has multiple IPs but only two levels (remote + local).
+	$ip_addresses = explode( ',', $ip_trail );
+	$ip_addresses = array_map( 'trim', $ip_addresses );
+	if ( 2 !== count( $ip_addresses ) ) {
+		return false;
+	}
+
+	list( $user_ip, $remote_proxy_ip ) = $ip_addresses;
+
+	// This should probably never happen, but validate just in case.
+	if ( $remote_proxy_ip !== $_SERVER['HTTP_X_FORWARDED_FOR'] ) {
+		return false;
+	}
+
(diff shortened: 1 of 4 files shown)


lib/proxy/ip-forward.php
```diff
@@ -0,0 +1,53 @@
...
+	list( $user_ip, $remote_proxy_ip ) = $ip_addresses;
+
+	// This should probably never happen, but validate just in case.
+	if ( $remote_proxy_ip !== $_SERVER['HTTP_X_FORWARDED_FOR'] ) {
+		return false;
+	}
+
+	require_once( __DIR__ . '/ip-utils.php' );
+
+	// Verify that our remote proxy matches out whitelist
```
nickdaugherty: Small typo - `s/out/our`

lib/proxy/ip-forward.php
```diff
@@ -0,0 +1,53 @@
...
+	$is_whitelisted_proxy_ip = IpUtils::checkIp( $remote_proxy_ip, $proxy_ip_whitelist );
+	if ( ! $is_whitelisted_proxy_ip ) {
```
nickdaugherty: An extra line between these statements would help readability.

lib/proxy/ip-utils.php
```diff
@@ -0,0 +1,136 @@
...
+            $left = ($left <= 16) ? $left : 16;
+            $mask = ~(0xffff >> $left) & 0xffff;
+            if (($bytesAddr[$i] & $mask) != ($bytesTest[$i] & $mask)) {
+                return false;
+            }
+        }
+
+        return true;
+    }
+}
```
nickdaugherty: Even though this was brought in from a 3rd party, what do you think about cleaning it up with the WP coding standards?

tests/test-lib-proxy-ip-forward.php
```diff
@@ -0,0 +1,90 @@
...
+	public function test__fix_remote_address__ip_in_whitelist() {
+		$_SERVER['HTTP_X_FORWARDED_FOR'] = '5.6.7.8';
+		$ip_trail = '1.2.3.4, 5.6.7.8';
+		$whitelist = [ '5.6.7.8' ];
+
+		$result = fix_remote_address_from_ip_trail( $ip_trail, $whitelist );
+
+		$this->assertTrue( $result );
+		$this->assertEquals( '1.2.3.4', $_SERVER['REMOTE_ADDR'] );
+	}
```
nickdaugherty: Should we add a test for invalid looking IP addresses, like `bad_ip` and an int like `12345689` (which could come from something like `ip2long()`)? That should never happen, but the internet is a strange place.

lib/proxy/ip-utils.php
mjangda: I've left it as-is (except for a small documentation update to add a link) to allow for easier updates.

tests/test-lib-proxy-ip-forward.php
mjangda: That's technically out-of-scope for this function as it would be handled by the IpUtils class, but I guess it wouldn't hurt to include a test here (since we are not running related tests for the IpUtils class)

lib/proxy/ip-utils.php
nickdaugherty: Alright, makes sense.

tests/test-lib-proxy-ip-forward.php
mjangda: Oh, I guess we don't technically validate that the user_ip in the trail is valid either, which should probably be happening.

nickdaugherty: Exactly - what we need to know is that given invalid input, the code still behaves as expected (no errors, no bypassing the protections). We're assuming that the IpUtils handles this correctly, but we don't _know_.

mjangda:
r87473-deploy

PR #438, opened 2017-01-25
432 alloptions size discrepency

### LINKED ISSUE #432: Alloptions: Disrepency in size counts
Email says "options is up to 786 KB [...] size when serialized: 805253"

wp-cli shows: 

```
Total size of all option values for this blog: 432 KB
Size of serialized alloptions for this blog: 786 KB
```

We should make sure the values are consistent across the board.

nickdaugherty:
Tweaks to improve the consistency of reported sizes.

FIxes #432


ethitter:
👍

mjangda:
Isn't `mb_strlen` a better option since that's the charset for our `wp_options` tables?

nickdaugherty:
It's my understanding that `mb_strlen()` counts multibyte characters as a single byte (to return an accurate string length). We don't actually care about string length, we want to know how the raw byte size of the data, without any fancy mb handling because there is no special size handling when it goes to memcache.

mjangda:
Ah, that makes sense :)

nickdaugherty:
`r87462-deploy`

PR #443, opened 2017-01-27
Validate user_ip

### LINKED ISSUE #442: Proxy Lib: Validate that user_ip is a valid IP address
From #437 suggested by @nickdaugherty:

> add a test for invalid looking IP addresses, like bad_ip and an int like 12345689 (which could come from something like ip2long())? That should never happen, but the internet is a strange place.

mjangda:
Make sure we're sent through a valid IPv4 or IPv6 address.

"Because the internet is a weird place"

Fixes #442


mjangda:
r87473-deploy

PR #445, opened 2017-02-01
Add meta_key_value index

joshbetz:
Add an index on the postmeta.meta_value for all new sites.

### DIFF
--- a/schema.php
+++ b/schema.php
@@ -1,0 +1,20 @@
+<?php
+
+/*
+ * Schema Customizations
+ */
+
+add_filter( 'dbdelta_create_queries', function( $queries ) {
+	global $wpdb;
+	
+	foreach( $queries as $k => $q ) {
+	
+		// Add meta_key_value index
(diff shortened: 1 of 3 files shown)


mjangda:
@joshbetz I thought about it some more and decided that this is probably the best approach :)

I added a test so we can catch if/when our `dbDelta` addition fails. Also adds a prefix to the index name and removes the default `meta_key` index.

ethitter:
Tested fine in my local environment. :shipit:

joshbetz:
Looks good to me too. I can't "review" this because it's my PR.

> I added a test so we can catch if/when our dbDelta addition fails.

Yeah, that's a good call.

joshbetz:
Once we test the index on some existing sites, I'll merge and deploy.

mjangda:
I think we can merge this now; INDEX has been live on a bunch of sites now.

joshbetz:
r89253-deploy

PR #451, opened 2017-02-02
Make Debug Bar available to all proxied requests

mjangda:
- [x] Need to send no-cache headers
- [ ] Make it work without admin-bar
- [ ] Test, test, test

### DIFF
--- a/debug-bar.php
+++ b/debug-bar.php
@@ -17,7 +17,7 @@ add_filter( 'debug_bar_enable', function( $enable ) {
 
 add_action( 'send_headers', function() {
 	if ( true === A8C_PROXIED_REQUEST ) {
-		header( 'Cache-Control: private' );
+		nocache_headers();
 	}
 });
 
--- a/debug-bar.php
+++ b/debug-bar.php
@@ -8,8 +8,9 @@
  Author URI: http://wordpress.org/
  */
 add_filter( 'debug_bar_enable', function( $enable ) {
-	$enable = is_automattician();
-
+	if ( true === A8C_PROXIED_REQUEST ) {
+		return true;
+	}
 	return $enable;
 }, 99 );
 
--- a/debug-bar.php
+++ b/debug-bar.php
@@ -7,12 +7,19 @@
  Version: 0.8.2
  Author URI: http://wordpress.org/
  */
+
 add_filter( 'debug_bar_enable', function( $enable ) {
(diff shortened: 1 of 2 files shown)


joshbetz:
We should probably send nocache headers to avoid serving Debug Bar to all users.

joshbetz:
Should we also enable debug bar for all logged in admins? I've previously thought about adding a `developer` role to enable tools like this for client devs, but not sure how we'd manage that.

debug-bar.php
```diff
@@ -7,13 +7,20 @@
...
 		return true;
 	}
 	return $enable;
 }, 99 );
 
+add_action( 'send_headers', function() {
+	if ( true === A8C_PROXIED_REQUEST ) {
+		header( 'Cache-Control: private' );
+	}
+});
```
mjangda: Might be good to call `nocache_headers()` as well.

mjangda:
Do we need to do something at the Varnish level? Proxied requests may end up getting a cached page and not see the debug bar.

debug-bar.php
joshbetz: Oh yeah, that's definitely better.

mjangda:
Bug: If a user is not logged in, the admin bar isn't shown. This means that the Debug doesn't render.

mjangda:
> If a user is not logged in, the admin bar isn't shown. This means that the Debug Bar doesn't render.

On WP.com, the `debug-bar` node is removed from `$wp_admin_bar`. A new "super admin bar" is created that floats on the page, which might be the right direction for this.

joshbetz:
What if, instead of making it work without the admin bar, we showed the admin bar to logged out, proxied users?

I think we can `add_filter( 'show_admin_bar', '__return_true' );`

mjangda:
Sounds like a good workaround for now.

simonwheatley:
> We should probably send nocache headers to avoid serving Debug Bar to all users.

Is this a good idea? It seems that it would make debugging caching issues harder. Should/could we have it off by default, and enable some toggle to turn it on (or vice versa)?

mjangda:
I think as discussed in #602, it's probably worthwhile adding a button or something to the little sandbox bar/label that allows you to toggle uncached mode which could also load the Debug Bar. 

Not sure if it's useful to also stay in cached mode but still access the debug bar.

mjangda:
Closing; we have Debug Mode now.

PR #454, opened 2017-02-05
Allow unauth-ed access to /vip/ sites endpoint

mjangda:
The endpoint does not have any secret data and is used for several internal tools. This will be replaced with a more thorough auth system in the near future.


mjangda:
r87755-deploy

PR #455, opened 2017-02-06
Generate PHPDocs

simonwheatley:
Generate documentation for our MU plugins repo, and publish it to GitHub pages.

### DIFF
--- a/phpdoc.dist.xml
+++ b/phpdoc.dist.xml
@@ -1,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<phpdoc>
+	<title>WordPress.com VIP – VIP Go Function Documentation</title>
+	<parser>
+		<!-- This places the cache dirs in the root dir `/phpdoc/` for your site -->
+		<target>../../phpdoc/</target>
+		<encoding>utf8</encoding>
+		<markers>
+			<item>TODO</item>
+			<item>FIXME</item>
+		</markers>
+		<visibility></visibility>
+	</parser>
+	<transformer>
+		<!-- This places the docs in the root dir `/phpdoc/` for your site -->
+		<target>../../phpdoc/</target>
+	</transformer>
+	<logging>
+		<level>quiet</level>
+		<paths>
+			<default>{APP_ROOT}/data/log/{DATE}.log</default>
+			<errors>{APP_ROOT}/data/log/{DATE}.errors.log</errors>
+		</paths>
+	</logging>
+	<transformations>
+		<template name="clean" />
+	</transformations>
+	<files>
+		<file>lib/proxy/ip-forward.php</file>
+		<file>vip-cache-manager/api.php</file>
+		<directory>vip-helpers</directory>
+	</files>
+</phpdoc>
--- a/phpdoc.dist.xml
+++ b/phpdoc.dist.xml
@@ -26,6 +26,7 @@
 		<template name="clean" />
 	</transformations>
 	<files>
+		<file>000-vip-init.php</file>
 		<file>lib/proxy/ip-forward.php</file>
 		<file>vip-cache-manager/api.php</file>
 		<directory>vip-helpers</directory>
--- a/phpdoc.dist.xml
+++ b/phpdoc.dist.xml
@@ -3,7 +3,7 @@
 	<title>WordPress.com VIP – VIP Go Function Documentation</title>
 	<parser>
 		<!-- This places the cache dirs in the root dir `/phpdoc/` for your site -->
-		<target>../../phpdoc/</target>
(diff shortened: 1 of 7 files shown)


simonwheatley:
Generating docs on the entire repo times out on Travis, but a limited scope works:

https://automattic.github.io/vip-go-mu-plugins/

Scope:

```
--directory="${TRAVIS_BUILD_DIR}/vip-helpers/" 
--filename="${TRAVIS_BUILD_DIR}/vip-cache-manager/api.php"
```

simonwheatley:
Questions:

* [ ] Are we covering all the areas of MU plugins we (initially) want to?
* [ ] Are we happy with the implementation?

simonwheatley:
> phpDoc supports XML based config (phpdoc.xml)- it might be nice to use that to keep the config in a standardized place and out of the shell script. That would make it easier to simply run phpdoc in the dir and have it just work.

@nickdaugherty I switched as much config as I could to `phpdoc.dist.xml` (you can override this file with a local `phpdoc.xml`). How does it look for you now? 

> phpDoc also supports log levels, so maybe we could use quiet in production to help with the timeout / /dev/null stuff. https://www.phpdoc.org/docs/latest/references/configuration.html#logging

It appears this is *log* levels, not *output* levels 😞  so I've added a `--quiet` switch instead.

nickdaugherty:
Looking good @simonwheatley - I just pushed a few more commits that make it even more modular, so we can easily build the docs locally using `make phpdoc`, then the doc script calls that same make command.

Not sure what I broke but locally I'm getting an error like:

```
Unable to find the `dot` command of the GraphViz package. Is GraphViz correctly installed and present in your path?
```

Could just be me though - I haven't dug into it. If my changes break your work, feel free to push back / trash them :).

simonwheatley:
Currently the command is failing with:

``` bash
+make phpdoc
./vendor/phpdocumentor/phpdocumentor/bin/phpdoc run --ignore-symlinks --no-interaction --quiet --sourcecode
make: ./vendor/phpdocumentor/phpdocumentor/bin/phpdoc: Command not found
make: *** [phpdoc] Error 127
```

Perhaps we can rely on `phpdoc` being in the path? To be continued…

simonwheatley:
> Looking good @simonwheatley - I just pushed a few more commits that make it even more modular, so we can easily build the docs locally using make phpdoc, then the doc script calls that same make command.

Good plan, I got a bit tripped up by having everything building and checking out in the same location but that's all working now. 

> Not sure what I broke but locally I'm getting an error like: (graphviz error)

Yeah, you need to install Graphviz, we do this with:

``` yml
addons:
  apt:
    packages:
    - graphviz
```

@nickdaugherty I think we're good to go now. I might squash and merge, as this history is messy. 🙂

simonwheatley:
Here's the output of the Graphviz stuff, by the way: https://automattic.github.io/vip-go-mu-plugins/graphs/class.html

phpdoc.dist.xml
```diff
@@ -0,0 +1,33 @@
...
+		<encoding>utf8</encoding>
+		<markers>
+			<item>TODO</item>
+			<item>FIXME</item>
+		</markers>
+		<visibility></visibility>
+	</parser>
+	<transformer>
+		<!-- This places the docs in the root dir `/phpdoc/` for your site -->
+		<target>../../phpdoc/</target>
```
nickdaugherty: @simonwheatley what's the use case for this path? 

Running locally, it dropped the files + cache in `/Users/nickdaugherty` on my setup - I would have expected `./phpdoc`, with a `.gitignore` entry.

I think the primary use-case for local doc generation will be testing - i.e., as you're working on the mu plugins code, you regenerate the docs prior to committing, to make sure they look right, are free from typos, etc. Keeping them in the same directory probably avoids confusion.

simonwheatley: > Keeping them in the same directory probably avoids confusion.

I was trying to keep the generated docs out of the MU plugins directory, out of concern that the nested Git clones (having the `gh-pages` branch checked out inside the other branch) would create some additional unanticipated complexity. Seems fine though.

simonwheatley:
@nickdaugherty Ready again. Changes since your last review: https://github.com/Automattic/vip-go-mu-plugins/compare/cee2ce1023ae4d5e837113539aab46c6be5fab8f...258-generate-docs 🙂

* Added a Phony as the make command failed without it, due to there being a clone of the `phpdocs` directory
* Cleaned up the paths for `VIP_PHPDOC_DIR` and `VIP_DOCS_DIR`
* Simpler `phpdoc` command in the `Makefile`
* Added an ignore for the `phpdoc` subdirectory to MU plugins
* Added `000-vip-init.php`

PR #457, opened 2017-02-07
456 add blog id to sites endpoint response

### LINKED ISSUE #456: Add blog id to vip/v1/sites REST endpoint
It would be useful to have the blog id for each multisite site, for a variety of things like importing media.

https://github.com/Automattic/vip-go-mu-plugins/blob/51945764f3835142817d93bc8e1c1c0c14c2f6d6/vip-rest-api.php#L125-L127

nickdaugherty:
Adds the blog id to vip/v1/sites endpoint.

Fixes #456

### DIFF
--- a/vip-rest-api.php
+++ b/vip-rest-api.php
@@ -123,6 +123,7 @@ class WPCOM_VIP_REST_API_Endpoints {
 					}
 
 					$sites[] = array(
+						'blog_id'     => $blog_id,
 						'domain_name' => $url,
 					);
 
--- a/vip-rest-api.php
+++ b/vip-rest-api.php
@@ -106,8 +106,8 @@ class WPCOM_VIP_REST_API_Endpoints {
 			// Inflate raw list of site IDs, if available
 			if ( is_array( $_sites ) ) {
 				// Switch to the blog to ensure certain domain filtering is respected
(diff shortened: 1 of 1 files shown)


vip-rest-api.php
```diff
@@ -106,8 +106,8 @@ public function list_sites() {
 			// Inflate raw list of site IDs, if available
 			if ( is_array( $_sites ) ) {
 				// Switch to the blog to ensure certain domain filtering is respected
-				foreach ( $_sites as $_site ) {
-					switch_to_blog( $_site );
+				foreach ( $_sites as $blog_id ) {
```
ethitter: Due to the risk of stomping on the `$blog_id` global, I opted for `$_site`. I'm fine with `$_blog_id`, but we shouldn't use `$blog_id`.

nickdaugherty: Good call, will change that.

nickdaugherty: Updated in 5682e62

nickdaugherty:
@ethitter - Also worth noting that this is only adding `blog_id` to the response when the site is multisite. For consistency, we should probably do it on regular sites too - what's the right value, `1` or `null`?

nickdaugherty:
Closing this per #456.

PR #458, opened 2017-02-07
Switch to two-factor core feature plugin

joshbetz:
TODO:

* [x] Require admins to enable two factor
* [x] SMS provider
* [x] Guarantee that exactly one of these is active -- if both are active, 🔥
* [x] Force jetpack-force-2fa if Jetpack SSO is enabled?


scottsweb:
Just taken a quick look at the plugin and sharing some observations. I am guessing some of these are deliberate and some will need to be reported upstream:

* Upon activation my user privs are reduced. I see an error `Two Step Authentication is required to publish to this site.`. Clicking on this takes me to my profile but there is no heading for `Two Step Authentication`. I think we should add one and keep the language consistent. 
* I activated three: email, dummy and backup codes. The settings were saved but I was shown the error: `Two Factor Authentication not activated, you must specify authcode to ensure it is properly set up. Please re-scan the QR code and enter the code provided by your application.` which is unexpected as the QR code method was not chosen. My privs came back.

screenshot from 2017-02-24 10-49-58

* Three methods have additional options (FIDO U2F, Authenticator and Backup codes) but the all behave differently. I think we should have a `View options` toggle for each one and move the FIDO U2F section from the bottom of the screen and into the same area as the others.
* Not sure we need the link to `You can find FIDO U2F Security Key devices for sale from here.`

Visually things are looking pretty decent. A couple of suggestions:

* We use bold or a darker colour for the 2-factor option name in the table, e.g: **FIDO U2F**
* The backup method options when signing in might be a little light and could use a little bit of line height or space:

screenshot from 2017-02-24 11-00-28

Will do some more testing on the train, got to dash!

joshbetz:
These issues would address some of your issues, @scottsweb:

https://github.com/georgestephanis/two-factor/pull/145
https://github.com/georgestephanis/two-factor/issues/86
https://github.com/georgestephanis/two-factor/issues/78
https://github.com/georgestephanis/two-factor/issues/43

I think 86 and 78 are things we should definitely fix. 43 would be nice as a way to replace your TOTP device without completely removing, but even wp.com doesn't have this yet, so I'd say that's not a hard requirement.

joshbetz:
"Fixed" the U2F issue by disabling U2F for now.

twostep.php
```diff
@@ -0,0 +1,55 @@
...
+	if ( ! wpcom_vip_plugin_is_loaded( 'shared-plugins/jetpack-force-2fa' ) && ! class_exists( 'Jetpack' ) && ! Jetpack::is_active() && ! Jetpack::is_module_active( 'sso' ) ) {
+		wpcom_vip_load_plugin( 'two-factor' );
+	} else if ( class_exists( 'Jetpack' ) && Jetpack::is_active() && Jetpack::is_module_active( 'sso' ) ) {
+		wpcom_vip_load_plugin( 'jetpack-force-2fa' );
+		remove_action( 'wp_login', array( 'Two_Factor_Core', 'wp_login' ) );
+	}
+}
+add_action( 'setup_theme', 'wpcom_enable_two_factor_plugin' );
+
+function wpcom_vip_twostep_filter_caps( $caps ) {
```
mjangda: We should have an inline comment explaining why we're removing caps.

twostep.php
```diff
@@ -0,0 +1,55 @@
...
+});
+
+function wpcom_vip_force_twostep() {
+	return ! wpcom_vip_plugin_is_loaded( 'shared-plugins/jetpack-force-2fa' ) && class_exists( 'Two_Factor_Core' ) && ! Two_Factor_Core::is_user_using_two_factor();
+}
+
+function wpcom_enable_two_factor_plugin() {
+	if ( ! wpcom_vip_plugin_is_loaded( 'shared-plugins/jetpack-force-2fa' ) && ! class_exists( 'Jetpack' ) && ! Jetpack::is_active() && ! Jetpack::is_module_active( 'sso' ) ) {
+		wpcom_vip_load_plugin( 'two-factor' );
+	} else if ( class_exists( 'Jetpack' ) && Jetpack::is_active() && Jetpack::is_module_active( 'sso' ) ) {
```
mjangda: Feels like `class_exists( 'Jetpack' ) && Jetpack::is_active() && Jetpack::is_module_active( 'sso' )` can be moved to a helper function and re-used within this function.

twostep.php
```diff
@@ -0,0 +1,55 @@
...
+add_filter( 'two_factor_providers', function( $p ) {
+        unset( $p[ 'Two_Factor_FIDO_U2F' ] );
+        return $p;
+});
+
+function wpcom_vip_force_twostep() {
+	return ! wpcom_vip_plugin_is_loaded( 'shared-plugins/jetpack-force-2fa' ) && class_exists( 'Two_Factor_Core' ) && ! Two_Factor_Core::is_user_using_two_factor();
+}
+
+function wpcom_enable_two_factor_plugin() {
```
mjangda: What happens if a theme decides to load the `two-factor` plugin?

twostep.php
```diff
@@ -0,0 +1,55 @@
...
+	if ( ! wpcom_vip_plugin_is_loaded( 'shared-plugins/jetpack-force-2fa' ) && ! class_exists( 'Jetpack' ) && ! Jetpack::is_active() && ! Jetpack::is_module_active( 'sso' ) ) {
+		wpcom_vip_load_plugin( 'two-factor' );
+	} else if ( class_exists( 'Jetpack' ) && Jetpack::is_active() && Jetpack::is_module_active( 'sso' ) ) {
+		wpcom_vip_load_plugin( 'jetpack-force-2fa' );
+		remove_action( 'wp_login', array( 'Two_Factor_Core', 'wp_login' ) );
```
mjangda: Should have an inline comment explaining why the action is being removed.

twostep.php
```diff
@@ -0,0 +1,55 @@
...
+}
+add_filter( 'map_meta_cap', 'wpcom_vip_twostep_filter_caps' );
+
+function wpcom_vip_twostep_admin_notice() {
+	if ( ! wpcom_vip_force_twostep() ) {
+		return;
+	}
+	?>
+	<div class="error">
+		<p><a href="<?php echo admin_url( 'profile.php' ); ?>">Two Step Authentication</a> is required to publish to this site.</p>
```
mjangda: Add `esc_url` for `admin_url`.

wpcom-vip-two-factor/sms-provider.php
```diff
@@ -0,0 +1,290 @@
...
+	/**
+	 * Validate the user token.
+	 *
+	 * @param int    $user_id User ID.
+	 * @param string $token User token.
+	 * @return boolean
+	 */
+	public function validate_token( $user_id, $token ) {
+		$hashed_token = get_user_meta( $user_id, self::TOKEN_META_KEY, true );
+		if ( wp_hash( $token ) !== $hashed_token ) {
```
mjangda: Should we use `hash_equals` here?

wpcom-vip-two-factor/sms-provider.php
```diff
@@ -0,0 +1,290 @@
...
+	 *
+	 * @param WP_User $user WP_User object of the logged-in user.
+	 */
+	public function authentication_page( $user ) {
+		if ( ! $user ) {
+			return;
+		}
+
+		$this->generate_and_send_token( $user );
+		require_once( ABSPATH .  '/wp-admin/includes/template.php' );
```
mjangda: Need an inline comment explaining why `template.php` is needed.

wpcom-vip-two-factor/sms-provider.php
```diff
@@ -0,0 +1,290 @@
...
+
+	/**
+	 * Send an SMS message to the specified numbers using the Twilio API
+	 *
+	 * @param int|array Nnumber(s) to send an SMS to. Accepts formatted and unformatted US numbers, e.g. +14155551212, (415) 555-1212 or 415-555-1212.
+	 * @param string Message to send. Longer than 140 chars will be split.
+	 *
+	 * @link http://www.twilio.com/docs/api/rest/sending-sms
+	 */
+	static function send_sms( $to, $message, $country_code = '+1' ) {
```
mjangda: `send_sms`, `send_single_sms_via_rest `, `get_rest_url`, and `split_words` should probably be in their own class.

twostep.php
```diff
@@ -0,0 +1,55 @@
+<?php
+/**
+ * Plugin Name: VIP Force Two Step
```
mjangda: Naming consistency: we use the terms `two step`, 2fa`, and `two factor` all interchangibly. We should probably pick one and stick to it.

joshbetz: Agree. On WordPress.com we call it Two-Step Authentication, but the plugin we're using is called Two-Factor, so I'm not sure which direction to go.

mjangda: Since the plugin UI uses "Two Factor" as well, we should probably just go with that.

wpcom-vip-two-factor/sms-provider.php
```diff
@@ -0,0 +1,167 @@
...
+		if ( ! $user ) {
+			return;
+		}
+
+		if ( ! isset( $_GET['action'] ) || 'validate_2fa' !== $_GET['action'] ) {
+			$this->generate_and_send_token( $user );
+		}
+
+		// Including template.php for submit_button()
+		require_once( ABSPATH .  '/wp-admin/includes/template.php' );
```
joshbetz: Sure, though it's also used in all the "core" providers:

https://github.com/georgestephanis/two-factor/blob/master/providers/class.two-factor-totp.php#L323
https://github.com/georgestephanis/two-factor/blob/master/providers/class.two-factor-fido-u2f.php#L118
https://github.com/georgestephanis/two-factor/blob/master/providers/class.two-factor-email.php#L124

mjangda: Ah, never mind then :)

joshbetz:
I think this is ready to deploy. The `wpcom_vip_force_twofactor` filter should mean we can deploy this without breaking any workflows. Users will be able to enable two factor, but we won't remove caps immediately. If any clients want to opt-in to forcing two factor, they can filter `wpcom_vip_force_twofactor` to `__return_true`. After some time, we'll remove the filter and force for all users.

PR #459, opened 2017-02-07
Disable term counting for bulk edit

mjangda:
Each post updated triggers a term count update which can be very slow, especially when updating lots of posts.

### DIFF
--- a/performance/bulk-edit.php
+++ b/performance/bulk-edit.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Automattic\VIP\Performance;
+
+// Bulk edits of lots of posts can trigger slow term count queries for each post updated
+add_action( 'load-edit.php', function() {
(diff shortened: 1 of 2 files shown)


performance/bulk-edit.php
```diff
@@ -0,0 +1,13 @@
...
+
+// Bulk edits of lots of posts can trigger slow term count queries for each post updated
+add_action( 'load-edit.php', function() {
+	if ( isset( $_REQUEST['bulk_edit'] ) ) {
+		wp_defer_term_counting( true );
+		add_action( 'shutdown', function() {
+			wp_defer_term_counting( false );
+		} );
+	}
+} );
```
mjangda: The nested anonymous functions aren't great, but okay as a quick fix.

mjangda:
r87850-deploy

PR #461, opened 2017-02-08
Split fix_remote_addr_from_ip_trail

mjangda:
Split into a more focused function that takes user ip + remote proxy ip which can be used on its own if needed.

- [x] Split up function.
- [x] Split up tests so new function has coverage.


mjangda:
r87901-deploy

PR #470, opened 2017-02-16
Don't autoload lastpostmodified options

mjangda:
We only need them for feeds and don't need them loaded for every request.


mjangda:
r88479-deploy

Issue #469, opened 2017-02-16
Add async_publish_post event


mjangda:
Similar to what we have on WP.com (see `queue_publish_post`).

Whenever a post is updated we could queue up a cron event that can be used to perform various non-essential tasks.

Some core examples include `update_posts_count` and `wpmu_update_blogs_date`. Some Go-specific ones could be cache purges and `lastpostmod` bumps. VIPs could hook into this as well for their own heavy events/actions.

ethitter:
For sites running Cron Control, we can treat this as a "internal event," which will give it priority and ensure reliable execution. Sites running WP-Cron Control should be fine with this as well. We do have a few sites that rely on traditional WordPress cron execution, which we'll want to exempt lest these async jobs never run.

mjangda:
Some more misc thoughts:

- WP.com only queues jobs for `publish_post`. Do we want to handle more status transitions (e.g. publish => trash)? 
- Should we queue one event where others hook in and we execute them all at the same time? (could result in cron events that take a long time to complete) Or should we offer a handler than allows queueing multiple events per task? (could result in lots of cron events per post update)

ethitter:
> WP.com only queues jobs for `publish_post`. Do we want to handle more status transitions (e.g. publish => trash)?

Seems worth queueing the job whenever either the new or old status is `publish` (skipping when they're both publish), since post counts and other things would change in that situation. We should be sure to cover CPTs, at least public CPTs, as `post` isn't guaranteed to be the primary CPT for every site.

> Should we queue one event where others hook in and we execute them all at the same time? Or should we offer a handler than allows queueing multiple events per task?

I'd rather queue discrete events. Too many long-running tasks in a single event are likely to run into the time limits imposed by our various cron runners, and separate entries will simplify debugging.

ethitter:
I changed my mind a bit while revising #471. 😄 

I opted for one job, because it's more natural for users to hook to an action. It also means that a lot of code can be moved to the hook with the addition of `async_` to the action. 

Right now, it applies to all post types. Should it?

Should transitioning from published to unpublished trigger a different action? Right now, it still fires `async_publish_post`.

Should we provide a filter to disable it for post types, statuses, etc?

Might something be missed, since we don't requeue if there's already a scheduled event?

rebeccahum:
Closing in favour of https://github.com/Automattic/vip-go-mu-plugins/pull/471.

PR #471, opened 2017-02-17
Add async transition_post_status for publish events

mjangda:
See #469


ethitter:
@mjangda I'm curious how an auto-draft scheduled an event, as it should've failed the status checks that are already in place. I added some logging and left a few browser tabs open to the editor, and thus far, haven't been able to capture an auto-draft. I'm surely missing something.

ethitter:
Another thought as I've begun the hunt for things to use with this: should we automatically offload things to the async hook during imports, CLI and XML-RPC requests, etc? The environment changes, which may or may not be a problem. Losing the `WP_IMPORTING` constant, for example, would probably cause trouble.

performance/async-publish-actions.php
```diff
@@ -0,0 +1,114 @@
...
+	if ( $new_status !== $old_status && in_array( 'publish', [ $new_status, $old_status ], true ) ) {
+		wp_schedule_single_event( time(), ASYNC_TRANSITION_EVENT, $args );
+	}
+}
+add_action( 'transition_post_status', __NAMESPACE__ . '\_queue_async_hooks', 10, 3 );
+
+/**
+ * Offload ping- and enclosure-related events
+ */
+remove_action( 'publish_post', '\_publish_post_hook', 5 );
```
mjangda: Looking at this callback, I'm not sure if there's actually much benefit in offloading since all it's doing is adding some post meta and maybe scheduling an event, neither of which are very expensive tasks.

performance/async-publish-actions.php
```diff
@@ -0,0 +1,114 @@
...
+ * Queue async event when status was or is `publish`
+ *
+ * @access private
+ *
+ * @param string $new_status New post status.
+ * @param string $old_status Old post status.
+ * @param object $post \WP_Post object.
+ */
+function _queue_async_hooks( $new_status, $old_status, $post ) {
+	if ( 'auto-draft' === $post->post_status ) {
```
mjangda: Should we add a WP_IMPORTING check as well?

tests/test-performance-async-publish-actions.php
```diff
@@ -0,0 +1,65 @@
...
+
+		parent::tearDown();
+	}
+
+	/**
+	 * Test that offload is scheduled when post is published
+	 */
+	function test_schedule_offload() {
+		$post = [
+			'post_title'   => 'Tommy Tutone - 867-5309/Jenny',
```
mjangda: :D

tests/test-performance-async-publish-actions.php
```diff
@@ -0,0 +1,65 @@
...
+		$args = [
+			'post_id'    => (int) $pid,
+			'new_status' => 'publish',
+			'old_status' => 'draft',
+		];
+
+		$next_transition    = wp_next_scheduled( ASYNC_TRANSITION_EVENT, $args );
+
+		// Event scheduled after publish.
+		$this->assertTrue( is_int( $next_transition ) );
```
mjangda: Ideally, this would be split off into multiple tests. A good rule of thumb is to avoid multiple state changes + multiple unrelated assert calls in a single test. Each code path + varying input should have its own tests.

Another good rule of thumb is that if you do end up with multiple assert calls in a single test, you should include the message parameter to add clues as which assertion fails (instead of just getting an error that says `expected true was false`.

Anyway, we could break these off into the following tests:

- Test auto-draft post doesn't schedule.
- Test already scheduled post doesn't schedule.
- Test when both new and old status are the same.
- Test when new status is publish.
- Test when old status is publish.
- Test that publishing a post and then another post schedules different events.

performance/async-publish-actions.php
ethitter: That's already skipped by `should_offload()`.

performance/async-publish-actions.php
```diff
@@ -0,0 +1,114 @@
...
+	( defined( 'XMLRPC_REQUEST' ) && \XMLRPC_REQUEST ) ||
+	( defined( 'WP_IMPORTING' ) && \WP_IMPORTING )
+) {
+	/**
+	 * Bypass offloading to the async hook, unless specifically requested
+	 *
+	 * @param bool $bypass Whether or not to bypass async offloading.
+	 */
+	if ( ! apply_filters( 'wpcom_async_transition_post_status_force_queue_event', false ) ) {
+		return;
```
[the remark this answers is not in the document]
ethitter: Agreed, this was a bad idea.

tests/test-performance-async-publish-actions.php
ethitter: 795e7b and ee0c038

performance/async-publish-actions.php
```diff
@@ -0,0 +1,114 @@
+<?php
+
+namespace Automattic\VIP\Performance;
```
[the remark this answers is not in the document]
ethitter: Yeah, a top-level mu-plugin makes more sense than its current location.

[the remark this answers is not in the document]
ethitter: 72aa3cac

async-publish-actions.php
```diff
@@ -0,0 +1,148 @@
...
+		$context = 'wp-cli';
+		$offload = false;
+	}
+
+	if ( defined( 'XMLRPC_REQUEST' ) && \XMLRPC_REQUEST ) {
+		$context = 'xml-rpc';
+		$offload = false;
+	}
+
+	if ( defined( 'WP_IMPORTING' ) && \WP_IMPORTING ) {
```
mjangda: Minor: should these be `elseif`?

tests/test-async-publish-actions.php
```diff
@@ -0,0 +1,197 @@
...
+ */
+
+namespace Automattic\VIP\Async_Publish_Actions\Tests;
+use Automattic\VIP\Async_Publish_Actions;
+use WP_UnitTestCase;
+
+/**
+ * @group async-publish-actions
+ */
+class async_publish_actions_Test extends WP_UnitTestCase {
```
mjangda: Minor: capitals for class name.

mjangda:
One additional thought (this doesn't need to block deploying this): we may want to consider allowing publish-to-publish transition (i.e. a simple update) as well.

Example use case: sending a ping to an external API any time a post is updated; that would be better done async.

async-publish-actions.php
ethitter: Oh yes! :man_facepalming:

ethitter:
That's a good point about updates. That's an simple change, with most of the effort in updating the tests.

Your question got me thinking about how I've handled a rapid set of updates that happen before the cron events are all run. Right now, it queues an event for reach distinct transition, and the serial nature of event execution (Cron Control locks by action, by default) ensures that they're processed in order. Canceling pending transitions seems unwise, but if the post object has changed since the event was queued, I'm also not sure what value it has. Perhaps the cron callback should check that the `$new_status` it's passed matches the post object's current status (we call `get_post()`, so it'll be fresh)?

mjangda:
> Perhaps the cron callback should check that the $new_status it's passed matches the post object's current status (we call get_post(), so it'll be fresh)?

That seems reasonable.

mjangda:
We'll want to update docs + Lobby post about this update including examples of things this is useful for.

ethitter:
r99181-deploy

PR #472, opened 2017-02-19
Performance and other tweaks for Akismet

mjangda:
- Limit cache increments for sites with lots of spam
- Limit number of spam comments in delete batch
- Identify Akismet requests as coming from vipgo


mjangda:
r88828-deploy

PR #475, opened 2017-02-21
Impose limits on bulk-editing

ethitter:
Bulk editing can have a considerable, negative impact on database servers, so by limiting the number of posts that can be processed at one time, the feature can be used without endangering stability.

WordPress is a little funny with this, and for hierarchical post types, uses a per-page value of `-1` to ensure all children can be nested under their parents. In that case, we block all bulk actions as there's no reliable way of knowing how many entries might be affected.

### DIFF
--- a/performance/bulk-edit.php
+++ b/performance/bulk-edit.php
@@ -22,6 +22,7 @@ class Bulk_Edit {
 			} );
 		}
 	}
+
 	/**
 	 * Suppress bulk actions when too many posts would be affected
 	 *
--- a/performance/bulk-edit.php
+++ b/performance/bulk-edit.php
@@ -65,9 +65,31 @@ function bulk_edit_admin_notice() {
 		return;
 	}
 
+	// HTML class doubles as key used to track dismissed notices
+	$id = 'notice-vip-bulk-edit-limited';
+
+	$dismissed_pointers = array_filter( explode( ',', (string) get_user_meta( get_current_user_id(), 'dismissed_wp_pointers', true ) ) );
+	if ( in_array( $id, $dismissed_pointers, true ) ) {
+		return;
+	}
+
 	?>
-	<div class="notice notice-error is-dismissible">
+	<div id="<?php echo esc_attr( $id ); ?>" class="notice notice-error is-dismissible">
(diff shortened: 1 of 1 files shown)


joshbetz:
I wonder if we can somehow make it clear what's going on so the box doesn't magically appear and disappear depending on posts_per_page.

performance/bulk-edit.php
```diff
@@ -2,12 +2,58 @@
...
-// Bulk edits of lots of posts can trigger slow term count queries for each post updated
-add_action( 'load-edit.php', function() {
-	if ( isset( $_REQUEST['bulk_edit'] ) ) {
-		wp_defer_term_counting( true );
-		add_action( 'shutdown', function() {
-			wp_defer_term_counting( false );
-		} );
-	}
-} );
+class Bulk_Edit {
```
mjangda: Do we really need to add a new class here?

ethitter:
@joshbetz We could include an admin notice. If we do, it should probably be dismissable, otherwise it could be annoying.

performance/bulk-edit.php
ethitter: Need? No. I just have some weird aversion to a pile of functions in a file.

joshbetz: I like the idea of using namespaces for things like this. I think we're all used to using classes in place of namespaces from before PHP 5.3. This pattern is nice though: https://github.com/Automattic/vip-go-mu-plugins/blob/518d47280954dd0b7f394a1a45fce5b9fe2e7dff/lib/proxy/ip-forward.php#L3

ethitter: I'll drop the class. I can live without it. :)

ethitter:
I added the admin notice, though it might be too annoying. It's dismissible, but I didn't realize that that applies only to the current page render. Dismissing it isn't persistent, but we could store that in a user option if we think it's necessary.

ethitter:
The admin notice is definitely too annoying if it's not permanently dismissible. Making dismissing stick, or putting a note in screen help, are two thoughts I've had. 

If we weren't removing something, we could use a pointer, which is natively dismissible. 😄

ethitter:
r88947-deploy

PR #476, opened 2017-02-21
Adding missing VIP helper function from WPCOM

emrikol:
`wpcom_vip_remove_playlist_styles()` is missing from VIP Go, and as far as I can tell not needed, so it should be deprecated.


joshbetz:
r88965-deploy

PR #477, opened 2017-02-21
Porting "VIP Performance Tweaks" from WPCOM

emrikol:
Bringing over performance tweak functionality from WordPress.com's `vip-helper-wpcom.php`


joshbetz:
r88965-deploy

PR #478, opened 2017-02-22
Adding bulk operation functions to VIP CLI Helper

emrikol:
These are pulled from WordPress.com, but are missing the PushPress and Elasticsearch components that don't exist on VIP Go.


joshbetz:
r88965-deploy

PR #480, opened 2017-02-23
Allow only authenticated requests to the VIP endpoints

ethitter:
To ensure we list all non-deleted sites, requests should be authenticated. With this PR, all requests will require a site-specific header.


ethitter:
This depends on changes to various WordPress.com uses of the endpoint. Those changes must be deployed first.

vip-rest-api.php
```diff
@@ -124,6 +121,32 @@ public function list_sites() {
...
 		return new WP_REST_Response( $sites );
 	}
+
+	/**
+	 * Check if necessary authentication header is present
+	 *
+	 * @return bool
+	 */
+	public function request_allowed( $request ) {
+		$header = $request->get_header( 'X-WPCom-Request-Secret' );
```
mjangda: I think we should use the standard `Authorization` header with a custom scheme instead of a custom header altogether (see http://stackoverflow.com/a/11420667). Maybe something like `Authorization: VIP-MACHINE-TOKEN a1b2c3d4...`

ethitter: Oh, definitely. TIL about custom schemes.

mjangda:
I think we can add the verification code directly in `mu-plugins`. It would also be good to break out the `request_allowed` method into its own class/module.

ethitter:
This is ready for another look. I moved the endpoints to a subdirectory, freeing up the top-level `vip-rest-api.php` file for the "helpers" that are needed to access endpoints.

vip-rest-api.php
```diff
@@ -1,129 +1,86 @@
...
+ * Check if necessary authentication header allows access to an endpoint
+ *
+ * Not always called as a REST API permission callback, hence going directly to the global
+ *
+ * @param  string $namespace RESET API route's namespace
+ * @return bool
+ */
+function wpcom_vip_go_rest_api_request_allowed( $namespace ) {
+	// Do we have a header to check?
+	if ( isset( $_SERVER['HTTP_AUTHORIZATION'] ) && ! empty( $_SERVER['HTTP_AUTHORIZATION'] ) ) {
```
mjangda: Can we change this to `if ( empty( ... ) ) { return false }` for simplicity?

ethitter:
Ready for another look @mjangda

rest-api/vip-endpoints.php
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	/**
+	 * PLUGIN FUNCTIONALITY
+	 */
+
+	/**
+	 * Some `/vip/` endpoints need to be accessible to requests from WordPress.com
+	 */
+	public function disable_auth( $result ) {
```
mjangda: The name `disable_auth` doesn't quite apply anymore since we do have auth now.

mjangda: The name `disable_auth` doesn't quite apply anymore since we do have auth now.

mjangda:
@ethitter Also, some general code review feedback that I didn't feel should block this but some suggestions for future.

For tests, it's usually better to be clear and explicit even if it means lots of repetition. I listened to a presentation (can't remember the source) that covered two really good points for me:
1) If you need to jump around to other parts of the source code to figure out what your test setup is doing, your tests are probably too complicated.
2) Tests should follow a pattern something like this:

```
function test__the_thing() {
    expected = 123

    setup_1()
    setup_2()

    value = execute_method()

   asset( value, expected )
}
```

This makes them super readable (try the "squint test" with a test structured like this) and it's very clear what's going on.

Obviously it's not possible to be as clear-cut as this all the time, but definitely something to strive for.

---

Unrelated to tests, I think namespaces may have been a good choice here to slim down the length of the auth function names.

Finally, if we move the auth functions to a separate file, that may simplify porting between here and WP.com as needed (although, hopefully, they don't need too many changes :))

ethitter:
r90341-deploy

PR #482, opened 2017-02-23
Prevent Jetpack version ping-pong on sandboxes

mjangda:
When a sandbox has an old version of stacks, this can lead to some serious issues because Jetpack tries to auto-correct the version on each load. The mismatch between the sandbox environment and live environment leads to a constant fight to override the option.

This just kills a request when an attempt to do a Jetpack version update is triggered from a sandbox.

### DIFF
--- a/vip-jetpack/vip-jetpack.php
+++ b/vip-jetpack/vip-jetpack.php
@@ -26,6 +26,13 @@ add_filter( 'jetpack_get_available_modules', function( $modules ) {
 	return $modules;
 }, 999 );
 
+// Prevent Jetpack version ping-pong when a sandbox has an old version of stacks
+if ( true === WPCOM_SANDBOXED ) {
+	add_action( 'updating_jetpack_version', function() {
+		wp_die( '😱😱😱 Oh no! Looks like your sandbox is trying to change the version of Jetpack. This is probably not a good idea. As a precaution, we\'re killing this request to prevent this from happening (this === 💥💥💥). Please run `vip stacks update` on your sandbox before doing anything else.', 400 );
+	}, 0 ); // No need to wait till priority 10 since we're going to die anyway
+}
+
 /**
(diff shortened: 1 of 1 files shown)


simonwheatley:
I suggest we move this to the VIP Jetpack plugin, rather than `misc.php`.

misc.php
```diff
@@ -178,3 +178,10 @@ function wpcom_vip_wp_link_query_args( $query ) {
 }
 
 add_filter( 'wp_link_query_args', 'wpcom_vip_wp_link_query_args', 10, 1 );
+
+// Prevent Jetpack version ping-pong when a sandbox has an old version of stacks
+if ( true === WPCOM_SANDBOXED ) {
+	add_action( 'updating_jetpack_version', function() {
```
simonwheatley: How about running with a priority of 0, so we're nice and early to the party?

ethitter: `PHP_MAX_INT * -1` 😉 

Okay, that might be _too_ early, but earlier is definitely better here.

mjangda: So, `PHP_INT_MIN`? :)

mjangda:
I've updated the priority for the hook.

@simonwheatley I'd like to bring the `vip-jetpack` plugin directly into `mu-plugins` (i.e. move away from a submodule to simplify development/deployment) and can move this snippet over after that's done. Any concerns with that?

simonwheatley:
> I'd like to bring the vip-jetpack plugin directly into mu-plugins (i.e. move away from a submodule to simplify development/deployment) and can move this snippet over after that's done. Any concerns with that?

My initial thinking was that there is value in separating plugins which others might want to cherry pick and use, but for this case perhaps it's unnecessary. Happy for this to be rolled into the main MU plugins repo.

mjangda:
Moved code to `vip-jetpack`

mjangda:
r89993-deploy

PR #485, opened 2017-02-24
Fix PHP notice from Akismet overrides

ethitter:
Use filter instead of constant, since Akismet doesn't check if constant is already defined before defining its default.

`PHP Notice:  Constant AKISMET_DELETE_LIMIT already defined in /.../wp-content/mu-plugins/akismet.php on line 18`

See 46faf19 and #472


mjangda:
Lame. We should get Akismet to fix this :)

ethitter:
r88986-deploy

PR #486, opened 2017-02-24
Have users contact us when we block Bulk Edit

mjangda:
So they have someone to talk to instead of feeling stuck.

h/t @andrealeebishop


ethitter:
Related to #475

r89040-deploy

PR #488, opened 2017-02-25
Track publish events

ethitter:
Track post-publish volume using a late, non-blocking request.

### DIFF
--- a/stats.php
+++ b/stats.php
@@ -3,8 +3,8 @@
 namespace Automattic\VIP\Stats;
 
 // Limit tracking to production
-if ( false !== WPCOM_IS_VIP_ENV ) {
-	add_action( 'transition_post_status', __NAMESPACE__ . '\track_publish_post', 9999, 10, 2 );
+if ( true === WPCOM_IS_VIP_ENV ) {
+	add_action( 'transition_post_status', __NAMESPACE__ . '\track_publish_post', 9999, 2 );
 }
 
 /**
--- a/stats.php
+++ b/stats.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace Automattic\VIP\Stats;
+
+// Limit tracking to production
+if ( false !== VIP_GO_ENV ) {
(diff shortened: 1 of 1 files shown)


stats.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace Automattic\VIP\Stats;
+
+// Limit tracking to production
+if ( false !== VIP_GO_ENV ) {
```
joshbetz: I think we're using `VIP_GO_ENV` to distinguish staging from production and using `WPCOM_IS_VIP_ENV` for local vs VIP Go. I guess we should change the comment here: https://github.com/Automattic/vip-go-mu-plugins/blob/master/000-vip-init.php#L21

stats.php
```diff
@@ -0,0 +1,24 @@
...
+	if ( 'publish' !== $new_status || 'publish' === $old_status ) {
+		return;
+	}
+
+	$pixel = add_query_arg( array(
+		'v'                     => 'wpcom-no-pv',
+		'x_vip-go-publish-post' => FILES_CLIENT_SITE_ID,
+	), 'http://pixel.wp.com/b.gif' );
+
+	wp_remote_get( $pixel, array( 'blocking' => false, ) );
```
mjangda: Do we need to set a low timeout as well (1s?) just in case the transport that's used doesn't support non-blocking requests?

stats.php
mjangda: Minor: `true === 'WPCOM_IS_VIP'` is slightly easier to parse/read.

stats.php
ethitter: Can't hurt. Tracks does: https://github.com/Automattic/jetpack/blob/d40db24971bec4fe0355d909fbe0a6ac6ce972d9/_inc/lib/tracks/class.tracks-client.php#L81. It also makes a blocking request, but I'm not sure we need to, unless we move this to the async hook coming with #471.

PR #490, opened 2017-02-27
Adding caching support for is_multi_author

emrikol:
`is_multi_author()` normally caches itself if the site has multiple authors, but if not, it tries to cache a `false` value which always makes the lookup happen.  This can cause a very slow query on sites where many tens of thousands of posts are attributed to the same author.

Can be used by adding this filter to the theme: `add_filter( 'pre_transient_is_multi_author', 'wpcom_vip_is_multi_author' );`

This helper function will cache the results for one hour as a string so even failures will be cached, since `is_multi_author()` casts its output as a boolean.


mjangda:
As discussed, the better option here would be to remove the use of the function altogether. Beyond that, we should look at storing it the multi_author state in a more permanent store instead.

emrikol:
> As discussed, the better option here would be to remove the use of the function altogether.

👍 I'm going to close this out for now then.  Thanks!

PR #495, opened 2017-03-01
Don't track publish events during imports

ethitter:
The stat shouldn't be bumped when `WP_IMPORTING` is defined and `true`.

### DIFF
--- a/stats.php
+++ b/stats.php
@@ -11,7 +11,7 @@ if ( true === WPCOM_IS_VIP_ENV ) {
  * Count publish events regardless of post type
  */
 function track_publish_post( $new_status, $old_status ) {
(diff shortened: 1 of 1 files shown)


stats.php
```diff
@@ -11,6 +11,10 @@
  * Count publish events regardless of post type
  */
 function track_publish_post( $new_status, $old_status ) {
+	if ( defined( 'WP_IMPORTING' ) && WP_IMPORTING ) {
```
mjangda: Minor: I usually prefer to do strict checks even with booleans (i.e. `true === WP_IMPORTING` ).

ethitter:
r89269-deploy

PR #497, opened 2017-03-03
Disable do_pings

### LINKED ISSUE #462: Disable do_pings by default
Something like:

```
add_filter( 'schedule_event', function( $event ) {
    if ( 'do_pings' === $event->hook ) {
        return false;
    }
    return $event;
} );
```

With the option to enable, if preferred.

joshbetz:
Fixes #462

### DIFF
--- a/performance/do-pings.php
+++ b/performance/do-pings.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Automattic\VIP\Performance;
+
+// Disable pings by default
+function disable_pings( $event ) {
+	if ( 'do_pings' === $event->hook ) {
+		return false;
+	}
+
+	return $event;
+}
+add_action( 'schedule_event', __NAMESPACE__ . '\disable_pings' );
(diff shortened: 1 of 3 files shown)


ethitter:
Should we remove the `VIP_CUSTOM_PINGS` stuff along with this change? https://github.com/Automattic/vip-go-mu-plugins/blob/0b4ad12862a2aef65982eadc94fd96521c7acf01/misc.php#L85

performance/do-pings.php
```diff
@@ -0,0 +1,13 @@
...
+
+// Disable pings by default
+function disable_pings( $event ) {
+	if ( 'do_pings' === $event->hook ) {
+		return false;
+	}
+
+	return $event;
+}
+add_action( 'schedule_event', 'disable_pings' );
```
ethitter: Core needs the namespace specified as well.

joshbetz:
Partially deployed in r89565-deploy. Will remove `VIP_CUSTOM_PINGS` code later today.

mjangda:
@joshbetz looks like `VIP_CUSTOM_PINGS` still needs to be removed from the `deploy` repo.

joshbetz:
Thanks for the reminder. Deployed in r90185-deploy.

PR #498, opened 2017-03-03
Bump http-concat to cea04c4

mjangda:
- Updated style_loader_tag filter with prefix to avoid conflicts.
- Support for subdirectory (single and multisite) installs.


mjangda:
r89389-deploy

PR #499, opened 2017-03-08
Kill AMP attachment lookups

mjangda:
These are pretty slow on large sites and don't work well on Go anyway, since most images have querystrings attached. This feature has already been removed in the AMP plugin upstream and this is a workaround until that is rolled out everywhere.

### DIFF
--- a/performance.php
+++ b/performance.php
@@ -4,3 +4,4 @@ require_once( __DIR__ . '/performance/lastpostmodified.php' );
 require_once( __DIR__ . '/performance/bulk-edit.php' );
 require_once( __DIR__ . '/performance/vip-tweaks.php' );
 require_once( __DIR__ . '/performance/do-pings.php' );
+require_once( __DIR__ . '/performance/plugins.php' );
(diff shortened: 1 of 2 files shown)


performance.php
```diff
@@ -4,3 +4,4 @@
 require_once( __DIR__ . '/performance/bulk-edit.php' );
 require_once( __DIR__ . '/performance/vip-tweaks.php' );
 require_once( __DIR__ . '/performance/do-pings.php' );
+require_once( __DIR__ . '/performance/plugins.php' );
```
joshbetz: Should we use `glob( __DIR__ . '/performance/*.php' )` here to include all files in the performance directory?

nickdaugherty: I would suggest avoiding `glob` in favor of hardcoded file inclusion. `glob` is convenient, but makes the app's behavior more 'magical' and less predictable.

It's also something we look for in client code: https://vip.wordpress.com/documentation/vip/code-review-what-we-look-for/#implicit-file-includes

mjangda:
r90084-deploy

PR #503, opened 2017-03-10
Add a simple svn wrapper to do pre-deploy checks.

joshbetz:
Not sure if this is where we'd want to put a script like this, but it seems as good a place as any.

### DIFF
--- a/bin/svn
+++ b/bin/svn
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+predeploy() {
+	# PHP Lint
+	find . -name \*.php -print0 | xargs -0 -n 1 -P 4 php -nl || exit 1
+
+	# Check for unverioned files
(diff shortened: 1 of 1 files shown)


bin/svn
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+predeploy() {
+	# PHP Lint
+	find . -name \*.php -print0 | xargs -0 -n 1 -P 4 php -nl || exit 1
```
mjangda: This is pretty noisy; I think we'll want to point output to `/dev/null`

joshbetz:
Alternatively: deploy directly from Github.

https://veetoop2.wordpress.com/2017/02/07/version-control/

mjangda:
I think we can merge this for now and get it mapped on our sandboxes; the larger discussion around deploying from Github probably shouldn't block this.

PR #504, opened 2017-03-10
Travis + PHP7 + phpunit compat

mjangda:
Borrowed from https://core.trac.wordpress.org/changeset/40255

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,12 +25,9 @@ before_install:
 install: true
 
 before_script:
-- |
-  if [[ ${TRAVIS_PHP_VERSION:0:2} == "7." ]]; then
-    composer global require "phpunit/phpunit=5.7.*"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -24,6 +24,14 @@ before_install:
 
 install: true
 
+before_script:
+- |
+  if [[ ${TRAVIS_PHP_VERSION:0:2} == "7." ]]; then
+    composer global require "phpunit/phpunit=5.7.*"
+  elif [[ ${TRAVIS_PHP_VERSION:0:3} != "5.2" ]]; then
```
joshbetz: Are we purposely not including phpunit for PHP 5.2?

mjangda: I just copied the core patch (https://core.trac.wordpress.org/changeset/40257). I guess since we are only testing 7.x right now, we don't need the conditional; we can just load phpunit 5.7.

mjangda:
@joshbetz can you take another look?

PR #506, opened 2017-03-12
Update Query Monitor to latest release

ethitter:
See https://github.com/johnbillion/query-monitor/releases


ethitter:
r90083-deploy

PR #507, opened 2017-03-13
Un-submodule vip-jetpack

mjangda:
See https://github.com/Automattic/vip-go-mu-plugins/pull/482#issuecomment-284635069

The code is fairly vip-specific and not particularly useful as a generic plugin. It can be used a reference instead.

Related #458 (which changes around where the force-2fa submodule lives and will need to be updated).


mjangda:
The `jetpack-force-2fa` change (https://github.com/Automattic/jetpack-force-2fa/pull/20) landed in r89992-deploy

PR #513, opened 2017-03-15
Make quality=100 an explicit choice

### LINKED ISSUE #511: a8c-files: Make it harder to set quality=100
This generates large, lossless images, which, if used incorrectly can cause issues. We should require the use of a constant when needing `quality=100` so changes can be detected and validated.

mjangda:
Require the use of a constant to set the quality value to 100. Otherwise bump other requests to 99.

I'm not sure if this is a good idea :)

Fixes #511


mjangda:
> So you would set `?quality=allow-quality-100` instead of `?quality=100`

Sort of. You would need this:

```
add_filter( 'vip_go_image_resize_pre_args', function( $args ) {
    $args['quality'] = WPCOM_VIP_IMAGE_QUALITY_100;
    return $args;
} );
```

Which our `image_downsize` code then coverts the arg to `quality=100`. Need to think on this some more.

a8c-files.php
```diff
@@ -690,6 +693,19 @@ function image_resize( $ignore, $id, $size ) {
 			$resized = true;
 		}
 
+		if ( isset( $args['quality'] ) ) {
+			$quality = $args['quality'];
+			// `quality=100` is a special value that generates large, lossless images which if not used correctly can cause issues. 
+			// There are valid use cases, but those should be concious choice and we're requiring the use of this constant to make it clear that the decision was thought through.
+			// If you're not sure whether to use this constant, please ask.
+			if ( WPCOM_VIP_IMAGE_QUALITY_100 === $quality ) {
```
mjangda: The constant is defined in the same file, so not strictly necessary.

a8c-files.php
```diff
@@ -19,6 +19,9 @@
 
 define( 'ALLOW_UNFILTERED_UPLOADS', false );
 
+// See inline for usage of this constant
+define( 'WPCOM_VIP_IMAGE_QUALITY_100', 'allow-quality-100' );
```
mjangda: As discussed, there was some confusion in how this is supposed to be used. My comment below clarifies usage: https://github.com/Automattic/vip-go-mu-plugins/pull/513#issuecomment-286887640

mjangda:
Still need to:

- [ ] Rebase
- [ ] Make it work with the new Jetpack Photon filters
- [ ] Add PHPDoc for the filter

simonwheatley:
@nickdaugherty Can you please re-review?

tests/test-a8c-files-utils.php
```diff
@@ -1,6 +1,40 @@
...
+				100,
+				99,
+				true,
+			],
+
+			'below_100_is_fine' => [
+				99,
+				99,
+				false,
+			],
```
nickdaugherty: Can we add some tests for the string version of these values? So if quality were `'100'` or `'99'`. They should _just work_ due to PHP's type juggling, but good to have tests that cover it.

mjangda:
Going to close this out. This just adds unnecessary complexity/unpredictability. The better fix (which we've implemented) is greater capacity on the files service.

PR #515, opened 2017-03-16
Add dynamic rewrites for intermediate image sizes

mjangda:
Relies on existing Jetpack Photon code that converts URL references to intermediate image sizes (`file-{w}x{h}.jpg`) to a dyanmic one (`?resize={w},{h}`) served by our Go Files service, which has similar capabilities as Photon.

/cc @mdrovdahl

### DIFF
--- a/a8c-files.php
+++ b/a8c-files.php
@@ -22,12 +22,12 @@ define( 'ALLOW_UNFILTERED_UPLOADS', false );
 class A8C_Files {
 
 	function __construct() {
-		
+
 		// Upload size limit is 1GB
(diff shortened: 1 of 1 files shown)


ethitter:
We need to be careful that this won't interfere with clients who've done similar things in their themes, as I believe some are using cookie-less domains and replacing the Photon URL as we do here. Not a concern until we ungate this, but important to keep in mind.

mjangda:
> We need to be careful that this won't interfere with clients who've done similar things in their themes, as I believe some are using cookie-less domains and replacing the Photon URL as we do here.

Very good point.

PR #516, opened 2017-03-16
Fix query monitor

joshbetz:
With the latest update, the `activate()` function was moved to a new `QM_Activation` class.

See: #506


joshbetz:
r90133-deploy

PR #518, opened 2017-03-17
Add `wpcomvip` to restricted usernames

mjangda:
And actually disable logins for the account.

- [x] Account for emails as usernames
- [x] Check that this doesn't block Jetpack communication with the site when connected via the `wpcomvip` account.
- [x] Tests?

Jetpack Start (#433) needs to be working well before we can merge this.


joshbetz:
Good idea adding this to the existing login limiting functionality.

mjangda:
We're going to hold off on deploying this until #638 lands and is stable.

mjangda:
This definitely needs thorough testing to make sure it doesn't block communication between WP.com and site (i.e. WP.com needs to be able to authenticate as the `wpcomvip` user with the blog token).

joshbetz:
Looks like this does break Jetpack if it's connected with the machine user.

joshbetz:
Hmm. So in my testing, it seemed to break Jetpack, but there is some code that seems to fix this:

https://github.com/Automattic/jetpack/blob/a6a58b71d2210c25e99fc380f55f218fb715c903/class.jetpack.php#L894

mjangda:
In my initial testing there were issues with this PR and the connection flow. However, since we've moved to Jetpack Start (and because of the auth fix within Jetpack) this is no longer an issue.

mjangda:
r108800-deploy

PR #519, opened 2017-03-17
Adding a funciton for flushing wp_old_slug_redirect_cache on post_update

david-binda:
In case a post_name is changed and item is published, we should flush all previous wp_old_slug_redirect cache in order to prevent possible redirect loops.


mjangda:
@sboisvert Did you deploy this? If so, can you please add the Deployed label and commit number?

sboisvert:
By deployed do we mean to wpcom? If so,  yes and r286292. I'll wait for you to confirm before putting the deployed label.

mjangda:
I mean to the VIP Go deploy repo :)

sboisvert:
Ah then no, I can do that Monday (not sure how yet)

mjangda:
r90510-deploy

PR #520, opened 2017-03-17
Run phpcs on Travis

joshbetz:
Initial effort to start enforcing PHPCS. The thresholds (`--error-severity`, `--warning-severity`) are currently set such that all existing code passes. I think we'd eventually want to dial these down until we're flagging all errors, at least.

The phpcbf target will do some automated fixing and you can also configure your editor to run phpcbf automatically, though probably after we get the existing errors fixed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ phpcs:
 	test -d /tmp/phpcs || git clone -b master --depth 1 https://github.com/squizlabs/PHP_CodeSniffer.git /tmp/phpcs
 	test -d /tmp/wpcs || git clone -b master --depth 1 https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards.git /tmp/wpcs
 	/tmp/phpcs/scripts/phpcs --config-set installed_paths /tmp/wpcs
-	/tmp/phpcs/scripts/phpcs -p . --standard=WordPress --extensions=php --error-severity=6 --warning-severity=8 --ignore=shared-plugins/*,advanced-post-cache/*,cron-control/*,debug-bar-cron/*,cron-control/*,jetpack/*,vaultpress/*,debug-bar/*,akismet/*,http-concat/*,query-monitor/*,rewrite-rules-inspector/*,vip-dashboard/*,vip-support/*,wp-cron-control/*,wordpress-importer/*
+	/tmp/phpcs/scripts/phpcs -p . --standard=WordPress --extensions=php --runtime-set ignore_warnings_on_exit true --ignore=shared-plugins/*,advanced-post-cache/*,cron-control/*,debug-bar-cron/*,cron-control/*,jetpack/*,vaultpress/*,debug-bar/*,akismet/*,http-concat/*,query-monitor/*,rewrite-rules-inspector/*,vip-dashboard/*,vip-support/*,wp-cron-control/*,wordpress-importer/*
 
 phpcbf:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -10,3 +10,15 @@ phpunit:
...
+	/tmp/phpcs/scripts/phpcs --config-set installed_paths /tmp/wpcs
+	/tmp/phpcs/scripts/phpcs -p . --standard=WordPress --extensions=php --error-severity=6 --warning-severity=8 --ignore=shared-plugins/*,advanced-post-cache/*,cron-control/*,debug-bar-cron/*,cron-control/*,jetpack/*,vaultpress/*,debug-bar/*,akismet/*,http-concat/*,query-monitor/*,rewrite-rules-inspector/*,vip-dashboard/*,vip-support/*,wp-cron-control/*,wordpress-importer/*
+
+phpcbf:
+	test -d /tmp/phpcs || git clone -b master --depth 1 https://github.com/squizlabs/PHP_CodeSniffer.git /tmp/phpcs
+	test -d /tmp/wpcs || git clone -b master --depth 1 https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards.git /tmp/wpcs
+	/tmp/phpcs/scripts/phpcs --config-set installed_paths /tmp/wpcs
+	/tmp/phpcs/scripts/phpcbf -p . --standard=WordPress --extensions=php --ignore=shared-plugins/*,advanced-post-cache/*,cron-control/*,debug-bar-cron/*,cron-control/*,jetpack/*,vaultpress/*,debug-bar/*,akismet/*,http-concat/*,query-monitor/*,rewrite-rules-inspector/*,vip-dashboard/*,vip-support/*,wp-cron-control/*,wordpress-importer/*
```
nickdaugherty: What do you think about moving this config into a `phpcs.xml` file to make it more readable and accessible?

https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#using-a-default-configuration-file

joshbetz: Probably a good idea. I didn't realize we'd need to skip so many directories when I started.

simonwheatley:
@davidbinovec is maintaining a set of rules for VIP, which we might prefer here.

joshbetz:
Oh yeah, I switched to the WordPress-VIP standard, but forgot to push.

Makefile
joshbetz: It's unclear to me if we can use a phpcs.xml file with the WordPress-VIP standard.

> If you run PHP_CodeSniffer without specifying a coding standard, PHP_CodeSniffer will look in the current directory, and all parent directories, for a file called either phpcs.xml or phpcs.xml.dist.

mjangda:
👎 as-is.

Either we should go through files and fix errors before we merge this or make CI not fail (i.e. only show a warning) when PHPCS errors are found. It will be much harder to catch actual test failures with this enabled as-is.

joshbetz:
> Either we should go through files and fix errors before we merge this or make CI not fail

Agree. My plan was to go through and fix things, but I wanted to make sure we're not scanning things that aren't specific to VIP Go in the first batch (hence the big list of excludes and #537). Then I'd run phpcbf to automatically fix a bunch of this. And finally, manually fix whatever is left.

simonwheatley:
We have a tool to scan only the diffs, but it only works with SVN right now. There is an outstanding to do to have it work with git too:

https://github.com/Automattic/phpcs-diff

/cc @david-binda

mjangda:
We can probably pick this up again after #711 lands. Although, I find many of the Coding Standards checks to be too strict so would prefer to not have them be merge-blocking (at least to start).

PR #522, opened 2017-03-19
Force disable dev mode for Photon

mjangda:
When using Jetpack Photon with Jetpack dev mode enabled, we end up with full size images everywhere enabled because jetpack_photon_url is short-circuited.

By force disabling dev mode for Photon, we can work around this.

### DIFF
--- a/a8c-files.php
+++ b/a8c-files.php
@@ -68,6 +68,10 @@ class A8C_Files {
 			return home_url();
 		} );
 
+		// If Jetpack dev mode is enabled, jetpack_photon_url is short-circuited.
+		// This results in all images being full size (which is not ideal)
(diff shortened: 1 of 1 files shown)


joshbetz:
Are you seeing jetpack dev mode enabled in production? I would think if JETPACK_DEV_DEBUG is set, it would typically be a local environment where we can't resize images with the Go files services anyway?

mjangda:
Yeah, ideally, no production site should have `JETPACK_DEV_DEBUG` on. But, there was a site in a pre-launch state that had it enabled and ran into this bug so I figured this workaround made sense. The a8c-files code only runs on Go servers right now anyway so not too worried about local environments.

We probably should have some other check/process watching for `JETPACK_DEV_DEBUG` in prod though.

mjangda:
r90350-deploy

PR #523, opened 2017-03-21
Fallback to Go image_downsize for is_admin

mjangda:
Continued from #515

Jetpack Photon's filters aren't currently supported for the Dashboard so all images end up full size since we don't allow intermediate sizes (and there is nothing else dynamically resizing them).

Until Jetpack allows using Photon in the admin (https://github.com/Automattic/jetpack/pull/6702), we can just fallback to Go's image downsize behaviour.

### DIFF
--- a/a8c-files.php
+++ b/a8c-files.php
@@ -47,17 +47,9 @@ class A8C_Files {
 		// This will be phased out and become the default eventually.
 		$use_jetpack_photon = $this->use_jetpack_photon();
 		if ( $use_jetpack_photon ) {
-			// The files service has Photon capabilities, but is served from the same domain.
-			// Force Jetpack to use it instead of the default Photon domains (`i*.wp.com`).
-			add_filter( 'jetpack_photon_domain', function() {
-				return home_url();
-			} );
-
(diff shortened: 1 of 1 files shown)


a8c-files.php
```diff
@@ -69,7 +61,37 @@ function __construct() {
...
 
-	function use_jetpack_photon() {
+	private function init_jetpack_photon_filters() {
+		// The files service has Photon capabilities, but is served from the same domain.
+		// Force Jetpack to use it instead of the default Photon domains (`i*.wp.com`).
+		add_filter( 'jetpack_photon_domain', function() {
+			return home_url();
+		} );
+
+		// The sizes metadata is not used and mostly useless on Go so let's strip it out.
```
ethitter: Does unsetting it, versus setting it to an empty array, have any consequences? It'll surely break `srcset` support, which was likely broken to begin with given the uselessness of the data.

mjangda: Yeah, the data is pretty much useless right now and I think empty array probably makes more sense. Would probably help fix the issue I ran into writing the script in #524 :)

mjangda:
r90335-deploy

PR #524, opened 2017-03-21
Add wp-cli command to remove `sizes` metadata

mjangda:
The data is not very useful on Go sites since all intermediate images are dynamically generated.

Related #515 and #523 (although this conflicts/breaks with the `wp_get_attachment_metadata` filter in the latter).


mjangda:
Not needed right now.

PR #525, opened 2017-03-21
Add function to remove core's meta_key INDEX

mjangda:
dbDelta does not drop indexes so we need to do it manually.

TODO: 
- [ ] Find a way to automagically call this after `dbDelta`

Related #445


joshbetz:
Do we need to be worried about calling it automatically in the case that the meta_key index doesn't exist?

mjangda:
Closing; we need to remove the index from the schema itself otherwise dbDelta ends up restoring it.

PR #530, opened 2017-03-24
Disable update checks as they're a waste of cron tasks

ethitter:
There's no need to schedule update checks, since we don't support updates through the dashboard, nor is the updates UI accessible.


mjangda:
We chatted about this before and kept them in because they also send useful stats to .org around versions and plugin usage.

ethitter:
I'm conflicted. I understand the desire to report, but the `wp_version_check` is probably the only useful one of the three. Most sites use a custom theme, and while some of their plugins are available on W.org, most are either custom or a fork. That all said, it's probably not worth reimplementing the Core check to bypass the other two.

This arose while working on Cron Control, and thinking about prioritizing client events over other things. Since that plugin already curates the event list, I can handle this there.

PR #531, opened 2017-03-24
Customize install defaults

ethitter:
* Stop creating the default post, comment, and page
* Allow other code to initialize itself during install

The `function_exists` check is a necessary quirk of Core. Without it, our `wp_install_defaults()` runs during install, but then causes a "cannot redeclare" fatal error once WP is installed.

Also worth noting that the contents of `wpcom_vip_install_defaults_from_core()`, with the exception of the `// Multisite love` line, are verbatim from Core's `wp_install_defaults()`. Hence the curiosities like `global_terms_enabled()`.

### DIFF
--- a/wp-install-defaults.php
+++ b/wp-install-defaults.php
@@ -4,17 +4,60 @@
  * Description: Set default values during install
  */
 
+if ( ! function_exists( 'wp_install_defaults' ) ) :
 /**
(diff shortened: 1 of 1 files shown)


ethitter:
In hindsight, this is unnecessary and introduces more duplication from Core, which we should forego.

PR #532, opened 2017-03-24
Fix vip plugins

pkevan:
Adds commands from WP.com for managing shared plugins via CLI


mjangda:
Nice. Is the `get_vip_plugins` method the only changed bit between here and WP.com? If so, we can add a filter which could simplify syncing (like you had noted).

pkevan:
Yes, apart from the removal of methods to get all plugins details by blog stickers, but a filter would still make this better.

mjangda:
r94216-deploy

PR #533, opened 2017-03-24
two-factor: Handle when contributor doesn't exist

mjangda:
Just use a hard-coded set of caps instead.


joshbetz:
We can probably wait to deploy this with #534

joshbetz:
r90763-deploy

PR #534, opened 2017-03-24
Two Factor: Only filter map_meta_cap for logged in users

joshbetz:
Otherwise we do many checks for whether the current user has two factor
enabled that are unnecessary.

### DIFF
--- a/two-factor.php
+++ b/two-factor.php
@@ -49,9 +49,7 @@ function wpcom_vip_force_two_factor() {
 	return apply_filters( 'wpcom_vip_force_two_factor', false );
 }
 
-function wpcom_vip_load_two_factor_plugin() {
-	wpcom_vip_load_plugin( 'two-factor' );
-
+function wpcom_vip_enforce_two_factor_plugin() {
 	if ( is_user_logged_in() ) {
 		add_action( 'admin_notices', 'wpcom_vip_two_factor_admin_notice' );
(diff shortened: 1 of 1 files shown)


two-factor.php
```diff
@@ -52,8 +52,10 @@ function wpcom_vip_force_two_factor() {
 function wpcom_vip_load_two_factor_plugin() {
 	wpcom_vip_load_plugin( 'two-factor' );
 
-	add_action( 'admin_notices', 'wpcom_vip_two_factor_admin_notice' );
-	add_filter( 'map_meta_cap', 'wpcom_vip_two_factor_filter_caps' );
+	if ( get_current_user_id() > 0 ) {
```
mjangda: `is_user_logged_in()`?

two-factor.php
```diff
@@ -70,7 +72,7 @@ function wpcom_enable_two_factor_plugin() {
 		wpcom_vip_load_two_factor_plugin();
 	}
 }
-add_action( 'setup_theme', 'wpcom_enable_two_factor_plugin' );
+add_action( 'set_current_user', 'wpcom_enable_two_factor_plugin' );
```
mjangda: Should we load the plugin on `plugins_loaded` and then delay adding the notice/filter caps until `set_current_user`? Feels that that would avoid any potential load order issues?

joshbetz:
r98392-deploy

PR #538, opened 2017-03-27
Move the WordPress Importer to shared-plugins

joshbetz:


mjangda:
Should we just remove completely since shared-plugins will be going away?

PR #539, opened 2017-03-27
Unsubmodule vip-support

joshbetz:


mjangda:
Will Travis automatically pick up the Travis config from the subdirectory?

joshbetz:
No. Linting should already be running from the mu-plugins repo and I can make sure we also run the tests in this directory.

PR #542, opened 2017-03-27
Jetpack Start fixes

mjangda:
Two fixes:
- Fix WP_CLI namespace for jetpack keys
- Fix Akismet conditional when setting keys

See #433


mjangda:
r90575-deploy

PR #543, opened 2017-03-28
Strip dimensions from intermediate images in the_content

mjangda:
Jetpack's [dimension stripping](https://github.com/Automattic/jetpack/blob/master/class.photon.php#L867-L874) doesn't quite work for us since it requires `file_exists` lookups. For now, filter `the_content` to strip out the dimensions from image file names.


mjangda:
r90619-deploy

PR #546, opened 2017-03-29
Prevent core from doing filename lookups

### LINKED ISSUE #545: Kill _wp_attached_file lookup for media search
Being discussed in core: https://core.trac.wordpress.org/ticket/39358

For really large sites, this can be super slow even with our meta_value index in place.

mjangda:
For media search. This results in meta queries that are super slow on large sites and can take upwards of 30+ seconds for search.

see https://core.trac.wordpress.org/ticket/39358

Fixes #545


mjangda:
r90742-deploy

PR #547, opened 2017-03-29
Don't unnecessarily fall back to Core cron

ethitter:
With the introduction of 4fa4d2c, we can ping any multisite subsite that hasn't been somehow marked inactive.


ethitter:
r90725-deploy

PR #548, opened 2017-03-29
Update Cron Control to use custom table

ethitter:
Update plugin to latest release, which replaces CPT with a custom table that includes tailored constraints.

Table is not created from front-end requests, and plugin disables itself without table present, which simplifies deploying this. In response to an admin request, or a request to the events-list endpoint, the plugin will create the table on `shutdown`.

See https://github.com/Automattic/Cron-Control/pull/89


ethitter:
r90733-deploy

PR #550, opened 2017-03-30
Don't bump stats from sandboxes

### LINKED ISSUE #549: Don't bump stats when sandboxed
Use `WPCOM_SANDBOXED` to disable things like the publish stats.

ethitter:
Fixes #549


ethitter:
r90724-deploy

PR #553, opened 2017-03-30
Display the master user when found to be active

simonwheatley:
Also make language clearer: "Jetpack is already active (connected)"

### DIFF
--- a/jetpack-start/wp-cli-api.php
+++ b/jetpack-start/wp-cli-api.php
@@ -63,13 +63,17 @@ class Jetpack_Start_CLI_Command extends WP_CLI_Command {
 
 		$force_connection = WP_CLI\Utils\get_flag_value( $assoc_args, 'force', false );
 		if ( ! $force_connection && Jetpack::is_active() ) {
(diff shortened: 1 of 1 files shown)


jetpack-start/wp-cli-api.php
```diff
@@ -63,9 +63,13 @@ public function connect( $args, $assoc_args ) {
 
 		$force_connection = WP_CLI\Utils\get_flag_value( $assoc_args, 'force', false );
 		if ( ! $force_connection && Jetpack::is_active() ) {
-			WP_CLI::error( 'Jetpack is already active; bailing. Run this command with `--force` to bypass this check or disconnect Jetpack before continuing.' );
+			if ( empty( Jetpack_Options::get_option( 'master_user' ) ) ) {
```
mjangda: Minor coding style thing: It's better to keep function calls (`Jetpack_Options::get_option( 'master_user' )`) out of conditionals for readability. Best to assign to a variable and then evaluate against that.

As a bonus, in the `else` case, this helps avoid an extra `get_option` call.

mjangda:
r93143-deploy

PR #556, opened 2017-03-30
Add es debug bar panel

nickdaugherty:
Adds logging of ES queries and bespoke Debug Bar panel

Panel includes both the request and the response, along with timing and backtrace information


vip-helpers/vip-debug-bar-panels.php
```diff
@@ -437,3 +437,77 @@ function log_http_request_result( $response, $context, $transport, $args, $url )
 		}
 	}
 }
+
+class WPCOM_VIP_Debug_Bar_Elasticsearch extends Debug_Bar_Panel {
```
mjangda: Let's put all new panels into their own files?

nickdaugherty: Good call - I moved it to the `debug-bar/panels` folder and removed some of the VIP specific stuff, so maybe it can be used as a more standalone plugin in the future.

nickdaugherty:
r90953-deploy

PR #557, opened 2017-03-31
Enable Cron Control for all sites but those we can't ping externally

ethitter:
WP-Cron Control is no longer necessary. This is dependent upon D4971-code.


ethitter:
Thanks!

PR #558, opened 2017-03-31
Cron Control: update to ensure caches are cleared during install

ethitter:
Performed this manually following #548.


ethitter:
r90738-deploy

PR #559, opened 2017-03-31
Allow remote images to be served from Photon

mjangda:
Otherwise passing those URLs through `jetpack_photon_url` ends up attempting to serve the image through the Go Files service, which does not work.

Without the patch (and with `WPCOM_VIP_USE_JETPACK_PHOTON` enabled), passing a remote image through a site ends up with the domain: `https://example.go-vip.co/example.com/wp-content/file.jpg`. With the patch, we get a correct Photon URL: `https://i0.wp.com/example.com/wp-content/file.jpg`

### DIFF
--- a/a8c-files.php
+++ b/a8c-files.php
@@ -63,10 +63,16 @@ class A8C_Files {
 
 	private function init_jetpack_photon_filters() {
 		// The files service has Photon capabilities, but is served from the same domain.
-		// Force Jetpack to use it instead of the default Photon domains (`i*.wp.com`).
-		add_filter( 'jetpack_photon_domain', function() {
-			return home_url();
-		} );
+		// Force Jetpack to use the files service instead of the default Photon domains (`i*.wp.com`) for internal files.
+		// Externally hosted files continue to use the remot Photon service.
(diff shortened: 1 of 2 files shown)


a8c-files.php
```diff
@@ -63,10 +63,16 @@ function __construct() {
...
 		// The files service has Photon capabilities, but is served from the same domain.
-		// Force Jetpack to use it instead of the default Photon domains (`i*.wp.com`).
-		add_filter( 'jetpack_photon_domain', function() {
-			return home_url();
-		} );
+		// Force Jetpack to use the files service instead of the default Photon domains (`i*.wp.com`) for internal files.
+		// Externally hosted files continue to use the remot Photon service.
+		add_filter( 'jetpack_photon_domain', function( $photon_url, $image_url ) {
+			$home_url = home_url();
+			if ( wp_startswith( $image_url, $home_url ) ) {
```
joshbetz: Should we also for *.go-vip.co domains to use the files service in case it's not the home_url? I don't think we'd ever want to send those through Photon.

mjangda: 97aaf6c

mjangda:
r91401-deploy

PR #560, opened 2017-03-31
Jetpack Start: set option, then verify for Akismet

mjangda:


mjangda:
Actually; going to re-work this; might be better to attempt the registration via Jetpack instead.

PR #562, opened 2017-03-31
Fixing error with filtering capabilities

jasonbahl:
We're getting some really funky behavior with a lot of things, such as REST requests, plugin activations, etc, and I tracked it down to an error with the filtering of capabilities. 

Since we don't have a "contributor" role, this is causing some issues. Not 100% sure this PR is "THE" fix, so ya'll might want to check out if there's a better way to address, but this is causing a lot of issues.


joshbetz:
Should be fixed by #533

jasonbahl:
Sweet! Thanks!

PR #564, opened 2017-04-05
Run the `wpcom_vip_set_old_slug_redirect_cache` early 

david-binda:
By running our callback as the very first one, we would capture the value returend from the database. That means custom callbacks to `old_slug_redirect_url` run on any later callback won't pollute the cache.

By adding the `apply_filters( 'old_slug_redirect_url', $redirect )` call to before the custom `wp_redirect` one the plugin now actually makes any custom callbacks working on cached old slug redirects.

### DIFF
--- a/vip-helpers/vip-caching.php
+++ b/vip-helpers/vip-caching.php
@@ -660,7 +660,8 @@ function wpcom_vip_wp_old_slug_redirect() {
 		$redirect = wp_cache_get( 'old_slug' . $wp_query->query_vars['name'] );
 
 		if ( false === $redirect ) {
-			add_filter( 'old_slug_redirect_url', 'wpcom_vip_set_old_slug_redirect_cache', 9999, 1 );
+			// Run the caching callback as the very firts one in order to capture the value returned by WordPress from database. This allows devs from using `old_slug_redirect_url` filter w/o polluting the cache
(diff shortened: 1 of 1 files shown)


vip-helpers/vip-caching.php
```diff
@@ -660,7 +660,7 @@ function wpcom_vip_wp_old_slug_redirect() {
 		$redirect = wp_cache_get( 'old_slug' . $wp_query->query_vars['name'] );
 
 		if ( false === $redirect ) {
-			add_filter( 'old_slug_redirect_url', 'wpcom_vip_set_old_slug_redirect_cache' );
+			add_filter( 'old_slug_redirect_url', 'wpcom_vip_set_old_slug_redirect_cache', 9999, 1 );
```
mjangda: Can we please add an inline comment explaining why we're using this priority?

david-binda: Sure! Good call.

david-binda:
I have discovered a bug which would be caused by this revision. I'm working on updating the diff.

david-binda:
@mjangda @sboisvert this is not ready for a review. Thanks!

david-binda:
@mjangda @sboisvert the "not" in previous comment should be "now". Sorry for the confusion!

david-binda:
This has been deployed as r90944

PR #566, opened 2017-04-07
rm `--` when calling `user create`

mjangda:
Ends up being treated as a separate arg, which breaks things.


mjangda:
r91060-deploy

nickdaugherty:
Bummer - I guess it's WP CLI that is responsible?

mjangda:
Yep.

PR #573, opened 2017-04-13
Don't skip empty events, as it causes them to be rescheduled infinitely

ethitter:
For "efficiency" reasons, Cron Control skips events whose actions have no callbacks. Unfortunately, this causes some jobs to be rescheduled infinitely. This functionality will be fixed or removed upstream, but this stops the runaway event creation in the meantime.


ethitter:
r91370-deploy

PR #575, opened 2017-04-14
Update Cron Control to set `DOING_CRON` consistently

ethitter:
See https://github.com/Automattic/Cron-Control/pull/99


ethitter:
r91429-deploy

PR #576, opened 2017-04-14
Update Cron Control

ethitter:
List endpoint now includes count of pending jobs, to permit monitoring and alerting

See https://github.com/Automattic/Cron-Control/pull/101


ethitter:
r91448-deploy

PR #577, opened 2017-04-16
Use require_once for loading mu-plugins in tests

mjangda:
To avoid issues in test environments where `mu_plugins` is already loaded when running tests.


mjangda:
No deploy here since it's just test code.

PR #581, opened 2017-04-18
Include our helpers for any network or site plugins loaded outside of our custom methods

### LINKED ISSUE #580: Plugins loaded via Core's option don't include `vipgo-helper.php`
Plugins loaded via the Core plugins screen, which updates the `active_plugins` option, won't include the `vipgo-helper.php` for the plugin; the helper is only included if `wpcom_vip_load_plugin()` is used. For plugins such as msm-sitemap, this is problematic, as there are performance improvements in the helper that won't be loaded.

One option might be to merge `active_plugins` into `wpcom_vip_active_plugins` on `shutdown` or some other action, allowing the existing handling to take over for the next and all subsequent requests.

ethitter:
Network-activated plugins, and community plugins activated via the wp-admin UI, won't have their `vipgo-helper.php` loaded. In many cases, performance- and architecture-related changes are omitted as a result.

For a plugin like msm-sitemap, this causes performance issues when generating sitemaps.

This is a continuation of https://github.com/Automattic/vip-dashboard/pull/71, properly situated and incorporating feedback from @mjangda.

Fixes #580

### DIFF
--- a/vip-helpers/vip-utils.php
+++ b/vip-helpers/vip-utils.php
@@ -1097,8 +1097,6 @@ add_action( 'muplugins_loaded', 'wpcom_vip_load_helpers_for_network_active_plugi
 /**
  * Load `vipgo-helper.php` if it exists for a plugin loaded outside of our custom UI and helpers
  *
- * Loaded at priority 6 because all plugins are typically loaded before 'plugins_loaded', and the UI-enabled plugins use priority 5
- *
  * Technically tries to include the main plugin file again, but we don't care, because it uses `include_once()` and is called after Core loads the plugin
  */
 function wpcom_vip_load_helpers_for_sites_core_plugins() {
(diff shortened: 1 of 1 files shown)


vip-helpers/vip-utils.php
```diff
@@ -1077,6 +1077,41 @@ function wpcom_vip_plugin_is_loaded( $plugin ) {
...
+	// wp_get_active_network_plugins() won't exist otherwise
+	if ( ! is_multisite() ) {
+		return;
+	}
+
+	foreach ( wp_get_active_network_plugins() as $plugin ) {
+		_wpcom_vip_include_plugin( $plugin );
+	}
+}
+add_action( 'muplugins_loaded', 'wpcom_vip_load_helpers_for_network_active_plugins' );
```
mjangda: I guess we don't need to worry about the priority here?

ethitter: Correct, since we aren't doing anything else related to plugin-loading on this action.

ethitter:
> I was thinking we could be clever and noop the `active_plugins` option and just take care of loading things ourselves but that might be over-complicating things :)

That was my first thought, but plugins UI that comes from the VIP Dashboard uses Core's plugins list table, so the `active_plugins` option will keep getting repopulated with "community" plugins. Were that not the case, moving the values from `active_plugins` to `wpcom_vip_active_plugins` would've solved most of this (save network-activated plugins), and was one of my earlier approaches.

ethitter:
r91558-deploy

PR #582, opened 2017-04-18
Force a lower timeout for VaultPress

mjangda:
To prevent issues in the rare cases when the VaultPress API is slow to respond. 10s is a bit arbitrary but better than the default 60s.


ethitter:
:shipit:

mjangda:
r91620-deploy

PR #583, opened 2017-04-19
Fix image_downsize filer for no image.

philipjohn:
When the `image_downsize` filter is called but there is no valid image for the given `$id`, the hooked function in the files service continues to try and manipulate the image URL despite there being no image.

The result of this is that functions like `wp_get_attachment_image_url()` that should return false in this situation return a URL based on the request URI instead.

This change checks if a valid image URL is available and returns false at the earliest opportunity if there is no image URL to process.


philipjohn:
Closing in favour of #392

PR #584, opened 2017-04-20
Override any client code that might block access to the Cron Control REST endpoints

ethitter:
Plugin performs its own authorization, and blocking this breaks cron execution for the site.


ethitter:
r91723-deploy

PR #585, opened 2017-04-20
Caching some expensive queries for media

CodeProKid:
Hi VIP folks, was going to add a plugin to our own repo to do this, but figured other sites on go might also want to take advantage of these new filters from 4.7.4? I assume we aren't the only sites that are seeing these queries be super slow. Let me know if that's not the case, and it would be better to just put this in our own codebase though.

This PR takes advantage of 3 new filters added from these commits -> https://core.trac.wordpress.org/changeset/40382 & https://core.trac.wordpress.org/changeset/40421

Essentially this just stores the results of 3 notoriously slow queries in a transient for ~1 hour. I'm also randomizing the expiration so we don't end up with a waterfall of expensive queries if they all expire at the same time. I didn't add any transient busting on media upload actions since that could be happening quite often on high traffic sites, but could potentially add that so the data returned from the query would always be accurate. Personally, I think that's a little overkill, but I'm open to discussion/suggestions. Thanks!

### DIFF
--- a/performance/media.php
+++ b/performance/media.php
@@ -1,7 +1,95 @@
 <?php
 
+namespace Automattic\VIP\Performance;
+
 // Prevent core from doing filename lookups for media search.
 // https://core.trac.wordpress.org/ticket/39358
 add_action( 'pre_get_posts', function() {
 	remove_filter( 'posts_clauses', '_filter_query_attachment_filenames' );
 } );
+
+/**
+ * Stores query for audio items in a transient so it doesn't have to run on every page load
+ *
+ * @param null $show_audio_playlist
+ * @return mixed|null|string
+ */
+function cache_has_audio_query( $show_audio_playlist ) {
+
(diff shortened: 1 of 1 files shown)


mjangda:
Thanks for the PR @CodeProKid! For audio and video what do you think about just returning true from the filter and skipping the query completely?

CodeProKid:
@mjangda I'd be open to that, and it would be ok for our uses, but I figured that other sites might actually want to use those audio/video filters in the UI, so caching the query results in the transient would limit how often the query would run, but still enable use of the filter in the UI.

joshbetz:
I think we could cache these forever if we hook into `save_post` and invalidate them as necessary.

CodeProKid:
Thanks for the feedback @joshbetz I'll address this in the coming days.

joshbetz:
@CodeProKid Check out https://github.com/Automattic/vip-go-mu-plugins/pull/592 - Looks like there's some overlap there.

CodeProKid:
@joshbetz yeah, that PR looks much more complete than mine. I'll close this out.

PR #586, opened 2017-04-25
Add helper for IP forwarding with key verifcation

mjangda:
TODO:
- [ ] Tests
- [ ] Should we pass in the actual key into the method instead of constant?

### DIFF
--- a/lib/proxy/ip-forward.php
+++ b/lib/proxy/ip-forward.php
@@ -2,6 +2,18 @@
 
 namespace Automattic\VIP\Proxy;
 
+function is_valid_ip( $ip ) {
+	if ( ! filter_var( $ip, FILTER_VALIDATE_IP, FILTER_FLAG_IPV4 )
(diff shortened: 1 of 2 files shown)


lib/proxy/ip-forward.php
```diff
@@ -79,3 +89,28 @@ function fix_remote_address_from_ip_trail( $ip_trail, $proxy_ip_whitelist ) {
...
+ *
+ * @return (bool) true, if REMOTE_ADDR updated; false, if not.
+ *
+ */
+function fix_remote_address_with_verification_key( $user_ip, $verification_key ) {
+	if ( ! is_valid_ip( $user_ip ) ) {
+		return false;
+	}
+
+	if ( ! hash_equals( $verification_key, WPCOM_VIP_IP_VERIFICATION_KEY ) ) {
```
simonwheatley: The constant we're setting is called `WPCOM_VIP_PROXY_VERIFICATION`.

mjangda:
@simonwheatley read for another look.

mjangda:
r94939-deploy

PR #589, opened 2017-04-26
Handle cases when QueryMonitor doesn't fully load

### LINKED ISSUE #588: Query Monitor + REST API Fatal
When logged in with QM enabled and you try to hit a REST API endpoint:

```
PHP Fatal error:  Class 'QM_Collector_DB_Queries' not found in wp-content/mu-plugins/vip-helpers/vip-query-monitor.php on line 3
```

mjangda:
We should stop loading additional bits like custom classes and other filters when this happens to avoid unexpected behaviour like fatals.

Fixes #588


mjangda:
r92265-deploy

PR #591, opened 2017-04-27
Don't trigger Jetpack Sync on shutdown if the request is for Cron Control

### LINKED ISSUE #590: When using Cron Control, should we disable Jetpack Sync's shutdown actions?
Cron Control ensures that Sync's two cron events run consistently, and the shutdown hook impacts all cron requests, including those that just ran the sync actions.

Should we use `jetpack_sync_sender_should_load` to disable it and confine the sync sender to its cron events? We could limit our callback to `DOING_CRON`, leaving the shutdown actions on admin requests, front-end requests for logged-in users, and POST requests.

https://github.com/Automattic/jetpack/blob/8abb597c27495be84e3b8dc5d04d0fd65cf15537/sync/class.jetpack-sync-actions.php#L60

ethitter:
Cron Control runs JP sync events, it shouldn't also trigger them on shutdown. In some cases, the sync action delays the REST response sufficiently to trigger a timeout in the runner.

Fixes #590

### DIFF
--- a/001-cron.php
+++ b/001-cron.php
@@ -49,6 +49,19 @@ function wpcom_vip_permit_cron_control_rest_access( $allowed ) {
 }
 
 /**
+ * Don't trigger Jetpack Sync's shutdown actions for cron requests
+ *
+ * Cron runs sync itself, and running sync on shutdown slows the endpoint response, sometimes beyond the 10-second timeout
+ */
+function wpcom_vip_disable_jetpack_sync_on_cron_shutdown( $load_sync ) {
+	if ( defined( 'DOING_CRON' ) && DOING_CRON ) {
+		return false;
+	}
+
+	return $load_sync;
+}
+
+/**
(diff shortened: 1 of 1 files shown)


ethitter:
Correct, the jobs are scheduled, their actions set, and the sender initialized via other callbacks in the `Jetpack_Sync_Actions` class. This filter is specific to the `shutdown` use, even though the name doesn't make that clear.

* The filter: https://github.com/Automattic/jetpack/blob/8abb597c27495be84e3b8dc5d04d0fd65cf15537/sync/class.jetpack-sync-actions.php#L60
* Cron's handlers: https://github.com/Automattic/jetpack/blob/8abb597c27495be84e3b8dc5d04d0fd65cf15537/sync/class.jetpack-sync-actions.php#L23.

ethitter:
r91969-deploy

PR #592, opened 2017-04-28
Update to reduce slow media queries

Kluny:
Ticket about slow queries on Laughing Squid: https://wordpressvip.zendesk.com/agent/tickets/65236

The slowness is the result of a select-distinct query when looking for media. Select-distinct is already expensive so it should be cached. But if it doesn't get any results, it still has to do a bunch of complicated processing, returns nothing, and nothing isn't cached, so the query stays slow. 

This filter caches the nothing, as it were. It is active on Wordpress.com and tests okay on LaughingSquid.com, a VIP site. 

Here's context for the bug:  https://keepingtheirblogsgoing.wordpress.com/2016/04/27/speeding-up-wp-admin/
Here's the same code that was pushed on wp.com on 19 April: https://wpcom.trac.automattic.com/changeset/154832

### DIFF
--- a/performance/vip-tweaks.php
+++ b/performance/vip-tweaks.php
@@ -76,12 +76,12 @@ if ( is_admin() ) {
 		$months = get_transient( 'wpcom_media_months_array' );
 
 		if ( false === $months ) {
-        	$months = $wpdb->get_results( $wpdb->prepare( "
-            	SELECT DISTINCT YEAR( post_date ) AS year, MONTH( post_date ) AS month
-            	FROM $wpdb->posts
-            	WHERE post_type = %s
-            	ORDER BY post_date DESC
-        	", 'attachment' ) );
+        		$months = $wpdb->get_results( $wpdb->prepare( "
+            		     SELECT DISTINCT YEAR( post_date ) AS year, MONTH( post_date ) AS month
+            			FROM $wpdb->posts
+            			WHERE post_type = %s
+            			ORDER BY post_date DESC
+        			", 'attachment' ) );
 			set_transient( 'wpcom_media_months_array', $months );
 		}
 
--- a/performance/vip-tweaks.php
+++ b/performance/vip-tweaks.php
@@ -33,19 +33,7 @@ function wpcom_vip_disable_performance_tweaks() {
 	remove_action( 'after_setup_theme', 'wpcom_vip_enable_performance_tweaks' );
 }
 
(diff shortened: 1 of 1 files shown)


Kluny:
Hey Stéphane, I've updated the commit with cache-busting functions and comments. Please review.

sboisvert:
Is deleting the cache for  wpcom_media_has_video still needed? Where is that transient set? Perhaps that's dead code on WordPress.com that we should remove now that 4.7.4 has landed?

performance/vip-tweaks.php
```diff
@@ -32,3 +32,72 @@ function wpcom_vip_enable_performance_tweaks() {
...
+		! $months->month == get_the_time( 'm', $post_id )
+	) {
+		// the new attachment is not in the same month/year as the
+		// most recent attachment, so we need to refresh the transient
+		delete_transient( 'wpcom_media_months_array' );
+	}
+
+}
+
+if ( wpcom_is_vip() && is_admin() ) {
```
joshbetz: `wpcom_is_vip()` is technically deprecated on VIP Go, though it should always return true.

performance/vip-tweaks.php
```diff
@@ -32,3 +32,72 @@ function wpcom_vip_enable_performance_tweaks() {
...
+	add_filter( 'media_library_show_audio_playlist', '__return_true' );
+
+	add_filter( 'media_library_months_with_files', 'wpcom_vip_media_library_months_with_files' );
+	function wpcom_vip_media_library_months_with_files() {
+ 		global $wpdb;
+
+		$months = get_transient( 'wpcom_media_months_array' );
+
+		if ( false === $months ) {
+        	$months = $wpdb->get_results( $wpdb->prepare( "
```
joshbetz: Looks like an indentation issue here?

Kluny:
Hey Stéphane, I removed the two bits of code that we discussed earlier. Please review.

mjangda:
@sboisvert Did you deploy this?

sboisvert:
No, I had talked with @joshbetz and was under the impression he was deploying it. I might be wrong. Happy to do it tomorrow when I'm on solid ground / solid wifi.

mjangda:
Okay, in the future, let's not merge PRs unless you're also deploying them.

joshbetz:
r92811-deploy

PR #597, opened 2017-05-03
remove blank_media_fix()

Kluny:
The function blank_media_fix() was a temporary fix for a tricky problem with queries related to post_mime_type (audio and video uploads).  The original ticket is here, https://core.trac.wordpress.org/ticket/31071 and the official fix, which properly caches media queries each time media is uploaded, is here: https://github.com/Automattic/vip-go-mu-plugins/pull/592

The function has been removed since it's no longer needed.


mjangda:
r92266-deploy

PR #600, opened 2017-05-04
Support different domains for home and site URL

ethitter:
For a consistent experience, particularly when the `home` URL _isn't_ secure, a number of accommodations are required:

- [X] Filter post preview URLs to use `siteurl`
- [X] Redirect logged-out requests from the `siteurl` to `home`
- [x] Rewrite static asset paths to use `home`, ensure that works with ngx-http-concat
- [ ] Filter upload URLs in post content, to use `home`


ethitter:
The redirect may need to be refined to limit the requests it applies to. Requires testing.

ethitter:
The static-asset handling needs fixing for subdirectory multisite, as my site is a subdomain install.

lib/split-home-site-urls/asset-urls.php
```diff
@@ -0,0 +1,309 @@
...
+	 * Silence is golden!
+	 */
+	private function __construct() {}
+
+	/**
+	 * Instantiate singleton
+	 */
+	public static function get_instance() {
+		if ( ! is_a( self::$instance, __CLASS__ ) ) {
+			self::$instance = new self;
```
nickdaugherty: Probably should use empty parens `()` here - I thought that was part of the WP standards but I didn't actually see it.

lib/split-home-site-urls/asset-urls.php
```diff
@@ -0,0 +1,309 @@
...
+	 * Ensures images not processed with Photon are at least served from the CDN
+	 *
+	 * @param string $url
+	 * @filter pre_option_upload_url_path
+	 * @return string
+	 */
+	public function filter_upload_url_path( $url ) {
+		$url = 'https://' . $this->static_host . '/wp-content/uploads';
+
+		if ( 2 === get_current_blog_id() ) {
```
nickdaugherty: Is this right? I would have expected something like `if ( 1 !== get_current_blog_id() )`.

lib/split-home-site-urls/asset-urls.php
```diff
@@ -0,0 +1,309 @@
...
+		// Don't rewrite PHP URLs
+		$extension = pathinfo( $src, PATHINFO_EXTENSION );
+		if ( 0 === strpos( $extension, 'php' ) ) {
+			return $src;
+		}
+
+		// Attempt to rewrite if proper conditions are met.
+		$parsed_host = parse_url( $src, PHP_URL_HOST );
+		if ( $parsed_host && $parsed_host !== $this->static_host && $this->should_staticize( $parsed_host ) && apply_filters( 'wpcom_vip_asset_urls_staticize', true, $parsed_host, $src, $context ) ) {
+			$src = str_replace( $parsed_host, $this->static_host, $src );
```
nickdaugherty: We should put a limit of `1` on the `str_replace()`, so that we don't affect other parts of the url, like GET params - we just want to change the hostname.

lib/split-home-site-urls/login.php
```diff
@@ -0,0 +1,52 @@
...
+ * Rewrite any admin URLs to use the site URL, not the home URL
+ */
+function rewrite_admin_url_to_site_url( $redirect_to ) {
+	// Nothing to correct if it's not a wp-admin link
+	if ( empty( $redirect_to ) || false === stripos( $redirect_to, 'wp-admin' ) ) {
+		return $redirect_to;
+	}
+
+	// Grab the path relative to wp-admin
+	preg_match( '#/wp-admin(/)?(.+?)$#i', $redirect_to, $admin_paths );
```
nickdaugherty: Another non-regex (maybe faster and simpler) approach would be to `explode()` the string on `/wp-admin` with a limit of 1.

lib/split-home-site-urls/asset-urls.php
ethitter: That was some cruft left from porting the class from my site. That whole method is going away, as it won't work if there's no SSL certificate for the `home` URL.

lib/split-home-site-urls/asset-urls.php
ethitter: Fair point. I'll switch this to `preg_replace()`, since `str_replace()` doesn't have a limit option (only count).

lib/split-home-site-urls/login.php
```diff
@@ -12,11 +12,11 @@ function rewrite_admin_url_to_site_url( $redirect_to ) {
 	}
 
 	// Grab the path relative to wp-admin
-	preg_match( '#/wp-admin(/)?(.+?)$#i', $redirect_to, $admin_paths );
+	$exploded = explode( '/wp-admin', $redirect_to );
```
nickdaugherty: On `explode()` here, we should set a limit of `1`, on the off chance that other parts of the url contain it, like in a GET param.

ethitter:
See #605 for continuation

PR #602, opened 2017-05-04
Don't cache sandboxed requests by default

### LINKED ISSUE #598: Disable page cache while sandboxed
We show a "Sandboxed" badge while sites are sandboxed, we should also disable the page cache. It's confusing to see "sandboxed" and get cached responses. Also, if you have the site sandboxed, you probably don't want it to be cached anyway.

joshbetz:
We show a sandboxed icon on sandboxed requests. This can be confusing coming from WordPress.com, where sandboxed requests skip batcache by default. We should also skip the page cache by default. If you want to test the page cache while sandboxed, you can easily unhook this.

Fixes #598

### DIFF
--- a/vip-helpers/sandbox.php
+++ b/vip-helpers/sandbox.php
@@ -4,9 +4,11 @@ if ( ! WPCOM_SANDBOXED ) {
 	return;
 }
 
+// Don't cache sandboxed requests by default
+add_action( 'send_headers', 'nocache_headers' );
+
 add_action( 'wp_footer', 'wpcom_do_sandbox_bar', 100 );
 add_action( 'admin_footer', 'wpcom_do_sandbox_bar', 100 );
-
 function wpcom_do_sandbox_bar() {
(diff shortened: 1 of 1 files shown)


mjangda:
I actually like that our behaviour is more closely in line with what you get in a production site. What if we allowed disabling the caching behaviour using a cookie/querystring or something?

joshbetz:
Sure. I guess we could add some UI to the "Sandboxed" button to show if the page is cached or not as well. Maybe even make it so that clicking the button sets a cookie and therefore disables cache. The specific case where this came up was some confusion about why sandbox changes weren't showing up. We could write it off as a learning curve, but I think I would help to make it super clear when the page is cached on sandboxes.

sboisvert:
> Sure. I guess we could add some UI to the "Sandboxed" button to show if the page is cached or not as well. 

That would be the best of both worlds!

pippercameron:
I agree with the "best of both worlds" approach. I think that would be super helpful.

PR #603, opened 2017-05-04
Prevent unfettered access to `wp-admin/options.php`

ethitter:
In a multisite environment (like WordPress.com VIP), only Super Admins are able to view the options list/edit screen at `wp-admin/options.php`. Core uses this screen to receive POST/GET requests when updating options, or when creating custom options through the Settings API, so it can't be blocked completely. However, since the UI allows editing of string options without the associated sanitization and validation normally applied to these settings, it should be blocked. It's also not particularly useful, since it doesn't support serialized data.


ethitter:
I'm fine with closing for now. This originated from a SIRT thread, which also received mixed reactions.

PR #605, opened 2017-05-05
Move split-domain support to a shared plugin

ethitter:
In hindsight, a lib was inappropriate as the theme will be too late to load this. A plugin is a happy medium for this, as `plugins_loaded` is the earliest hook used.


ethitter:
r92379-deploy

PR #607, opened 2017-05-09
Add object-cache drop-in

mjangda:
This change lets us provide our memcache drop-in code for easier local testing. The code currently lives in (private) VIP Go internals. Note that on it's own, this doesn't do anything as it needs to be loaded manually.

Once deployed, we can update the VIP Go internals to use this instead:

TODO post-merge and deploy:

- [ ] Update docs on how to use this drop-in locally.
- [ ] Update Go internal object-cache.php file to load this one.

 We can later look at including HyperDB as well.


mjangda:
r95961-deploy (but not actually in use yet)

Issue #166, opened 2016-04-06
Support client mu-plugins


joshbetz:
- Mount mu-plugins as `vip-mu-plugins` or something less confusing.
- Include all PHP files in this directory on `muplugins_loaded`

https://github.com/WordPress/WordPress/blob/703d5bdc8deb17781e9c6d8f0dd7e2c6b6353885/wp-settings.php#L210-L223

Can we do this without breaking the `muplugins_loaded` hook? How early is `add_action` available? Can we add a hook before mu-plugins are loaded and include them there?

mjangda:
Possible simpler option:

- Add a `client-mu-plugins` folder (or something similar to `vip-skeleton`).
- Add whatever to `vipd` to make sure that folder ^^ gets included in containers.
- Add a `zzz-client-mu-plugins.php` that just loads files from `client-mu-plugins the same way `wp_get_mu_plugins` does.

The trade-off is that client mu-plugins load later (although, still before `muplugins_loaded`) but I think that's okay.

joshbetz:
We could probably even call it `mu-plugins` in the vip-skeleton repo and map it as client-mu-plugins in the container, although that may be too confusing.

PR #611, opened 2017-05-10
Add support for additional mu-plugins

mjangda:
This change allows sites to add a `client-mu-plugins` folder (open to name change) in their VIP Go repo, which we load similar to mu-plugins. The client mu-plugins are loaded after the VIP Go mu-plugins are loaded ("enforced" by the filename prefix).

As part of this change, we will probably need a few others:

- [x] Update `vip-go-skeleton` to include the folder.
- [ ] Update `vip-go-api` to make the new folder required on site init.
- [x] Update `vipd` to map the folder inside `wp-content` within web containers.
- [ ] Update docs (https://vip.wordpress.com/documentation/vip-go/local-vip-go-development-environment/)

We will also probably need to re-create web containers for any sites that want to use this.

See #166

### DIFF
--- a/z-client-mu-plugins.php
+++ b/z-client-mu-plugins.php
@@ -1,0 +1,50 @@
+<?php
+
+// Note: This file is prefixed with `z-` for load order
+
+define( 'WPCOM_VIP_CLIENT_MU_PLUGIN_DIR', WP_CONTENT_DIR . '/client-mu-plugins' );
+
+function wpcom_vip_load_client_mu_plugins() {
+	static $loaded = false;
+
+	// Prevent running this multiple times
+	if ( $loaded ) {
+		return;
+	}
+
+	$loaded = true;
+
+	// Code below is adapted from wp_get_mu_plugins()
+	$client_mu_plugins = [];
+
+	if ( ! is_dir( WPCOM_VIP_CLIENT_MU_PLUGIN_DIR ) ) {
+		return;
+	}
+
+	$dh = opendir( WPCOM_VIP_CLIENT_MU_PLUGIN_DIR );
+	if ( ! $dh ) {
+		return;
+	}
+
+	do {
+		$plugin = readdir( $dh );
+		if ( false === $plugin ) {
+			break;
+		}
+
+		if ( substr( $plugin, -4 ) === '.php' ) {
+			$client_mu_plugins[] = WPCOM_VIP_CLIENT_MU_PLUGIN_DIR . '/' . $plugin;
+		}
+	} while ( false !== $plugin );
+
+	closedir( $dh );
+
+	// Make sure plugins load in a consistent, predictable order
+	sort( $client_mu_plugins );
+
+	foreach ( $client_mu_plugins as $plugin ) {
+		include_once( $plugin );
+	}
+}
+
+wpcom_vip_load_client_mu_plugins();


z-client-mu-plugins.php
```diff
@@ -0,0 +1,46 @@
...
+			break;
+		}
+
+		if ( substr( $plugin, -4 ) === '.php' ) {
+			$client_mu_plugins[] = WPCOM_VIP_CLIENT_MU_PLUGIN_DIR . '/' . $plugin;
+		}
+	} while ( false !== $plugin );
+
+	closedir( $dh );
+	sort( $client_mu_plugins );
```
joshbetz: Presumably this is so files load in a predictable, alphabetical order instead of whatever order they happen to appear on the filesystem?

mjangda: Yep. Probably worth an inline comment.

z-client-mu-plugins.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+// Note: This file is prefixed with `z-` for load order
+
+define( 'WPCOM_VIP_CLIENT_MU_PLUGIN_DIR', WP_CONTENT_DIR . '/client-mu-plugins' );
+
+function wpcom_vip_load_client_mu_plugins() {
```
joshbetz: I think we need to call this function somewhere?

mjangda: Heh, good catch. Fixed.

z-client-mu-plugins.php
mjangda: Added comment.

mikeselander:
Thanks for writing this up and pinging me @mjangda - really appreciate the responsiveness on this request!

CodeProKid:
What's the status on this? Would love to start leveraging this.

mjangda:
Thanks for the ping @CodeProKid. This fell off our radar a bit but hoping to get it ready soon.

mjangda:
`vip-go-skeleton` update in https://github.com/Automattic/vip-go-skeleton/pull/14

mjangda:
`vipd` has been updated to map the directory inside web containers.

mjangda:
r94201-deploy

mjangda:
We're going to hold off on the vip-go-api update until we've added the folder to all existing repos/branches.

fklein-lu:
I question the way that this feature was implemented:
- Is there a need to define a global constant for the client MU plugins directory? The preference should be not to place more elements in the root namespace if it can be avoided.
- Why define a function if it is invoked only once, immediately after its definition, and if the function contains protections against being called more than once? This code is not meant to be run more than once, so it shouldn't be placed in a reusable element.
- By including the MU plugin files inside of a function, variables defined inside in these functions are not placed in the global scope, but are restricted to the scope of the function. This behaviour differs from regular MU plugins.

The preferred approach should be to write this as a script:
- No need for constants or functions.
- Variables defined inside of MU plugins will be put in the right scope, without needing to do any of the awful things that `wpcom_vip_load_plugin()` does.
- MU plugin files only get loaded once, so there's no need to protect against the client MU plugins being loaded multiple times.
- You can use temporary variables, and as long as you unset them at the bottom of the script, they won't pollute the global namespace.

mjangda:
Thanks for the feedback, @fklein-lu, especially the point about scoping. The remaining points are somewhat subjective (e.g. putting the logic inside a function would allow us to more easily add tests) but will take them into consideration as we work to fix the scoping issue.

PR #615, opened 2017-05-11
Adding vip-scrub CLI command

emrikol:
I recently had to clear out 50K posts for a site older than 30 days.  This is the script I used--I thought it might be useful in case we need to do it again.


mjangda:
The WooCommerce theme has a similar script that uses direct DB queries for the scrubbing/pruning which would probably be much faster (than `wp_delete_post`) and likely worth going with instead.

joshbetz:
@emrikol thoughts on using direct sql queries for this instead?

emrikol:
Just now seeing that I have some outstanding PRs 🤦‍♂️ 

I worry about using direct SQL queries since it would require is to clean up meta, taxonomies, terms, and possibly other things.  Not to mention having to clear out the object cache.

jasonbahl:
@emrikol, every PR you submit is outstanding! 😄 

Oh, looks like we're using different definitions

Screen Shot 2019-04-11 at 2 02 15 PM

mjangda:
I worry about adding this to our production sites. I can definitely see value in a utility like this, but worry about someone running it somewhere they shouldn't and then 🔥.

I vote for closing this out and moving to a shared utility script repo instead.

emrikol:
> I worry about adding this to our production sites. I can definitely see value in a utility like this, but worry about someone running it somewhere they shouldn't and then 🔥.
> 
> I vote for closing this out and moving to a shared utility script repo instead.

Agree.  Now that we have the "private" scripts repo as well, I think anything we can pull out of mu-plugins we should :)

PR #616, opened 2017-05-12
Adding vip-cache.php

emrikol:
This file will hopefully contain helper scripts for cache related commands.

Right now, I've started it with a command to purge a site's Varnish cache.  This is useful for automating site cloning, so a support user doesn't have to log in to every newly cloned site and manually purge the cache.  It's pretty barebones (it's got one job) but I'm open to suggestions of other additions if needed.

### DIFF
--- a/wp-cli/vip-cache.php
+++ b/wp-cli/vip-cache.php
@@ -1,0 +1,15 @@
+<?php
+
+class VIP_Cache_CLI extends WPCOM_VIP_CLI_Command {
+	/**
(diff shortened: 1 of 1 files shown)


mjangda:
Nice. I think we probably want a matching command in `vip-cli` that creates a host action to run this on production sites (running it on a sandbox container will probably only PURGE your sandbox cache).

mjangda:
r94035-deploy

Issue #609, opened 2017-05-10
wp search-replace, but for big sites


joshbetz:
`wp search-replace` breaks big sites. We probably shouldn't run it on live sites in general. In addition to the load, we need to be careful about direct DB queries that run for a long time before the cache is flushed.

danielbachhuber:
Happy to discuss issues opened against https://github.com/wp-cli/search-replace-command/issues. I'd love to see `wp search-replace` improved for everyone.

mjangda:
Yep, we'll upstream whatever we end up with.

mjangda:
We're using https://github.com/Automattic/go-search-replace

PR #617, opened 2017-05-15
Add modified search-replace command

mjangda:
"Forking" the wp-cli `search-replace` command with a few modifications for better handling larger databases.

For each column in each table, we loop through in batches of 5000 based on primary using a `WHERE {key} BETWEEN {start} AND {end}` clause. This is better than a single `UPDATE` query, which on really large tables can sometimes take hours and days.

Sample output:

```
Checking: wp_posts.post_content
Max ID for column: 120625
- (DRY RUN) Processing row IDs 0 to 5000
-- 0 rows affected using SQL (in 0.001s).
- (DRY RUN) Processing row IDs 5001 to 10001
-- 82 rows affected using SQL (in 0.001s).
- (DRY RUN) Processing row IDs 10002 to 15002
-- 45 rows affected using SQL (in 0s).
- (DRY RUN) Processing row IDs 15003 to 20003
...
```

Once this has been tested as working well, we need to send our recommendations/updates upstream to: https://github.com/wp-cli/search-replace-command

See #609

### DIFF
--- a/wp-cli/vip-migrations.php
+++ b/wp-cli/vip-migrations.php
@@ -5,6 +5,152 @@ use \WP_CLI\Utils;
 class VIP_Go_Migrations_Command extends WPCOM_VIP_CLI_Command {
 
 	/**
+	 * Deletes senstive options
+	 *
+	 * @subcommand scrub-site
+	 */
+	public function clear_sensitive_options( $args, $assoc_args ) {
+		WP_CLI::log( 'Clearing sensitive options (like Jetpack connection).' );
+
+		// TODO: if called from preprod site with production values, does it disconnect the preprod site?
+//		WP_CLI::log( '- Disconnecting Jetpack' );
+//		Jetpack::disconnect();
+
+		WP_CLI::log( '- Deleting Options' );
+		$options = [
+			'jetpack_options',
+			'jetpack_private_options',
+			'vaultpress',
+			'wordpress_api_key',
+		];
+
+		foreach ( $options as $option_name ) {
+			WP_CLI::log( '-- ' . $option_name );
+			delete_option( $option_name );
+		}
+	}
+
+	/**
+	 * Update URLs across a whitelist of table columns
+	 *
+	 * ## OPTIONS
+	 *
+	 * --from=<from_url>
+	 * : The URL to search for.
+	 *
+	 * --to=<to_url>
+	 * : The URL to replace with.
+	 *
+	 * ## EXAMPLES
+	 *
+	 *     wp vip migration search-replace-url 'http://example.com' 'http://example.go-vip.co'
+	 *
+	 * @subcommand search-replace-url
+	 */
+	public function search_replace_url( $args, $assoc_args ) {
+		$from_url = $assoc_args['from'] ?? false;
+		$to_url = $assoc_args['to'] ?? false;
+
+		$from_url_parsed = parse_url( $from_url );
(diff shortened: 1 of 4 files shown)


mjangda:
May need to consider adding some pauses in between columns or batches of queries to give slave databases some time to catch up.

simonwheatley:
From a skim of the code, this addresses the replication concern with large datasets, but not the issue whereby the longer the search/replace takes the longer caches are drifting out of "true" from the database.

Could this be addressed with cache flushes in between batches or is there a concern with flushing the cache before completing the search and replace, e.g. having done the search/replace on posts but not on postmeta?

joshbetz:
> Could this be addressed with cache flushes in between batches or is there a concern with flushing the cache before completing the search and replace, e.g. having done the search/replace on posts but not on postmeta?

If we do something like that, I think we should make it optional. Another solution to the problem would be to block access to the site while search-replace is running.

wp-cli/vip-migrations.php
```diff
@@ -5,6 +5,152 @@
...
+
+		$to_url_parsed = parse_url( $to_url );
+		if ( empty( $to_url ) || ! $to_url_parsed ) {
+			WP_CLI::error( sprintf( 'Please provide a valid `--to` URL (current: %s)', $to_url ) );
+			return;
+		}
+
+		// TODO: should we have a flag to exclude home/siteurl option?
+
+		$tables_and_columns = [
```
mjangda: Probably should go with a blacklist instead since we miss custom tables.

mjangda:
> but not the issue whereby the longer the search/replace takes the longer caches are drifting out of "true" from the database.

@simonwheatley remind me why that's an issue?

joshbetz:
@mjangda I think this comes from my concern that while a search-replace is running on a live site, the cache is out of date and therefore any users updating content could overwrite our changes. It's probably not a huge problem and we can further mitigate against it by attempting to run search-replace before launch.

simonwheatley:
>> but not the issue whereby the longer the search/replace takes the longer caches are drifting out of "true" from the database.
> @simonwheatley remind me why that's an issue?

Maybe it's not a big issue? 😃 For a large dataset, the search and replace might take significant time. During this time the database is being changed, but the object cacheremains in the previous state, so the editorial team are effectively loading "old" data, and saving it to overwrite the work of the search/replace.

simonwheatley:
Certainly sometimes a search and replace can take a while 😳

> Good morning!  I ran the following command on a Multisite install (17K sites) starting at 3pm yesterday. It's still not done.
> `wp search-replace http://www.domain.tld/blogs https://blogs.domain.tld/blogs --network --dry-run`
> – https://wordpress.slack.com/archives/C02RP4T41/p1495801031781710

mjangda:
A few other things to look at:

- [ ] Skip a blacklist of columns where we know that URLs will definitely not exist (e.g. `autoload`, `to_ping`, etc.)
- [ ] Ability to parallelize PHP search-replace by doing it in batches
- [ ] [Cache flushes](https://github.com/Automattic/vip-go-mu-plugins/pull/617#issuecomment-304281699)?

joshbetz:
I was wondering if we should close this out since we have the `go-search-replace` tool, but it's so far along that it may be worth finishing up?

mjangda:
I think there are some non-VIP-specific improvements that we should send upstream, but good to close otherwise.

mjangda:
No immediate plans to work on this.

PR #618, opened 2017-05-15
Add more strict conditional for checking incorrect results

david-binda:
In case the `es_api_search_index` function returns any errors, those are assigned to `WPCOM_elasticsearch::search_result` property - it may either be a WP_Error, false or an array containing an `error` key.

The code is already properly checking for those in https://github.com/Automattic/vip-go-mu-plugins/blob/930f369d80be9d75eeb83b4c22a24e774880fb41/shared-plugins/vip-go-elasticsearch/vip-go-elasticsearch.php#L289 but the conditional in https://github.com/Automattic/vip-go-mu-plugins/blob/930f369d80be9d75eeb83b4c22a24e774880fb41/shared-plugins/vip-go-elasticsearch/vip-go-elasticsearch.php#L129 is doing looser check on the same data, and thus may produce a PHP Warning.

Example value of `$this->search_result` which produced a Warning is as follows:

```
array (
  'error' => 'Bad Request',
  'message' => 'Elasticsearch can never page beyond 10k results.',
)
```

This commit is using the same conditional as on line 289 on line 129 in order to address the PHP Warnings.

### DIFF
--- a/shared-plugins/vip-go-elasticsearch/vip-go-elasticsearch.php
+++ b/shared-plugins/vip-go-elasticsearch/vip-go-elasticsearch.php
@@ -126,7 +126,7 @@ class WPCOM_elasticsearch {
 		if ( ! $query->is_main_query() || ! $query->is_search() )
 			return $posts;
 
-		if ( ! is_array( $this->search_result ) )
+		if ( is_wp_error( $this->search_result ) || ! is_array( $this->search_result ) || empty( $this->search_result['results'] ) || empty( $this->search_result['results']['hits'] ) )
 			return $posts;
 
 		// This class handles the heavy lifting of transparently switching blogs and inflating posts


shared-plugins/vip-go-elasticsearch/vip-go-elasticsearch.php
```diff
@@ -126,7 +126,7 @@ public function filter__the_posts( $posts, $query ) {
 		if ( ! $query->is_main_query() || ! $query->is_search() )
 			return $posts;
 
-		if ( ! is_array( $this->search_result ) )
+		if ( is_wp_error( $this->search_result ) || ! is_array( $this->search_result ) || empty( $this->search_result['results'] ) || empty( $this->search_result['results']['hits'] ) )
```
mjangda: Would this check (`empty( $this->search_result['results']['hits'] )`) ignore the case where the search query finds `0` results?

david-binda: The `$this->search_result` may be an object of WP_Error ( see https://github.com/Automattic/vip-go-mu-plugins/blob/930f369d80be9d75eeb83b4c22a24e774880fb41/shared-plugins/vip-go-elasticsearch/vip-go-elasticsearch.php#L282 ) and thus treating the object as an array would produce PHP Fatal error.

mjangda: Right, which is covered by the addition of the is_wp_error check. My concern was the added check for `hits` since that's going beyond the scope of the issue. Might be better to more explicitly look for the `error` key instead? Not a huge deal since the behaviour ends up the same; just trying to prevent potential future bugs.

mjangda: /cc @nickdaugherty Thoughts?

nickdaugherty: Yes - `$this->search_result['results']['hits']` can be an empty array under normal circumstances, which would trigger the `empty()` check and incorrectly return the pre-filtered `$posts` instead of an empty array.

In practice the pre-filtered `$posts` is probably correct as-is due to how all the filters work in this plugin...but if we were to switch to `posts_pre_query` like we do in Jetpack Search, this would break and cause the query to fall back to a regular db query, as `$posts` will be `null` in that situation.

IMO it's better to handle an empty array of hits separately (since it's normal, expected behavior), explicitly returning an empty array like this:

https://github.com/Automattic/jetpack/blob/ee633d91d353eef2da953fd21791913121cfc45d/modules/search/class.jetpack-search.php#L187-L189

mjangda:
See #657

PR #621, opened 2017-05-16
Update Cron Control to latest version

ethitter:
Provides a way to run certain events in parallel, based on a whitelist.

See https://github.com/Automattic/Cron-Control/pull/108.


ethitter:
r92782-deploy

Issue #594, opened 2017-05-01
Jetpack Start: Multisite Support


mjangda:
It would nice to have a way to run the `connect` command against all sites in a network.

ethitter:
I started looking at this today; should have something to review either tonight or tomorrow.

Archived, spammed, and deleted sites will be omitted. Otherwise, this should be pretty straightforward for the user.

Should the `cancel` command also get MS support? (If so, that one would cover every site in the blogs table, regardless of status)

mjangda:
This is done.

PR #623, opened 2017-05-17
Jetpack Start for multisites: connections

ethitter:
We should be able to start a multisite without manually looping through subsites. To avoid doing so unexpectedly, the `--network` flag is introduced to the `wp jetpack-start api connect` command.

See #594

### DIFF
--- a/jetpack-start/wp-cli-api.php
+++ b/jetpack-start/wp-cli-api.php
@@ -66,8 +66,6 @@ class Jetpack_Start_CLI_Command extends WP_CLI_Command {
 			WP_CLI::warning( 'JETPACK_DEV_DEBUG mode is enabled. Please remove the constant after connection.' );
 		}
 
-		$sites = [ false ];
-
 		$network = WP_CLI\Utils\get_flag_value( $assoc_args, 'network', false );
 		if ( $network && is_multisite() ) {
 			$sites = get_sites( [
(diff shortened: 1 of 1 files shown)


ethitter:
The `WP_CLI::error()` create a poor experience in multisite, as one has to either force everything, or cancel everything first. Otherwise, the process gets interrupted and one has yet more things to cancel before trying again.

jetpack-start/wp-cli-api.php
```diff
@@ -61,34 +66,64 @@ public function connect( $args, $assoc_args ) {
...
+			$force_connection = WP_CLI\Utils\get_flag_value( $assoc_args, 'force', false );
+			if ( ! $force_connection && Jetpack::is_active() ) {
+				WP_CLI::error( 'Jetpack is already active; bailing. Run this command with `--force` to bypass this check or disconnect Jetpack before continuing.' );
 			}
-			WP_CLI::error( 'Failed to fetch keys from Jetpack Start: ' . $message );
-		}
 
-		WP_CLI::line( '-- Adding keys to site' );
-		$this->install_keys( $data );
+			WP_CLI::line( '-- Verifying VIP machine user exists (or creating one, if not)' );
```
mjangda: I think if we split off the connection logic into its own method, that would:

- simplify the multisite vs single site handling here (no more hacky `is_int` check)
- make it easier to handle the `WP_CLI::error` cases by returning early from the method.

ethitter:
I need to investigate why Jetpack doesn't get connected on subsites, meaning the command provisions new keys with each run for each subsite. It works if I pass the `--url` flag and treat a subsite as a single-site, so perhaps Jetpack is confused by the `switch_to_blog()` or there's some caching fun at play.

ethitter:
Ah, that was easy. It's the `install_keys()` method, which needs `--url` flags for the commands it runs.

jetpack-start/wp-cli-api.php
```diff
@@ -79,23 +79,35 @@ public function connect( $args, $assoc_args ) {
...
 			foreach ( $sites as $site ) {
 				switch_to_blog( $site );
 
 				WP_CLI::line( sprintf( 'Starting %s (site %d)', home_url( '/' ), $site ) );
 
-				$this->connect_site( $assoc_args );
+				$_success = $this->connect_site( $assoc_args );
+
+				// Just one failure to flip the message from success
+				if ( true === $success && $success !== $_success ) {
```
mjangda: Minor: We can just simplify this as:

```
if ( true !== $_success ) {
    $success = $_success;
}
```

Since the cost of assignment is not that high.

For clarity, I would also rename `$success` to something more verbose like `$did_any_fail` (although, that flips the boolean and would require changes below).

ethitter:
r93558-deploy

PR #630, opened 2017-05-24
Add debug helpers

mjangda:
Borrowed directly from WP.com.

Puts the file in a new `000-debug` directory and my thinking here is that we should move debug bar, query monitor and others into this folder as well.


mjangda:
r93992-deploy

PR #634, opened 2017-05-26
Switch to alt cron if basic auth is enabled

joshbetz:
Related: https://github.com/Automattic/vipd/pull/401

### DIFF
--- a/001-cron.php
+++ b/001-cron.php
@@ -16,7 +16,7 @@
 function wpcom_vip_use_core_cron() {
 	if ( defined( 'WPCOM_VIP_BASIC_AUTH' ) && WPCOM_VIP_BASIC_AUTH ) {
 		define( 'ALTERNATE_WP_CRON', true );
-		return false;
+		return true;
 	}
 	
 	// Bail early for anything that isn't a multisite subsite
--- a/001-cron.php
+++ b/001-cron.php
@@ -14,7 +14,7 @@
  * @return bool
  */
 function wpcom_vip_use_core_cron() {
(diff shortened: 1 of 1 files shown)


001-cron.php
```diff
@@ -14,6 +14,11 @@
  * @return bool
  */
 function wpcom_vip_use_core_cron() {
+	if ( VIP_GO_BASIC_AUTH ) {
```
ethitter: To avoid needing to rebuild all wp-configs straight away, can this include a `defined()`? Otherwise we're headed for notice-palooza.

001-cron.php
```diff
@@ -14,6 +14,11 @@
...
+		define( 'ALTERNATE_WP_CRON', true );
+		return false;
```
ethitter: This should be `true`, as we want to use core cron.

joshbetz:
@ethitter maybe we should only use ALTERNATE_WP_CRON as a workaround for basic auth and regular core cron otherwise?

ethitter:
@joshbetz I'm partial to setting the constant for all Core cron uses just so we reduce the number of variations in how we run cron. I don't think there's much harm, given how limited the use of Core cron is.

mjangda:
ALTERNATE_WP_CRON is pretty janky so I think better to just leave for the exceptional small number of use http auth cases:

> This method is a bit iffy sometimes, which is why it's not the default.

https://wordpress.org/support/topic/scheduled-posts-still-not-working-in-282/#post-1163465

mjangda:
r93440-deploy

PR #637, opened 2017-05-30
Log Jetpack Search queries to Debug Bar Elasticsearch Panel

nickdaugherty:
Adds JP Search queries to the Elasticsearch Panel.

Same as `wpcom_vip_did_elasticsearch_query()`

### DIFF
--- a/vip-jetpack/vip-jetpack.php
+++ b/vip-jetpack/vip-jetpack.php
@@ -32,3 +32,21 @@ if ( true === WPCOM_SANDBOXED ) {
 		wp_die( '😱😱😱 Oh no! Looks like your sandbox is trying to change the version of Jetpack. This is probably not a good idea. As a precaution, we\'re killing this request to prevent this from happening (this === 💥💥💥). Please run `vip stacks update` on your sandbox before doing anything else.', 400 );
 	}, 0 ); // No need to wait till priority 10 since we're going to die anyway
 }
+
+function wpcom_vip_did_jetpack_search_query( $query ) {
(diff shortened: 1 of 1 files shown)


vip-jetpack/vip-jetpack.php
```diff
@@ -32,3 +32,21 @@
...
 }
+
+function wpcom_vip_did_jetpack_search_query( $query ) {
+	if ( ! defined( 'SAVEQUERIES' ) || ! SAVEQUERIES ) {
+		return;
+	}
+
+	global $wp_elasticsearch_queries_log;
+
+	if ( ! $wp_elasticsearch_queries_log ) {
```
mjangda: Should this be `! isset( $wp_elasticsearch_queries_log )`?

nickdaugherty: Ya that would be better - updated and added an `is_array()` check for good measure.

mjangda:
Minor aside (not needed in this PR): I wonder if we should just provide the ability to fire an action which the debug panel and hook into to simplify some of the logic.

So, in `wpcom_vip_did_jetpack_search_query`, we would just do:

```
function wpcom_vip_did_jetpack_search_query( $query ) {
    // ...

    do_action( 'wpcom_vip_elasticsearch_debug_log_query', $query );
}
```

That way, we move the responsibility of the storing logging to the panel

nickdaugherty:
Good idea on the action for the ES panel - that would certainly simplify things and reduce code duplication.

nickdaugherty:
r93563-deploy

PR #639, opened 2017-06-02
Don't block JP version updates when running a different version locally

ethitter:
When running a different version of Jetpack, care of the `WPCOM_VIP_JETPACK_LOCAL` constant, version changes are expected.


mjangda:
We only block the version change for sandboxed requests so I'm not sure I understand what the issue is?

ethitter:
I'll reopen if I encounter this again. I was testing Jetpack 5 when I created this, but I don't recall further details.

PR #642, opened 2017-06-05
Add support for serving admin queries from ES

philipjohn:
Adds two new filters which give plugins/themes the power to have some admin queries served by Elasticsearch.

`wpcom_elasticsearch_enable_in_admin`, when set to true will cause the plugin to run in the admin instead of only the frontend.

When in the admin, `wpcom_elasticsearch_perform_admin_query` provides a mechanism to inspect the query and check if it's suitable for being served by ES (defaults to `false`).

This example would enable admin searches that are searching only published, public posts:

```
add_filter( 'wpcom_elasticsearch_enable_in_admin', '__return_true' );
add_filter( 'wpcom_elasticsearch_perform_admin_query', 'example_wpcom_elasticsearch_perform_admin_query' );
function example_wpcom_elasticsearch_perform_admin_query( $es, $query ) {

	// Explicitly stop attachment queries from using ES.
	if ( in_array( 'attachment', (array) $query->get('post_type') ) ) {
		return false;
	}

	// Only search published posts in ES.
	if ( 'publish' !== $query->get('post_status') ) {
		return false;
	}

	$public_post_types = get_post_types( [
		'public'   => true,
	] );

	// If any of the post types aren't public, don't use ES.
	if ( $query->get('post_type') !== array_intersect( (array) $query->get('post_type'), $public_post_types ) ) {
		return false;
	}

	// We have a query that is only looking for public post types with a
	// published status, so allow it to go through ES.
	return true;

}
```

### DIFF
--- a/shared-plugins/vip-go-elasticsearch/vip-go-elasticsearch.php
+++ b/shared-plugins/vip-go-elasticsearch/vip-go-elasticsearch.php
@@ -177,7 +177,7 @@ class WPCOM_elasticsearch {
 		 * Allow the admin query to be assessed for suitability for being served
 		 * by ES, For example, published posts only.
 		 *
-		 * @var [type]
+		 * @var bool $es Whether (true) or note (false) to run the query through ES.
 		 */
 		if ( is_admin() && false === apply_filters( 'wpcom_elasticsearch_perform_admin_query', false, $query ) ) {
 			return $sql;
--- a/shared-plugins/vip-go-elasticsearch/vip-go-elasticsearch.php
+++ b/shared-plugins/vip-go-elasticsearch/vip-go-elasticsearch.php
@@ -198,7 +198,7 @@ class WPCOM_elasticsearch {
 
 		$should_replace = false;
 
-		// By defauly, we only care about main search queries
+		// By default, we only care about main search queries
 		if ( $query->is_main_query() && $query->is_search() ) {
 			$should_replace = true;
 		}
@@ -209,7 +209,7 @@ class WPCOM_elasticsearch {
 		 * Allow the admin query to be assessed for suitability for being served
 		 * by ES, For example, published posts only.
 		 *
-		 * @var bool $es Whether (true) or note (false) to run the query through ES.
+		 * @var bool $es Whether (true) or not (false) to run the query through ES.
 		 */
 		$should_replace = apply_filters( 'wpcom_elasticsearch_should_replace_query', $should_replace, $query );
 
--- a/shared-plugins/vip-go-elasticsearch/vip-go-elasticsearch.php
+++ b/shared-plugins/vip-go-elasticsearch/vip-go-elasticsearch.php
@@ -171,6 +171,18 @@ class WPCOM_elasticsearch {
 		if ( ! $query->is_main_query() || ! $query->is_search() )
 			return $sql;
 
+		/**
+		 * Whether to run this admin query through ES.
+		 *
+		 * Allow the admin query to be assessed for suitability for being served
+		 * by ES, For example, published posts only.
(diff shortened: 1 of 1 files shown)


mjangda:
Let's make sure to update the wpcom version of this plugin too so the two don't get out of sync.

shared-plugins/vip-go-elasticsearch/vip-go-elasticsearch.php
```diff
@@ -163,6 +171,18 @@ public function filter__posts_request( $sql, $query ) {
 		if ( ! $query->is_main_query() || ! $query->is_search() )
 			return $sql;
 
+		/**
+		 * Whether to run this admin query through ES.
+		 *
+		 * Allow the admin query to be assessed for suitability for being served
+		 * by ES, For example, published posts only.
+		 *
+		 * @var [type]
```
nickdaugherty: Can you finish out the docblock here?

philipjohn: Ah yes sorry, I noticed that and forgot, thanks for spotting it.

philipjohn: Done in [19fe4cf]

mjangda:
@nickdaugherty noted that the `is_main_query` check may prevent this change from actually being useful; it may work for some queries but not others.

The better option would be to switch over to using https://github.com/alleyinteractive/es-wp-query instead and/or moving forward with the work on authenticated search requests (#644).

philipjohn:
> it may work for some queries but not others

@mjangda @nickdaugherty Yep, that's the intention. To get full ES support in the admin, we need the es-admin plugin. However, there are some queries (such as searching for content in the customiser, as some sites do) that could feasible be served by ES and would have a positive impact on performance before we're able to get full ES admin support.

david-binda:
@philipjohn I believe that @mjangda is referring to https://github.com/Automattic/vip-go-mu-plugins/pull/642/files#diff-88b217cf2de4d4f0af835bf8d95a2952R171 which excludes any WP_Query call which is not the primary one. So even the expected patch described above won't actually turn arbitrary `new WP_Query( array( 'post_type' => 'post', 'post_status' => 'publish' ) )` into a ES query, as it would not fulfil the `is_main_query` condition :-/

mjangda:
@philipjohn I tweaked the logic for this a bit.

`wpcom_elasticsearch_enable_in_admin` stays as-is but `wpcom_elasticsearch_perform_admin_query` has been renamed to `wpcom_elasticsearch_should_replace_query`.

The change  give us a bit more granular control and makes it so we can easily enable ES for any query (e.g. main post list search query in the admin). We can also disable ES for queries when we detect things that we know can't be queried via ES as well.

Updated usage example:

```
add_filter( 'wpcom_elasticsearch_enable_in_admin', '__return_true' );

add_filter( 'wpcom_elasticsearch_should_replace_query', function( $should_replace, $query ) {
    if ( ! $should_replace && ! $query->is_main_query() && 'publish' !== $query->get( 'post_status' ) ) {
        $should_replace = true;
    }
    return $should_replace;
} );

```

mjangda:
@nickdaugherty @philipjohn @david-binda I updated this PR to automatically authenticate all admin queries. Can you please take a look?

philipjohn:
Nice! I fixed a couple of typos.

mjangda:
This also requires translation of `post_status` into queries and maybe a blacklist of things that it doesn't do well that should rely on es-wp-query (or otherwise instead).

mjangda:
Going to close; to be replaced by VIP Search.

PR #644, opened 2017-06-06
Allow authenticated requests to ES

mjangda:
Allow setting an arg for `es_api_search_index` (`authenticated_request => true`), which would make an authenticated request to the search API using Jetpack internal auth.

Note that calling methods (e.g. `vip-go-elasticsearch`) need to filter through the returned results list to strip out any content that the user should not have access to.

### DIFF
--- a/vip-helpers/vip-elasticsearch.php
+++ b/vip-helpers/vip-elasticsearch.php
@@ -62,18 +62,36 @@ function es_api_search_index( $args ) {
 
     $args['blog_id'] = absint( $args['blog_id'] );
 
-    $service_url = 'https://public-api.wordpress.com/rest/v1/sites/' . $args['blog_id'] . '/search';
+    $endpoint = sprintf( '/sites/%s/search', $args['blog_id'] );
+    $service_url = 'https://public-api.wordpress.com/rest/v1' . $endpoint;
+
+    $do_authenticated_request = false;
+    if ( class_exists( 'Jetpack_Client' )
+            && isset( $args['authenticated_request'] )
+	    && true === $args['authenticated_request'] ) {
+        $do_authenticated_request = true;
+    }
 
     unset( $args['blog_id'] );
+    unset( $args['authenticated_request'] );
 
-    $start_time = microtime( true );
-
-    $request = wp_remote_post( $service_url, array(
+    $request_args = array(
         'headers' => array(
             'Content-Type' => 'application/json',
         ),
-        'body' => json_encode( $args ),
-    ) );
+    );
+    $request_body = json_encode( $args );
+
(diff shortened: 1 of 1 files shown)


vip-helpers/vip-elasticsearch.php
```diff
@@ -62,18 +62,36 @@ function es_api_search_index( $args ) {
 
     $args['blog_id'] = absint( $args['blog_id'] );
 
-    $service_url = 'https://public-api.wordpress.com/rest/v1/sites/' . $args['blog_id'] . '/search';
+    $endpoint = sprintf( '/sites/%s/search', $args['blog_id'] );
+    $service_url = 'https://public-api.wordpress.com/rest/v1' . $endpoint;
+
+    $do_authenticated_request = false;
+    if ( class_exists( 'Jetpack_Client' )
+            && isset( $args['authenticated_request'] )
```
mjangda: Maybe worthwhile to add a new filter for args (to allow setting the `authenticated_request` arg) and a filter for the results (and stripping out posts that don't match user permissions) that can be easily added/removed.

The use case would be when using something like `es-wp-query` (or other contexts) where we may not have the ability to manipulate the args passed into or returned from `es_api_search_index`.

mjangda: Oh, never mind. es-wp-query has the `es_posts_request` filter which we can use (https://github.com/alleyinteractive/es-wp-query/blob/master/class-es-wp-query-wrapper.php#L1033)

mjangda:
r93757-deploy

vip-helpers/vip-elasticsearch.php
nicoladj77: @mjangda for `es-wp-query` why can't I do simply

```
class ES_Query_Adapter extends \ES_WP_Query_Wrapper {

	protected function query_es( $es_args ) {
		$start_time = microtime( true );

		if ( isset( $es_args['query'], $es_args['query']['multi_match'] ) ) {
			$es_args['query']['multi_match']['fields'][] = 'file.content';
		}

		$es_args['authenticated_request'] = true;

		return es_api_search_index($es_args);
	}
}
```

shouldn't that work?

mjangda: You'll want to load and use the vip adapter (https://github.com/alleyinteractive/es-wp-query/blob/master/adapters/wpcom-vip.php) to make sure that the WP_Query args => ES fields are mapped correctly (and any other custom logic it includes).

You can obviously copy the adapter and modify it as needed but will need to keep it in sync with any future updates to the vip adapter, so it would be better just to use the filter to flag it as an `authenticated_request`.

PR #648, opened 2017-06-09
Add support for `srcset` image

### LINKED ISSUE #587: Jetpack Photon: Add responsive images support
As of 4.4, WordPress can automatically generate and add the necessary markup for `srcset` support (see [wp_make_content_images_responsive](https://developer.wordpress.org/reference/functions/wp_make_content_images_responsive/)). However, by default that relies on image sizes being defined in `_wp_attachment_metadata`, which don't exist on Go (since all images are dynamically generated).

We can use a combination of the `wp_calculate_image_srcset_meta` filter + Jetpack Photon to fake our way through and add the the necessary pieces for this to work.

Rough idea of how to make this work:

Hook in to `wp_calculate_image_srcset_meta` and add the `sizes` we want included in our `srcset`. The resulting `metadata` array + `sizes` should look something like:

```
array (
    'width' => 1500,
    'height' => 977,
    'file' => '2017/01/image.jpg',
    'sizes' =>
    array (
      'thumbnail' =>
      array (
        'file' => 'image-150x98.jpg',
        'width' => 150,
        'height' => 98,
        'mime-type' => 'image/jpeg',
      ),
      'medium' =>
      array (
        'file' => 'image-300x195.jpg',
        'width' => 300,
        'height' => 195,
        'mime-type' => 'image/jpeg',
      ),
   ),
)
```

Note that the `file` for each size doesn't have to physically exist but for Jetpack's Photon filters we need to include it in the array and follow the standard naming convention. We can either reuse or copy [`WP_Image_Editor::generate_filename`](https://developer.wordpress.org/reference/classes/wp_image_editor/generate_filename/) for this.

We also may want a custom filter for VIPs to adjust the sizes included in the srcset.

pkevan:
Mainly copied from WP.com implementation

- Allows overriding of thumbnails through `vipgo_srcset_image_thumbnails`.
- Blocks only creating `srcset` from images belonging to current site.

Fixes #587


ethitter:
Thanks for starting on this!

Two thoughts before looking too closely:

1. The `A8C_Files_Image_Srcset_Meta` class should be in its own file, in an `a8c-files` subdirectory or something similarly named. Multiple classes in a single file bothers me, particularly when both classes are significant in size. The `A8C_Files` should also move, with the file in the root serving as a stub/loader.
2. The `A8C_Files_Image_Srcset_Meta` class should only be included when a constant is defined, or conversely, the presence of a constant should prevent it from loading. Several clients have implemented srcset on their own, and we need an orderly way of launching this that doesn't also require simultaneous pushes to multiple client repos. Because we're in mu-plugins, and #611 is still open, removing the filter isn't a feasible approach to this.

I'll take a further look at the class shortly.

a8c-files.php
```diff
@@ -776,6 +776,232 @@ function a8c_files_init() {
...
+	 * @param int    $attachment_id The image attachment ID or 0 if not supplied.
+	 * @return array                The normalized image meta data or empty $image_meta['sizes'] on error.
+	 */
+	public static function generate_image_meta( $image_meta, $size_array, $image_src, $attachment_id ) {
+		// Reset $image_meta['sizes'] if not empty. Need to build it again matching the current img src.
+		$image_meta['sizes'] = array();
+
+		// Bail out early if the image is not hosted on the site, missing image meta or original image is too small.
+		// This will disable the generation of srcset.
+		if ( strpos( $image_src, home_url() ) === false ||
```
ethitter: This is rather fragile, and needs to be rethought. For starters, we should be comparing against just the hosts, using `parse_url()`. Otherwise, `https://www.example.com` and `https://www.example.com.bad.site` would pass, while a site with inconsistent HTTPS won't be parsed. Additionally, this doesn't account for differences between `siteurl` and `home`, and we have some sites with split settings. Either a filter, or checking both `home_url()` and `site_url()`, is needed.

a8c-files.php
```diff
@@ -776,6 +776,232 @@ function a8c_files_init() {
...
+			empty( $image_meta['height'] ) ||
+			( $image_meta['width'] < 100 && $image_meta['height'] < 100 ) ) {
+
+			return $image_meta;
+		}
+
+		$image_basename = wp_basename( $image_meta['file'] );
+
+		// Return early if filename mismatch
+		if ( strpos( wp_basename( $image_src ), $image_basename ) === false ) {
```
ethitter: To make it easier to follow what's being compared with what, the second `wp_basename()` call should also be set in a variable. I'm also curious why we don't just compare the two results. Why is the `strpos()` necessary?

a8c-files.php
```diff
@@ -776,6 +776,232 @@ function a8c_files_init() {
...
+				if ( $photon_args['crop'] === '1' ) {
+					$crop = true;
+				} else {
+					// No support for Photon cropping using pixel values or percents, yet.
+					return $image_meta;
+				}
+			}
+
+			if ( ! empty( $photon_args['resize'] ) ) {
+				// TODO: maybe see if there is a discrepancy between the width/height from the img tag and the values in 'resize'
```
ethitter: Still worth considering? Seems unlikely if this has been live on dotcom for some time, in which case, we should remove this note.

a8c-files.php
```diff
@@ -776,6 +776,232 @@ function a8c_files_init() {
...
+		if ( ! empty( $resize ) || $crop ) {
+			list( $full_width, $full_height ) = wp_expand_dimensions( $image_width, $image_height, $full_width, $full_height );
+		}
+
+		$width_2x = round( $image_width * 2 );
+		$height_2x = round( $image_height * 2 );
+
+		// Add the current size and 2x current size. That should be the largest available size.
+		// The current "retina" browsers don't show images larger than 2x even on displays with 3.5x or 4.0x pixel density.
+		// Prepend wpcom-current. Fixes iOS8 bug.
```
ethitter: Is there more to this? I don't see any other references to this, so I wonder how and what it fixes.

a8c-files.php
```diff
@@ -776,6 +776,232 @@ function a8c_files_init() {
...
+			$sizes['medium']['width'] = 300;
+			$sizes['medium']['height'] = 300;
+		}
+
+		if ( ! $sizes['large']['width'] && ! $sizes['large']['height'] ) {
+			$sizes['large']['width'] = 1024;
+			$sizes['large']['height'] = 0;
+		}
+
+		return apply_filters( 'vipgo_srcset_image_thumbnails', $sizes );
```
ethitter: Rather than namespace this just to VIP Go, let's use the same filter tag as dotcom. That way, clients migrating from from dotcom to Go get the support automatically.

pkevan:
@ethitter most of this has been copied from the WP.com implementation, i'll comment inline on the above

a8c-files.php
pkevan: This filter doesn't exist on dotcom.

a8c-files.php
pkevan: Not sure on this

a8c-files.php
pkevan: Probably best to just remove it, ideally we keep aligned with the dotcom version going forward.

a8c-files.php
pkevan: Yes, this part I was least sure about, on the dotcom implementation they have a dedicated files.wp domain check.

ethitter:
> @ethitter most of this has been copied from the WP.com implementation, i'll comment inline on the above

Perfect opportunity to iterate and/or improve! :)

a8c-files.php
ethitter: Okay. Let's prefix this with `wpcom_vip_` instead of `vipgo_` anyway. We don't use `vipgo_` as a prefix on any filters currently, so consistency is a plus.

mjangda:
Thanks for working on this @pkevan! I haven't looked too closely at the code, but some high-level thoughts:

1) +1 to Erick's point about using this as an opportunity to improve the code. For example, it would be nice break up the class into smaller, more focused methods and add tests where we can.

2) I'm a bit worried about adding "yet-another-constant". We already too much selective functionality locked behind constants and this will just add to the complexity. A better approach here would be how we handled the roll-out on WordPress.com:
a) default to on with an easy opt-out via filter (bonus if we can detect if the site already handles responsive images on their own and bail);
b) do a Lobby Post with a deployment deadline;
c) deploy on deadline.

3) Will this work locally? If not, anything we can do to enable local testing? Or should we disable the functionality in local environments?

a8c-files.php
david-binda: It's related to this ticket: https://core.trac.wordpress.org/ticket/35030 . Similar patch to the one in here can be found in core itself.

ethitter:
There are pieces of this missing from the WP.com implementation, which means it doesn't currently work. Most notably, we're missing `WPCOM_Responsive_Images::make_content_images_responsive()`, which is needed to replace `wp_make_content_images_responsive()` because we don't store image sizes.

ivankristianto:
Do we have an ETA for this PR merged and released?

mjangda:
Closing; continued in #860

PR #649, opened 2017-06-12
Don't disable Core updates outside of our environments

ethitter:
We should only block upgrades in our environments, where we know they can't succeed. For local environments, for example, we shouldn't interfere. See poqVs-hAR-P2.


mjangda:
Dupe of #650

ethitter:
See #650.

PR #650, opened 2017-06-12
Allow local core upgrades

mjangda:
Upgrade cap restrictions should only apply to production environments where we can't actually trigger upgrades due to the read-only file system.


mjangda:
r93988-deploy

PR #651, opened 2017-06-13
Add "validate-attachments" CLI command.

emrikol:
This command will iterate over attachments in the database and create a CSV of any missing (or other non-200 status) files.  This can be useful to determine if a media import has went wrong.


mjangda:
r94092-deploy

PR #652, opened 2017-06-13
Allow Cron Control to be disabled

ethitter:
Restore's the `wpcom_vip_go_use_core_cron` filter removed in fe898a7, allowing `wpcom_vip_use_core_cron()` to be changed at runtime.

Also, before we try overriding the REST API's authentication, we should be sure the class we need to reference is present.


mjangda:
How would someone use this filter?

ethitter:
I'm going to switch it to a constant instead. Originally, I thought this would be applicable only in local development, where someone could create a local helper to hook `__return_true`, but that seems too opaque. A constant avoids any ambiguity.

ethitter:
Is it worth adding a `WPCOM_IS_VIP_ENV` check, to prevent this from being applied in production?

mjangda:
Yeah, I think that's probably the bigger issue here; the fact the local environments currently load cron-control which mean cron doesn't work at all.

ethitter:
I flipped the logic around so that when `WPCOM_IS_VIP_ENV` is false, `WPCOM_VIP_LOAD_CRON_CONTROL_LOCALLY` must be defined true to load Cron Control.

mjangda:
r94217-deploy

PR #653, opened 2017-06-14
Graceful continue for dbdelta for network flag

mjangda:
If we pass the network flag on a single site, show a warning but continue.


mjangda:
r93991-deploy

PR #643, opened 2017-06-06
solved symbolic links issue in windows os

emgk:
Fixes #641

PR #654, opened 2017-06-14
Drop symlinks

mjangda:
For wider compat across environments and simpler deploys.

Previously #643


mjangda:
r93989-deploy
r93990-deploy

PR #657, opened 2017-06-14
Add more strict checking of API results before using as an array

### LINKED ISSUE #656: Invalid argument supplied for foreach() in ... (VIP Go Elasticsearch)
```PHP Warning:  Invalid argument supplied for foreach() in /var/www/wp-content/mu-plugins/shared-plugins/vip-go-elasticsearch/vip-go-elasticsearch.php on line 140```

We need to add some extra checking before using `$this->search_result['results']['hits']` as an array.

nickdaugherty:
Fixes #656


nickdaugherty:
r94030-deploy

PR #659, opened 2017-06-16
Adds update-user-ids CLI command

emrikol:
While this "works" it just feels wrong--so I'm open to any suggestions :)


mjangda:
This looks pretty scary :) I'm sure we really need it either? Might be good to have it in a separate migration scripts repo that we can pull and reference if needed?

emrikol:
Closing this out, moving it somewhere far away from production :)

PR #661, opened 2017-06-19
Adding import-user-meta CLI command

emrikol:
This CLI command imports user metadata from a CSV file, useful when migrating [user attributes](https://vip.wordpress.com/documentation/user_meta-vs-user_attributes/) from WordPress.com.


wp-cli/vip-migrations.php
```diff
@@ -147,6 +147,54 @@ public function validate_attachments( $args, $assoc_args ) {
 		WP_CLI\Utils\write_csv( $file_descriptor, $output );
 		fclose( $file_descriptor );
 	}
+
+	/**
+	 * Import user meta attributes from a CSV file.
+	 *
+	 * @subcommand import-user-meta
+	 * @synopsis <file> [--user_key=<userlogin>]
```
mjangda: We need more docs here.

What is `user_key`? What structure should the csv have?

wp-cli/vip-migrations.php
```diff
@@ -147,6 +147,54 @@ public function validate_attachments( $args, $assoc_args ) {
...
+	function import_user_meta( $args, $assoc_args ) {
+		$filename = $args[0];
+		$user_key = $assoc_args['user_key'] ?? 'user_login';
+
+		if ( ! file_exists( $filename ) ) {
+			WP_CLI::error( sprintf( 'Missing file: %s', $filename ) );
+		}
+
+		foreach ( new \WP_CLI\Iterators\CSV( $filename ) as $user_data ) {
+			$user_data = array_values( $user_data ); // Strip useless array keys.
```
mjangda: What if the first line in the CSV is column headers? Is that something we need to worry about?

wp-cli/vip-migrations.php
```diff
@@ -147,6 +147,54 @@ public function validate_attachments( $args, $assoc_args ) {
...
+			$user_value = $user_data[0];
+			$meta_key = $user_data[1];
+			$meta_value = json_decode( $user_data[2] );
```
mjangda: Minor: this can be simplified as:

```
list( $user_value, $meta_key, $meta_value ) = $user_data;
$meta_value = json_decode( $meta_value );
```

wp-cli/vip-migrations.php
```diff
@@ -147,6 +147,54 @@ public function validate_attachments( $args, $assoc_args ) {
...
+					$user = get_user_by( 'slug', $user_value );
+					break;
+				case 'user_email':
+					$user = get_user_by( 'email', $user_value );
+					break;
+				case 'user_login':
+					$user = get_user_by( 'login', $user_value );
+					break;
+				default:
+					WP_CLI::error( 'Error getting user ' . $user_value );
```
mjangda: Should we error out or throw a warning and move on to the next one?

wp-cli/vip-migrations.php
```diff
@@ -147,6 +147,54 @@ public function validate_attachments( $args, $assoc_args ) {
...
+			}
+
+			$add_meta = update_user_meta( $user->ID, $meta_key, $meta_value );
```
mjangda: We should probably add a dry run option.

wp-cli/vip-migrations.php
emrikol: I'm happy to add a switch for headers, but my plan was to have this work specifically with CSVs created from this WP-CLI command on Phabricator: D6019

Do you forsee a situation where we might be importing user meta from other places?

mjangda: Makes sense. Should we update the docs to mention that this is meant to be used with a CSV generated from the WP.com CLI command?

mjangda: 7e8ef6d

mjangda:
r95776-deploy

PR #664, opened 2017-06-20
Allow debug helpers to be used locally

### LINKED ISSUE #655: Allow debug helper to be used locally
The debug helper methods added in #630 are pretty useful and would be nice to allow them to be used for local development as well. Right now, they are sandbox only.

mjangda:
Fixes #655


mjangda:
r94463-deploy

PR #665, opened 2017-06-23
Jetpack: Disable Sync Listener for some requests.

philipjohn:
Sync Listener looks for events that need to be added to the sync queue. On many requests, such as frontend views, we wouldn't expect there to be any DB writes so there should be nothing for Jetpack to listen for.

This change stops Listener from loading on those requests.

See pp1HpG7-45O-p2

### DIFF
--- a/vip-jetpack/vip-jetpack.php
+++ b/vip-jetpack/vip-jetpack.php
@@ -50,3 +50,30 @@ function wpcom_vip_did_jetpack_search_query( $query ) {
 }
 
 add_action( 'did_jetpack_search_query', 'wpcom_vip_did_jetpack_search_query' );
+
+/**
+ * Decide when Jetpack's Sync Listener should be loaded.
+ *
+ * Sync Listener looks for events that need to be added to the sync queue. On
+ * many requests, such as frontend views, we wouldn't expect there to be any DB
+ * writes so there should be nothing for Jetpack to listen for.
+ *
+ * @param  bool $should_load Current value.
+ * @return bool              Whether (true) or not (false) Listener should load.
+ */
+function wpcom_vip_jetpack_sync_listener_should_load( $should_load ) {
+
+	// Don't run listener when we're on the frontend, not running cron and dealing
+	// with a GET request as opposed to POST/PUT etc.
+	if (
+		! is_admin() &&
+		! DOING_CRON &&
+		( isset( $_SERVER["REQUEST_METHOD"] ) && 'GET' === $_SERVER['REQUEST_METHOD'] )
+	) {
+		$should_load = false;
+	}
+
+	return $should_load;
+
+}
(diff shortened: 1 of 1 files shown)


philipjohn:
I'm not sure if my choice of conditional is really that good. I'm conscious that REST API requests could be making changes, so we don't want to exclude those. Cron actions definitely do make changes sometimes. What else?

joshbetz:
> See pp1HpG7-45O-p2

My enigma machine broke this morning. Was this supposed to be a link? 😬

joshbetz:
> I'm not sure if my choice of conditional is really that good. I'm conscious that REST API requests could be making changes, so we don't want to exclude those

In general GET requests shouldn't write to the database though.

vip-jetpack/vip-jetpack.php
```diff
@@ -50,3 +50,30 @@ function wpcom_vip_did_jetpack_search_query( $query ) {
...
+ */
+function wpcom_vip_jetpack_sync_listener_should_load( $should_load ) {
+
+	// Don't run listener when we're on the frontend, not running cron and dealing
+	// with a GET request as opposed to POST/PUT etc.
+	if (
+		! is_admin() &&
+		! DOING_CRON &&
+		( isset( $_SERVER["REQUEST_METHOD"] ) && 'GET' === $_SERVER['REQUEST_METHOD'] )
+	) {
```
mjangda: I would break this up with early returns so it's easier to read:

```
if ( is_admin() ) {
    return $should_load;
}

if ( DOING_CRON ) {
    return $should_load;
}

if ( isset( $_SERVER["REQUEST_METHOD"] ) && 'GET' === $_SERVER['REQUEST_METHOD'] ) {
    $should_load = false;
}

return $should_load;
```

vip-jetpack/vip-jetpack.php
```diff
@@ -50,3 +50,30 @@ function wpcom_vip_did_jetpack_search_query( $query ) {
...
+ * @param  bool $should_load Current value.
+ * @return bool              Whether (true) or not (false) Listener should load.
+ */
+function wpcom_vip_jetpack_sync_listener_should_load( $should_load ) {
+
+	// Don't run listener when we're on the frontend, not running cron and dealing
+	// with a GET request as opposed to POST/PUT etc.
+	if (
+		! is_admin() &&
+		! DOING_CRON &&
```
simonwheatley: Do we need this to be:

``` php
if ( ! defined('DOING_CRON') || ! DOING_CRON ) {
	return $should_load;
}
```

I don't think `DOING_CRON` is defined if it's false, just if it's true.

vip-jetpack/vip-jetpack.php
simonwheatley: +1

mjangda:
ping @philipjohn and @simonwheatley

philipjohn:
Thanks for the changes, this looks great!

mjangda:
r98200-deploy

PR #666, opened 2017-06-23
Introduce new plugins endpoint for querying all types of plugins

scottsweb:
This creates a new private endpoint for determining all active plugins on a site grouped by type:

screenshot-23-06-2017-18 45 10

Also added a helper function to grab client mu plugin information

### DIFF
--- a/rest-api/vip-endpoints.php
+++ b/rest-api/vip-endpoints.php
@@ -148,11 +148,8 @@ class WPCOM_VIP_REST_API_Endpoints {
 	 * Get all the plugins
 	 */
 	public function get_all_plugins() {
+		global $vip_loaded_plugins;
 		$all_plugins = array();
-		$standard_plugins = array();
-		$shared_plugins = array();
-		$mu_plugins = array();
-		$client_mu_plugins = array();
 
 		if ( ! function_exists( 'get_plugins' ) ) {
 			require_once ABSPATH . 'wp-admin/includes/plugin.php';
@@ -162,16 +159,42 @@ class WPCOM_VIP_REST_API_Endpoints {
 		$standard_plugins = get_plugins();
 		$tmp_plugins = array();
 		foreach ( $standard_plugins as $key => $plugin ) {
+			$vip_plugin_slug = 'plugins/' . dirname( $key );
 			if ( is_plugin_active( $key ) ) {
 				$tmp_plugins[ $key ] = array(
 					'name' => $plugin['Name'],
 					'version' => $plugin['Version'],
 					'description' => $plugin['Description'],
 					'type' => 'standard',
+					'active' => true,
+				);
+			} elseif ( ! in_array( $vip_plugin_slug, $vip_loaded_plugins, true ) ) {
+				$tmp_plugins[ $key ] = array(
+					'name' => $plugin['Name'],
+					'version' => $plugin['Version'],
+					'description' => $plugin['Description'],
+					'type' => 'standard',
+					'active' => false,
 				);
 			}
 		}
 		$all_plugins['standard'] = $tmp_plugins;
+
+		// array of all code activated standard plugins
+		$tmp_plugins = array();
+		foreach ( $standard_plugins as $key => $plugin ) {
+			$vip_plugin_slug = 'plugins/' . dirname( $key );
+			if ( in_array( $vip_plugin_slug, $vip_loaded_plugins, true ) ) {
+				$tmp_plugins[ $key ] = array(
+					'name' => $plugin['Name'],
+					'version' => $plugin['Version'],
+					'description' => $plugin['Description'],
+					'type' => 'standard-code',
+					'active' => true,
+				);
+			}
+		}
+		$all_plugins['standard-code'] = $tmp_plugins;
 
 		// array of all mu plugins
 		$mu_plugins = get_mu_plugins();
--- a/rest-api/vip-endpoints.php
+++ b/rest-api/vip-endpoints.php
@@ -133,9 +133,21 @@ class WPCOM_VIP_REST_API_Endpoints {
 	}
 
 	/**
-	 * Build list of active plugins on site ()
+	 * Endpoint to return active plugins on site
 	 */
 	public function list_plugins() {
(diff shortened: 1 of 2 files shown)


z-client-mu-plugins.php
```diff
@@ -47,4 +47,51 @@ function wpcom_vip_load_client_mu_plugins() {
 	}
 }
 
+function get_client_mu_plugins() {
```
mjangda: Function should be prefixed with `wpcom_vip_`

mjangda: There's a bunch of shared code between this and the function where we actually load the plugins. I think we can split this off into two functions and re-use some of the logic:

1) `wpcom_vip_get_client_mu_plugins()` which would get and return the plugins list, but not load plugin data.
2) `wpcom_vip_get_client_mu_plugins_data()` which would call `wpcom_vip_get_client_mu_plugins()` and then loop through to load plugin data and return.

rest-api/vip-endpoints.php
```diff
@@ -123,6 +131,94 @@ public function list_sites() {
...
+
+		// array of all shared plugins (activated via code and via UI)
+		if ( class_exists( 'WPCOM_VIP_Plugins_UI' ) ) {
+			$tmp_ui_plugins = array();
+			$tmp_code_plugins = array();
+			$vip_plugins = WPCOM_VIP_Plugins_UI::instance();
+			$shared_plugins = $vip_plugins->get_shared_plugins();
+
+			foreach ( $vip_plugins->get_shared_plugins() as $key => $plugin ) {
+				if ( $t = $vip_plugins->is_plugin_active( basename( dirname( $key ) ) ) ) {
```
mjangda: We should use a more descriptive variable name than `$t`. It's not totally obvious what it represents.

mjangda: We should try to avoid assignment of computation inside conditionals:

```
$t = $vip_plugins->is_plugin_active( basename( dirname( $key ) ) )
if ( $t ) {
    ...
}
```

rest-api/vip-endpoints.php
```diff
@@ -123,6 +131,94 @@ public function list_sites() {
...
+							'type' => 'vip-shared-ui',
+						);
+					}
+				}
+			}
+			$all_plugins['vip-shared-code'] = $tmp_code_plugins;
+			$all_plugins['vip-shared-ui'] = $tmp_ui_plugins;
+		}
+
+		return new WP_REST_Response( array( 'site' => get_home_url(), 'plugins' => $all_plugins ) );
```
mjangda: Why do we need to include the URL in the response? Making a call to the API would require knowing it already.

rest-api/vip-endpoints.php
```diff
@@ -123,6 +131,94 @@ public function list_sites() {
 
 		return new WP_REST_Response( $sites );
 	}
+
+	/**
+	 * Build list of active plugins on site ()
+	 */
+	public function list_plugins() {
```
mjangda: As a best practice, it's good to keep API endpoint callbacks lean and without much logic, which means that they should call other functions to collect the data they need. The endpoint callback could be something as simple as:

```
function list_plugins() {
    plugins = get_all_the_plugins();
    if ( is_error( plugins ) ) {
        return error
    }

    return wp_rest_response
}
```

The nice thing about this approach is we can then re-use the `get_all_the_plugins` function in other contexts, if needed (e.g. CLI).

rest-api/vip-endpoints.php
```diff
@@ -123,6 +131,94 @@ public function list_sites() {
...
+		if ( ! function_exists( 'get_plugins' ) ) {
+			require_once ABSPATH . 'wp-admin/includes/plugin.php';
+		}
+
+		// array of all standard plugins
+		$standard_plugins = get_plugins();
+		$tmp_plugins = array();
+		foreach ( $standard_plugins as $key => $plugin ) {
+			if ( is_plugin_active( $key ) ) {
+				$tmp_plugins[ $key ] = array(
```
mjangda: We should track active/inactive state for standard plugins. Also, do we need to differentiate standard plugins that may be UI- vs code-activated?

mjangda: Actually, another plugin type I realized that we need to track are plugins inside themes. Ideally though, we should stop supporting those on Go.

scottsweb: Added support for code activated standard plugins and also added inactive plugins to the output as well. All plugins have an `active` boolean too. 

It doesn't look like we are allowing theme activated plugins on Go already:

https://github.com/Automattic/vip-go-mu-plugins/blob/900fb80d9dbe4934caf014b712ca022fbbb3395e/vip-helpers/vip-utils.php#L993-L998

scottsweb:
@mjangda ready for another pass. See https://github.com/Automattic/vip-go-mu-plugins/pull/666#discussion_r125850631 for some notes

z-client-mu-plugins.php
```diff
@@ -93,5 +103,3 @@ function wpcom_vip_get_client_mu_plugins() {
 
 	return $wp_plugins;
 }
-
-wpcom_vip_load_client_mu_plugins();
```
mjangda: Was this accidentally removed?

rest-api/vip-endpoints.php
```diff
@@ -123,6 +131,133 @@ public function list_sites() {
 
 		return new WP_REST_Response( $sites );
 	}
+
+	/**
+	 * Endpoint to return active plugins on site
+	 */
+	public function list_plugins() {
+		$plugins = $this->get_all_plugins();
+		if ( is_wp_error( $plugins ) ) {
```
scottsweb: I was thinking the same thing :+1:

z-client-mu-plugins.php
```diff
@@ -47,4 +47,59 @@ function wpcom_vip_load_client_mu_plugins() {
 	}
 }
 
-wpcom_vip_load_client_mu_plugins();
```
scottsweb: Apologies, not sure how that happened.

PR #669, opened 2017-06-29
Add mail tracking header to outbound emails

mjangda:
This will allow us to validate and track where emails are coming from.


mjangda:
r95053-deploy

PR #672, opened 2017-06-29
Introduce a TTL Manager class

### LINKED ISSUE #658: JSON API: set a default TTL for all unauthenticated API requests
Should 1 minute; might be nice to be filterable.

mjangda:
Which sets a default TTL of 60s for all unauthenticated, read requests to REST API endpoints instead of the default 30 mins.

```
curl 'http://vagrant.local/?rest_route=/wp/v2/posts' -v -s > /dev/null
...
< Cache-Control: max-age=60
...
```

* [ ] Amend client facing docs here to document filter: https://vip.wordpress.com/documentation/vip-go/controlling-vip-go-page-cache/
* [ ] Amend client facing docs here to document filter: https://vip.wordpress.com/documentation/vip-go/vip-go-and-the-wordpress-rest-api/#caching

Fixes #658

### DIFF
--- a/tests/test-cache-ttl-manager-rest.php
+++ b/tests/test-cache-ttl-manager-rest.php
@@ -1,0 +1,119 @@
+<?php
+
+namespace Automattic\VIP\Cache;
+
+class TTL_Manager__REST_API__Test extends \WP_Test_REST_TestCase {
+	public function setUp() {
+		parent::setUp();
+
+		global $wp_rest_server;
+		$this->server = $wp_rest_server = new \WP_REST_Server;
+		do_action( 'rest_api_init' );
+
+		register_rest_route( 'tests/v1', '/endpoint', [
+			'methods' => [ 'GET', 'HEAD', 'POST', 'PUT', 'DELETE' ],
+			'callback' => '__return_null',
+		] );
+	}
+
+	public function tearDown() {
+		parent::tearDown();
+
+		global $wp_rest_server;
+		$wp_rest_server = null;
+	}
+
+	protected function dispatch_request( $method ) {
+		$request = new \WP_REST_Request( $method, '/tests/v1/endpoint' );
+		$response = $this->server->dispatch( $request );
+		$dispatched_response = apply_filters( 'rest_post_dispatch', rest_ensure_response( $response ), $this->server, $request );
+
+		return $dispatched_response;
+	}
+
+	public function get_rest_read_methods() {
+		return [
+			[ 'GET' ],
+			[ 'HEAD' ],
+		];
+	}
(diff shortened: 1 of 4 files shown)


joshbetz:
This looks pretty good. What do you think about getting rid of the singleton? With all static methods, having the class doesn't seem completely necessary.

mjangda:
I already removed the singleton so just wanted to confirm that your suggestion is to get rid of the class completely and just have functions within the namespace instead?

vip-cache-manager/ttl-manager.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+/*
+Plugin name: TTL Manager
+Description: Sets sane and reasonable cache TTLs for site responses.
```
simonwheatley: Perhaps more neutrally:

```
Description: Sets cache TTLs for site responses, see https://vip.wordpress.com/documentation/vip-go/controlling-vip-go-page-cache/
```

Incidentally we need to document this change in the client facing docs.

simonwheatley:
Added some to do items to the PR description above.

tests/test-cache-ttl-manager-rest.php
```diff
@@ -0,0 +1,119 @@
...
+
+		$response_headers = $response->get_headers();
+
+		$this->assertArrayNotHasKey( 'Cache-Control', $response_headers );
+	}
+
+	public function test__skip_ttl_for_error_responses() {
+		$request = new \WP_REST_Request( 'GET', '/tests/v1/this-does-not-exist' );
+		$response = $this->server->dispatch( $request );
+		$dispatched_response = apply_filters( 'rest_post_dispatch', rest_ensure_response( $response ), $this->server, $request );
```
simonwheatley: This filter needs inline docs

vip-cache-manager/ttl-manager.php
```diff
@@ -0,0 +1,51 @@
...
+		$response_headers = $response->get_headers();
+		if ( isset( $response_headers[ 'Cache-Control' ] ) ) {
+			return $response;
+		}
+
+		if ( is_user_logged_in() ) {
+			return $response;
+		}
+
+		$ttl = apply_filters( 'wpcom_vip_rest_response_ttl', self::DEFAULT_API_TTL, $response, $rest_server, $request );
```
simonwheatley: This filter needs inline docs

simonwheatley:
r94687-deploy

@mjangda @joshbetz Moved what appeared to be minor implementation issues into this issue: https://github.com/Automattic/vip-go-mu-plugins/issues/673

Issue #673, opened 2017-07-03
Iterate REST Request TTL


simonwheatley:
Remaining tasks from https://github.com/Automattic/vip-go-mu-plugins/pull/672 after deploy

* [x] Refactor as functions rather than a class ([#](https://github.com/Automattic/vip-go-mu-plugins/pull/672#issuecomment-312378014))
* [x] Plugin header rewording ([#](https://github.com/Automattic/vip-go-mu-plugins/pull/672#discussion_r125284689))
* [x] Document filter ([#](https://github.com/Automattic/vip-go-mu-plugins/pull/672#discussion_r125288012))
* [x] Amend client facing docs here to document filter: https://vip.wordpress.com/documentation/vip-go/controlling-vip-go-page-cache/
* [x] Amend client facing docs here to document filter: https://vip.wordpress.com/documentation/vip-go/vip-go-and-the-wordpress-rest-api/#caching

simonwheatley:
Transferred the to do items about writing client docs to the description above 😃

PR #678, opened 2017-07-07
Update TTL Manager

mjangda:
Drops the class and updates docs.

See #673

### DIFF
--- a/vip-cache-manager/ttl-manager.php
+++ b/vip-cache-manager/ttl-manager.php
@@ -2,11 +2,11 @@
 
 /*
 Plugin name: TTL Manager
-Description: Sets sane and reasonable cache TTLs for site responses.
+Description: Sets cache TTLs for site responses, see https://vip.wordpress.com/documentation/vip-go/controlling-vip-go-page-cache/
 Author: Automattic
-Author URI: http://automattic.com/
+Author URI: https://automattic.com/
 Version: 1.0
-License: GPL version 2 or later - http://www.gnu.org/licenses/old-licenses/gpl-2.0.html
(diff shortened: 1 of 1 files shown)


vip-cache-manager/ttl-manager.php
```diff
@@ -2,50 +2,59 @@
...
+	return $response;
+}
+
+function set_rest_response_ttl( \WP_REST_Response $response, $ttl ) {
+	$response->header( 'Cache-Control', sprintf( 'max-age=%d', $ttl ) );
 }
 
-TTL_Manager::init();
+// Let's do it!
+init();
```
joshbetz: We could probably even call `add_filter()` directly here?

mjangda: Keeping it in `init` to allow for future, additional TTL overrides we may want to add.

vip-cache-manager/ttl-manager.php
```diff
@@ -2,50 +2,59 @@
 
 /*
 Plugin name: TTL Manager
-Description: Sets sane and reasonable cache TTLs for site responses.
+Description: Sets cache TTLs for site responses, see https://vip.wordpress.com/documentation/vip-go/controlling-vip-go-page-cache/
```
simonwheatley: Last night it occurred to me we should add more links 🖖

mjangda: Pretty sure [this was your idea](https://github.com/Automattic/vip-go-mu-plugins/pull/672#discussion_r125284689) :)

mjangda:
r95041-deploy

vip-cache-manager/ttl-manager.php
simonwheatley: 🙃

PR #680, opened 2017-07-13
Simplify `fix_remote_address()`

simonwheatley:
The `$remote_proxy_ip` will always be `$_SERVER['HTTP_X_FORWARDED_FOR']`, so we
should simplify the developer experience and not bother asking for it.

Thoughts, @mjangda? At the time of writing, nobody seems to be using this function.

### DIFF
--- a/lib/proxy/ip-forward.php
+++ b/lib/proxy/ip-forward.php
@@ -14,7 +14,7 @@ namespace Automattic\VIP\Proxy;
  *
  * @return (bool) true, if REMOTE_ADDR updated; false, if not.
  */
-function fix_remote_address( $user_ip, $remote_proxy_ip, $proxy_ip_whitelist ) {
+function fix_remote_address( $user_ip, $proxy_ip_whitelist ) {
 	if ( ! is_valid_ip( $user_ip ) ) {
 		return false;
 	}
@@ -22,7 +22,7 @@ function fix_remote_address( $user_ip, $remote_proxy_ip, $proxy_ip_whitelist ) {
 	require_once( __DIR__ . '/ip-utils.php' );
 
 	// Verify that the remote proxy matches our whitelist
-	$is_whitelisted_proxy_ip = IpUtils::checkIp( $remote_proxy_ip, $proxy_ip_whitelist );
+	$is_whitelisted_proxy_ip = IpUtils::checkIp( $_SERVER['HTTP_X_FORWARDED_FOR'], $proxy_ip_whitelist );
 
 	if ( ! $is_whitelisted_proxy_ip ) {
 		return false;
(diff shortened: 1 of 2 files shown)


mjangda:
I believe I had originally gone with this approach to simplify testing and to keep things flexible for the future. But I think it makes sense.

We should probably do the same thing with `ip_trail` in `fix_remote_address_from_ip_trail` since it will always be `$_SERVER['HTTP_X_IP_TRAIL']`.

mjangda:
If we want to simplify things further, Symfony has a nice API that we can try to mimic/borrow fron: http://symfony.com/doc/current/request/load_balancer_reverse_proxy.html

lib/proxy/ip-forward.php
```diff
@@ -14,15 +14,15 @@
...
+function fix_remote_address( $user_ip, $proxy_ip_whitelist ) {
 	if ( ! is_valid_ip( $user_ip ) ) {
 		return false;
 	}
 
 	require_once( __DIR__ . '/ip-utils.php' );
 
 	// Verify that the remote proxy matches our whitelist
-	$is_whitelisted_proxy_ip = IpUtils::checkIp( $remote_proxy_ip, $proxy_ip_whitelist );
+	$is_whitelisted_proxy_ip = IpUtils::checkIp( $_SERVER['HTTP_X_FORWARDED_FOR'], $proxy_ip_whitelist );
```
mjangda: Minor: do we need to verify that this SERVER var is set before accessing it (to prevent a notice)?

lib/proxy/ip-forward.php
```diff
@@ -43,20 +43,19 @@ function fix_remote_address( $user_ip, $remote_proxy_ip, $proxy_ip_whitelist ) {
...
  *
- * @param (string) $ip_trail Comma-separated list of IPs (something like `user_ip, proxy_ip`)
  * @param (string|array) $proxy_ip_whitelist Whitelisted IP addresses for the remote proxy. Supports IPv4 and IPv6, including CIDR format.
  *
  * @return (bool) true, if REMOTE_ADDR updated; false, if not.
  */
-function fix_remote_address_from_ip_trail( $ip_trail, $proxy_ip_whitelist ) {
-	$ip_addresses = get_ip_addresses_from_ip_trail( $ip_trail );
+function fix_remote_address_from_ip_trail( $proxy_ip_whitelist ) {
+	$ip_addresses = get_ip_addresses_from_ip_trail( $_SERVER['HTTP_X_IP_TRAIL'] );
```
mjangda: Minor: do we need to verify that this SERVER var is set before accessing it (to prevent a notice)?

mjangda:
This is a backwards-incompatible fix and not a ton of benefit, so closing.

PR #683, opened 2017-07-14
Disable jetpack-force-2fa for Jetpack Staging sites

mjangda:
SSO is not allowed on those sites so users end up locked out.


mjangda:
r95687-deploy

mjangda:
r95688-deploy

PR #685, opened 2017-07-19
Cron Control: bump to latest to prevent deadlock when callback fails

ethitter:
See https://github.com/Automattic/Cron-Control/pull/118


ethitter:
r95370-deploy

PR #686, opened 2017-07-20
Add one-time script for forcing go-vip to SSL

mjangda:
Any sites using a go-vip.co domain should have their domain set to https since we enforce that at the server level.

### DIFF
--- a/wp-cli/vip-fixers.php
+++ b/wp-cli/vip-fixers.php
@@ -28,7 +28,8 @@ class VIP_Go_OneTimeFixers_Command extends WPCOM_VIP_CLI_Command {
 				restore_current_blog();
 			}
 
-			$this->flush_cache();
+			// Flush the cache one more time just to be safe
+			wp_cache_flush();
 		} else {
 			$old_home_url = home_url();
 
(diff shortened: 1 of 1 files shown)


mjangda:
Noting that we'll remove this script one it's been run and all sites have been fixed.

wp-cli/vip-fixers.php
```diff
@@ -1,48 +1,110 @@
...
-		$downloaded_url = download_url( $image_url, 30 );
+	private function enforce_govip_ssl_for_current_site() {
+		$updated_home = $this->enforce_govip_ssl_for_option_url( 'home' );
+		$updated_siteurl = $this->enforce_govip_ssl_for_option_url( 'siteurl' );
 
-		// We couldn't download and store to a temporary location, so bail.
-		if ( is_wp_error( $downloaded_url ) ) {
-			return $downloaded_url;
+		if ( $updated_home || $updated_siteurl ) {
+			$this->flush_cache( get_site_url() );
```
ethitter: Since this happens in the site context, couldn't we just call `wp_cache_flush()`? Does running it through a second WP-CLI process solve some problem?

mjangda: Oh, that's much cleaner :)

mjangda:
r95487-deploy

PR #687, opened 2017-07-20
Cron Control: update to further improve lock cleanup

ethitter:
See https://github.com/Automattic/Cron-Control/pull/119


ethitter:
r95450-deploy

PR #688, opened 2017-07-20
Log errors in event execution

ethitter:
Logs caught errors along with event details, since they won't be included now that they're caught.

Also logs event details for uncaught error, to help correlate the error with its source. Often it's something value like an object-cache memory exhaustion with no meaningful backtrace.

See https://github.com/Automattic/Cron-Control/pull/119


ethitter:
r95460-deploy so that we'd have some visibility into fataling events. This still needs review, and I need to sort out why these aren't being tagged as true fatal errors.

001-cron.php
```diff
@@ -79,6 +79,30 @@ function wpcom_vip_disable_jetpack_sync_on_cron_shutdown( $load_sync ) {
...
+	wpcom_vip_log_event_object( $event, 'Caught' );
+}
+
+/**
+ * Convert event object to log entry
+ *
+ * @param $event object
+ * @param $type string
+ */
+function wpcom_vip_log_event_object( $event, $type = 'Uncaught' ) {
```
mjangda: This should probably be named `wpcom_vip_log_cron_control_event_object`?

001-cron.php
```diff
@@ -79,6 +79,30 @@ function wpcom_vip_disable_jetpack_sync_on_cron_shutdown( $load_sync ) {
...
+ * Log details of fatal error in callback that Cron Control caught
+ *
+ * @param $event object
+ * @param $error \Throwable
+ */
+function wpcom_vip_log_cron_control_event_for_caught_error( $event, $error ) {
+	$message = sprintf( 'PHP Fatal error:  Caught Error: %1$s in %2$s:%3$d', $error->getMessage(), $error->getFile(), $error->getLine() );
+	error_log( $message );
+
+	wpcom_vip_log_event_object( $event, 'Caught' );
```
mjangda: Having two different "PHP Fatal error" log entries for the same error might be a bit confusing?

001-cron.php
```diff
@@ -99,5 +123,11 @@ function wpcom_vip_disable_jetpack_sync_on_cron_shutdown( $load_sync ) {
 	 */
 	add_filter( 'jetpack_sync_sender_should_load', 'wpcom_vip_disable_jetpack_sync_on_cron_shutdown' );
 
+	/**
+	 * Log details of events that fail
+	 */
+	add_action( 'a8c_cron_control_event_threw_catchable_error', 'wpcom_vip_log_cron_control_event_for_caught_error', 10, 2 );
+	add_action( 'a8c_cron_control_freeing_event_locks_after_uncaught_error', 'wpcom_vip_log_event_object' );
```
mjangda: Now that I look at these action names again, feels like the tenses are a bit off. Probably would have been better named `a8c_cron_control_event_throw_catchable_error` (`threw` => `throw`) and `a8c_cron_control_free_event_locks_after_uncaught_error` (`freeing` => `free`).

001-cron.php
ethitter: Since we separate fatals from other log entries, I hoped this would group the errors together to simplify correlating the error with its extra data. That's not happening right now (our custom "fatals" are ending up in the regular log), and if it won't work that way, I'm happy to change the prefix.

001-cron.php
ethitter: The tenses seem odd here, but in the context of where the actions fire, they make _a bit_ more sense. The first, "threw," happens inside of a `catch{}`, which is why I chose past tense. The second, "freeing," happens after the uncaught exception occurs, but before the locks are freed, hence the gerund.

ethitter:
r95778-deploy

PR #690, opened 2017-07-26
Port WP.com's custom error handler

### LINKED ISSUE #636: Port WP.com's PHP Error Handler
`wpcom_error_handler` and related code.

ethitter:
Introduce version of WP.com's custom error handler, to be loaded via an update to the `wp-config.php` template managed by `vipd`.

Fixes #636


ethitter:
To match WP.com behaviour and ensure that our custom reporting captures as many errors as possible, this is loaded via `wp-config.php`, between when `ABSPATH` is defined and when `wp-settings.php` is loaded. See the linked `vipd` PR for that piece.

ethitter:
Errors (perhaps just surpassed errors) turned up in WP-CLI output despite Core constants that should've directed that to a log. Needs further investigation.

ethitter:
I'm waiting to deploy this until I have confirmation that log ingestion won't break as a result: pMz3w-7qt-p2.

ethitter:
Systems confirmed that this change won't break things, proceeding.

ethitter:
r96306-deploy

PR #691, opened 2017-07-28
Enforce FS_METHOD as direct.

mjangda:
Filesystem interaction on Go will always use the `direct` method. This also helps simplify instantiation of the global filesystem object (via `WP_Filesystem`), which by default tries to write to `WP_CONTENT_DIR` and fails (since only /tmp is writeable).

The workaround without setting this constant would look something like:

```
WP_Filesystem( false, get_temp_dir(), true );
```


mjangda:
r95865-deploy

PR #692, opened 2017-07-29
Force is_ssl() in wp-cli based on siteurl

mjangda:
We need to make sure the correct URLs are returned by the site_url() function, which relies on is_ssl() which defaults to `false` in our currently environment. Without this, you end up with incorrect URLs returned which can be problematic when using them to generate data (e.g. Jetpack connections).


mjangda:
r95943-deploy

PR #693, opened 2017-07-29
Fix scoping issues for client-mu-plugins

### LINKED ISSUE #668: client-mu-plugins: Fix scoping issues
Because we load the plugins inside a function, we're not giving `client-mu-plugins` the same global scope that `mu-plugins` get. 

We could simplify this (and the entire loading process) by getting rid of the function and just doing the loading the files directly inside the mu-plugins script.

props @fklein-lu ([#](https://github.com/Automattic/vip-go-mu-plugins/pull/611#issuecomment-311002566))

mjangda:
Don't load them inside a function. Instead, load them in global scope inside the mu-plugin loader so the the loaded client-mu-plugins also have global scope.

Fixes #668


z-client-mu-plugins.php
```diff
@@ -1,107 +1,85 @@
...
+	if ( empty( $client_mu_plugins_files ) ) {
+		return [];
 	}
 
-	foreach ( $plugin_files as $plugin_file ) {
-		if ( ! is_readable( WPCOM_VIP_CLIENT_MU_PLUGIN_DIR . "/$plugin_file" ) ) {
+	$client_mu_plugins_data = [];
+
+	foreach ( $client_mu_plugins_files as $plugin_file ) {
+		if ( ! is_readable( WPCOM_VIP_CLIENT_MU_PLUGIN_DIR . '/' . $plugin_file ) ) {
```
scottsweb: For some reason this path is a little strange for me when running this locally: 

`/home/scott/Sites/wordpress/wp-content/client-mu-plugins//home/scott/Sites/wordpress/wp-content/client-mu-plugins/test-plugin.php`

Not sure what has really changed (could be something on my local setup) but it is causing the endpoint to fail for me.

mjangda: Hm, that looks broken. Will take a look.

mjangda: Good catch; the return signatures were a little different between the function I have and the one we were originally using. Should be fixed with defec1d.

mjangda: Hm, probably should add tests...

mjangda:
r95962-deploy

PR #694, opened 2017-07-31
Don't load plugin when URLs are the same, it will just cause a redirect loop

### LINKED ISSUE #674: split-home-site-urls plugin should handle un-split domains
If `home` and `siteurl` are identical, the "split-home-site-urls" plugin creates a redirect loop. Though it clearly shouldn't be applied in this case, it also shouldn't break the site.

ethitter:
Fixes #674

### DIFF
--- a/shared-plugins/split-home-site-urls/split-home-site-urls.php
+++ b/shared-plugins/split-home-site-urls/split-home-site-urls.php
@@ -30,7 +30,7 @@ namespace Automattic\VIP\Split_Home_Site_URLs;
  *
  * URLs must differ by more than just protocol for this plugin to be useful
  */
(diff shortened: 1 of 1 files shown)


shared-plugins/split-home-site-urls/split-home-site-urls.php
```diff
@@ -26,6 +26,16 @@
 namespace Automattic\VIP\Split_Home_Site_URLs;
 
 /**
+ * Disable when plugin's conditions aren't met
+ *
+ * URLs must differ by more than just protocol for this plugin to be useful
+ */
+if ( preg_replace( '#https?://#i', '', home_url( '/' ) ) === preg_replace( '#https?://#i', '', site_url( '/' ) ) ) {
```
mjangda: I wonder if `str_replace` might be slightly faster?

```
str_replace( [ 'http://', 'https://' ], '', home_url( '/' ) )
```

ethitter:
r95953-deploy

PR #700, opened 2017-08-03
Correctly identify Cron & Remove from Apdex

sboisvert:
Currently Cron's go under "index.php" which muddies the waters in terms of what is and what isn't slow. It also makes some sites with low traffic seem slow because cron is a bigger proportion of the requests.
I'm adding this as plugins_loaded 11 instead of earlier to account for https://github.com/Automattic/Cron-Control/blob/master/includes/class-events.php#L27 which may load on plugins_loaded.

### DIFF
--- a/vip-helpers/vip-utils.php
+++ b/vip-helpers/vip-utils.php
@@ -1230,6 +1230,17 @@ function wpcom_vip_add_URI_to_newrelic(){
 add_action( 'muplugins_loaded', 'wpcom_vip_add_URI_to_newrelic' );
 
 /**
+ * Name cron correctly in New Relic and do not count it as part of the Apdex score
+ */
+function wpcom_vip_cron_for_newrelic(){
+	if ( defined( 'DOING_CRON' ) && DOING_CRON && function_exists( 'newrelic_ignore_apdex' ) && function_exists( 'newrelic_name_transaction' ) ){
+		newrelic_name_transaction( 'cron-control' );
+		newrelic_ignore_apdex();
+	}
+}
+add_action( 'plugins_loaded', 'wpcom_vip_cron_for_newrelic', 11 );
+
+/**
  * Send a message to IRC
  *
  * $level can be an int of one of the following
--- a/vip-helpers/vip-utils.php
+++ b/vip-helpers/vip-utils.php
@@ -1241,6 +1241,18 @@ function wpcom_vip_cron_for_newrelic(){
 add_action( 'plugins_loaded', 'wpcom_vip_cron_for_newrelic', 11 );
 
 /**
+ * Name wp-cli correctly in New Relic and do not count it as part of the Apdex score
+ */
+function wpcom_vip_wpcli_for_newrelic(){
+	if ( defined( 'WP_CLI' ) && WP_CLI && ( ! defined( 'DOING_CRON' ) || ! DOING_CRON ) && function_exists( 'newrelic_ignore_apdex' ) && function_exists( 'newrelic_name_transaction' ) ){
+		newrelic_name_transaction( 'wp-cron' );
+		newrelic_ignore_apdex();
+	}
+}
+add_action( 'plugins_loaded', 'wpcom_vip_cron_for_newrelic', 11 );
+
+
+/**
  * Send a message to IRC
  *
  * $level can be an int of one of the following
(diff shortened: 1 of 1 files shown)


ethitter:
If this should apply only to Cron Control, `wpcom_vip_use_core_cron()` should be checked. We don't load Cron Control for Basic Auth sites, for example.

sboisvert:
Also Name WP-CLI

vip-helpers/vip-utils.php
```diff
@@ -1230,6 +1230,29 @@ function wpcom_vip_add_URI_to_newrelic(){
...
+	}
+}
+add_action( 'plugins_loaded', 'wpcom_vip_cron_for_newrelic', 11 );
+
+/**
+ * Name wp-cli correctly in New Relic and do not count it as part of the Apdex score
+ */
+function wpcom_vip_wpcli_for_newrelic(){
+	if ( defined( 'WP_CLI' ) && WP_CLI && ( ! defined( 'DOING_CRON' ) || ! DOING_CRON ) && function_exists( 'newrelic_ignore_apdex' ) && function_exists( 'newrelic_name_transaction' ) ){
+		newrelic_name_transaction( 'wp-cron' );
```
ethitter: This should be `wp-cli`, no?

sboisvert:
@ethitter fixed and also fixed using the wrong function name in the hook.

mjangda:
We should put this into a new file where we can put all New Relic-related code (instead of `vip-utils.php`).

vip-helpers/vip-utils.php
```diff
@@ -1230,6 +1230,29 @@ function wpcom_vip_add_URI_to_newrelic(){
 add_action( 'muplugins_loaded', 'wpcom_vip_add_URI_to_newrelic' );
 
 /**
+ * Name cron correctly in New Relic and do not count it as part of the Apdex score
```
mjangda: We should mention "why" we don't want it to count in the score in the inline comment.

vip-helpers/vip-utils.php
```diff
@@ -1230,6 +1230,29 @@ function wpcom_vip_add_URI_to_newrelic(){
...
+ */
+function wpcom_vip_cron_for_newrelic(){
+	if ( defined( 'DOING_CRON' ) && DOING_CRON && function_exists( 'newrelic_ignore_apdex' ) && function_exists( 'newrelic_name_transaction' ) ){
+		newrelic_name_transaction( 'wp-cron' );
+		newrelic_ignore_apdex();
+	}
+}
+add_action( 'plugins_loaded', 'wpcom_vip_cron_for_newrelic', 11 );
```
mjangda: Need an inline comment why we're using priority `11`.

vip-helpers/vip-utils.php
```diff
@@ -1230,6 +1230,29 @@ function wpcom_vip_add_URI_to_newrelic(){
...
+
+/**
+ * Name wp-cli correctly in New Relic and do not count it as part of the Apdex score
+ */
+function wpcom_vip_wpcli_for_newrelic(){
+	if ( defined( 'WP_CLI' ) && WP_CLI && ( ! defined( 'DOING_CRON' ) || ! DOING_CRON ) && function_exists( 'newrelic_ignore_apdex' ) && function_exists( 'newrelic_name_transaction' ) ){
```
mjangda: This conditional is starting to get pretty complex. Can we break it out a bit to make it easier to read (and less likely to break with changes)?

vip-helpers/vip-utils.php
```diff
@@ -1230,6 +1230,29 @@ function wpcom_vip_add_URI_to_newrelic(){
...
+		newrelic_name_transaction( 'wp-cli' );
+		newrelic_ignore_apdex();
+	}
+}
+add_action( 'plugins_loaded', 'wpcom_vip_wpcli_for_newrelic', 11 );
```
mjangda: Need an inline comment why we're using priority `11`.

vip-helpers/vip-utils.php
```diff
@@ -1230,6 +1230,29 @@ function wpcom_vip_add_URI_to_newrelic(){
...
+function wpcom_vip_cron_for_newrelic(){
+	if ( defined( 'DOING_CRON' ) && DOING_CRON && function_exists( 'newrelic_ignore_apdex' ) && function_exists( 'newrelic_name_transaction' ) ){
+		newrelic_name_transaction( 'wp-cron' );
+		newrelic_ignore_apdex();
+	}
+}
+add_action( 'plugins_loaded', 'wpcom_vip_cron_for_newrelic', 11 );
+
+/**
+ * Name wp-cli correctly in New Relic and do not count it as part of the Apdex score
```
mjangda: We should mention "why" we don't want it to count in the score in the inline comment.

vip-helpers/vip-utils.php
mjangda: Plus, why do we need the cron check? Should probably have an inline comment explaining that.

mjangda:
This was continued in #708

PR #701, opened 2017-08-04
Bump Akismet to 3.3.4

mjangda:
- Disabled Akismet’s debug log output by default unless AKISMET_DEBUG is defined.
- URL previews now begin preloading when the mouse moves near them in the comments section of wp-admin.
- When a comment is caught by the Comment Blacklist, Akismet will always allow it to stay in the trash even if it is spam as well.
- Fixed a bug that was preventing an error from being shown when a site can’t reach Akismet’s servers.


mjangda:
r96152-deploy

PR #702, opened 2017-08-05
Make plugins_url work with client-mu-plugins

### LINKED ISSUE #675: client-mu-plugins: make sure plugins_url works
We may need to port over the hack from WP.com to make sure `plugins_url` works when referencing a file in `client-mu-plugins`.

mjangda:
By default, it only works with plugins and mu-plugins. Any calls inside client-mu-plugins end up with a non-existent path and this change helps return a correct URL instead.

Fixes #675

### DIFF
--- a/z-client-mu-plugins.php
+++ b/z-client-mu-plugins.php
@@ -81,6 +81,37 @@ function wpcom_vip_get_client_mu_plugins_data( $directory = WPCOM_VIP_CLIENT_MU_
 	return $client_mu_plugins_data;
 }
 
+/**
+ * This callback hooked on `plugins_url` enables, allows us call `plugins_url` with a client-mu-plugins path (or inside any file in that path).
+ *
+ * e.g. plugins_url( 'file.js', WPCOM_VIP_CLIENT_MU_PLUGIN_DIR . '/plugin/file.php' );
+ */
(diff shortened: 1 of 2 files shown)


z-client-mu-plugins.php
```diff
@@ -81,6 +81,37 @@ function wpcom_vip_get_client_mu_plugins_data( $directory = WPCOM_VIP_CLIENT_MU_
...
 
+/**
+ * This callback hooked on `plugins_url` enables, allows us call `plugins_url` with a client-mu-plugins path (or inside any file in that path).
+ *
+ * e.g. plugins( 'file.js', WPCOM_VIP_CLIENT_MU_PLUGIN_DIR . '/plugin/file.php' );
+ */
+function wpcom_vip_filter_client_mu_plugins_url( $url, $url_path, $plugin_path ) {
+	static $mu_plugins_dir, $mu_plugins_url;
+
+	if ( ! isset( $plugins_dir ) ) {
```
ethitter: Shouldn't this be `$mu_plugins_dir`?

mjangda: Yep, it should.

mjangda:
r96364-deploy

PR #703, opened 2017-08-07
Support converting `latin1` tables to `utf8mb4`

ethitter:
Much of the work for this came from #297, but that was limited to converting `utf8` to `utf8mb4`. Per https://pento.net/2014/04/07/wordpress-and-utf-8/, the commands are the same to convert `latin1` to `utf8mb4`, so supporting that conversion is mostly a matter of removing existing blocks.


ethitter:
Related: p73hgi-tP-p2

ethitter:
Not so simple if there are non-ASCII characters in the `latin1` tables, which there will be in many of our cases. Conversion tricks to handle this:

* https://nicj.net/mysql-converting-an-incorrect-latin1-column-to-utf8/
* http://www.varesano.net/blog/fabio/latin1%20encoded%20tables%20or%20databases%20utf8%20data%20stored%20convert%20them%20native%20mysql%20utf8%20tables

ethitter:
This is ready for review. I still need to implement `enum` support, but that's defensive more than anything. WordPress.com doesn't use `enum` in any blog-specific tables.

ethitter:
Once the constants are updated, the command can't be run again, but I haven't tested what happens if it's run again before the constant was updated. I'm curious what that would do if, say, someone re-ran it because it failed the first time through.

ethitter:
`enum` support added, and I confirmed that running again doesn't break anything, as no convertible columns/tables are found.

ethitter:
Easiest way to test this is using an export from a WordPress.com site that uses non-latin characters.

1. Create an export of the site using `~/bin/dump-blog-sql.php` on your WP.com sandbox.
2. Import that into a test environment
3. Update the `DB_CHARSET` constant to `latin1` and clear any value set for `DB_COLLATE`.
4. Flush caches
5. Run `wp vip utf8mb4 convert --protect-latin1`
6. Update `DB_CHARSET` to `utf8mb4`
7. Flush caches
8. Check the site and/or its DB for mangled content.

wp-cli/vip-go-convert-to-utf8mb4.php
```diff
@@ -0,0 +1,348 @@
...
+		'mediumtext' => 'mediumblob',
+		'longtext'   => 'longblob',
+		'varchar'    => 'varbinary', // length handled in maybe_protect_column()
+		'enum'       => 'enum',
+	);
+
+	/**
+	 * Convert site using `utf8` or `latin1` to use `utf8mb4`
+	 *
+	 * @subcommand convert
```
mjangda: Minor: would be nice to note the `dry-run` and `protect-latin1` here.

wp-cli/vip-go-convert-to-utf8mb4.php
```diff
@@ -0,0 +1,348 @@
...
+			WP_CLI::success( "Set database to utf8mb4" );
+		} else {
+			WP_CLI::warning( "Could not update database default charset" );
+		}
+
+		// Wrap up
+		WP_CLI::line( '' );
+		WP_CLI::line( '' );
+		WP_CLI::line( 'DONE!' );
+		WP_CLI::line( 'Time to update sitemeta and reload web configs.' );
```
mjangda: Minor: would be nice to explicitly note what the meta key and value should be here.

wp-cli/vip-go-convert-to-utf8mb4.php
```diff
@@ -0,0 +1,348 @@
...
+		}
+
+		$null_not_null = 'yes' === strtolower( $col->Null ) ? ' NULL' : ' NOT NULL';
+
+		// On with it!
+		WP_CLI::line( "Converting column {$col->Field}" );
+
+		$pattern = 'ALTER TABLE %1$s CHANGE %2$s %2$s %3$s %4$s %5$s DEFAULT "%6$s"';
+		$convert = $wpdb->query( $wpdb->prepare( $pattern, $table, $col->Field, $to_type, '', $null_not_null, $col->Default ) );
+		$restore = $wpdb->query( $wpdb->prepare( $pattern, $table, $col->Field, $from_type, 'CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci', $null_not_null, $col->Default ) );
```
mjangda: Minor: Not totally clear why we need to convert and then restore; inline comments would be helpful.

ethitter:
r96518-deploy

PR #706, opened 2017-08-09
Don't call WPCOM_VIP_Cache_Manager::purge_site_cache() when saving posts under WP_IMPORTING

### LINKED ISSUE #705: Don't BAN entire site cache on clean_post_cache when WP_IMPORTING is true
Originally put in place for initial site imports - but `WP_IMPORTING` can be defined under a number of situations, such as content syndication, so we should not automatically BAN the entire site there.

https://github.com/Automattic/vip-go-mu-plugins/blob/33c65f5f72190578ba6f91315094ef3deeea8088/vip-cache-manager/vip-cache-manager.php#L271-L274

nickdaugherty:
WP_IMPORTING can be set under a number of conditions, such as
syndication via REST apis, and shouldn’t end up BAN-ing the entire site.

Question - what impact will this have on CLI-based imports? Lots of extra code below the removed block could end up running, which could really slow things down. Needs some thought.

Fixes #705

### DIFF
--- a/vip-cache-manager/vip-cache-manager.php
+++ b/vip-cache-manager/vip-cache-manager.php
@@ -63,7 +63,7 @@ class WPCOM_VIP_Cache_Manager {
 		$button_html .= '</p><p><span class="button"><a href="' . esc_url( $url ) . '"><strong>';
 		$button_html .= esc_html__( 'Purge Page Cache' );
 		$button_html .= '</strong></a></span>';
-		
+
 		echo "<p>$button_html</p>\n";
 	}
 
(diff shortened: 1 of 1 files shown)


nickdaugherty:
An alternative is to still exit early but not run the site BAN, though that's imperfect b/c can lead to stale caches.

ethitter:
> Question - what impact will this have on CLI-based imports? Lots of extra code below the removed block could end up running, which could really slow things down. Needs some thought.

I was wondering the same thing. Perhaps we should only remove the call to `purge_site_cache()`?

ethitter:
We might be okay with documentation, leaving it to CLI-based importers to manage caches themselves. Some clients are already actively managing their caches, so they may not notice at all.

nickdaugherty:
r96317-deploy

PR #707, opened 2017-08-09
Pass '' to dbDelta

sboisvert:
If nothing is passed to dbDelta it will not grab the schema and therefore the queries will be empty which will lead to not actually doing the dbDelta (but the command will tell us it was successful)

### DIFF
--- a/wp-cli/vip-migrations.php
+++ b/wp-cli/vip-migrations.php
@@ -30,13 +30,6 @@ class VIP_Go_Migrations_Command extends WPCOM_VIP_CLI_Command {
 		global $wpdb;
 
 		$tables =  $args[1];
-		$valid_tables = array( '', 'all', 'blog', 'global', 'ms_global' );
-
-		if ( empty( $tables ) ) {
-			$tables = '';
-		} else if ( ! in_array( $tables, $valid_tables, true ) ) {
-			WP_CLI::error( sprintf( '%s is not a valid entry. Valid entries are: %s', $tables, implode('", "', $valid_tables ) ) );
-		}
 
 		$network = Utils\get_flag_value( $assoc_args, 'network' );
 		if ( $network && ! is_multisite() ) {
--- a/wp-cli/vip-migrations.php
+++ b/wp-cli/vip-migrations.php
@@ -16,7 +16,15 @@ class VIP_Go_Migrations_Command extends WPCOM_VIP_CLI_Command {
 	 * ## OPTIONS
 	 *
 	 * [<tables>]
-	 * : Which tables to update (all, blog, global, ms_global)
+	 * : Which tables to update (all, blog, global, ms_global, "")
+	 * ---
+	 * default: ""
+	 * options:
+	 *   - all
+	 *   - blog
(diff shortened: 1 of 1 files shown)


wp-cli/vip-migrations.php
```diff
@@ -66,7 +66,7 @@ function dbdelta( $args, $assoc_args ) {
 		if ( in_array( $args[1], array( '', 'all', 'blog', 'global', 'ms_global' ), true ) ) {
 			$changes = dbDelta( $args[1], ! $dry_run );
 		} else {
-			$changes = dbDelta( null, ! $dry_run );
+			$changes = dbDelta( '', ! $dry_run );
```
mjangda: I think we should probably change the behaviour here. If `<tables>` is not set, we should default to `''` (this should happen very early in the command). And in this `else`, we should throw an error since an invalid `tables` value was specified.

wp-cli/vip-migrations.php
```diff
@@ -20,7 +20,15 @@ class VIP_Go_Migrations_Command extends WPCOM_VIP_CLI_Command {
 	 */
 	function dbdelta( $args, $assoc_args ) {
 		global $wpdb;
-
+		
+		$tables =  $args[1];
+		
+		if ( empty( $tables ) ){
```
mjangda: Minor: missing space.

wp-cli/vip-migrations.php
```diff
@@ -20,7 +20,15 @@ class VIP_Go_Migrations_Command extends WPCOM_VIP_CLI_Command {
...
+			$tables = '';
+		} else if ( ! in_array( $args[1], array( '', 'all', 'blog', 'global', 'ms_global' ), true ) ) {
```
mjangda: Minor: `$args[1]` should be `$tables`

wp-cli/vip-migrations.php
```diff
@@ -20,7 +20,15 @@ class VIP_Go_Migrations_Command extends WPCOM_VIP_CLI_Command {
...
+			WP_CLI::error( sprintf( '%s is not a valid entry. Valid entries are: "all", "blog", "global", "ms_global", "" (blank string)', $tables ) );
```
mjangda: If we move the list of valid values into a separate variable (array), we can just append that to the error message instead of hard-coding the values twice (something like `sprintf( '...Valid entries are one of: "%s"', implode( '","', $valid_tables ) );`

mjangda: We should also update the command docblock to mention the empty string option + default.

wp-cli/vip-migrations.php
```diff
@@ -20,7 +20,15 @@ class VIP_Go_Migrations_Command extends WPCOM_VIP_CLI_Command {
 	 */
 	function dbdelta( $args, $assoc_args ) {
 		global $wpdb;
-
+
```
mjangda: Minor: can we remove the unnecessary indents from the empty lines added in this changeset?

mjangda:
r96366-deploy

PR #708, opened 2017-08-09
Move New Relic separate file + Track WP_CLI and Cron

sboisvert:
Move all New Relic logic to one file and add logging of Cron and WP_CLI as separate transactions that don't impact the Apdex score of the site (since those transactions don't impact what users get in terms of performance)

### DIFF
--- a/vip-helpers/vip-newrelic.php
+++ b/vip-helpers/vip-newrelic.php
@@ -58,7 +58,8 @@ function wpcom_vip_cron_for_newrelic() {
 		&& function_exists( 'newrelic_add_custom_parameter' )
 		&& function_exists( 'newrelic_name_transaction' ) ) {
 		newrelic_name_transaction( 'wp-cron' );
-		newrelic_add_custom_parameter( 'wp-cron', 'true' );
+		// N.B. VIP Go sites execute cron events through WP CLI, so the event
+		// can be determined in the New Relic 'wp-cli-cmd-args' custom parameter
 		newrelic_ignore_apdex();
 	}
 }
--- a/vip-helpers/vip-newrelic.php
+++ b/vip-helpers/vip-newrelic.php
@@ -21,8 +21,8 @@ function wpcom_vip_disable_new_relic_js() {
 /**
  * Add the exact URI to NewRelic tracking but only if we're not in the admin
  */
-function wpcom_vip_add_URI_to_newrelic(){
-	if ( ! is_admin() && function_exists( 'newrelic_add_custom_parameter' ) ){
+function wpcom_vip_add_uri_to_newrelic() {
+	if ( ! is_admin() && function_exists( 'newrelic_add_custom_parameter' ) ) {
 		newrelic_capture_params();
 		newrelic_add_custom_parameter( 'HTTP_REFERER', isset( $_SERVER['HTTP_REFERER'] ) ? $_SERVER['HTTP_REFERER'] : '' );
 		newrelic_add_custom_parameter( 'HTTP_USER_AGENT', isset( $_SERVER['HTTP_USER_AGENT'] ) ? $_SERVER['HTTP_USER_AGENT'] : '' );
@@ -36,29 +36,29 @@ add_action( 'muplugins_loaded', 'wpcom_vip_add_URI_to_newrelic' );
  *
  * We don't want to count cron as part of the apdex because it is not a user facing function and if cron tasks are slow it doesn't imply that the site's performance is impacted. Without removing these, long running cron tasks could flag the site as having performance problems, which would cause false positives in the monitoring.
  */
-function wpcom_vip_cron_for_newrelic(){
-	if ( wp_doing_cron() && function_exists( 'newrelic_ignore_apdex' ) && function_exists( 'newrelic_name_transaction' ) ){
+function wpcom_vip_cron_for_newrelic() {
+	if ( wp_doing_cron() && function_exists( 'newrelic_ignore_apdex' ) && function_exists( 'newrelic_name_transaction' ) ) {
 		newrelic_name_transaction( 'wp-cron' );
 		newrelic_add_custom_parameter( 'wp-cron', 'true' );
 		newrelic_ignore_apdex();
 	}
 }
-add_action( 'muplugins_loaded', 'wpcom_vip_cron_for_newrelic', 11 ); //We are attaching this at muplugins_loaded because Cron-Control is loaded at muplugins_loaded priority 10
+add_action( 'muplugins_loaded', 'wpcom_vip_cron_for_newrelic', 11 ); // We are attaching this at muplugins_loaded because Cron-Control is loaded at muplugins_loaded priority 10
 
 /**
  * Name wp-cli correctly in New Relic and do not count it as part of the Apdex score
  *
  * We don't want to count ongoing WP-CLI requests as part of the apdex because it is not a user facing function and if a WP-CLI request is slow it doesn't imply that the site's performance is impacted. Without removing these WP-CLI requests from the apdex calculation it could flag the site as having performance problems, which would cause false positives in the monitoring.
  */
-function wpcom_vip_wpcli_for_newrelic(){
+function wpcom_vip_wpcli_for_newrelic() {
 	if ( defined( 'WP_CLI' )
 	     && WP_CLI
 	     && ! wp_doing_cron()  // Cron is going to be run via WP-CLI in the near term. We want to keep Cron and WP-CLI separated for better monitoring so we're not going to flag WP_CLI requests that are actually cron requests as WP-CLI.
 	     && function_exists( 'newrelic_ignore_apdex' )
-	     && function_exists( 'newrelic_name_transaction' ) ){
+	     && function_exists( 'newrelic_name_transaction' ) ) {
 		newrelic_name_transaction( 'wp-cli' );
 		newrelic_add_custom_parameter( 'wp-cli', 'true' );
 		newrelic_ignore_apdex();
 	}
 }
-add_action( 'muplugins_loaded', 'wpcom_vip_wpcli_for_newrelic', 11 );  //We are attaching this at muplugins_loaded because Cron-Control is loaded at muplugins_loaded priority 10
+add_action( 'muplugins_loaded', 'wpcom_vip_wpcli_for_newrelic', 11 );  // We are attaching this at muplugins_loaded because Cron-Control is loaded at muplugins_loaded priority 10
--- a/vip-helpers/vip-newrelic.php
+++ b/vip-helpers/vip-newrelic.php
@@ -95,23 +95,6 @@ function wpcom_vip_wpcli_for_newrelic() {
 		// e.g. `wp option get siteurl`
 		newrelic_add_custom_parameter( 'wp-cli-cmd', $cmd );
 
-		$assoc_args = \WP_CLI::get_runner()->assoc_args;
-		$config = \WP_CLI::get_runner()->config;
-		// Empty values clutter the result
-		foreach ( $config as $config_arg => $config_value ) {
-			if ( empty( $config_value ) ) {
-				unset( $config[$config_arg] );
-			}
-		}
-		$all_config = array_merge( $assoc_args, $config );
-		// Key sorting normalises the resultant string
-		ksort( $all_config );
-		// e.g. `allow-root=1&color=auto&format=json&skip-plugins=1` which is
-		// equivalent to `--allow-root --color=auto --format=json --skip-plugins`
-		// We have this format because building a query string is easy and understandable
-		// (and a hack) whereas rebuilding a WP CLI style command parameters is harder
-		newrelic_add_custom_parameter( 'wp-cli-cmd-args', urldecode(http_build_query( $all_config ) ) );
-
 		newrelic_ignore_apdex();
 	}
 }
--- a/vip-helpers/vip-newrelic.php
+++ b/vip-helpers/vip-newrelic.php
@@ -26,6 +26,7 @@ function wpcom_vip_add_URI_to_newrelic(){
 		newrelic_capture_params();
 		newrelic_add_custom_parameter( 'HTTP_REFERER', isset( $_SERVER['HTTP_REFERER'] ) ? $_SERVER['HTTP_REFERER'] : '' );
 		newrelic_add_custom_parameter( 'HTTP_USER_AGENT', isset( $_SERVER['HTTP_USER_AGENT'] ) ? $_SERVER['HTTP_USER_AGENT'] : '' );
+		newrelic_add_custom_parameter( 'HTTPS', isset( $_SERVER['HTTPS'] ) ? $_SERVER['HTTPS'] : '' );
 	}
 }
 add_action( 'muplugins_loaded', 'wpcom_vip_add_URI_to_newrelic' );
--- a/vip-helpers/vip-newrelic.php
+++ b/vip-helpers/vip-newrelic.php
@@ -29,8 +29,11 @@ function wpcom_vip_disable_new_relic_js() {
 
 if ( extension_loaded( 'newrelic' ) ){
 	add_action( 'muplugins_loaded', 'wpcom_vip_add_uri_to_newrelic' );
-	add_action( 'muplugins_loaded', 'wpcom_vip_cron_for_newrelic', 11 ); //We are attaching this at muplugins_loaded because Cron-Control is loaded at muplugins_loaded priority 10
-	add_action( 'muplugins_loaded', 'wpcom_vip_wpcli_for_newrelic', 11 );  //We are attaching this at muplugins_loaded because Cron-Control is loaded at muplugins_loaded priority 10
+	// We are attaching this at muplugins_loaded because Cron-Control is loaded at muplugins_loaded priority 10
+	add_action( 'muplugins_loaded', 'wpcom_vip_cron_for_newrelic', 11 );
+	// We are attaching this at muplugins_loaded because Cron-Control is loaded at muplugins_loaded priority 10
+	// This must be hooked later than wpcom_vip_cron_for_newrelic to allow values to be overwritten
+	add_action( 'muplugins_loaded', 'wpcom_vip_wpcli_for_newrelic', 12 );
 	add_action( 'rest_dispatch_request', 'wpcom_vip_rest_routes_for_newrelic', 10,4);
 }
 
--- a/vip-helpers/vip-newrelic.php
+++ b/vip-helpers/vip-newrelic.php
@@ -82,7 +82,16 @@ function wpcom_vip_wpcli_for_newrelic() {
 
 		newrelic_name_transaction( 'wp-cli' );
 
-		$cmd = 'wp ' . implode( ' ', \WP_CLI::get_runner()->arguments );
+		$wp_cli_arguments = \WP_CLI::get_runner()->arguments;
+		if ( empty( $wp_cli_arguments ) ) {
+			// Not much to do at this point
+			return;
+		}
+		if ( ! is_array( $wp_cli_arguments ) ) {
+			$wp_cli_arguments = [ $wp_cli_arguments ];
+		}
+		array_unshift( $wp_cli_arguments, 'wp' );
+		$cmd = implode( ' ', $wp_cli_arguments );
 		// e.g. `wp option get siteurl`
 		newrelic_add_custom_parameter( 'wp-cli-cmd', $cmd );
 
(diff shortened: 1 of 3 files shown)


vip-helpers/vip-newrelic.php
```diff
@@ -0,0 +1,61 @@
...
+}
+add_action( 'muplugins_loaded', 'wpcom_vip_add_URI_to_newrelic' );
+
+/**
+ * Name cron correctly in New Relic and do not count it as part of the Apdex score.
+ *
+ * We don't want to count cron as part of the apdex because it is not a user facing function and if cron tasks are slow it doesn't imply that the site's performance is impacted. Without removing these, long running cron tasks could flag the site as having performance problems, which would cause false positives in the monitoring.
+ */
+function wpcom_vip_cron_for_newrelic(){
+	if ( defined( 'DOING_CRON' ) && DOING_CRON && function_exists( 'newrelic_ignore_apdex' ) && function_exists( 'newrelic_name_transaction' ) ){
```
ethitter: I missed the addition in 4.8, but we can check `wp_doing_cron()` instead of using the constant. We don't have any sites running < 4.8, so I don't think a `function_exists()` check is needed.

vip-helpers/vip-newrelic.php
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Name wp-cli correctly in New Relic and do not count it as part of the Apdex score
+ *
+ * We don't want to count ongoing WP-CLI requests as part of the apdex because it is not a user facing function and if a WP-CLI request is slow it doesn't imply that the site's performance is impacted. Without removing these WP-CLI requests from the apdex calculation it could flag the site as having performance problems, which would cause false positives in the monitoring.
+ */
+function wpcom_vip_wpcli_for_newrelic(){
+	if ( defined( 'WP_CLI' )
+	     && WP_CLI
+	     && ( ! defined( 'DOING_CRON' ) || ! DOING_CRON ) // Cron is going to be run via WP-CLI in the near term. We want to keep Cron and WP-CLI separated for better monitoring so we're not going to flag WP_CLI requests that are actually cron requests as WP-CLI.
```
ethitter: Likewise, this can be `wp_doing_cron()` instead.

vip-helpers/vip-newrelic.php
```diff
@@ -0,0 +1,61 @@
...
+	     && function_exists( 'newrelic_ignore_apdex' )
+	     && function_exists( 'newrelic_name_transaction' ) ){
+		newrelic_name_transaction( 'wp-cli' );
+		newrelic_ignore_apdex();
+	}
+}
+add_action( 'plugins_loaded', 'wpcom_vip_wpcli_for_newrelic', 11 ); //We are attaching this at plugins_loaded because it's possible for Cron-Control to trigger at plugins_loaded priority 10 (at the latest), we therefore can't tell until then if this is a WP-CLI request that is also a cron request.
```
ethitter: It's safe to use `muplugins_loaded` here, though there's likely little practical difference. If Cron Control is loaded, it'll be done before `muplugins_loaded` fires. The higher priority is needed if the hook is switched.

vip-helpers/vip-newrelic.php
```diff
@@ -0,0 +1,61 @@
...
+	if ( defined( 'WP_CLI' )
+	     && WP_CLI
+	     && ! wp_doing_cron()  // Cron is going to be run via WP-CLI in the near term. We want to keep Cron and WP-CLI separated for better monitoring so we're not going to flag WP_CLI requests that are actually cron requests as WP-CLI.
+	     && function_exists( 'newrelic_ignore_apdex' )
+	     && function_exists( 'newrelic_name_transaction' ) ){
+		newrelic_name_transaction( 'wp-cli' );
+		newrelic_ignore_apdex();
+	}
+}
+add_action( 'muplugins_loaded', 'wpcom_vip_wpcli_for_newrelic', 11 ); //We are attaching this at plugins_loaded because it's possible for Cron-Control to trigger at plugins_loaded priority 10 (at the latest), we therefore can't tell until then if this is a WP-CLI request that is also a cron request.
```
mjangda: The hook mentioned in the inline comment (`plugins_loaded`) doesn't match the actual hook we're using (`muplugins_loaded`).

vip-helpers/vip-newrelic.php
```diff
@@ -0,0 +1,61 @@
+<?php
```
mjangda: Do we need to include this file? Or is it auto-loaded?

vip-helpers/vip-newrelic.php
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * We don't want to count cron as part of the apdex because it is not a user facing function and if cron tasks are slow it doesn't imply that the site's performance is impacted. Without removing these, long running cron tasks could flag the site as having performance problems, which would cause false positives in the monitoring.
+ */
+function wpcom_vip_cron_for_newrelic(){
+	if ( wp_doing_cron() && function_exists( 'newrelic_ignore_apdex' ) && function_exists( 'newrelic_name_transaction' ) ){
+		newrelic_name_transaction( 'wp-cron' );
+		newrelic_ignore_apdex();
+	}
+}
+add_action( 'muplugins_loaded', 'wpcom_vip_cron_for_newrelic', 11 ); //We are attaching this at plugins_loaded because it's possible for Cron-Control to trigger at plugins_loaded priority 10 (at the latest)
```
mjangda: Same comment about inline comment.

vip-helpers/vip-newrelic.php
sboisvert: Fixed

vip-helpers/vip-newrelic.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
+/**
+ * Disable New Relic's browser metrics
+ *
+ * Removes NR's JavaScript for tracking browser metrics, including page load times, Apdex score, and more.
+ *
```
simonwheatley: Do we prefer disabling this feature (RUM) in code or in the New Relic application settings?

vip-helpers/vip-newrelic.php
```diff
@@ -0,0 +1,61 @@
...
+		newrelic_disable_autorum();
+	}
+}
+
+/**
+ * Add the exact URI to NewRelic tracking but only if we're not in the admin
+ */
+function wpcom_vip_add_URI_to_newrelic(){
+	if ( ! is_admin() && function_exists( 'newrelic_add_custom_parameter' ) ){
+		newrelic_add_custom_parameter( 'REQUEST_URI', isset( $_SERVER['REQUEST_URI'] ) ? $_SERVER['REQUEST_URI'] : '' );
```
simonwheatley: Should we use `newrelic_capture_params` ([docs](https://docs.newrelic.com/docs/agents/php-agent/php-agent-api/newrelic_capture_params)) instead?

vip-helpers/vip-newrelic.php
```diff
@@ -0,0 +1,61 @@
...
+
+	if ( function_exists( 'newrelic_disable_autorum' ) ) {
+		newrelic_disable_autorum();
+	}
+}
+
+/**
+ * Add the exact URI to NewRelic tracking but only if we're not in the admin
+ */
+function wpcom_vip_add_URI_to_newrelic(){
```
simonwheatley: Is it worth including the scheme (`is_ssl()`)?

vip-helpers/vip-newrelic.php
```diff
@@ -0,0 +1,61 @@
...
+add_action( 'muplugins_loaded', 'wpcom_vip_add_URI_to_newrelic' );
+
+/**
+ * Name cron correctly in New Relic and do not count it as part of the Apdex score.
+ *
+ * We don't want to count cron as part of the apdex because it is not a user facing function and if cron tasks are slow it doesn't imply that the site's performance is impacted. Without removing these, long running cron tasks could flag the site as having performance problems, which would cause false positives in the monitoring.
+ */
+function wpcom_vip_cron_for_newrelic(){
+	if ( wp_doing_cron() && function_exists( 'newrelic_ignore_apdex' ) && function_exists( 'newrelic_name_transaction' ) ){
+		newrelic_name_transaction( 'wp-cron' );
```
simonwheatley: Should we add the event name as a custom parameter?

vip-helpers/vip-newrelic.php
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * We don't want to count ongoing WP-CLI requests as part of the apdex because it is not a user facing function and if a WP-CLI request is slow it doesn't imply that the site's performance is impacted. Without removing these WP-CLI requests from the apdex calculation it could flag the site as having performance problems, which would cause false positives in the monitoring.
+ */
+function wpcom_vip_wpcli_for_newrelic(){
+	if ( defined( 'WP_CLI' )
+	     && WP_CLI
+	     && ! wp_doing_cron()  // Cron is going to be run via WP-CLI in the near term. We want to keep Cron and WP-CLI separated for better monitoring so we're not going to flag WP_CLI requests that are actually cron requests as WP-CLI.
+	     && function_exists( 'newrelic_ignore_apdex' )
+	     && function_exists( 'newrelic_name_transaction' ) ){
+		newrelic_name_transaction( 'wp-cli' );
```
simonwheatley: Should we add the command as a custom parameter?

vip-helpers/vip-newrelic.php
sboisvert: That's how it was previously, I think in code is easier for devs to do since they might want to remove it but not have access to New Relic.

vip-helpers/vip-newrelic.php
sboisvert: IS it possible to be running SSL and non SSL?

vip-helpers/vip-newrelic.php
sboisvert: Why as a custom parameter? I think we want these to become their own transactions so we can monitor them (or remove them from the monitoring), if it's a custom param I don't think we can quite as easily remove them from calculations

vip-helpers/vip-newrelic.php
sboisvert: same as above

vip-helpers/vip-newrelic.php
sboisvert: Probably?

vip-helpers/vip-newrelic.php
mjangda: Yes.

vip-helpers/vip-newrelic.php
mjangda: It's also useful in other contexts like AMP.

vip-helpers/vip-newrelic.php
simonwheatley: Sorry to be unclear, I meant add the event name as a custom parameter as well as what you're already doing here.

vip-helpers/vip-newrelic.php
simonwheatley: Same as above: I've been unclear. Adding the name makes it easier to identify individual commands which have inefficiencies and to know what you're looking at in New Relic.

vip-helpers/vip-newrelic.php
simonwheatley: Got it, thanks both.

vip-helpers/vip-newrelic.php
sboisvert: Got it, done!

vip-helpers/vip-newrelic.php
sboisvert: Done

vip-helpers/vip-newrelic.php
sboisvert: done

mjangda:
In a future iteration, two improvements would be nice:

- Separate out lib functions (i.e. functions for client use like `wpcom_vip_disable_new_relic_js`) from others that are meant for internal use (like the CLI/cron naming functions).
- Only hook the "internal" functions/callbacks if New Relic is loaded.

vip-helpers/vip-newrelic.php
simonwheatley: I meant add the event name in a custom parameter, something like:

``` php
newrelic_add_custom_parameter( 'wp-cron', $cron_event_name );
```

vip-helpers/vip-newrelic.php
simonwheatley: Similar to above, I meant add the event name in a custom parameter, something like:

``` php
newrelic_add_custom_parameter( 'wp-cron', $cli_command_name );
```

simonwheatley:
> My boss told me to approve this before Simon got back. ;)

🙄

simonwheatley:
@sboisvert ICYMI: 

https://github.com/Automattic/vip-go-mu-plugins/pull/708#discussion_r133706275
https://github.com/Automattic/vip-go-mu-plugins/pull/708#discussion_r133706445

sboisvert:
@simonwheatley I have them added https://github.com/Automattic/vip-go-mu-plugins/pull/708/files#diff-f428b005c9f320c8c3ed89d2b82beea0R79 Unless I'm missing something?

simonwheatley:
@sboisvert I meant that we should pass the cron **event name** and the WP CLI **command name**… however I've just been looking into accessing the name of the currently running command in WP CLI, and I cannot see a way to do it. 

Same for Cron events, unless @ethitter can if Cron Control is running each Cron event as a separate WP CLI command, so we can hook in there…?

ethitter:
> ...I've just been looking into accessing the name of the currently running command in WP CLI, and I cannot see a way to do it.

WP-CLI doesn't provide an "official" way to do this, but it is possible. Cron Control does it, but it's pretty hacky: https://github.com/Automattic/Cron-Control/blob/6728765565e588674fc79d7c40fb18ecbf7fa0d0/includes/wp-cli.php#L9.

As for Cron Control, events are run through a special set of commands nested under the `cron-control orchestrate` namespace, so as long as the commands are identified, Cron Control need not do anything more.

vip-helpers/vip-newrelic.php
```diff
@@ -0,0 +1,125 @@
...
+				unset( $config[$config_arg] );
+			}
+		}
+		$all_config = array_merge( $assoc_args, $config );
+		// Key sorting normalises the resultant string
+		ksort( $all_config );
+		// e.g. `allow-root=1&color=auto&format=json&skip-plugins=1` which is
+		// equivalent to `--allow-root --color=auto --format=json --skip-plugins`
+		// We have this format because building a query string is easy and understandable
+		// (and a hack) whereas rebuilding a WP CLI style command parameters is harder
```
simonwheatley: Would we prefer to manually rebuild in the style of WP CLI, e.g. `--allow-root --color=auto --format=json --skip-plugins`, or are we happy enough with the query string style representation?

vip-helpers/vip-newrelic.php
```diff
@@ -0,0 +1,125 @@
...
+ * Name cron correctly in New Relic and do not count it as part of the Apdex score.
+ *
+ * We don't want to count cron as part of the apdex because it is not a user facing function and if cron tasks are slow it doesn't imply that the site's performance is impacted. Without removing these, long running cron tasks could flag the site as having performance problems, which would cause false positives in the monitoring.
+ */
+function wpcom_vip_cron_for_newrelic() {
+	if ( wp_doing_cron()
+		&& function_exists( 'newrelic_ignore_apdex' )
+		&& function_exists( 'newrelic_add_custom_parameter' )
+		&& function_exists( 'newrelic_name_transaction' ) ) {
+		newrelic_name_transaction( 'wp-cron' );
```
simonwheatley: Cron Control uses WP CLI to run Cron events, so Cron events will (I assume) show up as WP CLI commands (both `WP_CLI` being truthy and `wp_doing_cron()`). Because `wpcom_vip_wpcli_for_newrelic()` has a lower priority (12) than `wpcom_vip_cron_for_newrelic()` (11), and my assumption is that the later call to `newrelic_name_transaction()` will rename the transaction to Cron.

simonwheatley:
@sboisvert I added some code and some questions about my approach. If you're happy with these changes, I think we should throw this over to @mjangda or @ethitter as previously uninvolved parties to review our code.

sboisvert:
Looks good by me. Should we have some validation on \WP_CLI::get_runner()->arguments ? Or do we think it's pretty solid?

simonwheatley:
Ready for review, @ethitter @mjangda 😃

vip-helpers/vip-newrelic.php
ethitter: Shouldn't they be in the other order, then? Right now, `wpcom_vip_cron_for_newrelic()` will fire before `wpcom_vip_wpcli_for_newrelic()`.

vip-helpers/vip-newrelic.php
```diff
@@ -0,0 +1,134 @@
...
+			$wp_cli_arguments = [ $wp_cli_arguments ];
+		}
+		array_unshift( $wp_cli_arguments, 'wp' );
+		$cmd = implode( ' ', $wp_cli_arguments );
+		// e.g. `wp option get siteurl`
+		newrelic_add_custom_parameter( 'wp-cli-cmd', $cmd );
+
+		$assoc_args = \WP_CLI::get_runner()->assoc_args;
+		$config = \WP_CLI::get_runner()->config;
+		// Empty values clutter the result
```
ethitter: `array_filter()` can accomplish the same.

vip-helpers/vip-newrelic.php
ethitter: I'm not sure. Maybe we should start with this and see if further effort is warranted?

vip-helpers/vip-newrelic.php
```diff
@@ -0,0 +1,134 @@
...
+ * By default wp-api requests are tagged under index.php
+ * We'd want to have them tagged with the proper rest route used.
+ * While we are using the rest_dispatch_request filter, we're using it as an action without modifying the results.
+ */
+function wpcom_vip_rest_routes_for_newrelic( $dispatch_results, $request, $route, $handler ) {
+	if ( function_exists( 'newrelic_add_custom_parameter' )
+	     && function_exists( 'newrelic_name_transaction' ) ) {
+		newrelic_name_transaction( $route );
+		newrelic_add_custom_parameter( 'wp-api', 'true' );
+		newrelic_add_custom_parameter( 'wp-api-route', $route );
```
ethitter: Do we care that Cron Control will turn up here too? We only use the endpoints for monitoring at this point, so it won't skew things as much as including them with WP-CLI will.

vip-helpers/vip-newrelic.php
sboisvert: Line 78 
&& ! wp_doing_cron()  
Should take care of this.

vip-helpers/vip-newrelic.php
sboisvert: @simonwheatley Any reason not to use it?

vip-helpers/vip-newrelic.php
sboisvert: I think this is good enough for now

vip-helpers/vip-newrelic.php
sboisvert: I don't think we care if it shows up or not. We can change it if needed.

sboisvert:
@ethitter Any feedback on my feedback?

ethitter:
@sboisvert Nope, but there are still two open issues/questions in `vip-helpers/vip-newrelic.php`.

sboisvert:
@ethitter I'm not sure which ones you are referring to

ethitter:
@sboisvert: https://github.com/Automattic/vip-go-mu-plugins/pull/708#discussion_r138246061 and https://github.com/Automattic/vip-go-mu-plugins/pull/708#discussion_r138246775

vip-helpers/vip-newrelic.php
sboisvert: @simonwheatley ping about ^^

sboisvert:
The first one seems to be resolved no? pinged @simonwheatley on the second one.

vip-helpers/vip-newrelic.php
mjangda: We need to be careful here since wp-cli commands may pass secret args like passwords. We should not be logging those.

I think just the command and subcommand are probably good enough for now. If we want to include other args, we need to whitelist them (but I don't think that's scalable.

sboisvert:
REady for one last review.

simonwheatley:
Thanks @sboisvert, I'm interested to find if we need that command info back again, at least in part, to adequately analyse which commands are slow… deploying this will help us find out!

ethitter:
r99138-deploy

sboisvert:
Confirmed working!

image

PR #712, opened 2017-08-10
Fix WP-CLI with alternate cron

joshbetz:
Since alternate cron redirects to ?doing_cron, it breaks WP-CLI. This disables alternate cron in the CLI context.


mjangda:
r96365-deploy

PR #714, opened 2017-08-11
Enable Cron Control for Basic Auth sites

ethitter:
With the switch to the CLI runner, this exception is unnecessary.

See https://github.com/Automattic/Cron-Control/pull/114

### DIFF
--- a/001-cron.php
+++ b/001-cron.php
@@ -11,21 +11,13 @@
 /**
  * Determine if Cron Control is called for
  *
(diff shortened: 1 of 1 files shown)


ethitter:
**Until CLI containers exist for every site, this cannot be deployed.**

ethitter:
> We should probably ignore these from the cron queue monitoring?

Yup. They're already excluded, and the differential I have to convert the runner to a monitor maintains those exemptions.

ethitter:
r96544-deploy

PR #715, opened 2017-08-11
Via `sunrise.php`, log invalid multisite configurations

ethitter:
Depends on:
* software-stacks update: 1ad16-pb
* `vipd` update: https://github.com/Automattic/vipd/pull/448

### DIFF
--- a/lib/sunrise/sunrise.php
+++ b/lib/sunrise/sunrise.php
@@ -1,0 +1,54 @@
+<?php
+
+namespace Automattic\VIP\Sunrise;
+
+/**
+ * Nothing to see here for single sites
+ */
+if ( ! is_multisite() ) {
+	return;
+}
+
+/**
+ * Log errors retrieving network for a given request
+ *
+ * @param string $domain
+ * @param string $path
+ */
+function network_not_found( $domain, $path ) {
+	$data = [
+		'domain_requested' => $domain,
+		'path_requested'   => $path,
+	];
+	$data = json_encode( $data );
+
+	error_log( 'Network Not Found! ' . $data );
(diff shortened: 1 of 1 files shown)


lib/sunrise/sunrise.php
```diff
@@ -0,0 +1,54 @@
...
+ * @param string $path
+ */
+function network_not_found( $domain, $path ) {
+	$data = [
+		'domain_requested' => $domain,
+		'path_requested'   => $path,
+	];
+	$data = json_encode( $data );
+
+	error_log( 'Network Not Found! ' . $data );
```
mjangda: For both error messages, we should probably include some sort of unique hashtag? Could also use `trigger_error` with `E_USER_WARNING`?

lib/sunrise/sunrise.php
```diff
@@ -0,0 +1,54 @@
...
+}
+add_action( 'ms_network_not_found', __NAMESPACE__ . '\network_not_found', 9, 2 );
```
mjangda: Why do we need priority `9`? Should probably have an inline comment with the reason.

ethitter: Good call. It's because `WP_CLI` hooks at 10 and uses its `error()` method, so our callback won't fire.

lib/sunrise/sunrise.php
ethitter: I like the `trigger_error()` idea. The leading strings are unique enough for alerting.

ethitter:
r96390-deploy

The aforementioned dependences **have not** been approved/deployed.

PR #716, opened 2017-08-11
Remove recursion in Cron Control option to disable automatic execution

ethitter:
Prior to https://github.com/Automattic/Cron-Control/commit/ddf190124f5444f6a5a6e3468416c2240115440b, I attempted not to check or set the option unnecessarily. The changes from the aforementioned commit obviate those optimizations.


ethitter:
r96388-deploy

PR #718, opened 2017-08-11
Clearing transient on add_attachement

pkevan:
- checks new attachment against (removing query as it's not needed)
- make sure there is a transient before trying to remove it

### DIFF
--- a/performance/vip-tweaks.php
+++ b/performance/vip-tweaks.php
@@ -1,11 +1,12 @@
 <?php
+
 /**
  * This is a list of performance tweaks that will become default for All VIP sites
  */
 function wpcom_vip_enable_performance_tweaks() {
 	/**
 	 * Improves performance of all the wp-admin pages that load comment counts in the menu.
-	 * 
+	 *
 	 * This caches them for 30 minutes. It does not impact the per page comment count, only
 	 * the total comment count that shows up in the admin menu.
 	 */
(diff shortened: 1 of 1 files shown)


performance/vip-tweaks.php
```diff
@@ -40,29 +40,12 @@ function wpcom_vip_bust_media_months_cache( $post_id ) {
...
-		! $months->year == get_the_time( 'Y', $post_id ) &&
-		! $months->month == get_the_time( 'm', $post_id )
-	) {
-		// the new attachment is not in the same month/year as the
-		// most recent attachment, so we need to refresh the transient
+	// Grab the transient to see if it needs updating
+	$media_months = get_transient( 'wpcom_media_months_array' );
+	
+    	// If the transient exists, and the attachment uploaded doesn't match the first (latest) month or year in the transient, lets clear it.
+    	if ( false !== $media_months && ( $media_months[0]->year !== get_the_time( 'Y', $post_id ) || $media_months[0]->month !== get_the_time( 'm', $post_id ) ) ) {
```
mjangda: I wonder if we should change `false !== $media_months` to an `! empty` check in case we cached an empty array.

pkevan:
The query we are trying to avoid wouldn't be bad with no attachments (which
I'm assuming would come from an empty array )

mjangda:
Sorry, my concern was more around fatals if we hit an empty array and then try to access `$media_months[0]->year`.

pkevan:
@mjangda I misunderstood, added in some checks for the values existing.

Also fixed up the date value compare.

Fixes #718

mjangda:
@pkevan pushed some PHPCS and readability improvements.  Feel free to revert if you disagree with the changes but I think this is ready for deploy.

PR #719, opened 2017-08-14
Resolve several notices in custom Debug Bar panels

ethitter:
These are only an issue when our `db.php` and `object-cache.php` aren't loaded, such as in local development, but they generate enough noise that one can miss important errors.

* PHP Notice:  Undefined property: WP_Object_Cache::$time_total in /mu-plugins/vip-helpers/vip-debug-bar-panels.php on line 70
* PHP Notice:  Undefined index: query in /mu-plugins/vip-helpers/vip-debug-bar-panels.php on line 126
* PHP Notice:  Undefined index: elapsed in /mu-plugins/vip-helpers/vip-debug-bar-panels.php on line 149
* PHP Notice:  Undefined variable: query_time_pct in /mu-plugins/vip-helpers/vip-debug-bar-panels.php on line 171

### DIFF
--- a/vip-helpers/vip-debug-bar-panels.php
+++ b/vip-helpers/vip-debug-bar-panels.php
@@ -66,8 +66,12 @@ class WPCOM_VIP_Debug_Bar_Queries extends Debug_Bar_Panel {
 		$query_time = '<h2><span>Total query time:</span>' . number_format( sprintf( '%0.1f', $total_time * 1000 ), 1 ) . "ms</h2>\n";
 
 		$memory_usage  = '<h2><span>Peak Memory Used:</span>' . number_format( memory_get_peak_usage( ) ) . " bytes</h2>\n";
-		$memcache_time = property_exists( $wp_object_cache, 'time_total' ) ? number_format( sprintf( '%0.1f', $wp_object_cache->time_total * 1000 ), 1, '.', ',' ) : '?';
-		$memcache_time = '<h2><span>Total memcache query time:</span>' . $memcache_time . "ms</h2>\n";
+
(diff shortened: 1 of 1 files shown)


vip-helpers/vip-debug-bar-panels.php
```diff
@@ -66,8 +66,8 @@ function render() {
 		$query_time = '<h2><span>Total query time:</span>' . number_format( sprintf( '%0.1f', $total_time * 1000 ), 1 ) . "ms</h2>\n";
 
 		$memory_usage  = '<h2><span>Peak Memory Used:</span>' . number_format( memory_get_peak_usage( ) ) . " bytes</h2>\n";
-		$memcache_time = '<h2><span>Total memcache query time:</span>' .
-			number_format( sprintf( '%0.1f', $wp_object_cache->time_total * 1000 ), 1, '.', ',' ) . "ms</h2>\n";
+		$memcache_time = property_exists( $wp_object_cache, 'time_total' ) ? number_format( sprintf( '%0.1f', $wp_object_cache->time_total * 1000 ), 1, '.', ',' ) : '?';
+		$memcache_time = '<h2><span>Total memcache query time:</span>' . $memcache_time . "ms</h2>\n";
```
joshbetz: Maybe we should completely remove the memcached time in this case? Or add a link to a document about setting up memcached?

ethitter: Yeah, thinking about this more, I think it's better to just hide that section if the property isn't set. I'm not sure we want to maintain instructions for local dev memcache setups.

ethitter:
r96471-deploy

PR #721, opened 2017-08-15
Limit Jetpack connection management

mjangda:
To only our machine user. All connections should be happening through Jetpack Start and not managed by individual users.


mjangda:
Testing with an `administrator` user:

```
$ wp --allow-root shell --user=batmoox
wp> current_user_can( 'connect_jetpack' )
bool(false)
```

Testing with `wpcomvip`:

```
$ wp --allow-root shell --user=wpcomvip
wp> current_user_can( 'connect_jetpack' )
bool(true)
```

mjangda:
Typically, you would get an option to disconnect (under "Manage site connection"):

image

But with this enabled, it's no longer possible:

image

mjangda:
r98069-deploy

PR #722, opened 2017-08-17
Bump Cron Control to latest

ethitter:
Fixes event sorting when listing for CLI display.

See https://github.com/Automattic/Cron-Control/pull/133


ethitter:
r96580-deploy

PR #725, opened 2017-08-18
Move (unused) split-URLs plugin to its own repo

ethitter:
See https://github.com/Automattic/vip-go-split-home-site-urls


mjangda:
Is there anything VIP-specific in that plugin? Looks generic enough so maybe the repo is better prefixed or suffixed with `wp-` instead?

ethitter:
Nothing VIP-specific about it, but @scottsweb was hoping to enforce some naming conventions as we split these out, and I complied. :)

scottsweb:
I am proposing we take more ownership of our plugins going forward. So I think we should have a prefix of `vip-` for general plugins (e.g. `vip-coauthors`) and then platform specific prefixes too (e.g. `vip-go` and `vip-wpcom`) for plugins targeted at specific platforms.

mjangda:
r98201-deploy

PR #728, opened 2017-08-22
Adding image-delta WP CLI command

david-binda:
As there might be cases in which some images are imported to environment different from production, we need a way for moving the image delta to production upon content sync.

The newly introduced `wp vip migration image-delta` command makes it easy to import the image delta right after the SQL has been brought from preprod or dev to prod.

Using the `--uploaded-after-id` or `--uploaded-after-date` params allows you to import only images, which made it to preprod/dev after the initial import and thus import only the delta.

The ID or a date of the latest post from intial SQL import has to be determined manually.

### DIFF
--- a/wp-cli/vip-migrations.php
+++ b/wp-cli/vip-migrations.php
@@ -74,7 +74,7 @@ class VIP_Go_Migrations_Command extends WPCOM_VIP_CLI_Command {
 		require_once( ABSPATH . 'wp-admin/includes/upgrade.php' );
 
 		$changes = dbDelta( $tables, ! $dry_run );
-		
+
 		if ( empty( $changes ) ) {
 			WP_CLI::success( 'No changes.' );
 			return;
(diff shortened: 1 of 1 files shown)


mjangda:
Thanks @david-binda! I noted a question in Slack about how common the need for this command is.

Beyond that, at a high-level, there's too much actual logic within the CLI command class itself. If we proceed with this command, I think the file download / file exists functionality should be moved out into functions that can be used elsewhere.

david-binda:
> Beyond that, at a high-level, there's too much actual logic within the CLI command class itself. If we proceed with this command, I think the file download / file exists functionality should be moved out into functions that can be used elsewhere.

Totally agree. Ideally we would be able to reuse code from https://github.com/Automattic/vip-go-mu-plugins/blob/master/a8c-files.php#L353 and https://github.com/Automattic/vip-go-mu-plugins/blob/master/a8c-files.php#L232 . We'll have to make those methods static (or make the object available globally, or via `get_instance()` method) and generalise them a bit more. Let me look into it.

mjangda:
Closing; we should move this to the utility scripts repo if we still need it (and should refresh to standard upload functions).

PR #729, opened 2017-08-23
Introduce constant for sites with moved shared plugins

scottsweb:
This constant allows us to disable loading shared plugins - in particular it will stop `vip-go-elasticsearch` loading (or any other plugin we need to retire) in order for us to safely switch away.

Two arrays are added: One for retired plugins and one for protected shared plugins (2fa etc).

Test with:

`define( 'WPCOM_VIP_DISABLE_SHARED_PLUGINS', true );`

### DIFF
--- a/vip-helpers/vip-utils.php
+++ b/vip-helpers/vip-utils.php
@@ -990,29 +990,11 @@ function wpcom_vip_load_plugin( $plugin = false, $folder = false, $load_release_
 	// Array of directories to check for the above files in, in priority order
 	$test_directories = array();
 
-	// Array of shared plugins we are not deprecating
-	$protected_shared_plugins = array(
-		'two-factor',
-		'jetpack-force-2fa',
-	);
-
-	// Array of shared plugins that we cannot move
-	$retired_shared_plugins = array(
-		'vip-go-elasticsearch',
-	);
-
 	if ( $folder ) {
 		$test_directories[] = WP_PLUGIN_DIR . '/' . $folder;
 	} else {
 		$test_directories[] = WP_PLUGIN_DIR;
-		if ( defined( 'WPCOM_VIP_DISABLE_SHARED_PLUGINS' ) ) {
-			if ( in_array( $plugin, $protected_shared_plugins, true ) ) {
-				$test_directories[] = WPMU_PLUGIN_DIR . '/shared-plugins';
-			}
-			if ( in_array( $plugin, $retired_shared_plugins, true ) ) {
-				return;
-			}
-		} else {
+		if ( wpcom_vip_can_use_shared_plugin( $plugin ) ) {
 			$test_directories[] = WPMU_PLUGIN_DIR . '/shared-plugins';
 		}
 	}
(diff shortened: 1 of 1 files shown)


vip-helpers/vip-utils.php
```diff
@@ -990,11 +990,31 @@ function wpcom_vip_load_plugin( $plugin = false, $folder = false, $load_release_
...
+	$retired_shared_plugins = array(
+		'vip-go-elasticsearch',
+	);
+
 	if ( $folder ) {
 		$test_directories[] = WP_PLUGIN_DIR . '/' . $folder;
 	} else {
 		$test_directories[] = WP_PLUGIN_DIR;
-		$test_directories[] = WPMU_PLUGIN_DIR . '/shared-plugins';
+		if ( defined( 'WPCOM_VIP_DISABLE_SHARED_PLUGINS' ) ) {
```
nickdaugherty: We should also check that the value of the constant equals `true`, so that `define( 'WPCOM_VIP_DISABLE_SHARED_PLUGINS', false );` doesn't trigger it.

vip-helpers/vip-utils.php
```diff
@@ -990,11 +990,31 @@ function wpcom_vip_load_plugin( $plugin = false, $folder = false, $load_release_
...
+			if ( in_array( $plugin, $protected_shared_plugins, true ) ) {
+				$test_directories[] = WPMU_PLUGIN_DIR . '/shared-plugins';
+			}
+			if ( in_array( $plugin, $retired_shared_plugins, true ) ) {
+				return;
```
nickdaugherty: It seems like this block is unnecessary? If we don't add the shared-plugins dir to the $test_directories then it'll never match.

I guess what it accomplishes is it prevents the plugins in $retired_shared_plugins from ever being loaded via function?

Should we just move the ES plugin into $protected_shared_plugins for now since we have to keep it around for a bit?

mjangda: > It seems like this block is unnecessary? If we don't add the shared-plugins dir to the $test_directories then it'll never match.

I wonder if we can simplify it like so:

```
if ( defined( 'WPCOM_VIP_DISABLE_SHARED_PLUGINS' ) && true === WPCOM_VIP_DISABLE_SHARED_PLUGINS
    && ! in_array( $plugin, $protected_shared_plugins, true ) ) {
    return;
} else {
    $test_directories[] = WPMU_PLUGIN_DIR . '/shared-plugins';
}
```

And +1  to moving ES to the protected list.

scottsweb: It is not that simple. If we `return` after that first line then we don't ever get to the point where a once shared plugin is loaded from `/plugins` instead. We just give up loading. 

e.g. I have moved getty from `shared` to `/plugins` and it is no longer loading at all as we bail.

I think it probably needs to change to this:

```
if ( defined( 'WPCOM_VIP_DISABLE_SHARED_PLUGINS' ) && true === WPCOM_VIP_DISABLE_SHARED_PLUGINS && ! in_array( $plugin, $protected_shared_plugins, true ) ) {
    // do nothing
} else {
    $test_directories[] = WPMU_PLUGIN_DIR . '/shared-plugins';
}
```

Even with this, in my testing I end up triggering: `die( "Unable to load $plugin using wpcom_vip_load_plugin()!" );` - this doesn't happen on production luckily as it is wrapped in a check. Is that desired for local installs if we cannot find the plugin?

mjangda: Ah, good point. That seems fine. If we wanted, we could avoid the empty if by moving the check to a function (but not strictly necessary since this is temporary):

```
function wpcom_vip_can_use_shared_plugin( $plugin ) {
    if ( ! defined( 'WPCOM_VIP_DISABLE_SHARED_PLUGINS' ) ) {
        return true;
    }

    if ( true !== WPCOM_VIP_DISABLE_SHARED_PLUGINS ) {
         return true;
    }

    return in_array( $plugin, $protected_shared_plugins, true );
}

...

if ( wpcom_vip_can_use_shared_plugin( $plugin ) ) {
    $test_directories[] = WPMU_PLUGIN_DIR . '/shared-plugins';
}

mjangda:
r98199-deploy

PR #730, opened 2017-08-23
Fixing delete transient wpcom_media_months_array

pkevan:
When a transient is set, it always returns so isn't the expected behaviour.

We want to clear when data isn't correct and there is a transient set.


simonwheatley:
Fixes issues in https://github.com/Automattic/vip-go-mu-plugins/pull/718

PR #732, opened 2017-08-23
Add analytics tracking to vip_powered_wpcom_url to match wordpress.co…

### LINKED ISSUE #731: Add Analytics Tracking to vip_powered_wpcom_url()
Track the same parameters we are on wordpress.com and also add `utm_term` with the site's hostname for additional tracking.

chrisscott:
…m. Additionally, add the hostname as a tracked variable.

Fixes #731. 

Note: I've not sanitized the `$_SERVER` var since it isn't being stored, relying on any usage of this to escape the output URL properly. But happy to change that to sanitize and/or escape if it makes sense.

### DIFF
--- a/vip-helpers/vip-utils.php
+++ b/vip-helpers/vip-utils.php
@@ -120,7 +120,7 @@ function vip_powered_wpcom( $display = 'text', $before_text = 'Powered by ' ) {
  * @return string
  */
 function vip_powered_wpcom_url() {
-	return 'https://vip.wordpress.com/';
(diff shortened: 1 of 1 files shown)


ethitter:
> Note: I've not sanitized the `$_SERVER` var since it isn't being stored, relying on any usage of this to escape the output URL properly. But happy to change that to sanitize and/or escape if it makes sense.

I think it's safer to escape this just in case someone uses it for something unexpected in the future.

Could also use `add_query_arg()` with an array of the args, to make it easier to parse/update.

chrisscott:
@ethitter thanks, added the suggested changes.

ethitter:
r96868-deploy

PR #734, opened 2017-08-24
Clarify constant name used to identify CLI containers

ethitter:
I meant to include this in #733 and forgot to push it. While CLI containers do run on jobs hosts, the check we're performing has nothing to do with the host, but with the container.


ethitter:
r96885-deploy due to potential to break cron events, and approval from #733.

PR #737, opened 2017-08-30
Bypass Woocommerce file creation

### LINKED ISSUE #736: Stop Woocommerce from trying to install files in read-only filesystem
Following https://github.com/woocommerce/woocommerce/pull/16601, we can use the `woocommerce_install_skip_create_files` filter to stop Woocommerce from trying to create files under a read-only filesystem.

ethitter:
Prevents Woocommerce from trying to create files where they can't be created, which just leads to lots of `mkdir(): Read-only file system` errors.

Fixes #736

### DIFF
--- a/misc.php
+++ b/misc.php
@@ -148,6 +148,4 @@ add_filter( 'wp_link_query_args', 'wpcom_vip_wp_link_query_args', 10, 1 );
 /**
  * Stop Woocommerce from trying to create files on read-only filesystem
  */
-if ( false !== VIP_GO_ENV ) {
-	add_filter( 'woocommerce_install_skip_create_files', '__return_true' );
-}
+add_filter( 'woocommerce_install_skip_create_files', '__return_true' );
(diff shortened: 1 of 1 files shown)


ethitter:
It's worth noting that this won't have any effect until Woocommerce 3.2 is deployed to the sites that use it (see https://github.com/woocommerce/woocommerce/pull/16601).

misc.php
```diff
@@ -144,3 +144,10 @@ function wpcom_vip_wp_link_query_args( $query ) {
 }
 
 add_filter( 'wp_link_query_args', 'wpcom_vip_wp_link_query_args', 10, 1 );
+
+/**
+ * Stop Woocommerce from trying to create files on read-only filesystem
+ */
+if ( false !== VIP_GO_ENV ) {
```
mjangda: Should we do this everywhere to preserve production behaviour locally?

ethitter:
> Ideally, WooCommerce should be smart enough to detect when it's working with a read-only file system and behave appropriately.

Core really should be smarter about this, but I doubt they'll be receptive to a patch to `wp_mkdir_p()` simply to reduce notices. :) I may still try it though, because I like banging my head against the wall on occasion.

misc.php
ethitter: Yeah, I have no strong argument for keeping this limitation. Someone can always remove the filter in their local environment, if it's causing them trouble.

mjangda:
r98067-deploy

PR #738, opened 2017-09-05
Normalize URLs before adding them to purge list

mjangda:
Mostly, we want to strip off the fragment off any URLs since servers don't care about them.

So, a URL like `https://example.com/#q1` is added to the PURGE list as `https://example.com/` instead.

### DIFF
--- a/vip-cache-manager/vip-cache-manager.php
+++ b/vip-cache-manager/vip-cache-manager.php
@@ -564,16 +564,15 @@ class WPCOM_VIP_Cache_Manager {
 	}
 
 	protected function normalize_purge_url( $url ) {
-		$url = esc_url_raw( $url );
+		$normalized_url = esc_url_raw( $url );
 
-		$parsed_url = parse_url( $url );
-
-		// Strip off any URL fragments (i.e. `#search=xyz`) since that is ignored by the cache
-		if ( isset( $parsed_url['fragment'] ) ) {
-			unset( $parsed_url['fragment'] );
+		// Easy way to strip off any fragments since we don't have access to `http_build_url`.
+		$fragment_index = mb_strpos( $normalized_url, '#' );
+		if ( false !== $fragment_index ) {
+			$normalized_url = mb_substr( $normalized_url, 0, $fragment_index );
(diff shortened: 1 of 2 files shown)


vip-cache-manager/vip-cache-manager.php
```diff
@@ -556,6 +559,23 @@ public function queue_old_permalink_purge( $post_ID, $post_after, $post_before )
 		}
 	}
 
+	protected function normalize_purge_url( $url ) {
+		$url = esc_url_raw( $url );
+
+		$parsed_url = parse_url( $url );
+
+		// Strip off any URL fragments (i.e. `#search=xyz`) since that is ignored by the cache
```
simonwheatley: As our caching purges all query string variants of a URL, should we also strip the `query` portion of the URL? This might further reduce the number of URLs, but does make the system less DRY as a whole in that we would need to change the code here and in the cache config if we needed to remove/adjust this behaviour.

nickdaugherty: Is that part of our Varnish config, or Mangle?

simonwheatley: I think I've seen this in the Varnish config, and certainly the behaviour existed before Mangle was implemented.

joshbetz: Confirmed, it's in the Varnish config.

mjangda: I think it's okay to leave the `query` bits in the PURGE URLs for now. Can revisit later if we find PURGE counts are still high for sites.

mjangda:
`Error: Call to undefined function http_build_url()` :(

Looks like we need a polyfill like https://github.com/jakeasmith/http_build_url or an alternate approach.

joshbetz:
We should also ensure we're stripping anything after `?` because all query string permutations are also purged.

mjangda:
> We should also ensure we're stripping anything after ? because all query string permutations are also purged.

Done in 241244c

mjangda:
r99373-deploy

PR #739, opened 2017-09-05
Limit the number of PURGE and BAN urls

mjangda:
No need to PURGE thousands of URLs at a time.

### DIFF
--- a/vip-cache-manager/vip-cache-manager.php
+++ b/vip-cache-manager/vip-cache-manager.php
@@ -11,6 +11,9 @@ License: GPL version 2 or later - http://www.gnu.org/licenses/old-licenses/gpl-2
 require_once( __DIR__ . '/api.php' );
 
 class WPCOM_VIP_Cache_Manager {
+	const MAX_PURGE_URLS = 100;
+	const MAX_BAN_URLS   = 10;
+
 	private $ban_urls = array();
 	private $purge_urls = array();
 	private $site_cache_purged = false;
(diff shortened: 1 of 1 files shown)


vip-cache-manager/vip-cache-manager.php
```diff
@@ -11,6 +11,9 @@
 require_once( __DIR__ . '/api.php' );
 
 class WPCOM_VIP_Cache_Manager {
+	const MAX_PURGE_URLS = 100;
+	const MAX_BAN_URLS   = 10;
```
mjangda: Need to decide what makes sense for these max numbers.

joshbetz: This seems like a good place to start.

vip-cache-manager/vip-cache-manager.php
```diff
@@ -239,6 +242,18 @@ public function execute_purges() {
 		 */
 		do_action( 'wpcom_vip_cache_pre_execute_bans', $this->ban_urls );
 
+		$num_ban_urls = count( $this->ban_urls );
+		$num_purge_urls = count( $this->purge_urls );
+		if ( $num_ban_urls >= self::MAX_BAN_URLS ) {
```
joshbetz: I think there's an off-by-one error here. The condition should be `>` instead of `>=`. As is, this would generate an error if 10 URLs are provided, but all 10 would still be processed?

mjangda: Good catch!

mjangda:
r99376-deploy

PR #741, opened 2017-09-10
Don't try to block an already-blocked event

ethitter:
If the event was already blocked, either by Core or other code, `$event` won't be an object. Checking for this resolves a "trying to get property of non-object" notice.


ethitter:
r97818-deploy

PR #743, opened 2017-09-11
Bump Query Monitor to v2.14.0

ethitter:
See https://github.com/johnbillion/query-monitor/releases/tag/2.14.0


ethitter:
r97855-deploy

PR #744, opened 2017-09-17
Importer: Suspend cache

joshbetz:
Since sandboxes may not be in the origin datacenter, memcached
operations during imports might be slow. Suspending the cache and then
flush it afterwards may be faster than using the cache.


nickdaugherty:
Was about to leave a comment about `wp_suspend_cache_invalidation()` then saw you pushed it 😄

joshbetz:
@nickdaugherty Can you double check this once more before I deploy?

nickdaugherty:
Looks good to me. I was a little worried that we might be suspending cache invalidation for things like content syndication, but unless that process called the `import_start|end` hook (which seems unlikely) it'll be unaffected.

joshbetz:
r101480-deploy

PR #745, opened 2017-09-17
Adds "vip migrate clone-ms" command

emrikol:
This command runs all of the SQL/WP-CLI commands needed to when cloning a multisite on VIP Go.

Tested with a VIP Go Subfolder Multisite, no problems seen yet.

### DIFF
--- a/wp-cli/vip-migrations.php
+++ b/wp-cli/vip-migrations.php
@@ -245,6 +245,180 @@ class VIP_Go_Migrations_Command extends WPCOM_VIP_CLI_Command {
 			}
 		}
 	}
+	
+	/**
+	 * Clones a multisite to another multisite.
+	 *
+	 * "Cloning" is a misleading term here.  All of the actual cloning needs to be done outside this program
+	 * by copying the SQL from one site to another.  For VIP Go, this can be done using the `vip-cli` tool.
+	 *
+	 * ## OPTIONS
+	 *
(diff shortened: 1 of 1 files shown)


wp-cli/vip-migrations.php
```diff
@@ -245,6 +245,295 @@ function import_user_meta( $args, $assoc_args ) {
...
+	 * [--dry-run=<true>]
+	 * : Do a "dry run" and no data modification will be done.  Defaults to true.
+	 *
+	 * ## EXAMPLES
+	 *
+	 *     # Converts the data from the parent site test-subdir-ms-01.go-vip.co data to the child site test-subdir-ms-01-preprod.go-vip.co
+	 *     $ wp vip migration clone-ms test-subdir-ms-01.go-vip.co test-subdir-ms-01-preprod.go-vip.co --url=test-subdir-ms-01.go-vip.co --dry-run=false
+	 *
+	 *     # Does a "dry run" convert from "www.example.com" to "preprod.example.com"
+	 *     $ wp --allow-root vip migration clone-ms www.example.com preprod.example.com --url=www.example.com
```
david-binda: The `--url` param suggests that we can start the command for any blog in MS. Is that true? If not, I'd say that we might need "switch_to_blog()" from inside the command, or do some validation of the param.

joshbetz:
Closing in favor of #786

PR #747, opened 2017-09-20
Don't use placeholders for wpdb->prepare

mjangda:
Deprecated as of 4.8.2

h/t @soulseekah [#](https://github.com/Automattic/vip-go-mu-plugins/commit/3badb54961370724b7d24076b1feadc1c02d4bd2#commitcomment-24429023)

### DIFF
--- a/wp-cli/vip-go-convert-to-utf8mb4.php
+++ b/wp-cli/vip-go-convert-to-utf8mb4.php
@@ -326,7 +326,14 @@ class VIP_Go_Convert_To_utf8mb4 extends WPCOM_VIP_CLI_Command {
 
 			WP_CLI::line( "Converting column {$col->Field}" );
 
-			$convert = $wpdb->query( $wpdb->prepare( 'ALTER TABLE %1$s CHANGE %2$s %2$s ENUM(' . $options . ') CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci %3$s DEFAULT "%4$s"', $table, $col->Field, $null_not_null, $col->Default ) );
+			$convert = $wpdb->query( $wpdb->prepare(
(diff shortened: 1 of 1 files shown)


soulseekah:
Well, I was hoping you guys would weigh in and approve of actually supporting such placeholders in WordPress seeing how there are 1.2 million other cases of the use according to GitHub :)

But yes, fixing it in the code is one way to go.

There's a patch here already https://core.trac.wordpress.org/ticket/41925 that adds the support back in a safe way.

ethitter:
If we were using placeholders more, or could make a strong argument for why they're really needed here, I might feel differently about removing them. That said, they only snuck in here because I have a habit of including them, even when they aren't necessary.

soulseekah:
I see. So you believe that everyone should just be fixing this, the thousands of plugins and themes using this? I have never used them as far as I have grepped, but I am still very concerned about how this security vulnerability was handled. Instead of fixing it correctly, it was silently stubbed out and suddenly custom SQL around the world broke.

Yes, there's a vulnerability. But the way it was addressed was sub par. Was GitHub searched for the usage of numbered placeholders? Was the plugin repository combed through as it is customary when renaming, introducing, removing functionality from core?

What will happen if a function marked for internal use (but public) is used and the core team decides to rename it or remove it? Will they not grep the plugin and theme repo to make sure nobody's using it just in case?

It's nice that you guys can jump in and fix this within 24 hours, but admins around the world are stuck with three options:

1. either contact each plugin that broke for a fix,
2. fix it themselves until a fix is available,
3. rollback to vulnerable 4.8.1

Guess which one most will pick? :)

mjangda:
r98105-deploy

mjangda:
@soulseekah You bring up many valid points and I think it's best to continue the discussion over on the Core ticket: https://core.trac.wordpress.org/ticket/41925

ethitter:
@soulseekah No, my point was that I was sloppy in including placeholders, and since they aren't important to the query, I can't make a strong argument for or against how Core handled this change.

Your point about updating is valid, but I agree with Mo, this discussion should continue on the Core ticket.

PR #748, opened 2017-09-22
Stop notices related to a property that HyperDB provides

ethitter:
It's only an issue in local development, but it produces quite a bit of noise.


ethitter:
r98284-deploy

PR #751, opened 2017-09-29
Temporarily disable force two-step auth

joshbetz:
There was previously a bug causing this not to work. When that was fixed, it caused some confusion about why two-step authentication was suddenly being enabled. Anyone who still wants to force two-step auth can do so by enabling `jetpack-force-2fa` or `two-factor` in their site repository with the `wpcom_vip_load_plugin` helper function.


joshbetz:
r98669-deploy

PR #753, opened 2017-10-02
Load vip-secrets.php in wp-config

philipjohn:
Loading secrets is moved to wp-config, so loading here is now redundant.

### DIFF
--- a/000-vip-init.php
+++ b/000-vip-init.php
@@ -9,10 +9,6 @@
  * good reason not to.
  */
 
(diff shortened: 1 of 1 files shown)


mjangda:
@philipjohn I left a question for you in the vipd repo for some discussion.

Note: This cannot be merged until the related vipd change has been merged + deployed everywhere and all web configs reloaded.

We may also want to modify this to use `require_once` as an intermediary step to avoid fatals when we are in the process of rolling this out.

mjangda:
Closing; vipd PR was closed.

PR #754, opened 2017-10-02
Remove remaining shared plugins

scottsweb:
Once the shared plugin migrations are complete, we can remove the remainder of the shared plugins. This PR removes the following:

* chartbeat
* co-schedule
* disqus
* findthebest
* getty-images
* janrain-capture
* maintenance-mode
* multiple-post-thumbnails
* new-device-notification
* postrelease-vip
* wpcom-legacy-redirector
* vip-go-geouniques
* wp-parsely
* fieldmanager
* msm-sitemap

The constant and logic added here: https://github.com/Automattic/vip-go-mu-plugins/pull/729/files need to remain in place until we switch to Jetpack search and finalise the work on 2fa.

<strong>Potential breaking change</strong>

https://github.com/Automattic/vip-go-mu-plugins/pull/754/commits/1d4332ee027dff6021a9a4db12745d57a3b64bd9#diff-fb296e3bc7c2a02335f61196cce2f90fR1040

The file name and extension are added to the global array of code loaded plugins. This has implications for code that uses the `wpcom_vip_get_loaded_plugins` function. One instance of client use.


vip-helpers/vip-utils.php
```diff
@@ -1037,7 +1037,7 @@ function wpcom_vip_load_plugin( $plugin = false, $folder = false, $load_release_
 	}
 
 	if ( $includepath && file_exists( $includepath ) ) {
-		wpcom_vip_add_loaded_plugin( "{$plugin_type}/{$plugin}" );
+		wpcom_vip_add_loaded_plugin( "{$plugin_type}/{$plugin}/{$file}" );
```
mjangda: Why is this change needed?

scottsweb: The data we get back from `wpcom_vip_get_loaded_plugins()` is much closer to the option we store in `active_plugins` with this change. It makes comparing the two a little more accurate and easier.

I compare them here: https://github.com/Automattic/vip-dashboard/pull/75/files#diff-89ad60d19282ec7bc4917b8a060268ccR392

mjangda:
> The file name and extension are added to the global array of code loaded plugins. This has implications for code that uses the `wpcom_vip_get_loaded_plugins` function. One instance of client use.

What's the plan for handling this? Coordinate with the client before we deploy?

scottsweb:
Correct. We will need to reach out and make sure they are ready for the change. It should be a one line change.

vip-plugins/vip-plugins.php
```diff
@@ -0,0 +1,222 @@
...
+function wpcom_vip_fetch_vip_featured_plugins() {
+	$plugins = wp_cache_get( 'wpcom_vip_featured_plugins' );
+
+	if ( false === $plugins ) {
+		$plugins = array();
+		$url_for_featured_plugins = 'https://vip.wordpress.com/wp-json/vip/v1/plugins?type=technology';
+		$response = vip_safe_wp_remote_get( $url_for_featured_plugins, false, 3, 5 );
+
+		if ( ! $response || is_wp_error( $response ) ) {
+			trigger_error( 'The API on vip.wordpress.com is not responding (' . esc_url( $url_for_featured_plugins ) . ')' );
```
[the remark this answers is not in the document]
mjangda: We also need to cache an error or a non-`false` value here to avoid breaking things when the VIP site isn't available. Can cache it for a shorter period (e.g. `MINUTE_IN_SECONDS`).

vip-plugins/vip-plugins.php
```diff
@@ -0,0 +1,222 @@
...
+			return false;
+		}
+
+		$plugins = json_decode( $response['body'] );
+
+		if ( empty( $plugins ) ) {
+			return false;
```
mjangda: Should we trigger a warning here too? Something is probably broken if the response was empty. Also need caching here as well.

vip-plugins/vip-plugins.php
```diff
@@ -0,0 +1,222 @@
...
+	$screen = get_current_screen();
+
+	if ( ! ( 'plugins' === $screen->id || 'plugins-network' === $screen->id ) ) {
+		return;
+	}
+
+	$plugins = wpcom_vip_fetch_vip_featured_plugins();
+
+	if ( ! $plugins ) {
+		return;
```
mjangda: Should we show the user an error? Might be confusing if featured plugins suddenly don't show up when we run into an error.

vip-plugins/vip-plugins.php
```diff
@@ -0,0 +1,222 @@
...
+/**
+ * Render the featured partner plugins to the plugins screenReaderText
+ * Uses the notice hooks as that is all we have on these pages
+ *
+ * @return null
+ */
+function wpcom_vip_render_vip_featured_plugins() {
+	$screen = get_current_screen();
+
+	if ( ! ( 'plugins' === $screen->id || 'plugins-network' === $screen->id ) ) {
```
mjangda: Minor: can simplify the readability of this using negatives:

```
if ( 'plugins' !== $screen->id || 'plugins-network' !== $screen->id )
```

vip-plugins/vip-plugins.php
```diff
@@ -0,0 +1,222 @@
...
+				</a>
+			</div>
+			<?php
+		}
+		?>
+	</div>
+	<?php
+}
+add_action( 'admin_notices', 'wpcom_vip_render_vip_featured_plugins', 99 );
+add_action( 'network_admin_notices', 'wpcom_vip_render_vip_featured_plugins', 99 );
```
mjangda: Minor: would be good to note why we're using `admin_notices` for this and why priority 99.

vip-plugins/vip-plugins.php
```diff
@@ -0,0 +1,222 @@
...
+/**
+ * Custom CSS and JS for the plugins UIs
+ *
+ * @return null
+ */
+function wpcom_vip_plugins_ui_admin_enqueue_scripts() {
+	$screen = get_current_screen();
+	if ( 'plugins' === $screen->id || 'plugins-network' === $screen->id ) {
+		wp_enqueue_style( 'vip-plugins-style', WP_CONTENT_URL . '/mu-plugins/' . basename( __DIR__ ) . '/css/plugins-ui.css' , '3.0' );
+		wp_enqueue_script( 'vip-plugins-script', WP_CONTENT_URL . '/mu-plugins/' . basename( __DIR__ ) . '/js/plugins-ui.js', array( 'jquery' ), '3.0', true );
```
mjangda: We can simplify the URL-building: `plugins_url( 'css/plugins-ui.css', __FILE__ )`

vip-plugins/vip-plugins.php
```diff
@@ -0,0 +1,222 @@
...
+
+/**
+ * Custom CSS and JS for the plugins UIs
+ *
+ * @return null
+ */
+function wpcom_vip_plugins_ui_admin_enqueue_scripts() {
+	$screen = get_current_screen();
+	if ( 'plugins' === $screen->id || 'plugins-network' === $screen->id ) {
+		wp_enqueue_style( 'vip-plugins-style', WP_CONTENT_URL . '/mu-plugins/' . basename( __DIR__ ) . '/css/plugins-ui.css' , '3.0' );
```
mjangda: The 3rd param is dependencies so we should pass an empty array.

vip-plugins/vip-plugins.php
```diff
@@ -0,0 +1,222 @@
...
+
+/**
+ * Returns a filtered list of code activated plugins similar to core plugins Option
+ *
+ * @return array list of filtered plugins
+ */
+function wpcom_vip_get_filtered_loaded_plugins() {
+	$code_plugins = wpcom_vip_get_loaded_plugins();
+	foreach ( $code_plugins as $key => $plugin ) {
+		if ( substr( $plugin, 0, 8 ) === 'plugins/' ) {
```
mjangda: Minor: should have an inline comment explaining why we need to strip off `plugins/` from the path.

vip-plugins/vip-plugins.php
```diff
@@ -0,0 +1,222 @@
...
+/**
+ * Returns a filtered list of code activated plugins similar to network plugins option
+ *
+ * @return array list of filtered, active plugins
+ */
+function wpcom_vip_get_network_filtered_loaded_plugins() {
+	$code_plugins = wpcom_vip_get_filtered_loaded_plugins();
+	foreach ( $code_plugins as $key => $plugin ) {
+		unset( $code_plugins[ $key ] );
+		$code_plugins[ $plugin ] = filemtime( __FILE__ );
```
mjangda: Minor: should have inline comment noting why we need `filemtime` here.

vip-plugins/vip-plugins.php
```diff
@@ -0,0 +1,222 @@
...
+ *
+ * @param  array $actions
+ * @param  string $plugin_file
+ * @param  array $plugin_data
+ * @param  string $context
+ * @return array
+ */
+function wpcom_vip_plugin_action_links( $actions, $plugin_file, $plugin_data, $context ) {
+	$screen = get_current_screen();
+	if ( in_array( $plugin_file, wpcom_vip_get_filtered_loaded_plugins(), true ) ) {
```
mjangda: Minor: it's cleaner to move function calls like `wpcom_vip_get_filtered_loaded_plugins()` outside of `if` blocks and assign to a variable.

tests/test-vip-plugins.php
```diff
@@ -0,0 +1,164 @@
...
+
+		/**
+		 * Update the option with our list of active plugins
+		 */
+		update_option( 'active_plugins' , $this->option_active_plugins );
+
+		/**
+		 * Check that option is not empty
+		 */
+		$this->assertNotEmpty( get_option( 'active_plugins' ) );
```
mjangda: We should actually confirm that the value of `active_plugins` in the db matches what we expect.

tests/test-vip-plugins.php
```diff
@@ -0,0 +1,164 @@
...
+		 */
+		$merged_plugins = array_merge( wpcom_vip_get_filtered_loaded_plugins(), $this->option_active_plugins );
+		$this->assertEquals( $merged_plugins, get_option( 'active_plugins' ) );
+
+		/**
+		 * Check that updating the option is OK, add an extra plugin
+		 */
+		$plugin_change = array_merge( $this->option_active_plugins, array( 'amp-wp/amp.php' ) );
+		$option_update = update_option( 'active_plugins', $plugin_change );
+		$this->assertTrue( $option_update );
```
mjangda: We should check the value in the db as well, since we want to make sure that our code-activate plugins don't make it into the option.

tests/test-vip-plugins.php
```diff
@@ -0,0 +1,164 @@
...
+
+		/**
+		 * Update the option with our list of active network plugins
+		 */
+		update_site_option( 'active_sitewide_plugins' , $this->option_active_sitewide_plugins );
+
+		/**
+		 * Check that option is not empty
+		 */
+		$this->assertNotEmpty( get_site_option( 'active_sitewide_plugins' ) );
```
mjangda: Same as the single-site test, we should check the value in the db.

tests/test-vip-plugins.php
```diff
@@ -0,0 +1,164 @@
...
+		 */
+		$merged_plugins = array_merge( wpcom_vip_get_network_filtered_loaded_plugins(), $this->option_active_sitewide_plugins );
+		$this->assertEquals( $merged_plugins, get_site_option( 'active_sitewide_plugins' ) );
+
+		/**
+		 * Check that updating the option is OK, add an extra plugin
+		 */
+		$plugin_change = array_merge( $this->option_active_sitewide_plugins, array( 'amp-wp/amp.php' => 1507904134 ) );
+		$option_update = update_site_option( 'active_sitewide_plugins', $plugin_change );
+		$this->assertTrue( $option_update );
```
mjangda: Same as the single-site test, we should check the value in the db.

tests/test-vip-plugins.php
```diff
@@ -0,0 +1,164 @@
...
+
+		/**
+		 * Check that the option still makes sense again
+		 */
+		// emulates update
+		$track_plugin_change = array_diff( $plugin_change, wpcom_vip_get_network_filtered_loaded_plugins() );
+		// emulates get
+		$saved_plugins = array_merge( wpcom_vip_get_network_filtered_loaded_plugins(), $track_plugin_change );
+		$this->assertEquals( $saved_plugins, get_site_option( 'active_sitewide_plugins' ) );
```
mjangda: Any time we have multiple `assert*` calls in a single method, we should pass in the `message` parameter (with a unique string) to help us pinpoint where our test failed. Otherwise when this test fails, all we'll see is `array( 4, 5, 6 ) did not match array( 1, 2, 3 )` which doesn't really tell us what's wrong.

So, we can do something like this:

```
$this->assertEquals( $saved_plugins, get_site_option( 'active_sitewide_plugins' ), 'The value of the `active_sitewide_plugins` option did not match our expected `$saved_plugins`' );
```

This way, we can pinpoint that line 162 is where things failed.

tests/test-vip-plugins.php
```diff
@@ -0,0 +1,164 @@
...
+		if ( ! is_multisite() ) {
+			$this->markTestSkipped( 'Not relevant on single-site' );
+			return;
+		}
+
+		/**
+		 * Clear global of code activated plugins from previous test
+		 */
+		global $vip_loaded_plugins;
+		$vip_loaded_plugins = array();
```
mjangda: This should go in `setUp`. We can probably also move the empty array update for both tests into `setUp` as well.

vip-plugins/vip-plugins.php
scottsweb: Added to the function notes

vip-plugins/vip-plugins.php
scottsweb: Good call, didn't realised this worked for mu-plugins too

vip-plugins/vip-plugins.php
scottsweb: Could not get this logic to work

vip-plugins/vip-plugins.php
mjangda: FWIW, I actually find it's best to have details like this as close to the source as possible. If I'm looking through the code and trying to figure out why priority 99 is being used, my instinct is not to go the callback's docblock to get those details.

Additionally, if we change the priority or the action down the road, it's easier to forget to update the callback docblock since it's not right there, and we end up with outdated inline docs.

tests/test-vip-plugins.php
```diff
@@ -153,12 +171,20 @@ public function test__modify_network_active_plugins() {
 		$this->assertTrue( $option_update );
 
 		/**
+		 * Check the raw value in the DB matches what we sent above - skips any filters
+		 */
+		$option_db = $wpdb->get_row( $wpdb->prepare( "SELECT meta_value FROM $wpdb->sitemeta WHERE meta_key = %s LIMIT 1", 'active_sitewide_plugins' ) );
```
mjangda: Minor: can simplify this using `get_var` which will return a string value instead of an object. Also, `wpdb->prepare` isn't strictly necessary here. We could just hard-code the meta_key inside the query.

vip-plugins/vip-plugins.php
```diff
@@ -200,7 +202,9 @@ function wpcom_vip_site_option_active_sitewide_plugins( $value, $option ) {
 		$value = array_merge( $value, $code_plugins );
 	}
 
-	ksort( $value );
+	if ( ! empty( $value ) ) {
```
mjangda: Do we need to do the same thing in `wpcom_vip_pre_update_option_active_plugins` and `wpcom_vip_pre_update_site_option_active_sitewide_plugins`?

scottsweb: Good thinking. Better to be safe I think.

vip-plugins/vip-plugins.php
simonwheatley: Should this be:

``` php
if ( 'plugins' !== $screen->id && 'plugins-network' !== $screen->id )
```

mjangda:
Part 1: UI and functional bits in r99783-deploy
Part 2: plugin removals in r99784-deploy

mjangda:
One more change in r99786-deploy

PR #756, opened 2017-10-11
Fix replacement for wpcom_vip_enable_opengraph

mjangda:
The replacement is a filter, not a function.

props @paulschreiber.


mjangda:
r99115-deploy

PR #758, opened 2017-10-11
Fix confusing code comment

### LINKED ISSUE #757: WPCOM_IS_VIP_ENV defined as false; comments indicate it's defined as true
In `000-vip-init.php`, WPCOM_IS_VIP_ENV defined as false; comments indicate it's defined as true:

```
// For backwards compatibility - always true.
if ( ! defined( 'WPCOM_IS_VIP_ENV' ) ) {
}
```

I would expect the comment to reflect the actual value.

joshbetz:
Fixes #757

props @simonwheatley | h/t @paulschrieber

### DIFF
--- a/000-vip-init.php
+++ b/000-vip-init.php
@@ -27,7 +27,7 @@ if ( ! defined( 'VIP_GO_ENV' ) ) {
 	define( 'VIP_GO_ENV', false );
 }
 
-// For backwards compatibility - always true.
+// For backwards compatibility - always true on a VIP Go environment.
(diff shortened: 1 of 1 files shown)


000-vip-init.php
```diff
@@ -27,7 +27,7 @@
 	define( 'VIP_GO_ENV', false );
 }
 
-// For backwards compatibility - always true.
+// For backwards compatibility - always true on a VIP Go environment.
```
mjangda: The constant is pretty heavily used so I think we should probably remove the part about back compat.

mjangda:
r99375-deploy

PR #759, opened 2017-10-11
Updates for wpcom_vip_file_get_contents

emrikol:
Brings in changes from D6805 on WPCOM that fixes a bug where the "backup" cache fails when the remote server is down.

This function needs some PHPCS/PHPCBF fixes, but should probably be done separately from these specific bugfixes.

### DIFF
--- a/vip-helpers/vip-utils.php
+++ b/vip-helpers/vip-utils.php
@@ -567,7 +567,7 @@ function wpcom_vip_file_get_contents( $url, $timeout = 3, $cache_time = 900, $ex
 	// Okay, it wasn't successful. Perhaps we have a backup result from earlier.
 	elseif ( $content = wp_cache_get( $backup_key, $cache_group ) ) {
         //We found a backup copy,
-        if ( $server_up ){ //If we thought the server was up, let's store that it's not up for 60 seconds. We only do this if we thought the server was up so that we do retry every 1 minute.
+        if ( $server_up ) { //If we thought the server was up, let's store that it's not up for 60 seconds. We only do this if we thought the server was up so that we do retry every 1 minute.
 	        wp_cache_set( $disable_get_key, 1, $cache_group, 60 );
         }
 
--- a/vip-helpers/vip-utils.php
+++ b/vip-helpers/vip-utils.php
@@ -472,7 +472,7 @@ function vip_regex_redirects( $vip_redirects_array = array(), $with_querystring 
  *
  * We strongly recommend not using a $timeout value of more than 3 seconds as this
  * function makes blocking requests (stops page generation and waits for the response).
- *
+ * 
  * The $extra_args are:
  *  * obey_cache_control_header: uses the "cache-control" "max-age" value if greater than $cache_time.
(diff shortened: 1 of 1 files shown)


vip-helpers/vip-utils.php
```diff
@@ -569,46 +556,41 @@ function wpcom_vip_file_get_contents( $url, $timeout = 3, $cache_time = 900, $ex
...
-		wp_cache_set( $backup_key, $content, $cache_group );
+		wp_cache_add( $backup_key, $content, $cache_group );
 
 		// So we can hook in other places and do stuff
 		do_action( 'wpcom_vip_remote_request_success', $url, $response );
 	}
 	// Okay, it wasn't successful. Perhaps we have a backup result from earlier.
 	elseif ( $content = wp_cache_get( $backup_key, $cache_group ) ) {
+        //We found a backup copy,
+        if ( $server_up ){ //If we thought the server was up, let's store that it's not up for 60 seconds. We only do this if we thought the server was up so that we do retry every 1 minute.
```
mjangda: Minor: some spacing weirdness here that we should fix.

vip-helpers/vip-utils.php
```diff
@@ -495,28 +495,19 @@ function wpcom_vip_file_get_contents( $url, $timeout = 3, $cache_time = 900, $ex
...
 
-	// Legacy
-	if ( false !== $cache = wp_cache_get( $old_cache_key, $cache_group) )
-		return $cache;
-
 	// The timeout can be 1 to 10 seconds, we strongly recommend no more than 3 seconds
 	$timeout = min( 10, max( 1, (int) $timeout ) );
 
-	if ( $timeout > 3 )
+	if ( $timeout > 3 && ! is_admin() )
```
mjangda: The use of `is_admin()` inside this function doesn't feel right.

vip-helpers/vip-utils.php
```diff
@@ -495,28 +495,19 @@ function wpcom_vip_file_get_contents( $url, $timeout = 3, $cache_time = 900, $ex
 	$extra_args = wp_parse_args( $extra_args, $extra_args_defaults );
 
 	$cache_key       = md5( serialize( array_merge( $extra_args, array( 'url' => $url ) ) ) );
-	$backup_key      = $cache_key . '_backup';
-	$disable_get_key = $cache_key . '_disable';
+	$backup_key      = md5( $cache_key . '_backup' );
+	$disable_get_key = md5( $cache_key . '_disable' );
```
mjangda: Not totally clear why we need to md5 an already mostly md5 key? Plus, this makes debugging harder if we ever need to look up keys.

joshbetz:
Pinging @emrikol on the rest of the feedback

mjangda:
Re-ping @emrikol.

emrikol:
Ugh, honestly we should probably just rewrite this entire function.  It's kind of a mess.

emrikol:
This is so old, I'm going to close it out.  At this point, we should probably just consider Go the canonical source for the function and not worry about WPCOM unless we need to.

PR #762, opened 2017-10-12
Queue async transitions for posts force-published by Cron Control

ethitter:
Cron Control includes a job that runs every minute, force-publishing posts that missed their scheduled times. Since the async transitions code excludes the CLI context (#471), this manually queues the events.


001-cron.php
```diff
@@ -112,6 +112,13 @@ function wpcom_vip_cron_control_event_object_to_string( $event ) {
 }
 
 /**
+ * When Cron Control's Internal Event force-publishes posts that missed their scheduled time, enable async transitions
+ */
+function wpcom_vip_enable_async_actions_for_forced_posts( $pid ) {
+	Automattic\VIP\Async_Publish_Actions\_queue_async_hooks( 'publish', 'future', get_post( $pid ) );
```
mjangda: We should probably have a sanity check here to make sure that the method exists, or switch to firing an action instead.

ethitter: I was probably trying to keep all of the Cron Control stuff together, but this could just move to https://github.com/Automattic/vip-go-mu-plugins/blob/master/async-publish-actions.php.

WPprodigy:
Needs adjustments post-https://github.com/Automattic/vip-go-mu-plugins/pull/1999.

Of note, there's a good possibility of this module being deprecated and moved away from.

WPprodigy:
Confirming that we will be deprecating/removing the async publish posts module very soon.

Going to leave this PR open for a little in case I'll be migrating the plugin to an individual customer-repo where this code would be profitable.

PR #763, opened 2017-10-13
Bump VIP Support to latest

ethitter:
* Fixes incorrect a11n promotion, see https://github.com/Automattic/vip-support/pull/56
* Ensures Core constants are respected, see https://github.com/Automattic/vip-support/pull/53


ethitter:
r99227-deploy

PR #765, opened 2017-10-13
Add ability to specify TTL in wpcom_vip_wp_oembed_get

sboisvert:
matches r316138-wpcom
limit of 1 month and minimum of 5h

### DIFF
--- a/vip-helpers/vip-utils.php
+++ b/vip-helpers/vip-utils.php
@@ -910,17 +910,16 @@ function wpcom_vip_bulk_user_management_whitelist( $users ) {
  *
  * @param string $url The URL that should be embedded
  * @param array $args Addtional arguments and parameters the embed
- * @param int $ttl How long to cache for in seconds , minimum 18000 (5 Hours)
+ * @param int $ttl How long to cache for in seconds; minimum 18000 (5 Hours)
  * @return string
  */
 function wpcom_vip_wp_oembed_get( $url, $args = array(), $ttl = false ) {
-
-	//Set a Minimum of 5H (1800seconds)
-	//let's get a bit of variation in here to prevent stampedes.
-	//Set a max of 30 days after which memcache thinks you're giving it a timestamp.
-	if ( $ttl && $ttl > 5 * HOUR_IN_SECONDS && ( $ttl < MONTH_IN_SECONDS ) ){
+	// We want a min ttl of 5 hours (1800s).
+	// And a max of 30 days (after which memcache thinks you're giving it a timestamp).
+	// Let's also add a bit of variation to prevent stampedes.
+	if ( $ttl && $ttl > 5 * HOUR_IN_SECONDS && ( $ttl < MONTH_IN_SECONDS ) ) {
 		$ttl = $ttl + rand( 0, HOUR_IN_SECONDS );
-	}else{
+	} else {
 		$ttl = rand( 5 * HOUR_IN_SECONDS, 6 * HOUR_IN_SECONDS );
 	}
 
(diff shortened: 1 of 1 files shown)


joshbetz:
`MONTH_IN_MINUTES` is not part of core, so it would need to be added.

joshbetz:
`MONTH_IN_SECONDS` is part of core though and seems like that's actually what you want here.

vip-helpers/vip-utils.php
```diff
@@ -910,15 +910,25 @@ function wpcom_vip_bulk_user_management_whitelist( $users ) {
...
+ * @param int $ttl How long to cache for in seconds , minimum 18000 (5 Hours)
  * @return string
  */
-function wpcom_vip_wp_oembed_get( $url, $args = array() ) {
+function wpcom_vip_wp_oembed_get( $url, $args = array(), $ttl = false ) {
+
+	//Set a Minimum of 5H (1800seconds)
+	//let's get a bit of variation in here to prevent stampedes.
+	//Set a max of 30 days after which memcache thinks you're giving it a timestamp.
+	if ( $ttl && $ttl > 5 * HOUR_IN_SECONDS && ( $ttl < (30 * MONTH_IN_MINUTES * 60) ) ){
```
joshbetz: Your upper bound here is 30 months.

christianc1:
Hey, here from ticket 70562.  Thanks for this PR! 

One suggestion: While the random time pad makes sense in our use case, you might want to consider adding a fourth parameter `bool $rand_pad` as a flag.  I'm thinking of the broader VIP community here, and also of the many cases where I expect the cache to expire a value precisely when I say it should be expired 😈 .  Documentation could explain it's purpose, and CR's could mandate it if there is a risk of stampedes.

Cheers!

sboisvert:
Hey @christianc1, 

Thanks for the feedback! What we've noticed is that the risk of Stampede is far greater than usual because what tends to happen is that people have posts with lots of embeds. Sometimes up to 50, even with just a few, spreading out the remote requests is really useful to keep response times reasonable. 

>many cases where I expect the cache to expire a value precisely when I say it should be expired 

Can you walk me thru a use case for this where 5h as a minimum is acceptable but having a 0-60min padding isn't? I'm having a hard time coming up with one :)

mjangda:
@sboisvert For future PRs, let's try to include more detail in the PR description and commit messages, please.

For example, in the PR and the first commit, we say the following:

> Add ability to specify TTL
> matches r316138-wpcom
> limit of 1 month and minimum of 5h

This doesn't really tell reviewers for what feature we're adding the TTL customization to or why.

And then in the next commit, we say:

> Update vip-utils.php

This doesn't tell us anything useful at all about what's changing or why.

Here's a great post that covers how to write good, useful commit messages: https://robots.thoughtbot.com/5-useful-tips-for-a-better-commit-message

christianc1:
@sboisvert I guess I am also having trouble coming up with a potential scenario, and even if I were able, it would be purely contrived.  When I said "many cases" I was broadly referring to all cache invalidation, not necessarily just this case.  So probably you're right and it's better to address the issues you've currently identified and address future areas to improve if/when they arise. 👍

mjangda:
@sboisvert I pushed a minor whitespace + comment cleanup but also a fix for an edge case (where we can set the ttl to `MONTH_IN_SECONDS - 1` or `MONTH_IN_SECONDS - HOUR_IN_SECONDS + 1` (or anywhere in between) and end up passing the max.

Let me know what you think.

sboisvert:
@mjangda Looks great! 
cc @joshbetz

joshbetz:
r101482-deploy

PR #766, opened 2017-10-13
Add PHPCS on PR

joshbetz:
Started in #520, this is a somewhat weak check (only reports errors above priority 5), but currently passes and gives us a starting point.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,14 +15,14 @@ phpcs:
 	test -f /tmp/phpcs || curl -L https://squizlabs.github.io/PHP_CodeSniffer/phpcs.phar -o /tmp/phpcs && chmod +x /tmp/phpcs
 	test -d /tmp/wpcs || git clone -b master --depth 1 https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards.git /tmp/wpcs
 	/tmp/phpcs --config-set installed_paths /tmp/wpcs
-	/tmp/phpcs -p . --severity=6 --standard=WordPress-VIP --extensions=php --runtime-set ignore_warnings_on_exit true
+	/tmp/phpcs -p . --severity=6 --standard=phpcs.xml --extensions=php --runtime-set ignore_warnings_on_exit true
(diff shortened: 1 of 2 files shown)


joshbetz:
We could actually exclude any changes here if there are concerns about in-progress PRs.

mjangda:
> We could actually exclude any changes here if there are concerns about in-progress PRs.

It's annoying but probably easier just to rip the bandaid and then fix up the PRs :)

joshbetz:
@mjangda I think there are basically 3 ways to move this forward:

1. Revert the phpcbf changes and re-apply them later, per directory, depending on in-progress PRs, and lower the severity filter over time.
2. Merge as is, lower the severity filter over time.
3. Remove the severity filter now and fix everything, ripping off the band-aid as you say.

After thinking about this a bit, 2 seems suboptimal because we "rip the band-aid off" now and again when we lower the severity. 3 will be painful now, but it would be nice to have this done. 1 will prolong the process, but if we remember to keep checking on this, it could be the easiest.

I'm leaning toward 1, but don't feel super strongly.

lib/proxy/ip-utils.php
```diff
@@ -16,121 +16,117 @@
  *
  * @author Fabien Potencier <fabien@symfony.com>
  */
-class IpUtils
```
mjangda: We should probably exclude this lib and the related test file (`tests/test-lib-proxy-ip-utils.php`) since it's all taken from a 3rd party lib.

mjangda:
Yeah, `1` might be better given that we do have quite a few PRs open/pending/in-development.

GaryJones:
@joshbetz Note that this uses WPCS's `WordPress-VIP` standard, instead of VIPCS's `WordPress-VIP-Go` standard. They are very different. The former is deprecated and will be removed. The latter is under active development.

nickdaugherty:
Opened an issue (#1032) to switch over to the new rules.

PR #767, opened 2017-10-18
Add x-powered-by header

mjangda:
hackers_hacktheplanet

Also switched to use the `wp_headers` filter instead since it's cleaner and easier to remove the header if ever necessary.


ethitter:
Don't some internal tools (bot for newly-launched sites) check for the `X-Hacker` header? What about monitoring from Systems?

simonwheatley:
@ethitter Hello, stranger 🙂 We're looking into these concerns, and I recall you mentioning this in the past. We can switch the VIP tools relatively easily, and we'll make sure we don't cause issues with Systems level monitoring.

mjangda:
r99730-deploy

PR #769, opened 2017-10-20
Don't purge non-public taxonomies

mjangda:
As it just wastes resources. The URLs for terms in these taxonomies are not publicly accessible so we end up triggering purge requests for URLs that always 404.


mjangda:
r99677-deploy

PR #770, opened 2017-10-22
Fix for "undefined index" notice in jetpack.php

philipjohn:
As `$_REQUEST` doesn't always have an `action` key, the logging code
would trigger an error, which is visibly annoying in local development
environments.

This change takes that into account while also make the code more
readable by breaking it out onto multiple lines.


mjangda:
r100009-deploy

PR #771, opened 2017-10-23
Add check for WPCOM_VIP_DISABLE_SHARED_PLUGINS to /plugins api endpoint

scottsweb:
For quickly checking if any sites are missing the constant.


mjangda:
r99676-deploy

PR #773, opened 2017-10-25
Hotfix: Correctly handle Akismet API response order

mjangda:
Swap the order of the response processing for get_jetpack_user to match the order of the API calls.

This fix is being reviewed for the next version of Akismet but we're pushing through now so we can make use of the underlying code.

See `pZbK5-54x`


mjangda:
r99845-deploy

PR #774, opened 2017-10-26
Move two factor plugins

mjangda:
From `shared-plugins` to `wpcom-vip-two-factor` subfolder.

Background: https://vip.wordpress.com/documentation/vip-go/deprecating-shared-plugins-on-vip-go/

- [x] Move plugins
- [x] Update references in mu-plugins
- [ ] Update references in client code

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -25,12 +25,6 @@
 [submodule "debug-bar-cron"]
 	path = debug-bar-cron
 	url = https://github.com/tollmanz/debug-bar-cron.git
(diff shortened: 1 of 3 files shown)


joshbetz:
We'll need to deploy this in two parts. Adding the new code, then updating references in client code, and finally removing the old versions.

mjangda:
Still worth doing but probably better to just start fresh and move the plugins one at a time to reduce room for 💥 .

rinatkhaziev:
Closing this as it's stale and just a bunch of conflicts at this point.

PR #775, opened 2017-10-26
Require a flag when running `jetpack-start cancel`

mjangda:
As an added method of confirmation. `cancel` should rarely ever be used with JPSv2.


joshbetz:
We could add a prompt if the user doesn't pass `--confirm` https://make.wordpress.org/cli/handbook/internal-api/wp-cli-confirm/

mjangda:
Yeah, we could use the built-in confirm functionality but it's pretty easy to gloss over and just hit `y` and move on. I wanted the confirm bypass to require a bit more thought.

mjangda:
r100010-deploy

PR #778, opened 2017-11-01
Akismet: hotfix for API ordering quirk

mjangda:
This was breaking the register-via-Jetpack support

Upstream fix and details at https://plugins.trac.wordpress.org/changeset/1756297 (props @cfinke)

Previously: #773


mjangda:
r100203-deploy

PR #779, opened 2017-11-01
Less noise from vip-support

mjangda:
Only log removals if there were any.

Bumps the submodule to https://github.com/Automattic/vip-support/pull/57


mjangda:
r100206-deploy

PR #781, opened 2017-11-03
Fix shared plugin reporting

scottsweb:
Now the shared plugins UI is retired we are no longer adding the shared plugin information to our custom API endpoint. Whilst we still have 3 remaining shared plugins it will be useful if this continues to work. This re-instates the functionality.


joshbetz:
r101537-deploy

PR #783, opened 2017-11-03
Add an early exit if site is blocked

nickdaugherty:
Used during times of maintenance when the entire site must be blocked

Leaves cache healthchecks returning 200s (with slightly different
message) for quick recovery / no false positives


000-vip-init.php
```diff
@@ -9,6 +9,24 @@
...
+		}
+
+		http_response_code( 200 );
+
+		die( 'ok (site blocked)' );
+	}
+
+	http_response_code( 503 );
+
+	die( 'Site is currently under maintenance' );
```
mjangda: We should probably have a nice version of the maintenance page like this one: https://github.com/Automattic/vip-go-mu-plugins/blob/master/errors/alloptions-limit.html

joshbetz:
Looks good. Might be worth adding a comment explaining that this is for internal use only. Since the constant will be defined in wp-config, if a site repo defines the constant again, it will fatal.

mjangda:
> if a site repo defines the constant again, it will fatal

We can just force it to false in wp-config, so sites can't override it.

000-vip-init.php
```diff
@@ -9,6 +9,24 @@
  * good reason not to.
  */
 
+if ( defined( 'VIP_GO_SITE_BLOCKED' ) && VIP_GO_SITE_BLOCKED ) {
+	if ( '/cache-healthcheck?' === $_SERVER['REQUEST_URI'] ) {
```
mjangda: Can we avoid the duplication here somehow?

nickdaugherty: The tricky thing there is the 'regular' code comes from `WPCOM_VIP_Cache_Manager::__construct()`, which gets called when `vip-cache-manager.php` is included.

We could add our code after that mu-plugin gets loaded, the only downside is it's fairly 'late' (lots of other setup happens in between) and we'd essentially need a way to ensure our new code was loaded after it, something like `zzz-vip-init.php`.

nickdaugherty: Maybe we can just move the healthcheck out of the cache manager plugin, and into the top of `001-vip-init.php`?

mjangda: Moving the whole check probably makes sense. There was some initial work on that here: https://github.com/Automattic/vip-go-mu-plugins/pull/386 (although, that changes some of the logic as well which we can ignore for the purposes of this PR).

000-vip-init.php
nickdaugherty: Good call 👍

errors/site-blocked.html
```diff
@@ -0,0 +1,166 @@
...
+		<meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1" />
+		<meta name="viewport" content="width=device-width, initial-scale=1.0" />
+
+		<title>WordPress.com VIP - Error</title>
+
+		<link href='https://fonts.googleapis.com/css?family=Open+Sans:400,300' rel='stylesheet' type='text/css'>
+
+		<script type="text/javascript" src="https://stats.wp.com/wpcom.js?22"></script>
+		<script type="text/javascript">
+			_error = 'big-alloptions';
```
mjangda: Minor: `s/big-alloptions/site-blocked/`

nickdaugherty: Dang, good catch! Fixed.

nickdaugherty:
@mjangda / @joshbetz what do you think about _not_ exiting early on CLI commands? 

One one hand, we could end up with cron continuing to run even when a site is in the 'blocked' state.

On the other hand, we probably need to run some CLI commands (like `wp cache flush`) on sites when they are in this mode.

mjangda:
Oh, that's a good point. We could show a warning on sandboxes when someone tries to use wp-cli and the site is blocked?

nickdaugherty:
Yeah we could definitely do that. 

I think we'll need to allow WP CLI but bypass cron when we've blocked the site, and in a way that does not 'fail' the cron triggers (to prevent false positive warnings).

nickdaugherty:
To stop cron on Go, we can pre-filter the `Automattic\WP\Cron_Control::DISABLE_RUN_OPTION` option (`a8c_cron_control_disable_run`) option, which will cause the Golang runner to stop triggering, as it uses the `cron-control orchestrate` CLI command to check status.

https://github.com/Automattic/Cron-Control/blob/8d8aea4929ebb8fee55caa24dba1ebc342f97bec/includes/class-events.php#L510-L519

vip-cache-manager/vip-cache-manager.php
```diff
@@ -27,15 +27,6 @@ static public function instance() {
 	}
 
 	public function __construct() {
-		// Execute the healthcheck as quickly as possible
```
joshbetz: Should we add a note that the health check endpoint is defined somewhere else?

nickdaugherty: Either way - I probably wouldn't be looking in this class for the healthcheck endpoint first (it wouldn't be expected to me that it's in here). We know it was in here b/c historical reasons, but I dunno if future us or others would be expecting to see it here.

joshbetz: The reason I think I’d look here is I know the health checks are in varnish.

nickdaugherty:
r101491-deploy

PR #784, opened 2017-11-08
Only enable the VIP Cache Manager in VIP environments.

fklein-lu:
The Cache Manager breaks unit tests, and is useless on local.

By loading the files, the API is still publicly exposed. This avoids breaking code that uses these functions are extends them.

However the initialisation only happens in VIP environments.

### DIFF
--- a/vip-cache-manager.php
+++ b/vip-cache-manager.php
@@ -9,4 +9,10 @@ License: GPL version 2 or later - http://www.gnu.org/licenses/old-licenses/gpl-2
 */
 
 require_once( __DIR__ . '/vip-cache-manager/vip-cache-manager.php' );
+require_once( __DIR__ . '/vip-cache-manager/api.php' );
 require_once( __DIR__ . '/vip-cache-manager/ttl-manager.php' );
+
+if ( defined( 'WPCOM_IS_VIP_ENV' ) && WPCOM_IS_VIP_ENV === true ) {
+	WPCOM_VIP_Cache_Manager::instance();
+	Automattic\VIP\Cache\TTL_Manager\init();
+}
--- a/vip-cache-manager/ttl-manager.php
+++ b/vip-cache-manager/ttl-manager.php
@@ -55,6 +55,3 @@ function enforce_rest_api_read_ttl( $response, $rest_server, $request ) {
 function set_rest_response_ttl( \WP_REST_Response $response, $ttl ) {
 	$response->header( 'Cache-Control', sprintf( 'max-age=%d', $ttl ) );
 }
-
-// Let's do it!
-init();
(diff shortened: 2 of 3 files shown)


joshbetz:
Do you have more details about how this breaks unit tests?

fklein-lu:
Fails here https://github.com/Automattic/vip-go-mu-plugins/blob/master/vip-cache-manager/vip-cache-manager.php#L256  because tests modify a ton of data.

joshbetz:
Maybe we could stop `execute_purges()` from executing during tests instead, or even filter the purge URLs in the PHPUnit shutdown to avoid triggering that error.

fklein-lu:
This PR was just meant to fix the immediate issue. Your proposition would also just fix that.

But this issue is a sign of an underlying architectural issue. There is a lot that leaves to be desired in how the VIP Go MU Plugins are structured.

The main issue here is that these plugins, just like Jetpack or other plugins, needs a development mode. There is no point in having code run on local that depends on infrastructure components. 

Especially if there isn't even a local development setup specifically designed for VIP Go. As par the documentation, any old LEMP stack can be used to develop VIP Go sites locally. And the plugins used during development should reflect that.

joshbetz:
Our goal is to leave as much of the plugin intact as possible. It's true that you don't actually need to purge caches in a local environment, but it's useful to see how your code interacts with all the code that's running in production.

fklein-lu:
Thanks for the fix, even if it's not the approach I would have recommended.

"[S]ee[ing] how your code interacts with all the code that's running in production" does not seem to be a valid argument. The cache purging doesn't work, the CSS and JS concatenation doesn't work, Jetpack doesn't work, that leaves not much code to be worried about.

As said I consider that the overall approach to how the MU Plugins are set up is wrong. This code is running (successfully) in production, which is already a lot. So I'm disappointed that there seems to be no desire to take that extra step to make it actually good.

This fix is dealing with a symptom, not the cause. Define this random, undocumented constant, and then add this random check which bails in the middle of a method, without any inline documentation of course, and done deal.

Now it's up to me, as a client developer, to find out how all this works together to achieve what I need for my tests to pass. It's intransparent, and a horrible developer experience. People will _not_ understand it.

This approach seems to be part of this Automattic book of beloved anti-patterns: multiple classes and functions mixed in a file, auto-initialising classes on file inclusion, making everything under the sun a class, which is of course a singleton, or just contains static methods, and files that include files, which include other files, until somehow everything has been loaded, and actually starts working.

Why can't we just have the plugins, that have a main plugin file that initialises the plugin when a specific hook is called. Along with the plugins we have a collection of libraries for other APIs.

Then we have a single bootstrap file, instead of a bunch of them like `vip-dashboard.php` and `http-concat.php`. This file loads all the required files for libraries and plugins, according to the current environment.

This makes it very transparent what plugin runs in which environment. Plus it gives me as a developer all the flexibility I need: I can unhook plugin initialisation if I need it, and I can run plugins that aren't active on a specific environment by just including the main plugin file.

Plus by not having environment checks inside of methods and functions, I can test these easily, without messing around with constants, or needing to set global state so that things like `is_main_site()` work.

PR #786, opened 2017-11-10
Initial `vip migration cleanup` command

joshbetz:


emrikol:
Is this only for prod->non-prod syncing, or also for new imports?  If for new imports also, I think we should also make sure we get rid of any DB transients.  I've seen quite a large number of them come in with imports in the past.

joshbetz:
We need to be sure we pass the `--url` to child commands.

wp-cli/vip-migrations.php
```diff
@@ -4,6 +4,113 @@
...
+				"DELETE FROM $wpdb->options
+				WHERE option_name LIKE '\_transient\_%'
+				OR option_name LIKE '\_site\_transient\_%'"
+			);
+		}
+
+		do_action( 'vip_go_migration_cleanup', $dry_run );
+
+		// Flush the cache unless we're cleaning up the entire network
+		if ( ! $network ) {
```
nickdaugherty: Small note - this conditional may not be necessary due to the prior `if ( $network ) { ... return; }`.

joshbetz: We call the same command again, but with a `--url` specified. The goal is to avoid flushing the cache N times for multisites.

joshbetz: Although now I'm realizing this will always be false 🤔

joshbetz:
@nickdaugherty Can you sanity check that last change? We could add complex logic to avoid N cache flushes, but given that they're all going to be in quick succession, the 2nd, 3rd, Nth flushes will have pretty minimal impact.

nickdaugherty:
Looks good to me 👍

joshbetz:
r101651-deploy

PR #787, opened 2017-11-13
Update Cron Control to add fix for better handling of 'job not found' errors

nickdaugherty:
When a job is not found, it’s not _really_ an error because another runner may have picked it up.


nickdaugherty:
r100874-deploy

PR #789, opened 2017-11-14
Add Jetpack to list of $vip_loaded_plugins

### LINKED ISSUE #785: Filter jetpack into the active_plugins option
Jetpack is always enabled...so if a client is using a local version of Jetpack from `plugins`, we need to filter it into the `active_plugins` option like we do for everything loaded via `wpcom_vip_load_plugin()`, otherwise JP looks inactive which could lead to people attempting to manually activate it via the UI.

Probably should filter it in there for 'normal' mu-plugin usages of JP too (no local version).

nickdaugherty:
Because it needs to be part of the active_plugins filter, so that the
Plugins UIs (regular and Network) always show it as active and enabled
via code

Fixes #785

### DIFF
--- a/jetpack.php
+++ b/jetpack.php
@@ -79,6 +79,19 @@ if ( defined( 'WPCOM_VIP_JETPACK_LOCAL' ) && WPCOM_VIP_JETPACK_LOCAL ) {
 	}
 }
 
+// JP should always be marked as active, even if WPCOM_VIP_JETPACK_LOCAL is not set, b/c site could have it in /plugins
+// but not have it enabled via WPCOM_VIP_JETPACK_LOCAL, which would leave it open to manual activation/deactivation
+global $vip_loaded_plugins;
+
+if ( ! is_array( $vip_loaded_plugins ) ) {
+	$vip_loaded_plugins = array();
+}
+
+// Must include `plugins/` on it here, b/c wpcom_vip_get_filtered_loaded_plugins() filters out without it
+if ( ! in_array( 'plugins/jetpack/jetpack.php', $vip_loaded_plugins, true ) ) {
+	$vip_loaded_plugins[] = 'plugins/jetpack/jetpack.php';
+}
+
 require_once( $jetpack_to_load );
 
 require_once( __DIR__ . '/vip-jetpack/vip-jetpack.php' );


nickdaugherty:
cc @scottsweb as you've worked on plugins a lot recently.

jetpack.php
```diff
@@ -79,6 +79,19 @@ function wpcom_vip_log_updating_jetpack_version_option( $option_name, $option_va
...
+// but not have it enabled via WPCOM_VIP_JETPACK_LOCAL, which would leave it open to manual activation/deactivation
+global $vip_loaded_plugins;
+
+if ( ! is_array( $vip_loaded_plugins ) ) {
+	$vip_loaded_plugins = array();
+}
+
+// Must include `plugins/` on it here, b/c wpcom_vip_get_filtered_loaded_plugins() filters out without it
+if ( ! in_array( 'plugins/jetpack/jetpack.php', $vip_loaded_plugins, true ) ) {
+	$vip_loaded_plugins[] = 'plugins/jetpack/jetpack.php';
```
scottsweb: Do you think we should do a `file_exists` check on this as well to make sure Jetpack is actually in the plugins folder before adding it to the array? On my local test site without Jetpack in the plugins folder I get an error on the plugins screen:

`The plugin jetpack/jetpack.php has been deactivated due to an error: Plugin file does not exist.`

I also wonder if we should check for the presence of `WPCOM_VIP_JETPACK_LOCAL` before running any of this code.

scottsweb:
@simonwheatley is also suggesting a possible process change and recommending this goes in `client-mu-plugins` which would need to be considered.

jetpack.php
nickdaugherty: Ah, then yeah we should wrap it in `file_exists()` to prevent that warning.

> I also wonder if we should check for the presence of WPCOM_VIP_JETPACK_LOCAL before running any of this code.

I tried to explain this in my inline comment on the code:

```
+// JP should always be marked as active, even if WPCOM_VIP_JETPACK_LOCAL is not set, b/c site could have it in /plugins
 +// but not have it enabled via WPCOM_VIP_JETPACK_LOCAL, which would leave it open to manual activation/deactivation
```

Basically if you add the plugin, but don't set the constant, it'll still show as inactive and allow manual activation.

scottsweb: > Basically if you add the plugin, but don't set the constant, it'll still show as inactive and allow manual activation.

Ah of course! As you were :+1:

nickdaugherty:
I think moving everyone to client-mu-plugins is the best approach...then this PR is not actually needed...unless we wanted to ship a modified version to catch any sites that somehow left it in `/plugins`.

joshbetz:
> I think moving everyone to client-mu-plugins is the best approach...then this PR is not actually needed

🚢 🇮🇹

PR #792, opened 2017-11-14
VIP Plugins: Remove empty() checks for sites without plugins active

### LINKED ISSUE #790: VIP Plugins: no traditionally loaded plugins prevent option filtering
Test case:

1) There are no plugins on the site which would be activated via UI
2) A plugin is loaded via wpcom_vip_load_plugin
expected result: the code loaded plugin is in the option.
currently: option stays empty

In this case, the first `! empty()` in https://github.com/Automattic/vip-go-mu-plugins/blob/master/vip-plugins/vip-plugins.php#L172 is false and thus the merge does not happen.  I don't think we really need the conditional here. Merging two empty arrays should be rather fast and we really don't save any performance nor prevent any bugs when always merging the two values.

The same, about testing whether array is empty, imho, goes for https://github.com/Automattic/vip-go-mu-plugins/blob/master/vip-plugins/vip-plugins.php#L176 - there is little to no performance gain in not sorting an empty array, and the extra conditional makes code less readable.

### LINKED ISSUE #791: VIP Plugins: code active plugins get saved to database
There is a case in which plugins activated via `wpcom_vip_load_plugin` get saved in "active_plugins" to database. Which leads to a situation in which previously code activated plugin does not get deactivated when code is removed.

Test case:
1) One plugin active on the site via UI (has it's entry in database)
2) One plugin active via wpcom_vip_load_plugin
3) Visit wp-admin/plugins.php and activate some third plugin

Expected: only plugin mentioned in step 1 and 3 is in active_plugins option in database
Current: the plugin mentioned in step 2 is there as well
Symptom: get_option( 'active_plugins' ) contains a duplicate entry for plugin mentioned in step 2.

I'm not yet clear on why this happens :-/

scottsweb:
* Remove empty checks to ensure we always merge code activated plugins with our options. 
* Stop code activated plugins being saved to the DB.

Closes #790 
Closes #791

### DIFF
--- a/tests/test-vip-plugins.php
+++ b/tests/test-vip-plugins.php
@@ -13,21 +13,21 @@ class VIP_Go_Plugins_Test extends \WP_UnitTestCase {
 
 		// emulate the active plugins option
 		$this->option_active_plugins = [
-			'hello.php',
 			'airplane-mode/airplane-mode.php',
-			'msm-sitemap/msm-sitemap.php',
 			'gutenberg/gutenberg.php',
+			'hello.php',
 			'liveblog/liveblog.php',
-			'wp-instagram-widget/wp-instagram-widget.php',
+			'msm-sitemap/msm-sitemap.php',
 			'shortcake/shortcode-ui.php',
+			'wp-instagram-widget/wp-instagram-widget.php',
 		];
 
 		sort( $this->option_active_plugins );
 
 		// emulate the network active plugins option
 		$this->option_active_sitewide_plugins = [
-			'hello.php' => 1507904134,
 			'akismet/akismet.php' => 1507904154,
+			'hello.php' => 1507904134,
 			'msm-sitemap/msm-sitemap.php' => 1507904154,
 			'shortcake/shortcode-ui.php' => 1507904134,
 		];
--- a/tests/test-vip-plugins.php
+++ b/tests/test-vip-plugins.php
@@ -92,7 +92,7 @@ class VIP_Go_Plugins_Test extends \WP_UnitTestCase {
 		/**
 		 * Check that the returned option matches a merge of the filtered loaded plugins and active plugins
 		 */
-		$merged_plugins = array_merge( $this->option_active_plugins, wpcom_vip_get_filtered_loaded_plugins() );
+		$merged_plugins = array_unique( array_merge( $this->option_active_plugins, wpcom_vip_get_filtered_loaded_plugins() ) );
 		sort( $merged_plugins );
 		$this->assertEquals( $merged_plugins, get_option( 'active_plugins' ), 'The value of `$merged_plugins` does not match the returned value of `get_option( \'active_plugins\')`.' );
 
@@ -108,6 +108,11 @@ class VIP_Go_Plugins_Test extends \WP_UnitTestCase {
 		 */
 		$option_db = $wpdb->get_var( "SELECT option_value FROM $wpdb->options WHERE option_name = 'active_plugins' LIMIT 1" );
 		$option_db = maybe_unserialize( $option_db );
+		$plugin_change = array_merge( $this->option_active_plugins, array( 'amp-wp/amp.php' ) );
+		// because this was a dupe plugin we will be smart enough to remove it from the option
+		if ( ( $key = array_search( 'msm-sitemap/msm-sitemap.php', $plugin_change, true ) ) !== false ) {
+			unset( $plugin_change[ $key ] );
+		}
 		sort( $plugin_change );
 		$this->assertEquals( $plugin_change, $option_db, 'The database value `$option_db` does not match `$plugin_change`.' );
 
@@ -157,7 +162,7 @@ class VIP_Go_Plugins_Test extends \WP_UnitTestCase {
 		/**
 		 * Check that the returned option matches a merge of the filtered loaded plugins and active plugins
 		 */
-		$merged_plugins = array_merge( wpcom_vip_get_network_filtered_loaded_plugins(), $this->option_active_sitewide_plugins );
+		$merged_plugins = array_merge( $this->option_active_sitewide_plugins, wpcom_vip_get_network_filtered_loaded_plugins() );
 		ksort( $merged_plugins );
 		$this->assertEquals( $merged_plugins, get_site_option( 'active_sitewide_plugins' ), 'The value of `$merged_plugins` does not match the returned value of `get_site_option( \'active_sitewide_plugins\')`.' );
 
@@ -179,7 +184,7 @@ class VIP_Go_Plugins_Test extends \WP_UnitTestCase {
 		/**
 		 * Check that the option still makes sense again
 		 */
-		$saved_plugins = array_merge( wpcom_vip_get_network_filtered_loaded_plugins(), $plugin_change );
+		$saved_plugins = array_merge( $plugin_change, wpcom_vip_get_network_filtered_loaded_plugins() );
 		ksort( $saved_plugins );
 		$this->assertEquals( $saved_plugins, get_site_option( 'active_sitewide_plugins' ), 'The value of `$merged_plugins` does not match the returned value of `get_site_option( \'active_sitewide_plugins\')`.' );
 	}
--- a/vip-plugins/vip-plugins.php
+++ b/vip-plugins/vip-plugins.php
@@ -173,7 +173,7 @@ function wpcom_vip_option_active_plugins( $value, $option ) {
 		$value = array();
 	}
 
-	$value = array_merge( $value, $code_plugins );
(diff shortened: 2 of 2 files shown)


vip-plugins/vip-plugins.php
```diff
@@ -169,13 +169,13 @@ function wpcom_vip_plugin_action_links( $actions, $plugin_file, $plugin_data, $c
 function wpcom_vip_option_active_plugins( $value, $option ) {
 	$code_plugins = wpcom_vip_get_filtered_loaded_plugins();
 
-	if ( ! empty( $value ) && ! empty( $code_plugins ) ) {
-		$value = array_merge( $value, $code_plugins );
+	if ( empty( $value ) ) {
```
david-binda: This check does not feel right. It `$value` is empty, then create empty array. I believe that what situations is being solved here is whether `$value` is or is not an array. The check, for readability of the code, should if `if ( false === is_array( $value ) )`

vip-plugins/vip-plugins.php
```diff
@@ -191,7 +191,11 @@ function wpcom_vip_option_active_plugins( $value, $option ) {
 function wpcom_vip_site_option_active_sitewide_plugins( $value, $option ) {
 	$code_plugins = wpcom_vip_get_network_filtered_loaded_plugins();
 
-	if ( ! empty( $value ) && ! empty( $code_plugins ) ) {
+	if ( empty( $value ) ) {
```
david-binda: The same as above. We should use more specific check.

vip-plugins/vip-plugins.php
```diff
@@ -222,7 +224,11 @@ function wpcom_vip_site_option_active_sitewide_plugins( $value, $option ) {
 function wpcom_vip_pre_update_option_active_plugins( $value, $old_value, $option ) {
 	$code_plugins = wpcom_vip_get_filtered_loaded_plugins();
 
-	if ( ! empty( $value ) && ! empty( $code_plugins ) ) {
+	if ( empty( $value ) ) {
```
david-binda: The same as above. We should use more specific check.

vip-plugins/vip-plugins.php
```diff
@@ -191,7 +191,11 @@ function wpcom_vip_option_active_plugins( $value, $option ) {
 function wpcom_vip_site_option_active_sitewide_plugins( $value, $option ) {
 	$code_plugins = wpcom_vip_get_network_filtered_loaded_plugins();
 
-	if ( ! empty( $value ) && ! empty( $code_plugins ) ) {
+	if ( empty( $value ) ) {
+		$value = array();
+	}
+
+	if ( ! empty( $code_plugins ) ) {
```
david-binda: I wonder why the `! empty( $code_plugins )` check is performed here, but is not present in  `wpcom_vip_option_active_plugins` function. The `foreach` along with an empty array won't really do anything, and the `array_merge` should also be rather fast for an empty array, imho.

vip-plugins/vip-plugins.php
```diff
@@ -222,7 +224,11 @@ function wpcom_vip_site_option_active_sitewide_plugins( $value, $option ) {
 function wpcom_vip_pre_update_option_active_plugins( $value, $old_value, $option ) {
 	$code_plugins = wpcom_vip_get_filtered_loaded_plugins();
 
-	if ( ! empty( $value ) && ! empty( $code_plugins ) ) {
+	if ( empty( $value ) ) {
+		$value = array();
+	}
+
+	if ( ! empty( $code_plugins ) ) {
```
david-binda: Same as above. If we feel like the check helps the code to perform less work, it should be used even in `wpcom_vip_option_active_plugins`. I personally would omit this conditional, as it feels being extra.

vip-plugins/vip-plugins.php
```diff
@@ -253,7 +257,11 @@ function wpcom_vip_pre_update_option_active_plugins( $value, $old_value, $option
 function wpcom_vip_pre_update_site_option_active_sitewide_plugins( $value, $old_value, $option, $network_id ) {
 	$code_plugins = wpcom_vip_get_network_filtered_loaded_plugins();
 
-	if ( ! empty( $value ) && ! empty( $code_plugins ) ) {
+	if ( empty( $value ) ) {
```
david-binda: The same as above. We should use more specific check.

vip-plugins/vip-plugins.php
```diff
@@ -253,7 +257,11 @@ function wpcom_vip_pre_update_option_active_plugins( $value, $old_value, $option
...
+		$value = array();
+	}
+
+	if ( ! empty( $code_plugins ) ) {
```
david-binda: Same as above. Is this really necessary?

scottsweb:
Thanks for the feedback. All good ideas! Should be fixed up now.

tests/test-vip-plugins.php
```diff
@@ -92,7 +92,7 @@ public function test__modify_active_plugins() {
 		/**
 		 * Check that the returned option matches a merge of the filtered loaded plugins and active plugins
 		 */
-		$merged_plugins = array_merge( $this->option_active_plugins, wpcom_vip_get_filtered_loaded_plugins() );
+		$merged_plugins = array_unique( array_merge( $this->option_active_plugins, wpcom_vip_get_filtered_loaded_plugins() ) );
```
david-binda: The `array_unique` here feels like a patch for failing tests, where dupes exists. It might be covering some bugs in the tests. The `array_unique` should be removed once the tests get refactored and broken down to individual cases.

tests/test-vip-plugins.php
```diff
@@ -108,6 +108,11 @@ public function test__modify_active_plugins() {
 		 */
 		$option_db = $wpdb->get_var( "SELECT option_value FROM $wpdb->options WHERE option_name = 'active_plugins' LIMIT 1" );
 		$option_db = maybe_unserialize( $option_db );
+		$plugin_change = array_merge( $this->option_active_plugins, array( 'amp-wp/amp.php' ) );
+		// because this was a dupe plugin we will be smart enough to remove it from the option
+		if ( ( $key = array_search( 'msm-sitemap/msm-sitemap.php', $plugin_change, true ) ) !== false ) {
```
david-binda: This dynamic approach to generating control group of plugins might be needed now as we have this one huge test data situation here. But we might want to hardcore the values during the unit tests refactor, in order to have better overview of what is happening.

scottsweb:
We need to wait for some production code on the :sun_with_face: to go out before we deploy this.

vip-plugins/vip-plugins.php
```diff
@@ -169,13 +169,13 @@ function wpcom_vip_plugin_action_links( $actions, $plugin_file, $plugin_data, $c
...
-	if ( ! empty( $value ) && ! empty( $code_plugins ) ) {
-		$value = array_merge( $value, $code_plugins );
+	if ( false === is_array( $value ) ) {
+		$value = array();
 	}
 
-	if ( ! empty( $value ) ) {
-		sort( $value );
-	}
+	$value = array_unique( array_merge( $value, $code_plugins ) );
```
david-binda: I know that I have already approved the PR, I just wonder why this `array_unique` is necessary here, while it is not needed in the `wpcom_vip_site_option_active_sitewide_plugins ` ?

vip-plugins/vip-plugins.php
```diff
@@ -222,20 +215,13 @@ function wpcom_vip_site_option_active_sitewide_plugins( $value, $option ) {
 function wpcom_vip_pre_update_option_active_plugins( $value, $old_value, $option ) {
```
david-binda: The `wpcom_vip_pre_update_option_active_plugins ` and `wpcom_vip_pre_update_site_option_active_sitewide_plugins` are identical now. Except for the number of args it accepts. Perhaps we could merge those two into a single function and have the `$network_id ` param being set to 0 in case that param is not provided by `pre_update_option_active_plugins` filter? It would make the code cleaner.

vip-plugins/vip-plugins.php
david-binda: Once this discrepancy gets resolved, we also might want to merge this function with the `wpcom_vip_site_option_active_sitewide_plugins` and have a single one instead of two identical functions.

scottsweb: The array unique removes duplicates (plugins activated by both code and the UI). It is not needed on sitewide plugins as the array structure is different (keyed by a timestamp). It is possible that two timestamps could be the same and therefore we would not able to make that array  unique.

vip-plugins/vip-plugins.php
scottsweb: I think we need them to be different as per above.

joshbetz:
r101990-deploy

Issue #788, opened 2017-11-14
Disable Jetpack Activation/Deactivation


joshbetz:
- [x] Suppress deactivation hooks
- [x] Suppress activation hooks
- [ ] Disable connecting from wp-admin (in favor of Jetpack Start)
- [ ] Disable activating via UI (in case of custom versions)?

mjangda:
> Disable connecting from wp-admin (in favor of Jetpack Start)

This was implemented here: 

https://github.com/Automattic/vip-go-mu-plugins/pull/721

But is broken due to this Jetpack bug:

https://github.com/Automattic/jetpack/issues/8066

PR #796, opened 2017-11-27
Suppress Jetpack activation & deactivation hooks

joshbetz:
We don't want to disconnect Jetpack in case it is accidentally (or
otherwise) disabled on VIP Go.

See #788


joshbetz:
Needs testing

rinatkhaziev:
This is stale and things have changed since then, closing.

Issue #797, opened 2017-11-29
Move copy of Jetpack used by WPCOM_VIP_JETPACK_LOCAL to client MU plugins


simonwheatley:
We provide a method to override the version of Jetpack in MU plugins with a "local" copy, in order that clients can test new versions in different environments ([documentation](https://vip.wordpress.com/documentation/vip-go/testing-jetpack-on-vip-go/)).

Currently we have users place the "local" version in the `plugins` directory of their client site repository, we should change this to the `client-mu-plugins` directory.

PR #798, opened 2017-11-29
Use `client-mu-plugins` instead of `plugins` for the alternative Jetpack codebase

simonwheatley:
Closes https://github.com/Automattic/vip-go-mu-plugins/issues/797


simonwheatley:
We need to coordinate the merge and deployment of this with some documentation changes here:
https://vip.wordpress.com/documentation/vip-go/testing-jetpack-on-vip-go/

nickdaugherty:
This should be fine because of the `require_once` - the client mu plugins loading code would already automatically be including the local copy of JP, but I suppose it's better to force load this early, at the same time regular Jetpack would be loaded, for consistency (vs delaying and allowing regular client-mu-plugins code to do it).

nickdaugherty:
I guess the above would only apply on multisites.

joshbetz:
r101621-deploy

PR #794, opened 2017-11-21
WP Trac #42573: Fix for theme template file caching.

david-binda:
Flush the theme file cache each time the admin screens are loaded which uses the file list.

Copy of the plugin from https://gist.github.com/westonruter/6c2ca0e5a4da233bf4bd88a1871dd950 with additional `funciton_exists` check for our MU plugins.

PR #800, opened 2017-11-30
Removing the fix for theme template file caching.

david-binda:
The hotfix for core issue introduced via #794 is no longer needed as the platform is currently on 4.9.1 which patches the issue.


joshbetz:
r101989-deploy

PR #801, opened 2017-11-30
Apply DB schema changes on cleanup

joshbetz:
Run dbDelta on `wp vip migration cleanup` to ensure all the indexes are created properly.

**Note:** We need to wait for WP CLI commands to be async. Schema changes can be slow and blocking VIPd is bad.


joshbetz:
Before we merge this, we should offload wp-cli commands to WP CLI containers (not synchronous in VIPd) because dbDelta can be slow on big tables.

rinatkhaziev:
The issue is addressed, so I'm closing this.

PR #803, opened 2017-12-06
Fix cache manager in PHPUnit

joshbetz:
Tests can modify a lot of data, which causes lots of cache purges,
which generates an error. This bails before the errors are generated.

See #784

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -7,7 +7,7 @@ if ( ! $_tests_dir ) {
 
 require_once $_tests_dir . '/includes/functions.php';
 
-define( 'VIP_GO_PHPUNIT_RUNNING', true );
+remove_action( 'init', array( 'VIP_Cache_Manager', 'init' ) );
 
 // Constant configs
 // Ideally we'd have a way to mock these
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -7,6 +7,8 @@ if ( ! $_tests_dir ) {
 
 require_once $_tests_dir . '/includes/functions.php';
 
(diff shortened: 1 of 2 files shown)


vip-cache-manager/vip-cache-manager.php
```diff
@@ -212,6 +212,10 @@ public function execute_purges() {
 			return;
 		}
 
+		if ( defined( 'VIP_GO_PHPUNIT_RUNNING' ) && VIP_GO_PHPUNIT_RUNNING ) {
```
nickdaugherty: Alternatively, would it work to put something like this into the bootstrap file?

```
remove_action( 'init', array( 'VIP_Cache_Manager', 'init' ) );
```

tests/bootstrap.php
```diff
@@ -7,6 +7,8 @@
 
 require_once $_tests_dir . '/includes/functions.php';
 
+remove_action( 'init', array( 'VIP_Cache_Manager', 'init' ) );
```
mjangda: Tests failing because this is being called too early.
```
Error: Call to undefined function remove_action() in /home/travis/build/Automattic/vip-go-mu-plugins/tests/bootstrap.php on line 10
```

fklein-lu:
The "Too many purges" message also has appeared on local when running a WP CLI script that uses `wp_update_term()`.

I avoided the Cache Manager warning with:
```
$vip_cache_manager = \WPCOM_VIP_Cache_Manager::instance();
remove_action( 'clean_term_cache', [ $vip_cache_manager, 'queue_terms_purges' ] );
```

PR #805, opened 2017-12-06
Add Lightweight term counting plugin

sboisvert:
Some sites are having performance problems caused by term counting.
Term counting is slow and on WordPress.com is either sent to the jobs system or uses this plugin.

The plugin has been enabled on WordPress.com for around 8 months now and has proven stable.

This was tested on vip-test.go-vip.co by creating new tags, adding posts to them, removing posts from them and making sure the count matched.


joshbetz:
r102095-deploy

PR #806, opened 2017-12-08
Blacklist $helper_file variable in _wpcom_vip_include_plugin

david-binda:
As it's defined post our initial `get_defined_vars()` call, this variable is leaking to the global namespace.

This commit adds the variable to the blacklist, which is meant to prevent such situations.


mjangda:
r105342-deploy

PR #813, opened 2017-12-22
Only block sandbox jetpack downgrades (not upgrades)

### LINKED ISSUE #812: Only block JP upgrades on sandboxes when code version is less than db version
If JP is being upgraded via a sandbox, we should allow it and not die.

If JP is being rolled back via sandbox request (like the sandbox has an old version of JP), we _should_ continue to block the request.

The problem is, JP only runs the upgrade routine if it's active:

https://github.com/Automattic/jetpack/blob/e6d331bed7b58cbbbaadf7b8f4eb7526862d9645/class.jetpack.php#L355

So when attempting to connect JP via command line on a sandbox, the Jetpack::is_active() check starts to pass, but the versions mismatch (since the plugin upgrade code hasn't run yet previously), which stops the request.

nickdaugherty:
Otherwise we get into situations where we can’t connect JP, since the
upgrade wouldn’t have been triggered any other way (since there is an
`is_active()` check wrapping the upgrade routine)

Fixes #812

### DIFF
--- a/vip-jetpack/vip-jetpack.php
+++ b/vip-jetpack/vip-jetpack.php
@@ -34,6 +34,13 @@ if ( true === WPCOM_SANDBOXED ) {
 			return;
 		}
 
+		// If we're upgrading, then it's fine. We only want to prevent accidental downgrades
+		// Jetpack::maybe_set_version_option() already does this check, but other spots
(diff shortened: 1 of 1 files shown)


nickdaugherty:
r103524-deploy

PR #816, opened 2018-01-05
Use HTTPS for submodules

joshbetz:
SSH submodules require a Github account and associated SSH key. Since all of these submodules are public, we can use HTTPS instead to avoid requiring a Github account to clone the repo.


mjangda:
FWIW, this came up before and we left them as SSH since this is supposed to be a development repo and you need a GitHub account to develop it :)

Not a big deal though.

PR #817, opened 2018-01-09
Migration: Cleanup: Purge page cache

joshbetz:
Purge the page cache after data syncs.


mjangda:
Going to close this; hasn't been a huge problem so far. Can revisit later it needed.

PR #819, opened 2018-01-09
Bump JP Version number in mu-plugin file

nickdaugherty:
Forgot to bump the version here to match what’s in the JP submodule


joshbetz:
r103814-deploy

PR #827, opened 2018-01-26
Mark Cron transactions as background (non-web)

simonwheatley:
Props @spacedmonkey

Reference: https://docs.newrelic.com/docs/agents/php-agent/php-agent-api/newrelic_background_job

### DIFF
--- a/vip-helpers/vip-newrelic.php
+++ b/vip-helpers/vip-newrelic.php
@@ -61,6 +61,7 @@ function wpcom_vip_cron_for_newrelic() {
 		&& function_exists( 'newrelic_add_custom_parameter' )
 		&& function_exists( 'newrelic_name_transaction' ) ) {
 		newrelic_name_transaction( 'wp-cron' );
+		newrelic_background_job( true);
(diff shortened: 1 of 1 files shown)


vip-helpers/vip-newrelic.php
```diff
@@ -61,6 +61,7 @@ function wpcom_vip_cron_for_newrelic() {
 		&& function_exists( 'newrelic_add_custom_parameter' )
 		&& function_exists( 'newrelic_name_transaction' ) ) {
 		newrelic_name_transaction( 'wp-cron' );
+		newrelic_background_job( true);
```
mjangda: We have  `function_exists` checks for the other newrelic functions we're calling. Should we have a check for this one as well?

sboisvert: I don't think it's a bad thing to add them but I think the main purpose is to see if the New Relic extension is loaded. I guess depending on the version of the extension it could only have certain functions so might as well look for those :)

simonwheatley:
Ready for re-review, @mjangda @sboisvert 🙂

simonwheatley:
104792-deploy

PR #828, opened 2018-01-29
Convert A8C_Files to a singleton

sboisvert:
Beaver Builder is trying to use file_put_content() which doesn't work for the caches for CSS and JS files.

We modify A8C_Files to make it a singleton here so we can access it from outside the class and use it to upload the files.


mjangda:
Closing as we went with a different approach for this.

PR #830, opened 2018-01-30
Jetpack: Enforce some modules.

philipjohn:
The `jetpack_active_modules` filter allows developers to enable/disable some modules in code. We might need to ensure some modules are always enabled/disabled and so we hook in to that filter ourselves.

At the moment, this just forces Vaultpress to be enabled. It won't have any effect until https://github.com/Automattic/jetpack/pull/8465 is merged and released.

I'm not sure this is the finished code...

* Do we want to force-enable other modules?
* Should we put a ridiculous priority on the filter to try and make sure we're last?
* Do we want to force disable any modules?


chrean:
@simonwheatley do you think we can close this PR and address it somewhere else? Or should we reprioritize?

simonwheatley:
Not a priority for me at the moment, @chrean

PR #832, opened 2018-01-31
Allow Debug Bar when sandboxed

### LINKED ISSUE #831: Always enable Debug Bar when sandboxed
When VIP staff has a site sandboxed, they should be able to see the Debug Bar.

```
$enable = is_auttomattician() || WPCOM_SANDBOXED;
```

https://github.com/Automattic/vip-go-mu-plugins/blob/master/debug-bar.php#L12

nickdaugherty:
Regardles of `is_automattician()`, Debug Bar should be enabled when user has site sandboxed

Fixes #831

### DIFF
--- a/debug-bar.php
+++ b/debug-bar.php
@@ -9,7 +9,7 @@
  Text Domain: debug-bar
  */
 add_filter( 'debug_bar_enable', function( $enable ) {
-	$enable = is_automattician();
+	$enable = is_automattician() || WPCOM_SANDBOXED;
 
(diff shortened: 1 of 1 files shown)


mjangda:
We probably need to force the admin bar to display as well? The only downside is that this makes it harder to test unauthenticated requests when sandboxed (since the admin + debug bar add a lot of cruft).

mjangda:
Closing in favour of #1030

PR #834, opened 2018-02-02
Run WP CLI commands via WP Cron

david-binda:
This is not ready for client use. If you are interested in learning, please contact VIP Support.

The code in this PR is a mechanism for offloading WP CLI commands to WP Cron which allows those command to be run on CLI containers, instead of sandboxes, which takes the commands closer to the production servers and thus should provide better performance for the commands.

### DIFF
--- a/vip-wp-cli-to-cron.php
+++ b/vip-wp-cli-to-cron.php
@@ -1,9 +1,9 @@
 <?php
 
  /***************************************************************************************************
- *																									*
+ *                                                                                                  *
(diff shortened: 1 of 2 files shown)


david-binda:
Looks like the failing PHPUnit tests are not related to this PR. Have those ever been passing?

mjangda:
Yeah, something broke with the tests recently that we still need to look into.

mjangda:
r105066-deploy

Issue #809, opened 2017-12-18
Add private cache purge endpoint


joshbetz:
A new internal, rate-limited endpoint which clears the page cache for the current site. We can use this for clearing the page cache after deploys.

It should be:

- [ ] private (internal to VIP only)
- [ ] rate-limited

rinatkhaziev:
Given the plethora of cache implementation changes, this is not relevant anymore.

PR #836, opened 2018-02-05
Add/cache manager endpoints

chrean:
Added REST endpoints for purging Varnish cache.

https://github.com/Automattic/vip-go-mu-plugins/issues/809


vip-cache-manager/includes/class-rest-api.php
```diff
@@ -0,0 +1,130 @@
...
+			'permission_callback' => array( $this, 'check_secret' ),
+			'show_in_index'       => false,
+		) );
+
+	}
+
+	/**
+	 * Testing function
+	 */
+	public function debug_request( $request ) {
```
joshbetz: Looks like we need to flesh out the BAN method here

vip-cache-manager/includes/class-rest-api.php
```diff
@@ -0,0 +1,130 @@
...
+
+	/**
+	 * Register API routes
+	 */
+	public function rest_api_init() {
+
+		register_rest_route( self::API_NAMESPACE, '/' . self::ENDPOINT_PURGE, array(
+			'methods'             => 'POST',
+			'callback'            => array( $this, 'purge_urls' ),
+			'permission_callback' => array( $this, 'check_secret' ),
```
joshbetz: As mentioned in Slack, it would be best to reuse `wpcom_vip_go_rest_api_request_allowed()` here. Since it uses a header, it's slightly easier to obscure in logs as necessary.

joshbetz:
Any thoughts on rate limiting these?

vip-cache-manager/includes/class-rest-api.php
chrean: Sure, it was just a skeleton but now that we agreed on not using it - did we? -, it can be cleaned up.

vip-cache-manager/includes/class-rest-api.php
chrean: Sure, will work on this, makes definitely sense.

vip-cache-manager/includes/class-rest-api.php
joshbetz: Sure, let's remove it for now.

chrean: Done.

chrean: * Main class does not extend Singleton anymore.
* Ban route has been removed.
* Now using wpcom_vip_go_rest_api_request_allowed in place of shared secret.

chrean:
Hi @joshbetz , 

I've added exoskeleton and made some tests, and is apparently working. I say "apparently" because that plugin has been flagged as alpha, so I think it'd be better if we review it and perform further testing before thinking of using it for live envos.

mjangda:
No plans to implement this in the near future, so I think we can close this one out. We'll likely take a different approach if we revisit anyway.

PR #838, opened 2018-02-07
Don't override results of unique_filename_callback

mjangda:
`wp_unique_filename` allows passing in a custom callback as an added level of customization. In our filter for `wp_unique_filename`, we're just passing through the results as-is if the custom callback is used. Otherwise we can end up modifying the results, which may be undesirable.

One use case is when we want to override an existing file. We can use the custom callback to set the existing file's name. Without this change, our mu-plugins unique filename check runs and modifies the filename, which is not what we want.


mjangda:
r105341-deploy

PR #841, opened 2018-02-07
Extract migration cleanup into standalone functions

### LINKED ISSUE #840: Migration Cleanup: Convert to Cron
To better support client code (that could be very long running), we should convert the cleanup routines to a cron job instead of direct CLI command.

nickdaugherty:
So we can use it via CLI, or wp cron

Adds a wp cron entry point

Fixes #840

### DIFF
--- a/vip-helpers/vip-migrations.php
+++ b/vip-helpers/vip-migrations.php
@@ -1,0 +1,59 @@
+<?php
+
+add_action( 'vip_post_migration', 'vip_post_migration' );
+
+function vip_post_migration() {
+	if ( is_multisite() ) {
+		$sites = get_sites();
+
+		foreach( $sites as $site ) {
+			switch_to_blog( $site['blog_id'] );
+
+			vip_run_post_migration_cleanup();
+
+			restore_current_blog();
+		}
+
+		return true;
+	} else {
+		return vip_run_post_migration_cleanup();
+	}
+}
+
+function vip_run_post_migration_cleanup() {
+	/**
+	 * Fires on migration cleanup
+	 *
+	 * Migration cleanup runs on VIP Go during the initial site setup
+	 * and after database imports. This hook can be used to add additional
+	 * cleanup for a given site.
+	 */
+	do_action( 'vip_go_migration_cleanup' );
+
+	vip_delete_db_transients();
+
+	wp_cache_flush();
+
+	vip_connect_jetpack();
+	vip_connect_vaultpress();
+
+	return true;
+}
+
+function vip_delete_db_transients() {
+	global $wpdb;
+
+	return $wpdb->query(
+		"DELETE FROM $wpdb->options
+		WHERE option_name LIKE '\_transient\_%'
+		OR option_name LIKE '\_site\_transient\_%'"
+	);
+}
+
+function vip_connect_jetpack() {
+	return \WP_CLI::runcommand( sprintf( 'jetpack-start connect --url=%s', home_url() ) );
+}
+
+function vip_connect_vaultpress() {
+	return \WP_CLI::runcommand( sprintf( 'vaultpress register_via_jetpack --url=%s', home_url() ) );
+}
--- a/vip-helpers/vip-migrations.php
+++ b/vip-helpers/vip-migrations.php
@@ -53,11 +53,15 @@ function delete_db_transients() {
 function connect_jetpack() {
 	if ( defined( 'WP_CLI' ) && WP_CLI && class_exists( 'WP_CLI' ) ) {
 		\WP_CLI::runcommand( sprintf( 'jetpack-start connect --url=%s', home_url() ) );
+	} else {
+		trigger_error( 'Cannot connect JP outside of a WP_CLI context, skipping', E_USER_WARNING );
 	}
 }
 
 function connect_vaultpress() {
 	if ( defined( 'WP_CLI' ) && WP_CLI && class_exists( 'WP_CLI' ) ) {
 		\WP_CLI::runcommand( sprintf( 'vaultpress register_via_jetpack --url=%s', home_url() ) );
+	} else {
+		trigger_error( 'Cannot connect VaultPress outside of a WP_CLI context, skipping', E_USER_WARNING );
 	}
 }
--- a/vip-helpers/vip-migrations.php
+++ b/vip-helpers/vip-migrations.php
@@ -1,26 +1,28 @@
 <?php
 
-add_action( 'vip_after_data_migration', 'vip_after_data_migration' );
+namespace Automattic\VIP\Migration;
 
-function vip_after_data_migration() {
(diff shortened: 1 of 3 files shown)


vip-helpers/vip-migrations.php
```diff
@@ -0,0 +1,59 @@
...
+
+	return $wpdb->query(
+		"DELETE FROM $wpdb->options
+		WHERE option_name LIKE '\_transient\_%'
+		OR option_name LIKE '\_site\_transient\_%'"
+	);
+}
+
+function vip_connect_jetpack() {
+	return \WP_CLI::runcommand( sprintf( 'jetpack-start connect --url=%s', home_url() ) );
```
joshbetz: Is this going to work in a non-WP_CLI context?

nickdaugherty: 🤷‍♂️

simonwheatley: I think this will work:

```
$ wp shell
wp> \WP_CLI::runcommand( sprintf( 'jetpack-start connect --url=%s', home_url() ) );
- Jetpack is already connected
-- Everything looks good!
Success: All done! Welcome to Jetpack! ✈️️✈️️✈️️
=> phar:///root/roles/docker-container-vip-web/usr/local/bin/wp/vendor/wp-cli/shell-command/src/WP_CLI/REPL.php:43:
NULL
```

joshbetz: `wp shell` is a WP_CLI context though.

joshbetz: I think we should at least wrap these in `function_exists()` checks.

nickdaugherty: Added a check in f16530a

nickdaugherty:
Ready for review - have tested both CLI and cron on standard and multisite installs.

vip-helpers/vip-migrations.php
```diff
@@ -0,0 +1,67 @@
+<?php
+
```
mjangda: Should we use a namespace for consistency with new code? Maybe `Automattic\VIP\Migration`?

vip-helpers/vip-migrations.php
```diff
@@ -0,0 +1,67 @@
...
+	/**
+	 * Fires on migration cleanup
+	 *
+	 * Migration cleanup runs on VIP Go during the initial site setup
+	 * and after database imports. This hook can be used to add additional
+	 * cleanup for a given site.
+	 */
+	do_action( 'vip_go_migration_cleanup' );
+
+	vip_delete_db_transients();
```
mjangda: Should we just hook these functions into `vip_go_migration_cleanup` instead? It would simplify the wp-cli as well as get rid of the duplication. For the dry run, we could just list out the callback names that would fire (or just remove the dry-run).

vip-helpers/vip-migrations.php
```diff
@@ -0,0 +1,67 @@
...
+
+	wp_cache_flush();
+
+	vip_connect_jetpack();
+	vip_connect_vaultpress();
+
+	return true;
```
mjangda: The `return` is unnecessary here.

vip-helpers/vip-migrations.php
```diff
@@ -0,0 +1,67 @@
...
+		WHERE option_name LIKE '\_transient\_%'
+		OR option_name LIKE '\_site\_transient\_%'"
+	);
+}
+
+function vip_connect_jetpack() {
+	if ( defined( 'WP_CLI' ) && WP_CLI && class_exists( 'WP_CLI' ) ) {
+		return \WP_CLI::runcommand( sprintf( 'jetpack-start connect --url=%s', home_url() ) );
+	}
+
```
mjangda: Should we trigger a warning here since the function is being called incorrectly (i.e. outside wp-cli context)?

vip-helpers/vip-migrations.php
```diff
@@ -0,0 +1,67 @@
...
+	return false;
+}
+
+function vip_connect_vaultpress() {
+	if ( defined( 'WP_CLI' ) && WP_CLI && class_exists( 'WP_CLI' ) ) {
+		return \WP_CLI::runcommand( sprintf( 'vaultpress register_via_jetpack --url=%s', home_url() ) );
+	}
+
```
mjangda: Same as above about the warning.

vip-helpers/vip-migrations.php
```diff
@@ -0,0 +1,67 @@
...
+			switch_to_blog( $site->blog_id );
+
+			vip_run_after_data_migration_cleanup();
+
+			restore_current_blog();
+		}
+
+		return true;
+	} else {
+		return vip_run_after_data_migration_cleanup();
```
mjangda: The `return`s are unnecessary here.

vip-helpers/vip-migrations.php
```diff
@@ -0,0 +1,67 @@
...
+	return true;
+}
+
+function vip_delete_db_transients() {
+	global $wpdb;
+
+	return $wpdb->query(
+		"DELETE FROM $wpdb->options
+		WHERE option_name LIKE '\_transient\_%'
+		OR option_name LIKE '\_site\_transient\_%'"
```
mjangda: I know this is old code, but probably worth adding a sane `LIMIT` (plus a loop to make sure we get all the entries)? Could probably do in a separate PR.

vip-helpers/vip-migrations.php
nickdaugherty: 👍 - f393c5c

vip-helpers/vip-migrations.php
nickdaugherty: Good idea - 1c271e3

vip-helpers/vip-migrations.php
nickdaugherty: IMO it's ok to keep them separate - the CLI command has a slightly different use case as it is meant to also cleanup jp/vp options (which the cron event doesn't as that's already happened). I can imagine it could diverge further in the future.

vip-helpers/vip-migrations.php
nickdaugherty: Good idea - f393c5c

vip-helpers/vip-migrations.php
nickdaugherty: 👍 https://github.com/Automattic/vip-go-mu-plugins/issues/842

vip-helpers/vip-migrations.php
nickdaugherty: Moved it to the same func that's called by cron in https://github.com/Automattic/vip-go-mu-plugins/pull/841/commits/a04c34c80a944813ec0cc49007556278e6cccbf7

nickdaugherty:
r105534-deploy

PR #843, opened 2018-02-14
Update to Query Monitor 2.17.0

simonwheatley:
/cc @johnbillion @fklein-lu


mjangda:
We're up to 3.0 so this PR needs a refresh: https://github.com/johnbillion/query-monitor/releases

PR #844, opened 2018-02-20
Moving the sprintf to act on `$paging_endpoint` only

### LINKED ISSUE #823: [VIP Cache Manager] Escape percent signs in term URLs
In cases where the slug has percent signs, this causes a 'too few arguments warning' at https://github.com/Automattic/vip-go-mu-plugins/blob/master/vip-cache-manager/vip-cache-manager.php#L518 

For example, category name वर्ग has the slug %e0%a4%b5%e0%a4%b0%e0%a5%8d%e0%a4%97

Are there any downsides to a simple str_replace fix in the line above? e.g.
`str_replace( '%', '%%', $maybe_purge_url )`

david-binda:
In case the slug contains some percentage signs, for instance due to encoding non-ascii chars, the `sprintf` call fails.

Instead of proposed str_replace escaping ( #823 ), this approach does no mess up with the URL, but only calls sprintf with `$paging_endpoint` as a param - that variable is supposed to contain  single `%d` conversion specifier and thus the `sprintf` should not fail due to the confusion over missing params.

Fixes #823

### DIFF
--- a/vip-cache-manager/vip-cache-manager.php
+++ b/vip-cache-manager/vip-cache-manager.php
@@ -514,8 +514,7 @@ class WPCOM_VIP_Cache_Manager {
 			$term_purge_urls[] = $maybe_purge_url;
 			// Now add the pages for the archive we're clearing
 			for( $i = 2; $i <= $max_pages; $i++ ) {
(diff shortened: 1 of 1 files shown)


mjangda:
r106223-deploy

PR #846, opened 2018-02-21
WP CLI to Cron logging improvements

philipjohn:
Some improvements to the logging for CLI commands offloaded to cron;

* Improved the URL reported when running on multisite
* Added additional IRC logging to record the completed run


mjangda:
We should probably add similar improvements to the emails?

philipjohn:
There's already an output e-mail, so I didn't think it'd be worth doing a "finished" e-mail, and I've changed the URL on all the e-mails where it's used, I think.

mjangda:
> I've changed the URL on all the e-mails where it's used, I think.

Sorry, I meant adding the url to emails that don't currently include them.

mjangda:
> Sorry, I meant adding the url to emails that don't currently include them.

Added in dc3e2095

mjangda:
r106238-deploy

Issue #848, opened 2018-03-07
Remove wpcom_vip_include_active_plugins()


nickdaugherty:
Now that (almost) all plugins are removed, I think we can remove `wpcom_vip_include_active_plugins()` and its `plugins_loaded` hook. 

An edge case scenario arises, when a site had `vip-go-elasticsearch` enabled via UI at some point in the past (and there is a subsequent entry in the `wpcom_vip_active_plugins` option), but has since migrated to loading via code, then onto Jetpack Search - `vip-go-elasticsearch` is still loaded with no way to disable other than removing the option, which is unexpected.

What do you think @scottsweb?

nickdaugherty:
The other aspect of this is that we don't show these plugins as enabled in the Plugins UI anymore (right?), so there is no indication of it being enabled and no way to disable.

scottsweb:
Correct, these are not shown as active anymore because that option (`wpcom_vip_active_plugins`) is effectively out of action these days.

There are only three plugins I believe that could be impacted by this bug:

https://github.com/Automattic/vip-go-mu-plugins/blob/a05aa70ce1cddc60e5106829e63166ad3c036d48/vip-helpers/vip-utils.php#L1075-L1079

and only sites without the `WPCOM_VIP_DISABLE_SHARED_PLUGINS` constant will suffer from the issue. All sites that went through the transition process should have this in place. If my memory serves me correctly the site that has prompted this issue decided to implement a patch without our help and did not include the `WPCOM_VIP_DISABLE_SHARED_PLUGINS` constant or it is possible it was later removed. 

I am just going to check the plugin stats but I am in favour of retiring `wpcom_vip_include_active_plugins` and will put together a PR if the stats look sensible.

scottsweb:
See our P2 for a follow up

WPprodigy:
Closing per https://github.com/Automattic/vip-go-mu-plugins/pull/849#issuecomment-990257085

PR #849, opened 2018-03-07
Remove wpcom_vip_include_active_plugins()

scottsweb:
As discussed in #848


brandon-m-skinner:
@scottsweb is this PR still valid? If so I'd like to pursue its completion/deployment.

scottsweb:
I believe it would be still valid yes... might be good to double check with @nickdaugherty or @mjangda in case anything has changed.

nickdaugherty:
We'd need to check again for any uses of the function across our own or client code first.

brandon-m-skinner:
@nickdaugherty There's only one occurrence and it's a copy of the mu-plugins-built branch in a client repo https://github.com/wpcomvip/insidehook/blob/724da6dc7c26a7c4db5bc3c6980fde9b085fe877/mu-plugins/vip-plugins/vip-plugins.php#L268-L280

pauarge (approved):
LGTM

pauarge:
It looks like the reference that @brandon-m-skinner mentions in the comment above no longer exists.

pauarge (changes requested):
Actually, we probably should have to delete this: https://github.com/Automattic/vip-go-mu-plugins/blob/71552b9adeb42595fb1bc44f06cf25d4dde450cc/rest-api/vip-endpoints.php#L373

and this: https://github.com/Automattic/vip-go-mu-plugins/blob/71552b9adeb42595fb1bc44f06cf25d4dde450cc/rest-api/vip-endpoints.php#L343

pauarge:
Noting that we found quite a few sites that would break if this would be merged.

WPprodigy:
Next step is probably adding some deprecation notices/warnings in the methods we want to remove. Then we can traverse logs for a bit until we've knocked out all existing codebases that would cause a problem.

WPprodigy:
Took another look. So as long as these two plugins still exist, we can't really remove this function as far as I can tell: https://github.com/Automattic/vip-go-mu-plugins/blob/83d7293cef0221ab753fbf262939988e0f74081d/vip-helpers/vip-utils.php#L1246-L1249

Going to close out this PR.

PR #852, opened 2018-03-08
Add doing_it_wrong for late loading plugins

mjangda:
For now, plugins loaded via code, should be loaded earlier before the theme. Preferably via `client-mu-plugins` so that any callbacks on `plugins_loaded` fire as expected.


GaryJones:
> For now

Does this imply that it might change at some point in the future?

mjangda:
> Does this imply that it might change at some point in the future?

We may consider alternate ways to load plugins via code, e.g. config files.

brandon-m-skinner:
@mjangda Is this still PR still valid? If so I'd like to pursue its completion/deployment.

mjangda:
Yep, still valid.

mjangda:
PR refreshed. We'll probably want to keep an eye on warning/notice counts post-deploy.

brandon-m-skinner:
r899-stacks

PR #854, opened 2018-03-09
Jetpack Start: Make args for `connect_site` optional

### LINKED ISSUE #853: PHP 7.1 Compatibility
`Jetpack_Start_CLI_Command::connect_site()` requires 1 argument:
https://github.com/Automattic/vip-go-mu-plugins/blob/9ab4bc13f34eabc71f613686cc334487de3e01a7/jetpack-start/wp-cli-api.php#L108

but is called with none here:

https://github.com/Automattic/vip-go-mu-plugins/blob/9ab4bc13f34eabc71f613686cc334487de3e01a7/jetpack-start/wp-cli-api.php#L218

This will throw a fatal since PHP 7.1 - http://php.net/manual/en/migration71.incompatible.php#migration71.incompatible.too-few-arguments-exception

It doesn't appear that `$assoc_args` is heavily used within `connect_site`, perhaps just default it to an empty array?

mjangda:
Since we don't use them heavily in the method anyway. This also fixes compatibility issues with PHP 7.1 because we have at least one call to the method that doesn't pass in the arg.

Fixes #853

### DIFF
--- a/jetpack-start/wp-cli-api.php
+++ b/jetpack-start/wp-cli-api.php
@@ -105,7 +105,7 @@ class Jetpack_Start_CLI_Command extends WP_CLI_Command {
 		}
 	}
 
-	private function connect_site( $assoc_args ) {
+	private function connect_site( $assoc_args = [] ) {
 		if ( Jetpack::is_active() ) {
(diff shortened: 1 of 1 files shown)


mjangda:
r107310-deploy

PR #855, opened 2018-03-09
Jetpack Start: Fix disconnect-and-connect routine in Jetpack Start

mjangda:
Use the direct `Jetpack::disconnect` method so that the connection state actually gets cleared in the current request. Otherwise, our `is_active()` check on reconnection fails because Jetpack is still connected in the current session.


mjangda:
r107311-deploy

PR #856, opened 2018-03-14
Update VIP Support plugin to latest master version

david-binda:
Fetches the "Hook the custom user_has_cap callback as late as possible" commit.


david-binda:
r107524-deploy

PR #857, opened 2018-03-16
Add VIP Go stats pixel

joshbetz:
Right now we aggregate Jetpack stats, but this has some flaws. Namely:

* Stats are tracked per day, but ideally we'd want per hour stats
* Stats are only tracked for WordPress, we can extend this to other types of apps
* We miss stats for sites that don't have Jetpack

Note: When we deploy this, we should consider the current stats aggregation. Ideally we should
wait until just after the stats aggregation runs at 00:00 UTC, shift everything back a day (because
right now the stats are actually off by 1 day), and then turn this on to avoid double counting.

Alternatively, it would be much easier to disable the stats aggregation job before 00:00 and then deploy
this as close to 00:00 as possible. We'll technically be missing a day of stats, but that might be fine.

### DIFF
--- a/misc.php
+++ b/misc.php
@@ -154,5 +154,5 @@ add_filter( 'woocommerce_install_skip_create_files', '__return_true' );
  * Record WordPress pageviews for VIP Go
  */
 add_action( 'wp_footer', function() {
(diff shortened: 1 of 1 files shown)


brandon-m-skinner:
@joshbetz Is this still valid?

mjangda:
I think we can close this. We have other stats tracking in place now.

PR #860, opened 2018-04-04
Image sizes meta support for VIP Go

david-binda:
In order to support srcset and other places where intermediate image sizes are expected by WordPress, the implementation in this commit generates sizes key of the image metadata w/o actually touching or generating new images. It computes and prepares VIP Go File service URLs and produces all necessary data for WordPress to handle those as if the intermediate images existed.

The commit also contains unit and integration tests and conditional way for loading the implementation for easy rollout. The implementation is only enabled on non-production environments with possibility of loading it via filters in site's theme:

```
add_filter( 'vip_go_srcset_enabled', '__return_true' );
```


joshbetz:
Is it only enabled for non-production so that we can test before rolling it out everywhere? We should avoid more permanent differences between environments.

david-binda:
Right, that was my intention - let clients to test it outside production with possibility to turn it on/off in their environment, and then, upon known date, turn it on for all environments.

However, I'm open to any suggestions to how the best roll this change out on the platform.

a8c-files.php
```diff
@@ -71,15 +71,6 @@ private function init_jetpack_photon_filters() {
 		// This results in all images being full size (which is not ideal)
 		add_filter( 'jetpack_photon_development_mode', '__return_false', 9999 );
 
-		// The sizes metadata is not used and mostly useless on Go so let's empty it out.
```
david-binda: It has just occurred to me that we might need to keep this in place or backfill in case the srcset implementation is turned off.

I'll look into it.

a8c-files/ImageSizes.php
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	/**
+	 * @param array
+	 *
+	 * @return array|\WP_Error Array for usage in $metadata['sizes']; WP_Error on failure.
+	 */
+	protected function resize( $size_data ) {
+
+		$resized_image = $this->image->resize( $size_data );
```
mjangda: Unless I'm missing something, it feels like we could remove some complexity by having this method just return the data in the format of the array below, instead of changing the internal state of the image object. This would mean that we can remove the `is_resized` method and related conditionals.

tests/test-ImageSizes.php
```diff
@@ -0,0 +1,581 @@
...
+
+		// This means the sizes are not being created via any filters.
+		$this->assertEmpty( $metadata['sizes'] );
+	}
+
+	/**
+	 * Integration test of the virtual creation of intermediate sizes.
+	 * No physical copies are being created.
+	 */
+	public function test__inject_image_sizes() {
```
mjangda: It would be good to have a test for a non-image mime type as well.

tests/test-Image.php
```diff
@@ -0,0 +1,274 @@
...
+		wp_update_attachment_metadata( $attachment_id, wp_generate_attachment_metadata( $attachment_id, $this->test_image ) );
+
+		$postmeta = get_post_meta( $attachment_id, '_wp_attachment_metadata', true );
+
+		$image = new A8C_Files\Image( $postmeta, $attachment_id );
+
+		$this->assertEquals( $postmeta['width'], $image->get_width() );
+		$this->assertEquals( $postmeta['height'], $image->get_height() );
+		$this->assertEquals( 'image/jpeg', $image->get_mime_type() );
+		$this->assertFalse( $image->is_resized() );
```
mjangda: Minor: Any time there are multiple assert calls within a single test, it helps to include the `message` param to help identify which assertion failed. Otherwise, phpunit will report something like `Failed asserting that 125 matches expected 100`, which doesn't really give us a clear idea of what failed.

tests/test-Image.php
```diff
@@ -0,0 +1,274 @@
...
+
+		$width = $this->getProperty( 'width' );
+		$width->setValue( $image, 150 );
+
+		$height = $this->getProperty( 'height' );
+		$height->setValue( $image, 150 );
+
+		$get_resized_filename = $this->getMethod( 'get_resized_filename' );
+
+		$this->assertEquals( $this->test_image . '?resize=150,150', $get_resized_filename->invokeArgs( $image, [] ) );
```
mjangda: Minor: I find it's cleaner to keep logic and execution out of assertion calls. It helps with readability and debugging.

One testing pattern that helps with this (and is always nice to follow) is arrange-act-assert (e.g. http://wiki.c2.com/?ArrangeActAssert). It encourages a visual and structural separation of the various stages of a test:

```
// Arrange: setup the test condition
$attachment_id = self::factory()->attachment->create_object(
...
);
wp_update_attachment_metadata( $attachment_id, wp_generate_attachment_metadata( $attachment_id, $this->test_image ) );
...

// Act: the method under test
$get_resized_filename = $this->getMethod( 'get_resized_filename' );
$actual_resized_filename = $get_resized_filename->invokeArgs( $image, [] )

// Assert: did it work?
$this->assertEquals( $this->test_image . '?resize=150,150', $actual_resized_filename );
```

 A good tip I once heard was that you should be able to squint at a test method and visually be able to separate the three stages. It doesn't always work as cleanly as this (some tests are just too complex) but it's a good guideline to try and follow.

a8c-files.php
mjangda: The main reason we added this in was to avoid issues on migrated sites with stale attachment data that reference hard-coded image path in the sizes meta, which would break things.

If we restore this, we'll need to remember to fix the logic in `maybe_inject_image_sizes`

a8c-files/Image.php
```diff
@@ -0,0 +1,202 @@
+<?php
+
+namespace A8C_Files;
```
mjangda: Should we use a cleaner namespace for this? Maybe something like `Automattic\VIP\Files`?

a8c-files/ImageSizes.php
david-binda: I don't think you are missing something. While reducing the Image class to a mere utilility class which would produce the array for direct usage in `$metadata['sizes']` might make the implementation less complex, it would also make it, at least from my point of view, less object oriented and less flexible.

I see the Image object in this implementation as a virtual image copy which can be modified similarly to the `WP_Image_Editor`. I have originally explored a design where there would be a class extending the `WP_Image_Editor` but performing the modifications via VIP Go File Service, but that really felt unnecessarily complex.

Unless you have strong opinions on not doing this the way it's done in the PR, I would prefer to keep it this way.

On the other hand, I have added `get_size` method, which does what you described for the `resize` one.

Please, let me know your thoughts here.

tests/test-ImageSizes.php
david-binda: Very good call. Addressed via https://github.com/Automattic/vip-go-mu-plugins/pull/860/commits/df55f2db5f971cb6232ac7b1b29d08c9c379deee

david-binda:
@mjangda 👋 Sorry for the delay with addressing the feedback.

I have finally gone through all the feedback and implemented the bit which were advised, except for the bit in https://github.com/Automattic/vip-go-mu-plugins/pull/860#discussion_r179998232 where I have slightly different opinion on the matter - expressed in https://github.com/Automattic/vip-go-mu-plugins/pull/860#discussion_r183665268

My brief testing on a VIP Go testing site shows the srcset being properly generated. Also the unit test coverage is rather good, so it should be easy to iterate on the code in case we discover some flaws.

I'd like to point out the necessity off adding an extra filter for blocking generation of intermediate sizes for non-image attachments. It looks like the flaw exists on VIP Go already. I did not want to patch it in terms of the new feature, but it was borking the srcset implementation, so I have addressed that. Please, see https://github.com/Automattic/vip-go-mu-plugins/pull/860/commits/681bb93b79b068cd8a68e78540aac3f978e5df5a for more details.

Hit me with any related questions!

a8c-files/ImageSizes.php
mjangda: Your reasoning makes sense, thanks!

mikeyarce:
@mjangda are there any plans to add this to VIP Go in the near future?  Not sure what the status of this is now that @david-binda is not focusing on VIP related things.

mjangda:
The code is ready to go from my perspective but we need to figure out a communication +release plan since sites will want to test+verify and/other disable this functionality.

a8c-files.php
```diff
@@ -71,7 +71,7 @@ private function init_jetpack_photon_filters() {
 		// This results in all images being full size (which is not ideal)
 		add_filter( 'jetpack_photon_development_mode', '__return_false', 9999 );
 
-		if ( false === is_vip_go_srcset_enabled() ) {
+		if ( false === is_vip_go_srcset_enabled() || ! isset( $_GET['enable_vip_srcset'] ) || '1' !== $_GET['enable_vip_srcset'] ) {
```
mjangda: For simplicity, we should move the GET param checks to `is_vip_go_srcset_enabled`.

mjangda:
r118915-deploy

PR #862, opened 2018-04-09
Add CLI command to import theme-related options.

vaurdan:
When importing from WordPress.com, we need to reactivate the theme, and since the theme slug is different between WordPress.com and VIP Go, every menu assignment, widget assignment, and theme-related settings are lost. This happens because in WordPress.com all the VIP themes are prefixed with `vip/`. 

This command allows to import these options from a JSON file. The JSON file is generated in a WordPress.com sandbox, with the command `wp vip-export theme-options`.

### DIFF
--- a/wp-cli/vip-migrations.php
+++ b/wp-cli/vip-migrations.php
@@ -318,6 +318,121 @@ class VIP_Go_Migrations_Command extends WPCOM_VIP_CLI_Command {
 			}
 		}
 	}
+
+	/**
+	 * Import theme options, sidebars options and widgets options from a JSON file.
+	 *
+	 * ## OPTIONS
+	 *
+	 * <file>
+	 * : The JSON file with all the settings.
+	 *
+	 * [--dry-run=<true>]
+	 * : Do a "dry run" and no data modification will be done.  Defaults to true.
+	 *
+	 * [--yes]
+	 * : Skip the confirmation prompt
+	 *
+	 * ## EXAMPLES
+	 *
+	 *     # Copies theme options from the JSON file.
+	 *     $ wp vip migration import-theme-options file.json
+	 *
+	 *
+	 * @subcommand import-theme-options
+	 */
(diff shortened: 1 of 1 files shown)


mjangda:
This should probably go here instead: https://github.com/Automattic/vip-go-wpcom-compat/

vaurdan:
@mjangda Oh, I wasn't aware of that repository. I can definitely move the script there, and I understand this is a very WordPress.com specific issue. 

However, as this can be an issue for most WordPress.com imports to Go, installing `vip-go-wpcom-compat` will be a requirement for every migration. 

Since this can be used for more generic situations, like a theme switch, I think it fits here too. What do you think is the best approach? :)

mjangda:
Oh, I guess I should have looked at the code :)

> Since this can be used for more generic situations, like a theme switch, I think it fits here too. What do you think is the best approach? :)

How frequently do you see this being used? How would someone generate the JSON file?

vaurdan:
> How frequently do you see this being used? How would someone generate the JSON file?

On WordPress.com to VIP Go migrations, I believe it will need to be used pretty much every time. Regarding theme switches, I can't give you a concrete number, but I would say that around 25% of the theme switches can use this tool, as the developers might change the theme name, but keep using the same sidebars/menus for compatibility. 

> How would someone generate the JSON file?

Right now it's only available on WordPress.com (see Differencial D11606). However, if this command is going to live here, it makes all the sense to include the exporter too.

mjangda:
@vaurdan Let's either move this over to the wpcom-compat repo or a new repo entirely. Non-WP.com sites that need it can cherry-pick it into their own repo for now. If it's used frequently enough, we can move it back here.

vaurdan:
Created a [PR in Automattic/vip-go-wpcom-compat](https://github.com/Automattic/vip-go-wpcom-compat/pull/7). Closing.

PR #868, opened 2018-04-19
Files: Extract Files API communication into new class

mjangda:
A new class that will be cleaner, more re-usable to use and easier to test.

Most of the logic already exists in a8c-files and we're abstracting that out into something simpler. This PR will also attempt to switch over to use the WordPress HTTP API (instead of curl), which will simplify debugging/testing.

This PR will not try not to modify any of the existing code in a8c-files. Once we have this class added and deployed, we can look at transitioning over the a8c-files code.

### TODOs:

- [x] Core class structure
- [x] Tests for core class
- [x] is_file
- [x] Tests for is_file
- [x] delete_file
- [x] Tests for delete_file
- [x] get_file
- [x] Tests for get_file
- [x] upload_file (or maybe put_file)
- [x] Tests for upload_file

(More TODOs may be added.)


files/class-api-client.php
```diff
@@ -0,0 +1,179 @@
...
+		if ( ! file_exists( $local_path ) ) {
+			/* translators: 1: local file path 2: remote upload path */
+			return new WP_Error( 'upload_file-failed-invalid_path', __( 'Failed to upload file `%1$s` to `%2$s`; the file does not exist.' ), $local_path, $upload_path );
+		}
+
+		$file_size = filesize( $local_path );
+		$file_name = basename( $local_path );
+		[
+			'type' => $file_mime,
+		] = wp_check_filetype( $file_name );
```
joshbetz: Minor: What do you think about collapsing this to one line?

```
[ 'type' => $file_mime ] = wp_check_filetype( $file_name );
```

It could just be me getting used to new syntax, but it's a little hard to read at first glance.

mjangda: I originally had it as a single line but changed it because PHPCS was complaining. I think you're right that it does look a bit odd, especially with just one element.

mjangda:
r109651-deploy

sboisvert:
>This PR will also attempt to switch over to use the WordPress HTTP API (instead of curl), which will simplify debugging/testing.

Do you mind elaborating on that? Reading the code it felt a bit more fragile, but it could be I'm not understanding the benefits as it relates to testing :)

mjangda:
> Do you mind elaborating on that?

The HTTP API is well-tested and can be incorporated into "unit" tests like this PR does (we can't do that with `curl`). Plus, it integrates with debugging tools like Debug Bar and Query Monitor, which `curl` also doesn't do.

There's one specific element where we need to fallback on `curl` (the upload file streaming), which I'm not super happy about, but helpful/necessary for handling large files. It would be good to revisit that at some point (e.g. we probably don't need it for small files) and improve it a bit.

> Reading the code it felt a bit more fragile

What's fragile about it?

sboisvert:
>What's fragile about it?
Just the `enforce_curl_transport()` part of it. Having filters ensure we use curl feels like it could be overwritten. 
Since we have full control over the environment that's not a problem. I just recall Barry suggesting that we ensure all VIP code doesn't rely on curl but use just the wp_http functions, but I guess really this PR helps us with that in that we're less all in on curl :)

mjangda:
> we ensure all VIP code doesn't rely on curl but use just the wp_http functions

I think it's a good principle to follow for the reasons I mentioned (testability, reliability, debugging, etc.). However, sometimes, it's not possible to do everything we need using the high-level API and need to get creative :)

PR #874, opened 2018-04-27
Restore tempnam (and drop the use of wp_tempnam)

mjangda:
To avoid dependency loading issues if the Filesystem class is called too early. `wp_tempname` uses `wp_generate_password` which is loaded later.


mjangda:
r109813-deploy

PR #875, opened 2018-04-27
WP_Filesystem_VIP:  Add naive implementation of `is_dir` 

mjangda:
We don't have an API for managing directories so let's just assume we can create files on all paths.


mjangda:
r109815-deploy

PR #879, opened 2018-05-01
Add path blocks

mjangda:
To prevent known, bad paths (like `/autodiscover/autodiscover.xml`) from being hit on sites.

- [ ] Consider adding long cache expiry
- [ ] Tests


mjangda:
Not planning to implement this.

PR #881, opened 2018-05-02
Upgrade Query Monitor to 3.0.0

mjangda:
https://github.com/johnbillion/query-monitor/releases/tag/3.0.0

Previously #843

### DIFF
--- a/query-monitor.php
+++ b/query-monitor.php
@@ -34,8 +34,8 @@
  * Determines if Query Monitor should be enabled. We don't
  * want to load it if we don't have to.
  *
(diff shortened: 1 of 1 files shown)


mjangda:
https://github.com/johnbillion/query-monitor/commit/019b158a92fc7d075d6007bb22514bea0a664a9c switches to runtime based capability assignment instead of on activation (all users with `manage_options` on single sites can access it). This doesn't impact us since we do the cap check before loading QM (so it's filtering code doesn't actually run).

mjangda:
r112451-deploy

PR #884, opened 2018-05-08
Don't override upload limit on Multisite

mjangda:
If a limit has been explicitly set.

TODO:

- [x] Tests


rinatkhaziev:
I'm marking this as abandoned, but feel free to resurrect.

PR #886, opened 2018-05-08
Disable email sharing when recaptcha is not set-up

### LINKED ISSUE #826: Block JP share by email if reCaptcha is not enabled
Email based sharing should not be supported unless recaptcha is setup, to prevent abuse.

https://jetpack.com/2013/04/15/recaptcha-email-sharing-button/

rebeccahum:
As per #826, we should only allow [email sharing](https://jetpack.com/support/sharing/) if recaptcha is set-up.

## Testing

1) If it's not enabled already, go to Settings > Sharing and drag & drop the "Email" button into "Enabled Services"
2) Assuming recaptcha isn't set-up yet, go to a post and you should see the email sharing icon not visible.
3) [Set-up recapatcha](https://jetpack.com/tag/recaptcha/)
4) Go back to post and you should see that the email sharing icon is visible.

props @mjangda 

Fixes #826

### DIFF
--- a/vip-jetpack/vip-jetpack.php
+++ b/vip-jetpack/vip-jetpack.php
@@ -114,8 +114,9 @@ add_filter( 'jetpack_sync_listener_should_load', 'wpcom_vip_disable_jetpack_sync
 /**
  * Disable Email Sharing if Recaptcha is not setup.
  *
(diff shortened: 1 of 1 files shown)


mjangda:
To be deployed on Thursday, May 10, 2018.

simonwheatley:
r110363-deploy

PR #888, opened 2018-05-10
Merge Aggregation changes from WordPress.com wpcom-elasticsearch

mdbitz:
To upgrade all clients to 2.x we need to ensure "Facets" is no longer in use and instead aggregations are in place. This PR pulls across the wpcom-elasticsearch changes for aggregations. 

Needs to be fully tested that the update doesn't impact the results for clients using the VIP Go Search Add-On


shared-plugins/vip-go-elasticsearch/vip-go-elasticsearch.php
```diff
@@ -106,12 +114,13 @@ public function init_hooks() {
 	}
 
 	public function admin_notice_no_index() {
-		echo '<div class="error"><p>' . __( 'VIP Go Search Add-On needs a little extra configuration behind the scenes. Please contact support to make it happen.' ) . '</p></div>';
+		echo '<div class="error"><p>' . __( 'WordPress.com VIP Search Add-On needs a little extra configuration behind the scenes. Please contact support to make it happen.' ) . '</p></div>';
```
mjangda: Minor: This probably shouldn't be changed :)

mdbitz:
@mjangda sorry for the ambiguity. These changes are compliant with the existing 1.x indexes. We don't need to do any changes to the ES index (rebuild/update/etc). 

Next step on this is to do some Regression Testing on the Go sites. 

I'm about 2.5 weeks paused on the ES 1.x -> 2.x migration and so can keep this moving once i get all the random tasks offloaded again. June looks promising.

mdbitz:
Note this change is missing a dependency on an update to vip-helpers/vip-elasticsearch.php

To handle the aggregations.... Added missing logic is showing 1-1 behavior for all DFM sites (DenverPost.com, etc)

mdbitz:
PR Fixed so that helper function wpcom_search_api_wp_to_es_args is updated to handle `aggregations` . (Code 1-1 from wp.com)

mdbitz:
As a note syncing code from WP.com platform to VIP is a bit more onerous than originally anticipated.

mjangda:
r117362-deploy

PR #889, opened 2018-05-14
Added a call to remove the `vaultpress` option from the db prior to plugin registration

scarstocea:
This is to avoid a VaultPress fatal during the sync, which is caused by
the `vaultpress` option being present in the child site db.


mjangda:
Noting that code itself is fine here but for the sync use case, we do already attempt to delete this option earlier in the process so something else might be broken there and needs some investigation.

joshbetz:
r111145-deploy

PR #890, opened 2018-05-14
Plugin Fixes Plugin

mdbitz:
Intended for addition of modifications to known plugins to resolve platform level issues while fixes get integrated back up to core repos.

Modified AMP Query Var filter to handle relative links. Existing AMP-WP plugin causes the AMP page to be rendered regardless of what comes after the /AMP/ portion of the link. To ensure that we don't get in infinite indexing loops we are enforcing that AMP triggers 404s appropriately instead of re-rendering the page. This is temporary until merged into core.

### DIFF
--- a/plugin-fixes.php
+++ b/plugin-fixes.php
@@ -1,0 +1,40 @@
+<?php
+/*
+Plugin Name: VIP GO Plugin Fixes
+Description: Handles Plugin Specific Fixes for VIP GO Environments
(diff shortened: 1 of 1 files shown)


mjangda:
This is the upstream fix: https://github.com/Automattic/amp-wp/pull/910

This workaround can be removed when that is merged.

mjangda:
r110583-deploy

PR #891, opened 2018-05-15
add Gutenberg loading functionality to Go

mattoperry:
This PR allows VIPs to load Gutenberg conditionally from their themes using a helper called `wpcom_vip_load_gutenberg`.  This function is a wrapper for `ramp_for_gutenberg_load_gutenberg` and accepts the same parameters.  In order to use this function pre-5.0, VIPs will need to separately install the Gutenberg plugin in the `/plugins` directory of their repos.

Included in this PR:
- submodule for the Gutenberg Ramp plugin.
- provide a wrapper for `ramp_for_gutenberg_load_gutenberg` called `wpcom_vip_load_gutenberg`
- turn off Ramp UI by default

### DIFF
--- a/vip-plugins.php
+++ b/vip-plugins.php
@@ -10,3 +10,23 @@
  */
 
 require_once( __DIR__ . '/vip-plugins/vip-plugins.php' );
+
+/** loading helper and behaviors for Gutenberg on VIP Go **/
+
+/** load Gutenberg Ramp **/
+if ( file_exists( __DIR__ . '/ramp-for-gutenberg/ramp-for-gutenberg.php' ) ) {
+	require_once( __DIR__ . '/ramp-for-gutenberg/ramp-for-gutenberg.php' );
+}
+
+/** Turn off the UI for Ramp **/
+add_action( 'plugins_loaded', function() {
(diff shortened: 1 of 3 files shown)


vip-plugins.php
```diff
@@ -10,3 +10,23 @@
...
+}
+
+/** Turn off the UI for Ramp **/
+add_action( 'plugins_loaded', function() {
+	remove_action( 'admin_init', 'ramp_for_gutenberg_initialize_admin_ui' );
+} );
+
+/** Loading helper **/
+function wpcom_vip_load_gutenberg( $criteria ) {
+	if ( !function_exists( 'ramp_for_gutenberg_load_gutenberg' ) ) {
```
mjangda: Isn't this included as part of Ramp, which we load above?

mattoperry: It is indeed ... I was thinking of a strange case where this function would be moved into another file, making it possible to call it without loading ramp, perhaps in a local environment or something?

mattoperry: That said, there's no immediate need for it so I'd also be happy to remove it.

mjangda:
r110755-deploy

PR #893, opened 2018-05-23
Allow Developers to Opt Out from Gutenberg Ramp

fklein-lu:
This PR moves the Gutenberg Ramp loading inside of the `wpcom_vip_load_gutenberg()` function. This enables developers to opt out from loading Gutenberg Ramp.

Other changes include moving the usage information to the doc block of the loading function, making it easier for developers to understand the options that they have.

This was tested manually using the following scenarios:
- _Not_ calling the `wpcom_vip_load_gutenberg()` function does _not_ load Gutenberg Ramp.
- Calling the `wpcom_vip_load_gutenberg()` function _does_ load Gutenberg Ramp.
- The Gutenberg Ramp UI is not visible in the admin.

### DIFF
--- a/ramp-for-gutenberg.php
+++ b/ramp-for-gutenberg.php
@@ -2,30 +2,31 @@
 
 /**
  * Plugin Name: Gutenberg Ramp
- * Description: Allows theme authors to control the circumstances under which the Gutenberg editor loads. Options include "load" (1 loads all the time, 0 loads never) "post_ids" (load for particular posts) "post_types" (load for particular posts types.)
- * Version:     0.1
+ * Description: Control the circumstances under which the Gutenberg editor loads in code, using Gutenberg Ramp.
+ * Version:     0.2
  * Author:      Automattic, Inc.
  * License:     GPL-2.0+
  * License URI: http://www.gnu.org/licenses/old-licenses/gpl-2.0.html
  * Text Domain: ramp-for-gutenberg
  */
 
-// This file loads Ramp, and modifies behaviors for Gutenberg on VIP Go
-
-/** load Gutenberg Ramp **/
-if ( file_exists( __DIR__ . '/ramp-for-gutenberg/ramp-for-gutenberg.php' ) ) {
-	require_once( __DIR__ . '/ramp-for-gutenberg/ramp-for-gutenberg.php' );
-}
-
-/** Turn off the UI for Ramp **/
-add_action( 'plugins_loaded', function() {
-	remove_action( 'admin_init', 'ramp_for_gutenberg_initialize_admin_ui' );
-} );
-
-/** Loading helper **/
+/**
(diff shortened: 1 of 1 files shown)


fklein-lu:
@justnorris Cheers, back to you.

ramp-for-gutenberg.php
```diff
@@ -12,6 +12,10 @@
 
 // This file loads Ramp, and modifies behaviors for Gutenberg on VIP Go
 
+if ( defined( 'VIP_GO_DISABLE_RAMP' ) && VIP_GO_DISABLE_RAMP ) {
```
justnorris: A strong comparisson yoda condition would keep this in alignment with VIP standards ( `&& true === VIP_GO_DISABLE_RAMP` )

ramp-for-gutenberg.php
```diff
@@ -22,7 +26,11 @@
 	remove_action( 'admin_init', 'ramp_for_gutenberg_initialize_admin_ui' );
 } );
 
-/** Loading helper **/
+/**
+ * Load Gutenberg via the Gutenberg Ramp plugin.
+ *
+ * @param array|false $criteria Use `false` or [ 'load' => 1 ] to always load, [ 'load' => 0 ] to never load, [ 'post_ids' => [] ] to load for particular posts, and [ 'post_types' => [] ] to load for particular post types.
```
justnorris: Splitting the documentation into multiple lines would really improve the inline-documentation readability!

justnorris:
Thanks, @fklein-lu - this looks good to me. I left 2 inline comments that are totally not critical but would be nice to have improvements. 

Let's hold off for a couple hours until @mattoperry can have one final look just to make sure everything's a-ok.

ramp-for-gutenberg.php
mattoperry: It might also be ok to not introduce this.  Obviously docs are good, but since this is a wrapper function, it should accept the same params as the wrapped function, and we might just be able to rely on the docs for that function, rather than needing to replicate them here.  (They are about to change/expand anyway.)

In any case, this PR should probably only address what it is meant to address -- ie: loading behavior.

On the other hand, not critical either way :)

ramp-for-gutenberg.php
mattoperry: Yes please.

mattoperry:
Overall, I do like the idea of this if we can make it work.

Also noting that I'll be pushing a new Gutenberg Ramp release (v0.2) to Go that should correct the unfortunate bug that might have given rise to this originally.

mattoperry:
Sorry -- regarding the above, @maevelander helpfully reminded me that it could be done in client sunrise or vip-config.  So ignore me :)

fklein-lu:
And back... again. Any further code changes that are needed, please execute yourself. I need this merged before the end of the week. Thanks.

mjangda:
r111393-deploy

PR #895, opened 2018-05-25
Add custom handlers for WordPress 4.9.6 personal data export feature

mjangda:
Core relies on the ability to write files directly to the uploads folder which doesn't currently work on Go. Core also expects the PHP `zip` module to always be available even though it won't for all applications, so we fall back to `PclZip` in those instances.

### TODO

- [x] Test on Go servers
- [x] Local fallback

### Testing

Generate a request; then:

- On a site with `zip` module enabled, make sure the export is generated and downloadable.
- On a site without the `zip` module enabled, make sure the export is generated and downloadable.
- Make sure that archives for requests are being deleted after 3 days and no longer accessible past that time.

### DIFF
--- a/001-core/privacy.php
+++ b/001-core/privacy.php
@@ -168,6 +168,7 @@ function generate_personal_data_export_file( $request_id ) {
 	}
 
 	if ( is_wp_error( $zip_result ) ) {
+		/* translators: %s: error message */
 		$error = sprintf( __( 'Unable to generate export file (archive) for writing: %s' ), $zip_result->get_error_message() );
 	} else {
 		/** This filter is documented in wp-admin/includes/file.php */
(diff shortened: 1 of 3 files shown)


mjangda:
This needs some thorough testing (which I'm working on) and would be good to get some eyes on the code as well. Paging @simonwheatley @joshbetz @chrean.

mjangda:
Worth noting that the code is largely a copy of core's implementation, which is not the cleanest. In the interest of getting this out, I minimized the number of changes even though we could refactor this pretty heavily to make it more readable and testable.

001-core/privacy.php
```diff
@@ -0,0 +1,315 @@
...
+	if ( is_wp_error( $zip_result ) ) {
+		/* translators: %s: error message */
+		$error = sprintf( __( 'Unable to generate export file (archive) for writing: %s' ), $zip_result->get_error_message() );
+	} else {
+		/** This filter is documented in wp-admin/includes/file.php */
+		do_action( 'wp_privacy_personal_data_export_file_created', $archive_pathname, $archive_url, $html_report_pathname, $request_id );
+	}
+
+	// And remove the HTML file.
+	unlink( $html_report_pathname );
```
joshbetz: Probably not a major concern right now, but I wonder if we need to consider what happens if a request times out before it gets here.

mjangda: I moved it a bit earlier in the flow which should help. The timeout, if it happens, is going to be during the zip process which needs the file, so not much we can do there.

We probably need an external cleanup routine for handling those edge cases.

joshbetz: > We probably need an external cleanup routine for handling those edge cases.

Yeah, this is what I was thinking too. I think we can use `find` to list files older than X, so we could probably add a daily cron job to web containers or something.

mjangda:
r111452-deploy

PR #896, opened 2018-05-25
Privacy Compat Fixes

mjangda:
Follow-up from #895 

- [x] Switch to init hook. The delete handler is triggered during cron and need to be initialized in that context, not just admin.
- [ ] ...


mjangda:
r111585-deploy

PR #900, opened 2018-05-28
Prevent healthcheck alerts for site-not-found errors

### LINKED ISSUE #865: Custom Error Page for Multisite "Site Not Found"
We should show a nice, helpful error page as a default when a site cannot be found on a multisite instance and not throw a 500.

Would be good to make sure sites can override this as well.

mjangda:
These can trigger during Data Sync which is undesirable and we should handle it appropriately.

- [x] Account for WP CLI / REST and other contexts
- [x] Account for Maintenance Mode
- [x] Tests

Fixes #865


simonwheatley:
I added this to the "Next" column in the Project board, @mjangda, feel free to move it to Launch Blockers or In Progress as appropriate 🙂

lib/sunrise/sunrise.php
```diff
@@ -41,10 +47,38 @@ function site_not_found( $network, $domain, $path ) {
...
+		header( 'Content-type: text/plain' );
+		echo sprintf( '%s not found; but still OK', esc_html( $error_type ) );
+		exit;
+	}
+
+	$is_web_request = Context::is_web_request();
+	if ( $is_web_request ) {
+		$is_maintenance_mode = Context::is_maintenance_mode();
+		if ( $is_maintenance_mode ) {
+			$status_code = 200;
```
joshbetz: 200 will cause this to be cached. Is that ideal? If we send 503 here, will it trigger health checks?

https://yoast.com/http-503-site-maintenance-seo/

mjangda: Ah, good catch. We're setting a `503` in `000-vip-init.php` as well.

mjangda: > If we send 503 here, will it trigger health checks?

The healthcheck handler fires earlier, so shouldn't be an issue.

mjangda:
r113564-deploy

PR #901, opened 2018-05-28
Switch to `sanitize_file_name`

mjangda:
Works better with unicode characters than our previous `sanitize_title_with_dashes` implementation for sanitization.


mjangda:
r112158-deploy

PR #906, opened 2018-05-31
Deprecates `wpcom_vip_load_wp_rest_api`

emrikol:
This will better allow themes to be migrated from WordPress.com without causing a fatal error.

### DIFF
--- a/vip-helpers/vip-deprecated.php
+++ b/vip-helpers/vip-deprecated.php
@@ -1,6 +1,15 @@
 <?php
 
 /**
+ * Loads the built-in WP REST API endpoints in WordPress.com VIP context.
+ *
+ * @deprecated Not applicable since VIP 2.0.0
+ */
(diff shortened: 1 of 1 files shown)


vip-helpers/vip-deprecated.php
```diff
@@ -1,5 +1,14 @@
 <?php
 
+/**
+ * Loads the built-in WP REST API endpoints in WordPress.com VIP context.
+ *
+ * @deprecated Not applicable since VIP 2.0.0
+ */
+function wpcom_vip_load_wp_rest_api() {
```
mjangda: Can we move this to https://github.com/Automattic/vip-go-wpcom-compat instead? And make sure all migrating sites load that plugin?

emrikol: Yep!  I'll close this out and make a PR there.  Thanks :)

PR #907, opened 2018-06-04
Allow VaultPress registration to fail gracefully

scarstocea:
Allow VaultPress registration to fail gracefully, but still output the error

### DIFF
--- a/vaultpress/class.vaultpress-cli.php
+++ b/vaultpress/class.vaultpress-cli.php
@@ -18,7 +18,7 @@ class VaultPress_CLI extends WP_CLI_Command {
 	public function register_via_jetpack() {
 		$result = VaultPress::init()->register_via_jetpack( true );
 		if ( is_wp_error( $result ) ) {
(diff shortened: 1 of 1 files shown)


scarstocea:
This is for running the sync from CLI to start to finish – otherwise this fatal error will occur during the cleanup action on some sites, which causes the entire action to fail.

We're still debugging the VaultPress issues which cause this, but some clients require weekly syncs and their cleanup actions take a lot of time when ran manually, so we'd rather have this work for now (even as a stop gap measure).

mjangda:
As discussed in Slack, let's go ahead and release this since it's breaking sync. It does modify the VaultPress plugin directly so we'll have to be careful not to overwrite the change with future releases (and work with the VP team to see if we can upstream this or something similar).

mjangda:
r112157-deploy

scarstocea:
Thanks, Mo! I'll chat with the VaultPress devs about upstreaming it.

PR #910, opened 2018-06-07
Don't silently handle missing plugins on VIP environments

### LINKED ISSUE #883: Don't silently handle missing plugins on VIP environments
If `wpcom_vip_load_plugin()` is called with a file name that doesn't exist, it exits with an error message, but only on non-VIP environments. Ref: https://github.com/Automattic/vip-go-mu-plugins/blob/95a3597ecc988e846f9dcaf4c8d8ed693c79098e/vip-helpers/vip-utils.php#L1061-L1063

If the same situation is encountered on a VIP environment then the problem is silently discarded. If, for example, the theme on the site directly calls functions from the missing plugin then it'll raise a fatal error, but if it has guard conditions in place then everything will fail silently.

As a safeguard, I would like this condition to raise a PHP warning on VIP environments. Thoughts?

chrean:
Fixes #883

Question: since New Relic extension is not loaded on sandboxes I could not test it properly. Is there a way to test this on non-production environments?
The rest of the code seems to work and it's fairly trivial, but I would prefer testing it anyway.

### DIFF
--- a/vip-helpers/vip-utils.php
+++ b/vip-helpers/vip-utils.php
@@ -1058,8 +1058,16 @@ function wpcom_vip_load_plugin( $plugin = false, $folder = false, $load_release_
 
 		return _wpcom_vip_include_plugin( $includepath );
 	} else {
+		$error_msg = "Unable to load plugin '$plugin' using wpcom_vip_load_plugin()!";
 		if ( ! WPCOM_IS_VIP_ENV ) {
-			die( "Unable to load $plugin using wpcom_vip_load_plugin()!" );
+			die( $error_msg );
+		} else {
+			// On VIP we try to both notify the user...
+			trigger_error( $error_msg, E_USER_WARNING );
+			// ...And trigger a New Relic notice, if the extension is available
+			if ( extension_loaded( 'newrelic' ) && function_exists( 'newrelic_notice_error' ) ) {
+				newrelic_notice_error( $error_msg );
+			}
 		}
 	}
 }
--- a/vip-helpers/vip-utils.php
+++ b/vip-helpers/vip-utils.php
@@ -1058,7 +1058,7 @@ function wpcom_vip_load_plugin( $plugin = false, $folder = false, $load_release_
 
 		return _wpcom_vip_include_plugin( $includepath );
 	} else {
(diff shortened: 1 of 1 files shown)


vip-helpers/vip-utils.php
```diff
@@ -1058,8 +1058,16 @@ function wpcom_vip_load_plugin( $plugin = false, $folder = false, $load_release_
 
 		return _wpcom_vip_include_plugin( $includepath );
 	} else {
+		$error_msg = "Unable to load plugin '$plugin' using wpcom_vip_load_plugin()!";
```
mjangda: It might be good to fix up the error a bit:

```
$error_msg = sprint( 'wpcom_vip_load_plugin: Unable to load plugin `%s`; the path `%s` does not exist.', $plugin, $includepath );
```

vip-helpers/vip-utils.php
```diff
@@ -1058,8 +1058,16 @@ function wpcom_vip_load_plugin( $plugin = false, $folder = false, $load_release_
...
 		if ( ! WPCOM_IS_VIP_ENV ) {
-			die( "Unable to load $plugin using wpcom_vip_load_plugin()!" );
+			die( $error_msg );
+		} else {
+			// On VIP we try to both notify the user...
+			trigger_error( $error_msg, E_USER_WARNING );
+			// ...And trigger a New Relic notice, if the extension is available
+			if ( extension_loaded( 'newrelic' ) && function_exists( 'newrelic_notice_error' ) ) {
```
mjangda: Aside: I wonder if there's value in abstracting this into a separate logging library / function (i.e. which calls both `trigger_error` and `newrelic_notice_error`)? Out of scope for this PR so no change needed, but might be good to open an issue and discuss what that could look like.

mjangda:
> Is there a way to test this on non-production environments

Unfortunately, not at the moment. I think we can trust that the new relic extension will do its job.

vip-helpers/vip-utils.php
chrean: Good point. Done.

vip-helpers/vip-utils.php
chrean: https://github.com/Automattic/vip-go-mu-plugins/issues/913

mjangda:
r112762-deploy

Issue #904, opened 2018-05-31
No index non production sites with a custom domain


scottsweb:
By default all *.go-vip.co domains have a no-index HTTP header automagically added. If the non-production environment uses a different domain this does not happen and protection needs to be added explicitly.

Possible fix:

```
<?php 
 
if ( 'production' !== VIP_GO_ENV ) {
    header( 'X-Robots-Tag: noindex, nofollow', true );
}
```

chrean:
Adding refs for tracking purposes:

- https://veetoop2.wordpress.com/2018/05/24/id-like-to-remove-the/
- https://veetoop2.wordpress.com/2018/05/24/smart-defaults-after-data-sync/

simonwheatley:
> No index non production sites with a custom domain

I think we could try doing this for all domains, not just custom domains.

chrean:
Merged https://github.com/Automattic/vip-go-mu-plugins/pull/911

chrean:
r112439-deploy

PR #911, opened 2018-06-07
No index non production sites with a custom domain

chrean:
Discussed here: https://github.com/Automattic/vip-go-mu-plugins/issues/904


joshbetz:
r112439-deploy

mjangda:
One small downside with this implementation is that it will add the tag to local development environments as well. Not sure if that's an issue but could be unexpected for some devs?

chrean:
Can't see why a dev may want to index a local environment, however I agree, it may be unexpected and perhaps confusing?

If so, we may add a check on `WPCOM_IS_VIP_ENV` e.g. 

```
if ( 'production' !== VIP_GO_ENV && WPCOM_IS_VIP_ENV ) {
                $headers['X-Robots-Tag'] = 'noindex, nofollow';
}

mjangda:
Could also do `false !== VIP_GO_ENV && 'production' !== VIP_GO_ENV`.

This also feels dangerous and would be good to add some tests :)

Issue #908, opened 2018-06-04
(PHP7) wpcom-error-handler: throws Array to String when certain notices are handled


rogertheriault:
This line
https://github.com/Automattic/vip-go-mu-plugins/blob/master/lib/wpcom-error-handler/wpcom-error-handler.php#L60
Will throw a warning on Array to String conversion when the argument is an array - previous checks only look for objects.

Seen on multiple VIP Go sites in conjunction with errors, particularly "Cannot modify header". e.g. 
```
PHP Notice:  Array to string conversion in wp-content/mu-plugins/lib/wpcom-error-handler/wpcom-error-handler.php on line 60
```

This is likely only happening on PHP 7.

PR #912, opened 2018-06-07
Update Error Handler

mjangda:
Two related fixes:

- Properly render class methods
- Handle array args for filters (to prevent an array-conversion notice)

### DIFF
--- a/lib/wpcom-error-handler/wpcom-error-handler.php
+++ b/lib/wpcom-error-handler/wpcom-error-handler.php
@@ -53,7 +53,9 @@ function wpcom_get_error_backtrace( $last_error_file, $last_error_type, $for_irc
 			$path .= isset( $call['line'] ) ? ':' . $call['line'] : '';
 		}
 
-		if ( in_array( $call['function'], array( 'do_action', 'apply_filters' ) ) ) {
+		if ( isset( $call['class'] ) ) {
+			$path .= " {$call['class']}{$call['type']}{$call['function']}()";
+		} elseif ( in_array( $call['function'], array( 'do_action', 'apply_filters' ) ) ) {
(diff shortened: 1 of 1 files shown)


lib/wpcom-error-handler/wpcom-error-handler.php
```diff
@@ -53,9 +53,13 @@ function wpcom_get_error_backtrace( $last_error_file, $last_error_type, $for_irc
 			$path .= isset( $call['line'] ) ? ':' . $call['line'] : '';
 		}
 
-		if ( in_array( $call['function'], array( 'do_action', 'apply_filters' ) ) ) {
+		if ( isset( $call['class'] ) ) {
+			$path .= " {$call['class']}{$call['type']}{$call['function']}()";
```
rogertheriault: Nice.

Considering how many different combinations of params there might be, it might be worthwhile adding a check for type, or maybe preceding this line with
`$c_type = isset( $call['type'] ) ? $call['type'] : '???';`

lib/wpcom-error-handler/wpcom-error-handler.php
```diff
@@ -53,9 +53,13 @@ function wpcom_get_error_backtrace( $last_error_file, $last_error_type, $for_irc
...
+		} elseif ( in_array( $call['function'], array( 'do_action', 'apply_filters' ) ) ) {
 			if ( is_object( $call['args'][0] ) && ! method_exists( $call['args'][0], '__toString' ) ) {
 				$path .= " {$call['function']}(Object)";
+			} elseif ( is_array( $call['args'][0] ) ) {
+				$path .= " {$call['function']}(Array)";
```
rogertheriault: Thank you! Fixes #908

mjangda:
r112455-deploy

PR #915, opened 2018-06-12
Only Enforce Restricted Usernames on VIP Environments

mikeselander:
Currently, the restricted usernames functionality enforces usernames on every single environment that the mu-plugins library is used on, including local. However, quite a lot of local development environments use the admin username by default, which makes it impossible to login with a default configuration. Chassis, in particular, sets up WordPress with a super-admin `admin` user by default and it is a recommended local environment for developing on VIP.

This simple change only enforces the restricted usernames when on a WPCOM_VIP environment instead of doing so indiscriminately.

P.S. I know I opened this up without any discussion in an Issue, I'm happy to open one up for this topic if preferred as a first step.

### DIFF
--- a/security.php
+++ b/security.php
@@ -41,7 +41,10 @@ add_action( 'wp_login', 'wpcom_vip_login_limiter_on_success', 10, 2 );
 
 function wpcom_vip_limit_logins_for_restricted_usernames( $user, $username, $password ) {
 	$is_restricted_username = wpcom_vip_is_restricted_username( $username );
-	if ( $is_restricted_username ) {
+	if (
+		( defined( 'WPCOM_IS_VIP_ENV' ) && WPCOM_IS_VIP_ENV )
+		&& $is_restricted_username
+	) {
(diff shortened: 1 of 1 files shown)


mjangda:
We're trying to keep differences between Go servers and local environments to a minimum and unfortunately, this change doesn't help with that.

Chassis allows you to set a different username for the default user (http://docs.chassis.io/en/latest/config/#default-admin). Should we recommend that path instead?

chrisscott:
I somehow missed this in searching before opening #921 so that's a dup of this. I've pushed an update in https://github.com/Automattic/vip-go-mu-plugins/tree/921_allow_restricted_usernames_outside_vip but held off on a PR after seeing this. I went about it a bit differently and added a test but functionally it looks like it does the same thing.

@mjangda there's another use case for this in my issue which is that of content imports locally that already use the `admin` user (and could potentially happen if there's a `vip` user) or plain vanilla setups where `admin` is used before adding these mu plugins.

mjangda:
Going to close this out for now based on my earlier point:

> We're trying to keep differences between Go servers and local environments to a minimum

PR #923, opened 2018-06-19
Fix plugin load warning

mjangda:
In some cases, `$includepath` may not be defined leading to an incorrect error:

```
wpcom_vip_load_plugin: Unable to load plugin `photonfill`; the path `` does not exist.
```

Instead, we selectively build the error message based on the context.

### DIFF
--- a/vip-helpers/vip-utils.php
+++ b/vip-helpers/vip-utils.php
@@ -1058,7 +1058,13 @@ function wpcom_vip_load_plugin( $plugin = false, $folder = false, $load_release_
 
 		return _wpcom_vip_include_plugin( $includepath );
 	} else {
-		$error_msg = sprintf( 'wpcom_vip_load_plugin: Unable to load plugin `%s`; the path `%s` does not exist.', $plugin, $includepath );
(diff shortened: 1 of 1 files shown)


vip-helpers/vip-utils.php
```diff
@@ -1058,7 +1058,13 @@ function wpcom_vip_load_plugin( $plugin = false, $folder = false, $load_release_
 
 		return _wpcom_vip_include_plugin( $includepath );
 	} else {
-		$error_msg = sprintf( 'wpcom_vip_load_plugin: Unable to load plugin `%s`; the path `%s` does not exist.', $plugin, $includepath );
+		$error_msg = sprintf( 'wpcom_vip_load_plugin: Unable to load plugin `%s`', $plugin );
+		if ( $includepath ) {
```
chrean: If the variable is not defined, this will throw an error.
Or, is it always defined but sometimes empty?

mjangda: Defined as `null` here: https://github.com/Automattic/vip-go-mu-plugins/pull/923/files#diff-fb296e3bc7c2a02335f61196cce2f90fR1026

mjangda:
r113354-deploy

PR #925, opened 2018-06-20
replace deprecated wpcom_initiate_flush_rewrite_rules call

paulschreiber:
replace wpcom_initiate_flush_rewrite_rules() with flush_rewrite_rules().

Fixes `PHP Notice:  wpcom_initiate_flush_rewrite_rules is <strong>deprecated</strong> since version 2.0.0! Use flush_rewrite_rules instead. in /srv/www/theundefeated/public_html/wp-includes/functions.php on line 3840`


mjangda:
r113352-deploy

PR #928, opened 2018-06-28
Use Singleton for A8C_ Files class

kjbenk:


mjangda:
Can you please provide some background on this PR? What are you trying to accomplish?

kjbenk:
Since this class is not a singleton it is very hard to call functions within it without adding duplicate hooks and filters. The hooks and filters are attached on the `__construct` method which means that they are added each time a new instance is created. This can lead to problems in the future and using a Singleton will solve those problems.

ethitter:
We're working on a project that uses S3 for uploads, as access controls are necessary. Since we can't use standard PHP functions like `file_exists()` or `unlink()`, we need access to equivalent methods in `A8C_Files` without instantiating multiple instances of the class.

For our particular case, after offloading the file to S3, it needs to be removed from Go's filesystem but should remain in the media library. With access to `A8C_Files`' `attachment_file_exists()` and `delete_file()` methods, we can leverage an existing S3 solution without needing to duplicate logic from `A8C_Files`.

mjangda:
Thanks for the added context. As discussed, `A8C_Files` is largely meant to be an internal / private class so it's a not good fit for direct use. The `Api_Client` or `WP_Filesystem_VIP` classes would be better options.

PR #932, opened 2018-07-09
Fix multisite not found error message

joshbetz:
The site not found page for multisites can be confusing. We should make it more clear that it's a multisite, but the domain is not mapped to a site.


mjangda:
r114412-deploy

PR #939, opened 2018-07-17
Add fallbacks for new VIP_GO_APP_* constants

mjangda:
These new constants will be available for use on Go sites soon and will be defined early in `wp-config`.

This PR adds fallbacks for local development.


mjangda:
Some of these are going to change, so needs a bit of work.

rinatkhaziev:
👋 @mjangda what do you want to do with this?

mjangda:
The VIPd issue needs be sorted out first. Feel free to close this if we want to keep things clean.

PR #941, opened 2018-07-18
Change a8c-file interactions to use  API_Client functions rather than direct curl calls

### LINKED ISSUE #933: Update A8C Files to use Api_Client
The API interactions in [`a8c-files.php`](https://github.com/Automattic/vip-go-mu-plugins/blob/master/a8c-files.php) should be updated to use our [Api_Client class](https://github.com/Automattic/vip-go-mu-plugins/blob/master/files/class-api-client.php) (i.e. instead of the direct `curl` calls).

- [x] `A8c_Files::_check_uniqueness_with_backend()`
- [x] `A8c_Files::attachment_file_exists()`
- [x] `A8c_Files::upload_file()`
- [x] `A8c_Files::delete_file()`

Would also be great to add some test coverage for these methods as well.

dchymko:
Using API Client interactions makes things more standardized, modular, and easier to test.
Implemented unique_filename API call in API_Client & added tests.

Fixes #933


a8c-files.php
```diff
@@ -381,60 +367,23 @@ function upload_file( $details, $upload_type ) {
...
-					if ( isset(  $obj->filename ) && basename( $obj->filename ) != basename( $post_url ) ) {
-						$uploads = wp_upload_dir();
-						if ( false === $uploads['error'] ) {
-							@copy( $details['file'], $uploads['path'] . '/' . $obj->filename );
-							register_shutdown_function( 'unlink', $uploads['path'] . '/' . $obj->filename );
-						}
-						$details['file'] = str_replace( basename( $post_url ), basename( $obj->filename ), $details['file'] );
-					}
+		if ( is_wp_error( $upload_result ) ) {
+			$details['error'] = $upload_result->get_error_message();
```
mjangda: Might be cleaner to just return early here and get rid of the `else`.

a8c-files.php
```diff
@@ -381,60 +367,23 @@ function upload_file( $details, $upload_type ) {
...
+		} else {
+			$result_filename = $upload_result;
+			if ( isset( $result_filename ) && basename( $result_filename ) != basename( $post_url ) ) {
```
mjangda: Is this block still necessary?

tests/test-a8c-files-utils.php
```diff
@@ -1,6 +1,48 @@
...
+	public function setUp() {
+		parent::setUp();
+	}
+
+	public function tearDown() {
+		$this->api_client = null;
+		parent::tearDown();
+	}
+
+	private function setup_mock_a8c_files( $method_name, $expected ) {
```
mjangda: Minor: Based on the name alone, this seems like it's mocking `a8c_files`. Maybe `setup_a8c_files_and_mock_client` is a bit clearer (even though it's longer :))

tests/test-a8c-files-utils.php
```diff
@@ -1,6 +1,48 @@
 <?php
 
 class VIP_Go_A8C_Files_Utils_Test extends WP_UnitTestCase {
```
mjangda: These tests are technically meant for the `A8C_Files_Utils` class. Should we have a new testcase class for our new tests?

files/class-api-client.php
```diff
@@ -70,7 +70,22 @@ private function call_api( $path, $method, $request_args = [] ) {
...
+		] );
+
+		if ( is_wp_error( $response ) ) {
+			return $response;
+		}
+		
+		$response_code = wp_remote_retrieve_response_code( $response );
+		$response_content   = wp_remote_retrieve_body( $response );
+
+		return ['http_code' => $response_code, 'content' => $response_content ];
```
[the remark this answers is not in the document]
dchymko: I was mimicking how the existing methods in `A8C_files` were working. It checks for the response code in `filter_unique_filename()` but we could refactor that to look for a boolean or WP_Error coming back.

a8c-files.php
dchymko: I think you are correct. At first glance, it looked like it was doing more with $post_url but I think most of that could be refactored out. I was a bit concerned with breaking existing functionality since the existing logic above and below the curl call was a bit tricky to get a complete understanding of. I think some tests around that could help with the understanding and refactoring.

tests/test-a8c-files-utils.php
dchymko: Yeas, I thought about that. Also the fact that in some of tests I'm mocking an object that I pass _another_ mock object into. :)

tests/test-a8c-files-utils.php
dchymko: Right, good point, I'll split them off.

a8c-files.php
```diff
@@ -298,6 +297,10 @@ public function filter_filetype_check( $filetype_data, $file, $filename, $mimes
 
 		$check = $this->_check_uniqueness_with_backend( $filename );
 
+		if ( is_wp_error( $check ) ) {
+			return $filetype_data;
```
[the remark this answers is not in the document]
dchymko: I changed this to return the error instead.

a8c-files.php
```diff
@@ -25,7 +25,9 @@
 
 class A8C_Files {
 
-	function __construct() {
+	private $api_client;
+
+	function __construct($api_client = null) {
```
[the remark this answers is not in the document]
dchymko: We're only calling this from the init method in the very same file so I removed the null fallback.

a8c-files.php
```diff
@@ -447,30 +396,15 @@ function delete_file( $file_name ) {
...
-		curl_setopt( $ch, CURLOPT_TIMEOUT, 10 );
-
-		curl_exec( $ch );
-		$http_code = curl_getinfo( $ch, CURLINFO_HTTP_CODE );
-		curl_close( $ch );
-
-		if ( 200 != $http_code ) {
-			error_log( sprintf( __( 'Error deleting the file from the remote servers: Code %d' ), $http_code ) );
+		if ( is_wp_error( $delete_result ) ) {
+			error_log( $delete_result->get_error_message() );
```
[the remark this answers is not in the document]
dchymko: We may want to do something different in the case of a WP_Error coming back from `delete_file` but currently this method always returns null

mjangda:
Closing; most of a8c-files will be going away with the Stream Wrapper implementation.

PR #945, opened 2018-07-30
Remove Try Gutenberg callout introduced as part of 4.9.8

justnorris:
<strike>
**What does it do?**
Removes the following Gutenberg Callout introduced in Core WordPress 4.9.8

screen shot 2018-07-27 at 6 44 39 pm

**But why?**
VIP Go doesn't allow direct plugin installations, so the "Install Plugin" action leads to an error page. There are no filters in the Core to remove/modify the action buttons, so we have to remove the whole message.

**Has this been tested?**
Yes, tested with a local VIP Go WordPress 4.9.8-RC2 installation

**How can I test this?**
To make sure everything is in order, you can test it yourself by applying this patch and making sure that the Gutenberg notification doesn't appear in WordPress 4.9.8 when the patch is applied. 
</strike>

---

Based on some further discussion, the original premise of the PR was flawed as the banner does to appropriate capability checks. We're now keeping the banner as-is, but changing the "Learn More" link to point to the VIP Docs instead.

### To Test

- Download and install 4.9.8 RC3
- Apply this PR
- Verify that the "Try" Banner shows up but links to the VIP site.


001-core.php
```diff
@@ -27,3 +27,8 @@ function wpcom_vip_disable_core_update_cap( $caps, $cap ) {
 	}
 	return $caps;
 }
+
+/**
+ * Remove Try Gutenberg callout introduced as part of 4.9.8
+ */
+remove_action( 'try_gutenberg_panel', 'wp_try_gutenberg_panel' );
```
mjangda: Should we move this to the Ramp loader file?

justnorris: This is core related, specific to VIP Go and core compatibility. I think it's better not to mix Ramp in when possible to avoid confusion.

mjangda:
r115741-deploy

PR #948, opened 2018-08-01
Added/improved header section for some MU plugins 

TheCrowned:
Some MU plugins lack proper name/description and thus show up in the plugins view just with their slugs, as shown below:
screenshot from 2018-08-01 09-13-44

This PR aims at fixing them, providing all MU plugins with an apt header section. This is particularly important since we are going to show those in the general plugins view, as per #947.

### DIFF
--- a/000-vip-init.php
+++ b/000-vip-init.php
@@ -1,8 +1,13 @@
 <?php
 
 /**
- * Initialisation for various VIP functionality
- *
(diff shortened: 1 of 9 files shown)


mjangda:
Thanks for fixing these up @TheCrowned!

mjangda:
r116148-deploy

PR #949, opened 2018-08-07
Bump http-concat to 10025c4

mjangda:
Brings over fix to print extra/localized JS: https://github.com/Automattic/nginx-http-concat/pull/40


mjangda:
r116079-deploy

PR #950, opened 2018-08-08
Don't override Cache-Control headers in REST API

mjangda:
If the header is set via `header()` we should not override it.

Previously, we were only checking if the Cache-Control header was set via the REST Response object. Now, we look at both.

## To Test

- Set `header( 'Cache-Control: max-age=404' );` somewhere in your pageload cycle and make a request to a `wp-json` endpoint.
- Verify that the `Cache-Control` header is set to `max-age=404`.

Then:

- Remove the `header()` call and make the request again. Verify that the `Cache-Control` header now has `max-age=60`.

### DIFF
--- a/vip-cache-manager/ttl-manager.php
+++ b/vip-cache-manager/ttl-manager.php
@@ -27,6 +27,15 @@ function enforce_rest_api_read_ttl( $response, $rest_server, $request ) {
 		return $response;
 	}
 
+	// Don't override existing Cache-Control headers sent via PHP
+	$php_headers = headers_list();
+	foreach ( $php_headers as $header ) {
+		if ( 0 === stripos( $header, 'Cache-Control:' ) ) {
+			return $response;
+		}
+	}
+
+	// Don't override existing Cache-Control headers set via REST Response
 	$response_headers = $response->get_headers();
 	if ( isset( $response_headers['Cache-Control'] ) ) {
 		return $response;
--- a/tests/test-cache-ttl-manager-rest.php
+++ b/tests/test-cache-ttl-manager-rest.php
@@ -106,7 +106,7 @@ class TTL_Manager__REST_API__Test extends \WP_Test_REST_TestCase {
 		$this->assertArrayNotHasKey( 'Cache-Control', $response_headers );
 	}
 
-	public function test__skip_ttl_if_already_set() {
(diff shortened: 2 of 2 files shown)


vip-cache-manager/ttl-manager.php
```diff
@@ -27,6 +27,16 @@ function enforce_rest_api_read_ttl( $response, $rest_server, $request ) {
...
 	}
 
+	// Don't override existing Cache-Control headers sent via PHP
+	$php_headers = headers_list();
+	foreach ( $php_headers as $header ) {
+		if ( 0 === stripos( $header, 'Cache-Control' ) ) {
+			return $response;
+		}
+	}
+	unset( $header );
```
nickdaugherty: We only need to explicitly `unset()` if the value is passed by reference in the loop.

vip-cache-manager/ttl-manager.php
```diff
@@ -27,6 +27,16 @@ function enforce_rest_api_read_ttl( $response, $rest_server, $request ) {
 		return $response;
 	}
 
+	// Don't override existing Cache-Control headers sent via PHP
+	$php_headers = headers_list();
+	foreach ( $php_headers as $header ) {
+		if ( 0 === stripos( $header, 'Cache-Control' ) ) {
```
nickdaugherty: Should this be a stricter check?

```
if ( strtolower( $header ) === 'cache-control' ) {
```

That way we don't have false positive matches with things like `Cache-Control-Something-Else` (which could conceivably exist in the future at some point, or be a custom header from client).

mjangda: `$header` will look like `Cache-Control: max-age=60` so we can't do an exact check as suggested.

It's a good point though. `stripos` takes care of the case-insensitive matching for us already and to make it more specific, I'm now checking for `Cache-Control:` (with the colon).

vip-cache-manager/ttl-manager.php
mjangda: Done.

vip-cache-manager/ttl-manager.php
nickdaugherty: Ah ok, adding the colon will achieve the same. 👍

Issue #867, opened 2018-04-13
Vaultpress - Deprecated function - PHP 7.2


rodruiz:
Hi,
PHP 7.2 was [Deployed on VIP Go](https://lobby.vip.wordpress.com/2018/04/06/php-7-2-deployed-on-vip-go/). I'm using VVV for local development and updated the local environment to 7.2 as well.
```error
PHP Deprecated: The each() function is deprecated. This message will be suppressed on further calls in /srv/www/wordpress-default/public_html/wp-content/mu-plugins/vaultpress/cron-tasks.php on line 111
```
See: [Line 111](https://github.com/Automattic/vip-go-mu-plugins/blob/master/vaultpress/cron-tasks.php#L111)

See [PHP documentation](http://php.net/manual/en/function.each.php) and [deprecations_php_7_2#each](https://wiki.php.net/rfc/deprecations_php_7_2#each)

Thank you.

GaryJones:
This appears to now be fixed.

PR #953, opened 2018-08-08
Rewrite code that uses deprecated function each().

smistephen:
As of PHP 7.2, `each()` is a deprecated function, so it is being replaced with a similar but non-deprecated algorithm:
 - extract the first array key (`key()`)
 - extract the first array value (`current()`)
 - advance the array pointer (`next()`)

This is related to #867.


mjangda:
Can we submit this upstream to the VaultPress team first?

mjangda:
Closing; fix should happen upstream.

PR #955, opened 2018-08-09
Change default behaviour of wpcom_vip_load_gutenberg() function.

smistephen:
Currently, calling `wpcom_vip_load_gutenberg()` makes `$criteria` default to `false`. That means calling `wpcom_vip_load_gutenberg()` does not load Gutenberg. This needs to change.

The current state of the code has led to at least three instances of clients thinking they have enabled Gutenberg when they have not, and filing high priority support tickets.

The new version of this code sets `$criteria` to `true` by default, making for a more logical connection between function name and function behaviour.


mjangda:
r116224-deploy

PR #957, opened 2018-08-10
Hotfix: HTTP Concat for Gutenberg

justnorris:
**The Why**

HTTP Concat plugin currently breaks Gutenberg Admin. This is a hot fix that will selectively disable HTTP Concat in the admin when editing posts that have Gutenberg enabled.

This change will not affect people that have defined `VIP_GO_DISABLE_RAMP` as true.

**Testing**
1. Activate Gutenberg on a fresh VIP GO Installation
2. Add this to the active theme functions.php:
```
wpcom_vip_load_gutenberg( [ 'post_types' => [ 'page' ] ] );
```
3. Reload the Page edit page twice, observe a white screen of death ( JavaScript errors in console )
4. Apply this patch
5. Problem fixed - Gutenberg should now work in "Add new Page".
6. Visit "Add new post" - observe concatenated scripts because Gutenberg is only enabled in "page" post type.


gutenberg-ramp.php
```diff
@@ -38,5 +38,26 @@ function wpcom_vip_load_gutenberg( $criteria = true ) {
 	if ( ! function_exists( 'gutenberg_ramp_load_gutenberg' ) ) {
 		return;
 	}
+
 	gutenberg_ramp_load_gutenberg( $criteria );
+
+	if ( ! has_action( 'admin_init', 'wpcom_vip_disable_gutenberg_concat' ) ) {
```
mjangda: Why do we need this check?

mjangda:
r116279-deploy

PR #959, opened 2018-08-14
Add Pull Request template

mjangda:


pkevan:
Looks like a good start for this. Perhaps the only improvement at this stage could be having the test plan in sequential numbers 😉

mjangda:
We expect all the steps to be done at the same time. Efficiency!

Issue #961, opened 2018-08-27
Add an 'A11n Only' note to the Purge Cache button


nickdaugherty:
The button lives in WP Admin dashboard but is only accessible to Automatticians, so it should be labeled as such to make it clear to ourselves that end-clients don't have access to the button.

https://github.com/Automattic/vip-go-mu-plugins/blob/da7fcd24b773e1957687c6eb56f0b56ddd402d9e/vip-cache-manager/vip-cache-manager.php#L57-L70

WPprodigy:
Located here (TIL): 
![dashboard](https://user-images.githubusercontent.com/8536129/44708094-d9607500-aa6b-11e8-8351-891785a2f3cc.png)

PR #962, opened 2018-08-28
Update “purge page cache” button description

WPprodigy:
Closes https://github.com/Automattic/vip-go-mu-plugins/issues/961

Update the text to be more clear about what it does and that it will mostly only appear for a12s.

Updated:
updated


mjangda:
r117299-deploy

PR #964, opened 2018-09-05
Adding a filter to prevent async post status event from being scheduled

CodeProKid:
## Description
We have been having some issues with too many cron events being scheduled with posts getting published. This is happening because we use custom post types as a storage mechanism for various things that don't contain any public information.

To prevent this from happening I've added a filter here where we can conditionally set the value to `false` and it will prevent the cron event from being scheduled.

For our implementation, we plan on doing something like:
```php
add_filter( 'wpcom_async_transition_post_status_schedule_async', function( $value, $args ) {
    if ( 'my-post-type' === get_post_type( $args['post_id'] ) ) {
        $value = false;
    }
    return $value;
}, 10, 2 );
```

## Checklist

## Steps to Test

1. Check out PR.
2. Add the above code (replacing CPT name with `page`)
3. Publish a new page
4. No event shout have gotten scheduled
5. 💰

### DIFF
--- a/async-publish-actions.php
+++ b/async-publish-actions.php
@@ -122,6 +122,16 @@ function _queue_async_hooks( $new_status, $old_status, $post ) {
 		'old_status' => $old_status,
 	];
 
+	/**
+	 * Filter whether or not a cron event should be scheduled when the post transitions status.
+	 *
+	 * @param bool  $true Whether or not to schedule the cron event.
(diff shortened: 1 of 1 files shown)


mjangda:
Deployed (r117707-deploy)

mjangda:
Updated docs: https://vip.wordpress.com/documentation/vip-go/asynchronous-publishing-actions-on-vip-go/

PR #965, opened 2018-09-05
Prevent Notice when calling `nocache_headers()` super early

### LINKED ISSUE #935: Calling nocache_headers() before the main query triggers PHP Notice
```
Notice: is_404 was called incorrectly. Conditional query tags do not work before the query is run. Before then, they always return false. Please see <a href="https://codex.wordpress.org/Debugging_in_WordPress">Debugging in WordPress</a> for more information. (This message was added in version 3.1.0.)
```

The issue is from `wpcom_vip_check_for_404_and_remove_cache_headers`, which calls `is_404()` without checking if it's safe to.

shantanu2704:
## Description

Added a check to see if `global $wp_query` is set before using `is_404()` to avoid the following notice:
```
Notice: is_404 was called incorrectly. Conditional query tags do not work before the query is run. Before then, they always return false. Please see <a href="https://codex.wordpress.org/Debugging_in_WordPress">Debugging in WordPress</a> for more information. (This message was added in version 3.1.0.)
```

Fixes #935

## Checklist

## Steps to Test

1. Check out PR.
2. Add the following code snippet to a PHP file in `client-mu-plugins` directory in your sandbox
```
if ( isset( $_GET['dont-cache'] ) ) {
    nocache_headers();
}
```
3. Visit a page that you know will throw a 404 error with the following appended to the URL:
`/?dont-cache=1`
4. With the fix, the aforementioned notice will not be seen.
5. Also, confirm the contents of the `$headers` array in `misc.php`

### DIFF
--- a/tests/test-misc.php
+++ b/tests/test-misc.php
@@ -1,0 +1,67 @@
+<?php
+
+class VIP_Go_Misc_Test extends WP_UnitTestCase {
+
+	public static function setUpBeforeClass() {
+		parent::setUpBeforeClass();
+		require_once( __DIR__ . '/../misc.php' );
+	}
+
+	/**
+	 * Test cache headers removal if page is 404.
+	 *
+	 * @dataProvider data_provider
+	 * @param bool $is_wp_query_set Is the global $wp_query object set for a particular test case.
+	 * @param bool $is_404_set Is this page a 404.
+	 */
+	function test_check_for_404_and_remove_cache_headers( $is_wp_query_set, $is_404_set ) {
+		global $wp_query;
+		$headers = array(
+			'Expires'       => 1,
+			'Cache-Control' => 1,
+			'Pragma'        => 1,
+		);
+
+		$temp = null;
+		if ( $is_404_set ) {
+			$wp_query->set_404();
+		}
+		if ( ! $is_wp_query_set ) {
+			$temp     = $wp_query;
+			$wp_query = null;
+		}
+
+		$result = wpcom_vip_check_for_404_and_remove_cache_headers( $headers );
+		$this->assertEquals( empty( $result ), ( $is_404_set && $is_wp_query_set ) );
+		if ( ! isset( $wp_query ) ) {
+			$wp_query = $temp;
+		}
+	}
+
+	/**
+	 * The data provider method
+	 *
+	 * @return array
+	 */
+	function data_provider() {
+		return array(
+			array(
+				true,
+				true,
+			),
+			array(
+				true,
+				false,
+			),
+			array(
+				false,
+				false,
+			),
+			array(
+				false,
+				true,
+			),
+		);
+	}
+
+}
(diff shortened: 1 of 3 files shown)


tests/test-misc.php
```diff
@@ -0,0 +1,67 @@
+<?php
+
+class VIP_Go_Misc_Test extends WP_UnitTestCase {
```
mjangda: I wonder if we should just make this a class + file specific to the `check_for_404_and_remove_cache_headers` function. That way, we can utilize `setUp` and/or `tearDown` functions instead of relying on cleanup within the test itself.

tests/test-misc.php
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	/**
+	 * The data provider method
+	 *
+	 * @return array
+	 */
+	function data_provider() {
+		return array(
+			array(
```
mjangda: Can we add labels for the various conditions? i.e.

```
return [
     'is_404-and-wp_query-set' => [

    ],

    'not-is_404-and-wp_query-set' => [
        // ...
    ],
    // ...
];
```

tests/test-misc.php
```diff
@@ -0,0 +1,67 @@
...
+		if ( $is_404_set ) {
+			$wp_query->set_404();
+		}
+		if ( ! $is_wp_query_set ) {
+			$temp     = $wp_query;
+			$wp_query = null;
+		}
+
+		$result = wpcom_vip_check_for_404_and_remove_cache_headers( $headers );
+		$this->assertEquals( empty( $result ), ( $is_404_set && $is_wp_query_set ) );
```
mjangda: It would be better to actually check the explicit headers rather than just checking if empty or not. The data provider can pass in the input headers and expected headers as well.

tests/test-misc.php
```diff
@@ -0,0 +1,67 @@
...
+		if ( ! isset( $wp_query ) ) {
```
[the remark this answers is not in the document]
mjangda: I think we should always try to revert state changes. I think `$wp_query = $original_wp_query` should be enough in `tearDown`.

tests/test-misc.php
mjangda: No preference; either is fine :)

shantanu2704:
I renamed the file, but somehow a new file got created and the old file `test-misc.php` was left unchanged. I don't understand what went wrong :(

shantanu2704:
I think this happened because I renamed the file from my IDE. Looking through SO, I figure I should've done `git mv old-file-name.php new-file-name.php`

Since I've already pushed to the remote branch, is there a way to safely undo the last commit/push?
Or should I just remove the old file(`test-misc.php`) and commit/push again?

mjangda:
You can add the file in a new commit and push up. If you're feeling adventurous, you could use git rebase to combine the two commits before you push. If you need help, just ping me on Slack and I can walk you through it.

PR #966, opened 2018-09-05
Bump concurrency of the transition events

mjangda:
With lots of publishing activity on a site, we can end up with a backlog. Enabling some concurrency allows that to be cleared a bit faster.

Main caveat is that the events are no longer atomic and anything relying on post state should check the current value in the event callback.

Related #964

## Checklist

## Steps to Test

n/a


mjangda:
> Main caveat is that the events are no longer atomic and anything relying on post state should check the current value in the event callback.

Docs have been updated to prepare for this: https://vip.wordpress.com/documentation/vip-go/asynchronous-publishing-actions-on-vip-go/#caveats

mjangda:
> Do you think we would ever need this as a dedicated function, rather than an anonymous function?

If we do, we can switch it up pretty easily.

mjangda:
r117733-deploy

PR #968, opened 2018-09-10
Privacy: Use different hooks for overriding filters

mjangda:
`init` is too early to override the core privacy file generation hook, so switch to `admin_init`. Keep the cron handler on `init`.

Previously, we would have both our override and the core ones hooked, which meant we would fall through to the core handle and trigger their `ZipArchive` check and fail the entire request. 

## Checklist

## Steps to Test

- Apply patch to a site without the `zip` extension.
- Generate a privacy report for a user
- Verify that it gets generated successfully.
- Verify that our custom cron job runs and deletes old files.


mjangda:
r119774-deploy

PR #972, opened 2018-09-25
Move files from a8c-files => files

mjangda:
To keep things consistent and in one place.

Continued from #860


mjangda:
r118915-deploy

PR #974, opened 2018-09-27
Add lost password limits

brettshumaker:
This adds rate limits to lost password requests using similar logic to our login limiter. It generalizes a few of the existing functions used in limiting login attempts so the logic could be applied to multiple contexts. We're limiting lost password requests to 3 attempts. I've also added two tests related to the new functionality.

## Steps to Test

1. Check out PR.
1. Go to `/wp-login.php?action=lostpassword`
1. Fill out and submit the form.
1. Verify that an error message is returned after the third attempt.
1. Verify that no password reset email is received after the third attempt.


mjangda:
r120355-deploy

PR #975, opened 2018-09-27
Updated PHPUnit setup instructions

brettshumaker:
I've updated the instructions for running tests with PHPUnit to include specific instruction for using VVV. These are the steps @mjangda provided me to get my local setup working correctly.

### DIFF
--- a/README-PUBLIC.md
+++ b/README-PUBLIC.md
@@ -1,48 +1,13 @@
-# VIP MU Plugins
+# VIP Go mu-plugins
 
-This is the complete MU plugins folder for the WordPress.com VIP next generation hosting platform, assembled without submodules for convenience. 
+This is the mu-plugins folder for VIP Go, assembled without submodules for convenience.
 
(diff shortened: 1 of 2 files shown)


README-PUBLIC.md
```diff
@@ -19,15 +19,28 @@ make lint
 
 **To set up PHPUnit locally** (requires a working WordPress development environment, specifically PHP and MySQL):
 
-Notes:
```
mjangda: Hm, I'm not sure why these were originally added to README-PUBLIC; I think they should be in the main README instead.

brettshumaker: Should README-PUBLIC.md just be fully combined with the main README? If we remove the testing instructions the only unique content in that file will be a sentence about what this repo is and a link to the repo itself.

mjangda: Yeah, I moved most of the content to README. The PUBLIC one is used here: https://github.com/Automattic/vip-go-mu-plugins-built

Issue #977, opened 2018-10-08
php warning - hash_equals() expects strings


davidsword:
I came across

```
Warning: hash_equals(): Expected user_string to be a string, integer given 
in lib/proxy/ip-forward.php on line 175
```

https://github.com/Automattic/vip-go-mu-plugins/blob/master/lib/proxy/ip-forward.php#L175

The integer looks like it can come from `get_proxy_verification_key()` which returns `rand()` if `WPCOM_VIP_PROXY_VERIFICATION` is undefined.

https://github.com/Automattic/vip-go-mu-plugins/blob/master/lib/proxy/ip-forward.php#L157

I am wondering if there was a deliberate reason for returning an integer, or if `rand()` should be casted as a string.

mjangda:
> I am wondering if there was a deliberate reason for returning an integer, or if rand() should be casted as a string.

Not intentional; casting as a string seems reasonable.

mjangda:
This was fixed.

PR #980, opened 2018-10-11
Fix hash_equals() php warning

davidsword:
This solves #977

`hash_equals()` expects both values passed into it to be strings. However `get_proxy_verification_key()` may return an integer when `WPCOM_VIP_PROXY_VERIFICATION` is undefined, triggering a PHP warning. Casting `rand()` to a  string prevents this.

phpunit tests updated to reflect string instead of integer. Added `assertNotEmpty()` to verify that the string of numbers are properly returned when `WPCOM_VIP_PROXY_VERIFICATION` undefined or blank.

### DIFF
--- a/lib/proxy/ip-forward.php
+++ b/lib/proxy/ip-forward.php
@@ -159,8 +159,8 @@ function get_proxy_verification_key() {
 		return (string) WPCOM_VIP_PROXY_VERIFICATION;
 	}
 
-	// If not properly defined for some reason, return a string of random numbers to avoid guessing the key.
-	return (string) rand();
+	// If not properly defined for some reason, return a string of random chars to avoid guessing the key.
+	return bin2hex( random_bytes( 32 ) );
 }
 
 /**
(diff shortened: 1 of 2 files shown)


lib/proxy/ip-forward.php
```diff
@@ -152,15 +152,15 @@ function set_remote_address( $ip ) {
 /**
  * Return the defined verification key for a site
  *
- * @return string|int The verification key if available, or a random integer if no key is configured.
+ * @return string The verification key if available, or a string of random numbers if no key is configured.
  */
 function get_proxy_verification_key() {
 	if ( defined( 'WPCOM_VIP_PROXY_VERIFICATION' ) && ! empty( WPCOM_VIP_PROXY_VERIFICATION ) ) {
 		return WPCOM_VIP_PROXY_VERIFICATION;
```
mjangda: Just to be safe, we should probably cast this as a string as well and add a test for it.

tests/test-lib-proxy-ip-forward.php
```diff
@@ -226,7 +226,8 @@ public function test__not_defined() {
 
 		$actual_key = get_proxy_verification_key();
 
-		$this->assertTrue( is_numeric( $actual_key ) );
+		$this->assertNotEmpty( $actual_key );
+		$this->assertTrue( is_string( $actual_key ) );
```
mjangda: Any time you have multiple assertions in a test, it's helpful to add the message parameter with details about the failure.

davidsword:
Thanks! Very good points. I've added a commit https://github.com/Automattic/vip-go-mu-plugins/pull/980/commits/7c9a0e0eb36096c90670fedfa5f3619ed2d1179e with these changes.

mjangda:
r119769-deploy

PR #983, opened 2018-10-15
Fix spelling error

pkevan:
Makes the comment sound less Scottish 😉


chrean:
Although I like scottish accent.

pkevan:
Me too, 50% Scottish over here 👋

mjangda:
r120355-deploy

PR #984, opened 2018-10-16
Liveblog: Purge Varnish for each new liveblog entry.

philipjohn:
Given that Varnish caches posts for 30 minutes, a liveblog post can contain important meta data that is as much as 30 minutes out of date. This is because the post includes data about the last updated timestamp and this is included in the cached output.

This data can therefore differ from the actual entries, which are loaded asynchronously.

## Checklist

## Steps to Test

1. Sandbox a test site
1. Apply the patch
1. Start a new Liveblog, and load it in the browser, in incognito
1. Reload until you see the `X-Cache: hit` header
1. Note the associated meta data in the source code, it should include `latest_entry_timestamp`
1. In a logged-in browser, add a new entry
1. The new entry should appear in the incognito window
1. Reload the page
1. The liveblog entries should remain the same
1. Check the meta data again - the timestamp should have changed, and the `X-Cache: miss` header should be present


vip-helpers/vip-liveblog.php
```diff
@@ -0,0 +1,29 @@
...
+ */
+function wpcom_vip_liveblog_purge_on_new_entries( $comment_id, $post_id ){
+
+	// Get the URL for this Liveblog post.
+	$permalink = get_permalink( absint( $post_id ) );
+	if ( ! $permalink ) {
+		return;
+	}
+
+	if ( ! function_exists( 'wpcom_vip_purge_edge_cache_for_url' ) ) {
```
GaryJones: I would presume that this `function_exists()` check is less expensive than trying to do a `get_permalink()`, so I'd swap over these guard clauses, and avoid a potentially un-necessary database query.

vip-helpers/vip-liveblog.php
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Purge the Varnish cache for a liveblog on each new entry.
+ *
+ * Ensures that a Liveblog page isn't cached with stale mete data during an
+ * active liveblog.
+ *
+ * @param  int $comment_id ID of the comment for this new entry.
+ * @param  int $post_id    ID for this liveblog post.
+ */
+function wpcom_vip_liveblog_purge_on_new_entries( $comment_id, $post_id ){
```
GaryJones: Any reason not to use scalar type declarations on these arguments?

vip-helpers/vip-liveblog.php
```diff
@@ -0,0 +1,29 @@
...
+	if ( ! function_exists( 'wpcom_vip_purge_edge_cache_for_url' ) ) {
+		return;
+	}
+
+	// Purge the Varnish cache for the liveblog post so that new loads of the
+	// post include the newest entries.
+	wpcom_vip_purge_edge_cache_for_url( $permalink );
+
+}
+add_action( 'liveblog_insert_entry', 'wpcom_vip_liveblog_purge_on_new_entries' );
```
GaryJones: The callback is expecting 2 parameters, but the default args here would suggest only 1 will be passed through. Does it need to be:

```
add_action( 'liveblog_insert_entry', 'wpcom_vip_liveblog_purge_on_new_entries', 10, 2 );
```
?

vip-helpers/vip-liveblog.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+/**
+ * Purge the Varnish cache for a liveblog on each new entry.
+ *
+ * Ensures that a Liveblog page isn't cached with stale mete data during an
```
GaryJones: Typo: mete data -> metadata

vip-helpers/vip-liveblog.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+/**
+ * Purge the Varnish cache for a liveblog on each new entry.
+ *
+ * Ensures that a Liveblog page isn't cached with stale meta data during an
+ * active liveblog.
+ *
+ * @param  int $comment_id ID of the comment for this new entry.
```
GaryJones: Only one space needed between `@param` and arg type. This line and the next.

vip-helpers/vip-liveblog.php
```diff
@@ -0,0 +1,29 @@
...
+
+	// Get the URL for this Liveblog post.
+	$permalink = get_permalink( absint( $post_id ) );
+	if ( ! $permalink ) {
+		return;
+	}
+
+	// Purge the Varnish cache for the liveblog post so that new loads of the
+	// post include the newest entries.
+	wpcom_vip_purge_edge_cache_for_url( $permalink );
```
dhanendrannewscorp: @philipjohn don't we need to clear the AMP cache separately?

WPprodigy:
I think this change may make more sense in the liveblog plugin itself now?

WPprodigy:
Closing this as it is quite stale now. Opened up an issue in the liveblog repo for tracking: https://github.com/Automattic/liveblog/issues/677

PR #985, opened 2018-10-17
Impliment concatention whitelist for Ramp

dchymko:
## Description
When certain js files are run through the existing concatenation on VIP, some are not properly rendered which breaks things in the Gutenberg editor. This fix whitelists a subset of files so that the functionality works correctly.

## Pros
Gutenberg works

## Cons
Performance can be slightly affected if not all files aren't concatenated

## Steps to Test

1. Enable Gutenberg for the Go site if not enabled already
2. Set `vip-go-mu-plugins` on the site to one that does not include any concat fixes (https://github.com/Automattic/vip-go-mu-plugins/commit/8335bb772bf5784b453df2634e5aa187489054ae or older)
3. Verify that the Gutenberg editor does not load for a Gutenberg-enabled post in wp-admin (the post are will be white space and you will see JavaScript errors in the console like: `TypeError: f is undefined`)
4. Update the `vip-go-mu-plugins` on the site to this version.
5. Verify that the Gutenberg-editor does load for a Gutenberg-enabled post in wp-admin
6. To ensure concatenation is working but excluding the files in the whitelist, use the Network tab in the browser console. Set the filter to show just the JS files, and you should see `lodash.js`, `editor.js` and about 20 or so `/_static/??`... files. (With the original concat fix (#957) the Gutenberg editor works but this list is 100+ files.)

### DIFF
--- a/gutenberg-ramp.php
+++ b/gutenberg-ramp.php
@@ -41,7 +41,7 @@ function wpcom_vip_load_gutenberg( $criteria = true ) {
 
 	gutenberg_ramp_load_gutenberg( $criteria );
 
-	add_action( 'admin_init', 'wpcom_vip_disable_gutenberg_concat' );	
+	add_action( 'admin_init', 'wpcom_vip_disable_gutenberg_concat' );
 }
 
 /**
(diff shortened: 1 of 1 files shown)


dchymko:
I still need to test this on a sandbox.

mjangda:
r120263-deploy

mjangda:
Reverted in r120264-deploy; needs testing.

PR #987, opened 2018-10-18
WPCOM Compat: 'plugins' means the default plugin location

joshbetz:
In order to allow themes to be more easily copied as-is from
WordPress.com, this makes wpcom_vip_load_plugin interpret 'plugins'
as the default plugin location. This is because it really is the default
location, but WordPress.com themes commonly set it anyway because they
need to set a release candidate version. When we copy the theme to VIP
Go, it tries to load the theme from wp-content/plugins/plugins.


mjangda:
r120385-deploy

PR #988, opened 2018-10-18
Connect akismet during migration cleanup

joshbetz:
### Test:

First, in a sandbox:

```
wp --allow-root cron event schedule vip_after_data_migration
```

After the command has finished, verify Akismet is connected:

```
wp --allow-root shell
wp> Akismet_Admin::display_status();
<div class="akismet-alert akismet-active">
	<h3 class="akismet-key-status">Akismet is now protecting your site from spam. Happy blogging!</h3>
	</div>
```

### DIFF
--- a/vip-helpers/vip-migrations.php
+++ b/vip-helpers/vip-migrations.php
@@ -41,6 +41,8 @@ function run_after_data_migration_cleanup() {
 	
 	if ( class_exists( 'Akismet_Admin' ) && method_exists( 'Akismet_Admin', 'connect_jetpack_user' ) ) {
 		Akismet_Admin::connect_jetpack_user();
+	} else {
+		trigger_error( sprintf( '%s: Failed to call `Akismet_Admin::connect_jetpack_user` as it does not exist; Akismet will need to be connected manually', __FUNCTION__ ), E_USER_WARNING );
 	}
 }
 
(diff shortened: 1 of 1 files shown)


vip-helpers/vip-migrations.php
```diff
@@ -38,6 +38,7 @@ function run_after_data_migration_cleanup() {
 
 	connect_jetpack();
 	connect_vaultpress();
+	Akismet_Admin::connect_jetpack_user();
```
mjangda: Should probably have a `method_exists` check here?

vip-helpers/vip-migrations.php
```diff
@@ -38,6 +38,10 @@ function run_after_data_migration_cleanup() {
 
 	connect_jetpack();
 	connect_vaultpress();
+	
+	if ( class_exists( 'Akismet_Admin' ) && method_exists( 'Akismet_Admin', 'connect_jetpack_user' ) ) {
+		Akismet_Admin::connect_jetpack_user();
+	}
```
mjangda:
```suggestion
	} else {
		trigger_error( sprintf( '%s: Failed to call `Akismet_Admin::connect_jetpack_user` as it does not exist; Akismet will need to be connected manually', __FUNCTION__ ), E_USER_WARNING );
	}
```

joshbetz:
r121479-deploy

PR #989, opened 2018-10-18
Run dbDelta during migration cleanup

joshbetz:
If it's a multisite, ensure we update the global tables. For each blog, update the blog tables.

### Test:

First, in a sandbox:

```
wp --allow-root cron event schedule vip_after_data_migration
```

After the command has finished, verify the schema is correct:

```
$ vipgo db <site>
mysql> show indexes from wp_postmeta;
```

Ensure the `vip_meta_key_value` index exists.

PR #993, opened 2018-10-23
vip-migration: Fix dbDelta fatals

mjangda:
It's defined in an wp-admin includes file which isn't loaded everywhere. To be safe, we should require that file before calling dbDelta.

Introduced in #989


mjangda:
r120516-deploy

PR #997, opened 2018-10-29
Add concatenation efficiency range statistic

dchymko:
## Description
Calculates an efficiency ratio of how well the js_concat functionality is combining files. This will primarily be used for before and after comparisons of the js_concat changes made for the Gutenberg editor.

See https://github.com/Automattic/nginx-http-concat/pull/43/


mjangda:
r120961-deploy

PR #998, opened 2018-10-30
Changed group metric calculation 

dchymko:
## Description
Changed our concatenation stats metric to calculate the  efficiency ratio based on the average group size relative to the number of scripts in all groups:

 `average_groupsize`/`total_number_scripts_in_groups`:
1) 1 group of 10 scripts:   10/10 = 100%
2) 2 groups, one of 5, one of 15 = 10/20 = 50%
2) 3 groups of 3 scripts each : 3/9 = 33%
3)  10 groups of 1 script each = 1/10 = 10%

Also see: https://github.com/Automattic/vip-go-mu-plugins/pull/997

### DIFF
--- a/stats.php
+++ b/stats.php
@@ -61,19 +61,18 @@ class Concat_Metrics {
 	}
 
 	private function calculate_efficiency_ratio( $scripts ) {
-		$total_scripts = count( $scripts );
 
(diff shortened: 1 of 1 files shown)


mjangda:
r121054-deploy

PR #1010, opened 2018-10-30
Use named function to allow re-enabling filename media searches

mdbitz:
## Description

Currently Filename searches are disabled for performance reasons. However clients have requested this functionality be enabled. (see Ticket 81547 in Zendesk). This update names the anonymous function so that clients can remove it if needed.

I'm not sure if we want to have documentation on this feature as it is a performance concern.

## Checklist

## Steps to Test

1. Check out PR.
1. Go to `wp-admin` > `Media`
1. Perform a search for a filename
1. Verify expected media is *not* returned.
1. Add `remove_action( 'pre_get_posts', 'vip_filter_query_attachment_filenames');` to a themes functions.php file.
1. Go to `wp-admin` > `Media`
1. Perform a search for a filename
1. Verify expected media *is* returned.

### DIFF
--- a/performance/media.php
+++ b/performance/media.php
@@ -2,8 +2,8 @@
 
 // Prevent core from doing filename lookups for media search.
 // https://core.trac.wordpress.org/ticket/39358
-if( ! ( defined( 'VIP_GO_ENABLE_FILENAMES_SEARCH' ) && VIP_GO_ENABLE_FILENAMES_SEARCH ) ) {
-	add_action( 'pre_get_posts', function() {
-		remove_filter( 'posts_clauses', '_filter_query_attachment_filenames' );
-	} );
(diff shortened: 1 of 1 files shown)


mjangda:
As discussed, I think a better approach to this would be for site to just restore the filter:

```
add_action( 'pre_get_posts', function() {
	add_filter( 'posts_clauses', '_filter_query_attachment_filenames' );
}, 9999 ); // run later after VIP
```

mdbitz:
Hi @mjangda as discussed we can't simply re-add the filter as it is only added in specific queries. Instead how are you with us giving a name to the anonymous function? This way clients can remove the filter using the named function to enable the filename search?

`remove_action( 'pre_get_posts', 'vip_go_filter_query_attachment_filenanes');`

mjangda:
r123698-deploy

PR #1011, opened 2018-10-30
Remove the concatentation stats as they are no longer needed. 

dchymko:
## Description
We have verified the concat changes are functioning correctly and not affecting performance.


mjangda:
r121092-deploy

PR #1012, opened 2018-10-31
Files: Globally enable Photon's Local Filters

rinatkhaziev:
### What

 Remove all the hacks meant to work around the issues with Photon initialization in is_admin context: `A8C_Files::init_vip_photon_filters`, and corresponding `A8C_Files::image_resize`, which meant to act as a replacement/shim for Photon's `image_downsize` filter, because there was no way to enable Photon downsize in `is_admin()` 2 years ago.

Remove init_vip_photon_filter, because it's has a comment 'This will be phased out and become the default eventually.' dating back 2 years, so it's safe to assume that 'eventually' is now.

### Why 

Having to set `WPCOM_VIP_USE_JETPACK_PHOTON` in vip-config.php or elsewhere seems like an extra step that was meant to be a stop-gap measure until the Files service is stable, it's confusing, and adds extra friction when moving to Go. 

## Checklist

### DIFF
--- a/a8c-files.php
+++ b/a8c-files.php
@@ -49,14 +49,7 @@ class A8C_Files {
 		add_filter( 'wp_save_image_file',        array( &$this, 'save_image_file' ), 10, 5 );
 		add_filter( 'wp_save_image_editor_file', array( &$this, 'save_image_file' ), 10, 5 );
 
-		// Limit to certain contexts for the initial testing and roll-out.
-		// This will be phased out and become the default eventually.
-		$use_jetpack_photon = $this->use_jetpack_photon();
-		if ( $use_jetpack_photon ) {
-			$this->init_jetpack_photon_filters();
-		} else {
-			$this->init_vip_photon_filters();
-		}
+		$this->init_jetpack_photon_filters();
 
 		// Automatic creation of intermediate image sizes is disabled via `wpcom_intermediate_sizes()`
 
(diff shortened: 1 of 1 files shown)


joshbetz:
I don't think we want Photon enabled everywhere since the VIP Go Files Service supports all the same filters and this will lead to double caching everything. Ideally we want the Photon magic to apply query parameters and devicepx, but not to send files through `i[0-2].wp.com` (Photon).

mjangda:
> I don't think we want Photon enabled everywhere

This change loads the local Photon module within Jetpack but images continue to be served from the VIP Go Files Service (not the remote Photon service).

Naming is hard :)

joshbetz:
Aha, thanks for the clarification. I was confused because I thought we still needed the `image_resize()`, but I guess Photon will fully handle image resizing now 🤘.

rinatkhaziev:
@joshbetz you're not the only one confused, that's why I made this PR specifically in this fashion - not keeping the image_resize and the constant. 

@mjangda pointed out last night that we have issues with devicepx, I have some availability this week and can take a stab at it, in order to get this PR merged.

joshbetz:
@rinatkhaziev what are the next steps on this?

rinatkhaziev:
@joshbetz  Ideally, we'd want to test this and make sure we don't break sites. As for the correct testing strategy, I'm at a complete loss. 

My basic unicorns-and-rainbows vision is grabbing a fully uncached version of output before the change, and then compare that to a fully uncached version after the change, the bells-and-whistles version would also include adding/editing content and comparing that as well.

joshbetz:
> As for the correct testing strategy, I'm at a complete loss.

I would start by sandboxing some sites. If you're logged in, you'll bypass the page cache, so that won't be an issue.

rinatkhaziev:
Superseded by #2185

PR #1014, opened 2018-10-31
Jetpack: Make Protect a mandatory module

mjangda:
This was disabled earlier due to login issues with reverse proxies. Those have since been fixed and we can restore this.

## Checklist

## Steps to Test

1. Check out PR.
1. Go to `wp-admin` > `Jetpack`
1. Verify that Protect is turned on.


mjangda:
We have other protections in place now, so this is not necessary.

PR #1018, opened 2018-11-05
Add PHPUnit test script

### LINKED ISSUE #1017: PHP Unit Test Script
The current PHPUnit scripts are built for Travis CI and don't work out of the box locally. We should include a script to setup PHPUnit locally with Docker.

I'm thinking we can use `docker run mariadb` and `docker run phpunit/phpunit` with an adapted version of the install-wp-tests script.

joshbetz:
```
usage: ./bin/phpunit-docker.sh [wp-version]
```

Adds a new script that uses docker to create a temporary database
container, installs the WP Unit Test dependencies, and runs the tests
in a phpunit container.

We're also splitting the existing install-wp-tests.sh script into two parts
so the applicable parts are reusable for this. The part that sets up a
local database is unnecessary here since we're using Docker for that.

The existing script still references the new download-wp-tests.sh script
internally, so that should continue to function the same as before.

Fixes #1017

### DIFF
--- a/bin/phpunit-docker.sh
+++ b/bin/phpunit-docker.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+WP_VERSION=${1-latest}
+
+MYSQL_ROOT_PASSWORD='wordpress'
+db=$(docker run -p 3306:3306 -e MYSQL_ROOT_PASSWORD=$MYSQL_ROOT_PASSWORD -d mariadb)
+function cleanup() {
+	docker rm -f $db
+}
+trap cleanup EXIT
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+WP_VERSION=$($DIR/download-wp-tests.sh wordpress_test root $MYSQL_ROOT_PASSWORD 127.0.0.1 $WP_VERSION)
+
+## Ensure there's a database connection for the rest of the steps
+until docker exec -it $db mysql -u root --password="wordpress" -e 'CREATE DATABASE wordpress_test' > /dev/null; do
+	echo "Waiting for database connection..."
+	sleep 5
+done
+
+docker run \
+	-v $(pwd):/app \
+	-v /tmp/wordpress-tests-lib-$WP_VERSION:/tmp/wordpress-tests-lib \
+	-v /tmp/wordpress-$WP_VERSION:/tmp/wordpress \
+	--network host \
+	--rm phpunit/phpunit \
+	--bootstrap /app/tests/bootstrap.php
--- a/bin/phpunit-docker.sh
+++ b/bin/phpunit-docker.sh
@@ -10,7 +10,7 @@ function cleanup() {
 trap cleanup EXIT
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
-WP_VERSION=$($DIR/download-wp-tests.sh wordpress_test root $MYSQL_ROOT_PASSWORD 127.0.0.1 $WP_VERSION)
+WP_VERSION=$($DIR/download-wp-tests.sh wordpress_test root "$MYSQL_ROOT_PASSWORD" "127.0.0.1" "$WP_VERSION")
 
 ## Ensure there's a database connection for the rest of the steps
 until docker exec -it $db mysql -u root --password="wordpress" -e 'CREATE DATABASE wordpress_test' > /dev/null; do
--- a/bin/phpunit-docker.sh
+++ b/bin/phpunit-docker.sh
@@ -3,22 +3,25 @@
 WP_VERSION=${1-latest}
 
 MYSQL_ROOT_PASSWORD='wordpress'
-db=$(docker run -p 3306:3306 -e MYSQL_ROOT_PASSWORD=$MYSQL_ROOT_PASSWORD -d mariadb)
+docker network create tests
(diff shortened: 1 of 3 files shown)


bin/phpunit-docker.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+WP_VERSION=${1-latest}
+
+MYSQL_ROOT_PASSWORD='wordpress'
+db=$(docker run -p 3306:3306 -e MYSQL_ROOT_PASSWORD=$MYSQL_ROOT_PASSWORD -d mariadb)
```
dchymko: Minor suggestion: If we avoid using the host network, then there won't be a conflict if MySQL is already running there.
```suggestion
docker network create tests
db=$(docker run -p 3306:3306 --network tests --name db -e MYSQL_ROOT_PASSWORD=$MYSQL_ROOT_PASSWORD -d mariadb)
```

bin/phpunit-docker.sh
```diff
@@ -0,0 +1,27 @@
...
+function cleanup() {
+	docker rm -f $db
+}
```
dchymko:
```suggestion
	docker network rm tests
}
```

bin/phpunit-docker.sh
```diff
@@ -0,0 +1,27 @@
...
+trap cleanup EXIT
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+WP_VERSION=$($DIR/download-wp-tests.sh wordpress_test root "$MYSQL_ROOT_PASSWORD" "127.0.0.1" "$WP_VERSION")
```
dchymko:
```suggestion
WP_VERSION=$($DIR/download-wp-tests.sh wordpress_test root "$MYSQL_ROOT_PASSWORD" "db" "$WP_VERSION")
```

bin/phpunit-docker.sh
```diff
@@ -0,0 +1,27 @@
...
+
+## Ensure there's a database connection for the rest of the steps
+until docker exec -it $db mysql -u root --password="wordpress" -e 'CREATE DATABASE wordpress_test' > /dev/null; do
```
dchymko:
```suggestion
until docker exec -it $db mysql -u root -h db --password="wordpress" -e 'CREATE DATABASE wordpress_test' > /dev/null; do
```

bin/phpunit-docker.sh
```diff
@@ -0,0 +1,27 @@
...
+	echo "Waiting for database connection..."
+	sleep 5
+done
+
+docker run \
+	-v $(pwd):/app \
```
dchymko:
```suggestion
 	--network tests \
	-v $(pwd):/app \
```

bin/phpunit-docker.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+WP_VERSION=${1-latest}
+
+MYSQL_ROOT_PASSWORD='wordpress'
+docker network create tests
+db=$(docker run -p 3306:3306 --network tests --name db -e MYSQL_ROOT_PASSWORD=$MYSQL_ROOT_PASSWORD -d mariadb)
```
dchymko:
```suggestion
db=$(docker run --network tests --name db -e MYSQL_ROOT_PASSWORD=$MYSQL_ROOT_PASSWORD -d mariadb)
```
I missed removing the host port mapping.

bin/phpunit-docker.sh
```diff
@@ -0,0 +1,30 @@
...
+	echo "Waiting for database connection..."
+	sleep 5
+done
+
+docker run \
+ 	--network tests \
+	-v $(pwd):/app \
+	-v /tmp/wordpress-tests-lib-$WP_VERSION:/tmp/wordpress-tests-lib \
+	-v /tmp/wordpress-$WP_VERSION:/tmp/wordpress \
+	--network host \
```
dchymko:
```suggestion
	--network host \
```

bin/phpunit-docker.sh
```diff
@@ -0,0 +1,29 @@
...
+	echo "Waiting for database connection..."
+	sleep 5
+done
+
+docker run \
+ 	--network tests \
+	-v $(pwd):/app \
+	-v /tmp/wordpress-tests-lib-$WP_VERSION:/tmp/wordpress-tests-lib \
+	-v /tmp/wordpress-$WP_VERSION:/tmp/wordpress \
+	--rm phpunit/phpunit \
```
mjangda: One minor downside to this image is that it's running PHP 7.0 (https://hub.docker.com/r/phpunit/phpunit/)

mjangda:
This is pretty awesome. Works well locally.

Not critical but would be nice to have a way to keep the `tests` network + `phpunit` container around to avoid the bootup time every time you run tests.

bin/phpunit-docker.sh
joshbetz: > One minor downside to this image is that it's running PHP 7.0

I think it depends what version we use. For example, I see the PHP Unit 7.4.0 container runs Alpine 3.8. If I `docker run -it alpine:3.8` and install the `php7` package, it actually installs `PHP 7.2.10`. It looks like there's no consistency for what version of Alpine they're running though. Some of the Dockerfiles say 3.8, some say 3.7, and some say latest.

https://github.com/JulienBreux/phpunit-docker/blob/master/7.4.0/dockerfile#L2

PR #1020, opened 2018-11-07
Add VIP Filesystem plugin

hanifn:
## Description

Add plugin to provide a custom stream wrapper for the VIP Filesystem Service

## Checklist


vip-filesystem/class.vip-filesystem-loader.php
```diff
@@ -0,0 +1,121 @@
...
+namespace Automattic\VIP\Filesystem;
+
+/**
+ * Register all actions and filters for the plugin.
+ *
+ * Maintain a list of all hooks that are registered throughout
+ * the plugin, and register them with the WordPress API. Call the
+ * run function to execute the list of actions and filters.
+ */
+class Vip_Filesystem_Loader {
```
mjangda: What's the benefit to having this Loader class? Feels a bit like an unnecessary abstraction (and maybe created as part of some boilerplate?) If there are some clear benefits, we can keep it; otherwise, probably just simpler to call `add_action`/`add_filter` directly in the main `Vip_Filesystem` class.

vip-filesystem.php
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Currently plugin version.
+ * Start at version 1.0.0 and use SemVer - https://semver.org
+ */
+define( 'VIP_FILESYSTEM_VERSION', '1.0.0' );
+
+/**
+ * The core plugin class
+ */
+require __DIR__ . '/vip-filesystem/class.vip-filesystem.php';
```
mjangda: Should we put all these files into `/files/` alongside all the other files related to the Files Service? (If so, we'll want to update the Namespace, even though we're not super consistent with naming :))

vip-filesystem/class.vip-filesystem-loader.php
```diff
@@ -0,0 +1,121 @@
+<?php
+/**
+ * Created by PhpStorm.
```
mjangda: We should remove these auto-generated metadata comments.

vip-filesystem/class.vip-filesystem-stream.php
```diff
@@ -0,0 +1,551 @@
...
+	 * @return  bool    True on success or FALSE on failure
+	 */
+	public function stream_open( $path, $mode, $options, &$opened_path ) {
+		$path = $this->trim_path( $path );
+
+		if ( $this->client->is_file( $path ) ) {
+			$result = $this->client->get_file( $path );
+
+			if ( is_wp_error( $result ) || $result instanceof \WP_Error ) {
+				trigger_error( $result->get_error_message(), E_USER_WARNING );
```
mjangda: For this and all other `trigger_error` calls, thoughts on prefixing them with `__METHOD__`? Would make debugging/tracing a lot easier.

vip-filesystem/class.vip-filesystem-stream.php
```diff
@@ -0,0 +1,551 @@
...
+	 *
+	 * @return  bool    True on success or FALSE on failure
+	 */
+	public function stream_open( $path, $mode, $options, &$opened_path ) {
+		$path = $this->trim_path( $path );
+
+		if ( $this->client->is_file( $path ) ) {
+			$result = $this->client->get_file( $path );
+
+			if ( is_wp_error( $result ) || $result instanceof \WP_Error ) {
```
mjangda: `|| $result instanceof \WP_Error` is probably redundant?

```suggestion
			if ( is_wp_error( $result ) ) {
```

vip-filesystem/class.vip-filesystem-stream.php
```diff
@@ -0,0 +1,551 @@
...
+			}
+
+			// Converts file contents into stream resource
+			$file = $this->string_to_resource( $result );
+		} else {
+			// File doesn't exist on File service so create new file
+			$file = $this->string_to_resource( '' );
+		}
+
+
```
mjangda: Minor: extra space.

```suggestion
```

vip-filesystem/class.vip-filesystem-stream.php
```diff
@@ -0,0 +1,551 @@
...
+		if ( FALSE === $length ) {
+			trigger_error( 'Error writing to file: ' . $this->path,
+				E_USER_WARNING );
+			return FALSE;
+		}
+
+		// Upload to file service
+		$result = $this->client
+			->upload_file( $this->uri, $this->path );
+		if ( is_wp_error( $result ) || $result instanceof \WP_Error ) {
```
mjangda: Redundant WP_Error check?

```suggestion
		if ( is_wp_error( $result ) ) {
```

vip-filesystem/class.vip-filesystem-stream.php
```diff
@@ -0,0 +1,551 @@
...
+	 * @access  public
+	 * @param   string  $path
+	 *
+	 * @return  bool    True if success. False on failure
+	 */
+	public function unlink( $path ) {
+		$path = $this->trim_path( $path );
+		$result = $this->client->delete_file( $path );
+
+		if ( is_wp_error( $result ) || $result instanceof \WP_Error ) {
```
mjangda: Redundant WP_Error check?

```suggestion
		if ( is_wp_error( $result ) ) {
```

vip-filesystem/class.vip-filesystem-stream.php
```diff
@@ -0,0 +1,551 @@
...
+			trigger_error( $result->get_error_message(), E_USER_WARNING );
+			return FALSE;
+		}
+
+		$this->close_handler();
+
+		return TRUE;
```
mjangda: Minor: we're using mixed cases for true/false (sometimes `true`, sometimes `TRUE`). We should be consistent (lowercase is preferred).

vip-filesystem.php
```diff
@@ -0,0 +1,43 @@
+<?php
+/**
+ * Plugin Name:       VIP Filesystem Plugin
+ * Description:       Provides a custom stream wrapper for handling file uploads to VIP/A8C Files API.
+ * Version:           1.0.0
+ * Author:            Automattic
+ * Author URI:        http://automattic.com/
+ * License:           GPL-2.0+
+ * License URI:       http://www.gnu.org/licenses/gpl-2.0.txt
```
mjangda: Let's add some credits / thank yous / links to the various plugins and code we've referenced, used, or drawn inspiration from.

vip-filesystem/class.vip-filesystem-stream.php
```diff
@@ -0,0 +1,551 @@
...
+		 *
+		 * Hanif's note: Copied from humanmade's S3 plugin
+		 */
+		if ( ! $extension ) {
+			return $stats;
+		}
+
+		$info = array();
+		$result = $this->client->is_file( $path, $info );
+		if ( is_wp_error( $result ) || $result instanceof \WP_Error ) {
```
mjangda: Redundant check?

```suggestion
		if ( is_wp_error( $result ) ) {
```

vip-filesystem/class.vip-filesystem-stream.php
```diff
@@ -0,0 +1,551 @@
...
+		$extension = pathinfo( $path, PATHINFO_EXTENSION );
+		/**
+		 * If the file is actually just a path to a directory
+		 * then return it as always existing. This is to work
+		 * around wp_upload_dir doing file_exists checks on
+		 * the uploads directory on every page load.
+		 *
+		 * Added by Joe Hoyle
+		 *
+		 * Hanif's note: Copied from humanmade's S3 plugin
```
mjangda: Let's add a link to the code on their GitHub repo.

vip-filesystem/class.vip-filesystem-stream.php
```diff
@@ -0,0 +1,551 @@
...
+	 * @param   string  $path_to    New path to the file
+	 *
+	 * @return  bool    True on successful rename
+	 */
+	public function rename( $path_from, $path_to ) {
+		$path_from = $this->trim_path( $path_from );
+		$path_to = $this->trim_path( $path_to );
+
+		// Get original file first
+		// Note: Subooptimal. Should figure out a way to do this without downloading the file as this could
```
mjangda: I guess probably best to add a new endpoint for this? Although, we may want to see how frequently this is used before investing too much time in that.

vip-filesystem/class.vip-filesystem-stream.php
```diff
@@ -0,0 +1,551 @@
...
+			trigger_error( $result->get_error_message(), E_USER_WARNING );
+			return FALSE;
+		}
+
+		return TRUE;
+	}
+
+	/**
+	 * Called in response to mkdir()
+	 * Currently, it will always return true as directories are automatically created on the Filesystem API
```
mjangda: Let's move this inline as it's easy to miss in the docblock.

```suggestion
```

vip-filesystem/class.vip-filesystem-stream.php
```diff
@@ -0,0 +1,551 @@
...
+	 * @access  public
+	 *
+	 * @param   string  $path
+	 * @param   int     $mode
+	 * @param   in      $options
+	 *
+	 * @return  bool
+	 */
+	public function mkdir( $path, $mode, $options ) {
+		return TRUE;
```
mjangda:
```suggestion
		// Currently, it will always return true as directories are automatically created on the Filesystem API
		return TRUE;
```

vip-filesystem/class.vip-filesystem-stream.php
```diff
@@ -0,0 +1,551 @@
...
+	 * @since   1.0.0
+	 * @access  public
+	 *
+	 * @param   string  $path
+	 * @param   int     $option
+	 * @param   mixed   $value
+	 *
+	 * @return  bool
+	 */
+	public function stream_metadata($path, $option, $value) {
```
mjangda:
```suggestion
	public function stream_metadata( $path, $option, $value ) {
```

vip-filesystem/class.vip-filesystem-stream.php
```diff
@@ -0,0 +1,551 @@
...
+	 */
+	public function rename( $path_from, $path_to ) {
+		$path_from = $this->trim_path( $path_from );
+		$path_to = $this->trim_path( $path_to );
+
+		// Get original file first
+		// Note: Subooptimal. Should figure out a way to do this without downloading the file as this could
+		//       get really inefficient with large files
+		$result = $this->client->get_file( $path_from );
+		if ( is_wp_error( $result ) || $result instanceof \WP_Error ) {
```
mjangda:
```suggestion
		if ( is_wp_error( $result ) ) {
```

vip-filesystem/class.vip-filesystem-stream.php
```diff
@@ -0,0 +1,551 @@
...
+		}
+
+		// Convert to actual file to upload to new path
+		$file = $this->string_to_resource( $result );
+		$meta = stream_get_meta_data( $file );
+		$filePath = $meta['uri'];
+
+		// Upload to file service
+		$result = $this->client->upload_file( $filePath, $path_to );
+		if ( is_wp_error( $result ) || $result instanceof \WP_Error ) {
```
mjangda:
```suggestion
		if ( is_wp_error( $result ) ) {
```

vip-filesystem/class.vip-filesystem-stream.php
```diff
@@ -0,0 +1,551 @@
...
+			trigger_error( $result->get_error_message(), E_USER_WARNING );
+			return FALSE;
+		}
+
+		// Delete old file
+		$result = $this->client->delete_file( $path_from );
+		if ( is_wp_error( $result ) || $result instanceof \WP_Error ) {
```
mjangda:
```suggestion
		if ( is_wp_error( $result ) ) {
```

vip-filesystem/class.vip-filesystem-stream.php
```diff
@@ -0,0 +1,551 @@
...
+	 * @access  public
+	 *
+	 * @param   string  $path_from  Path to file to rename
+	 * @param   string  $path_to    New path to the file
+	 *
+	 * @return  bool    True on successful rename
+	 */
+	public function rename( $path_from, $path_to ) {
+		$path_from = $this->trim_path( $path_from );
+		$path_to = $this->trim_path( $path_to );
```
mjangda: Minor: Do we need a check to make sure that the to/from values are different?

vip-filesystem/class.vip-filesystem-stream.php
```diff
@@ -0,0 +1,551 @@
...
+	 * @access  protected
+	 *
+	 * @param   string     $data   The file content to be written
+	 *
+	 * @return  resource   Returns resource or FALSE on write error
+	 */
+	protected function string_to_resource( $data ) {
+		// Create a temporary file
+		$tmp_handler = tmpfile();
+		if (false === fwrite( $tmp_handler, $data ) ) {
```
mjangda:
```suggestion
		if ( false === fwrite( $tmp_handler, $data ) ) {
```

vip-filesystem/class.vip-filesystem-stream.php
```diff
@@ -0,0 +1,551 @@
...
+	/**
+	 * Closes the open file handler
+	 *
+	 * @since   1.0.0
+	 * @access  protected
+	 *
+	 * @return  bool        True on success. False on failure.
+	 */
+	protected function close_handler() {
+		if (! $this->file ) {
```
mjangda:
```suggestion
		if ( ! $this->file ) {
```

vip-filesystem/class.vip-filesystem-stream.php
```diff
@@ -0,0 +1,551 @@
...
+			$this->path = null;
+			$this->uri  = null;
+		}
+
+		return $result;
+	}
+
+	/**
+	 * Converted the protocol file path into something the File Service
+	 * API can use
```
mjangda:
```suggestion
	 * API client can use
```

vip-filesystem/class.vip-filesystem.php
```diff
@@ -0,0 +1,150 @@
...
+	 * @return array Modified output of `wp_upload_dir`
+	 */
+	public function filter_upload_dir( $params ) {
+		/**
+		 * This is to account for the a8c-files plugin and should be temporary.
+		 * Eventually, this plugin will replace a8c-files so this check can
+		 * then be removed.
+		 * - Hanif
+		 */
+		if ( 0 === stripos( $params['basedir'], '/tmp/uploads' ) ) {
```
mjangda:
```suggestion
		if ( 0 === stripos( $params['basedir'], LOCAL_UPLOADS ) ) {
```

vip-filesystem/class.vip-filesystem.php
```diff
@@ -0,0 +1,150 @@
...
+			$params['path']    = str_replace(
```
mjangda: Just to be on the safe side, should we limit the replaces (this one and the ones below) to only act on the beginning of the string? Something like https://stackoverflow.com/a/1252710

Would avoid issues with situations where someone has an unfortunate path like `/tmp/uploads/files/tmp/uploads`. Very edge case, but since it's a simple fix, worth considering.

vip-filesystem/class.vip-filesystem-stream.php
hanifn: You're right. It probably is.

vip-filesystem/class.vip-filesystem-stream.php
hanifn: But the error messages include the stack trace so we can easily find that sort of information.

vip-filesystem/class.vip-filesystem-stream.php
hanifn: Thats probably a good idea

vip-filesystem/class.vip-filesystem-stream.php
hanifn: Done!

vip-filesystem/class.vip-filesystem-stream.php
hanifn: Done

vip-filesystem.php
mjangda:
```suggestion
 * License URI:       http://www.gnu.org/licenses/gpl-2.0.txt
 *
 * Inspired by the S3 Uploads plugin from Human Made: https://github.com/humanmade/S3-Uploads
```

files/class-vip-filesystem.php
```diff
@@ -0,0 +1,143 @@
...
+			$params['basedir']    = substr_replace( $params['basedir'],
+				self::PROTOCOL . '://wp-content',
+				$pos,
+				strlen( WP_CONTENT_DIR ) );
+		}
+
+		return $params;
+	}
+
+	public function prefilter_move_tmp_file( $file ) {
```
mjangda: I think we can remove this and the `wp_handle_upload_prefilter` filter? Don't think they'll be needed.

hanifn: That is true. Removing

hanifn: Removed

mjangda:
Looking good 👍 Thanks for the work on this @hanifn!

As discussed, we're going to ship the stream wrapper without hooking into any parts of WordPress right now. In future, PRs we're going to work through adding tests and transitioning over from the current `a8c-files`-based workarounds.

mjangda:
r125073-deploy

PR #1029, opened 2018-11-21
Use mangle cache invalidation for file purges

darthhexx:
## Description

If the PURGE_SERVER_TYPE is configured for mangle, use it for files purges instead of direct calls to Varnish. Also, don't fire off deletes to the files service for every image size as this slows down the request needlessly.

## Checklist

- [x] This change works and has been tested locally.
- [x] This change works and has been tested in production.

## Steps to Test

1. Set the purge server type constant on you sandbox to 'mangle'.
2. Delete a file.
3. Ask systems to check the DC LB host logs to ensure only a single DELETE hit the files service and only a single PURGE took place to the mangle cache invalidation service.

### DIFF
--- a/a8c-files.php
+++ b/a8c-files.php
@@ -454,19 +454,21 @@ class A8C_Files {
 		else
 			$file_uri = '/' . $url_parts['path'];
 
-		if ( in_array( $file_uri, $deleted_uris ) ) {
-			// This file has already been successfully deleted from the file service in this request
-			return;
-		}
-
 		$headers = array(
 					'X-Client-Site-ID: ' . constant( 'FILES_CLIENT_SITE_ID' ),
 					'X-Access-Token: ' . constant( 'FILES_ACCESS_TOKEN' ),
 				);
 
+		$delete_uri = $file_uri;
 		$service_url = $this->get_files_service_hostname() . '/' . $this->get_upload_path();
 		if ( is_multisite() && ! ( is_main_network() && is_main_site() ) ) {
 			$service_url .= '/sites/' . get_current_blog_id();
+			$delete_uri = '/sites/' . get_current_blog_id() . $delete_uri;
+		}
+
+		if ( in_array( $delete_uri, $deleted_uris ) ) {
+			// This file has already been successfully deleted from the file service in this request
+			return;
 		}
 
 		$ch = curl_init( $service_url . $file_uri );
@@ -487,7 +489,7 @@ class A8C_Files {
 		}
 
 		// Set our static so we can later recall that this file has already been deleted and purged
-		$deleted_uris[] = $file_uri;
+		$deleted_uris[] = $delete_uri;
 
 		// We successfully deleted the file, purge the file from the caches
 		$invalidation_url = get_site_url() . '/' . $this->get_upload_path();
--- a/a8c-files.php
+++ b/a8c-files.php
@@ -444,11 +444,20 @@ class A8C_Files {
 	}
 
 	function delete_file( $file_name ) {
+		// To ensure we don't needlessly fire off deletes for all sizes of the same image, of
+		// which all except the first result in 404's, keep accounting of what we've deleted.
+		static $deleted_uri;
+
 		$url_parts = parse_url( $file_name );
 		if ( false !== stripos( $url_parts['path'], constant( 'LOCAL_UPLOADS' ) ) )
 			$file_uri = substr( $url_parts['path'], stripos( $url_parts['path'], constant( 'LOCAL_UPLOADS' ) ) + strlen( constant( 'LOCAL_UPLOADS' ) ) );
 		else
 			$file_uri = '/' . $url_parts['path'];
+
+		if ( isset( $deleted_uri ) && ( $deleted_uri == $file_uri ) ) {
+			// This file has already been successfully deleted from the file service in this request
+			return;
+		}
 
 		$headers = array(
 					'X-Client-Site-ID: ' . constant( 'FILES_CLIENT_SITE_ID' ),
@@ -476,6 +485,9 @@ class A8C_Files {
 			error_log( sprintf( __( 'Error deleting the file %s from the remote servers: Code %d' ), $file_uri, $http_code ) );
 			return;
 		}
+
+		// Set our static so we can later recall that this file has already been deleted and purged
+		$deleted_uri = $file_uri;
 
 		// We successfully deleted the file, purge the file from the caches
 		$invalidation_url = get_site_url() . '/' . $this->get_upload_path();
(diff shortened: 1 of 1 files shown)


a8c-files.php
```diff
@@ -444,12 +444,21 @@ function( $ch, $fd, $length ) use( $stream ) {
...
+		static $deleted_uris = array();
+
 		$url_parts = parse_url( $file_name );
 		if ( false !== stripos( $url_parts['path'], constant( 'LOCAL_UPLOADS' ) ) )
 			$file_uri = substr( $url_parts['path'], stripos( $url_parts['path'], constant( 'LOCAL_UPLOADS' ) ) + strlen( constant( 'LOCAL_UPLOADS' ) ) );
 		else
 			$file_uri = '/' . $url_parts['path'];
 
+		if ( in_array( $file_uri, $deleted_uris ) ) {
+			// This file has already been successfully deleted from the file service in this request
```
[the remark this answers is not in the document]
darthhexx: Good catch. It's not possible to do from the media library, but custom calls could definitely do it. I have amended the code to add the multisite path to the deleted file path, to ensure we would delete the files for all sites that deleted the same file path in the same request.

a8c-files.php
```diff
@@ -473,10 +482,13 @@ function delete_file( $file_name ) {
 		curl_close( $ch );
 
 		if ( 200 != $http_code ) {
-			error_log( sprintf( __( 'Error deleting the file from the remote servers: Code %d' ), $http_code ) );
+			error_log( sprintf( __( 'Error deleting the file %s from the remote servers: Code %d' ), $file_uri, $http_code ) );
```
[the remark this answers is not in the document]
darthhexx: The only signal we get is the error code. Either 200, 404 (file is already deleted), or 500s for service related issues. At least now we log which file it was :)

a8c-files.php
```diff
@@ -490,24 +502,53 @@ function delete_file( $file_name ) {
...
 
+		$parsed = parse_url( $url );
+		if ( empty( $parsed['host'] ) ) {
+			return $requests;
+		}
+
+		$uri = '/';
+		if ( isset( $parsed['path'] ) )
+			$uri = $parsed['path'];
+		if ( isset( $parsed['query'] ) )
```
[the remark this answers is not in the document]
darthhexx: We only purge the file at `/2018/01/01/image.jpg` and all variants are purged in Varnish at the same time. This is a key feature of Varnish for us.

a8c-files.php
```diff
@@ -490,24 +502,53 @@ function delete_file( $file_name ) {
...
+			$http_code = curl_getinfo( $curl, CURLINFO_HTTP_CODE );
+			curl_close( $curl );
+
+			if ( 200 != $http_code ) {
+				error_log( sprintf( __( 'Error purging %s from the cache service: Code %d' ), $url, $http_code ) );
+			}
+			return;
+		}
+
+		if ( ! isset( $file_cache_servers ) || empty( $file_cache_servers ) )
```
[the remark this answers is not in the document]
darthhexx: This would only execute for sandboxes, but it would still make sense to log as an error. I added it.

a8c-files.php
```diff
@@ -473,10 +484,13 @@ function delete_file( $file_name ) {
 		curl_close( $ch );
 
 		if ( 200 != $http_code ) {
-			error_log( sprintf( __( 'Error deleting the file from the remote servers: Code %d' ), $http_code ) );
+			error_log( sprintf( __( 'Error deleting the file %s from the remote servers: Code %d' ), $file_uri, $http_code ) );
```
mjangda: To ensure the error actually gets indexed: 

```suggestion
			trigger_error( sprintf( __( 'Error deleting the file %s from the remote servers: Code %d' ), $file_uri, $http_code ), E_USER_WARNING );
```

a8c-files.php
```diff
@@ -490,24 +504,55 @@ function delete_file( $file_name ) {
...
+					'Content-Type: application/json',
+					'Content-Length: ' . strlen( $json )
+				) );
+			curl_setopt( $curl, CURLOPT_TIMEOUT, 5 );
+			curl_exec( $curl );
+			$http_code = curl_getinfo( $curl, CURLINFO_HTTP_CODE );
+			curl_close( $curl );
+
+			if ( 200 != $http_code ) {
+				error_log( sprintf( __( 'Error purging %s from the cache service: Code %d' ), $url, $http_code ) );
```
mjangda: To ensure the error actually gets indexed: 

```suggestion
				trigger_error( sprintf( __( 'Error purging %s from the cache service: Code %d' ), $url, $http_code ), E_USER_WARNING );
```

a8c-files.php
```diff
@@ -490,24 +504,55 @@ function delete_file( $file_name ) {
...
+			}
+			return;
+		}
+
+		if ( ! isset( $file_cache_servers ) || empty( $file_cache_servers ) ) {
+			error_log( sprintf( __( 'Error purging the file cache for %s: There are no file cache servers defined.' ), $url ) );
```
mjangda: To ensure the error actually gets indexed: 

```suggestion
			trigger_error( sprintf( __( 'Error purging the file cache for %s: There are no file cache servers defined.' ), $url ), E_USER_WARNING );
```

a8c-files.php
```diff
@@ -490,24 +504,55 @@ function delete_file( $file_name ) {
 	private function purge_file_cache( $url, $method ) {
 		global $file_cache_servers;
 
+		$parsed = parse_url( $url );
+		if ( empty( $parsed['host'] ) ) {
+			return $requests;
+		}
+
+		$uri = '/';
+		if ( isset( $parsed['path'] ) )
```
mjangda: Minor coding standards thing: we should always use braces, even for single line `if` clauses. Without them, we open up room for easy-to-make mistakes.

a8c-files.php
```diff
@@ -490,24 +504,55 @@ function delete_file( $file_name ) {
 	private function purge_file_cache( $url, $method ) {
 		global $file_cache_servers;
 
+		$parsed = parse_url( $url );
+		if ( empty( $parsed['host'] ) ) {
+			return $requests;
+		}
+
+		$uri = '/';
```
mjangda: Minor: I don't think there's any major concerns with with this default since this is a `private` function, but we could end up sending a purge for `example.com/` if a non-file URL is passed through. Probably not worth changing but figured I'd flag it just in case.

a8c-files.php
darthhexx: Yeah, I just moved existing code and never changed it. I'll fix it.

a8c-files.php
darthhexx: It wouldn't cause any issues if we did, so I think it's okay.

mjangda:
r123942-deploy

PR #1030, opened 2018-11-22
Add/debug mode

shantanu2704:
## Description

This PR contains the initial work for enabling Debug Mode for logged out proxied in users. 
Added a new function to `vip-helpers/vip-utils.php` - `is_proxied_request()`.
The `vip-go-cb`(cache buster) cookie is set for 2 hours(no particular reason, seemed like a reasonable amount of time, open to changes) based on the handler(`/?a8c-debug=true`). This turns on debug mode and calls `nocache_headers()`. Similarly, `/?a8c-debug=false` turns off debug mode.

The debug mode indicator is similar to the sandbox indicator. It's in the bottom right corner in purple colour to distinguish itself from the sandbox indicator.
We could add more info to it if we like.

## Checklist*


mjangda:
Continued in #1164

PR #1043, opened 2018-12-12
Retain < 5.0 behaviour for tinyMCE scripts.

mjangda:
These used to be output individually and bypass concat. 5.0 registers
them as scripts which gets them picked up by our concat.

However, tinymce.min.js (root) does not handle that situation well.
It ends up initializing with the baseURL set to that of the script.

Which then leads to problems loading additional dependencies and
everything breaks.

By skipping concat, we restore earlier behaviour and keep sites working.

## Steps to Test

1. Apply PR
1. Load classic editor on 5.0 
1. Verify it works


trepmal:
👍
Tested on a site with known issues and this works

mjangda:
r124327-deploy

Issue #1040, opened 2018-12-10
Migration: Reset roles if it's empty


joshbetz:
WordPress.com doesn't store roles in the database. Therefore, when we import the database from WordPress.com, we wipe out the roles. If we try to run `wp jetpack-start connect --url=%s` when the administrator role doesn't exist, it fails silently. To fix this, I've run `wp roles reset --all`. We may want to do something similar in our migration cleanup if the roles option is empty.

https://github.com/wp-cli/role-command/blob/master/src/Role_Command.php#L271-L283

PR #1044, opened 2018-12-13
Migration: Reset all the site roles if the core roles doesn't exist.

vaurdan:
## Description

When importing sites from WordPress.com, since it doesn't store the roles in the database, the default core roles are not presented on the target site, requiring to run `wp role reset --all` after the database import.

This PR wants to remove that need and automate that step as part of the migration cleanup script. It will check for the existence of the `administrator` role. In case it isn't available on the site, it will run the `wp role reset --all` command.

I have decided on calling the command directly instead of duplicating the logic in the [`role reset` command](https://github.com/wp-cli/role-command/blob/master/src/Role_Command.php#L271-L283), as it allows reusing that code, making the cleanup command less complex, but I'm open to different approaches!

Ref: #1040 

## Checklist

## Steps to Test

1. Check out the PR.
2. Run `wp role list` in a Go site with a fresh import from WordPress.com, and check if the default roles exist.
3. Run `wp vip migration cleanup` and if there are no default roles, you should see them being reseted and created; if the roles exist you shouldn't see anything.
4. Check if the roles exist with `wp role list`

### DIFF
--- a/wp-cli/vip-migrations.php
+++ b/wp-cli/vip-migrations.php
@@ -70,6 +70,12 @@ class VIP_Go_Migrations_Command extends WPCOM_VIP_CLI_Command {
 			}
 		}
 
+		// Cleanup roles
+		if ( get_role( 'administrator' ) === null ) {
+			// If the `administrator` role is not present, it's required to reset the default roles.
+			WP_CLI::runcommand( 'role reset --all' );
+		}
+
 		WP_CLI::line( 'Calling Automattic\VIP\Migration\run_after_data_migration_cleanup()' );
 
 		if ( ! $dry_run ) {


wp-cli/vip-migrations.php
```diff
@@ -70,6 +70,12 @@ function cleanup( $args, $assoc_args ) {
 			}
 		}
 
+		// Cleanup roles
+		if ( get_role( 'administrator' ) === null ) {
```
mjangda: I think this may be an edge case but any concerns with a site that explicitly removes the `administrator` role? One issue would be that we'd then wipe out any custom roles as well. Not sure if this is anything we need to worry about but thought I'd flag it just in case.

joshbetz: Yeah, I was thinking about that. The original issue is that `jetpack-start connect` silently fails if the Administrator role doesn't exist (because our machine user is an Administrator).

I guess we could create a custom role for the machine user that's defined in code, so it's always there.

We still need to populate roles after a migration if the option is empty, but maybe that's a separate issue.

mjangda: Do none of the roles exist? Would that mean that `wp_roles()->roles` would return empty (https://developer.wordpress.org/reference/classes/wp_roles/)? Should we just check for that?

joshbetz: It returns the vip-support role because it's hard-coded. We could check for that, but again, if someone adds a role in their theme the same way we add vip-support, it will also be there.

vaurdan:
> Thoughts on moving this to `run_after_data_migration_cleanup()` (maybe right before we flush memcached?) in `vip-helpers/vip-migration.php` so it's not specific to this CLI?

I have thought about that, as I was unsure if it would make sense to have it directly on the CLI or having it on the `vip-migrations.php` file. 

I think it makes more sense to have this particular step on that file, so I'm going to refactor the code and move it there.

wp-cli/vip-migrations.php
vaurdan: I believe the odds of having someone removing the `administrator` role are higher than the odds of someone hard-coding roles like we do for `vip-support`, so I'm not sure which one is the best approach, but I think checking for the `administrator` role is the less risky option. 

I'm fully open to suggestions and alternative on how to address this edge case :)

joshbetz: > I believe the odds of having someone removing the administrator role are higher than the odds of someone hard-coding roles like we do for vip-support

The problem is that our helper functions hard-code the role and any sites coming from WP.com use those helper functions, so there will certainly be some sites that do this.

What if we unconditionally reset the roles after a migration?

joshbetz:
To reset the conversation a bit, there are two main problems to solve:

1. After migrations from WordPress.com, we have to `populate_roles()` because the way core roles are treated differ between the two platforms (on WP.com they're defined in code, on VIP Go we follow core and leave them in the database).
2. If the `wpcomvip` user doesn't have a role (and therefore no caps), connecting Jetpack fails.

The `wpcomvip` user's role is `administrator`, so forcing `populate_roles()` will probably fix it in all cases. Since the core roles are hard-coded on WordPress.com, `populate_roles()` will probably also not cause any issues for these migrations. The problem will be if any site in the future ever wants to remove core roles. Even then, it would only be a problem if we did an import to a launched, production site.

We could potentially deal with this differently though:

1. Instead of introducing `populate_roles()` to the migration process for all sites, we build this into the import process just for WordPress.com migrations
2. We add a `wpcomvip` role in code in `mu-plugins` and add the `wpcomvip` user to that role. We could probably even enforce that only the username `wpcomvip` can belong to that role.

Thoughts? @vaurdan @mjangda

mjangda:
> Instead of introducing populate_roles() to the migration process for all sites, we build this into the import process just for WordPress.com migrations.

👍

PR #1045, opened 2018-12-14
Improving the filesystem experience on VIP Go multisite

brettshumaker:
## Description

Working with the filesystem on VIP Go multisite is difficult to work with. In my testing, I was unable to place a file in `/wp-content/uploads/sites/{site_id}` and instead all files were placed in `/wp-content/uploads`. This makes storing site-specific files impossible on VIP Go. Plugins like Beaver Builder rely on being able to write to those folders to avoid potentially overwriting files from other sites on the network. Plugins like the newest version of Multilingual Press needs to be able to copy media files from one site on the network to another but are unable to do so.

I've updated the `upload_dir` filter in a8c-files.php respect multisite by adding in the `/sites/{site_id}` portion of the paths. By standardizing the output of `wp_get_upload_dir()`, I need to update `WP_Filesystem_VIP_Uploads->sanitize_uploads_path()` to allow access to both `/tmp/uploads` as well as `/tmp/uploads/sites/{any_site_id}`. Finally, `WP_Filesystem_VIP->is_uploads_path()` also needed to be updated to correctly determine the transport class that should be used for a given file path.

This code should be a good starting point for discussion about what needs to happen to improve the filesystem experience on VIP Go multisite.

## Checklist

## Steps to Test

1. Check out PR
1. Sandbox a VIP Go Multisite
1. Use the example code from the [VIP Documentation](https://vip.wordpress.com/documentation/using-wp_filesystem-instead-of-direct-file-access-functions/#example) to create new or copy existing files.
1. Verify that the created or copied files are in the expected location

### DIFF
--- a/a8c-files.php
+++ b/a8c-files.php
@@ -232,14 +232,14 @@ class A8C_Files {
 	function get_upload_dir( $upload ) {
 
 		// Maybe fix paths for multisite
-		if ( ! is_multisite() || ( is_main_network() && is_main_site() ) ) {
-			// If we're not on multisite, do nothing
-			$sites_path = '';
-		} elseif ( false !== stripos( $upload['subdir'], '/sites/' ) ) {
-			// If we maybe already have `/sites/` in the path, do nothing
-			$sites_path = '';
-		} else {
-			$sites_path = '/sites/' . get_current_blog_id();
+		$sites_path = '';
+		if (
+			// We only want to manipulate the paths if...
+			is_multisite() && // we're on multisite
+			1 !== get_current_blog_id() && // the blog ID is NOT 1, the uploads path for blog ID 1 doesn't need to include the site ID
+			false === stripos( $upload['subdir'], '/sites/' ) // the `subdir` hasn't already been updated to include /sites/{site_id}
+			) {
+				$sites_path = '/sites/' . get_current_blog_id();
 		}
 
 		$upload['path'] = constant( 'LOCAL_UPLOADS' ) . $sites_path . $upload['subdir'];
--- a/a8c-files.php
+++ b/a8c-files.php
@@ -230,8 +230,20 @@ class A8C_Files {
 	}
 
 	function get_upload_dir( $upload ) {
-		$upload['path'] = constant( 'LOCAL_UPLOADS' ) . $upload['subdir'];
-		$upload['basedir'] = constant( 'LOCAL_UPLOADS' );
+
+		// Maybe fix paths for multisite
+		if ( ! is_multisite() || ( is_main_network() && is_main_site() ) ) {
+			// If we're not on multisite, do nothing
+			$sites_path = '';
+		} elseif ( false !== stripos( $path, '/sites/' ) ) {
+			// If we maybe already have `/sites/` in the path, do nothing
+			$sites_path = '';
+		} else {
+			$sites_path = '/sites/' . get_current_blog_id();
+		}
+
+		$upload['path'] = constant( 'LOCAL_UPLOADS' ) . $sites_path . $upload['subdir'];
+		$upload['basedir'] = constant( 'LOCAL_UPLOADS' ) . $sites_path;
 
 		return $upload;
 	}
--- a/a8c-files.php
+++ b/a8c-files.php
@@ -235,7 +235,7 @@ class A8C_Files {
 		if ( ! is_multisite() || ( is_main_network() && is_main_site() ) ) {
 			// If we're not on multisite, do nothing
 			$sites_path = '';
-		} elseif ( false !== stripos( $path, '/sites/' ) ) {
+		} elseif ( false !== stripos( $upload['subdir'], '/sites/' ) ) {
 			// If we maybe already have `/sites/` in the path, do nothing
 			$sites_path = '';
 		} else {
--- a/files/class-wp-filesystem-vip.php
+++ b/files/class-wp-filesystem-vip.php
@@ -68,10 +68,11 @@ class WP_Filesystem_VIP extends \WP_Filesystem_Base {
 	}
 
 	private function is_uploads_path( $file_path ) {
-		$upload_dir  = wp_get_upload_dir();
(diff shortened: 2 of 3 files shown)


a8c-files.php
```diff
@@ -230,8 +230,20 @@ function save_image_file( $override, $filename, $image, $mime_type, $post_id ) {
...
+		// Maybe fix paths for multisite
+		if ( ! is_multisite() || ( is_main_network() && is_main_site() ) ) {
+			// If we're not on multisite, do nothing
+			$sites_path = '';
+		} elseif ( false !== stripos( $upload['subdir'], '/sites/' ) ) {
+			// If we maybe already have `/sites/` in the path, do nothing
+			$sites_path = '';
+		} else {
+			$sites_path = '/sites/' . get_current_blog_id();
+		}
```
hanifn: Perhaps this clause can be simplified as:
```
               $sites_path = '';
               if ( ( is_multisite() || ( ! is_main_network() && ! is_main_site() ) && 
                   ( false === stripos( $upload['subdir'], '/sites/' ) ) ) {
			$sites_path = '/sites/' . get_current_blog_id();
		}
```

Since we're not really doing anything with the first 2 cases

files/class-wp-filesystem-vip.php
```diff
@@ -68,10 +68,11 @@ private function is_tmp_path( $file_path ) {
 	}
 
 	private function is_uploads_path( $file_path ) {
-		$upload_dir  = wp_get_upload_dir();
-		$upload_base = $upload_dir['basedir'];
+		// This probably needs to be more robust as I'm not sure if there are ever scenarios where the uploads directory would _not_ be /tmp/uploads on VIP Go
+		// We just need to look for /tmp/uploads here - we don't really care about the /sites/{site_id} part to determine if it's an uploads path or not
+		$upload_path = trailingslashit( get_temp_dir() ) . 'uploads';
```
hanifn: I am a little concerned with this since it makes it less flexible with the assumption that the uploads directory will always be in `/tmp`. Especially in the context of the Filesystem Stream Wrapper work I'm doing.

brettshumaker: > I am a little concerned with this

Same. I'm not really sure how to use something more dynamic, though, unless we can implement a method that calculates where the uploads directory _should_ be and exclude the `/sites/{site_id}` portion. The reason is that the purpose of the `is_uploads_path()` method is to determine the transport needed to manipulate a given file path - `WP_Filesystem_VIP_Uploads` if it's in the "uploads" directory, or `WP_Filesystem_Direct` if it isn't. The problem is that if we use `wp_get_upload_dir()['basedir']` to check, paths that are _actually_ in the uploads directory and _need_ `WP_Filesystem_VIP_Uploads` to be manipulated will fail this check. Example: You're working on site ID 3 and want to duplicate it to a new subsite. Trying to copy media files from site ID 3 to site ID 7 would fail because `wp_get_upload_dir()['basedir']` is `/tmp/uploads/sites/3` and it's trying to determine the transport for a path (`/tmp/uploads/sites/7/some-file.jpg`) that doesn't include `/tmp/uploads/sites/3`. So the existing logic prevents you from being able to write to any other site on the multisite which is a blocker for some of our clients. Here we need a reliable way to determine whether or not `WP_Filesystem_VIP_Uploads` should be used to deal with the file path given.

Do you know how the `/tmp/uploads` path is determined to be the "uploads" directory? If it's hard coded somewhere, then I think we're fine to make this assumption but if it can be changed by clients we'll need a way to get the value they've set it to.

> Especially in the context of the Filesystem Stream Wrapper work I'm doing

Would this mean that `WP_Filesystem_VIP_Uploads` may need to be tweaked to account for that? A way for it to determine whether or not it's needed or if it can just use `WP_Filesystem_Direct`.

hanifn: > Trying to copy media files from site ID 3 to site ID 7 would fail because wp_get_upload_dir()['basedir'] is /tmp/uploads/sites/3 and it's trying to determine the transport for a path (/tmp/uploads/sites/7/some-file.jpg) that doesn't include /tmp/uploads/sites/3

Perhaps your change to `get_upload_dir` shouldn't change how `basedir` is set? As per your example, `path` can remain as `/tmp/uploads/sites/3/2018/12/` but `basedir` is simply `/tmp/uploads/`. Would that resolve this issue?

> Do you know how the /tmp/uploads path is determined to be the "uploads" directory? If it's hard coded somewhere, then I think we're fine to make this assumption but if it can be changed by clients we'll need a way to get the value they've set it to.

It is set as a constant, `LOCAL_UPLOADS` in `a8c-files.php`. So it is hard coded but I still feel that it's better to stick to using `wp_get_upload_dir` to get the upload paths since any changes made to the `LOCAL_UPLOADS` would be automatically reflected by that function. In the StreamWrapper plugin I've been working on, we're hooking into a filter to change the output of `wp_get_upload_dir` so there's that situation too.

> Would this mean that WP_Filesystem_VIP_Uploads may need to be tweaked to account for that?

Actually, from our tests so far, it looks like the `WP_Filesystem_VIP_Uploads` class is not even needed.

a8c-files.php
brettshumaker: Ok, I can check into this a bit more to make sure I understand it correctly.

hanifn:
@brettshumaker You might want to update the tests for these changes too as it's currently failing
```
1) Automattic\VIP\Files\WP_Filesystem_VIP_Test::test__is_uploads_path with data set "valid uploads path" ('/tmp/wordpress/wp-content/upl...04.jpg', true)
Failed asserting that false matches expected true.

/home/travis/build/Automattic/vip-go-mu-plugins/tests/files/test-wp-filesystem-vip.php:75
```

mjangda:
Thanks for the PR, @brettshumaker. I'm going to close this out as we're moving away from much of the older logic in a8c-files with the stream wrapper work. We can re-open if it continues to be a problem though.

PR #1052, opened 2018-12-25
Add conditional loading of streamwrapper

hanifn:
## Description

Load either the stream wrapper class and plugin filters or a8c-files based on a predefined constant.

The constant, `VIP_FILESYSTEM_USE_STREAM_WRAPPER`, should be defined in somewhere like `wp-config.php` where it's loaded before the MU plugins.

## Checklist

Please make sure the items below have been covered before requesting a review:

- [x] This change works and has been tested locally (or has an appropriate fallback).
- [x] This change works and has been tested on a Go sandbox.

## Steps to Test

Test that media upload functions work with and without the constant being defined.

PR #1053, opened 2018-12-26
Only init VIP Filesystem with FILES constants

mjangda:
To prevent breaking local environments where those are not set. The notices with WP_DEBUG can break uploads.

Temp fix until we deploy #1052


mjangda:
r125250-deploy

PR #1057, opened 2018-12-31
Update comment for check uniqueness method

hanifn:
## Description

Update the comment for `_check_uniqueness_with_backend` method to reflect it's implementation more accurately.


hanifn:
r126390-deploy

PR #1067, opened 2019-01-09
Don't process images with missing metadata

mjangda:
If an attachment doesn't have the appropriate meta (width/height/file), we can't really resize it, and we should skip attempting to process it anyway.

Prevents warnings that are triggering because of this:

```
Warning: Illegal string offset 'width' in /var/www/wp-content/mu-plugins/files/class-image.php on line 48 [wp-content/mu-plugins/files/class-image-sizes.php:31 Automattic\VIP\Files\Image->__construct(), wp-content/mu-plugins/a8c-files.php:918 Automattic\VIP\Files\ImageSizes->__construct(), wp-includes/class-wp-hook.php:286 a8c_files_maybe_inject_image_sizes() ... ]
```

### DIFF
--- a/a8c-files.php
+++ b/a8c-files.php
@@ -911,6 +911,13 @@ function a8c_files_maybe_inject_image_sizes( $data, $attachment_id ) {
 		return $data;
 	}
 
+	// Missing some critical data we need to determine sizes, so bail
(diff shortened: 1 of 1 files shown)


mjangda:
r126016-deploy

PR #1076, opened 2019-01-14
Fix partner-provision.sh on CLI containers

joshbetz:
After migrations, we run `wp vip migration cleanup` on a CLI container.
This runs `wp jetpack-start connect` internally, which calls the
partner-provision script.

Apparently the PATH variable is not set in the PHP exec(), which breaks
the partner-provision script. Here we're setting the full path to
executables that are used in the partner-provision script.

Alternatively, we could fix the PATH environment variable.

Related: https://github.com/Automattic/vip-go-api/pull/1702

### DIFF
--- a/jetpack-start/bin/partner-provision.sh
+++ b/jetpack-start/bin/partner-provision.sh
@@ -1,10 +1,10 @@
-#!/usr/bin/env sh
+#!/bin/sh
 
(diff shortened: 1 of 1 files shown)


joshbetz:
r126852-deploy

PR #1078, opened 2019-01-15
Add fixes for translation concatenation

dchymko:
## Description
Brings in translation concat  fixes
Automattic/nginx-http-concat#45


dchymko:
r126953-deploy

PR #1079, opened 2019-01-18
Update Filesystem/Stream Wrapper errors

hanifn:
## Description

Update `trigger_error` calls to include `#vip-go-streams` tag so that we can search for Filesystem specific errors easier.

## Checklist

### DIFF
--- a/files/class-vip-filesystem-stream-wrapper.php
+++ b/files/class-vip-filesystem-stream-wrapper.php
@@ -127,7 +127,10 @@ class VIP_Filesystem_Stream_Wrapper {
 			$result = $this->client->get_file( $path );
 
 			if ( is_wp_error( $result ) ) {
-				trigger_error( $result->get_error_message(), E_USER_WARNING );
+				trigger_error(
+					sprintf( '%s #vip-go-streams', $result->get_error_message() ),
+					E_USER_WARNING
+				);
 				return false;
 			}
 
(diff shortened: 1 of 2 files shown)


files/class-vip-filesystem-stream-wrapper.php
```diff
@@ -523,8 +552,10 @@ protected function string_to_resource( $data ) {
 		// Create a temporary file
 		$tmp_handler = tmpfile();
 		if ( false === fwrite( $tmp_handler, $data ) ) {
-			trigger_error( "Error creating temporary resource for data:\n" . $data,
-				E_USER_ERROR );
+			trigger_error(
+				sprintf( "Error creating temporary resource for data:\n%s\n#vip-go-streams", $data ),
```
mjangda: What's the reason for the newlines? 

```suggestion
				sprintf( "Error creating temporary resource for data: %s #vip-go-streams", $data ),
```

Might be better to just remove them for consistency?

hanifn: I put them in there because that error will be called with the data that we're trying to write to file. That data can be pretty arbitrary so the newlines will make the error logs easier to read.

Now that I think about it though, depending on the type file, that `data` variable could pretty much result in a gigantic wall of nonsense in the logs. Not sure if we should even include them in the logs now.

mjangda: Yeah, probably better to remove it.

hanifn:
r127766-deploy

PR #1084, opened 2019-01-24
Add more verbose errors for streamwrapper methods

mjangda:
Specifically, add the calling function and the path we were operating on. This will make debugging much easier.


mjangda:
r128216-deploy

PR #1088, opened 2019-01-25
Remove unused, deprecated functions

joshbetz:
These functions have been deprecated on VIP Go and are currently unused
by any VIP Go sites. We'll move this code to the vip-go-wpcom-compat
plugin so sites that are moving from WordPress.com are safe, but this
code is not loaded on the rest of VIP Go sites.

### DIFF
--- a/vip-helpers/vip-deprecated.php
+++ b/vip-helpers/vip-deprecated.php
@@ -41,63 +41,6 @@ function vip_allow_title_orphans() {
 }
 
 /**
- * VIP Legacy Related Posts (HTML formatted results)
- *
- * Don't use for new projects, just use WPCOM_RelatedPosts directly, since it has hooks
- * like jetpack_relatedposts_filter_args, jetpack_relatedposts_filter_filters
- *
- * @deprecated No longer supported since VIP 2.0.0 - Use Jetpack Related Posts
- * @param int $max_num Optional. Maximum number of results you want (default: 5).
- * @param array $additional_stopwords No longer used, we leave the stopwords magic to ES which knows more about word frequencies across articles.
- * @param bool $exclude_own_titles No longer used.
- * @return string Returns an HTML unordered list of related posts from the same blog.
- */
-function wpcom_vip_flaptor_related_posts( $max_num = 5, $additional_stopwords = array(), $exclude_own_titles = true ){
-    _deprecated_function( __FUNCTION__, '2.0.0' );
-
-    return '';
-}
-
-/**
- * VIP Legacy Related Posts (get post_id, title, url)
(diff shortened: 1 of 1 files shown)


joshbetz:
The next step will be removing all the functions that currently don't do anything other than throw deprecation warnings and don't even return anything.

```
wpcom_vip_load_helper
wpcom_vip_load_helper_wpcom
wpcom_vip_load_helper_stats
vip_allow_title_orphans
wpcom_vip_disable_enhanced_feeds
vip_remove_enhanced_feed_images
make_tags_local
wpcom_vip_allow_full_size_images_for_real
wpcom_vip_remove_polldaddy_rating
wpcom_vip_disable_postpost
wpcom_vip_enable_opengraph
wpcom_disable_mobile_app_promotion
wpcom_vip_allow_more_html_in_comments
wpcom_vip_disable_devicepx_js
wpcom_vip_disable_hovercards
wpcom_vip_disable_custom_customizer
wpcom_vip_remove_opensearch
vary_cache_on_function
disable_autosave
vip_multiple_moderators
wpcom_vip_disable_zemanta_for_all_users
wpcom_vip_disable_smilies
wpcom_vip_stats_roles
wpcom_print_sitemap_item
wpcom_print_xml_tag
wpcom_print_sitemap
wpcom_print_news_sitemap
do_sitemap_pings
sitemap_discovery
sitemap_handle_update
wpcom_vip_remove_feed_tracking_bug
wpcom_vip_plugins_ui_disable_activation
wpcom_vip_load_custom_cdn
wpcom_vip_remove_playlist_styles
wpcom_vip_disable_instapost
```

hanifn:
r128977-deploy

hanifn:
Reverted in r128978-deploy

`500 Internal Server Error` for many sites on Nagios

hanifn:
Found in logs: `Uncaught Error: Call to undefined function _wpcom_vip_load_plugin_sanitizer()`

Looks like somme functions were still in use

mjangda:
We'll want to take another pass at this but with a few changes:

- Start with logging only changes to confirm that the function are not actually in use.
- Reduce the size a bit, preferably one function at a time (which will take longer but is safer).
- Add some tests for the `wpcom_vip_load_plugin` function.
- We're also looking at more thorough integration and pre-deploy tests internally.

joshbetz:
> Found in logs: Uncaught Error: Call to undefined function _wpcom_vip_load_plugin_sanitizer()
> Looks like some functions were still in use

😬 we still use this within mu-plugins. Opened #1095 to move that out of `vip-deprecated.php`.

PR #1090, opened 2019-01-29
Add Stream cache

hanifn:
## Description

Add request based, local file cache to reduce number of requests to the Fileservice API.

## Checklist

### DIFF
--- a/files/class-api-client.php
+++ b/files/class-api-client.php
@@ -5,12 +5,14 @@ namespace Automattic\VIP\Files;
 use WP_Error;
 
 require( __DIR__ . '/class-curl-streamer.php' );
+require( __DIR__ . '/class-vip-filesystem-api-cache.php' );
 
 function new_api_client() {
 	return new API_Client(
 		'https://' . constant( 'FILE_SERVICE_ENDPOINT' ),
 		constant( 'FILES_CLIENT_SITE_ID' ),
-		constant( 'FILES_ACCESS_TOKEN' )
+		constant( 'FILES_ACCESS_TOKEN' ),
+		API_Cache::get_instance()
 	);
 }
 
@@ -21,12 +23,19 @@ class API_Client {
 	private $files_site_id;
 	private $files_token;
 
-	public function __construct( $api_base, $files_site_id, $files_token ) {
+	/**
+	 * @var API_Cache
+	 */
+	private $cache;
+
+	public function __construct( $api_base, $files_site_id, $files_token, $cache ) {
 		$api_base = untrailingslashit( $api_base );
 		$this->api_base = $api_base;
 
 		$this->files_site_id = $files_site_id;
 		$this->files_token = $files_token;
+
+		$this->cache = $cache;
 	}
 
 	protected function is_valid_path( $path ) {
@@ -122,6 +131,9 @@ class API_Client {
 		}
 
 		// response looks like {"filename":"/wp-content/uploads/path/to/file.ext"}
+		// save to cache
+		$this->cache->cache_file( $response_data->filename, file_get_contents( $local_path ) );
+
 		return $response_data->filename;
 	}
 
@@ -132,6 +144,13 @@ class API_Client {
 	}
 
 	public function get_file( $file_path ) {
+		// check in cache first
+		$file_content = $this->cache->get_file( $file_path );
+		if ( $file_content ) {
+			return $file_content;
+		}
+
+		// not in cache so get from API
 		$response = $this->call_api( $file_path, 'GET' );
 
 		if ( is_wp_error( $response ) ) {
--- a/files/class-vip-filesystem-api-cache.php
+++ b/files/class-vip-filesystem-api-cache.php
@@ -64,13 +64,18 @@ class API_Cache {
 	public function cache_file( $filepath, $data ) {
 		$file_name = basename( $filepath );
 
+		// create file with unique filename
+		$tmp_file = tempnam( $this->tmp_dir, 'vip' );
+
 		if ( ! isset( $this->files[ $file_name ])) {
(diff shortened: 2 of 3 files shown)


files/class-api-client.php
```diff
@@ -122,6 +131,9 @@ public function upload_file( $local_path, $upload_path ) {
 		}
 
 		// response looks like {"filename":"/wp-content/uploads/path/to/file.ext"}
+		// save to cache
+		$this->cache->cache_file( $response_data->filename, file_get_contents( $local_path ) );
```
mjangda: We'll want to clear out the cache for a successful `delete_file` call as well.

files/class-vip-filesystem-api-cache.php
```diff
@@ -0,0 +1,81 @@
...
+	 * @var string  Temp directory to cache file in
+	 */
+	private $tmp_dir = '/tmp';
+
+	/**
+	 * API_Cache constructor.
+	 */
+	protected function __construct() {
+		// sys_get_temp_dir is pretty inconsistent regarding trailing slashes
+		$this->tmp_dir = rtrim( sys_get_temp_dir(), '/' );
```
mjangda: Maybe better to use the WordPress `get_temp_dir` function? It also returns a consistent slash so we may want to update other uses to take that into account.

```suggestion
		$this->tmp_dir = get_temp_dir();
```

https://developer.wordpress.org/reference/functions/get_temp_dir/

tests/files/test-api-client.php
```diff
@@ -16,7 +16,11 @@ public static function setUpBeforeClass() {
 	public function setUp() {
 		parent::setUp();
 
-		$this->api_client = new API_Client( 'https://files.go-vip.co', 123456, 'super-sekret-token' );
+		$this->api_client = new API_Client(
+			'https://files.go-vip.co',
+			123456,
+			'super-sekret-token',
+			API_Cache::get_instance() );
```
mjangda: Thoughts on passing in a mock/stub version of `API_Cache` instead? I guess with the current approach it's nice that we end up testing the integration between `API_Client` and `API_Cache`

files/class-vip-filesystem-api-cache.php
```diff
@@ -0,0 +1,81 @@
+<?php
+
+namespace Automattic\VIP\Files;
+
+use WP_Error;
+
+
+class API_Cache {
```
mjangda: Would be nice to add some basic tests for this new class.

files/class-vip-filesystem-api-cache.php
```diff
@@ -0,0 +1,81 @@
+<?php
```
mjangda: Minor: Let's rename this to `class-api-cache.php` to keep the naming consistent. (It's also not specific to the vip-filesystem either.)

files/class-api-client.php
```diff
@@ -21,12 +23,19 @@ class API_Client {
 	private $files_site_id;
 	private $files_token;
 
-	public function __construct( $api_base, $files_site_id, $files_token ) {
+	/**
+	 * @var API_Cache
+	 */
+	private $cache;
+
+	public function __construct( $api_base, $files_site_id, $files_token, $cache ) {
```
mjangda: How do we suggest handling direct instantiation of this class? i.e. `new API_Client( ... )` Should the caller just pass in `API_Cache::get_instance()`. Is it worth creating an interface or abstract class and typecasting the param?

files/class-api-client.php
hanifn: Good point! I'll add that in.

files/class-vip-filesystem-api-cache.php
hanifn: Ah I did not know about that function. It's probably a better idea to use it then.

files/class-vip-filesystem-api-cache.php
hanifn: I already did. Perhaps you missed it?

files/class-vip-filesystem-api-cache.php
hanifn: That is true. I'll change it then

tests/files/test-api-client.php
hanifn: I'm already working on using a mock/stub cache for those test

files/class-api-client.php
hanifn: I guess it depends on the likelihood of there being other Client types or Cache types. On the other hand, using interfaces might make writing tests for these much simpler

files/class-vip-filesystem-api-cache.php
mjangda: Ah, sorry. Forgot to refresh the page :(

files/class-api-client.php
mjangda: No, I wouldn't worry too much about it.

files/class-vip-filesystem-api-cache.php
mjangda: Test looks good! 💯

files/class-vip-filesystem-api-cache.php
```diff
@@ -0,0 +1,81 @@
...
+		$file_name = basename( $filepath );
+
+		// create file with unique filename
+		$tmp_file = tempnam( $this->tmp_dir, 'vip' );
+
+		if ( ! isset( $this->files[ $file_name ])) {
+			$this->files[ $file_name ] = $tmp_file;
+		} else {
+			// update cache; delete old file
+			unlink( $this->files[ $file_name ] );
```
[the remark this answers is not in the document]
hanifn: Good idea 👍🏽

hanifn:
r128817-deploy

PR #1093, opened 2019-01-31
Set shorter timeouts for low-cost API calls

### LINKED ISSUE #1089: Files\Api_Client: adjust timeouts based on request
Not all requests that the `Api_Client` makes need the standard 10s timeout. Example, `file_exists` should be fast (< 1s). We should adjust the timeout for various requests based on their expected time.

mjangda:
We don't need to use the default 10s timeout for all API calls. Some of them like `is_file` and `unique_filename` should be fast and we should set those timeouts accordingly.

Fixes #1089 

## Checklist

### DIFF
--- a/files/class-api-client.php
+++ b/files/class-api-client.php
@@ -167,7 +167,9 @@ class API_Client {
 	}
 
 	public function delete_file( $file_path ) {
-		$response = $this->call_api( $file_path, 'DELETE' );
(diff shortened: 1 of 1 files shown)


mjangda:
Ha! 🖖

mjangda:
r128964-deploy

Issue #1095, opened 2019-02-01
Move WP.com deprecated functions to vip-go-wpcom-helper


joshbetz:
Related: https://github.com/Automattic/vip-go-wpcom-compat/pull/29

We made a first attempt at this in #1088. That broke because we were removing `_wpcom_vip_load_plugin_sanitizer`, but that's used internally in mu-plugins. This is an important change to get ahead of WordPress.com sites moving to VIP Go.

Opening this issue for more thorough planning.

First of all, we should move `_wpcom_vip_load_plugin_sanitizer` out of the deprecated functions list and remove the deprecation warning. I don't think we can consider this to be deprecated given that we're still using it.

It would also be good to double check that no other functions that are currently marked as deprecated are actually still being used.

Then, per [Mo's feedback](https://github.com/Automattic/vip-go-mu-plugins/pull/1088#issuecomment-459617676) on the original issue, we should add some logging to all these deprecated functions so we can ensure they're not still being used before removing them.

The original list was not used in any client code at the time, but we should double check that for each function when we deploy as well.

rinatkhaziev:
#1096 is brought up to par with master and is ready to land, once that's done I'll close this.

GaryJones:
Since #1096 has been merged, I'll close this issue.

PR #1096, opened 2019-02-01
Fix dependencies on deprecated functions

joshbetz:
These functions are actually not deprecated given that we use them internally.

See #1095


rinatkhaziev:
I've fixed the merge conflict to make sure this applies cleanly, automated tests are passing, I have also applied the patch manually and tested in a browser.

nickdaugherty:
r662-stacks

PR #1099, opened 2019-02-04
Don't update the $wp_user_roles global

mjangda:
This should be on the core / lower-level functions to do. Usage is also very rare.

Us overriding the global can lead to unexpected behaviour in multisite  contexts like switching blogs and overriding/removing core roles.

Updating it is not actually necessary since core tracks changes to role objects in the `wp_roles()` global instead.

## Checklist

## Steps to Test / Verify

- Start by verifying that the VIP Support roles are still there with the PR in place (`wp role list`).
- Add the `$wp_user_roles` global with some custom roles to `vip-config` and verify that the VIP Support roles are still there (again, `wp role list`).

Next, verify the previously broken flow doesn't break:

1. On a multisite install, make sure you have two sites.
1. Verify initial roles for both sites via `wp role list`
1. `wp shell`
1. Add a new role using our helper: ` wpcom_vip_add_role( 'testing-2', 'Testing 2', [] )`
1. Verify role was added successfully (can inspect `wp_roles()`).
1. Switch to different subsite: `switch_to_blog( 2 )` // where `2` is the ID
1. Add a new role using our helper: ` wpcom_vip_add_role( 'testing-3', 'Testing 3', [] )`
1. Verify role was added successfully and existing roles were retained (can inspect `wp_roles()` or run `wp role list`).

Repeat with `$wp_user_roles` global set.

### DIFF
--- a/vip-helpers/vip-roles.php
+++ b/vip-helpers/vip-roles.php
@@ -1,9 +1,6 @@
 <?php
 /**
- * Helper functions that make it easy to add roles for WordPress.com sites.
- *
- * We use the core API functions as well as modifying the $wp_user_roles global
- * in case roles are re-initialized and our mods are lost.
+ * Helper functions that make it easy to add roles for VIP sites.
  */
 
 /**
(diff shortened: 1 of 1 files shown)


mjangda:
Note: need to test this change on a site using hard-coded roles via `$wp_user_roles` to see if we just need to harden things rather than removing them.

mjangda:
r129986-deploy

PR #1106, opened 2019-02-07
Optimize Stream_Wrapper::stream_open

### LINKED ISSUE #1105: Stream Wrapper: optimize stream_open
Right now, we do two API calls: `is_file` and `get_file`. We can just do one by calling `get_file and checking for the appropriate response / error code.

mjangda:
Instead of doing two separate API calls (is_file and get_file), just do one and use the signal from get_file to determine if the file exists.

In the case where files don't exist, we'll still do 1 call as previously, but in the case where they do, we drop down the count from 2 to 1.

Fixes #1105

## Checklist

## Steps to Test

1. Check out PR.
1. `wp shell`
1. `file_put_contents( 'vip://wp-content/uploads/file-' . time() '.txt', 'testing ' . time() );`
1. Verify that works and that fetching the file returns the correct content.

### DIFF
--- a/files/class-api-client.php
+++ b/files/class-api-client.php
@@ -158,7 +158,10 @@ class API_Client {
 		}
 
 		$response_code = wp_remote_retrieve_response_code( $response );
-		if ( 200 !== $response_code ) {
+		if ( 404 === $response_code ) {
+			/* translators: 1: file path */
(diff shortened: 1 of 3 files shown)


mjangda:
Ah, you need to have `define( 'VIP_FILESYSTEM_USE_STREAM_WRAPPER', true );` in `vip-config.php`.

nickdaugherty:
Thanks! Worked great with the constant defined.

mjangda:
r130095-deploy

PR #1107, opened 2019-02-08
Fix `wp --skip-plugins`

### LINKED ISSUE #1069: Fix --skip-plugins
Currently, the WP CLI `--skip-plugins` flag doesn't work as expected (plugins end up loaded anyway). This can cause issues because important WP CLI command can fail.

A few things we need to do when WP CLI is called with `--skip-plugins` to avoid 3rd party code from conflicting with critical WP CLI commands:

- [x] Don't load `client-mu-plugins`
- [x] Don't load plugins via `wpcom_vip_load_plugin()`
- [x] Don't load plugins via `wpcom_vip_load_helpers_for_sites_core_plugins()`

mjangda:
Leads to unnecessarily loading plugins which can sometimes break WP-CLI because of unexpected fatals. This also defeats the purpose of the skip-plugins flag.

We're now preventing breaking WP-CLI by skipping plugin loading when that context is set.

Fixes #1069 

## Checklist

## Steps to Test

*Scenario 1: UI-activated plugin with fatal*

1. Add a plugin and activate via UI (wp-admin > Plugins)
1. Add a fatal to the main plugin file (e.g. `poop();`)
1. Call `wp shell --skip-plugins`.and verify that shell loads without error. Previously, this would fatal.
1. Call `wp shell` and verify that it fatals.

*Scenario 2: client-mu-plugin with fatal*

1. Add a client-mu-plugin file.
1. Add a fatal to the plugin file (e.g. `poop();`)
1. Call `wp shell --skip-plugins` and verify that shell loads without error. Previously, this would fatal.
1. Call `wp shell` and verify that it fatals.

*Scenario 1: UI-activated plugin with fatal*

1. Add a plugin and activate via code (`wpcom_vip_load_plugin( 'my-plugin' );`)
1. Add a fatal to the main plugin file (e.g. `poop();`)
1. Call `wp shell --skip-plugins` and verify that shell loads without error. Previously, this would fatal.
1. Call `wp shell` and verify that it fatals.

### DIFF
--- a/vip-helpers/vip-utils.php
+++ b/vip-helpers/vip-utils.php
@@ -956,6 +956,10 @@ function wpcom_vip_load_plugin( $plugin = false, $folder = false, $load_release_
 		}
 	}
 
+	if ( ! wpcom_vip_should_load_plugins() ) {
+		return;
+	}
+
 	/**
 	 * wpcom compat
 	 *
--- a/vip-helpers/vip-utils.php
+++ b/vip-helpers/vip-utils.php
@@ -1138,6 +1138,29 @@ function wpcom_vip_can_use_shared_plugin( $plugin ) {
 }
 
 /**
+ * Helper function to check if we can load plugins or not.
+ */
+function wpcom_vip_should_load_plugins() {
+	static $should_load_plugins;
+
+	if ( isset( $should_load_plugins ) ) {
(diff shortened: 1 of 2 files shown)


mjangda:
r130257-deploy

PR #1110, opened 2019-02-11
Register Stream Wrapper `init` hook earlier

### LINKED ISSUE #1104: Stream Wrapper: Register earlier in the WP bootup
Currently, we wait until `init` to load the stream wrapper. The downside to this is that calls to `wp_upload_dir` earlier in the boot process will trigger warnings because they attempt to `w_mkdir_p` on the local, read-only filesystem.

hanifn:
## Description

Move a8c files init to use the `muplugins_loaded` hook instead of `init`.

Fixes #1104 

## Checklist

### DIFF
--- a/a8c-files.php
+++ b/a8c-files.php
@@ -935,7 +935,7 @@ function a8c_files_maybe_inject_image_sizes( $data, $attachment_id ) {
 
 if ( defined( 'FILES_CLIENT_SITE_ID' ) && defined( 'FILES_ACCESS_TOKEN' ) ) {
 	// Kick things off
-	add_action( 'init', 'a8c_files_init' );
+	add_action( 'muplugins_loaded', 'a8c_files_init' );
 
 	// Disable automatic creation of intermediate image sizes.
 	// We generate them on-the-fly on VIP.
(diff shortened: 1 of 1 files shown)


a8c-files.php
```diff
@@ -935,7 +935,7 @@ function a8c_files_maybe_inject_image_sizes( $data, $attachment_id ) {
 
 if ( defined( 'FILES_CLIENT_SITE_ID' ) && defined( 'FILES_ACCESS_TOKEN' ) ) {
 	// Kick things off
-	add_action( 'init', 'a8c_files_init' );
+	add_action( 'muplugins_loaded', 'a8c_files_init' );
```
mjangda: One downside to waiting till `muplugins_loaded` is that any other code that runs while loading remaining `mu-plugins` and `client-mu-plugins` will not be able to use the stream wrapper. Maybe we should just call `a8c_files_init` right away instead of waiting till a hook to avoid that?

```suggestion
	a8c_files_init();
```

hanifn: Sounds like a good idea. I just tested it on my sandboxed test site and it doesn't seem to break anything

mjangda:
r130592-deploy

mjangda:
This was reverted due to fatals and then restored in r130613 (along with #1115)

PR #1114, opened 2019-02-11
Add debug log for API client

hanifn:
## Description

Debug log for the VIP Filesystem API Client. Depends on the `VIP_FILESYSTEM_STREAM_WRAPPER_DEBUG` constant to be set to `true`.

## Checklist

Please make sure the items below have been covered before requesting a review:

- [x] This change works and has been tested locally (or has an appropriate fallback).
- [x] This change works and has been tested on a Go sandbox.

PR #1117, opened 2019-02-13
Allow E_USER_NOTICE for debugging in API Client

mjangda:
By default, WordPress skips them via error_reporting, which is causing our trigger_error calls to be useless. If debugging is enabled for stream wrappers, we're now allowing E_USER_NOTICE so the debug gets tracked properly.

Previously #1114 

## Checklist

## Steps to Test

1. With `define( 'VIP_FILESYSTEM_STREAM_WRAPPER_DEBUG', false )`, verify that nothing is logged when browsing attachment pages and uploading and deleting files.
1. With `define( 'VIP_FILESYSTEM_STREAM_WRAPPER_DEBUG', true )` and `define( 'WP_DEBUG', false );`, verify that nothing is logged when browsing attachment pages and uploading and deleting files.
1. With `define( 'VIP_FILESYSTEM_STREAM_WRAPPER_DEBUG', true )` and `define( 'WP_DEBUG', true );` verify that API client actions are logged as notices.

### DIFF
--- a/files/class-api-client.php
+++ b/files/class-api-client.php
@@ -261,7 +261,19 @@ class API_Client {
 		return $obj->filename;
 	}
 
+	// Allow E_USER_NOTICE to be logged since WP blocks it by default.
+	private function allow_E_USER_NOTICE() {
+		static $updated_error_reporting = false;
+		if ( ! $updated_error_reporting ) {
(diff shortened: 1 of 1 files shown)


mjangda:
r130670-deploy

PR #1119, opened 2019-02-13
Add lockout plugin

### LINKED ISSUE #1112: Lockout: Implement base functionality
When the `VIP_LOCKOUT_STATE` constant is defined and set to `warning`, we should:

- display an admin notice to **users with the `manage_options` capability** with whatever message is set in the `VIP_LOCKOUT_MESSAGE` constant.
- We should keep track of which users see a lockout message and when. We can track the seen/not seen state as one user meta entry and time as another user meta entry.
- This should not apply to VIP Support users.

When the `VIP_LOCKOUT_STATE` constant is defined and set to `locked`, we should:

- display an admin notice to **all users** with whatever message is set in the `VIP_LOCKOUT_MESSAGE` constant.
- remove all caps from users except `read` (so they can't take any action on the site).
- This should not apply to VIP Support users.

If the `VIP_LOCKOUT_STATE` constant is not set or set to another value, we should do nothing.

---

Other notes:

- We may want to use anonymous functions as callbacks for `add_action`/`add_filter` to make them harder to unhook.

hanifn:
## Description

Add notices for sites and locking out users

Resolves #1112 

## Checklist


hanifn:
There is a problem with current implementation of `filter_user_has_cap`. I'm encountering errors when certain plugins are activated. Right now it's only happening with the Elementor plugin on my test site but the risk of similar things happening with other plugins is there. The root of the problem is the `$submenu` global which returned a null value for the Elementor plugin slug.

I'm considering force setting the `$submenu` global in here but i'm not if that's a good idea.

mjangda:
> I'm considering force setting the `$submenu` global in here but i'm not if that's a good idea.

I don't think we need to optimize for plugin-specific problems (at least ones that we don't have much control over or that aren't going to be major issues). If we can isolate it, we can report the issue upstream.

mjangda:
Thanks for pulling this together @hanifn. What are some unit tests that would be good to add here?

hanifn:
> Thanks for pulling this together @hanifn. What are some unit tests that would be good to add here?

I've added some tests for the filter method and the meta method. Not sure how to effectively test the notice rendering though. Any ideas?

vip-lockout.php
```diff
@@ -0,0 +1,113 @@
...
+     * If site is in lockout mode then all user will only have capabilities of a subscriber.
+     *
+	 * @param array $user_caps
+	 * @param array $caps
+	 * @param array $args
+	 * @param WP_User $user
+	 *
+	 * @return array
+	 */
+	public function filter_user_has_cap( array $user_caps, array $caps, array $args, WP_User $user ) {
```
mjangda: Just to be on the safe side, I think it's better to avoid type-hints for callbacks on hooks:

```suggestion
	public function filter_user_has_cap( $user_caps, $caps, $args, $user ) {
```

Since we don't control the data being passed in, it may not always be the format we expect (usually a case that someone is doing something wrong, but the type-hints would trigger fatals.

vip-lockout.php
```diff
@@ -0,0 +1,113 @@
...
+	protected function render_warning_notice( WP_User $user ) {
+		if ( ! $user->has_cap( 'manage_options' ) ) {
+			return;
+		}
+
+		?>
+		<div id="lockout-warning" class="notice-warning wrap clearfix" style="align-items: center;background: #ffffff;border-left-width:4px;border-left-style:solid;border-radius: 6px;display: flex;margin-top: 30px;padding: 30px;" >
+			<div class="dashicons dashicons-warning" style="display:flex;float:left;margin-right:2rem;font-size:38px;align-items:center;margin-left:-20px;color:#ffb900;"></div>
+			<div class="vp-message" style="display: flex;align-items: center;" >
+				<h3><?php _e( VIP_LOCKOUT_MESSAGE ); ?></h3>
```
mjangda: Translation functions only work for string literals, not variables. In this case, we should pass the message through kses just to be safe:

```suggestion
				<h3><?php wp_kses_post( VIP_LOCKOUT_MESSAGE ); ?></h3>
```

vip-lockout.php
```diff
@@ -0,0 +1,113 @@
...
+		</div>
+		<?php
+	}
+
+	protected function render_locked_notice() {
+		?>
+		<div id="lockout-warning" class="notice-error wrap clearfix" style="align-items: center;background: #ffffff;border-left-width:4px;border-left-style:solid;border-radius: 6px;display: flex;margin-top: 30px;padding: 30px;" >
+			<div class="dashicons dashicons-warning" style="display:flex;float:left;margin-right:2rem;font-size:38px;align-items:center;margin-left:-20px;color:#dc3232;"></div>
+			<div class="vp-message" style="display: flex;align-items: center;" >
+				<h3><?php _e( VIP_LOCKOUT_MESSAGE ); ?></h3>
```
mjangda: Same note as earlier:

```suggestion
				<h3><?php wp_kses_post( VIP_LOCKOUT_MESSAGE ); ?></h3>
```

vip-lockout.php
```diff
@@ -0,0 +1,113 @@
...
+	 *
+	 * @return array
+	 */
+	public function filter_user_has_cap( array $user_caps, array $caps, array $args, WP_User $user ) {
+		if ( is_automattician( $user->ID ) ) {
+			return $user_caps;
+		}
+
+		if ( defined( 'VIP_LOCKOUT_STATE' ) && 'locked' === VIP_LOCKOUT_STATE ) {
+			$subscriber = get_role( 'subscriber' );
```
mjangda: Probably worth adding a fallback in case the subscriber role doesn't exist. Can be a future PR though.

vip-lockout.php
```diff
@@ -0,0 +1,113 @@
...
+     *
+	 * @param array $user_caps
+	 * @param array $caps
+	 * @param array $args
+	 * @param WP_User $user
+	 *
+	 * @return array
+	 */
+	public function filter_user_has_cap( array $user_caps, array $caps, array $args, WP_User $user ) {
+		if ( is_automattician( $user->ID ) ) {
```
mjangda: It would be better to use `is_proxied_automattician` here, although looks like we currently don't allow passing an ID to that.

tests/test-vip-lockout.php
```diff
@@ -0,0 +1,130 @@
...
+		require_once __DIR__ . '/../vip-support/class-vip-support-role.php';
+
+		define( 'VIP_LOCKOUT_STATE', 'locked' );
+
+		$email = 'user@automattic.com';
+		$user = $this->factory->user->create_and_get( [
+			'role' => 'administrator',
+			'user_email' => $email,
+		] );
+		add_user_meta( $user->ID, \Automattic\VIP\Support_User\User::META_EMAIL_VERIFIED, $email );
```
mjangda: It might be easier just to call `\Automattic\VIP\Support_User\User:add` here (since it handles the email verify meta.

vip-lockout.php
```diff
@@ -0,0 +1,113 @@
...
+				case 'warning':
+					$user = wp_get_current_user();
+
+					$this->render_warning_notice( $user );
+
+					$this->user_seen_warning( $user );
+					break;
+
+				case 'locked':
+					$this->render_locked_notice();
```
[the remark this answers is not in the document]
hanifn: Would this be a different meta value? Currently it's just set to `true` but we could change it to eiither `warning` or `locked`

vip-lockout.php
```diff
@@ -0,0 +1,113 @@
...
+
+	const USER_SEEN_WARNING_KEY = 'seen_lockout_warning';
+
+	const USER_SEEN_WARNING_TIME_KEY = 'seen_lockout_warning_time';
+
+	/**
+	 * VIP_Lockout constructor.
+	 */
+	public function __construct() {
+		add_action( 'admin_notices', [ $this, 'add_admin_notice' ], 1 );
```
mjangda: Did some testing and noticed two things:

1) It would be good to only hook into these if `VIP_LOCKOUT_STATE` and `VIP_LOCKOUT_MESSAGE` are set.

2) When locked, we'll need to hook into `pre_site_option_site_admins` and return an empty array when the site is locked.

hanifn: > 1. It would be good to only hook into these if `VIP_LOCKOUT_STATE` and `VIP_LOCKOUT_MESSAGE` are set. No need to add the extra overhead to all pageloads.

Good idea 👍🏽

> 2. When locked, we'll need to hook into `pre_site_option_site_admins` and return an empty array when the site is locked.

I can't find any references for a `pre_site_option_` filter. Did you mean `pre_option_`?

vip-lockout.php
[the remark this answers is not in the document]
mjangda: > we could change it to eiither `warning` or `locked`

Yeah, let's run with that for now.

vip-lockout.php
hanifn: It's also dependent on the `A8C_PROXIED_REQUEST` constant which I can't redefine in the tests

vip-lockout.php
mjangda: > I can't find any references for a `pre_site_option_` filter.

It's a dynamic filter (Googling it is not easy :))

https://developer.wordpress.org/reference/hooks/pre_site_option_option/

vip-lockout.php
```diff
@@ -0,0 +1,113 @@
...
+	protected function render_warning_notice( WP_User $user ) {
+		if ( ! $user->has_cap( 'manage_options' ) ) {
+			return;
+		}
+
+		?>
+		<div id="lockout-warning" class="notice-warning wrap clearfix" style="align-items: center;background: #ffffff;border-left-width:4px;border-left-style:solid;border-radius: 6px;display: flex;margin-top: 30px;padding: 30px;" >
+			<div class="dashicons dashicons-warning" style="display:flex;float:left;margin-right:2rem;font-size:38px;align-items:center;margin-left:-20px;color:#ffb900;"></div>
+			<div class="vp-message" style="display: flex;align-items: center;" >
+				<h3><?php wp_kses_post( VIP_LOCKOUT_MESSAGE ); ?></h3>
```
mjangda: Whoops, forgot to add an echo :)

```suggestion
				<h3><?php echo wp_kses_post( VIP_LOCKOUT_MESSAGE ); ?></h3>
```

vip-lockout.php
```diff
@@ -0,0 +1,113 @@
...
+		</div>
+		<?php
+	}
+
+	protected function render_locked_notice() {
+		?>
+		<div id="lockout-warning" class="notice-error wrap clearfix" style="align-items: center;background: #ffffff;border-left-width:4px;border-left-style:solid;border-radius: 6px;display: flex;margin-top: 30px;padding: 30px;" >
+			<div class="dashicons dashicons-warning" style="display:flex;float:left;margin-right:2rem;font-size:38px;align-items:center;margin-left:-20px;color:#dc3232;"></div>
+			<div class="vp-message" style="display: flex;align-items: center;" >
+				<h3><?php wp_kses_post( VIP_LOCKOUT_MESSAGE ); ?></h3>
```
mjangda:
```suggestion
				<h3><?php echo wp_kses_post( VIP_LOCKOUT_MESSAGE ); ?></h3>
```

vip-lockout.php
hanifn: Maybe just pass an array like `[ 'read' => true, 'level_0' => true ]`?

mjangda:
r131154-deploy

Issue #1122, opened 2019-02-15
Vary Cache: Improvements to group get/set functionality


dchymko:
We should enhance the group functions to make them more flexible for a variety of use cases:

- [x] Prefix the cookie value with a version `v1__group_--_value` to allow dynamic changes to the cookie over time, if needed (e.g. if we change the delimiter or change the structure).
- [x] Add a `register_group()` method to accept a single group
- [x] Change the group functions to work when the encrypted flag is set
- [x] Optimize the group parsing function to only load/run when required (i.e. if cookie has been parsed with no changes, skip)
- [x] Simplify `is_user_in_group` (a bit difficult to follow the logic)

dchymko:
Cookie prefix added as part of #1127

mjangda:
This is done.

PR #1124, opened 2019-02-15
Vary Cache: Simplify logic for is_user_in_group

mjangda:
By making the value optional and add some tests.

See #1122 

## Checklist

## Steps to Test

Adds new unit tests for the method.

### DIFF
--- a/cache/class-vary-cache.php
+++ b/cache/class-vary-cache.php
@@ -96,29 +96,42 @@ class Vary_Cache {
 	}
 
 	/**
-	 * Checks if the request has some in with agroup cookie matching a given group and optionally a value
+	 * Checks if the request has a group cookie matching a given group, regardless of segment value.
 	 *
 	 * @param  string $group Group name.
-	 * @param  string $value (optional) Which segment within the group to check? Omit to check if the user is in the group at all.
 	 *
 	 * @return bool   True on success. False on failure.
 	 */
-	public static function is_user_in_group( $group, $value = null ) {
+	public static function is_user_in_group( $group ) {
 		self::parse_group_cookie();
 
 		// The group isn't defined, or the user isn't in it.
-		if ( ! isset( self::$groups[ $group ] ) || null === self::$groups[ $group ] ) {
+		if ( ! array_key_exists( $group, self::$groups ) ) {
 			return false;
 		}
 
-		// The user is the group, and we don't care about the specific segment.
-		if ( null === $value ) {
-			return true;
+		return true;
+	}
+
+	/**
+	 * Checks if the request has a group cookie matching a given group and segment. e.g. 'dev-group', 'yes'
+	 *
+	 * @param  string $group Group name.
+	 * @param  string $segment Which segment within the group to check.
+	 *
+	 * @return bool   True on success. False on failure.
+	 */
+	public static function is_user_in_group_segment( $group, $segment ) {
+		self::parse_group_cookie();
+
+		if ( ! self::is_user_in_group( $group ) ) {
+			return false;
 		}
 
 		// Check for a specific group segment.
-		return self::$groups[ $group ] === $value;
+		return self::$groups[ $group ] === $segment;
 	}
+
 
 	/**
 	 * Returns the associated groups for the request.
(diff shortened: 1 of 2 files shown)


cache/class-vary-cache.php
```diff
@@ -98,18 +98,26 @@ public static function set_group_for_user( $group, $value ) {
...
 	 *
 	 * @return bool   True on success. False on failure.
 	 */
-	public static function is_user_in_group( $group, $value ) {
+	public static function is_user_in_group( $group, $value = null ) {
 		self::parse_group_cookie();
-		if ( ! isset( self::$groups[ $group ] ) ) {
+
+		// The group isn't defined, or the user isn't in it.
+		if ( empty( self::$groups[ $group ] ) ) {
```
dchymko: Would this break for groups that are named "0"?

mjangda: Yeah, it would, but that's like an unlikely case. We could harden things a bit by changing the default value for groups to `null` and checking explicitly for that?

```suggestion
		if ( ! isset( self::$groups[ $group ] ) || null === self::$groups[ $group ] ) {
```

Although, this maybe gets confusing with the null default for the `$value` param.

Thinking about this some more, maybe we should separate the two uses: `is_user_in_group` (is the user is in the group at all, don't care about the value) vs `is_user_in_group_segment` (is the use in the group with a specific segment value?).

dchymko: Yeah, I think splitting it into two functions would make sense (since we are planning on doing the something similar for `register_group` and `register_groups`

dchymko: @mjangda Can you take a quick peek at the updates I made?

tests/cache/test-vary-cache.php
```diff
@@ -0,0 +1,175 @@
...
+				'dev-group',
+				'0',
+				true,
+			],
+		];
+	}
+
+	public function get_test_data__is_user_in_group() {
+		return [
+			'user-not-in-group' => [
```
mjangda: Should we have a `group-not-defined` test?

tests/cache/test-vary-cache.php
```diff
@@ -0,0 +1,175 @@
...
+			'user-in-group-with-empty-segment' => [
+				[
+					'vip-go-seg' => 'dev-group_--_',
+				],
+				[
+					'dev-group',
+				],
+				'dev-group',
+				'',
+				true,
```
mjangda: Should an empty segment mean that the user is in the group? What would be a use case for that?

tests/cache/test-vary-cache.php
```diff
@@ -0,0 +1,175 @@
...
+				[
+					'vip-go-seg' => 'design-group_--_yes',
+				],
+				[
+					'dev-group',
+				],
+				'dev-group',
+				false,
+			],
+			'user-not-in-group' => [
```
mjangda:
```suggestion
			'user-in-group' => [
```

tests/cache/test-vary-cache.php
```diff
@@ -0,0 +1,175 @@
...
+			],
+			'user-in-group-and-empty-segment' => [
+				[
+					'vip-go-seg' => 'dev-group_--_',
+				],
+				[
+					'dev-group',
+				],
+				'dev-group',
+				true,
```
mjangda: Same as the other note; what would be a valid use case for this?

cache/class-vary-cache.php
mjangda: Done!

tests/cache/test-vary-cache.php
dchymko: Nothing specific but I think if the user explicitly sets a group with an empty value and then checks for it, we should match it. Unless we want to make `value` mandatory

cache/class-vary-cache.php
```diff
@@ -61,6 +61,7 @@ public static function remove_no_cache_for_user() {
 	 * @param  array $groups  One or more groups to vary on.
 	 */
 	public static function register_groups( $groups ) {
+		self::$groups = [];
```
mjangda: This is probably going to create problems in cases where different parts of a codebase (e.g. different plugins) need to register groups.

mjangda: Should be good to merge after this.

dchymko: Oops, that was left that in unintentionally when adding the group clear method

dchymko:
r131592-deploy

PR #1127, opened 2019-02-16
Enable Vary Cache with encrypted groups/data

dchymko:
## Description

Implement the encrypted cookie approach to allow encrypted segmentation.

See #1122 

## Checklist

Example:

1. Start with the [GDPR example](https://github.com/wpcomvip/vip-vary-cache-testing)
2. Set `VIP_GO_AUTH_COOKIE_KEY` and `VIP_GO_AUTH_COOKIE_IV` constants for the site
3. Enable encryption by setting `Vary_Cache::set_encryption( true );` before the register_groups call
4. Ensure that the segmentation functionality works as expected

### DIFF
--- a/cache/class-vary-cache.php
+++ b/cache/class-vary-cache.php
@@ -211,14 +211,17 @@ class Vary_Cache {
 	 * @access  private
 	 */
 	private static function parse_group_cookie() {
-		if ( isset( $_COOKIE[ self::COOKIE_SEGMENT ] ) ) {
-			$cookie_value = sanitize_text_field( wp_unslash( $_COOKIE[ self::COOKIE_SEGMENT ] ) );
+		if ( isset( $_COOKIE[ self::COOKIE_SEGMENT ] ) || isset( $_COOKIE[ self::COOKIE_AUTH ] ) ) {
+
 			if ( self::is_encryption_enabled() ) {
+				$cookie_value = sanitize_text_field( wp_unslash( $_COOKIE[ self::COOKIE_AUTH ] ) );
 				$cookie_value = self::decrypt_cookie_value( $cookie_value );
+			} else {
+				$cookie_value = sanitize_text_field( wp_unslash( $_COOKIE[ self::COOKIE_SEGMENT ] ) );
 			}
+
 			$groups = explode( self::GROUP_SEPARATOR, $cookie_value );
 			foreach ( $groups as $group ) {
-				// TODO: error handling (what if it's not in the right format?)?
 				list( $group_name, $group_value ) = explode( self::VALUE_SEPARATOR, $group );
 				self::$groups[ $group_name ] = $group_value ?? '';
 			}
@@ -262,8 +265,11 @@ class Vary_Cache {
 	 */
 	public static function add_vary_headers() {
 		if ( ! empty( self::$groups ) ) {
-			header( 'Vary: X-VIP-Go-Segmentation' );
-			header( 'X-VIP-Go-Segmentation-Debug: ' . self::stringify_groups() );
+			if ( self::is_encryption_enabled() ) {
+				header( 'Vary: X-VIP-Go-Auth' );
+			} else {
+				header( 'Vary: X-VIP-Go-Segmentation' );
+			}
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


cache/class-vary-cache.php
```diff
@@ -211,14 +211,17 @@ private static function decrypt_cookie_value( $cookie_value ) {
 	 * @access  private
 	 */
 	private static function parse_group_cookie() {
-		if ( isset( $_COOKIE[ self::COOKIE_SEGMENT ] ) ) {
-			$cookie_value = sanitize_text_field( wp_unslash( $_COOKIE[ self::COOKIE_SEGMENT ] ) );
+		if ( isset( $_COOKIE[ self::COOKIE_SEGMENT ] ) || isset( $_COOKIE[ self::COOKIE_AUTH ] ) ) {
+
 			if ( self::is_encryption_enabled() ) {
+				$cookie_value = sanitize_text_field( wp_unslash( $_COOKIE[ self::COOKIE_AUTH ] ) );
```
mjangda: What's the reason for using `sanitize_text_field` and `wp_unslash` here? One concern would be that they strip or modify the cookie value in a way that breaks things so we should just make sure that they're actually needed. It might be better to sanitize further down when we actually parse the group information.

mjangda:
> 2\. Enable encryption by setting `Vary_Cache::set_encryption( true );`

We also need to set the encryption constants (`VIP_GO_AUTH_COOKIE_KEY` and `VIP_GO_AUTH_COOKIE_IV`)

cache/class-vary-cache.php
dchymko: Yeah, this might be overkill and you are right about the wp_unslash, especially since the backlash is one the characters it can contain (I had actually removed it at one point during testing)

cache/class-vary-cache.php
```diff
@@ -211,14 +212,17 @@ private static function decrypt_cookie_value( $cookie_value ) {
 	 * @access  private
 	 */
 	private static function parse_group_cookie() {
-		if ( isset( $_COOKIE[ self::COOKIE_SEGMENT ] ) ) {
-			$cookie_value = sanitize_text_field( wp_unslash( $_COOKIE[ self::COOKIE_SEGMENT ] ) );
+		if ( isset( $_COOKIE[ self::COOKIE_SEGMENT ] ) || isset( $_COOKIE[ self::COOKIE_AUTH ] ) ) {
+
 			if ( self::is_encryption_enabled() ) {
+				$cookie_value = str_replace( self::VERSION_PREFIX, '', $_COOKIE[ self::COOKIE_AUTH ] );
```
mjangda: This line should be moved after the decryption. Varnish isn't aware of it so the version will need to be part of the encrypted content.

cache/class-vary-cache.php
```diff
@@ -211,14 +212,17 @@ private static function decrypt_cookie_value( $cookie_value ) {
 	 * @access  private
 	 */
 	private static function parse_group_cookie() {
-		if ( isset( $_COOKIE[ self::COOKIE_SEGMENT ] ) ) {
-			$cookie_value = sanitize_text_field( wp_unslash( $_COOKIE[ self::COOKIE_SEGMENT ] ) );
+		if ( isset( $_COOKIE[ self::COOKIE_SEGMENT ] ) || isset( $_COOKIE[ self::COOKIE_AUTH ] ) ) {
```
mjangda: Edge case but we could run into potential complications here where a user has an auth cookie but encryption not enabled (also vice versa). It would be better to more explicitly match cases:

```
if ( self::is_encryption_enabled() && ! empty( $_COOKIE[ self::COOKIE_AUTH ] ) ) {
    $cookie_value = self::decrypt_cookie_value( $_COOKIE[ self::COOKIE_AUTH ] );
} elseif ( ! empty( $_COOKIE[ self::COOKIE_SEGMENT ] ) ) {
    $cookie_value = $_COOKIE[ self::COOKIE_SEGMENT ];
}

if ( empty( $cookie_value ) {
    return;
}

// strip version
// parse
```

cache/class-vary-cache.php
dchymko: Yes, this was an untested WIP :)

cache/class-vary-cache.php
dchymko: This occurred to me when I was writing the original logic but this is much simpler check than what I had in mind.

mjangda:
r131740-deploy

PR #1135, opened 2019-02-19
Vary Cache: Set group separator to be more unique 

### LINKED ISSUE #1129: Vary Cache: Make GROUP_SEPARATOR more unique
Currently, too generic (`__`) and would be better to be more unique (e.g. `-__-` or `--__`)

dchymko:
## Description
Set separator value to be more Unique and also make sure that the seprator values aren't used in the group name or segment 

Fixes #1129 

## Checklist

### DIFF
--- a/cache/class-vary-cache.php
+++ b/cache/class-vary-cache.php
@@ -59,21 +59,25 @@ class Vary_Cache {
 	 * @access  public
 	 *
 	 * @param  array $groups  One or more groups to vary on.
+	 * @return WP_error|boolean
 	 */
 	public static function register_groups( $groups ) {
 		if ( is_array( $groups ) ) {
 			foreach ( $groups as $group ) {
 				if ( strpos( $group, self::GROUP_SEPARATOR ) !== false || strpos( $group, self::VALUE_SEPARATOR ) !== false ) {
-					trigger_error( sprintf( 'Failed to register group; cannot use the delimiter values (`%s` or `%s`) in the group name', self::GROUP_SEPARATOR, self::VALUE_SEPARATOR ), E_USER_WARNING );
-					return;
+					return new WP_Error( 'invalid_vary_group_name', sprintf( 'Failed to register group; cannot use the delimiter values (`%s` or `%s`) in the group name', self::GROUP_SEPARATOR, self::VALUE_SEPARATOR ) );
 				}
 				self::$groups[ $group ] = '';
 			}
 		} else {
+			if ( strpos( $groups, self::GROUP_SEPARATOR ) !== false || strpos( $groups, self::VALUE_SEPARATOR ) !== false ) {
+				return new WP_Error( 'invalid_vary_group_name', sprintf( 'Failed to register group; cannot use the delimiter values (`%s` or `%s`) in the group name', self::GROUP_SEPARATOR, self::VALUE_SEPARATOR ) );
+			}
 			self::$groups[ $groups ] = '';
 		}
 
 		self::parse_group_cookie();
+		return true;
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


cache/class-vary-cache.php
```diff
@@ -86,6 +90,11 @@ public static function set_group_for_user( $group, $value ) {
 		// TODO: only send header if we added or changed things
 		// TODO: don't set the cookie if was already set on the request
 		// validate, process $group, etc.
+		if ( strpos( $group, self::GROUP_SEPARATOR ) !== false || strpos( $group, self::VALUE_SEPARATOR ) !== false ||
+			strpos( $value, self::GROUP_SEPARATOR ) !== false || strpos( $value, self::VALUE_SEPARATOR ) !== false ) {
+				trigger_error( 'Cannot use the group or value separator text in the group/value name', E_USER_WARNING );
+			return;
```
mjangda: Should the function return a boolean or a `WP_Error` object now that is has a success/failure result?

cache/class-vary-cache.php
```diff
@@ -86,6 +90,11 @@ public static function set_group_for_user( $group, $value ) {
 		// TODO: only send header if we added or changed things
 		// TODO: don't set the cookie if was already set on the request
 		// validate, process $group, etc.
+		if ( strpos( $group, self::GROUP_SEPARATOR ) !== false || strpos( $group, self::VALUE_SEPARATOR ) !== false ||
+			strpos( $value, self::GROUP_SEPARATOR ) !== false || strpos( $value, self::VALUE_SEPARATOR ) !== false ) {
```
mjangda: It would be good to split out the validation for group vs value so we can provide a more specific error message.

cache/class-vary-cache.php
```diff
@@ -63,6 +63,10 @@ public static function remove_no_cache_for_user() {
 	public static function register_groups( $groups ) {
 		if ( is_array( $groups ) ) {
 			foreach ( $groups as $group ) {
+				if ( strpos( $group, self::GROUP_SEPARATOR ) !== false || strpos( $group, self::VALUE_SEPARATOR ) !== false ) {
+					trigger_error( 'Cannot use the group separator text in the group name', E_USER_WARNING );
```
mjangda: The error message should ideally include the separator:

```suggestion
					trigger_error( sprintf( 'Failed to register group; cannot use the delimiter values (`%s` or `%s`) in the group name', self::GROUP_SEPARATOR, self::VALUE_SEPARATOR ), E_USER_WARNING );
```

cache/class-vary-cache.php
```diff
@@ -86,6 +90,11 @@ public static function set_group_for_user( $group, $value ) {
 		// TODO: only send header if we added or changed things
 		// TODO: don't set the cookie if was already set on the request
 		// validate, process $group, etc.
+		if ( strpos( $group, self::GROUP_SEPARATOR ) !== false || strpos( $group, self::VALUE_SEPARATOR ) !== false ||
+			strpos( $value, self::GROUP_SEPARATOR ) !== false || strpos( $value, self::VALUE_SEPARATOR ) !== false ) {
+				trigger_error( 'Cannot use the group or value separator text in the group/value name', E_USER_WARNING );
```
mjangda: Same as earlier, the error message should be a bit more explicit about what happened and why:

> Failed to add user to group; cannot use the group delimiter value (`%s`) in the group name...

cache/class-vary-cache.php
dchymko: Good call, I made the change to register_groups as well

dchymko:
I added tests here, although the headers already sent error was causing the valid `set_group_for_user()` test to fail so I marked it as skipped until we refactor the cookie setting logic.

cache/class-vary-cache.php
```diff
@@ -90,19 +98,26 @@ public static function clear_groups() {
...
 	 * @param  string $value A value for the group.
+	 * @return WP_Error|boolean
 	 */
 	public static function set_group_for_user( $group, $value ) {
 		// TODO: make sure headers aren't already sent
 		// TODO: only send header if we added or changed things
 		// TODO: don't set the cookie if was already set on the request
-		// validate, process $group, etc.
+		if ( strpos( $group, self::GROUP_SEPARATOR ) !== false || strpos( $group, self::VALUE_SEPARATOR ) !== false ) {
+			return new WP_Error( 'invalid_vary_group_name', sprintf( 'Failed to register group; cannot use the delimiter values (`%s` or `%s`) in the group name', self::GROUP_SEPARATOR, self::VALUE_SEPARATOR ) );
```
mjangda: I think it's worth still triggering a warning via `trigger_error` as that's a good signal that something is wrong.

mjangda: Actually, looking at anticipated usage, `register_groups` is more of a fire-and-forget function (there's not much happening) so `trigger_error` is probably better there.

Whereas `set_group_for_user` should probably require handling the result (and just the WP_Error should be good enough).

mjangda:
r131747-deploy

PR #1138, opened 2019-02-20
Fix group check logic to return false for groups with empty segments

dchymko:
## Description

Fixes an issue where checking if the user is in a group will return true if user hasn't been assigned a segment in that group yet.

We'll have to be clear in the documentation that the user must set a non-empty value for the segment for the user to be considered in the group.

## Checklist

### DIFF
--- a/cache/class-vary-cache.php
+++ b/cache/class-vary-cache.php
@@ -114,9 +114,8 @@ class Vary_Cache {
 	 */
 	public static function is_user_in_group( $group ) {
 		self::parse_group_cookie();
-
 		// The group isn't defined, or the user isn't in it.
-		if ( ! array_key_exists( $group, self::$groups ) ) {
+		if ( ! array_key_exists( $group, self::$groups ) || null == self::$groups[ $group ] ) {
 			return false;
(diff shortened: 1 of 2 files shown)


cache/class-vary-cache.php
```diff
@@ -114,9 +114,8 @@ public static function set_group_for_user( $group, $value ) {
 	 */
 	public static function is_user_in_group( $group ) {
 		self::parse_group_cookie();
-
 		// The group isn't defined, or the user isn't in it.
-		if ( ! array_key_exists( $group, self::$groups ) ) {
+		if ( ! array_key_exists( $group, self::$groups ) || null == self::$groups[ $group ] ) {
```
mjangda: Should probably use a strict check here:

```suggestion
		if ( ! array_key_exists( $group, self::$groups ) || null === self::$groups[ $group ] ) {
```

dchymko: You're too fast! I changed it to a trim() to catch the empty string as well.

mjangda:
r131742-deploy

PR #1142, opened 2019-02-21
Vary Cache: Fix version prefix

mjangda:
Should only set it when stringifying groups. Previously we were appending it to the value which meant that encrypted cookies and cb cookies were getting incorrect values set.

## Checklist

## Steps to Test

1. Using the GDPR example, verify that all things work.
1. Enable encryption for the GDPR example; make sure that the cookie value does not have the version prefix visible.
1. Call `Vary_Cache::set_no_cache_for_user()` on a page; verify that the `vip-go-cb` cookie is set with a value of `1`.


mjangda:
r131962-deploy

PR #1143, opened 2019-02-21
Vary Cache: Group WP hooks / callbacks together

### LINKED ISSUE #1132: Vary Cache: group WordPress actions/filters
Currently, we call `add_action( 'send_headers', ... )` in a global context. It would be nice to group this and any future WP hooks together (either in a new class or a within `Vary_Cache`).

mjangda:
- Moves the `send_headers` call into the class alongside other WP hooks
- Rather than re-parsing the cookie every time a method is called, which is a bit of a waste, parse it once on init.

Fixes #1132
See #1122

## Checklist

## Steps to Test

With the GDPR example, make sure that the 3 different states work as expected and get correctly cached values on multiple refreshes.

### DIFF
--- a/cache/class-vary-cache.php
+++ b/cache/class-vary-cache.php
@@ -48,9 +48,29 @@ class Vary_Cache {
 
 	/** Clears the cache-busting flag */
 	public static function remove_no_cache_for_user() {
-		if ( isset( $_COOKIE[   self::COOKIE_NO_CACHE ] ) ) {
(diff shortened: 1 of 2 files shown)


mjangda:
Tests are failing because we don't parse the cookie any more between tests.

dchymko:
LGTM. Tested with the most recent GDPR example.

dchymko:
r132024-deploy

Issue #1123, opened 2019-02-15
Vary Cache: Validation and debugging improvements


dchymko:
We should add validation to ensure that invalid use is handled or rejected

- [x] Validate that the group name and value are limited to alphanumerics, '-' and '_' (or escape them)
- [ ] Add tests on the public methods
- [x] Add debug mode which would include a custom header and also output in the footer source to assist with developer troubleshooting

mjangda:
> Add tests on the public methods

There are some private / protected methods that we may want to add tests for as well.

mjangda:
Going to close this out; we'll add more tests over time.

PR #1144, opened 2019-02-21
Vary Cache: Send a debug header if WP_DEBUG enabled

mjangda:
To simplify debugging, it can be helpful to see which groups we're varying on.

#1123

## Checklist

## Steps to Test

1. Enable WP_DEBUG.
1. Go to the GDPR example. Verify that requests show the `X-VIP-Go_Segementation_Debug` response header.
1. Turn off WP_DEBUG; verify that requests no longer have the header.

### DIFF
--- a/cache/class-vary-cache.php
+++ b/cache/class-vary-cache.php
@@ -322,6 +322,10 @@ class Vary_Cache {
 			} else {
 				header( 'Vary: X-VIP-Go-Segmentation' );
 			}
+
+			if ( defined( 'WP_DEBUG' ) && true === WP_DEBUG ) {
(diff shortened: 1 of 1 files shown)


dchymko:
LGTM. Do we maybe want to raise a warning/error/alert if this is on for a production site and set_encryption is true? (To avoid accidental leaking of secrets if it gets left on there)

mjangda:
> Do we maybe want to raise a warning/error/alert if this is on for a production site and set_encryption is true

I think we'd have bigger problems if `WP_DEBUG` is enabled on a production site :)

mjangda:
r131961-deploy

PR #1145, opened 2019-02-21
Vary Cache: Validate auth keys and cookie values

### LINKED ISSUE #1131: Vary Cache: improve encryption handling
- [x] Right now, we only check if they are defined. Should confirm their value before attempting to use them otherwise throw a warning.
- [x] Rename `set_encryption`  to `enable_encryption` (should just be a one-way function)
- [x] Move checks to `enable_encryption`
- [x] <strike>Set a random value for the key and IV if not properly defined (to avoid leaking sensitive cookie data).</strike> Throw fatal if key/iv are note properly set up.

dchymko:
## Description

Adds validation for
- IV/Key set if using the encrypted approach
- Cookie values only being alphanumeric, dash or underscore
- Cookie values not containing separator (refactored into single validation method)

Fixes #1131 

## Checklist

### DIFF
--- a/cache/class-vary-cache.php
+++ b/cache/class-vary-cache.php
@@ -65,16 +65,18 @@ class Vary_Cache {
 	public static function register_groups( $groups ) {
 		if ( is_array( $groups ) ) {
 			foreach ( $groups as $group ) {
-				if ( strpos( $group, self::GROUP_SEPARATOR ) !== false || strpos( $group, self::VALUE_SEPARATOR ) !== false ) {
-					trigger_error( sprintf( 'Failed to register group; cannot use the delimiter values (`%s` or `%s`) in the group name', self::GROUP_SEPARATOR, self::VALUE_SEPARATOR ), E_USER_WARNING );
+				$validate_group = self::validate_cookie_values( $group );
+				if ( true !== $validate_group ) {
+					trigger_error( sprintf( 'Failed to register group; ', $validate_group ), E_USER_WARNING );
 					return false;
 				}
 
 				self::$groups[ $group ] = '';
 			}
 		} else {
-			if ( strpos( $groups, self::GROUP_SEPARATOR ) !== false || strpos( $groups, self::VALUE_SEPARATOR ) !== false ) {
-				trigger_error( sprintf( 'Failed to register group; cannot use the delimiter values (`%s` or `%s`) in the group name', self::GROUP_SEPARATOR, self::VALUE_SEPARATOR ), E_USER_WARNING );
+			$validate_group = self::validate_cookie_values( $groups );
+			if ( true !== $validate_group ) {
+				trigger_error( sprintf( 'Failed to register group; ', $validate_group ), E_USER_WARNING );
 				return false;
 			}
 
@@ -109,11 +111,13 @@ class Vary_Cache {
 		// TODO: make sure headers aren't already sent
 		// TODO: only send header if we added or changed things
 		// TODO: don't set the cookie if was already set on the request
-		if ( strpos( $group, self::GROUP_SEPARATOR ) !== false || strpos( $group, self::VALUE_SEPARATOR ) !== false ) {
-			return new WP_Error( 'invalid_vary_group_name', sprintf( 'Failed to register group; cannot use the delimiter values (`%s` or `%s`) in the group name', self::GROUP_SEPARATOR, self::VALUE_SEPARATOR ) );
+		$validate_group = self::validate_cookie_values( $group );
+		if ( true !== $validate_group ) {
+			return new WP_Error( 'invalid_vary_group_name', sprintf( 'Failed to register group; ', $validate_group ) );
 		}
-		if ( strpos( $value, self::GROUP_SEPARATOR ) !== false || strpos( $value, self::VALUE_SEPARATOR ) !== false ) {
-			return new WP_Error( 'invalid_vary_group_segment', sprintf( 'Failed to register group; cannot use the delimiter values (`%s` or `%s`) in the group segment', self::GROUP_SEPARATOR, self::VALUE_SEPARATOR ) );
+		$validate_value = self::validate_cookie_values( $value );
+		if ( true !== $validate_value ) {
+			return new WP_Error( 'invalid_vary_group_segment', sprintf( 'Failed to register group segment; ', $validate_group ) );
 		}
 		self::$groups[ $group ] = $value;
 		if ( self::is_encryption_enabled() ) {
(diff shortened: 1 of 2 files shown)


mjangda:
@dchymko I had added a few more suggested improvements in #1131 -- would be good to include those as well (namely, renaming `set_encryption` and adding a fallback for the key/iv).

mjangda:
There's a change to the `vip-support` submodule which seems unintentional?

cache/class-vary-cache.php
```diff
@@ -344,6 +341,21 @@ private static function set_cookie( $name, $value ) {
...
 	}
+	/**
+	 * Only allow alphanumerics, dash and underscore
+	 *
+	 * @param string $value The string you want to test on.
+	 * @return boolean
+	 */
+	private static function validate_cookie_values( $value ) {
+		if ( preg_match( '/[^_0-9a-zA-Z-]+/', $value ) > 0 ) {
+			return 'Invalid character(s). Can only use alphanumerics, dash and underscore';
```
mjangda: Should probably return a `WP_Error` object here rather than a string?

Also, tests?

dchymko:
I guess we may also want to validate for empty ('') values?

cache/class-vary-cache.php
dchymko: You're right, I was thinking it should be up to the calling function in how it wants to deal with the errors. But we can do:
```
if ( is_wp_error( $value) ) {
  trigger_error(...
```
in the calling functions instead.

It's tested indirectly from the tests I added in the register_groups() but I should probably add explicit tests too.

dchymko: Done

cache/class-vary-cache.php
```diff
@@ -64,8 +64,9 @@ public static function remove_no_cache_for_user() {
 	 */
 	public static function register_groups( array $groups ) {
 		foreach ( $groups as $group ) {
-			if ( strpos( $group, self::GROUP_SEPARATOR ) !== false || strpos( $group, self::VALUE_SEPARATOR ) !== false ) {
-				trigger_error( sprintf( 'Failed to register group (%s); cannot use the delimiter values (`%s` or `%s`) in the group name', $group, self::GROUP_SEPARATOR, self::VALUE_SEPARATOR ), E_USER_WARNING );
+			$validate_result = self::validate_cookie_values( $group );
+			if ( is_wp_error( $validate_result ) ) {
+				trigger_error( sprintf( 'Failed to register group (%s) ; ', $group, $validate_result->get_error_message() ), E_USER_WARNING );
```
mjangda:
```suggestion
				trigger_error( sprintf( 'Failed to register group (%s): %s', $group, $validate_result->get_error_message() ), E_USER_WARNING );
```

cache/class-vary-cache.php
```diff
@@ -116,11 +117,13 @@ public static function set_group_for_user( $group, $value ) {
...
-			return new WP_Error( 'invalid_vary_group_name', sprintf( 'Failed to register group; cannot use the delimiter values (`%s` or `%s`) in the group name', self::GROUP_SEPARATOR, self::VALUE_SEPARATOR ) );
+		$validate_group_result = self::validate_cookie_values( $group );
+		if ( is_wp_error( $validate_group_result ) ) {
+			return new WP_Error( 'invalid_vary_group_name', sprintf( 'Failed to register group (%s): %s', $group, $validate_group_result->get_error_message() ) );
 		}
-		if ( strpos( $value, self::GROUP_SEPARATOR ) !== false || strpos( $value, self::VALUE_SEPARATOR ) !== false ) {
-			return new WP_Error( 'invalid_vary_group_segment', sprintf( 'Failed to register group; cannot use the delimiter values (`%s` or `%s`) in the group segment', self::GROUP_SEPARATOR, self::VALUE_SEPARATOR ) );
+		$validate_value_result = self::validate_cookie_values( $value );
+		if ( is_wp_error( $validate_value_result ) ) {
+			return new WP_Error( 'invalid_vary_group_segment', sprintf( 'Failed to register group segment (%s); %s ', $group, $validate_value_result->get_error_message() ) );
```
mjangda:
```suggestion
			return new WP_Error( 'invalid_vary_group_segment', sprintf( 'Failed to register group segment (%s): %s', $group, $validate_value_result->get_error_message() ) );
```

cache/class-vary-cache.php
```diff
@@ -188,10 +191,17 @@ public static function get_groups() {
...
+	 * @return WP_Error|null
 	 */
-	public static function set_encryption( $value = true ) {
-		static::$encryption_enabled = $value;
+	public static function enable_encryption() {
+
+		// Validate that we have the secret values.
+		if ( ( ! defined( 'VIP_GO_AUTH_COOKIE_KEY' ) || ! defined( 'VIP_GO_AUTH_COOKIE_IV' ) ||
+			empty( constant( 'VIP_GO_AUTH_COOKIE_KEY' ) ) || empty( constant( 'VIP_GO_AUTH_COOKIE_IV' ) ) ) ) {
+			return new WP_Error( 'vary-cache-secrets-not-defined', sprintf( 'Constants not defined for encrypted vary cache cookies (%s and %s)', 'VIP_GO_AUTH_COOKIE_KEY', 'VIP_GO_AUTH_COOKIE_IV' ) );
```
mjangda: Let's fatal here as discussed.

cache/class-vary-cache.php
```diff
@@ -348,13 +344,28 @@ private static function set_cookie( $name, $value ) {
...
 	private static function unset_cookie( $name ) {
 		setcookie( $name, '', time() - 3600 );
 	}
+	/**
+	 * Only allow alphanumerics, dash and underscore
+	 *
+	 * @param string $value The string you want to test on.
+	 * @return WP_Error|boolean
+	 */
+	private static function validate_cookie_values( $value ) {
```
mjangda: Minor: This would be better named as a singular: `validate_cookie_value` (no `s`)

tests/cache/test-vary-cache.php
```diff
@@ -310,4 +321,69 @@ public function test__set_group_for_user_invalid( $group, $value, $expected_erro
...
+
+		$actual_result = Vary_Cache::enable_encryption( );
+
+		$this->assertWPError( $actual_result, 'Not WP_Error object' );
+		$actual_error_code = $actual_result->get_error_code();
+		$this->assertEquals( 'vary-cache-secrets-not-defined', $actual_error_code, 'Incorrect error code' );
+
+	}
+
+	public function test__enable_encryption_true_valid() {
```
mjangda: Will need to run this test in a separate process otherwise the `define` will persist across other tests.

See https://github.com/Automattic/vip-go-mu-plugins/blob/2280341588b2fe2fff26fec8b4cce984d2aa27df/tests/lib/utils/context.php#L40-L43

tests/cache/test-vary-cache.php
```diff
@@ -310,4 +321,69 @@ public function test__set_group_for_user_invalid( $group, $value, $expected_erro
 		$this->assertEquals( $expected_error_code, $actual_error_code, 'Incorrect error code' );
 	}
 
+	public function test__enable_encryption_true_invalid() {
```
mjangda: Also need a test for empty constants

tests/cache/test-vary-cache.php
dchymko: Cool, was wondering on that

dchymko:
This should be good to go. Note, I had to skip the enable_encryption error-checking tests due to an incompatibility in the annotation of Error exceptions between the local version of phpunit (6.5.13) and the Travis version (5.7). We may want to bump phpunit on Travis if possible

dchymko:
r132024-deploy

PR #1146, opened 2019-02-22
Stream Wrappers: Improve exception handling

### LINKED ISSUE #1136: Stream Wrapper: Improve exception handling
Exceptions thrown from lower level code are not being caught by the Stream Wrapper class resulting in unexpected behaviour and also making it harder to track errors related to the Stream Wrapper

hanifn:
## Description

Wrap Fileservice API requests in `try catch` blocks to catch exceptions thrown from further down stream. Currently exceptions thrown from classes used by `API_Client` aren't handled properly sometimes resulting in fatal errors.
 
Resolves #1136 

## Checklist


mjangda:
> exceptions thrown from further down stream

Do you have some examples? I don't think API_Client has many dependencies that throw exceptions. Mostly trying to understand if we need the handling everywhere or just some spots.

hanifn:
@mjangda The exceptions I've been seeing mainly come from `wp-includes/Requests/Transport/cURL.php`. I've not seen any other dependencies throwing exceptions but I think I'll spend time to go through the `API_Client` dependencies and see if we need this handling elsewhere.

hanifn:
r633-vip-go-stacks

PR #1149, opened 2019-02-22
Add skipped tests back in now that PHPUnit versions are in sync

dchymko:
## Description

Re-nable to tests that check for PHP errors now that the PHPUnit version in Travis has been bumped up

### DIFF
--- a/tests/cache/test-vary-cache.php
+++ b/tests/cache/test-vary-cache.php
@@ -251,7 +251,7 @@ class Vary_Cache_Test extends \WP_UnitTestCase {
 	 * @dataProvider get_test_data__register_groups_invalid
 	 */
 	public function test__register_groups__invalid( $invalid_groups ) {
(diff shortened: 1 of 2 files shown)


mjangda:
Do we need to update this as well? 

https://github.com/Automattic/vip-go-mu-plugins/blob/da7fcd24b773e1957687c6eb56f0b56ddd402d9e/tests/test-vip-cache-manager.php#L68

dchymko:
> Do we need to update this as well?
> 
> [vip-go-mu-plugins/tests/test-vip-cache-manager.php](https://github.com/Automattic/vip-go-mu-plugins/blob/da7fcd24b773e1957687c6eb56f0b56ddd402d9e/tests/test-vip-cache-manager.php#L68)
> 

Probably a good idea to keep things consistent. Updated. 

The new style is with the slashes although they've aliases all of the old ones (Error was new in 6.0 though which is why we had to do the bump in the first place)

PR #1150, opened 2019-02-22
Add send_headers checks

### LINKED ISSUE #1130: Vary Cache: Add `headers_sent` checks
If we attempt to use call `register_group` or `set_group_for_user` too late, we'll end up hitting a `headers already sent" error. We should those cases, throw a warning, and bail.

mjangda:
Fail when changes are attempted after send_headers.

We can't set cookies or the vary header after headers have already been sent and should warn users when they attempt to do that.

Fixes #1130

## Checklist

## Steps to Test

1. Check out PR.
1. Add `add_action( 'template_redirect', function() { Vary_Cache::register_group( 'dev-group' ); } );`
1. Verify warning is thrown.
1. Add `add_action( 'template_redirect', function() { var_dump( Vary_Cache::set_group_for_user( 'dev-group', 'yay' ) ); } );`
1. Verify WP_Error is output.

### DIFF
--- a/cache/class-vary-cache.php
+++ b/cache/class-vary-cache.php
@@ -5,7 +5,7 @@ namespace Automattic\VIP\Cache;
 use WP_Error;
 
 class Vary_Cache {
-	private const COOKIE_NO_CACHE = 'vip-go-cb';
+	private const COOKIE_NOCACHE = 'vip-go-cb';
 	private const COOKIE_SEGMENT = 'vip-go-seg';
 	private const COOKIE_AUTH = 'vip-go-auth';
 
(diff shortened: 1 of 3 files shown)


tests/cache/test-vary-cache.php
```diff
@@ -391,22 +422,58 @@ public function test__validate_cookie_values_invalid( $value, $expected_error_co
...
+	 */
+	public function test__send_vary_headers__sent_for_group_with_encryption() {
+		define( 'VIP_GO_AUTH_COOKIE_KEY', 'abc' );
+		define( 'VIP_GO_AUTH_COOKIE_IV', '123' );
+		Vary_Cache::register_group( 'dev-group' );
+		Vary_Cache::enable_encryption();
+
+		do_action( 'send_headers' );
+
+		$this->assertContains( 'Vary: X-VIP-Go-Auth', xdebug_get_headers() );
```
dchymko: :+1:

cache/class-vary-cache.php
```diff
@@ -331,10 +374,16 @@ public static function set_cookie_expiry( int $expiry ) {
...
+		self::$did_send_headers = true;
+
+		self::send_vary_headers();
+	}
+
 	/**
 	 * Add the vary cache headers to indicate that the response should be cached
 	 */
-	public static function add_vary_headers() {
+	private static function send_vary_headers() {
```
dchymko: Is there any reason we make the public method just a wrapper for the private function?

mjangda: Will be adding the cookie handling here so it made sense to have a general wrapper method.

tests/cache/test-vary-cache.php
mjangda: The xdebug method is a super weird quirk when dealing with headers in phpunit.

dchymko: yeah, it immediately jumped out at me but is cool to know. TIL.

cache/class-vary-cache.php
```diff
@@ -132,13 +173,15 @@ public static function clear_groups() {
 	 * @return WP_Error|boolean
 	 */
 	public static function set_group_for_user( $group, $value ) {
-		// TODO: make sure headers aren't already sent
-		// TODO: only send header if we added or changed things
-		// TODO: don't set the cookie if was already set on the request
+		if ( self::did_send_headers() ) {
+			return new WP_Error( 'did_send_headers', sprintf( 'Failed to set group (%s => %s) for user; cannot be called after the `send_headers` hook has fired.', $group, $value ) );
```
dchymko: This probably needs to be `trigger_error` It's not firing if you call it in the `template_redirect` action

cache/class-vary-cache.php
```diff
@@ -41,36 +50,63 @@ class Vary_Cache {
...
 
-	/** Clears the cache-busting flag */
+	/**
+	 * Clears the No-Cache cookie for the user.
+	 *
+	 * Restores caching behaviour for all future requests.
+	 */
 	public static function remove_no_cache_for_user() {
+		if ( self::did_send_headers() ) {
+			return new WP_Error( 'did_send_headers', sprintf( 'Failed to remove nocache cookie; cannot be called after the `send_headers` hook has fired.', implode( ', ', $groups ) ) );
```
dchymko: This probably needs to be `trigger_error` It's not firing if you call it in the `template_redirect` action. (Same as below)
Or we need to be clear that the developer needs to check for the error and deal with it accordingly.

mjangda: Ah, sorry, my testing instructions were flawed :)

We should follow the example of the `set_group` method here and return a WP_Error. We'll definitely want to update the usage docs.

cache/class-vary-cache.php
mjangda: See https://github.com/Automattic/vip-go-mu-plugins/pull/1150#discussion_r259562607

mjangda:
r132090-deploy

PR #1151, opened 2019-02-23
Vary Cache: Set cookies only once

### LINKED ISSUE #1133: Vary Cache: only call setcookie once
If call `set_group_for_user` multiple times in a pageload, `setcookie` will also be called multiple times. 

We should defer that a bit until `send_headers` to avoid the repetition. Will also need a fallback to run on `shutdown` in case execution ends early.

Related #1122

mjangda:
Only send the cookie once instead of calling setcookie every time a change is triggered. This also allows for easier testing and organizes the code a bit more cleanly.

Also introduces a new action `vip_vary_cache_did_send_headers` that must be used for early exits. We attempted falling back to `shutdown` for those cases, but can't reliably do so since output has already been sent in many cases by the time our shutdown callback fires.

Fixes #1133 

## Checklist

## Steps to Test

Make sure the GDPR example works (just pushed an update to work with this PR).

### DIFF
--- a/cache/class-vary-cache.php
+++ b/cache/class-vary-cache.php
@@ -33,6 +33,33 @@ class Vary_Cache {
 	private static $did_send_headers = false;
 
 	/**
+	 * Flag to indicate if we're in nocache mode.
+	 *
+	 * @since   1.0.0
+	 * @access  private
+	 * @var     bool  true if nocache enabled
+	 */
+	private static $is_user_in_nocache = false;
+
+	/**
+	 * Flag to indicate if we should update the nocache cookie.
+	 *
+	 * @since   1.0.0
+	 * @access  private
+	 * @var     bool
+	 */
+	private static $should_update_nocache_cookie = false;
+
+	/**
+	 * Flag to indicate if we should update the group/segment cookie
+	 *
+	 * @since   1.0.0
+	 * @access  private
+	 * @var     bool
+	 */
+	private static $should_update_group_cookie = false;
+
+	/**
 	 * Member variable to store the parsed group values.
 	 *
 	 * @since   1.0.0
@@ -51,16 +78,31 @@ class Vary_Cache {
 	private static $cookie_expiry = MONTH_IN_SECONDS;
 
 	/**
+	 * Check if the user is in nocache mode.
+	 *
+	 * Should only be used after the `init` hook.
+	 *
+	 * @since   1.0.0
+	 * @access  public
+	 *
+	 * @return boolean
+	 */
+	public static function is_user_in_nocache() {
+		return (bool) self::$is_user_in_nocache;
+	}
+
+	/**
 	 * Add nocache cookie for the user.
 	 *
 	 * This bypasses all requests from the VIP Cache.
 	 */
 	public static function set_nocache_for_user() {
-		if ( self::did_send_headers() ) {
+		if ( self::$did_send_headers ) {
 			return new WP_Error( 'did_send_headers', 'Failed to set nocache cookie; cannot be called after the `send_headers` hook has fired.' );
 		}
 
-		self::set_cookie( self::COOKIE_NOCACHE, 1 );
+		self::$is_user_in_nocache = true;
+		self::$should_update_nocache_cookie = true;
 
 		return true;
 	}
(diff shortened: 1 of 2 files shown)


tests/cache/test-vary-cache.php
```diff
@@ -323,13 +323,26 @@ public function get_test_data__set_group_for_user_invalid() {
...
 
-		$this->assertEquals( $expected_result, $actual_result );
+		$this->assertTrue( $actual_result, 'Return value was not true' );
+
+		$this->assertEquals( [ 'dev-group' => 'yep' ], Vary_Cache::get_groups(), 'Groups did not match expected value' );
+
+		$this->assertTrue( self::get_vary_cache_property( 'should_update_group_cookie' ), 'Did not update group cookie' );
+
+		// Verify cookie actions were taken
+		add_action( 'vip_vary_cache_did_send_headers', function( $sent_vary, $sent_cookie ) {
```
mjangda: We need an assertion that `vip_vary_cache_did_send_headers` actually triggers.

tests/cache/test-vary-cache.php
```diff
@@ -499,7 +512,17 @@ public function test__set_nocache_for_user() {
 
 		$this->assertTrue( $actual_result, 'Result was not true' );
 
-		// TODO: verify cookie was set
+		$this->assertTrue( self::get_vary_cache_property( 'is_user_in_nocache' ), 'Did not switch on nocache mode' );
+		$this->assertTrue( self::get_vary_cache_property( 'should_update_nocache_cookie' ), 'Did not update nocache cookie' );
+
+		// Verify cookie actions were taken
+		add_action( 'vip_vary_cache_did_send_headers', function( $sent_vary, $sent_cookie ) {
```
mjangda: We need an assertion that `vip_vary_cache_did_send_headers` actually triggers.

tests/cache/test-vary-cache.php
```diff
@@ -520,10 +543,20 @@ public function test__remove_nocache_for_user__did_send_headers() {
...
+		$actual_result = Vary_Cache::remove_nocache_for_user();
 
 		$this->assertTrue( $actual_result, 'Result was not true' );
 
-		// TODO: verify cookie was removed
+		$this->assertFalse( self::get_vary_cache_property( 'is_user_in_nocache' ), 'Did not switch off nocache mode' );
+		$this->assertTrue( self::get_vary_cache_property( 'should_update_nocache_cookie' ), 'Did not update nocache cookie' );
+
+		// Verify cookie actions were taken
+		add_action( 'vip_vary_cache_did_send_headers', function( $sent_vary, $sent_cookie ) {
```
mjangda: We need an assertion that `vip_vary_cache_did_send_headers` actually triggers.

cache/class-vary-cache.php
```diff
@@ -370,17 +445,104 @@ public static function set_cookie_expiry( int $expiry ) {
...
+				 * Vary or Set-Cookie header were sent.
+				 *
+				 * Can be used to handle things like early redirects after a user action and group assignment.
+				 *
+				 * @since 1.0.0
+				 *
+				 * @param boolean $sent_vary Was a Vary header sent?
+				 * @param boolean $sent_cookie Was a Set-Cookie header sent?
+				 */
+				do_action( 'vip_vary_cache_did_send_headers', $sent_vary, $sent_cookie );
```
dchymko: I've noted to add this to the docs

cache/class-vary-cache.php
```diff
@@ -370,17 +445,104 @@ public static function set_cookie_expiry( int $expiry ) {
...
+		return $sent_cookie;
+	}
+
+	/**
+	 * Sets the group/segment cookie based on the user's current groupings.
+	 *
+	 * @since   1.0.0
+	 * @access  private
+	 */
+	private static function set_group_cookie() {
```
dchymko: I feel like were having a lot of methods for the various code paths here. It's better for testing but IMO makes it a bit harder to follow (initially). Probably fine as is, but we add any more logic to it might be worth splitting into a separate class (as we had discussed)

tests/cache/test-vary-cache.php
```diff
@@ -323,13 +323,26 @@ public function get_test_data__set_group_for_user_invalid() {
...
+		$this->assertEquals( [ 'dev-group' => 'yep' ], Vary_Cache::get_groups(), 'Groups did not match expected value' );
+
+		$this->assertTrue( self::get_vary_cache_property( 'should_update_group_cookie' ), 'Did not update group cookie' );
+
+		// Verify cookie actions were taken
+		add_action( 'vip_vary_cache_did_send_headers', function( $sent_vary, $sent_cookie ) {
+			$this->assertTrue( $sent_vary, 'Vary was not sent' );
+			$this->assertTrue( $sent_cookie, 'Cookie was not sent' );
+		}, 10, 2 );
+
```
dchymko:
```suggestion
$this->assertEquals( 1, did_action( 'vip_vary_cache_did_send_headers' ) );
```

cache/class-vary-cache.php
mjangda: I actually broke it down to make the flow a little bit easier, but maybe I failed :)

Definitely would be better extracting out in the future.

tests/cache/test-vary-cache.php
mjangda: Needs to actually run after the `do_action` call (done in e1a1ace8)

mjangda:
r132273-deploy

cache/class-vary-cache.php
dchymko: Nah, I think it just varies by person and this case  is totally fine. I just find that sometimes splitting things into too many functions can be a bit hard to trace through on initial read (but might be the better structure once you are familiar/working with it)

PR #1152, opened 2019-02-25
Stream Wrapper: Add WP CLI command to update filesize in metadata

### LINKED ISSUE #1137: Stream Wrapper: WP-CLI script to update attachment file sizes
Create a new WP-CLI command to run through all file attachments and ensure that their file size are properly saved in `postmeta`.

This is to reduce `file_exists` calls to the Fileservice API.

hanifn:
## Description

Add a custom WP CLI command to update attachment metadata with it's file size to reduce calls to the Fileservice API using `file_exists` and `filesize`.

Fixes #1137 

## Checklist


wp-cli/vip-filesystem.php
```diff
@@ -0,0 +1,149 @@
...
+		$meta = wp_get_attachment_metadata( $attachment_id );
+
+		if ( ! isset( $meta['filesize'] ) ) {
+			$filesize = $this->get_filesize_from_file( $attachment_id );
+
+			if ( 0 < $filesize ) {
+				$meta['filesize'] = $filesize;
+
+				if ( ! $this->dry_run ) {
+					wp_update_attachment_metadata( $attachment_id, $meta );
```
mjangda: Might be worth logging the result to a file? Similarly above for files we can't get the filesize for.

wp-cli/vip-filesystem.php
```diff
@@ -0,0 +1,149 @@
...
+
+		$offset = 0;
+
+		WP_CLI::line( 'Updating attachment filesize metadata...' );
+
+		// Parse arguments
+		$_dry_run = WP_CLI\Utils\get_flag_value( $assoc_args, 'dry-run', true );
+		if ( 'false' === $_dry_run ) {
+			$this->dry_run = false;
+		}
```
mjangda: Should probably have a fallback here?

```suggestion
		} else {
			$this->dry_run = true;
		}
```

wp-cli/vip-filesystem.php
```diff
@@ -0,0 +1,149 @@
...
+		if ( 0 >= $attachment_count ) {
+			WP_CLI::error( 'No attachments found' );
+			WP_CLI::halt( 1 );
+		}
+
+		$this->progress = \WP_CLI\Utils\make_progress_bar(
+			'Checking ' . number_format( $attachment_count ) . ' attachments', $attachment_count );
+
+		do {
+			$sql = $wpdb->prepare( 'SELECT ID FROM ' . $wpdb->posts . ' WHERE post_type= "attachment" limit %d, %d',
```
mjangda: Minor syntax fix:

```suggestion
			$sql = $wpdb->prepare( 'SELECT ID FROM ' . $wpdb->posts . ' WHERE post_type = "attachment" LIMIT %d, %d',
```

mjangda:
Forgot to note: the script itself worked nicely on a test site 👌

wp-cli/vip-filesystem.php
```diff
@@ -0,0 +1,149 @@
...
+
+	/**
+	 * Update attachment's filesize metadata
+	 *
+	 * @param int $attachment_id
+	 */
+	private function update_attachment_filesize( $attachment_id ): void {
+		$meta = wp_get_attachment_metadata( $attachment_id );
+
+		if ( ! isset( $meta['filesize'] ) ) {
```
mjangda: Actually, spotted this:

```
Warning: Illegal string offset 'filesize' in /chroot/var/www/wp-content/mu-plugins/wp-cli/vip-filesystem.php on line 120
```

Probably need a check to verify that `wp_get_attachment_metadata` actually returned an array.

wp-cli/vip-filesystem.php
hanifn: Maybe I should just set the default value for `$dry_run` to `true` in line 12?

mjangda: Oh that's a better idea :+1:

mjangda:
@hanifn I added logging support and a few other small tweaks. Let me know if you spot any issues.

wp-cli/vip-filesystem.php
```diff
@@ -0,0 +1,177 @@
...
+	 * Update attachment's filesize metadata
+	 *
+	 * @param int $attachment_id
+	 */
+	private function update_attachment_filesize( $attachment_id ): array {
+		$this->progress->tick();
+
+		$meta = wp_get_attachment_metadata( $attachment_id );
+
+		if ( ! is_array( $meta ) ) {
```
hanifn: Would simply checking for `false` be inadequate here? Since the possible values are either `false` or the `_wp_attachment_metadata` array.

wp-cli/vip-filesystem.php
```diff
@@ -0,0 +1,177 @@
...
+	/**
+	 * Update attachments' metadata
+	 *
+	 * @param array $attachments
+	 */
+	private function update_attachments( array $attachments ): void {
+		foreach ( $attachments as $attachment ) {
+			list( $did_update, $result ) = $this->update_attachment_filesize( $attachment->ID );
+
+			fputcsv( $this->log_file, [
```
hanifn: Should we make writing to a log file optional? I'm thinking that if the user doesn't specify a log file option, `--log-to-file`?, then maybe we should just print a report to `STDOUT` after everything is done

mjangda: Yep, that seems reasonable.

wp-cli/vip-filesystem.php
mjangda: I've come across cases before where the value is neither and we end up with warnings about undefined indexes (and then corrupted data saved).

wp-cli/vip-filesystem.php
mjangda: Thoughts on deploying this and adding the `log-to-file` / stdout logic in a separate PR?

hanifn: I'm ok with this. I'm actually having 2nd thoughts about that `STDOUT` logic because the logs will include the result of every attachment update and could get really big.

hanifn:
r133208-deploy

PR #1153, opened 2019-02-25
Migrations CLI: Use cURL multi for attachements validation

vaurdan:
## Description

The `validate-attachments` command in the `vip migration` command was doing a synchronous HTTP request for each attachment, which was really slow for sites with a big amount of attachments.

I have implemented the HTTP status code validation using `curl_multi`, so it tests 10 URLs asynchronously. From my tests, I believe this change makes the process ~10x faster than the previous method using `wp_remote_get`.

Additionally, I have included two optional arguments: `--start_date` and `--end_date`, to time box the validation, which is useful when dealing with delta imports of media.

There was a problem with 301 redirections from http to https being included in the CSV, despite of the `--log-found-files` argument. This was addressing by having the cURL request follow any redirections that might exist.

There are a few breaking changes, though: the CSV file now includes an header (`url,status`), while the `$response_message` has been removed from the file - I couldn't find a good way to extract that information from the cURL request.

I'm open to any suggestions on how to improve the code.

## Checklist

## Steps to Test

1. Check out PR.
2. On the VIP Go sandbox, run `wp vip migration validate-attachments attachments.csv --start_date=2019-02-20`
3. You should be able to see all the missing (404) attachments in the `attachments.csv` file, otherwise it will be an empty file with just the header.

### DIFF
--- a/wp-cli/vip-migrations.php
+++ b/wp-cli/vip-migrations.php
@@ -241,6 +241,8 @@ class VIP_Go_Migrations_Command extends WPCOM_VIP_CLI_Command {
 					$log_request = false;
 
 					$response_code = curl_getinfo( $handle, CURLINFO_HTTP_CODE );
+					$url = curl_getinfo( $handle, CURLINFO_EFFECTIVE_URL );
+					
 					curl_multi_remove_handle( $mh, $handle );
 
 					if ( 200 === $response_code ) {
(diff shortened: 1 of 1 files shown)


vaurdan:
Worth nothing that it would be great if we could have this reviewed and approved by Friday, as there is a migration happening on Friday that would benefit a lot from this improvement.

emrikol:
This is really awesome!  I was thinking about adding this as well, since it took me a long time to test thedrive's media.

I'll test this out with them later, and see if I have any feedback, but I'm super excited!

vaurdan:
@joshbetz is there anything else I need to do to deploy this?

joshbetz:
r132633-deploy

PR #1154, opened 2019-02-26
Stats: Set default for context param for wp_handle_upload

mjangda:
Some plugins (like `enable-media-replace`) call the action manually but don't pass the correct number of params (1 instead of 2), which can result in fatals:

```
Uncaught ArgumentCountError: Too few arguments to function Automattic\VIP\Stats\handle_file_upload(), 1 passed in /var/www/wp-includes/class-wp-hook.php on line 286 and exactly 2 expected
```

Set a default to prevent those fatals.

## Checklist

## Steps to Test

1. Check out PR.
1. Install enable-media-replace plugin
1. Upload a file and then replace it.
1. Verify both work as expected.

### DIFF
--- a/stats.php
+++ b/stats.php
@@ -44,7 +44,7 @@ function track_publish_post( $new_status, $old_status ) {
 /**
  * Count uploaded files
  */
-function handle_file_upload( $upload, $context ) {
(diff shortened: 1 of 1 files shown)


mjangda:
r133951-deploy

PR #1155, opened 2019-02-26
Fix multicall to always return 429 once rate limit exceeded

dchymko:
Fixes issue with incorrect error being returned in the case of multiple calls(system.multicall) in single xmlrpc request after rate limit was hit.

## Checklist

## Steps to Test

Example:

1. Check out PR.
2. Send a payload to xmlrpc.php with 50+ login calls via system.multicall
```
<?xml version="1.0"?>
<methodCall>
<methodName>system.multicall</methodName>
<params>
  <param>
      <value><array><data>
          <value><struct>
          <member>
            <name>methodName</name>
            <value><string>wp.getUsersBlogs</string></value>
          </member>
          <member>
            <name>params</name><value><array><data>
            <value><array><data>
            <value><string>admin20</string></value>
            <value><string>password</string></value>
            </data></array></value>
            </data></array></value>
          </member>
          </struct></value>
.....repeat value block 50 times...
  </data></array></value>
  </param>
</params>
</methodCall>
```
3.  Verify that after 50 method calls, all subsequent response codes are 503  not 403)

### DIFF
--- a/security.php
+++ b/security.php
@@ -97,9 +97,13 @@ function wpcom_vip_login_limit_dont_show_login_form() {
 }
 add_action( 'login_form_login', 'wpcom_vip_login_limit_dont_show_login_form' );
 
+
 function wpcom_vip_login_limit_xmlrpc_error( $error, $user ) {
-	if ( is_wp_error( $user ) && ERROR_CODE_LOGIN_LIMIT_EXCEEDED == $user->get_error_code() )
+	static $login_limit_error;
+	if ( ( is_wp_error( $user ) && ERROR_CODE_LOGIN_LIMIT_EXCEEDED === $user->get_error_code() ) || ERROR_CODE_LOGIN_LIMIT_EXCEEDED === $login_limit_error ) {
+		$login_limit_error = ERROR_CODE_LOGIN_LIMIT_EXCEEDED;
 		return new IXR_Error( 503, $user->get_error_message() );
+	}
 
 	return $error;
 }
--- a/security.php
+++ b/security.php
@@ -105,7 +105,7 @@ function wpcom_vip_login_limit_xmlrpc_error( $error, $user ) {
 			$login_limit_error = $user;
 	}
 
-	if ( ( is_wp_error( $login_limit_error ) && ERROR_CODE_LOGIN_LIMIT_EXCEEDED === $login_limit_error->get_error_code() ) ) {
+	if ( is_wp_error( $login_limit_error ) ) {
 		return new IXR_Error( 503, $login_limit_error->get_error_message() );
 	}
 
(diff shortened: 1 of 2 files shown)


security.php
```diff
@@ -97,9 +97,13 @@ function wpcom_vip_login_limit_dont_show_login_form() {
 }
 add_action( 'login_form_login', 'wpcom_vip_login_limit_dont_show_login_form' );
 
+
 function wpcom_vip_login_limit_xmlrpc_error( $error, $user ) {
-	if ( is_wp_error( $user ) && ERROR_CODE_LOGIN_LIMIT_EXCEEDED == $user->get_error_code() )
+	static $login_limit_error;
```
nickdaugherty: Can we add some clarifying comments on why this is necessary, for the World of Tomorrow?

dchymko: Sure. Was thinking a test would also be helpful for context.

dchymko: Hopefully this comment makes sense.

security.php
```diff
@@ -97,9 +97,14 @@ function wpcom_vip_login_limit_dont_show_login_form() {
...
+	if ( ( is_wp_error( $user ) && ERROR_CODE_LOGIN_LIMIT_EXCEEDED === $user->get_error_code() ) || ERROR_CODE_LOGIN_LIMIT_EXCEEDED === $login_limit_error ) {
+		// We need to set a persistent error here, as once there is an auth error in a system.multicall, core will no longer trigger any of the rate limit filters for further login attempts in the set.
+		$login_limit_error = ERROR_CODE_LOGIN_LIMIT_EXCEEDED;
 		return new IXR_Error( 503, $user->get_error_message() );
```
mjangda: The faultString for `1+n` returns blank:

```
  <value><struct>
  <member><name>faultCode</name><value><int>503</int></value></member>
  <member><name>faultString</name><value><string>You have exceeded the login limit.  Please wait a few minutes and try again.</string></value></member>
</struct></value>
  <value><struct>
  <member><name>faultCode</name><value><int>503</int></value></member>
  <member><name>faultString</name><value><string></string></value></member>
</struct></value>
```

dchymko: Good catch. Changed the logic to persist the error across calls so that we have access to the message.

seanosh:
Happened to notice the PR - is there a reason or history behind using `503` as the response when the limit has been reached vs. `429`? Just curious...

mjangda:
> is there a reason or history behind using `503` as the response when the limit has been reached vs. `429`?

Copied over from WP.com. Seems reasonable to change it though.

seanosh:
> Copied over from WP.com. Seems reasonable to change it though.

Thanks. Was just curious. If it's `503` there then probably so for good reason :)

dchymko:
r132688-deploy

PR #1157, opened 2019-02-28
Remove unused args from hooked functions

WPprodigy:
A client-site was manually calling the `wp_login` hook during their own login logic, but failed to pass the second parameter which ended up causing a fatal error in the MU plugin.

While not passing all the args that a core WP hook has is error-some by itself, I figure if we aren't using these arguments we might as well remove them.

I also wonder if it might be worth it to go a step further in our MU plugins and check to make sure arguments exist as expected before continuing (or at least catch the fatal error)? Essentially just extra defensive measure against client code. This isn't the first time this issue has come up with a core mu-plugin filter.

### DIFF
--- a/security.php
+++ b/security.php
@@ -50,7 +50,7 @@ function wpcom_vip_login_limiter( $username ) {
 }
 add_action( 'wp_login_failed', 'wpcom_vip_login_limiter' );
 
-function wpcom_vip_login_limiter_on_success( $username, $user ) {
+function wpcom_vip_login_limiter_on_success( $username ) {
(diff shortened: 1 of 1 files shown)


mjangda:
@WPprodigy Thanks for this. Can you include some testing/verification steps for this as well?

brandon-m-skinner:
@WPprodigy Do you still want to pursue this?

WPprodigy:
Hmm, not particularly at the moment 😬

Probably needs a wider effort with more defensive programming with all our filters.

PR #1160, opened 2019-02-28
Optimize cookie output by removing empty groups

### LINKED ISSUE #1139: Vary Cache: Add tests for parse_group_cookie

### LINKED ISSUE #1159: Vary Cache: Don't send/output groups that are empty

dchymko:
## Description

To conserve resources, we probably don't need to embed empty groups in the cookie value. This removes empty groups before we generate the cookie text.

Fixes #1159 
Fixes #1139 

## Checklist

### DIFF
--- a/cache/class-vary-cache.php
+++ b/cache/class-vary-cache.php
@@ -335,7 +335,7 @@ class Vary_Cache {
 	 * @throws string If credentials ENV Variables aren't defined.
 	 * @return string encrypted version of string
 	 */
-	public static function encrypt_cookie_value( $value ) {
+	private static function encrypt_cookie_value( $value ) {
 		$client_key = constant( 'VIP_GO_AUTH_COOKIE_KEY' );
 		$client_iv = constant( 'VIP_GO_AUTH_COOKIE_IV' );
 		$cookie_value = random_bytes( 32 ) . '|' . $value . '|' . ( time() + self::$cookie_expiry );
--- a/cache/class-vary-cache.php
+++ b/cache/class-vary-cache.php
@@ -429,17 +429,14 @@ class Vary_Cache {
 	 * @return string A string representation of the groups
 	 */
 	private static function stringify_groups() {
-		$groups = array_filter(
-			self::$groups,
-			function( $value ) {
-				return '' !== $value;
+		ksort( self::$groups ); // make sure the string order is the same every time.
+		$flattened = [];
+		foreach ( self::$groups as $key => $value ) {
+			if ( '' === trim( $value ) ) {
+				continue;
 			}
-		);
-		ksort( $groups ); // make sure the string order is the same every time.
-		$flatten = function ( $key, $value ) {
-			return $key . self::VALUE_SEPARATOR . $value;
-		};
-		$flattened = array_map( $flatten, array_keys( $groups ), $groups );
+			$flattened[] = $key . self::VALUE_SEPARATOR . $value;
+		}
 
 		return self::VERSION_PREFIX . implode( self::GROUP_SEPARATOR, $flattened );
 	}
--- a/cache/class-vary-cache.php
+++ b/cache/class-vary-cache.php
@@ -442,6 +442,10 @@ class Vary_Cache {
 			$flattened[] = $key . self::VALUE_SEPARATOR . $value;
 		}
 
+		if ( empty( $flattened ) ) {
+			return;
+		}
+
 		return self::VERSION_PREFIX . implode( self::GROUP_SEPARATOR, $flattened );
 	}
 
--- a/cache/class-vary-cache.php
+++ b/cache/class-vary-cache.php
@@ -429,6 +429,10 @@ class Vary_Cache {
 	 * @return string A string representation of the groups
 	 */
 	private static function stringify_groups() {
+		if ( empty( self::$groups ) ) {
+			return;
+		}
+
 		ksort( self::$groups ); // make sure the string order is the same every time.
 		$flattened = [];
 		foreach ( self::$groups as $key => $value ) {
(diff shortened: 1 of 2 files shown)


cache/class-vary-cache.php
```diff
@@ -426,11 +426,17 @@ private static function parse_group_cookie() {
...
+			function( $value ) {
+				return '' !== $value;
+			}
+		);
+		ksort( $groups ); // make sure the string order is the same every time.
 		$flatten = function ( $key, $value ) {
 			return $key . self::VALUE_SEPARATOR . $value;
 		};
-		$flattened = array_map( $flatten, array_keys( self::$groups ), self::$groups );
+		$flattened = array_map( $flatten, array_keys( $groups ), $groups );
```
mjangda: I wonder if a `foreach` would be simpler here?

```
$flattened = [];
foreach ( self::$groups as $key => $value ) {
    if ( empty( $value ) ) {
        continue;
    }

    $flattened[] = $key . self::VALUE_SEPARATOR . $value;
}
```

dchymko:
Added tests for the parse group cookie and fixed an issue with parsing an empty group set.

tests/cache/test-vary-cache.php
```diff
@@ -580,4 +580,121 @@ public function test__remove_nocache_for_user() {
...
+		$get_stringify_groups_method = self::get_vary_cache_method( 'stringify_groups' );
+		Vary_Cache::register_groups( $groups );
+		foreach($group_values as $key => $value) {
+			Vary_Cache::set_group_for_user($key, $value);
+		}
+
+		$actual_result = $get_stringify_groups_method->invokeArgs(null, [ ] );
+
+		$this->assertEquals( $expected_result, $actual_result );
+		Vary_Cache::unload();
```
mjangda: This shouldn't be needed since we call it on `tearDown`.

```suggestion
```

tests/cache/test-vary-cache.php
```diff
@@ -580,4 +580,121 @@ public function test__remove_nocache_for_user() {
...
+	/**
+	 * @dataProvider get_test_data__parse_group_cookies
+	 * @runInSeparateProcess
+	 * @preserveGlobalState disabled
+	 */
+	public function test__parse_group_cookie_valid( $secrets, $initial_cookie, $expected_result )
+	{
+		$_COOKIE = $initial_cookie;
+		$get_parse_group_cookie_method = self::get_vary_cache_method( 'parse_group_cookie' );
+		if ( ! empty( $secrets ) ) {
```
mjangda: Not necessary to change now, but branches like this are good sign that we should probably split the test into two.

tests/cache/test-vary-cache.php
```diff
@@ -580,4 +580,121 @@ public function test__remove_nocache_for_user() {
...
+					'dev-group' => 'yes',
+				],
+				'vc-v1__dev-group_--_yes'
+			],
+			'values_for_all_empty_groups' => [
+				[
+				],
+				[
+				],
+				'vc-v1__'
```
mjangda: If the groups are empty, should we just return empty as well? And then `set_group_cookie` would need to either `unset_cookie` or just skip setting the cookie.

cache/class-vary-cache.php
dchymko: Hmm, I'm on the fence if that makes it simpler or not.
Perhaps a comment explaining the array_map options here?
I feel like there's a way to do it with an array_merge and the spread operator but that might be even more obscure.

tests/cache/test-vary-cache.php
dchymko: Agreed here.

tests/cache/test-vary-cache.php
dchymko: Perhaps. Although don't we need to set the header and cookie in any case?

mjangda: The Vary header will get set thanks to `register_groups`.

The downside to this mostly empty cookie is that we end up with two groups with the same content (users with no cookie, and users with mostly empty cookie) which reduces our cache efficacy.

cache/class-vary-cache.php
mjangda: With the foreach, we have one loop. With the array_map and array_filter we have two. 

Plus there's this: https://stackoverflow.com/questions/18144782/performance-of-foreach-array-map-with-lambda-and-array-map-with-static-function

Obviously we're looking at really tiny arrays so the performance costs with be mostly negligable but still worth optimizing where we can.

dchymko: Okay, I'm convinced :)

tests/cache/test-vary-cache.php
dchymko: Okay, I've addressed this. Setting the string to null in the case of no groups, which I _think_ is right.

cache/class-vary-cache.php
```diff
@@ -511,11 +521,16 @@ private static function set_cookies() {
 	 * @access  private
 	 */
 	private static function set_group_cookie() {
+		$group_string = self::stringify_groups();
+		if ( empty( $group_string ) ) {
+			return;
```
mjangda: Something to track in a separate issue: we may want to consider a `remove_user_from_group` and unsetting cookies when a user goes from being in a group with a cookie to being in no groups.

cache/class-vary-cache.php
```diff
@@ -335,7 +335,7 @@ public static function is_encryption_enabled() {
 	 * @throws string If credentials ENV Variables aren't defined.
 	 * @return string encrypted version of string
 	 */
-	private static function encrypt_cookie_value( $value ) {
+	public static function encrypt_cookie_value( $value ) {
```
mjangda: What's the reason for switching this to `public`?

cache/class-vary-cache.php
```diff
@@ -426,11 +429,18 @@ private static function parse_group_cookie() {
...
-			return $key . self::VALUE_SEPARATOR . $value;
-		};
-		$flattened = array_map( $flatten, array_keys( self::$groups ), self::$groups );
+		$flattened = [];
+		foreach ( self::$groups as $key => $value ) {
+			if ( '' === trim( $value ) ) {
+				continue;
+			}
+			$flattened[] = $key . self::VALUE_SEPARATOR . $value;
+		}
```
mjangda: To handle the case where I have groups registered but not actually set for a user (or set to an empty value).

```suggestion
		}

		if ( empty( $flattened ) ) {
			return;
		}
```

mjangda:
Spotted two things; otherwise this is good to ship.

cache/class-vary-cache.php
dchymko: Whoops, that was a left-over artifact from calling it to get values for the encrypted cookie tests. Flipped back

dchymko:
r132734-deploy

PR #1167, opened 2019-03-05
Update wpcom_vip_url_to_postid() to be interchangeable with url_to_postid()

WPprodigy:
## Description

I found that `wpcom_vip_url_to_postid()` couldn't be directly substituted with `url_to_postid()` when debugging a performance issue last month. The reason was that the sanity check we do doesn't account for when only a path is passed and not a full URL. Core does this same sanity check, so we shouldn't need to: https://github.com/WordPress/WordPress/blob/6fd8080e7ee7599b36d4528f72a8ced612130b8c/wp-includes/rewrite.php#L477-L483

And then I'm 90% sure the first condition is no longer needed either as we just directly use the site/home urls on VIP GO rather than the mapping logic on wpcom.

### DIFF
--- a/vip-helpers/vip-caching.php
+++ b/vip-helpers/vip-caching.php
@@ -250,15 +250,6 @@ add_action( 'transition_post_status', 'wpcom_vip_flush_get_page_by_path_cache', 
  * @return int Post ID, or 0 on failure.
  */
 function wpcom_vip_url_to_postid( $url ) {
-	// Can only run after init, since home_url() has not been filtered to the mapped domain prior to that,
-	// which will cause url_to_postid to fail
-	// @see https://vip.wordpress.com/documentation/vip-development-tips-tricks/home_url-vs-site_url/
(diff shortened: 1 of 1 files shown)


vip-helpers/vip-caching.php
```diff
@@ -250,20 +250,6 @@ function wpcom_vip_flush_get_page_by_path_cache( $new_status, $old_status, $post
  * @return int Post ID, or 0 on failure.
  */
 function wpcom_vip_url_to_postid( $url ) {
-	// Can only run after init, since home_url() has not been filtered to the mapped domain prior to that,
-	// which will cause url_to_postid to fail
-	// @see https://vip.wordpress.com/documentation/vip-development-tips-tricks/home_url-vs-site_url/
-	if ( ! did_action( 'init' ) ) {
```
mjangda: 💯

mjangda:
r133804-deploy

PR #1173, opened 2019-03-07
Stream Wrapper: Tweaks to `vip files update-filesizes` script

mjangda:
A couple of optimizations and improvements:

- Create meta if it doesn't exist
- BETWEEN instead of LIMIT (much faster when using INDEX)
- Confirmation step

And a few other small changes.

## Checklist

### DIFF
--- a/wp-cli/vip-filesystem.php
+++ b/wp-cli/vip-filesystem.php
@@ -46,8 +46,6 @@ class VIP_Files_CLI_Command extends \WPCOM_VIP_CLI_Command {
 	public function update_filesizes( $args, $assoc_args ) {
 		global $wpdb;
 
-		$offset = 0;
-
 		if ( ! defined( 'VIP_FILESYSTEM_USE_STREAM_WRAPPER' ) || true !== VIP_FILESYSTEM_USE_STREAM_WRAPPER ) {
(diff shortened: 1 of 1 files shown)


wp-cli/vip-filesystem.php
```diff
@@ -77,14 +96,21 @@ public function update_filesizes( $args, $assoc_args ) {
 			WP_CLI::halt( 1 );
 		}
 
-		$this->progress = \WP_CLI\Utils\make_progress_bar(
-			'Checking ' . number_format( $attachment_count ) . ' attachments', $attachment_count );
+		WP_CLI::confirm( sprintf( 'Should we start processing %s attachments?', number_format( $attachment_count ) ), $assoc_args );
+
+		$max_id = $wpdb->get_var( 'SELECT ID FROM ' . $wpdb->posts . ' ORDER BY ID DESC LIMIT 1' );
+
+		$end_index = $start_index + $batch_size;
```
hanifn: Are we at all concerned about how accurate the `$end_index` would be? It is very likely that the post ID will not be sequential

mjangda: With the way it's written, we just loop through all post IDs until we hit the highest one. We will probably have batches where there are 0 results or 0 attachments but that's okay.

mjangda:
r133803-deploy

PR #1175, opened 2019-03-12
Vary Cache: Mark methods that should not be directly called as 'private'

dchymko:
Some filter-related methods are public in order to be called by hooks but we aren't intended to be called directly (by custom code outside of the plugin)

This marks them as private in the phpdocs to make this more clear.

### DIFF
--- a/cache/class-vary-cache.php
+++ b/cache/class-vary-cache.php
@@ -149,11 +149,19 @@ class Vary_Cache {
 		self::$cookie_expiry = MONTH_IN_SECONDS;
 	}
 
+	/**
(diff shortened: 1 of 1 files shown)


mjangda:
I added `@access private` to a bunch more. And removed the redundant `@access private` on already private functions.

@dchymko Unless you spot any issues, we can 🚢

mjangda:
@dchymko don't forget to deploy this!

dchymko:
Whoops, I was thinking this was just a docs update that phpdocs would
build, but of course we need to deploy the changed class file.

dchymko:
r134038-deploy

PR #1183, opened 2019-03-14
Stream Wrapper: Add update meta cron job

### LINKED ISSUE #1182: Stream Wrapper: Cron event for updating filesize
We should move the logic from the `update-filesize` CLI command to a recurring cron event so we can run it globally across all environments easily.

hanifn:
## Description

Add a scheduled cron event to update attachment metadata with file sizes to improve performance of sites with Stream Wrappers enabled

Resolves #1182 

## Checklist

### DIFF
--- a/files/class-vip-filesystem.php
+++ b/files/class-vip-filesystem.php
@@ -483,14 +483,14 @@ class VIP_Filesystem {
 		$start_index = get_option( self::OPT_NEXT_FILESIZE_INDEX, 0 );
 		$end_index = $start_index + $updater->get_batch_size();
 
-		if ( $start_index > $max_id ) {
-			// This means all attachments have been processed so marking as done
-			update_option( self::OPT_ALL_FILESIZE_PROCESSED, 1 );
-
-			return;
-		}
-
 		do {
+			if ( $start_index > $max_id ) {
+				// This means all attachments have been processed so marking as done
+				update_option( self::OPT_ALL_FILESIZE_PROCESSED, 1 );
+
+				return;
+			}
+
 			$attachments = $updater->get_attachments( $start_index, $end_index );
 
 			$start_index = $end_index + 1;
--- a/files/class-vip-filesystem.php
+++ b/files/class-vip-filesystem.php
@@ -451,7 +451,7 @@ class VIP_Filesystem {
 		}
 
 		if (! wp_next_scheduled ( self::CRON_EVENT_NAME )) {
-			wp_schedule_event(time(), 'five_minutes', self::CRON_EVENT_NAME );
+			wp_schedule_event(time(), 'vip_five_minutes', self::CRON_EVENT_NAME );
 		}
 
 		add_action( self::CRON_EVENT_NAME, [ $this, 'update_attachment_meta' ] );
@@ -490,7 +490,12 @@ class VIP_Filesystem {
 			return;
 		}
 
-		$attachments = $updater->get_attachments( $start_index, $end_index );
+		do {
+			$attachments = $updater->get_attachments( $start_index, $end_index );
+
+			$start_index = $end_index + 1;
+			$end_index = $start_index + $updater->get_batch_size();
+		} while ( empty( $attachments ) );
 
 		if ( $attachments ) {
 			$updater->update_attachments( $attachments );
--- a/files/class-vip-filesystem.php
+++ b/files/class-vip-filesystem.php
@@ -27,6 +27,14 @@ class VIP_Filesystem {
 	const OPT_NEXT_FILESIZE_INDEX = 'vip_next_attachment_filesize_index';
 
 	/**
+	 * Option name for storing Max ID.
+	 *
+	 * We do not need to keep this updated as new attachments will already have file sizes
+	 * included in their meta.
+	 */
+	const OPT_MAX_POST_ID = 'vip_attachment_max_post_id';
+
+	/**
 	 * The unique identifier of this plugin.
 	 *
 	 * @since    1.0.0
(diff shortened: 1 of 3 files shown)


files/class-vip-filesystem.php
```diff
@@ -385,4 +410,77 @@ public function filter_get_attached_file( $file, $attachment_id ) {
...
+			'interval' => 300,
+			'display' => __( 'Once every 5 minutes' ),
+		];
+
+		return $schedule;
+	}
+
+	public function schedule_update_job() {
+		if ( get_option( self::OPT_ALL_FILESIZE_PROCESSED ) ) {
+			wp_clear_scheduled_hook( self::CRON_EVENT_NAME );
```
mjangda: We'll end up triggering `wp_clear_scheduled_hook` on every pageload after attachments are processed which may not be ideal. We should probably check `wp_next_scheduled` first?

```suggestion
			if ( wp_next_scheduled( self::CRON_EVENT_NAME ) ) {
				wp_clear_scheduled_hook( self::CRON_EVENT_NAME );
			}
```

files/class-vip-filesystem.php
```diff
@@ -385,4 +410,77 @@ public function filter_get_attached_file( $file, $attachment_id ) {
...
+		if ( get_option( self::OPT_ALL_FILESIZE_PROCESSED ) ) {
+			// already done. Nothing to update
+			trigger_error(
+				sprintf( 'Already completed. Exiting %s... $vip-go-streams-debug', self::CRON_EVENT_NAME ),
+				E_USER_NOTICE );
+			return;
+		}
+
+		$updater = new Meta_Updater( 1000 );
+		$max_id = $updater->get_max_id();
```
mjangda: We should probably store the max ID in an option. It will keep shifting over time and could mean that on some sites we never hit the end. Plus, new media will get the filesize on upload so we don't need to worry about them.

Basically:

```suggestion
		$max_id = (int) get_option( self::OPT_MAX_POST_ID );
		if ( ! $max_id ) {
			$max_id = $updater->get_max_id();
			update_option( self::OPT_MAX_POST_ID, $max_id, false );
		}
```

files/class-vip-filesystem.php
```diff
@@ -385,4 +410,77 @@ public function filter_get_attached_file( $file, $attachment_id ) {
...
+		$end_index = $start_index + $updater->get_batch_size();
+
+		if ( $start_index > $max_id ) {
+			// This means all attachments have been processed so marking as done
+			update_option( self::OPT_ALL_FILESIZE_PROCESSED, 1 );
+
+			return;
+		}
+
+		$attachments = $updater->get_attachments( $start_index, $end_index );
```
mjangda: One thought I had: on some sites, there will be batches where there are no attachments, in which case, we're just wasting time waiting for the next cron run. I wonder if we should move this to a `while` loop and run until we get some returned?

files/class-vip-filesystem.php
```diff
@@ -11,6 +11,21 @@ class VIP_Filesystem {
 	 */
 	const PROTOCOL = 'vip';
 
+	/**
+	 * The name of the scheduled cron event to update attachment metadata
+	 */
+	const CRON_EVENT_NAME = 'update_attachment_filesizes';
```
mjangda: Just noticed that we should prefix this with `vip_`

```suggestion
	const CRON_EVENT_NAME = 'vip_update_attachment_filesizes';
```

files/class-vip-filesystem.php
```diff
@@ -385,4 +410,77 @@ public function filter_get_attached_file( $file, $attachment_id ) {
...
+	 * Filter `cron_schedules` output
+	 *
+	 * Add a custom schedule for a 5 minute interval
+	 *
+	 * @param   array   $schedule
+	 *
+	 * @return  mixed
+	 */
+	public function filter_cron_schedules( $schedule ) {
+		$schedule['five_minutes'] = [
```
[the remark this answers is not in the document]
hanifn: Why not both?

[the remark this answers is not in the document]
mjangda: Good point :)

files/class-vip-filesystem.php
```diff
@@ -490,7 +490,12 @@ public function update_attachment_meta() {
 			return;
 		}
 
-		$attachments = $updater->get_attachments( $start_index, $end_index );
+		do {
+			$attachments = $updater->get_attachments( $start_index, $end_index );
```
mjangda: I think we'll need to move the `$start_index > $max_id` check in here otherwise we risk getting into an infinite loop.

hanifn: That's probably a good idea

files/class-vip-filesystem.php
```diff
@@ -385,4 +418,94 @@ public function filter_get_attached_file( $file, $attachment_id ) {
...
+		} while ( empty( $attachments ) );
+
+		if ( $attachments ) {
+			$updater->update_attachments( $attachments );
+		}
+
+		// All done, update next index option
+		trigger_error(
+			sprintf( 'Batch %d to %d completed. Updating options... $vip-go-streams-debug',
+				$start_index, $end_index ),
```
mjangda: Minor: Just realized, now that we have the loop, we could end up with a gap in these messages. Maybe we should have an `original_start_index` variable that we would use here instead?

hanifn: Yeah I noticed that too in the logs. I'll change it along with sending the messages to IRC

hanifn:
r134132-deploy

PR #1184, opened 2019-03-14
Add cache-nav-menus submodule

jenkoian:
## Description

I'm sure there's a good reason this hasn't been included already but added it to at least evoke the discussion.

Context, I'm looking to use this plugin on a VIP build and was surprised it wasn't including in the mu-plugins.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -40,3 +40,6 @@
 [submodule "gutenberg-ramp"]
 	path = gutenberg-ramp
 	url = https://github.com/Automattic/gutenberg-ramp.git
+[submodule "cache-nav-menus"]
+	path = cache-nav-menus
+	url = https://github.com/Automattic/cache-nav-menus.git


mjangda:
It would be better just to include this in your repo/project directly. That way you have more control over the caching behavior (e.g. if you wanted to tweak things).

jenkoian:
@mjangda thanks for the response. I've read it as 'we expect users will likely want to make changes to the plugin after installed' which surprises me a little? That's not typical of plugins right? If this is the case shouldn't we add some better extension points (hooks etc.)? I'm quite new to the world of WordPress so please excuse any naivety on my part.

joshbetz:
> I've read it as 'we expect users will likely want to make changes to the plugin after installed' which surprises me a little?

It's more that this doesn't need to be part of the platform and there are benefits of including it only on sites where it's needed. For example, if you want to make changes, you can. You can also upgrade on your schedule instead of being forced to upgrade only when we do. And if you want to use it on some sites but not others, you can do that.

PR #1187, opened 2019-03-15
Update Filesize: clear hook when we hit the max

mjangda:
Because of issues with notoptions, we can end up in a situation where the cron event is never cleared.

To avoid this, clear the hook from within the event itself rather than waiting until the next pageload.


mjangda:
Actually, this won't work. We'll just hit a forever loop of scheduling and unscheduling :/

PR #1189, opened 2019-03-16
Plugin Compat: Add fix for Contact Form 7 attachments

mjangda:
The plugin attempts to write to a `wp-content` path which will fail. These files are transient and only meant to be included as attachment, so let's just tell CF7 to put them in `/tmp/` which will work.

## Checklist

## Steps to Test

1. Check out PR.
1. Install and activate Contact Form 7.
1. Add a form with an file field.
1. Something a form with a file and verify it's attached to the outgoing email.

### DIFF
--- a/plugin-fixes.php
+++ b/plugin-fixes.php
@@ -1,15 +1,28 @@
 <?php
 /*
-Plugin Name: VIP Go Plugin Fixes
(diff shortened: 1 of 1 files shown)


mjangda:
r135496-deploy

PR #1191, opened 2019-03-19
Stream Wrapper: Don’t schedule meta update for non-prod env

### LINKED ISSUE #1190: Stream Wrapper: Don't schedule file size meta update on non-production site
We are currently scheduling the file size meta update cron event on all sites but this leads to a lot of noise on the IRC channel we are using to monitor the scheduled events. Since non-production environments can run a data sync to update the attachment metadata, it would be best if we disable scheduling and unschedule any existing events on non-production sites.

hanifn:
## Description

Only schedule the meta update event on production environments.

Also removes any existing scheduled meta update event from non-production environments.

Fixes #1190 

## Checklist

### DIFF
--- a/files/class-vip-filesystem.php
+++ b/files/class-vip-filesystem.php
@@ -442,6 +442,16 @@ class VIP_Filesystem {
 	}
 
 	public function schedule_update_job() {
+		// Only schedule meta update job on production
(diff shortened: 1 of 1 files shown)


hanifn:
r134453-deploy

PR #1192, opened 2019-03-19
Bump vip-support

mjangda:
- Don't send new email on Support User creation
- Hide personal emails
- Verify user is sucessfully found before using


mjangda:
r134451-deploy

FYI @davidsword

PR #1196, opened 2019-04-01
Check if constant defined first

mjangda:
Check that `WPCF7_UPLOADS_TMP_DIR` isn't already defined before defining it to avoid notices/warnings.

Continued from #1189


mjangda:
r135496-deploy

PR #1199, opened 2019-04-04
Cast returned max ID as int

mjangda:
If the site has no posts, we get a null and fatal because the return value is typecast as an int.


mjangda:
r135896-deploy

PR #1200, opened 2019-04-04
Move update filesizes job

hanifn:
## Description

Move the scheduled attachment filesize metadata updater into `a8c-files` so that it can run on all sites regardless of wether Stream Wrappers is active or not. This will allow us to run the job before activating the Stream Wrapper for a site and avoid the initial performance hit.

## Checklist

### DIFF
--- a/a8c-files.php
+++ b/a8c-files.php
@@ -894,16 +894,6 @@ class A8C_Files {
 	}
 
 	public function schedule_update_job() {
-		// Only schedule meta update job on production
-		if ( 'production' !== VIP_GO_ENV ) {
-			// Remove any existing meta update jobs from non-production
-			if ( wp_next_scheduled( self::CRON_EVENT_NAME ) ) {
-				wp_clear_scheduled_hook( self::CRON_EVENT_NAME );
-			}
-
-			return;
-		}
-
 		if ( get_option( self::OPT_ALL_FILESIZE_PROCESSED ) ) {
 			if ( wp_next_scheduled( self::CRON_EVENT_NAME ) ) {
 				wp_clear_scheduled_hook( self::CRON_EVENT_NAME );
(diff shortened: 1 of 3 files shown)


mjangda:
This will need a refresh as there have been lots of changes in the deploy repo (once those are stable).

mjangda:
@hanifn I've synced over the changes. Can you take a look to make sure everything seems okay?

a8c-files.php
```diff
@@ -832,6 +870,148 @@ function image_resize( $ignore, $id, $size ) {
...
+			'interval' => 180,
+			'display' => __( 'Once every 3 minutes, unlike what the slug says. Originally used to be 5 mins.' ),
+		];
+
+		return $schedule;
+	}
+
+	public function schedule_update_job() {
+		// Only schedule meta update job on production
+		if ( 'production' !== VIP_GO_ENV ) {
```
hanifn: @batmoo with the new `VIP_FILESYSTEM_SCHEDULE_FILESIZE_UPDATE` constant check I added, would this `ENV` be needed anymore? I feel like it's a better to leave that sort of details to deployment configs

mjangda: Yep, I think we can safely drop this.

I guess to for the rollout, we'd do something like this?

```
if ( ! defined( 'VIP_FILESYSTEM_SCHEDULE_FILESIZE_UPDATE' ) ) {
   if ( 'production' !== VIP_GO_ENV ) {
        define( 'VIP_FILESYSTEM_SCHEDULE_FILESIZE_UPDATE', true );
   }
}
```

hanifn: Yeap. This way we can easily deactivate or change batch sizes for individual sites should we need to

mjangda: Awesome, should be good to go after that change.

mjangda:
r136845-deploy

PR #1203, opened 2019-04-05
Production sync of update-filesize changes

mjangda:
- Fix index bug that was causing batches to be skipped
- `get_attachments` only returns attachments without filesize for faster iteration through jobs
- `update_attachments` now returns counts of the various results
- use a HEAD request and header sniffing to get the content size (few requests and faster)
- Speed up the jobs by increasing batch size and frequency
- Add more detail to output

### DIFF
--- a/files/class-vip-filesystem.php
+++ b/files/class-vip-filesystem.php
@@ -19,12 +19,12 @@ class VIP_Filesystem {
 	/**
 	 * Option name to mark all attachment filesize update completed
 	 */
-	const OPT_ALL_FILESIZE_PROCESSED = 'vip_all_attachment_filesize_processed';
+	const OPT_ALL_FILESIZE_PROCESSED = 'vip_all_attachment_filesize_processed_v2';
 
 	/**
 	 * Option name to mark next index for starting the next batch of filesize updates
 	 */
-	const OPT_NEXT_FILESIZE_INDEX = 'vip_next_attachment_filesize_index';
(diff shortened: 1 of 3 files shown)


files/class-vip-filesystem.php
```diff
@@ -494,16 +495,20 @@ public function update_attachment_meta() {
 			return;
 		}
 
-		$updater = new Meta_Updater( 1000 );
+		$batch_size = 3000;
```
hanifn: I notice that this change doesn't include the batch size modifications for specific sites

files/class-vip-filesystem.php
```diff
@@ -433,8 +433,9 @@ public function filter_cron_schedules( $schedule ) {
 			return $schedule;
 		}
 
+		// Not actually five minutes; we want it to run faster though to get through everything.
 		$schedule['vip_five_minutes'] = [
-			'interval' => 300,
+			'interval' => 180,
 			'display' => __( 'Once every 5 minutes' ),
```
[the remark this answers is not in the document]
mjangda: I don't think we can easily change the slug but can definitely change the display.

files/class-vip-filesystem.php
mjangda: Yeah, was trying to avoid referencing clients directly since this is a public repo.

hanifn: AH that makes sense

mjangda:
r136219-deploy

PR #1204, opened 2019-04-05
WordPress Importer 0.6.4

mjangda:
```
Improve PHP7 compatibility.
Fix bug that caused slashes to be stripped from imported comments.
Fix for various deprecation notices including wp_get_http() and screen_icon().
Fix for importing export files with multiline term meta data.
```

https://en-ca.wordpress.org/plugins/wordpress-importer/#developers


mjangda:
r136217-deploy

PR #1206, opened 2019-04-08
Prevent modifications of the VIP machine user

mjangda:
It should never be editable/removable.

- [x] Need more tests to account for non-administrator roles. Should also have a test for superadmin + machine user editing itself.

## Checklist

## Steps to Test

Adds several automated tests for this logic.

For manual testing:

- Make sure you've run Jetpack Start so that the machine user is created.
- Then using a super admin, admin, and other user, try to edit / delete / role change thee machine user. All of which should fail (or not be allowed at least).

Not that on multisite, there is a pathway to remove a user from a blog which is not ideal because core doesn't do the right permissions checks.

### DIFF
--- a/bin/phpunit-docker.sh
+++ b/bin/phpunit-docker.sh
@@ -5,7 +5,13 @@
 # Stop on failures: ./bin/phpunit-docker.sh --stop-on-failure
 # etc.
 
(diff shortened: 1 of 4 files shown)


mjangda:
r136840-deploy

PR #1121, opened 2019-02-13
Make feed cache create method static

gmazzap:
## Description

See #1120

## Checklist

Please make sure the items below have been covered before requesting a review:

- [x] This change works and has been tested locally (or has an appropriate fallback).
- [ ] This change works and has been tested on a Go sandbox.

Tested only locally, but the change is *very* minimal and backward compatible.

PR #1207, opened 2019-04-09
Revert "Make feed cache create method static"

mjangda:
Reverts Automattic/vip-go-mu-plugins#1121: `Fatal error: Cannot make non static method SimplePie_Cache::create() static in class VIP_Go_Feed_Cache`


mjangda:
r136224-deploy

PR #1213, opened 2019-04-11
Enable 2FA by default

joshbetz:
Enables `two-factor` by default and requires it for anyone with `edit_posts` capabilities.

### Deploy Notes

We need to deploy a fix to two-factor that lets it work with Jetpack SSO: georgestephanis/two-factor#276 / #1214

### Testing

1. Load this PR on a sandbox.
2. You may need to comment out the line that disables this for proxied users.
3. See that most of the sidebar options are gone except for Profile.
4. Enable two-factor in your profile.
5. See that sidebar options are back as expected.

### DIFF
--- a/two-factor.php
+++ b/two-factor.php
@@ -51,16 +51,16 @@ function wpcom_enable_two_factor_plugin() {
 function wpcom_vip_two_factor_filter_caps( $caps, $cap, $user_id, $args ) {
 	if ( wpcom_vip_force_two_factor() ) {
 		// Use a hard-coded list of caps that give just enough access to set up 2FA
-		$contributor_caps = [
+		$subscriber_caps = [
 			'read',
 		];
 
 		// You can edit your own user account (required to set up 2FA)
 		if ( $cap === 'edit_user' && ! empty( $args ) && $user_id === $args[ 0 ] ) {
-			$contributor_caps[] = 'edit_user';
+			$subscriber_caps[] = 'edit_user';
 		}
 
-		if ( ! in_array( $cap, $contributor_caps ) ) {
+		if ( ! in_array( $cap, $subscriber_caps, true ) ) {
 			return array( 'do_not_allow' );
 		}
 	}
--- a/two-factor.php
+++ b/two-factor.php
@@ -64,6 +64,7 @@ function wpcom_vip_two_factor_filter_caps( $caps, $cap, $user_id, $args ) {
 		// Use a hard-coded list of caps that give just enough access to set up 2FA
 		$subscriber_caps = [
 			'read',
+			'level_0',
 		];
 
 		// You can edit your own user account (required to set up 2FA)
--- a/two-factor.php
+++ b/two-factor.php
@@ -86,8 +86,23 @@ function wpcom_vip_two_factor_admin_notice() {
 	}
 
 	?>
-	<div class="error">
-		<p><a href="<?php echo esc_url( admin_url( 'profile.php' ) ); ?>">Two Factor Authentication</a> is required to publish to this site.</p>
+	<div id="vip-2fa-error" class="notice-error wrap clearfix" style="align-items: center;background: #ffffff;border-left-width:4px;border-left-style:solid;border-radius: 6px;display: flex;margin-top: 30px;padding: 30px;line-height: 2em;">
+			<div class="dashicons dashicons-warning" style="display:flex;float:left;margin-right:2rem;font-size:38px;align-items:center;margin-left:-20px;color:#ffb900;"></div>
+			<div>
+				<p style="font-weight:bold; font-size:16px;">
+					<a href="https://wpvip.com/documentation/vip-go/two-factor-authentication-on-vip-go/">Two Factor Authentication</a> is required to edit content on this site.
+				</p>
+
+				<p>For the safety and security of this site, your account access has been downgraded. Please enable two-factor authentication to restore your access.</p>
+
+				<p>
+					<a href="<?php echo esc_url( admin_url( 'profile.php' ) ); ?>" class="button button-primary">
+						Enable Two-factor Authentication
+					</a>
+
+					<a href="https://wpvip.com/documentation/vip-go/two-factor-authentication-on-vip-go/" class="button" target="_blank">Learn More</a>
+				</p> 
+			</div>
 	</div>
 	<?php
 }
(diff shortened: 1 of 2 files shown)


joshbetz:
We should just need https://github.com/Automattic/jetpack/pull/12029 to make this all work.

joshbetz:
Instead of the previous Jetpack PR, we need https://github.com/georgestephanis/two-factor/pull/276.

two-factor.php
```diff
@@ -50,54 +24,44 @@ function wpcom_vip_force_two_factor() {
 		return false;
 	}
 
-	return apply_filters( 'wpcom_vip_force_two_factor', false );
+	// We can't use current_user_can because it calls map_meta_cap which creates an infinite loop
+	$user = wp_get_current_user();
+	$caps = array_keys( $user->allcaps );
+	return apply_filters( 'wpcom_vip_force_two_factor', in_array( 'edit_posts', $caps ) );
```
joshbetz: This is a bit of a hack because we can't use `current_user_can()`. Also, almost every user has `edit_posts`, including contributors. It might make more sense to default this to on for everyone?

joshbetz:
I'm thinking we get this deployed with `wpcom_vip_force_two_factor` defaulting to false for now. So, nobody is locked out, but you can still enable 2FA or opt-in to force it on your site.

Then we announce via the lobby that you can filter that now to force 2FA for any users you want and starting some time in the future (maybe 1 month from now?) we will default it for all users that can edit posts.

mjangda:
We'll want to test this out with the following scenarios:

- [ ] Applications using the Jetpack Force 2fa plugin
- [ ] Applications using the force 2fa setting in Jetpack
- [ ] Applications manually enabling two-factor and our filter

two-factor.php
```diff
@@ -6,41 +6,15 @@
  * License: GPL version 2 or later - http://www.gnu.org/licenses/old-licenses/gpl-2.0.html
  */
 
-add_filter( 'two_factor_providers', function( $p ) {
-	if ( wpcom_vip_have_twilio_keys() ) {
-		$p['Two_Factor_SMS'] = __DIR__ . '/wpcom-vip-two-factor/sms-provider.php';
-	}
-
-	// https://github.com/georgestephanis/two-factor/issues/78
-	unset( $p[ 'Two_Factor_FIDO_U2F' ] );
```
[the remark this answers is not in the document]
mjangda: Ah, I see the issue referenced in the inline comment is fixed as noted in the commit: https://github.com/Automattic/vip-go-mu-plugins/pull/1213/commits/abef3f27bd2a90a005e02f258410839f34446228

two-factor.php
```diff
@@ -6,41 +6,15 @@
...
-	return class_exists( 'Jetpack' )
-		&& Jetpack::is_active()
-		&& Jetpack::is_module_active( 'sso' )
-		// SSO does not work for staging sites
-		&& ! Jetpack::is_staging_site();
-}
+// Custom list of providers
+require_once __DIR__ . '/wpcom-vip-two-factor/set-providers.php';
 
 function wpcom_vip_force_two_factor() {
```
mjangda: Minot: thoughts on renaming this from `wpcom_vip_force_two_factor` to `wpcom_vip_is_two_factor_forced`?

This always trips me up when reading this file; especially since the function isn't forcing anything, but rather checking a state.

two-factor.php
mjangda: I think for releasing this, we should look at updating the admin notice notifying that this will be enforced at `x` date with clear links to enabling 2fa.

mjangda: > It might make more sense to default this to on for everyone?

This would impact non-editorial users on WooCommerce and other sites (e.g. sites with subscriptions or other community features), which is probably not ideal.

mjangda: Also, maybe doing this on-the-fly every time, we should make the determination around whether to enforce 2fa on the `set_current_user` filter and store as a `static` somewhere? We could then check that rather than relying on the hack.

two-factor.php
joshbetz: Done in a831d3b

two-factor.php
joshbetz: Fixed in https://github.com/Automattic/vip-go-mu-plugins/pull/1213/commits/c66a3df50d5a96d7c7e67787ac13f796c6c798c2

joshbetz:
In https://github.com/Automattic/vip-go-mu-plugins/pull/1213/commits/1f8dc497ecdd0cfe7d7388c38cc0805144e16679 we time gate forcing for everyone with `edit_posts` after Thursday, May 23, 2019 at 18:00 UTC / 2 PM ET. Happy to change that date and time as necessary.

joshbetz:
Whatever date we decide, we should set up a couple Slack reminders :)

joshbetz:
Tested variations sites that are manually loading `two-factor` and our filter.

```
wpcom_vip_load_plugin( 'two-factor' );
add_action( 'muplugins_loaded', 'wpcom_enable_two_factor_plugin' );
```

Also, sites that load another copy of `two-factor` in the admin.

I didn't experience any issues. The sites I tested on a sandbox transparently switched to our version of `two-factor` and showed the admin notice as expected.

joshbetz:
Also tested sites with jetpack-force-2fa enabled, the require two-step setting in Jetpack, and both at the same time.

joshbetz:
@mjangda I think we should be all set here as long as we're OK with May 23 @ 1800 UTC for forcing 2FA for everyone.

two-factor.php
```diff
@@ -52,67 +27,88 @@ function wpcom_vip_force_two_factor() {
...
-		$contributor_caps = [
-			'edit_posts' => true,
-			'read' => true,
-			'level_1' => true,
-			'level_0' => true,
+function wpcom_vip_two_factor_filter_caps( $caps, $cap, $user_id, $args ) {
+	if ( wpcom_vip_is_two_factor_forced() ) {
+		// Use a hard-coded list of caps that give just enough access to set up 2FA
+		$subscriber_caps = [
+			'read',
```
mjangda: Do we need `level_0` in there just-in-case?

```suggestion
			'read',
			'level_0',
```

https://codex.wordpress.org/Roles_and_Capabilities#User_Levels

two-factor.php
```diff
@@ -52,67 +27,88 @@ function wpcom_vip_force_two_factor() {
 		return false;
 	}
 
-	return apply_filters( 'wpcom_vip_force_two_factor', false );
+	return apply_filters( 'wpcom_vip_is_two_factor_forced', false );
```
mjangda: Should we use the constant here so it auto-flips on the day? Or maybe better to have it intentionally switched over?

```suggestion
	return apply_filters( 'wpcom_vip_is_two_factor_forced', VIP_IS_AFTER_2FA_TIME_GATE );
```

two-factor.php
```diff
@@ -52,67 +27,88 @@ function wpcom_vip_force_two_factor() {
...
+		$timezone = 'UTC';
+	}
+
+	$date = new DateTime( "now", new DateTimeZone( $timezone ) );
+	$date->setTimestamp( VIP_2FA_TIME_GATE );
+	$date = $date->format( 'M d, Y \a\t g:i a T' );
+	$message = "will be required to publish to this site after {$date}.";
+
+	?>
+	<div class="error">
```
mjangda: Going to push up a change to tweak the style a bit and add more context.

two-factor.php
joshbetz: That would enable it for everyone. In `wpcom_vip_enforce_two_factor_plugin()`, we calculate `current_user_can( 'edit_posts' )` and then force two factor if that's true.

two-factor.php
mjangda: Done in dd49167

mjangda:
Only blocker now is being able to link directly to the "Two-Factor Options". Everything else is good to go.

mjangda:
To be deployed on Tuesday.

joshbetz:
r138198-deploy

PR #1214, opened 2019-04-12
Update two-factor

joshbetz:
Allows two-factor to work with Jetpack SSO

https://github.com/georgestephanis/two-factor/pull/276


nickdaugherty:
Just noting for visibility that the latest version of the 2FA plugin enforces it on API requests ([#](https://github.com/georgestephanis/two-factor/compare/4315c905b1a5d785df99b6ef9cb5248a69f81683...5ae6b1ba964e9654a27bfdf1f764c0086443adaf#diff-fa3c61990ea53ab19ab294efedcfeffa))

nickdaugherty:
Correction - it adds the _ability_ to enforce it on API requests - the filter used defaults to `false`.

mjangda:
> Correction - it adds the ability to enforce it on API requests - the filter used defaults to false.

Good catch.

I think we probably should look at enforcing it. However, we'll want to double-check that it doesn't block Jetpack and other legitimate requests. Maybe, in a separate PR, worth adding some tracking to see how many and which requests this would block?

joshbetz:
> Correction - it adds the ability to enforce it on API requests - the filter used defaults to false.

Just noticed this is enabled by default. Note the name of the filter is `two_factor_user_api_login_enable`, which defaults to false. So, API login is disabled.

I will flip it back to true before deploying and we can follow up with the testing @mjangda mentioned.

joshbetz:
Noting this is not deployed. We need to merge https://github.com/Automattic/vip-go-mu-plugins/pull/1222 first.

joshbetz:
r137564-deploy

PR #1212, opened 2019-04-11
Adds missing image sizes to attachment metadata

emrikol:
## Description

Due to the way that images are generated dynamically on VIP Go, when new image sizes are added to a site, they do not show up in the attachment metadata.  This will dynamically add missing image sizes to attachment metadata.

## Checklist

Please make sure the items below have been covered before requesting a review:

- [x] This change works and has been tested locally (or has an appropriate fallback).
- [x] This change works and has been tested on a Go sandbox.
- [ ] This change has relevant unit tests (if applicable).
- [ ] This change has relevant documentation additions / updates (if applicable).

## Steps to Test

Outline the steps to test and verify the PR here.

Example:

1. Add new image size to site's theme/plugins
1. Visit a media REST API endpoint (ex: `https://example.com/wp-json/wp/v2/media/12345`)
1. Check out PR.
1. Compare before and after of API response for new size.

PR #1216, opened 2019-04-16
Cache a local copy of generated image sizes

mjangda:
wp_get_attachment_metadata can be called a lot on image heavy pages which triggers our image processing code many many time as well.

Once we've processed the sizes we don't really need to do it again so cache in memory and re-use.

Includes a few minor small order optimizations to bail out early if we don't want to process the attachement (e.g. not an image, don't have the necessary data).

## Checklist

## Steps to Test

TODO

### DIFF
--- a/a8c-files.php
+++ b/a8c-files.php
@@ -918,6 +918,28 @@ function a8c_files_maybe_inject_image_sizes( $data, $attachment_id ) {
 		return $data;
 	}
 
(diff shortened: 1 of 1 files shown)


mjangda:
Follow-up from #1212

mjangda:
r136836-deploy and r136837

PR #1217, opened 2019-04-16
Prevent warnings when sizes key not set

mjangda:
Continued from (and deployed already with) #1216


mjangda:
r136836-deploy and r136837-deploy

Issue #1221, opened 2019-04-23
Block WP-API for users with 2FA enabled


joshbetz:
There's a new filter in `two-factor` that we can use to block the API for users with 2FA enabled. We'll need to do some testing to ensure it doesn't break Jetpack though.

https://github.com/Automattic/vip-go-mu-plugins/pull/1214#issuecomment-485670867

GaryJones:
Closed by #1222.

PR #1222, opened 2019-04-23
Ensure API logins are not blocked by two-factor

joshbetz:
We need to do some testing before blocking API logins this way.

https://github.com/Automattic/vip-go-mu-plugins/issues/1221


joshbetz:
r137564-deploy

PR #1224, opened 2019-04-24
Jetpack Connection Pilot

WPprodigy:
## Description

Monitor Jetpack connections. Send alerts when there are connection issues.

Currently will only attempt a reconnection under one scenario: if the site was previously successfully connected and the domains haven't changed.

I can squash the commits when we're ready to merge.

## Steps to Test

1. Check out PR.
1. Pull the branch into the vip-go-mu-plugins on a sandbox
1. define the constant in `wp shell` : `define( 'WPCOM_VIP_RUN_CONNECTION_PILOT', true );`
1. Run `WPCOM_VIP_Jetpack_Connection_Pilot::run_connection_pilot()` in `wp shell`
1. Check the value of the `vip_jetpack_connection_pilot_healthcheck` option if it was successfully connected.
1. Otherwise, add a `return` statement [here](https://github.com/Automattic/vip-go-mu-plugins/blob/jetpack-connection-pilot/vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php#L74) so you can see the error messages in the console.


vip-jetpack/connection-pilot/class-jetpack-connection-status-check.php
```diff
@@ -0,0 +1,200 @@
...
+
+		return 'Jetpack (re)connection attempt failed.';
+	}
+
+	/**
+	 * Check if the current site is using a default VIP GO domain name.
+	 *
+	 * @return bool True if this is a default URL.
+	 */
+	private function is_default_vip_domain() {
```
mjangda: default => placeholder.

This should ideally be a utility function from elsewhere.

vip-jetpack/connection-pilot/class-jetpack-connection-status-check.php
```diff
@@ -0,0 +1,200 @@
...
+			return false;
+		}
+
+		$current_site = get_blog_details();
+		if ( empty( $current_site->registered ) ) {
+			return false;
+		}
+
+		$time_diff = time() - strtotime( $current_site->registered );
+		if ( 2 >= ( $time_diff / 3600 ) ) {
```
mjangda: What's the thinking behind the 2 hour mark? Would be nice to have an inline comment stating that.

vip-jetpack/connection-pilot/class-jetpack-connection-status-check.php
```diff
@@ -0,0 +1,200 @@
...
+		return $this->notify_pilot( 'Jetpack is disconnected, and it appears the domain has changed.' );
+	}
+
+	/**
+	 * Connection issue handler for new sites.
+	 *
+	 * This includes either a default go-vip domain
+	 * or a recently created multi-site site.
+	 */
+	private function handle_new_sites() {
```
mjangda: Looks like this doesn't actually attempt to connect the site; is that intentional?

vip-jetpack/connection-pilot/class-jetpack-connection-status-check.php
```diff
@@ -0,0 +1,200 @@
...
+		// The connection check failed, let's handle the problem the best we can.
+		$this->handle_connection_issues();
+	}
+
+	/**
+	 * Handle connection issues. (Re)connect when possible, else add a notification.
+	 */
+	private function handle_connection_issues() {
+		if ( in_array( $this->connection_error->get_error_code(), array( 'jp-cxn-pilot-missing-constants', 'jp-cxn-pilot-development-mode' ), true ) ) {
+			return $this->notify_pilot( 'Jetpack cannot currently be connected on this site.' );
```
mjangda: It would be nice for the errors to include specific the "why" and what to do about it (otherwise we end up wasting time trying to figure that out.

vip-jetpack/connection-pilot/class-jetpack-connection-status-check.php
mjangda: It also might be better to just separate into two functions since w're repeating the `if` logic here and where it's called.

vip-jetpack/connection-pilot/class-jetpack-connection-status-check.php
```diff
@@ -0,0 +1,200 @@
...
+
+		if ( is_wp_error( $connection_attempt ) ) {
+			$this->connection_error = $connection_attempt;
+		}
+
+		return 'Jetpack (re)connection attempt failed.';
+	}
+
+	/**
+	 * Check if the current site is using a default VIP GO domain name.
```
mjangda:
```suggestion
	 * Check if the current site is using a default VIP Go domain name.
```

vip-jetpack/connection-pilot/bin/partner-provision.sh
```diff
@@ -0,0 +1,237 @@
...
+	[--allow-root] \
+	[--partner-tracking-id=1]'
+}
+
+# Note: this script should always be designed to keep wp-cli OPTIONAL
+# so that it can be run outside of the WordPress installation.
+WP_CLI_COMMAND="wp"
+WP_CLI_ARGS=""
+
+# EDIT: Added this block for VIP GO compatability
```
mjangda:
```suggestion
# EDIT: Added this block for VIP Go compatability
```

vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
```diff
@@ -0,0 +1,278 @@
...
+	 * Creates the VIP user if needed, provisions the plan with WP.com, and re-runs the connection checks to ensure it all worked.
+	 *
+	 * @param bool $skip_connection_tests Skip if we've already run the checks before this point.
+	 * @param bool $disconnect Set to true if it should disconnect Jetpack at the start.
+	 *
+	 * @return mixed bool|WP_Error True if JP was (re)connected, WP_Error otherwise.
+	 */
+	public static function connect_site( $skip_connection_tests = false, $disconnect = false ) {
+		if ( ! self::validate_constants() ) {
+			return new WP_Error( 'jp-cxn-pilot-missing-constants', 'This is not a valid VIP GO environment.' );
```
mjangda:
```suggestion
			return new WP_Error( 'jp-cxn-pilot-missing-constants', 'This is not a valid VIP Go environment or some constants are missing.' );
```

mjangda:
A couple other misc thoughts (don't necessarily have to be addressed in this PR):

- We should probably pause the Pilot when VIP Go's global maintenance mode is enabled.
- There are some sites which we'll never be able to connect due to access restrictions and we need to handle those correctly (rather than alerting all the time).
- It would be nice to add some tests to some of the new methods being added here.

rogertheriault:
Just tested, with the following results (I have added the define step to the note above)

```
$ wp shell
wp> class_exists( 'WPCOM_VIP_Jetpack_Connection_Pilot' );
bool(true)
wp> define( 'WPCOM_VIP_RUN_CONNECTION_PILOT', true );
bool(true)
wp> WPCOM_VIP_Jetpack_Connection_Pilot::run_connection_pilot();
NULL
wp> get_option( 'vip_jetpack_connection_pilot_healthcheck' );
array(3) {
  'site_url' =>
  string(26) "https://[redacted]"
  'cache_site_id' =>
  int([redacted])
  'last_healthcheck' =>
  int(1557854534)
}

... (time passes) ...

wp> WPCOM_VIP_Jetpack_Connection_Pilot::run_connection_pilot();
NULL
wp> get_option( 'vip_jetpack_connection_pilot_healthcheck' );
array(3) {
  'site_url' =>
  string(26) "https://[redacted]"
  'cache_site_id' =>
  int([redacted])
  'last_healthcheck' =>
  int(1557854678)
}
```

So I've verified this works for a connected site.

rogertheriault:
Also confirmed this for a site that is not (was not) connected:

```
wp> define( 'WPCOM_VIP_RUN_CONNECTION_PILOT', true );
bool(true)
wp> WPCOM_VIP_Jetpack_Connection_Pilot::run_connection_pilot();
array (
  'message' => 'Jetpack is disconnected, though it appears this is a new site.',
  'error' => 
  WP_Error::__set_state(array(
     'errors' => 
    array (
      'jp-cxn-pilot-not-active' => 
      array (
        0 => 'Jetpack is not currently active.',
      ),
    ),
     'error_data' => 
    array (
    ),
  )),
  'healthcheck' => false,
)
NULL
```

vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
```diff
@@ -0,0 +1,128 @@
...
+		return wpcom_vip_irc( '#vip-jp-cxn-monitoring', $message );
+	}
+
+	/**
+	 * Checks if the connection pilot should run.
+	 *
+	 * @return bool True if the connection pilot should run.
+	 */
+	private static function should_run_connection_pilot() {
+		if ( defined( 'WPCOM_VIP_RUN_CONNECTION_PILOT' ) && WPCOM_VIP_RUN_CONNECTION_PILOT ) {
```
rogertheriault: this should have an explicit true check as well

Also - if someone manually runs this (like I did) without the define being set, they may not realize it's necessary. I wonder what's the best way to detect/warn about that?

And do we need the ability to run this without scheduling an event?

WPprodigy: Ideally, there will be CLI commands with the shared functionality in the future. ATM, jetpack-start is sufficient when needing to do manual tests and fixes though. So I'm not sure if this is the right place to optimize/change.

vip-jetpack/connection-pilot/class-jetpack-connection-status-check.php
WPprodigy: Yes, this is intentional at the moment. The scope for the initial rollout was to only auto-reconnect in the one scenario. Connecting new sites was out of scope, but definitely something that I want to do, just perhaps needs a little more discussion first. That said, technically speaking, it's all ready to be done with like a single line change.

The conditional won't really need to exist inside the `handle_new_sites()` method once we are auto-connecting since both will be handled the same way. I've just left it there as it made it nice while testing to ensure I'm getting the right message.

vip-jetpack/connection-pilot/class-jetpack-connection-status-check.php
WPprodigy: The cron job is going to be run hourly, so I figured 2 hours was a safe time period. I can add some clarification around this decision there.

vip-jetpack/connection-pilot/bin/partner-provision.sh
```diff
@@ -0,0 +1,237 @@
+#!/usr/bin/env sh
+
+# Duplicate of JP core's script, except for the two changes marked below by "EDIT" comments.
```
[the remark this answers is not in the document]
WPprodigy: I did leave a note about this here: https://github.com/Automattic/vip-go-mu-plugins/blob/jetpack-connection-pilot/vip-jetpack/connection-pilot/class-jetpack-connection-controls.php#L144-L145. Hopefully the script can be removed entirely in the future.

vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
```diff
@@ -0,0 +1,278 @@
...
+		}
+
+		if ( Jetpack::is_development_mode() ) {
+			return new WP_Error( 'jp-cxn-pilot-development-mode', 'Jetpack is in development mode.' );
+		}
+
+		// The Jetpack::is_active() method just checks if there are user/blog tokens in the database.
+		if ( ! Jetpack::is_active() || ! Jetpack_Options::get_option( 'id' ) ) {
+			return new WP_Error( 'jp-cxn-pilot-not-active', 'Jetpack is not currently active.' );
```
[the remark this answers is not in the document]
WPprodigy: I've given each a specific error code so the IRC messages can be quickly tracked down to a location in the code / potentially docs in the future.

vip-jetpack/connection-pilot/class-jetpack-connection-status-check.php
WPprodigy: Every slack/irc message will also include the WP_Error that occurred when checking the connection status. So these will also help provide context.

Ideally, we'll be auto-resolving issues where possible as we start to see them occurring in the wild. If that's not possible though, then updating the messaging / adding a list of errors in the FG would be the next step.

vip-jetpack/connection-pilot/class-jetpack-connection-status-check.php
```diff
@@ -0,0 +1,199 @@
...
+	 *
+	 * @return bool True if this is a newly created site.
+	 */
+	private function is_new_multisite_site() {
+		if ( ! is_multisite() ) {
+			return false;
+		}
+
+		$current_site = get_blog_details();
+		if ( empty( $current_site->registered ) ) {
```
nickdaugherty: In what situations will `registered` be empty? Do we want to attempt a connection anyway?

vip-jetpack/connection-pilot/class-jetpack-connection-status-check.php
```diff
@@ -0,0 +1,199 @@
...
+
+		return false;
+	}
+
+	/**
+	 * Adds to the array of notifications that need to be sent out by the Pilot.
+	 *
+	 * @param string $message The first line of the notification that will be sent.
+	 */
+	private function notify_pilot( $message ) {
```
nickdaugherty: IMO it would be better to pass the error object into `notify_pilot()` explicitly, rather than using the class property. Using the single property gets messy because something else could overwrite it at any time, at which point the message and underlying error can be different things.

Better to be explicit and pass the error in as an argument at runtime.

vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
```diff
@@ -0,0 +1,278 @@
...
+	 *
+	 * @param int $user_id The VIP machine user ID.
+	 *
+	 * @return mixed bool|WP_Error True if provisioning worked, WP_Error otherwise.
+	 */
+	private static function provision_site( $user_id ) {
+		// TODO: This is a high-risk possible problem on the CLI/CRON containers. Needs testing.
+		$script_path = __DIR__ . '/bin/partner-provision.sh';
+
+		// Can use JP core's script once the changes here are made: https://github.com/Automattic/vip-go-mu-plugins/pull/986.
```
nickdaugherty: Rather than include a 3rd, slightly different copy of this script, can we use the one that currently works, `jetpack-start/bin/partner-provision.sh`, then update this as part of #986?

000-vip-init.php
```diff
@@ -92,13 +92,24 @@
...
 		'action'   => 'wpcom_vip_support_remove_user_via_cron', // Automattic\VIP\Support_User\User::CRON_ACTION
 		'callback' => array( 'Automattic\VIP\Support_User\User', 'do_cron_cleanup' ),
-	),
-) );
+	)
+);
+
+if ( defined( 'WPCOM_VIP_RUN_CONNECTION_PILOT' ) && true === WPCOM_VIP_RUN_CONNECTION_PILOT ) {
+	$internal_cron_events[] = array(
+		'schedule'  => 'hourly',
```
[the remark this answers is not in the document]
WPprodigy: Possibly. Ideally we dig into the disconnection issues to find the underlining cause. If possible, we could then try to prevent the issue in the first place.

If sites are disconnecting often, then I'm not sure if the best course of action is to try and fix every few minutes.

vip-jetpack/connection-pilot/class-jetpack-connection-status-check.php
```diff
@@ -0,0 +1,199 @@
...
+	public $pilot_notifications = array();
+
+	/**
+	 * Class constructor.
+	 * Populates the class properties and sets a fallback error.
+	 */
+	public function __construct() {
+		$this->healthcheck_option = get_option( self::HEALTHCHECK_OPTION_NAME );
+		$this->site_url           = get_site_url();
+		$this->connection_error   = new WP_Error( 'jp-cxn-pilot-default-error', 'The Jetpack connection is in an unknown state.' );
```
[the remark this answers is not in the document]
WPprodigy: At the moment, it should always be overwritten. This is more-so a safeguard so we can always safely assume `$this->connection_error` is a WP_Error without conditional checks everywhere.

vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
WPprodigy: For sure. Need to make the two adjustments - the one in the above issue, as well as changing the `partner_tracking_id` arg (I've checked in with the JP team about this and got the go-ahead).

vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
```diff
@@ -0,0 +1,278 @@
...
+		if ( is_wp_error( $response ) ) {
+			return new WP_Error( 'jp-cxn-pilot-test-fail', sprintf( 'Failed to test connection (#%s: %s)', $response->get_error_code(), $response->get_error_message() ) );
+		}
+
+		$body = wp_remote_retrieve_body( $response );
+		if ( ! $body ) {
+			return new WP_Error( 'jp-cxn-pilot-empty-body', 'Failed to test connection (empty response body).' );
+		}
+
+		$result       = json_decode( $body );
```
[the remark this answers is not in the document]
WPprodigy: Will the below checks handle this already? `isset( $result->connected )` will fail gracefully, and we'll end up with right error message.

vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
```diff
@@ -0,0 +1,278 @@
...
+			}
+		}
+
+		if ( $update_notoptions ) {
+			wp_cache_set( 'notoptions', $notoptions, 'options' );
+		}
+
+		if ( $update_alloptions ) {
+			wp_cache_set( 'alloptions', $alloptions, 'options' );
+		}
```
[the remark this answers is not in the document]
WPprodigy: Makes sense. I went this route trying to interfere with the live site as little as possible, but seems like just clearing the cache values is better overall (also simplifies logic).

vip-jetpack/connection-pilot/class-jetpack-connection-status-check.php
WPprodigy: No scenarios that I know of, just a safeguard before assuming the property is there.

I imagine if we come across scenarios where new MS sites aren't being auto-connected in the future, this would be adjusted.

vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
```diff
@@ -0,0 +1,128 @@
...
+	 * Class constructor.
+	 * Ensures the cron job is set up correctly and ready.
+	 */
+	private function __construct() {
+		if ( ! self::should_run_connection_pilot() ) {
+			return;
+		}
+
+		// Ensure the internal cron job has been added. Should already exist as an internal Cron Control job.
+		if ( ! has_action( self::CRON_ACTION ) ) {
```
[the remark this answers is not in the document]
WPprodigy: Took inspiration from the other internal cron-control event on this part: https://github.com/Automattic/vip-support/blob/00dcbc16b1a9fc54a4fe2dc9075edf6f570d7ea7/class-vip-support-user.php#L166

Cron-control should be declaring the action, so it seemed like a good check to ensure cron-control is working as needed.

vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
```diff
@@ -0,0 +1,278 @@
...
+			$connection_test = self::jetpack_is_connected();
+
+			if ( true === $connection_test ) {
+				// Abort since the site is already connected to JP, and we aren't okay with disconnecting.
+				return new WP_Error( 'jp-cxn-pilot-already-connected', 'Jetpack is already properly connected.' );
+			}
+		}
+
+		if ( $disconnect ) {
+			Jetpack::disconnect();
```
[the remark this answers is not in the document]
WPprodigy: > It deletes all the cache site content

Hmm, are you sure about this? From what I see, it mainly just removes the blog_token and user_tokens. 

I had the param there for some parity with the jetpack-start commands: https://github.com/Automattic/vip-go-mu-plugins/blob/master/jetpack-start/wp-cli-api.php#L187-L196. (Hoping to remove that module at some point in favor of this one)

vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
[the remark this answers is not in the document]
nickdaugherty: True, that should be sufficient here.

vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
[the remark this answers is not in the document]
nickdaugherty: I think you're right - I was going off of this comment:

https://github.com/Automattic/jetpack/blob/6a536225dd441269d2b0427d2e8a2b5ee46b0c71/class.jetpack.php#L3274-L3276

But tracing that function through, it looks like it only deletes _local_ stuff (doesn't attempt to clear out the cache site).

vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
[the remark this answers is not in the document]
nickdaugherty: But what happens if, in 6 months, we decide that we want something new to run on `wpcom_vip_run_jetpack_connection_pilot`, so we add:

```
add_action( 'wpcom_vip_run_jetpack_connection_pilot', 'do_stuff' )
```

in some other file? We'll unknowingly break the main pilot code because now there's already an action registered, so `has_action()` returns `true`, and we never register this action.

Not saying that's a likely scenario, but it makes this code fragile if simply hooking into the same action anywhere else will cause it to stop working.

vip-jetpack/connection-pilot/class-jetpack-connection-status-check.php
nickdaugherty: Alright, sounds good 👍

vip-jetpack/connection-pilot/class-jetpack-connection-status-check.php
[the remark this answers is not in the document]
nickdaugherty: It looks like `handle_connection_issues()` is the only place that uses it as a WP_Error - if we default to just null, it's an easy check to see if it was set or not.

vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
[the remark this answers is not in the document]
nickdaugherty: Addressed in 8491985d

vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
```diff
@@ -0,0 +1,128 @@
...
+	 *
+	 * Needs to be static due to how it is added to cron control.
+	 */
+	public static function run_connection_pilot() {
+		if ( ! self::should_run_connection_pilot() ) {
+			return;
+		}
+
+		require_once __DIR__ . '/class-jetpack-connection-controls.php';
+		require_once __DIR__ . '/class-jetpack-connection-status-check.php';
```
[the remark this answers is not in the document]
nickdaugherty: Requires moved in 222534df

vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
```diff
@@ -0,0 +1,128 @@
...
+	 *
+	 * Schedule changes can take up to 24 hours to take effect.
+	 * See the a8c_cron_control_clean_legacy_data event for more details.
+	 */
+	const CRON_SCHEDULE = 'hourly';
+
+	/**
+	 * Initiate an instance of this class if one doesn't exist already.
+	 */
+	public static function init() {
```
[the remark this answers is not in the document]
nickdaugherty: Moved the setup code into `init()` (and dropped `__construct()`) in e9bb45fe

vip-jetpack/connection-pilot/class-jetpack-connection-status-check.php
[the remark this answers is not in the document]
nickdaugherty: I went ahead and defaulted it to `null` in ae2ce2d5 - better to start off with no error set, until we actually know there is an error condition.

vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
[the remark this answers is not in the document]
nickdaugherty: 100b1ecc

vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
nickdaugherty: @WPprodigy I'm trying to sort this out - our `jetpack-start` already has a `partner-provision.sh` script (which is different than the core JP one)...this is what we currently use, is there any issue using that one instead of this version?

The first issue only applies to using the core provision script (which we're not currently)...and it's not clear what the `partner_tracking_id` one is referring to.

nickdaugherty: I think I answered my question - the script in this PR returns JSON to be parsed by PHP, where our existing one from JP start does not, doh.

nickdaugherty:
Did various refactoring today...the biggest TODO here is reducing duplication with JP Start, if we can. Lots of the core code in the Controls class is duplicated from JP Start, and the bin script is a fork of JP core's, so I'd like to see what can be made reusable. Having several different code paths for doing things like creating the machine user, testing the connection, etc will bite us.

WPprodigy:
Agreed. Was hoping to deprecate the JP Start module and move it into this. 

Didn't want to group too many changes together though, especially since this uses the new method of establishing connections that JP start does not do yet. So I wanted to make these these changes separately after the initial merge.

nickdaugherty:
r644-stacks

PR #1225, opened 2019-04-25
Switch the plugins dashboard to use wpvip.com

joshbetz:
Since we're using wpvip.com now, we need to point the plugins dashboard to use the API there instead of vip.wordpress.com.


joshbetz:
Thanks! Will get this deployed later today after the domain is live.

joshbetz:
r137843-deploy

PR #1227, opened 2019-04-30
Jetpack: Disable CSS for unused modules

joshbetz:
Jetpack ships a concatenated CSS file that includes CSS for all available modules. Since we concatenate CSS files on VIP Go, we should only include CSS for active modules, which will reduce the transfer size of CSS and eliminate unused styles.


joshbetz:
After further consideration, we're going to close this out to avoid unintended side effects. Namely, sites that unknowingly rely on styles for modules that aren't active. If we want to reconsider this at a later date, it would be helpful to automatically run a visual diff of every site with and without this code.

PR #1229, opened 2019-04-30
Update two-factor enforcement date to May 29

joshbetz:
May 23 is right before the Memorial Day weekend in the US, so we'll push this back to the follow week to avoid enforcement kicking in right as people are leaving for a long weekend.


joshbetz:
r138198-deploy

PR #1230, opened 2019-04-30
Don't force 2FA for OneLogin or SimplSaml

joshbetz:
These popular SSO solutions already provide a way for users to login
via multi-factor authentication, we don't need to add yet another
authentication mechanism to the mix.

### DIFF
--- a/two-factor.php
+++ b/two-factor.php
@@ -33,7 +33,7 @@ function wpcom_vip_should_force_two_factor() {
 	}
 
 	// Don't force 2FA for SimpleSaml
-	if ( function_exists( '\HumanMade\SimpleSaml\instance()' ) ) {
+	if ( function_exists( '\HumanMade\SimpleSaml\instance' ) && \HumanMade\SimpleSaml\instance() ) {
 		return false;
 	}
 
--- a/two-factor.php
+++ b/two-factor.php
@@ -129,6 +129,11 @@ function wpcom_vip_two_factor_prep_admin_notice() {
 	if ( wpcom_vip_is_two_factor_forced() ) {
 		return;
 	}
+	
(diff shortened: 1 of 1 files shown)


joshbetz:
Also adds a `wpcom_vip_two_factor_prep_hide_admin_notice` filter to site owners can hide the prep notice.

```
add_filter( 'wpcom_vip_is_two_factor_forced', '__return_false' );
add_filter( 'wpcom_vip_two_factor_prep_hide_admin_notice', '__return_true' );
```

I originally wanted to use the `wpcom_vip_is_two_factor_forced` filter for this purpose, but we're already using it to decide if 2FA is enabled. If we switch that to true, 2FA is forced on now, which allows site owners to opt-in to 2FA early.

two-factor.php
```diff
@@ -27,6 +27,24 @@ function wpcom_vip_is_two_factor_forced() {
 		return false;
 	}
 
+	// Don't force 2FA for OneLogin SSO
+	if ( function_exists( 'is_saml_enabled' ) && is_saml_enabled() ) {
+		return false;
+	}
+
+	// Don't force 2FA for SimpleSaml
+	if ( function_exists( '\HumanMade\SimpleSaml\instance()' ) ) {
```
mjangda: For this, is it simply enough that that method exists? Or do we need to actually check if it's enabled?

joshbetz: Good point. It looks like we need to bail if the instance function return falsey, which indicates there's not a valid SAML config.

https://github.com/wpcomvip/skyscanner/blob/940e0be2dcf2fc9891db4fbe0ae96057d827cc6a/client-mu-plugins/wp-simple-saml/inc/namespace.php#L156-L159

joshbetz:
r138198-deploy

mjangda:
👍

PR #1231, opened 2019-05-01
Don't show 2fa notice on Edit Post screen

mjangda:
As it interferes with Gutenberg.

## Steps to Test

1. Log in unproxied to an account without 2fa
1. Verify the message is shown everywhere in wp-admin except the Posts > Add New screen.
1. Enable 2fa
1. Verify the message goes away everywhere.


mjangda:
r138254-deploy

PR #1215, opened 2019-04-15
Update Rewrite Rules Inspector plugin to latest version to resolve deprecated warnings

rebeccahum:
## Description

Navigating to VIP > Rewrite Rules throws the below errors locally:

```
Notice: screen_icon is deprecated since version 3.8.0 with no alternative available. in /Users/Shared/IH/Integrated/wp-includes/functions.php on line 4332

Notice: get_screen_icon is deprecated since version 3.8.0 with no alternative available. in /Users/Shared/IH/Integrated/wp-includes/functions.php on line 4332
```

## Checklist

Please make sure the items below have been covered before requesting a review:

- [x] This change works and has been tested locally (or has an appropriate fallback).
- [x] This change works and has been tested on a Go sandbox.

## Steps to Test

1. Check out PR and ensure `WP_DEBUG` is set to `true`.
1. Go to VIP > Rewrite Rules
1. Verify that the below warnings are no longer appearing:
Screen Shot 2019-04-15 at 4 41 21 PM

PR #1211, opened 2019-04-11
Bump VIP Dashboard to introduce cap filter

mjangda:
https://github.com/Automattic/vip-dashboard/pull/89

## Checklist

Please make sure the items below have been covered before requesting a review:

- [x] This change works and has been tested locally (or has an appropriate fallback).
- [x] This change works and has been tested on a Go sandbox.
- n/a This change has relevant unit tests (if applicable).
- [ ] This change has relevant documentation additions / updates (if applicable).

## Steps to Test

- Verify that the VIP Page continues to load.
- Then, add the following and verify that only admins can access the VIP Page:

```
add_filter( 'vip_dashboard_page_cap', function( $cap ) {
    // restrict to administrators
    return 'manage_options';
} );
```

PR #1232, opened 2019-05-01
Restore accidentally revert submodules

mjangda:
Changes were accidentally reverted in #1213.

See #1215 and #1211


mjangda:
No deploy needed since they were only reverted on GitHub.

PR #1235, opened 2019-05-02
Better screen check

mjangda:
Use `is_block_editor()` method to capture all cases where the block editor is active to hide the 2fa notice

## Steps to Test

1. Check out PR.
1. Start with an account without 2fa active.
1. Verify that when the block editor is loaded, the 2fa notice is not displayed.
1. Verify that the notice shows up on other pages.


mjangda:
r138378-deploy

PR #1237, opened 2019-05-02
Add wpcom_vip_two_factor_enforcement_cap filter

joshbetz:
Adds a filter to customize the capability that's used to decide whether to enforce 2fa.


mjangda:
r138374-deploy

PR #1240, opened 2019-05-06
Two Factor: Load MS constants

joshbetz:
We load the cookie constants early. This ensures we properly define the MS cookie constants.


joshbetz:
r138669-deploy

PR #1244, opened 2019-05-07
Exif compat for Streams.

mjangda:
The iptc and exif functions don't always work with streams. So, download a local copy of the file and use that to read the exif data from that file instead.

Props S3-Uploads and humanmade for the fix (https://github.com/humanmade/S3-Uploads/blob/942eb80dd289235d6bb60b0f68ca9fecb3fe227c/inc/class-s3-uploads.php#L233-L249)

## Checklist

## Steps to Test

1. Check out PR.
1. Upload an image with a caption.
1. Verify that the caption is added to the attachment.


hanifn:
r138898-deploy

PR #1245, opened 2019-05-08
Only lock out contributors and up

mjangda:
Sites may have external registered users that have a subscriber role (with `read`) who we don't want to display the message to.

## Steps to Test

1. Check out PR.
1. Enable lockout by setting the constants.
1. Login as subscriber and verify you don't see the message.
1. Login as contributor and verify you do see the message.


mjangda:
r138931-deploy

PR #1256, opened 2019-05-15
VaultPress 1.9.10

BrookeDot:
```
1.9.10 – 4 April 2019

 * Bugfix: Fix a PHP fatal error caused by passing an object to the current() function.

1.9.9 – 28 March 2019

 * PHP 7.2.0 compatibility fix.
 * Adding button to delete all VaultPress settings
```

@scarstocea This includes the same core modification that we made in https://github.com/Automattic/vip-go-mu-plugins/pull/907 do you have an update on if this can be resolved upstream?


mjangda:
r139713-deploy

PR #1252, opened 2019-05-14
Remove executing full sync on user authorization to Jetpack

rebeccahum:
## Description

Since Jetpack executes a full sync on user connection, it overrides an ongoing full sync and thus, the ongoing full sync never ends up completing:

https://github.com/Automattic/jetpack/blame/19f9d99879d28bcdc51da94539cb26b0d1b117be/sync/class.jetpack-sync-actions.php#L529

## Checklist

Please make sure the items below have been covered before requesting a review:

- [x] This change works and has been tested locally (or has an appropriate fallback).
- [x] This change works and has been tested on a Go sandbox.
- [x] This change has relevant unit tests (if applicable).
- [x] This change has relevant documentation additions / updates (if applicable).

## Steps to Test

Unfortunately, this needs to be deployed on production, as it communicates via the Jetpack production code per Enej.

PR #1257, opened 2019-05-15
Disable do_initial_sync() for all sites now

rebeccahum:
## Description

Same as #1252 but removing checks.


nickdaugherty:
r139542-deploy

PR #1258, opened 2019-05-15
Add `two-factor report` CLI command

joshbetz:
Adds a CLI command that we can use to report adoption of 2FA on a VIP Go site.

### DIFF
--- a/wp-cli/vip-two-factor.php
+++ b/wp-cli/vip-two-factor.php
@@ -1,0 +1,24 @@
+<?php
+
+class VIP_Two_Factor_Command extends WPCOM_VIP_CLI_Command {
+	function report() {
+		$fields = array( 'ID', 'display_name' );
+		$users = get_users( array( 'fields' => $fields ) );
+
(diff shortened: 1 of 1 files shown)


wp-cli/vip-two-factor.php
```diff
@@ -0,0 +1,24 @@
...
+		$cap = apply_filters( 'wpcom_vip_two_factor_enforcement_cap', 'edit_posts' );
+		foreach ( $users as $user ) {
+			if ( Two_Factor_Core::is_user_using_two_factor( $user->ID ) ) {
+				$user->two_factor_enabled = 'true';
+			} else if ( ! user_can( $user->ID, $cap ) ) {
+				$user->two_factor_enabled = 'n/a';
+			} else {
+				$user->two_factor_enabled = 'false';
+			}
+		}
```
mjangda: Thoughts, on including which providers are in use? (can use `Two_Factor_Core ::get_enabled_providers_for_user( $user )` to do that)

joshbetz: Ah yes, good call.

joshbetz:
r140053-deploy

PR #1265, opened 2019-05-23
Send alloptions alerts to vip-deploy-on-call

mjangda:
Better fits our internal workflows.


trepmal:
This is still going to be in IRC land?

mjangda:
@trepmal WIll go to both Slack and IRC.

mjangda:
> 4th parameter is necessary/doing anything unless you also set the 5th/interval param: 

An interval of `0` means no expiry (https://developer.wordpress.org/reference/functions/wp_cache_add/)

WPprodigy:
I see. The conditional will fail though: `if ( $kind && $interval ... )` = `if ( 'a8c-alloptions' && 0 ... ), which will fail.

mjangda:
Oh, hah!

mjangda:
r140294-deploy

PR #1276, opened 2019-06-06
Force all VIP sites to use local domain for Photon

### LINKED ISSUE #1274: Jetpack: default Photon i*.wp.com domains not fully phased out for VIP sites
VIP sites should not be running through the default WP.com CDN for images. This was confirmed as per my chat with @darthhexx. There is also various compatibility issues that can occur under these current circumstances.

I do not see a meta item set for `WPCOM_VIP_USE_JETPACK_PHOTON` (which can toggle this on) but the home page still does have instances of i*.wp.com (recent uploads too).

https://github.com/Automattic/vip-go-mu-plugins/blob/c102bbe1d57bceba70255561a161a1d12be0a230/a8c-files.php#L147

Original change for this is here
https://github.com/Automattic/vip-go-mu-plugins/blob/5d8d29afc7da243e398074d3b309ac944edd5d9f/vip-jetpack/vip-jetpack.php#L21-L31

mjangda:
The VIP File Service has Photon capabilities, but is served from the same domain. If the Photon module is enabled, force Jetpack to use the files service instead of the default Photon domains (`i*.wp.com`) for internal files.

Externally hosted files continue to use the remote Photon service.

Previously, this was only enabled if you had the `WPCOM_VIP_USE_JETPACK_PHOTON` constant enabled. This enables it for all since we want this behaviour everywhere. We also don't need to wait until #1012 is deployed.

Fixes #1274 

## Checklist

## Steps to Test

1. Check out PR.
1. Enable Photon module (`wp jetpack module enable photon`)
1. Ensure `WPCOM_VIP_USE_JETPACK_PHOTON` constant isn't set.
1. Load a post with a local image and an externally hosted image.
1. Verify that the local image is loaded from the site itself and external image from the `i*.wp.com` domain.

Then, enable the `WPCOM_VIP_USE_JETPACK_PHOTON` constant and verify that the behaviour doesn't change.

Then, retest with the Photon module disabled and verify that all images load from their respective domains.

### DIFF
--- a/a8c-files.php
+++ b/a8c-files.php
@@ -119,6 +119,11 @@ class A8C_Files {
 		} else {
 			$this->init_vip_photon_filters();
 		}
+
+		// The VIP File Service has Photon capabilities, but is served from the same domain.
+		// If the Photon module is enabled, force Jetpack to use the files service instead of the default Photon domains (`i*.wp.com`) for internal files.
+		// Externally hosted files continue to use the remote Photon service.
+		add_filter( 'jetpack_photon_domain', [ 'A8C_Files_Utils', 'filter_photon_domain' ], 10, 2 );
 	}
 
 	private function init_jetpack_photon_filters() {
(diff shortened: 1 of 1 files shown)


nickdaugherty (dismissed):
Looks like a straightforward change that shouldn't have any side effects.

NOTE - haven't actually tested this.

nickdaugherty:
@mjangda - does it get confusing that `A8c_Files::use_jetpack_photon()` can return `true` but the image domain is still the site's domain (not `i*.wp.com`)?

pschoffer:
Hi @mjangda / @nickdaugherty .

Is this still a wanted change?

If so could you please point me to a good direction on how to test this? I tried on my vip go site to setup post with a local image and remote image - https://pschoffer-develop.go-vip.net/2021/08/09/testing-photon/. None of them are sourced from `i*.wp.com` with or without this change.

But, I guess I am missing a key understanding of what the photon service is :) I did look up that it is part of [JP accelerator](https://jetpack.com/support/site-accelerator/), but I failed to find the switch to enable it as described in the linked doc.

This is how my JP settings page looks like:
image

WPprodigy:
CC @itsTallulah, thought you might have thoughts here :)

PR #1277, opened 2019-06-07
Add phone verification step when configuring SMS two-factor

### LINKED ISSUE #535: SMS two factor provider configuration confirmation
Unlike the TOTP method, there’s no confirmation that the phone number you’ve entered is correct. 

It might be nice to send the user a “check code” which they enter, the provider confirms the code worked, and then we know it’s all configured correctly before they next have to log in (which would be a bad time to discover a misconfiguration).

dchymko:
## Description
Add a confirmation step to the SMS provider to make sure the user is able to receive messages at the phone number they entered

Fixes #535 

## Steps to Test
1. Go into the Edit profile for a user in wp-admin
2. Enter your SMS number and click Submit
3. The verification fields should now be where the phone number fields were.Enter the verification code and click submit"
4. The 'Correctly configured for <phone number>` message should now appear

Delete the number and repeat steps 1-4, this time checking the "enabled" checkbox on each step. It should only persist as 'enabled' once the 'correctly configured' message appears

### DIFF
--- a/wpcom-vip-two-factor/sms-provider.php
+++ b/wpcom-vip-two-factor/sms-provider.php
@@ -158,7 +158,7 @@ class Two_Factor_SMS extends Two_Factor_Provider {
 				Correctly configured for <?php echo esc_attr( $sms ); ?>.
 				<input type="submit" class="button" name="vip-two-factor-phone-delete"
 					   value="<?php esc_attr_e( 'Delete', 'two-factor' ); ?>"/>
-			<?php elseif ( !empty( $hashed_token ) ) : ?>
+			<?php elseif ( ! empty( $hashed_token ) ) : ?>
 				<p>
 					Verification code has been sent to <?php echo esc_attr( $sms ); ?>
 					<input type="submit" class="button" name="vip-two-factor-phone-send-code"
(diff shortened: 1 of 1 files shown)


dchymko:
Is there a reason why we only send the code in the SMS message? Would even a "Here's the security code you requested" be better?

https://github.com/Automattic/vip-go-mu-plugins/blob/e0df2c94cb27b93eb9b68fe5474f5ae605336b21/wpcom-vip-two-factor/sms-provider.php#L91

wpcom-vip-two-factor/sms-provider.php
```diff
@@ -147,13 +149,38 @@ public function is_available_for_user( $user ) {
...
 		<div>
-			<label>Phone Number
-				<input name="vip-two-factor-phone" type="tel" placeholder="+14158675309" value="<?php echo esc_attr( $sms );?>" />
-			</label>
-			<p><strong>Note:</strong> Please include your country calling code (e.g. +44, +1, +61, etc.) to ensure SMS messages are correctly sent.</p>
+			<?php if ( '1' === $sms_configured ) : ?>
+				Correctly configured for <?php echo esc_attr( $sms ); ?>.
+				<input type="submit" class="button" name="vip-two-factor-phone-delete"
+					   value="<?php esc_attr_e( 'Delete', 'two-factor' ); ?>"/>
+			<?php elseif ( !empty( $hashed_token ) ) : ?>
```
dchymko:
```suggestion
			<?php elseif ( ! empty( $hashed_token ) ) : ?>
```
whitespace

joshbetz:
> Is there a reason why we only send the code in the SMS message? Would even a "Here's the security code you requested" be better?

Yeah, or "Your security code for home_url() is: "

dchymko:
r622-vip-go-stacks-deploy

PR #1278, opened 2019-06-10
Allow VIP_2FA_TIME_GATE to be overwritten by other mu-plugins

joshbetz:
We'll use this to enable 2FA for all newly created sites.


joshbetz:
r614-vip-go-stacks

PR #1281, opened 2019-06-12
Disable current theme validation

joshbetz:
By default, WordPress falls back to a default theme if it can't find
the active theme. This is undesirable because it requires manually
re-activating the correct theme and can lead to data loss in the form
of missing sidebar widgets.

## Checklist

## Steps to Test

1. Checkout PR
1. Activate a non-default theme.
1. Delete the active theme.
1. Navigate to the Appearance > Themes page.
1. Observe the error message, but the theme does not revert to default.

Expected notice with this change:

```
ERROR: The theme directory "twentyfifteen" does not exist.
```

Expected notice without this change:

```
The active theme is broken. Reverting to the default theme.
```

### DIFF
--- a/001-core.php
+++ b/001-core.php
@@ -9,6 +9,16 @@
 
 require_once( __DIR__ . '/001-core/privacy.php' );
 
+/**
+ * Disable current theme validation
+ *
+ * By default, WordPress falls back to a default theme if it can't find
+ * the active theme. This is undesirable because it requires manually
+ * re-activating the correct theme and can lead to data loss in the form
+ * of missing sidebar widgets.
+ */
+add_filter( 'validate_current_theme', '__return_false' );
+
 if ( false !== WPCOM_IS_VIP_ENV ) {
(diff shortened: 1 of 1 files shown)


001-core.php
```diff
@@ -9,6 +9,16 @@
 
 require_once( __DIR__ . '/001-core/privacy.php' );
 
+/**
+ * Disable current theme validation
+ *
+ * By default, WordPress falls back to a default theme if it can't find
+ * the active theme. This is undesirable because it requires manually
+ * re-activating the correct theme and can lead to data loss in the form
+ * of missing sidebar widgets.
```
GaryJones: "...can lead to data loss in the form of deactivated widgets and menu location assignments."

- Not all widgets are in a _sidebar_ (front end appearance), but they are in widget areas. 
- They don't get lost completely; they get moved to the Inactive section of the widgets page.
- Menu locations assignments are another practical thing they would otherwise need to fix.

joshbetz:
r643-vip-go-stacks

PR #1285, opened 2019-06-14
Cache Personalization: Tweaks to encrypted segmentation flow

dchymko:
## Description
Sets the encrypted cookie in orer for Varnish to properly decrypte and deconstruct on future requests
## Steps to Test

1. Use a segmentation example such as:
https://github.com/Automattic/vip-go-mu-plugins/blob/master/cache/examples/segmentation-beta/beta.php
2. After the register_groups command, add `Vary_Cache::enable_encryption();`
3. Ensure the group logic works as it did in the unencrypted version

### DIFF
--- a/cache/class-vary-cache.php
+++ b/cache/class-vary-cache.php
@@ -362,8 +362,7 @@ class Vary_Cache {
 	private static function encrypt_cookie_value( $value ) {
 		$client_key = constant( 'VIP_GO_AUTH_COOKIE_KEY' );
 		$client_iv = constant( 'VIP_GO_AUTH_COOKIE_IV' );
-		$random = random_int( 0,  PHP_INT_MAX ) . random_int( 0,  PHP_INT_MAX );
-		$cookie_value = $random . '|' . $value . '|' . ( time() + self::$cookie_expiry );
+		$cookie_value = random_int( 0,  PHP_INT_MAX ) . '|' . $value . '|' . ( time() + self::$cookie_expiry );
 		$cipher_cookie = openssl_encrypt( $cookie_value, 'aes-128-cbc', $client_key, 0, $client_iv );
 
 		return $cipher_cookie;
--- a/cache/class-vary-cache.php
+++ b/cache/class-vary-cache.php
@@ -540,7 +540,8 @@ class Vary_Cache {
 
 		if ( self::is_encryption_enabled() ) {
 			$cookie_value = self::encrypt_cookie_value( $group_string );
-			self::set_cookie( self::COOKIE_AUTH, $cookie_value );
+			$base_64 = base64_encode( $cookie_value ) ;
+			self::set_cookie( self::COOKIE_AUTH, VIP_GO_APP_ID . '.' . $base_64 );
 		} else {
 			self::set_cookie( self::COOKIE_SEGMENT, $group_string );
 		}
--- a/cache/class-vary-cache.php
+++ b/cache/class-vary-cache.php
@@ -576,7 +576,7 @@ class Vary_Cache {
 			if ( self::is_encryption_enabled() ) {
 				header( 'Vary: X-VIP-Go-Auth' );
 			} else {
-				header( 'Vary: X-VIP-Go-Segmentation: '. $group_string );
+				header( 'Vary: X-VIP-Go-Segmentation' );
 			}
 
 			if ( defined( 'WP_DEBUG' ) && true === WP_DEBUG ) {
--- a/cache/class-vary-cache.php
+++ b/cache/class-vary-cache.php
@@ -591,7 +591,7 @@ class Vary_Cache {
 	 */
 	private static function set_cookie( $name, $value ) {
 		$expiry = time() + self::$cookie_expiry;
-		setcookie( $name, $value, $expiry, COOKIEPATH, COOKIE_DOMAIN );
+		setrawcookie( $name, $value, $expiry, COOKIEPATH, COOKIE_DOMAIN );
 	}
 
 	/**
--- a/cache/class-vary-cache.php
+++ b/cache/class-vary-cache.php
@@ -420,8 +420,8 @@ class Vary_Cache {
 			// If the header auth isn't set (in case of a logged-in user), fall back to decrypting the cookie itself.
 			$auth_cookie = null;
 			// $_COOKIE is automatically urldecoded, so we need to search through the $_SERVER version to get the unencoded one.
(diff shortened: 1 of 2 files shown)


cache/class-vary-cache.php
```diff
@@ -361,7 +362,8 @@ public static function is_encryption_enabled() {
 	private static function encrypt_cookie_value( $value ) {
 		$client_key = constant( 'VIP_GO_AUTH_COOKIE_KEY' );
 		$client_iv = constant( 'VIP_GO_AUTH_COOKIE_IV' );
-		$cookie_value = random_bytes( 32 ) . '|' . $value . '|' . ( time() + self::$cookie_expiry );
+		$random = random_int( 0,  PHP_INT_MAX ) . random_int( 0,  PHP_INT_MAX );
```
mjangda: Can we add some more context around why we're generating the random this way?

cache/class-vary-cache.php
```diff
@@ -413,9 +414,27 @@ private static function parse_nocache_cookie() {
...
 	 */
 	private static function parse_group_cookie() {
-		if ( self::is_encryption_enabled() && ! empty( $_COOKIE[ self::COOKIE_AUTH ] ) ) {
-			$cookie_value = self::decrypt_cookie_value( $_COOKIE[ self::COOKIE_AUTH ] );
-		} elseif ( ! empty( $_COOKIE[ self::COOKIE_SEGMENT ] ) ) {
+		//use the decrypted value provided from the cache layer
+		if ( self::is_encryption_enabled() && isset( $_SERVER[ self::HEADER_AUTH ] ) &&  ! empty( $_SERVER[ self::HEADER_AUTH ] ) ) {
+			$cookie_value = $_SERVER[ self::HEADER_AUTH ];
+		} elseif ( self::is_encryption_enabled() && ! empty( $_COOKIE[ self::COOKIE_AUTH ] ) ) {
+			// if the header auth isn't set (in case of a logged-in user), fall back to decrypting the cookie itself.
```
mjangda:
```suggestion
			// If the header auth isn't set (in case of a logged-in user), fall back to decrypting the cookie itself.
```

cache/class-vary-cache.php
```diff
@@ -413,9 +414,27 @@ private static function parse_nocache_cookie() {
...
+
+			$auth_cookie = null;
+			//$_COOKIE is automatically urldecoded, so we need to search through the $_SERVER version to get the unencoded one.
```
mjangda:
```suggestion
			// $_COOKIE is automatically urldecoded, so we need to search through the $_SERVER version to get the unencoded one.
```

cache/class-vary-cache.php
```diff
@@ -413,9 +414,27 @@ private static function parse_nocache_cookie() {
...
+			foreach(explode('; ',$_SERVER['HTTP_COOKIE']) as $rawcookie)
```
mjangda: We should fix the spacing / formatting in this block.

cache/class-vary-cache.php
```diff
@@ -413,9 +414,27 @@ private static function parse_nocache_cookie() {
...
+				if( self::COOKIE_AUTH === $k) {
+					$auth_cookie = $v;
+					break;
+				}
+			}
+
+			$value = ltrim( $auth_cookie, VIP_GO_APP_ID . '.' ); // remove the site prefix
+			$cookie_value = self::decrypt_cookie_value( $value );
+
+		}elseif ( ! empty( $_COOKIE[ self::COOKIE_SEGMENT ] ) ) {
```
mjangda:
```suggestion
		} elseif ( ! empty( $_COOKIE[ self::COOKIE_SEGMENT ] ) ) {
```

cache/class-vary-cache.php
```diff
@@ -590,7 +609,8 @@ private static function send_vary_headers() {
 	 */
 	private static function set_cookie( $name, $value ) {
 		$expiry = time() + self::$cookie_expiry;
-		setcookie( $name, $value, $expiry, COOKIEPATH, COOKIE_DOMAIN );
+		//need to use setrawcookie() here to prevent PHP from URLEncoding the base-64 terminator (==) on encrypted payloads
```
mjangda:
```suggestion
		// Need to use setrawcookie() here to prevent PHP from URLEncoding the base-64 terminator (==) on encrypted payloads
```

tests/cache/test-vary-cache.php
```diff
@@ -669,9 +675,13 @@ public function get_test_data__parse_group_cookies() {
...
 				],
 				[
-					'vip-go-auth' => 'qSME2LdfVuNvZa0GfeUJ45/uHCu7Auqj3iesSL4CITteGfva/N0wQ5TCIzcyeBImeTf3On2P4f7EtQyviw2ooA==',
 				],
 				[
+					'HTTP_X_VIP_GO_AUTH' => 'vc-v1__design-group_--_yes---__dev-group_--_no'
+				],
+				[
+					'design-group' => 'yes' ,
+					'dev-group' => 'no'
```
mjangda: I think we still need a test case for the case where header doesn't exist and we fall to the cookie? (i.e. `if the header auth isn't set (in case of a logged-in user)`)

cache/class-vary-cache.php
dchymko: I simplified this, I think trying to duplicate the 32 byte random characters was overkill

cache/class-vary-cache.php
```diff
@@ -413,8 +414,9 @@ private static function parse_nocache_cookie() {
 	 * Parses the group/segment cookie into the local groups array of key-values.
 	 */
 	private static function parse_group_cookie() {
-		if ( self::is_encryption_enabled() && ! empty( $_COOKIE[ self::COOKIE_AUTH ] ) ) {
-			$cookie_value = self::decrypt_cookie_value( $_COOKIE[ self::COOKIE_AUTH ] );
+		//use the decrypted value provided from the cache layer
```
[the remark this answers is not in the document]
dchymko: Done. Ideally we want to use the decrypted version passed from the cache layer, as that saves having to do it a second time. 

The fallback to the cookie is only intended as a convenience to have the same functionality to logged in users ( for testing, etc). I wonder if we can/should add some sort of tracking here to flag cases where for some reason we're hitting a lot of these (uncached) calls?

dchymko:
> I had a few changes originally that I see have been made already, so LGTM now except for the spacing in this code segment:

Corrected

dchymko:
r620-vip-go-stacks-deploy

PR #1287, opened 2019-06-24
Fixes attachments with completely missing meta

emrikol:
If an attachment exists, but has no attachment meta, we should still add the filesize.  This is already set up to happen in `Meta_Updater::update_attachment_filesize()` but the check in `Meta_Updater::get_attachments()` does not allow the attachment ID to be fixed if it does not already have attachment metadata.

## Checklist

## Steps to Test

Example:

1. Check out PR.
1. Delete an attachment meta, or find an attachment that is missing meta
1. Run the WP-CLI command to fix missing filesizes
1. Verify filesizes are added to attachment meta.


brandon-m-skinner:
r907-stacks

PR #1289, opened 2019-06-28
Remove limit from `get_sites`, when running with the `--network` flag.

vaurdan:
## Description

By default, `get_sites()` will only return, at most, 100 sites. On larger networks, with more than 100 sites, running `wp jetpack-start connect --network` would not get past these first hundred. By including the `number` argument, `get_sites()` will return all the sites in the network, making it  possible to connect all the existing sites. 

## Checklist

## Steps to Test

Run `wp jetpack-start connect --network` on a site with more than 100 subsites. See if the script tries to connect all the subsites.


brandon-m-skinner:
@vaurdan Is this PR still valid? If so I'd like to pursue its completion/deployment.

brandon-m-skinner:
Still valid 👍

brandon-m-skinner:
r898-stacks

Issue #1280, opened 2019-06-12
Bulk editing limitations


WPprodigy:
"Pages" are hierarchical in WordPress, meaning you can nest pages underneath a parent. As a result, WP sets the main query's posts_per_page limit to `-1` on hierarchical post types for sorting reasons (so all of the children can show under the parent I assume).

This happens here in WordPress core: https://github.com/WordPress/WordPress/blob/c77e771c849a85b63cef1b9457ad1a501a331afa/wp-admin/includes/post.php#L1169-L1176

And on VIP, we limit bulk editing actions when there is more than a 20 `posts_per_page` limit. This happens here: https://github.com/Automattic/vip-go-mu-plugins/blob/a053b6e8b4f64e4a8ad37b39149d1a636f76b563/performance/bulk-edit.php#L71-L73

So on any page editing screen, we are imposing the bulk editing limit. This used to not be the case, but I believe this change in core amplified the bug: https://github.com/WordPress/WordPress/commit/921fd015826a442dc58444f9f07d21def54f7e03 CC @david-binda 

I see that the page edit screen still limits to 20 per page by default though, so I feel there is something we could do on the VIP logic side to accompany this behavior better.

### Replication

- On any VIP test site, add at least 21 pages. 
- Note how bulk editing is disabled.

david-binda:
@WPprodigy 👋Thanks for the ping.

However, I don't think that the WordPress/WordPress@921fd01 is the culprit here. That commit hotfixes an issue introduced in https://github.com/WordPress/WordPress/commit/5a30da57c68595949e917645130ac01094f3c44b which only existed in a RC version of the WordPress.

I mean, reverting the changeset you identified possibly fixes the issue you are describing, but break something else. I believe that in case you were testing the version of the code prior https://github.com/WordPress/WordPress/commit/5a30da57c68595949e917645130ac01094f3c44b it would still be reproducible.

If this really were working at some point in the past, I would look for some breaking change prior https://github.com/WordPress/WordPress/commit/5a30da57c68595949e917645130ac01094f3c44b, not the https://github.com/WordPress/WordPress/commit/921fd015826a442dc58444f9f07d21def54f7e03 .

Please, let me know if it makes sense, or whether you have any related questions!

PR #1291, opened 2019-07-02
Allow bulk editing for hierarchical post types

WPprodigy:
## Description

Fixes https://github.com/Automattic/vip-go-mu-plugins/issues/1280

At the moment, pages (and other hierarchical post types) cannot be bulk edited when there are more than 20 total, regardless of how many are set to show per page. This is because WP core sets the `posts_per_page` query arg to -1 here for these post types, [seen here](https://github.com/WordPress/WordPress/blob/c77e771c849a85b63cef1b9457ad1a501a331afa/wp-admin/includes/post.php#L1169-L1176). I assume this is so it can properly list children under their parents.

The `per_page` value/option is still respected though when it comes to how many posts show per pages in the lists table though, [seen here](https://github.com/WordPress/WordPress/blob/cf3fa9f7c8038447303eea1fbd3636c5fece28cd/wp-admin/includes/class-wp-posts-list-table.php#L155). So we should check this before assuming bulk editing needs to be disabled.

## Checklist

## Steps to Test

Example:

1. On any VIP test site, add at least 21 pages. Note how bulk editing is disabled.
1. Check out this PR and test again. Bulk editing should work when set to 20 or less per page.

### DIFF
--- a/performance/bulk-edit.php
+++ b/performance/bulk-edit.php
@@ -2,8 +2,7 @@
 
 namespace Automattic\VIP\Performance;
 
-// Core defaults to 20, so let's assume that's safe
-const BULK_EDIT_LIMIT = 20;
+const BULK_EDIT_LIMIT = 40;
 
 /**
  * Bulk edits of lots of posts can trigger slow term count queries for each post updated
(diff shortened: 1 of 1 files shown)


WPprodigy:
Also, what are your thoughts on maybe increasing this number to be higher than 20? I understand the need to prevent huge edits, but even ~100 posts at a time seems reasonable.

dchymko:
vip-go-stacks-deploy-r632

PR #1294, opened 2019-07-02
Add/delete attachment url post id cache

paulschreiber:
This is an improved version of https://github.com/Automattic/vip-go-mu-plugins/pull/1284 that addresses @nickdaugherty's comments.

### DIFF
--- a/vip-helpers/vip-caching.php
+++ b/vip-helpers/vip-caching.php
@@ -605,7 +605,6 @@ add_action( 'delete_attachment', function ( $post_id ) {
 	$url = wp_get_attachment_url( $post_id );
 	$cache_key = wpcom_vip_attachment_cache_key( $url );
 	wp_cache_delete( $cache_key );
- );
 } );
 
 /**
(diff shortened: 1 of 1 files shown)


vip-helpers/vip-caching.php
```diff
@@ -576,22 +576,37 @@ function wpcom_vip_get_adjacent_post( $in_same_term = false, $excluded_terms = '
...
 
+/**
+ * Remove cached post ID when attachment post deleted
+ *
+ * @see wpcom_vip_attachment_url_to_postid()
+ */
+add_action( 'delete_attachment', function ( $post_id ) {
+	$url = wp_get_attachment_url( $post_id );
+	$cache_key = wpcom_vip_attachment_cache_key( $url );
+	wp_cache_delete( $cache_key );
```
nickdaugherty: Here we'll need to pass in `default` for the group as well - `wp_cache_delete()` defaults to an empty string, which will get passed into `$wp_object_cache->delete()` as an empty string and the default value of `default` on that function won't be used (since an empty string was passed in, rather than `null`).

```suggestion
	wp_cache_delete( $cache_key, 'default' );
```

paulschreiber:
Great. Let's get this fix out to everyone.

nickdaugherty:
Thanks @paulschreiber! This has now been deployed to VIP Go.

nickdaugherty:
r634-stacks

PR #1306, opened 2019-07-16
Update Two Factor Report Cap to manage_options

joshbetz:
Per https://github.com/Automattic/vip-go-mu-plugins/commit/c8fde7992ab2101915ae493bfa4e21fa2793352a#diff-9f0b8e5764792448766b0150acdd71a6, we're only requiring two factor for admins to start.


joshbetz:
r627-vip-go-stacks

PR #1309, opened 2019-07-19
Don't enforce 2FA by default in local dev

joshbetz:
Adds a `wpcom_vip_is_two_factor_local_testing` filter. 2FA enforcement will be disabled by default in local environments, but if you need to test the 2FA functionality, you can easily enable it in local environments with this filter.

We generally try to make the local environment as close as possible to production, but this is an instance where it would just get in the way most of the time.

Some things to consider during review:

1. The idea of conditionally enabling 2FA enforcement in local dev. Generally we try to keep local dev as close as possible to production, but this seems like a good example of where it will be more annoying than helpful most of the time.
2. The approach used here. I don't know of other functionality that you can optionally enable in local dev. How do we feel about this constant + filter approach?


joshbetz:
r630-vip-go-stacks

PR #1310, opened 2019-07-24
Two Factor: Add `wpcom_vip_is_two_factor_forced` before default to allow overriding

joshbetz:
By adding this filter at default priority, it is applied after code in client-mu-plugins and therefore overrides client code unless clients specifically add it at priority 11. By changing this to 9, we honor client code at the default priority.


joshbetz:
r635-vip-go-stacks

PR #1311, opened 2019-07-26
2FA: Don't restrict wpcomvip, logins are blocked anyway

joshbetz:
The wpcomvip user is responsible for the Jetpack connection, among other things. We don't want to restrict access here and logins are blocked anyway.

### Testing

I've manually tested on a sandbox with my support user and just changed the username instances to `wpcomvip`.


joshbetz:
r637-vip-go-stacks

PR #1326, opened 2019-09-27
Strip query string from file path when deleting files

nickdaugherty:
## Description

Before deleting files, strip off any query string params that snuck through.

WIP

## Checklist

## Steps to Test

TBD

### DIFF
--- a/files/class-vip-filesystem.php
+++ b/files/class-vip-filesystem.php
@@ -292,6 +292,13 @@ class VIP_Filesystem {
 			$file_path = substr( $file_path, strlen( $upload_path['basedir'] ) + 1 );
 		}
 
+		// Strip any query params that snuck through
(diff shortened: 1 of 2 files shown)


nickdaugherty:
It does do that, but we don't use `VIP_Filesystem::get_file_uri_path()` when calling `unlink()`:

https://github.com/Automattic/vip-go-mu-plugins/blob/master/files/class-vip-filesystem.php#L258-L260

AFAICT, that can end up sending query params through to the filesystem backend, and the delete will fail b/c that path doesn't exist.

hanifn:
Looks like you're right. This concern was actually something we were aware of when developing the stream wrappers but we must have missed that we the strip query results were being used in the actual `unlink`. Good catch @nickdaugherty !

hanifn:
r670-stacks

PR #1329, opened 2019-10-02
VIP FS: Remove Stream Wrapper exclusion for AMS

GaryJones:
The AMS datacenter now has support for the new file system, so the exclusion can be removed.


nickdaugherty:
NOTE - Stream Wrappers aren't _quite_ ready for AMS so we'll have to hold off on this until we get the go-ahead from the SysEng team.

GaryJones:
Closing in favour of @mjangda's merged #1474.

PR #1331, opened 2019-10-07
Fix namespacing for remove_action of do_initial_sync

rebeccahum:
## Description

Initial syncs are being triggered again overwriting full syncs. See #1257.  However, Jetpack was re-factored and we need to update it with the correct namespace: https://github.com/Automattic/jetpack/blob/a5a11059d24eb7ee50b25dbdb08f04ead55a0d50/packages/sync/src/Main.php#L32


rebeccahum:
Just a note, this will need to be removed in the next JP release because of https://github.com/Automattic/jetpack/pull/13546.

nickdaugherty:
r667-stacks

PR #1332, opened 2019-10-12
Introduce new wpcom_vip_load_plugin_args filter

WPprodigy:
## Description

Add new filter to `wpcom_vip_load_plugin()`. Allows filtering the $plugin and $folder parameters.

Main purpose is for WPcom -> Go migrations, so we can allow for the previous functionality around versioning as well as resolve the deprecated approach of setting the $folder to `theme`. Secondary purposes could be to give clients some more flexibility around conditionally loading different plugins, versions, or using different locations.

I decided to pass `$load_release_candidate_not_used` as an extra argument to the filter rather than including it in the return. However, I'm pretty impartial to this as it could go either way. Doesn't matter at all as this variable currently is never used, as least right now.

The filter is designed to be as flexible as possible, to avoid getting in the way of any future edits to the function. Future args could easily be added, or even removed.
________

Example usage:

```
function wpcom_compat_load_plugin_args( $args, $version ) {
  if ( ! empty( $args['plugin'] ) && ! empty( $version ) && is_string( $args['plugin'] ) && is_string( $version ) ) {
    $args['plugin'] = "{$args['plugin']}-{$version}/{$args['plugin']}.php";
  }

  if ( ! empty( $args['folder'] ) && in_array( $args['folder'], [ 'theme', 'plugins' ], true ) ) {
    $args['folder'] = false;
  }

  return $args;
}
add_filter( 'wpcom_vip_load_plugin_args','wpcom_compat_load_plugin_args', 10, 2 );
```

### DIFF
--- a/vip-helpers/vip-utils.php
+++ b/vip-helpers/vip-utils.php
@@ -964,6 +964,7 @@ function _wpcom_vip_load_plugin_sanitizer( $folder ) {
  * @return bool True if the include was successful
  */
 function wpcom_vip_load_plugin( $plugin = false, $folder = false, $load_release_candidate_not_used = null ) {
+	// This filter is used in the WP.com compatibility plugin (https://github.com/Automattic/vip-go-wpcom-compat).
 	$filtered_args = apply_filters( 'wpcom_vip_load_plugin_args', array( 'plugin'  => $plugin, 'folder'  => $folder ), $load_release_candidate_not_used );
 	if ( is_array( $filtered_args ) ) {
 		$plugin = isset( $filtered_args['plugin'] ) ? $filtered_args['plugin'] : $plugin;
--- a/vip-helpers/vip-utils.php
+++ b/vip-helpers/vip-utils.php
@@ -964,6 +964,12 @@ function _wpcom_vip_load_plugin_sanitizer( $folder ) {
  * @return bool True if the include was successful
  */
 function wpcom_vip_load_plugin( $plugin = false, $folder = false, $load_release_candidate_not_used = null ) {
+	$filtered_args = apply_filters( 'wpcom_vip_load_plugin_args', array( 'plugin'  => $plugin, 'folder'  => $folder ), $load_release_candidate_not_used );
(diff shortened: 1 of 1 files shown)


WPprodigy:
Some other options:

1) Pass the third param in the original array that is returned. Would be a bit weird in the future if we used this third param for something else since the naming in the filter would be "locked in". 

```
$filtered_args = apply_filters( 'wpcom_vip_load_plugin_args', array( 'plugin'  => $plugin, 'folder'  => $folder, 'load_release_candidate' =>  $load_release_candidate_not_used ) );
if ( is_array( $filtered_args ) ) {
  $plugin = isset( $filtered_args['plugin'] ) ? $filtered_args['plugin'] : $plugin;
  $folder = isset( $filtered_args['folder'] ) ? $filtered_args['folder'] : $folder;
}
```

2) Return `false` by default and only action on the filter if it's changed.

```
$filtered_args = apply_filters( 'wpcom_vip_load_plugin_args', false, array( 'plugin'  => $plugin, 'folder'  => $folder, 'load_release_candidate' =>  $load_release_candidate_not_used ) );
if ( ! empty( $filtered_args ) ) {
  $plugin = isset( $filtered_args['plugin'] ) ? $filtered_args['plugin'] : $plugin;
  $folder = isset( $filtered_args['folder'] ) ? $filtered_args['folder'] : $folder;
}
```

3) One liner with extract(). Not super keen on this as it lacks the ability for extra validation though.

```
extract( apply_filters( 'wpcom_vip_load_plugin_args', array( 'plugin'  => $plugin, 'folder'  => $folder, 'load_release_candidate_not_used' => $load_release_candidate_not_used ) ) );
```

_____

Let me know if you have any questions or ideas. Happy to make any changes needed to get this pushed through.

rinatkhaziev:
Looks good, but let's add a code comment to provide more context for that filter, something like 
```
// This filter is used in WP.com compatibility plugin (https://github.com/Automattic/vip-go-wpcom-compat)
```

PR #1334, opened 2019-11-04
Do not purge cache if VIP_GO_DISABLE_CACHE_PURGING is enabled

chrean:
## Description

Adds an option to disable cache purging on content changes.

## Checklist

## Steps to Test

1. Apply changes to a platform sandbox host.
1. Sandbox a site (delete and restart if using an existing one)
1. Add the following to `vip-config/vip-config.php`: `define( 'VIP_GO_DISABLE_CACHE_PURGING', true );`
1. Add `error_log( 'Skipping cache purge' );` just before the `return` inside of `init()`, vip-cache-manager/vip-cache-manager.php`
1. Load a page, you should see `[date time] Skipping cache purge` in `/chroot/tmp/php-errors`.
1. Now take away the `define` and reload: you shouldn't see any additional line in `/chroot/tmp/php-errors`

### DIFF
--- a/vip-cache-manager/vip-cache-manager.php
+++ b/vip-cache-manager/vip-cache-manager.php
@@ -291,6 +291,10 @@ class WPCOM_VIP_Cache_Manager {
 		if ( $this->site_cache_purged )
 			return false;
 
+		if ( defined( 'WP_IMPORTING' ) ) {
+			return false;
+		}
+
 		$post = get_post( $post_id );
 		if ( empty( $post ) ||
 		     'revision' === $post->post_type ||
--- a/vip-cache-manager/vip-cache-manager.php
+++ b/vip-cache-manager/vip-cache-manager.php
@@ -290,8 +290,8 @@ class WPCOM_VIP_Cache_Manager {
 	public function queue_post_purge( $post_id ) {
 		if ( $this->site_cache_purged )
 			return false;
-
-		if ( defined( 'WP_IMPORTING' ) ) {
+		// Do not queue anything if an import is running or cache purging is disabled
+		if ( defined( 'WP_IMPORTING' ) || defined( 'VIP_GO_DISABLE_CACHE_PURGING' ) ) {
 			return false;
 		}
 
--- a/vip-cache-manager/vip-cache-manager.php
+++ b/vip-cache-manager/vip-cache-manager.php
@@ -291,7 +291,7 @@ class WPCOM_VIP_Cache_Manager {
 		if ( $this->site_cache_purged )
 			return false;
 		// Do not queue anything if an import is running or cache purging is disabled
-		if ( defined( 'WP_IMPORTING' ) || defined( 'VIP_GO_DISABLE_CACHE_PURGING' ) ) {
+		if ( defined( 'WP_IMPORTING' ) || ( defined( 'VIP_GO_DISABLE_CACHE_PURGING' ) && true === VIP_GO_DISABLE_CACHE_PURGING ) ) {
 			return false;
 		}
 
--- a/vip-cache-manager/vip-cache-manager.php
+++ b/vip-cache-manager/vip-cache-manager.php
@@ -215,6 +215,11 @@ class WPCOM_VIP_Cache_Manager {
 			return;
 		}
 
+		// Cache purging disabled, bail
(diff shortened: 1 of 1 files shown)


chrean:
I think we'd need to add some documentation for this... Shoudn't we?

nickdaugherty:
> I think we'd need to add some documentation for this... Shoudn't we?

Probably not super important, currently this will only be used internally by VIP to disable cache purging in certain circumstances, there isn't really a use-case for clients disabling purging entirely - they can also filter the purged urls, which _is_ documented.

GaryJones:
> this will only be used internally by VIP to disable cache purging in certain circumstances

I'm curious what those circumstances would be?

nickdaugherty:
> I'm curious what those circumstances would be?

Unlaunched sites that are importing large amounts of content via an automated process without setting `WP_IMPORTING`, as this ends up triggering _tons_ of unnecessary PURGE requests - currently there is an unlaunched site hitting 50k PURGEs per second 😱

vip-cache-manager/vip-cache-manager.php
```diff
@@ -215,6 +215,11 @@ public function execute_purges() {
 			return;
 		}
 
+		// Cache purging disabled, bail
+		if ( defined( 'VIP_GO_DISABLE_CACHE_PURGING' ) && true === VIP_GO_DISABLE_CACHE_PURGING ) {
+			return;
```
mjangda: Should we leave an `error_log` in there (like in the PR description) for added visibility?

nickdaugherty: The volume could be quite large for the use case we're adding this for (certain imports)...

vip-cache-manager/vip-cache-manager.php
```diff
@@ -215,6 +215,11 @@ public function execute_purges() {
 			return;
 		}
 
+		// Cache purging disabled, bail
+		if ( defined( 'VIP_GO_DISABLE_CACHE_PURGING' ) && true === VIP_GO_DISABLE_CACHE_PURGING ) {
```
joshbetz: We should also check the core `WP_IMPORTING` constant.

vip-cache-manager/vip-cache-manager.php
chrean: Makes sense, thanks for taking a look!

chrean: Added with 05d64815 . Totally open to reword the message, not 100% sure about it.

vip-cache-manager/vip-cache-manager.php
joshbetz: I see we check for `WP_IMPORTING` when URLs are queued. I wonder if we should also check for `VIP_GO_DISABLE_CACHE_PURGING` there as well? The benefit would be reduced memory usage since we wouldn't even put these URLs into the purge array.

https://github.com/Automattic/vip-go-mu-plugins/blob/76a6152d68542c54992ed2a462837d86a11bf060/vip-cache-manager/vip-cache-manager.php#L288

chrean: Makes totally sense, early check added with 532c0c72

vip-cache-manager/vip-cache-manager.php
```diff
@@ -285,7 +291,7 @@ public function queue_post_purge( $post_id ) {
 		if ( $this->site_cache_purged )
 			return false;
 
-		if ( defined( 'WP_IMPORTING' ) ) {
+		if ( ( defined( 'WP_IMPORTING' ) && true === WP_IMPORTING ) ) {
```
nickdaugherty:
```suggestion
		if ( defined( 'WP_IMPORTING' ) && true === WP_IMPORTING ) {
```

Think we have an extra set of parenthesis here.

chrean:
OK, tested, doesn't break anything.

vip-cache-manager/vip-cache-manager.php
chrean: Not logging anything anymore because we have chosen to put the code in `init()` and that would make the log flooded with cache purging messages.

chrean:
r674-deploy

PR #1335, opened 2019-11-07
Allow SSL verification to be turned off for local dev.

jenkoian:
## Description

Whilst developing locally it's useful to be able to turn off SSL
verification easily when using a self signed certificate. This change
will allow a constant to be set `WPCOM_VIP_DISABLE_SSL_VERIFY` (name up for debate) which
will disable ssl verify on `vip_safe_wp_remote_get()` calls.

There is some protection to ensure this doesn't ever occur on VIP envs
where ssl verify should always be turned on.

## Checklist

## Steps to Test

Example:

1. Have a local environment using a self signed certificate
2. Install [WPCOM-Legacy-Redirector](https://github.com/Automattic/WPCOM-Legacy-Redirector)
3. Attempt to add a redirect notice error 'Error: Redirects need to be from URLs that have a 404 status.' even if the URL in question is a 404.
4. Checkout this PR.
5. Add in vip-config.php `define( 'WPCOM_VIP_DISABLE_SSL_VERIFY', true );`
6. Re-attempt the above.

### DIFF
--- a/vip-helpers/vip-utils.php
+++ b/vip-helpers/vip-utils.php
@@ -492,6 +492,11 @@ function wpcom_vip_file_get_contents( $url, $timeout = 3, $cache_time = 900, $ex
 		'http_api_args' => array(), // See http://codex.wordpress.org/Function_API/wp_remote_get
 	);
 
+	if ( defined( 'WPCOM_VIP_DISABLE_SSL_VERIFY' ) && WPCOM_VIP_DISABLE_SSL_VERIFY ) {
+		// Only allow this to be false on non VIP envs.
+		$extra_args_defaults[ 'sslverify' ] = WPCOM_IS_VIP_ENV;
+	}
+
 	$extra_args = wp_parse_args( $extra_args, $extra_args_defaults );
 
 	$cache_key       = md5( serialize( array_merge( $extra_args, array( 'url' => $url ) ) ) );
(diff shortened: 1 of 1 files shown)


rinatkhaziev:
Thanks for the PR!.

Sorry that it took so long to review, but, unfortunately, this is not something we're willing to merge. The same can be achieved by using a local filter in your dev environment.

jenkoian:
No problem @rinatkhaziev - are you able to let me know what the filter is?

GaryJones:
@jenkoian I can't read Rinat's mind, but possibly https://developer.wordpress.org/reference/hooks/https_local_ssl_verify/ ?

jenkoian:
Thanks @GaryJones

rinatkhaziev:
@jenkoian Apologies, it was a bit late, so I forgot to add the link and my response lacked clarity. 

What I meant is using a local development helper (something like a `client-mu-plugins/no-ssl-verify.php`) that would not be a part of the site's repo. The filter I meant is [http_request_args](https://developer.wordpress.org/reference/hooks/http_request_args/) but [https_ssl_verify](https://developer.wordpress.org/reference/hooks/https_ssl_verify/) should work too and is probably better for your case.

https://developer.wordpress.org/reference/classes/wp_http/request/

PR #1337, opened 2019-11-08
VIP FS: Add append mode support to streamwrapper

hanifn:
## Description

Add support for append mode file operations into the streamwrapper class.

Currently using `fopen()` with `a` or `a+` mode or `file_put_contents()` with the `FILE_APPEND` flag does not work as expected. This should fix that.

## Checklist

## Steps to Test

Example:

1. Sandbox a test site
1. `wp shell`
1. Create new file: `file_put_contents( 'vip://wp-content/uploads/append-test.txt', 'new file' );`
1. Append to file: `file_put_contents( 'vip://wp-content/uploads/append-test.txt', ' appended', FILE_APPEND );`
1. Verify file contents is now `new file appended`.

### DIFF
--- a/files/class-vip-filesystem-stream-wrapper.php
+++ b/files/class-vip-filesystem-stream-wrapper.php
@@ -717,7 +717,7 @@ class VIP_Filesystem_Stream_Wrapper {
 	*
 	* @return  bool
 	 */
-	private function validate( $path, $mode ) {
+	public function validate( $path, $mode ) {
 		if ( ! in_array( $mode, self::ALLOWED_MODES, true ) ) {
 			trigger_error( "Mode not supported: { $mode }. Use one 'r', 'w', 'a', or 'x'." );
 
--- a/files/class-vip-filesystem-stream-wrapper.php
+++ b/files/class-vip-filesystem-stream-wrapper.php
@@ -237,6 +237,11 @@ class VIP_Filesystem_Stream_Wrapper {
 
 		if ( 'r' === $this->mode ) {
 			// No writes in 'read' mode
+			trigger_error(
+				sprintf( 'stream_flush failed for %s with error: No writes allowed in "read" mode #vip-go-streams', $this->path ),
+				E_USER_WARNING
+			);
+
 			return false;
 		}
 
--- a/files/class-vip-filesystem-stream-wrapper.php
+++ b/files/class-vip-filesystem-stream-wrapper.php
@@ -235,6 +235,11 @@ class VIP_Filesystem_Stream_Wrapper {
 			return false;
 		}
 
+		if ( 'r' === $this->mode ) {
+			// No writes in 'read' mode
+			return false;
+		}
+
 		try {
 			// Upload to file service
 			$result = $this->client
--- a/files/class-vip-filesystem-stream-wrapper.php
+++ b/files/class-vip-filesystem-stream-wrapper.php
@@ -10,6 +10,16 @@ class VIP_Filesystem_Stream_Wrapper {
 	const DEFAULT_PROTOCOL = 'vip';
 
 	/**
+	 * Allowed fopen modes
+	 *
+	 * We are ignoring `b`, `t` and `+` modes as they do not affect how this
+	 * Stream Wrapper works.
+	 * Not supporting `c` and `e` modes as these are rarely used and adds complexity
+	 * to support.
+	 */
+	const ALLOWED_MODES = [ 'r', 'w', 'a', 'x' ];
+
+	/**
 	 * The Stream context. Set by PHP
 	 *
 	 * @since   1.0.0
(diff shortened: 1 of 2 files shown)


nickdaugherty:
Crap, somehow missed the Review Request notification for this...looking now...

files/class-vip-filesystem-stream-wrapper.php
```diff
@@ -650,4 +680,69 @@ protected function close_handler() {
...
+	*
+	* @since   1.0.0
+	* @access  private
+	* @param   string    $path   Path to file
+	* @param   string    $mode   fopen mode
+	*
+	* @return  bool
+	 */
+	private function validate( $path, $mode ) {
+		if ( ! in_array( $mode, [ 'r', 'w', 'a', 'x' ], true ) ) {
```
nickdaugherty: Will `$mode` _always_ have a value?

files/class-vip-filesystem-stream-wrapper.php
```diff
@@ -122,6 +131,11 @@ public function register() {
 	 */
 	public function stream_open( $path, $mode, $options, &$opened_path ) {
 		$path = $this->trim_path( $path );
+		$mode = rtrim( $mode, 'bt' );
```
nickdaugherty: What does the `rtrim()` for `bt` do?

files/class-vip-filesystem-stream-wrapper.php
```diff
@@ -220,6 +235,11 @@ public function stream_flush() {
 			return false;
 		}
 
+		if ( 'r' === $this->mode ) {
+			// No writes in 'read' mode
+			return false;
```
nickdaugherty: Should this `trigger_error()` too?

files/class-vip-filesystem-stream-wrapper.php
nickdaugherty: PHP supports more modes than this:

https://www.php.net/manual/en/function.fopen.php

Specifically `r+`, `w+`, `a+`, `x+`, `c`, `c+`, and `e` (not really used). We don't necessarily need to implement those all right away, but we should at least document what is and isn't supported (and probably move the supported modes to a constant).

hanifn: > Will `$mode` _always_ have a value?

Yes as `$mode` is from `fopen()` and is a required argument.

hanifn: > We don't necessarily need to implement those all right away, but we should at least document what is and isn't supported (and probably move the supported modes to a constant).

This is a good idea. I'll add that in. Also, we're ignoring the `+` modes as in our current implementation, all the file handles are read/write capable.

files/class-vip-filesystem-stream-wrapper.php
hanifn: `b` is force binary mode and `t` is Windows-only text mode. It's used along with the regular modes of `fopen()` like so: `rb` or `r+b`. Both of which does not apply, I believe, to the functionality of the stream wrapper thus we're stripping them here.

files/class-vip-filesystem-stream-wrapper.php
hanifn: I originally did not think so but a `trigger_error()` could help with debugging so I'll add that in.

files/class-vip-filesystem-stream-wrapper.php
hanifn: Changes made in a145ce99ed9c3643e9f5db25e440a0894d283007

files/class-vip-filesystem-stream-wrapper.php
hanifn: Changes made in f60d94b3aba76cc7357adb618ed9e13448ab9063
I've also added the same check to `stream_write()`

hanifn:
Hmmm... `A8C_Files_ImageSizes_Test` tests are failing

files/class-vip-filesystem-stream-wrapper.php
nickdaugherty: Awesome!

nickdaugherty:
> Hmmm... A8C_Files_ImageSizes_Test tests are failing

Once we figure that out, how difficult would it be to add a test for the new `a` mode support?

mjangda:
The test failures are because of an upstream change in 5.3 (core adds two new image sizes)

chrean:
@hanifn what's the status on this one? Do we need to add tests as @nickdaugherty suggested, or?

hanifn:
@chrean TBH, I'm not exactly sure how to test this in PHPUnit. Should we just merge this and then figure that out later?

files/class-vip-filesystem-stream-wrapper.php
```diff
@@ -650,4 +706,69 @@ protected function close_handler() {
...
+	 * Validates the provided stream arguments for fopen.
+	*
+	* @since   1.0.0
+	* @access  private
+	* @param   string    $path   Path to file
+	* @param   string    $mode   fopen mode
+	*
+	* @return  bool
+	 */
+	private function validate( $path, $mode ) {
```
nickdaugherty: If we make this a `public` method, it becomes testable with PHPUnit.
```suggestion
	public function validate( $path, $mode ) {
```

hanifn:
r681-stacks

PR #1338, opened 2019-11-08
Revert #1257 now that https://github.com/Automattic/jetpack/pull/1354 is live

rebeccahum:
## Description

#1257 was done as stopgap for initial syncs being interrupted.  Now that https://github.com/Automattic/jetpack/pull/1354 is in 7.9, we should remove it.


rebeccahum:
@chrean That is correct, the whole block of code is irrelevant now.

chrean:
r676-deploy

PR #1340, opened 2019-11-12
Disable the big image threshold in anticipation of WordPress 5.3 rollout

david-binda:
WordPress 5.3 comes with a new feature capping the max size of uploaded images. This does not seem to be needed on VIP Go as we are using Photon for media handling.

A possible side-effects of limiting the max size of images are users unable to upload hi-res images to their sites.

This can safely land prior 5.3 rollout.

### DIFF
--- a/misc.php
+++ b/misc.php
@@ -151,3 +151,9 @@ add_filter( 'wp_link_query_args', 'wpcom_vip_wp_link_query_args', 10, 1 );
  * Stop Woocommerce from trying to create files on read-only filesystem
  */
 add_filter( 'woocommerce_install_skip_create_files', '__return_true' );
+
+/**
+ * WordPress 5.3 adds "big image" processing, for images over 2560px (by default).
+ * This is not needed on VIP Go since we use Photon for dynamic image work.
+ */
+add_filter( 'big_image_size_threshold', '__return_false' );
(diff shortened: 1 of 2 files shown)


misc.php
```diff
@@ -151,3 +151,9 @@ function wpcom_vip_wp_link_query_args( $query ) {
  * Stop Woocommerce from trying to create files on read-only filesystem
  */
 add_filter( 'woocommerce_install_skip_create_files', '__return_true' );
+
+/**
+ * WordPress 5.3 adds "big image" processing, for images over 2560px (by default).
+ * This is not needed on VIP Go since we use Photon for dynamic image work.
+ */
+add_filter( 'big_image_size_threshold', '__return_false' );
```
mjangda: I'm thinking this should probably be moved to `a8c-files.php` since that contains all image processing code.

david-binda: Good call!

mjangda:
Looks like 5.3 added two new image sizes so some of our test assertions now fail (they just need to be updated): https://github.com/WordPress/wordpress-develop/blob/db4f746b4aa13ce87b065038c8dfea6f7dc9253b/src/wp-includes/media.php#L4410-L4428

mjangda:
Tests have been fixed; updated branch and will merge + deploy after CI.

mjangda:
r686-stacks

PR #1341, opened 2019-11-13
Remove 2FA time gate

joshbetz:
The time for enforcing 2FA has come and gone. We can simplify the logic to remove this unnecessary check.


joshbetz:
r685-vip-go-stacks

PR #1342, opened 2019-11-15
Two Factor: Add wpcom_vip_use_custom_sso filter

justnorris:
This will make `wpcom_vip_should_force_two_factor` more flexible by allowing other third party SSO solutions or custom logic to help figure out whether or not 2FA should be forced.

## Description
This PR will allow customization of whether or not 2FA should be forced. 

Two Usage examples:
* Implement a custom SSO Plugin other than OneLogin or SimpleSaml
* Force 2FA for users that aren't using SSO. For example:

```php
add_filter( 'wpcom_vip_use_custom_sso', function( $enabled ) {
	
	// The user `justnorris` isn't using OneLogin, force 2FA for him
	$user = wp_get_current_user();
	if ( isset( $user->user_login ) && 'justnorris' === $user->user_login  ) {
		return false;
	}

	// Check if OneLogin SSO is enabled
	if ( function_exists( 'is_saml_enabled' ) && is_saml_enabled() ) {
		return true;
	}

	return $enabled;

} );
```

## Checklist

## Steps to Test

1. Apply PR
2. Validate that nothing has changed to the login functionality when SSO is used
3. Use the code snippet from above to force Two-Factor on your user (You'll likely also need to disable proxy/local checks in `wpcom_vip_should_force_two_factor() )

### DIFF
--- a/two-factor.php
+++ b/two-factor.php
@@ -47,34 +47,32 @@ function wpcom_vip_should_force_two_factor() {
 		return false;
 	}
 
-	// Don't force 2FA for OneLogin SSO
-	if ( function_exists( 'is_saml_enabled' ) && is_saml_enabled() ) {
-		return false;
-	}
-
-	// Don't force 2FA for SimpleSaml
-	if ( function_exists( '\HumanMade\SimpleSaml\instance' ) && \HumanMade\SimpleSaml\instance() ) {
+	// Allow custom SSO solutions
+	if ( wpcom_vip_use_custom_sso() ) {
 		return false;
 	}
 
 	return true;
 }
 
-function wpcom_vip_is_jetpack_authorize_request() {
-	return (
-		// XML-RPC Jetpack authorize request
-		// This works with the classic core XML-RPC endpoint, but not
-		// Jetpack's alternate endpoint.
-		defined( 'XMLRPC_REQUEST' ) && XMLRPC_REQUEST
-		&& isset( $_GET['for'] ) && 'jetpack' === $_GET['for']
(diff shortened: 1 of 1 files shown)


two-factor.php
```diff
@@ -47,34 +47,32 @@ function wpcom_vip_should_force_two_factor() {
...
-	if ( function_exists( '\HumanMade\SimpleSaml\instance' ) && \HumanMade\SimpleSaml\instance() ) {
+	// Allow custom SSO solutions
+	if ( wpcom_vip_use_custom_sso() ) {
 		return false;
 	}
 
 	return true;
 }
 
-function wpcom_vip_is_jetpack_authorize_request() {
```
joshbetz: Accidentally deleted this function?

justnorris: 🤔

justnorris: I have no idea why it's red. That function is still there.

justnorris: https://github.com/Automattic/vip-go-mu-plugins/blob/5a5059e1a110a7d41dcecbe4280afe7e377e9f29/two-factor.php#L83-L90

joshbetz: I don't think so https://github.com/Automattic/vip-go-mu-plugins/blob/5a5059e1a110a7d41dcecbe4280afe7e377e9f29/two-factor.php

joshbetz: @justnorris that's a different function. `wpcom_vip_is_jetpack_authorize_request` vs `wpcom_vip_is_jetpack_authorize_rest_request`.

justnorris: Fixed

joshbetz:
r696-vip-go-stacks

PR #1345, opened 2019-11-21
Accept multisite paths with no `site` sub-directory

hanifn:
## Description

`Path_Utils::trim_leading_multisite_directory()` will return `false` for file paths that doesn't include the `/sites/#` sub-directory. This causes problems for multisite sites that are running their default site without the `/sites/#` sub-directory as their files are uploaded to the wrong directory.

## Checklist

### DIFF
--- a/files/class-path-utils.php
+++ b/files/class-path-utils.php
@@ -4,13 +4,13 @@ namespace Automattic\VIP\Files;
 
 class Path_Utils {
 	public static function is_subdirectory_multisite_path( $file_path, $uploads_path ) {
-		$pattern = '#^/[_0-9a-zA-Z-]+/' . $uploads_path . '/sites/[0-9]+/#';
+		$pattern = '#^/[_0-9a-zA-Z-]+/' . $uploads_path . '/#';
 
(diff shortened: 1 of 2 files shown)


darthhexx:
@hanifn this is fine from SysEng's side, because we don't rely on the `sites/[0-9]+/` being present in our edge regex substitutions anymore either. I can't approve though, since I'm not on your team apparently :)

darthhexx:
Also, I noticed the one checkbox isn't ticked, it would be a good idea to test on a sandbox too if you haven't already.

mjangda:
> causes problems for multisite sites that are running their default site without the /sites/# sub-directory as their files are uploaded to the wrong directory.

To clarify, the main site typically does not have the `/sites/#/` in their path and this issue crops up when the main site is hosted in a subdirectory (e.g. `example.com/mysite` for which the a sample upload path would be `example.com/mysite/wp-content/uploads/file.jpg` but would fail the checks in `trim_leading_multisite_directory`).

hanifn:
Thanks for the clarification @mjangda 👍🏽

hanifn:
r680-stacks

PR #1346, opened 2019-11-21
Skip externals for test repo

mjangda:
The only external included is the wordpress importer, which we don't really need. It's also currently breaking on CircleCI due to rate limits.

## Steps to Test

Make sure the tests on Circle work :)


nickdaugherty:
💯 💯 💯

PR #1349, opened 2019-11-21
Add upper and lower limits to JP sync queue settings

nickdaugherty:
## Description

VIP sites need higher defaults for the `max_queue_size` and `max_queue_lag` Jetpack Sync settings - this adds two option filters to ensure that the settings are within a predefined range, while still allowing the exact setting to be tweaked on a per-site basis (compared to a `pre_option` filter).

## Checklist

## Steps to Test

1. Check a site's settings - `wp option get jetpack_sync_settings_max_queue_size`
1. Should be default (`1000`)
1. Check out PR.
1. Check a site's settings - `wp option get jetpack_sync_settings_max_queue_size`
1. Should now be the minimum size allowed - `10000`


vip-jetpack/vip-jetpack.php
```diff
@@ -9,6 +9,12 @@
  * License: GPL2+
  */
 
+define( 'VIP_GO_JETPACK_SYNC_MAX_QUEUE_SIZE_LOWER_LIMIT', 10000 );
+define( 'VIP_GO_JETPACK_SYNC_MAX_QUEUE_SIZE_UPPER_LIMIT', 50000 );
+
+define( 'VIP_GO_JETPACK_SYNC_MAX_QUEUE_LAG_LOWER_LIMIT', 2 * HOUR_IN_SECONDS );
+define( 'VIP_GO_JETPACK_SYNC_MAX_QUEUE_LAG_UPPER_LIMIT', 6 * HOUR_IN_SECONDS );
```
mjangda: Should we wrap these in a `defined` check? Would allow customizing on a per-site basis, if needed.

nickdaugherty: I see these more as hard limits - the sites can still be customized via the option, but these are the "bumpers" that sites can't go above or below (too big or too small can break the site). We can allow changing these though if you think it'll be valuable.

mjangda: Oh, you're right :)

I thought the goal here was to set higher defaults across VIP, not just better limits? Or is that coming as a separate PR?

mjangda: Never mind -- my brain is clearly not working. I see the min bumps those up :)

Let's add some comments to the constant definitions noting why they're set to those specific values.

nickdaugherty: 👍 added explanatory comments to the constants in da6da27f

vip-jetpack/vip-jetpack.php
```diff
@@ -9,6 +9,34 @@
...
+ * headroom as they tend to publish more than average
+ */
+define( 'VIP_GO_JETPACK_SYNC_MAX_QUEUE_SIZE_LOWER_LIMIT', 10000 );
+
+/**
+ * The largest incremental sync queue size allowed - items will not get enqueued if there are already this many pending
+ * 
+ * The queue is stored in the option table, so if the queue gets _too_ large, site performance suffers
+ */
+define( 'VIP_GO_JETPACK_SYNC_MAX_QUEUE_SIZE_UPPER_LIMIT', 50000 );
```
nickdaugherty: @lezama / @mdbitz what do you think about this upper limit? Should it be higher, perhaps 100k?

vip-jetpack/vip-jetpack.php
```diff
@@ -9,6 +9,34 @@
...
+ * The default is 15 minutes, but VIP sites often have more busy queues and we must prevent dropping items if the sync is
+ * running behind
+ */
+define( 'VIP_GO_JETPACK_SYNC_MAX_QUEUE_LAG_LOWER_LIMIT', 2 * HOUR_IN_SECONDS );
+
+/**
+ * The maximum incremental sync queue lag allowed - just sets a reasonable upper bound on this limit to prevent extremely
+ * stale incremental sync queues
+ */
+define( 'VIP_GO_JETPACK_SYNC_MAX_QUEUE_LAG_UPPER_LIMIT', 6 * HOUR_IN_SECONDS );
```
nickdaugherty: @lezama / @mdbitz same question - is 6 hours too conservative for a max queue lag setting?

nickdaugherty:
Going to merge / deploy - we can fine-tune the exact values later if needed.

nickdaugherty:
r683-stacks

vip-jetpack/vip-jetpack.php
lezama: I guess it depends what we have promised to our customers. On a healthy site the queue should never be bigger than a dozen actions. If they reach this numbers it is probably because something else is wrong, failing in wpcom or in the communication channel. We need to keep this big just for when there's a situation where we don't want to lose this actions while we fix things.

lezama: It also depends on how much the size of the queue affects  the site.
We need to remember two things:
- The queue gets stored as blog options
- Both conditions have to be met in order to stop adding to the queue. So the queue could have more than `VIP_GO_JETPACK_SYNC_MAX_QUEUE_SIZE_UPPER_LIMIT` if the lag hasn't reached `VIP_GO_JETPACK_SYNC_MAX_QUEUE_LAG_UPPER_LIMIT`

nickdaugherty: > Both conditions have to be met in order to stop adding to the queue

Ah - in that case, I may have made the limits _too high_ in https://github.com/Automattic/vip-go-mu-plugins/pull/1352

PR #1350, opened 2019-11-22
Update wp-memcached submodule.

jenkoian:
## Description

The object-cache drop in hasn't been updated for a while and there's
been an important fix for the `$found` parameter added.

This is particularly important for some plugins, e.g. MLP3 depends on this functionality and performance declines without it.

## Checklist

## Steps to Test

Example:

1. Install MLP3
2. Notice a number of queries constantly being run on the homepage
3. Update the submodule to this PR
4. The queries should now be gone as the cache should be getting hit.


nickdaugherty:
Thanks for the PR @jenkoian!

The issue with `$found` is not fully fixed upstream in `wp-memcached` - after merging the original PR (https://github.com/Automattic/wp-memcached/pull/34), I identified an unresolved bug in the implementation, which I documented here:

https://github.com/Automattic/wp-memcached/issues/40

So, that still needs to be addressed before we can update to the latest, or we'll likely cause more problems than we solve.

jenkoian:
Thanks for the response @nickdaugherty!

This is quite frustrating. The contract of the original methods (https://developer.wordpress.org/reference/functions/wp_cache_get/) and (https://developer.wordpress.org/reference/classes/wp_object_cache/get/) says there should be a 4th parameter of `$found` so I can understand why MLP3 are using this. (unfortunately this version of their plugin is not open source so I cannot link to it).

I can open a support ticket with them, but in general I would have expected that the drop in would support the contract of the documented function.

Is there any progress being made on this issue do you know? I'm happy to take a look too after the weekend. 

Thanks!

jenkoian:
Just for reference, hopefully a fix for https://github.com/Automattic/wp-memcached/issues/40 in this PR: https://github.com/Automattic/wp-memcached/pull/43

jenkoian:
@nickdaugherty just updated this one too 👍

rinatkhaziev:
Superseded by #2006

PR #1355, opened 2019-11-26
Improve stream wrapper cache memory efficiency

hanifn:
## Description

The current implementation of caching for the Stream Wrapper have issues with `Out of Memory` errors when uploading large files. This is due to the fact the fact the caching works by creating local, temporary copies of the file which can lead to PHP running out of memory when uploading large files.

## Checklist

## Steps to Test

Example:

1. Check out PR.
1. Upload a file that's close to the PHP `upload_max_filesize` limit
1. Ensure file is properly uploaded.


hanifn:
r692-stacks

PR #1357, opened 2019-11-28
Add conditional to check for non-production domains based on HTTP_HOST to properly set X-Robots-Tag for non-prod domains

rinatkhaziev:
It has been discovered that at least in one case a non-prod domain didn't have a proper robots.txt blocking bots. This commit is intended as a fallback for situations like that.

### DIFF
--- a/000-vip-init.php
+++ b/000-vip-init.php
@@ -180,7 +180,7 @@ add_filter( 'wp_headers', function( $headers ) {
 	$headers['X-Powered-By'] = 'WordPress.com VIP <https://wpvip.com>';
 
 	// All non-production domains should not be indexed.
-	if ( 'production' !== VIP_GO_ENV || false !== stristr( $_SERVER[ 'HTTP_HOST' ], '.go-vip.' ) ) {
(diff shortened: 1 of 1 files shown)


000-vip-init.php
```diff
@@ -180,8 +180,7 @@
 	$headers['X-Powered-By'] = 'WordPress.com VIP <https://wpvip.com>';
 
 	// All non-production domains should not be indexed.
```
```suggestion
	// Non-production applications and go-vip.(co|net) domains should not be indexed.
```

000-vip-init.php
```diff
@@ -180,8 +180,7 @@
...
-	// This should not apply only to *.vip-go.co
-	if ( 'production' !== VIP_GO_ENV ) {
+	if ( 'production' !== VIP_GO_ENV || false !== stristr( $_SERVER[ 'HTTP_HOST' ], '.go-vip.' ) ) {
```
mjangda: `strpos` is generally preferred over `strstr`, especially when we only care about checking if the string has something:

```suggestion
	if ( 'production' !== VIP_GO_ENV || false !== strpos( $_SERVER[ 'HTTP_HOST' ], '.go-vip.' ) ) {
```

mjangda:
r689-stacks

PR #1358, opened 2019-11-28
Add back-compat for window.$

mjangda:
Copies over a patch from WP.com which restores `window.$` which was being leaked by the mediaelement script. WP 5.3 "fixed" that leak but there are many instances of code that still rely on that leaked global.

This will help prevent possible issues until we get all those instances fixed.

props @david-binda 

## Checklist

## Steps to Test

1. Add the following snippet to a 5.3 site:

```
add_action( 'admin_print_footer_scripts', function() {
    ?>
    <script>console.log( $.fn.jquery );</script>
    <?php
}, PHP_INT_MAX );
```

2. Load the post edit page (with classic editor). Console should show an error (related to $ being undefined)

3. Apply the patch.

4. Error should go away.

### DIFF
--- a/back-compat.php
+++ b/back-compat.php
@@ -1,0 +1,21 @@
+<?php
+
+/**
+ * Plugin Name: VIP Back-compat
+ * Description: Adds relevant backwards compatibility code.
+ * Author: Automattic
+ * Version: 1.0
+ * License: GPL version 2 or later - http://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+ */
+
+/*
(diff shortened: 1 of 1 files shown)


mjangda:
Tested locally with 5.2 and 5.3. On 5.3, without patch leads to this error: `Uncaught TypeError: Cannot read property 'fn' of undefined`. With patch, the error goes away.

Also tested on Go environment running 5.3 with Edit Flow 0.9 (which has this bug). With the patch, the error goes away.

mjangda:
r690-deploy

PR #1359, opened 2019-11-28
Akismet 4.1.3

mjangda:
- Prevented an attacker from being able to cause a user to unknowingly recheck their Pending comments for spam.
- Improved compatibility with Jetpack 7.7+.
- Updated the plugin activation page to use consistent language and markup.
- Redirecting users to the Akismet connection/settings screen upon plugin activation, in an effort to make it easier for people to get setup.


mjangda:
r735-stacks

PR #1360, opened 2019-11-28
WP-CLI: When `url` not set, force the main site

mjangda:
We routinely run into cases where a mismatch in the wp-config and the multisite config leads to failures on WP_CLI (which can make debugging, fixing the issue, and other work challenging).

This forces loads the "main site" based on the IDs specified in config when the url flag is not set. This way we avoid "site not found" errors when a site is in a mismatched state and allows WP_CLI to continue working.

TODO:
- [ ] Need to add unit tests for this.
- [ ] Need to fix issues with invalid sites.

## Checklist

## Steps to Test

1. Get a site into a mismatched state: create a multisite install and then change the URL in `wp_blogs` table to a different domain.
2. Any WP_CLI calls should now fail with a "Site Not Found" error

```
$ wp option get home
Error: Site 'also-multisite.local/' not found. Verify DOMAIN_CURRENT_SITE matches an existing site or use `--url=<url>` to override.
```

3. Apply the patch.

4. Try again and it should work:

```
$ wp option get home
http://multisite.local
```

5. Try again with the url set to the main site (should work):

```
$ wp option get home --url=multisite.local
http://multisite.local
```

6. Try again with a valid subsite (should also work):

```
$ wp option get home --url=multisite.local/test/
http://multisite.local/test
```

7. Try again with an invalid site (should fail):

```
$ wp option get home --url=multisite.local/not-a-site/
<SHOULD BE ERROR BUT CURRENTLY BROKEN>
```


mjangda:
@nickdaugherty let me know what you think of this approach. Should help fix a lot of the failures we see so sites that are in the middle of migration.

Separately, we probably need monitoring to catch these sites as well.

mjangda:
Going to close this out. It's bound to cause more problems and better to fix the root of the problem(s) instead.

PR #1364, opened 2019-11-29
VaultPress: Allow local instance

GaryJones:
## Description

If there is a VaultPress install in the `client-mu-plugins` directory, use that instead of the version in vip-go-mu-plugins.

## Checklist

## Steps to Test

Example:

1. Add a die statement to `client-mu-plugins/vaultpress/vaultpress.php`
1. Load a WP instance.
1. Apply the PR
1. Reload a WP instance and get the die statement to show this is working.
1. Can add a full VaultPress install to `client-mu-plugins/`, and reload to check the dashboard page is still accessible.

### DIFF
--- a/vaultpress.php
+++ b/vaultpress.php
@@ -17,4 +17,14 @@ if ( ! defined( 'VAULTPRESS_TIMEOUT' ) ) {
 	define( 'VAULTPRESS_TIMEOUT', 10 );
 }
 
-require_once( __DIR__ . '/vaultpress/vaultpress.php' );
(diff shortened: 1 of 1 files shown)


GaryJones:
The use case for this was to allow some temporary debugging code to be added to a version of VaultPress which could then be added to a client site, but we can add it directly to the vip-go-mu-plugin version directly. 

As such, this PR isn't needed, but it helps to document how it would be done in if really needed.

The drawback of having this though, is that it can get messy if clients don't update their local versions, and that can lead to more bugs, etc.

PR #1365, opened 2019-12-02
Updated Max Queue Settings to 50,000 queue size and 8 hour lag.

mdbitz:
## Description

Reducing max Jetpack Settings to account for recent performance updates.

## Checklist

## Steps to Test

1. Apply Patch
1. Manually set settings higher than thresholds in Jetpack Debugger
1. Verify settings reset to max limit
1. Manually set settings lower than thresholds in Jetpack Debugger
1. Verify settings reset to min limit

## Follow-up Items

The current tests have settings related to earlier limits. Need to review to get them aligned with final settings that we want.


mjangda:
> Reducing max Jetpack Settings to account for recent performance updates.

The code increases the queue size but decreases the lag time, which seems counter to the PR description. Also, can you share some details about the performance updates just so we have more context available for the future?

nickdaugherty:
@mdbitz code looks good but the tests will need to be updated to account for the new values.

brandon-m-skinner:
@mdbitz Is this still valid? I have no problems taking over this one and getting it live if it's still good.

mdbitz:
I'd say we can ax this. The limits on VIP sites are quite high but they actually protected metro from data loss 2 JP versions ago.

It'll be up to the VIP team to determine if they want to make further changes to settings as future changes to Sync roll-out.

PR #1366, opened 2019-12-02
Increase jp sync throughput on cron

nickdaugherty:
## Description

Increases the throughput of JP Sync when running via cron by:

* Allowing the cron job to run for 4 minutes (default is 30s)
* Reducing the wait time between batches from 10s to 1s

The cron job for incremental syncing runs every 5 minutes, but by default is only allowed to run for 30s. This is much too short for VIP sites, so this PR increases the limit to 4 minutes, to allow more items to be processed per cron event, while leaving a small buffer between completion and the start of the next event.

The `sync_wait_time` change affects how long cron waits in between batches of content before restarting, in an effort to reduce performance impacts. By default this is 10 seconds, but VIP can be more aggressive and doesn't need to wait as long (we'll still wait a small amount).

## Checklist

## Steps to Test

Example:

1. Check out PR.
1. Go to `wp shell`
1. `Automattic\Jetpack\Sync\Settings::get_setting( 'cron_sync_time_limit' );`
1. Should be 240 (4 minutes)
1. `Automattic\Jetpack\Sync\Settings::get_setting( 'sync_wait_time' );`
1. Should be 1

### DIFF
--- a/vip-jetpack/vip-jetpack.php
+++ b/vip-jetpack/vip-jetpack.php
@@ -100,6 +100,15 @@ add_filter( 'option_jetpack_sync_settings_cron_sync_time_limit', function( $valu
 	return 4 * MINUTE_IN_SECONDS;
 }, 9999 );
 
+/**
+ * Reduce the time between sync batches on VIP
+ * 
+ * Default is 10 seconds
+ */
+add_filter( 'option_jetpack_sync_settings_sync_wait_time', function( $value ) {
+	return 1;
+}, 9999 );
+
 // Prevent Jetpack version ping-pong when a sandbox has an old version of stacks
 if ( true === WPCOM_SANDBOXED ) {
 	add_action( 'updating_jetpack_version', function( $new_version, $old_version ) {
--- a/vip-jetpack/vip-jetpack.php
+++ b/vip-jetpack/vip-jetpack.php
@@ -106,7 +106,6 @@ add_filter( 'option_jetpack_sync_settings_cron_sync_time_limit', function( $valu
  *
  * By default, this is 10 seconds, but VIP can be more aggressive and doesn't need to wait as long (we'll still wait a small amount).
  * 
- * Default is 10 seconds
  */
 add_filter( 'option_jetpack_sync_settings_sync_wait_time', function( $value ) {
 	return 1;
--- a/vip-jetpack/vip-jetpack.php
+++ b/vip-jetpack/vip-jetpack.php
@@ -91,6 +91,15 @@ add_filter( 'option_jetpack_sync_settings_max_queue_lag', function( $value ) {
 	return $value;
 }, 9999 );
 
+/**
+ * Allow incremental syncing via cron to take longer than the default 30 seconds
+ * 
+ * The default interval for this cronjob is 5 minutes
+ */
+add_filter( 'option_jetpack_sync_settings_cron_sync_time_limit', function( $value ) {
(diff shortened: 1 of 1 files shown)


mdbitz:
@nickdaugherty , quick question for the other settings we allow ranges. I wanted to understand why we are going with a strict setting in these options cases. Based on our conversations I think these values are where we want to start at just wondering why the different approach.

vip-jetpack/vip-jetpack.php
```diff
@@ -91,6 +91,24 @@
 	return $value;
 }, 9999 );
 
+/**
+ * Allow incremental syncing via cron to take longer than the default 30 seconds
+ * 
+ * The default interval for this cronjob is 5 minutes
```
mjangda:
```suggestion
```

vip-jetpack/vip-jetpack.php
```diff
@@ -91,6 +91,24 @@
 	return $value;
 }, 9999 );
 
+/**
+ * Allow incremental syncing via cron to take longer than the default 30 seconds
```
mjangda:
```suggestion
 * Allow incremental syncing via cron to take longer than the default 30 seconds.
 *
 * This will allow more items to be processed per cron event, while leaving a small buffer between completion and the start of the next event (the event interval is 5 mins).
```

vip-jetpack/vip-jetpack.php
```diff
@@ -91,6 +91,24 @@
...
+ * 
+ * The default interval for this cronjob is 5 minutes
+ */
+add_filter( 'option_jetpack_sync_settings_cron_sync_time_limit', function( $value ) {
+	return 4 * MINUTE_IN_SECONDS;
+}, 9999 );
+
+/**
+ * Reduce the time between sync batches on VIP
```
mjangda:
```suggestion
 * Reduce the time between sync batches on VIP for performance gains.
 *
 * By default, this is 10 seconds, but VIP can be more aggressive and doesn't need to wait as long (we'll still wait a small amount).
```

vip-jetpack/vip-jetpack.php
```diff
@@ -91,6 +91,24 @@
...
+ * 
+ * Default is 10 seconds
```
mjangda:
```suggestion
```

nickdaugherty:
> just wondering why the different approach.

Good question - for these, there isn't much benefit to be gained from other values so it's less likely to be overridden (we're basically trying to max it out). They're still overrideable to anything, we'd just need a later priority filter.

The downsides of going outside of an acceptable range here is also smaller - cron will run differently, but we're less likely to have _lost_ data or major performance problems compared to when the queue limits are changed.

nickdaugherty:
These updates will be in a future version of Jetpack core, but we should still merge this for the meantime and for fine-grained control on VIP.

joshbetz:
Is this 4 minutes instead of 5 minutes needed to give "breathing room" between runs?

nickdaugherty:
> Is this 4 minutes instead of 5 minutes needed to give "breathing room" between runs?

That's right, b/c Jetpack doesn't terminate the process strictly at 4 minutes, or 30s, or whatever the time limit is - it lets the current batch finish, then checks to see if that time limit has been exceeded. So in reality, a configured value of 4 minutes will always run longer than 4 minutes.

nickdaugherty:
r693-stacks

PR #1367, opened 2019-12-03
VaultPress: Add extra debug info

GaryJones:
Amend the value that is already logged when pinging a site from VaultPress.com fails in a specific way (signature mismatch).

This output can be seen in the “ping test” panel in VPMC.

This is needed to debug why a live client production site isn't able to connect to VaultPress.


nickdaugherty:
@GaryJones is this PR still valuable?

GaryJones:
@nickdaugherty I no longer need it for a specific customer, but it could be useful for general logging for VaultPress issues.

WPprodigy:
Is there are particular piece of `$_POST` that could be logged rather than the whole thing?

Otherwise, I think maybe it's okay to go ahead and close this one out.

GaryJones:
Closing as not needed.

PR #1371, opened 2019-12-05
Disable concat for react and react-dom

joshbetz:
## Description

Disables concatenation for the two core javascript files that use strict mode.

Noting that this is not a complete solution as any plugin could also use strict mode. It does fix the cases we've seen so far though. We'll use this to fix the immediate issue and follow up with a more robust solution.

> This syntax [use strict] has a trap that has already bitten a major site: it isn’t possible to blindly concatenate conflicting scripts. Consider concatenating a strict mode script with a non-strict mode script: the entire concatenation looks strict! The inverse is also true: non-strict plus strict looks non-strict.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode

## Checklist

## Steps to Test

1. Go to new post screen and confirm undefined reference errors in the console.
1. Check out this PR.
1. Confirm javascript errors are resolved.

If you're having trouble reproducing the issue, make sure HTTP Concat is enabled in your test environment. You'll also need to be using a plugin that's affected by this. I can help find test cases if necessary.

### DIFF
--- a/http-concat.php
+++ b/http-concat.php
@@ -23,6 +23,14 @@ if ( true === WPCOM_IS_VIP_ENV ) {
 			if ( 'wp-tinymce-root' === $handle || 'wp-tinymce' === $handle ) {
 				$do_concat = false;
 			}
+			
+			// Don't concat core scripts that use strict mode
+			// We can't blindly concatenate strict and non-strict scripts because
+			// strict mode applies to the whole file and changes functionality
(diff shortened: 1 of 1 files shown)


dchymko:
> Should we put this in the concat plugin rather than mu-plugins?

Yeah, probably makes sense to have all the js logic in one place here?
https://github.com/Automattic/nginx-http-concat/blob/master/jsconcat.php#L65

PR #1372, opened 2019-12-06
Bump http-concat to disable concat with strict mode

joshbetz:
https://github.com/Automattic/nginx-http-concat/pull/54


joshbetz:
r694-vip-go-stacks

PR #1379, opened 2019-12-14
Error page: Use descriptive code for alloptions

### LINKED ISSUE #1210: Alloptions: Better error code when size exceeds limit
When the alloptions size exceeds the defined limit (1MB), we display an error page with a `1024` code. Let's make that code more human friendly so it's easier to identify the actual problem...`ALLOPTIONS_SIZE_LIMIT_EXCEEDED` or similar.

GaryJones:
## Description

When the alloptions size exceeds the defined limit (1MB), we displayed an error page with a `1024` code. We can make that code more human-friendly so it's easier to identify the actual problem.

Fixes #1210

## Checklist

## Steps to Test

No real testing needed (unless there are e2e tests hiding somewhere) - it's just a change in the visible error code on an error page.


mjangda:
Background on the error code: we intentionally left it not-obvious, since end users are the ones primarily seeing the error page. An error like `ALLOPTIONS_SIZE_LIMIT_EXCEEDED` can be panic inducing especially considering they have no control over it.

There's probably a middle ground we can find like `Error 1024 (alloptions)` instead?

errors/alloptions-limit.html
```diff
@@ -146,7 +146,7 @@
 				<h1>Site Unavailable</h1>
 				<p>This site is temporarily unavailable, please try again in a few moments.</p>
 				<footer>
-					<span class="error">Error 1024</span>
+					<span class="error">Error <code>ALLOPTIONS_SIZE_LIMIT_EXCEEDED</code></span>
```
GaryJones:
```suggestion
					<span class="error">Error 1024 (<code>alloptions</code>)</span>
```

GaryJones:
Background makes sense - message updated.

mjangda:
r858-stacks

PR #1381, opened 2019-12-14
wpcom_vip_maybe_skip_old_slug_redirect(): Improve compatibility with local setups

### LINKED ISSUE #976: "Undefined index" warning in vip-caching.php
The function `wpcom_vip_maybe_skip_old_slug_redirect` (https://github.com/Automattic/vip-go-mu-plugins/blob/master/vip-helpers/vip-caching.php#L712-L720) in our local environment produces a warning: `Undefined index: DOCUMENT_URI`.

In our systems, the server variable `DOCUMENT_URI` is not defined, we have the same value in `REQUEST_URI`, instead.

This might not be an issue on VIP servers, but should be took into account for local environements. This is filling our local logs as it happens at every frontend request.

Considering that min PHP version is now 7.2, I would suggest something like this:

```php
$uri = $_SERVER['DOCUMENT_URI'] ?? $_SERVER['REQUEST_URI'] ?? null;

// We look to see if a malformed url (represented by 'http:' ) is right after the starting / in DOCUMENT_URI hence position 1
if ( is_404() && ( 1 === strpos( $uri, 'http:' ) || 1 === strpos( $uri, 'https:' ) ) ) {
```

which is both readable and fix the issue.

An alternative could be using `add_query_arg()` function so avoid direct access to `$_SERVER` values at all.

It worth nothing that `add_query_arg()` uses `REQUEST_URI` not `DOCUMENT_URI`.

GaryJones:
## Description

Add support for local development servers that do not have `$_SERVER['DOCUMENT_URI']` defined.

Fixes #976.

_Needs actual testing._

## Checklist

## Steps to Test

1. Create a setup where `$_SERVER['DOCUMENT_URI']` is undefined.
2. Check that function does what it was still meant to, once PR is applied.

### DIFF
--- a/vip-helpers/vip-caching.php
+++ b/vip-helpers/vip-caching.php
@@ -707,18 +707,22 @@ function wpcom_vip_flush_wp_old_slug_redirect_cache( $post_id, $post, $post_befo
 }
 
 /**
- * We're seeing an increase of urls that match this pattern: http://example.com/http://othersite.com/random_text
- * These then cause really slow lookups inside of wp_old_slug_redirect, since wp_old_slug redirect does not match on full urls but rather former slugs it's safe to skip the lookup for these. (Most of the calls are from bad ad providers that generate random urls)
+ * Potentially skip redirect for old slugs.
+ *
+ * We're seeing an increase of URLs that match this pattern: http://example.com/http://othersite.com/random_text.
+ *
+ * These then cause really slow lookups inside of wp_old_slug_redirect. Since wp_old_slug redirect does not match
+ * on full URLs but rather former slugs, it's safe to skip the lookup for these. Most of the calls are from bad ad
+ * providers that generate random URLs.
  */
-function wpcom_vip_maybe_skip_old_slug_redirect(){
-
-	//We look to see if a malformed url (represented by 'http:' ) is right after the starting / in DOCUMENT_URI hence position 1
-	if ( is_404() && ( 1 === strpos( $_SERVER['DOCUMENT_URI'], 'http:' ) || 1 === strpos( $_SERVER['DOCUMENT_URI'], 'https:' ) ) ) {
+function wpcom_vip_maybe_skip_old_slug_redirect() {
+	if ( is_404() && ( 0 === strpos( $_SERVER['REQUEST_URI'], '/http:' ) || 0 === strpos( $_SERVER['REQUEST_URI'], '/https:' ) ) ) {
 		remove_action( 'template_redirect', 'wp_old_slug_redirect' );
 		remove_action( 'template_redirect', 'wpcom_vip_wp_old_slug_redirect', 8 );
 	}
 
 }
+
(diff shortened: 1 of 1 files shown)


vip-helpers/vip-caching.php
```diff
@@ -710,15 +710,17 @@ function wpcom_vip_flush_wp_old_slug_redirect_cache( $post_id, $post, $post_befo
  * We're seeing an increase of urls that match this pattern: http://example.com/http://othersite.com/random_text
  * These then cause really slow lookups inside of wp_old_slug_redirect, since wp_old_slug redirect does not match on full urls but rather former slugs it's safe to skip the lookup for these. (Most of the calls are from bad ad providers that generate random urls)
  */
-function wpcom_vip_maybe_skip_old_slug_redirect(){
+function wpcom_vip_maybe_skip_old_slug_redirect() {
+	$uri = $_SERVER['DOCUMENT_URI'] ?? $_SERVER['REQUEST_URI'] ?? '';
 
 	//We look to see if a malformed url (represented by 'http:' ) is right after the starting / in DOCUMENT_URI hence position 1
-	if ( is_404() && ( 1 === strpos( $_SERVER['DOCUMENT_URI'], 'http:' ) || 1 === strpos( $_SERVER['DOCUMENT_URI'], 'https:' ) ) ) {
+	if ( is_404() && ( 1 === strpos( $uri, 'http:' ) || 1 === strpos( $uri, 'https:' ) ) ) {
```
brandon-m-skinner, line 717: Since you're comparing to e.g.: http://othersite.com/random_text and looking to match the start of the string, this should be comparisons to 0 since string indexes start at 0.
```suggestion
	if ( is_404() && ( 0 === strpos( $uri, 'http:' ) || 0 === strpos( $uri, 'https:' ) ) ) {
```

GaryJones: @brandon-m-skinner How does that match up with the comment then?

```
//We look to see if a malformed url (represented by 'http:' ) is right after the starting / in DOCUMENT_URI hence position 1
```

As per the function DocBlock, what we're looking for is an invalid URL e.g. `http://example.com/http://othersite.com/random_text`, which means the DOCUMENT_URI and REQUEST_URI would be `/http:...`, so looking for `http:` or `https:` starting at index 1.

We could just start at index 0 and look for `/http:` or `/https:` if that makes it easier to understand? (I think it would be)

Also, do we really need to check for DOCUMENT_URI if REQUEST_URI seems to be the most popular key to check? Compare [instances of DOCUMENT_URI](https://github.com/Automattic/vip-go-mu-plugins/search?q=DOCUMENT_URI) to [instances of REQUEST_URI](https://github.com/Automattic/vip-go-mu-plugins/search?q=REQUEST_URI) just in this repo.

GaryJones:
Re-jigged, and rebased on to latest `master`.

brandon-m-skinner:
r961-stacks

PR #1384, opened 2019-12-17
Add trailing slash to make valid upload path

emrikol:
## Description

Some sites may not use the year/month upload scheme, and the current `check_filetype_with_backend()` will make all uploads fail with this warning:

```
PHP Warning:  The specified file path (`wp-content/uploadsfilename.jpg`) does not begin with `/wp-content/uploads/`. #vip-go-streams in /var/www/wp-content/mu-plugins/files/class-vip-filesystem.php on line 211
```

Unit tests are already written properly at https://github.com/Automattic/vip-go-mu-plugins/blob/e09535fd8b56577479b96033fee9d195f717752a/tests/files/test-api-client.php#L71 so there shouldn't need to be any changes there :)

## Checklist

## Steps to Test

1. Check out PR.
1. Add `add_filter( 'pre_option_uploads_use_yearmonth_folders', function() {return '0';}, 9999 );` to a site.
1. Upload a media file via the admin.
1. Verify upload is correct.

### DIFF
--- a/files/class-vip-filesystem.php
+++ b/files/class-vip-filesystem.php
@@ -197,7 +197,7 @@ class VIP_Filesystem {
 	 * @return  bool        True if filetype is supported. Else false
 	 */
 	protected function check_filetype_with_backend( $filename ) {
-		$upload_path = $this->get_upload_path();
+		$upload_path = trailingslashit( $this->get_upload_path() );
 
 		$file_path = $upload_path . $filename;
 


mjangda:
Thanks for tracking this down and fixing it @emrikol!

> Some sites may not use the year/month upload scheme, and the current check_filetype_with_backend() will make all uploads fail with this warning

For added context (for our future selves), the bug was occurring regardless of the upload folder scheme. However, the year/month case got lucky and validated just fine because the path we checked was `wp-content/uploads/yy/mm`.

joshbetz:
r708-vip-go-stacks

PR #1385, opened 2019-12-17
Makes concat_js/css logic easier to read

joshbetz:
Undefined or equals 'yes' isn't consistent with how `concat_js` and `concat_js` are used in practice, which makes the original hard to read. The only time this is used is for `concat_js=no`, so looking for 'no' equals concat_js makes more sense.

### DIFF
--- a/http-concat.php
+++ b/http-concat.php
@@ -10,7 +10,7 @@
 // We don't want concat running outside VIP Go environments.
 if ( true === WPCOM_IS_VIP_ENV ) {
 	// Activate concatenation
-	if ( ! isset( $_GET['concat_js'] ) || 'yes' === $_GET['concat_js'] ) {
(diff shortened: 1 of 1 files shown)


http-concat.php
```diff
@@ -10,7 +10,7 @@
 // We don't want concat running outside VIP Go environments.
 if ( true === WPCOM_IS_VIP_ENV ) {
 	// Activate concatenation
-	if ( ! isset( $_GET['concat_js'] ) || 'yes' === $_GET['concat_js'] ) {
+	if ( ! isset( $_GET['concat_js'] ) || 'no' !== $_GET['concat_js'] ) {
```
mjangda: A better option would be to move this to a function which we can then more easily test.

PR #1386, opened 2019-12-19
Add conditional around add_filter() to prevent PHP fatal

nickdaugherty:
## Description

Older versions of ElasticPress only pass 2 args to the `ep_index_name` filter which triggers a PHP fatal, so to avoid that, only add the filter when the constant is defined

## Checklist

Example:

1. Check out PR.
1. Install ElasticPresss 2.1
1. Ensure no "too few arguments" fatals are thrown

### DIFF
--- a/tests/vip-helpers/test-vip-elasticsearch.php
+++ b/tests/vip-helpers/test-vip-elasticsearch.php
@@ -12,6 +12,9 @@ class VIP_ElasticSearch_Test extends \WP_UnitTestCase {
 
 		define( 'USE_VIP_ELASTICSEARCH', true );
 
+		// Hack to get around the constant not being defined early enough...there is probably a proper PHPUnit way to do that
+		add_filter( 'ep_index_name', 'vip_elasticsearch_filter_ep_index_name', PHP_INT_MAX, 3 );
+
 		$index_name = apply_filters( 'ep_index_name', 'index-name', 1, $mock_indexable );
 
 		$this->assertEquals( 'vip-123-slug-1', $index_name );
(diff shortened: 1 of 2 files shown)


nickdaugherty:
We may want to rethink where we're putting this filter...it really should only run if the client has activated the plugin (maybe a shared plugin) and the constant is set.

tests/vip-helpers/test-vip-elasticsearch.php
```diff
@@ -12,6 +12,9 @@ public function test__vip_elasticsearch_filter_ep_index_name() {
 
 		define( 'USE_VIP_ELASTICSEARCH', true );
 
+		// Hack to get around the constant not being defined early enough...there is probably a proper PHPUnit way to do that
```
mjangda: There's no clean way to do this since the constant is defined in the bootstrap. We'd have to remove it from there to be able to get more control around these tests.

vip-helpers/vip-elasticsearch.php
```diff
@@ -463,4 +463,8 @@ function vip_elasticsearch_filter_ep_index_name( $index_name, $blog_id, $indexab
 
 	return $index_name;
 }
-add_filter( 'ep_index_name', 'vip_elasticsearch_filter_ep_index_name', PHP_INT_MAX, 3 ); // We want to enforce the naming, so run this really late.
+
+// Only add filter when using VIP Elasticsearch, b/c older versions of ElasticPress only pass 2 args to this which is a PHP warning
```
nickdaugherty:
```suggestion
// Only add filter when using VIP Elasticsearch, b/c older versions of ElasticPress only pass 2 args to this which is a PHP fatal
```

PR #1387, opened 2019-12-20
Add default EP_SYNC_CHUNK_LIMIT constant for bulk syncing

nickdaugherty:
## Description

Adds a default `EP_SYNC_CHUNK_LIMIT` value for ElasticPress bulk syncing, when using VIP Elasticsearch.

## Checklist

## Steps to Test

1. Check out PR.
1. `define( 'USE_VIP_ELASTICSEARCH', true );` in site code
1. `echo EP_SYNC_CHUNK_LIMIT` - should be 250 (default)
1. `define( 'EP_SYNC_CHUNK_LIMIT', 500 );` - should not cause any problems

### DIFF
--- a/tests/vip-helpers/test-vip-elasticsearch.php
+++ b/tests/vip-helpers/test-vip-elasticsearch.php
@@ -83,4 +83,18 @@ class VIP_ElasticSearch_Test extends \WP_UnitTestCase {
 
 		$this->assertEquals( EP_SYNC_CHUNK_LIMIT, 500 );
 	}
+
+	/**
+	 * Test that the default bulk index chunk size limit is not defined if we're not using VIP Elasticsearch
+	 *
+	 * @runInSeparateProcess
+	 * @preserveGlobalState disabled
+	 */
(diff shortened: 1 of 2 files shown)


tests/vip-helpers/test-vip-elasticsearch.php
```diff
@@ -53,4 +53,34 @@ public function test__vip_elasticsearch_filter_ep_index_name__constant_is_false(
...
 	}
+
+	/**
+	 * Test that we set a default bulk index chunk size limit
+	 *
+	 * @runInSeparateProcess
+	 * @preserveGlobalState disabled
+	 */
+	public function test__vip_elasticsearch_bulk_chunk_size_default() {
+		define( 'USE_VIP_ELASTICSEARCH', true );
```
hanifn: Should we add a test for when the constant is `false`? But I guess it doesn't really affect much even if the chunk size is set in that case

nickdaugherty: Good idea - I've added that test in 1e3e6f22

nickdaugherty:
r715-stacks

PR #1388, opened 2019-12-26
Conditionally add blog_id to index name

nickdaugherty:
## Description

Global type Elasticsearch indexes won't have a blog id for ElasticPress index names, so we need to handle that condition to avoid names like `vip-1-users-` (with a trailing dash).

## Checklist

## Steps to Test

1. Check out PR
1. Do a bulk index of users
1. Check ES indexes, should now have one without the trailing hyphen


nickdaugherty:
r716-stacks

PR #1390, opened 2019-12-31
Streamwrapper: Cache file_exists requests

hanifn:
## Description

Start caching `file_exists()`/`stat()` requests to reduce number requests to the Fileservice.

Some actions like image editing can result in multiple `file_exists()` calls which may lead to timeouts on high traffic sites.

## Checklist

## Steps to Test

1. Set `VIP_FILESYSTEM_STREAM_WRAPPER_DEBUG` constant as `true` on a test site.
1. Do some media action i.e.: upload file, edit image etc.
1. Check number of `file_exists` requests ( Has `X-action: file_exists` header )
1. Checkout PR and repeat media actions
1. Verify number of `file_exists` requests is reduced.


files/class-api-cache.php
```diff
@@ -40,14 +45,17 @@ public static function get_instance() {
...
 			return;
 		}
 
 		foreach( $this->files as $name => $path ) {
 			unlink( $path );
 			unset( $this->files[ $name ] );
 		}
+
+    // empty file stats cache
+    $this->file_stats = [];
```
nickdaugherty: Looks like the whitespace on these lines got a little funky.

files/class-api-client.php
```diff
@@ -219,6 +226,9 @@ public function is_file( $file_path, &$info = null ) {
 			$response_body = wp_remote_retrieve_body( $response );
 			$info = json_decode( $response_body, true );
 
+			// cache file info
+			$this->cache->cache_file_stats( $file_path, $info );
```
nickdaugherty: Would we also want to cache failures (404s), or is that overkill?

files/class-api-cache.php
nickdaugherty: Actually, I think most of the new lines of code have funky whitespace.

files/class-api-cache.php
```diff
@@ -95,5 +120,10 @@ public function remove_file( $filepath ) {
 			unlink( $this->files[ $file_name ] );
 			unset( $this->files[ $file_name ] );
 		}
+
+		// Remove cached stats too if any
+		if ( isset( $this->file_stats[ $file_name ] ) ) {
```
nickdaugherty: Technically we can omit the `isset()` check here for simplicity b/c PHP won't raise a warning if you `unset()` an undefined variable.

files/class-api-cache.php
```diff
@@ -95,5 +120,10 @@ public function remove_file( $filepath ) {
...
+			unset( $this->file_stats[ $file_name ] );
```
nickdaugherty: Should this be encapsulated into a reusable function, like we have for `cache_file_stats` and `get_file_stats`?

files/class-api-client.php
hanifn: I think it's overkill. I can't think of a situation where that would be necessary.

files/class-api-cache.php
hanifn: I'm not sure if that's necessary. It will basically be a one line function wouldn't it?

hanifn:
> It would also be good to add a test case to verify that calling API_Cache::remove_file() also clears the stat cache.

Added the test in 37373212202674e2c021e8994ea2034c67d1cac7

files/class-api-cache.php
hanifn: Ah looks like vim isn't remembering my `noexpandtab` setting for some reason 😐

Fixed in 6ec94e12a2fc368541ee9e57e0d236630e4033a7

files/class-api-cache.php
hanifn: Good point. Changed in 990f7cd84cc27773ecc94026cdb9773cf77821e8

chrean:
> This is a good start - are there other scenarios where we need to clear the cache though?

> Specifically, I'm thinking if a file is modified during the request - I don't see where that causes the stats cache to be cleared (maybe missing it).

Looks good @hanifn, my only concern is the same as @nickdaugherty raised just above.
I think we should invalidate/clear the cache whenever a file is modified otherwise the system may encounter inconsistencies and this will bite us back in the future.

hanifn:
>  I think we should invalidate/clear the cache whenever a file is modified otherwise the system may encounter inconsistencies and this will bite us back in the future.

Good point. Changed in e4d80d0bd6903b04e1342cd33d4f074b19ed84af

hanifn:
r731-stacks

PR #1392, opened 2019-12-31
add support for parsing and supplying the blog_url in the WXR parser

xmattus:
## Description

Gets the blog_url property from WXR files and passes it along to the output data. Also supplies the blog_url to the category filter so it is possible to act on categories conditionally based on the URL of the source site. The base URL property for WXRs generated on WP.com is not sufficient for this purpose since it is always just "wordpress.com".

## Checklist

## Steps to Test

1. Check out PR.
2. Write a filter elsewhere in your codebase that hooks to `wp_import_categories`.
3. Use the `blog_url` parameter somehow in the hooked function.


mjangda:
Thanks for the PR. However, this should be sent upstream https://en-ca.wordpress.org/plugins/wordpress-importer/

PR #1393, opened 2020-01-02
Add WP-CLI command to validate index post counts

hanifn:
## Description

Adds a new VIP WP-CLI command to validate post type counts between the DB and Elasticsearch index.

Requires ElasticPress to be installed and activated.

## Checklist

## Steps to Test

1. Setup Elasticsearch and ElasticPress on test VIP Go site
1. Checkout PR
1. Run `wp vip-es health validate-counts`
1. Verify post counts and command output


mjangda:
Aside: I think @nickdaugherty flagged this already, but we should probably move all the ElasticPress-related code into a new subfolder (can just add a new top-level one called `elasticsearch`? the `vip` is somewhat redundant :)) rather than scattering them across the mu-plugins repo like we have historically.

hanifn:
@mjangda Gotcha. I moved the files into its own `elasticsearch` directory. I also added a generic loader class so that we can move the Elasticpress related hooks into it in a future PR.

elasticsearch/class-health-command.php
```diff
@@ -0,0 +1,123 @@
...
+			} else {
+				$msg = 'found inconsistent counts for post types.';
+			}
+			WP_CLI::error( $msg );
+		}
+
+		WP_CLI::success( 'counts for public post types are all equal!' );
+	}
+
+	protected function is_elasticpress_active() {
```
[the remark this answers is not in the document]
hanifn: I'm not so sure about that. I wanted the check to be done every time the command is called and I also don't want to create a circular dependency between the command class and the `Elasticsearch` class. Perhaps a separate `Validator` class?

elasticsearch/class-elasticsearch.php
```diff
@@ -0,0 +1,38 @@
...
+
+use \WP_CLI;
+
+class Elasticsearch {
+
+	/**
+	 * Initialize the VIP Elasticsearch plugin
+	 */
+	public function init() {
+		if ( defined( 'USE_VIP_ELASTICSEARCH' ) && USE_VIP_ELASTICSEARCH ) {
```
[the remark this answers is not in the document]
hanifn: Good idea! Changed in 45959c58c0ebbdd41948192bb93e7261f73a7f52

elasticsearch/class-health-command.php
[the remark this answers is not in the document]
mjangda: Would `use \ElasticPress\Indexables as Indexables;` fatal if the plugin isn't loaded?

@nickdaugherty and I chatted a bit about this earlier today, and we're thinking that it makes sense for us to bundle and auto-load the ElasticPress plugin via `Automattic\VIP\Elasticsearch\Elasticsearch` (we can do this is in a future PR). 

Anyway, I think it's okay to leave as-is but probably can be moved (or removed) in the future.

000-vip-init.php
```diff
@@ -174,6 +174,14 @@
 	require_once( __DIR__ . '/vip-helpers/vip-wp-cli.php' );
 }
 
+// Load elasticsearch helpers
+if ( defined( 'USE_VIP_ELASTICSEARCH' ) && USE_VIP_ELASTICSEARCH ) {
+	require_once( __DIR__ . '/elasticsearch/class-elasticsearch.php');
```
```suggestion
	require_once( __DIR__ . '/elasticsearch/class-elasticsearch.php' );
```

elasticsearch/class-health-command.php
```diff
@@ -0,0 +1,123 @@
...
+	 *     wp vip-es health validate-counts
+	 *
+	 * @subcommand validate-counts
+	 */
+	public function validate_counts( $args, $assoc_args ) {
+		// Make sure ElasticPress is installed and active
+		$this->is_elasticpress_active();
+
+		// Get indexable objects
+		$indexable = Indexables::factory()->get( 'post' );
```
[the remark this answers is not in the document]
nickdaugherty: +1 - ideally this logic runs automatically, such as via cron, so we'll want to abstract it into a standalone function that gets called in the command - that also makes it easier to add tests.

elasticsearch/class-health-command.php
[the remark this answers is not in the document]
nickdaugherty: That sounds like a plan - since this PR includes the base `Elasticsearch` class that we'll need to use to load ElasticPress, let's leave it as-is in this PR then follow up by adding ElasticPress to the `elasticsearch` directory, and auto-load it when `USE_VIP_ELASTICSEARCH` is `true`.

elasticsearch/class-health-command.php
```diff
@@ -0,0 +1,123 @@
...
+
+			$icon = "\u{2705}"; // unicode check mark
+			$diff = '';
+			if ( $db_total !== $es_total ) {
+				$error = true;
+
+				$icon = "\u{274C}"; // unicode cross mark
+
+				if ( $es_result ) {
+					$diff = sprintf( ', diff: %d', $es_total - $db_total );
```
nickdaugherty: The DB could also have more items than the ES index, which would cause this to display a negative number.

elasticsearch/class-health-command.php
```diff
@@ -0,0 +1,123 @@
...
+			$query = new WP_Query( $query_args );
+
+			$formatted_args = $indexable->format_args( $query->query_vars, $query );
+
+			$es_result = $indexable->query_es( $formatted_args, $query->query_vars );
+			
+			if ( ! $es_result ) {
+				$es_total = 'N/A';
+				$error = true;
+				$es_conn_err = true;
```
nickdaugherty: Should this condition also break the `foreach()`? Or, maybe we should just put the logic from line 81 into the `else` statement, so that it doesn't try to run after an error.

elasticsearch/class-health-command.php
```diff
@@ -0,0 +1,123 @@
...
+	public function validate_counts( $args, $assoc_args ) {
+		// Make sure ElasticPress is installed and active
+		$this->is_elasticpress_active();
+
+		// Get indexable objects
+		$indexable = Indexables::factory()->get( 'post' );
+
+		$post_types = $indexable->get_indexable_post_types();
+
+		WP_CLI::line( sprintf( "Checking %d post types (%s)\n", count( $post_types ), implode( ',', $post_types ) ) );
```
nickdaugherty:
```suggestion
		WP_CLI::line( sprintf( "Checking %d post types (%s)\n", count( $post_types ), implode( ', ', $post_types ) ) );
```

For readability, probably should add a space to the `implode()`.

elasticsearch/class-health-command.php
hanifn: That is deliberate. A negative diff would indicate the ES index is missing items while a positive diff would mean the opposite

elasticsearch/class-health-command.php
hanifn: I didn't break the `foreach()` loop as I'm thinking it could only be temporary and the number of post types could vary depending on the site.

Good point about putting that logic in the `else` statement though.

hanifn: Changed in 21a5ac773c63707f872a89c74b79b56698fa3683

elasticsearch/class-health-command.php
nickdaugherty: That might be confusing to read as someone who runs the command, because you may not know which is the "left" side vs "right" side of the equation (so without knowledge of the code, you don't know if negative means posts missing from the DB or from the index).

For clarity, I'd suggest a more explicit message, like `index has 18 more posts than db` and `db has 90 more posts than index`.

mjangda: Thoughts on shipping as-is, and fixing up the output to be nicer in a future PR?

nickdaugherty: Ya that would be fine, it's minor.

elasticsearch/class-health-command.php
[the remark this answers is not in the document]
nickdaugherty: We can punt this change to a later time, when we automate these checks.

nickdaugherty:
r725-stacks

PR #1394, opened 2020-01-02
Replace ElasticPress network layer with vip_safe_wp_remote_get()

nickdaugherty:
## Description

Switch out the default WP HTTP API for Elasticsearch requests with `vip_wp_safe_remote_get()` for added safety (timeout and limited global failure counts).

Depends on #1395

## Checklist

## Steps to Test

1. On a local env with Elasticsearch...
1. Switch to this branch
1. Perform some searches regularly
1. Reindex everything
1. Reindex single item
1. Verify above all behaves as expected
1. Stop `elasticsearch` container
1. Repeat above, but verify that requests are erroring with the expected error from our interception layer (`There was an error connecting to the upstream Elasticsearch server`). This can be viewed in the ElasticPress Debug Bar panel


vip-helpers/vip-elasticsearch.php
```diff
@@ -469,6 +469,12 @@ function vip_elasticsearch_filter_ep_index_name( $index_name, $blog_id, $indexab
 	return $index_name;
 }
 
+function vip_elasticsearch_filter_ep_do_intercept_request( $request, $query, $args, $failures ) {
+	$request = vip_safe_wp_remote_request( $query['url'], null, 3, 1, 20, $args );
```
nickdaugherty: Instead of passing `null` for the fallback, we need to pass a `WP_Error()`, otherwise we'll literally get `null` back when it fails due to how PHP's default function args work.

nickdaugherty: This is done, in b90e0d2e

PR #1395, opened 2020-01-02
Add vip_safe_wp_remote_request()

nickdaugherty:
## Description

Same function as `vip_safe_wp_remote_get()`, except made generic to support all methods. The `vip_safe_wp_remote_get()` function is now a simple wrapper to this.

## Checklist

## Deploy Notes

Should update the docs here: http://vip.wordpress.com/documentation/fetching-remote-data/

## Steps to Test

Do something like this, which tests both `vip_safe_wp_remote_request()` and `vip_safe_wp_remote_get()` along with a POST request (which should get cached differently):

```
$res = vip_safe_wp_remote_request( 'http://slowwly.robertomurray.co.uk/delay/4000/url/http://www.google.co.uk' );

var_dump( $res );

$res = vip_safe_wp_remote_get( 'http://slowwly.robertomurray.co.uk/delay/4000/url/http://www.google.co.uk' );

var_dump( $res );

$res = vip_safe_wp_remote_request( 'http://slowwly.robertomurray.co.uk/delay/4000/url/http://www.google.co.uk',      null, 3, 1, 20, array( 'method' => 'POST' ) );

var_dump( $res );

$res = vip_safe_wp_remote_get( 'http://slowwly.robertomurray.co.uk/delay/4000/url/http://www.google.co.uk' );

var_dump( $res );
```


vip-helpers/vip-utils.php
```diff
@@ -718,26 +718,29 @@ function vip_get_random_posts( $number = 1, $post_type = 'post', $return_ids = f
...
-function vip_safe_wp_remote_get( $url, $fallback_value='', $threshold=3, $timeout=1, $retry=20, $args = array() ) {
+function vip_safe_wp_remote_request( $url, $fallback_value='', $threshold=3, $timeout=1, $retry=20, $args = array() ) {
 	global $blog_id;
 
-	$cache_group = "$blog_id:vip_safe_wp_remote_get";
-	$cache_key = 'disable_remote_get_' . md5( parse_url( $url, PHP_URL_HOST ) );
+	$default_args = array( 'method' => 'GET' );
+	$parsed_args = wp_parse_args( $args, $default_args );
+
+	$cache_group = "$blog_id:vip_safe_wp_remote_request";
```
mjangda: Any concerns around this causing a mass cache invalidation?

nickdaugherty: I'm not too concerned - the function only caches (and for 20 seconds by default) when the remote server is down or responding slowly (which theoretically is only a small portion of requests) - regular requests won't be relying on a cache entry, and it expires quickly anyway.

But, would be fine to leave the group as-is too.

vip-helpers/vip-utils.php
```diff
@@ -796,6 +799,30 @@ function vip_safe_wp_remote_get( $url, $fallback_value='', $threshold=3, $timeou
...
+ *
+ * @author tottdev
+ * @link http://vip.wordpress.com/documentation/fetching-remote-data/ Fetching Remote Data
+ * @param string $url URL to fetch
+ * @param string $fallback_value Optional. Set a fallback value to be returned if the external request fails.
+ * @param int $threshold Optional. The number of fails required before subsequent requests automatically return the fallback value. Defaults to 3, with a maximum of 10.
+ * @param int $timeout Optional. Number of seconds before the request times out. Valid values 1-3; defaults to 1.
+ * @param int $retry Optional. Number of seconds before resetting the fail counter and the number of seconds to delay making new requests after the fail threshold is reached. Defaults to 20, with a minimum of 10.
+ * @param array Optional. Set other arguments to be passed to wp_remote_get().
+ * @return string|WP_Error|array Array of results. If fail counter is met, returns the $fallback_value, otherwise return WP_Error.
```
mjangda: Minor: Should we just skip the duplicate `@param` docs here since they're the same as the new function?

tests/vip-helpers/test-vip-utils.php
```diff
@@ -1,6 +1,20 @@
 <?php
 
 class WPCOM_VIP_Get_Resized_Attachment_Url_Test extends \WP_UnitTestCase {
```
mjangda: Should the tests go in a new class? The name of this class implies it's just for `Get_Resized_Attachment_Url` :)

tests/vip-helpers/test-vip-utils.php
```diff
@@ -22,4 +36,170 @@ public function test__valid_attachment() {
...
+	 * Test vip_safe_wp_remote_request() behavior with normal responses
+	 */
+	public function test__vip_safe_wp_remote_request_with_normal_response() {
+		$url = 'https://localhost';
+		$response = 'mock_response';
+
+		$this->mock_http_response( $response, 0.10 );
+
+		// We can call it 4 times (more than the default threshold of 3) and it always returns the expected response (no failure / fallback)
+		for( $i = 0; $i < 4; $i++ ) {
```
mjangda: Aside: I'm always vary of loops in things like tests (adds too much additional logic), but I think they're okay here as we have things very well-documented; kudos on that btw).

vip-helpers/vip-utils.php
nickdaugherty: The docblock is used to generate documentation, so I think if we omitted it here, the docs would be incomplete? Could leave a note to see `vip_safe_wp_remote_request()`, but that feels clunky as a UX for someone reading the docs. Thoughts?

tests/vip-helpers/test-vip-utils.php
nickdaugherty: Maybe we should add a `tests/vip-helpers/vip-utils` directory to contain the individual parts? `vip-utils.php` is pretty big and unrelated.

vip-helpers/vip-utils.php
```diff
@@ -718,26 +718,29 @@ function vip_get_random_posts( $number = 1, $post_type = 'post', $return_ids = f
...
 
 /**
- * This is a sophisticated extended version of wp_remote_get(). It is designed to more gracefully handle failure than wpcom_vip_file_get_contents() does.
+ * This is a sophisticated extended version of wp_remote_request(). It is designed to more gracefully handle failure than wpcom_vip_file_get_contents() does.
  *
- * Note that like wp_remote_get(), this function does not cache.
+ * Note that like wp_remote_request(), this function does not cache.
  *
  * @author tottdev
  * @link http://vip.wordpress.com/documentation/fetching-remote-data/ Fetching Remote Data
```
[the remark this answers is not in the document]
nickdaugherty: Good catch - added a note to the PR description.

tests/vip-helpers/test-vip-utils.php
nickdaugherty: Good call, done in d80ecdaf

tests/vip-helpers/test-vip-utils.php
```diff
@@ -1,6 +1,20 @@
 <?php
 
 class WPCOM_VIP_Get_Resized_Attachment_Url_Test extends \WP_UnitTestCase {
+	public function tearDown() {
+		remove_all_filters( 'pre_http_request' );
```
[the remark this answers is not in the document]
nickdaugherty: Ah, was going off some prior code in the Filesystem cache stuff, removed in cac70ab6

vip-helpers/vip-utils.php
mjangda: It's a pretty common practice to crosslink functions in docs. 

The main concern with duplicating them is that these could get out of date which would be much worse than bad UX.

tests/vip-helpers/test-vip-utils.php
mjangda: Splitting into smaller test files is a great idea.

mjangda:
@nickdaugherty I realized that I forgot to mention that I didn't add inline suggestions for the other remaining multi-assert calls, so we should make sure we add the `message` arg for all of them.

One other thought (that shouldn't hold up the PR): the tests for `wp_remote_get` feel somewhat redundant, since they're mostly just verifying the behavior of the underlying `wp_remote_request` function. It would be better to test the core logic of the `wp_remote_get` function, the bulk of which is to make sure the remote request uses `GET`. All the other tests are probably just unnecessary.

vip-helpers/vip-utils.php
nickdaugherty: True - switched it out in 70d79679

nickdaugherty:
> that I didn't add inline suggestions for the other remaining multi-assert calls

Oops, I should have double checked - added in d2806117

nickdaugherty:
r726-stacks

PR #1396, opened 2020-01-09
Move VIP ES setup code to the new class

nickdaugherty:
## Description

Now that we have the `Automattic\VIP\Elasticsearch\Elasticsearch` class, move the existing constant/filter setup code to it.

## Checklist

## Steps to Test

1. Pull down PR
1. Use the site without setting `USE_VIP_ELASTICSEARCH` constant - class shouldn't load, index names aren't filtered, etc
1. Set `USE_VIP_ELASTICSEARCH` to `true` and ensure the class loads and sets the constants / filters


tests/elasticsearch/test-class-elasticsearch.php
```diff
@@ -57,63 +59,41 @@ public function test__vip_elasticsearch_filter_ep_index_name__no_constant() {
...
 	 * @runInSeparateProcess
 	 * @preserveGlobalState disabled
 	 */
 	public function test__vip_elasticsearch_bulk_chunk_size_already_defined() {
-		define( 'USE_VIP_ELASTICSEARCH', true );
-
 		define( 'EP_SYNC_CHUNK_LIMIT', 500 );
 
-		vip_elasticsearch_setup_constants();
+		$es = new \Automattic\VIP\Elasticsearch\Elasticsearch();
```
mjangda: Should we move this to `setUp`?

tests/elasticsearch/test-class-elasticsearch.php
```diff
@@ -57,63 +59,41 @@ public function test__vip_elasticsearch_filter_ep_index_name__no_constant() {
...
+		$es->setup_hooks();
```
mjangda: Should this be `init` so we load the full state of the ES class?

elasticsearch/class-elasticsearch.php
```diff
@@ -24,13 +25,35 @@ protected function load_dependencies() {
...
+	}
+
+	public function setup_hooks() {
+		add_filter( 'ep_index_name', [ $this, 'filter__ep_index_name' ], PHP_INT_MAX, 3 ); // We want to enforce the naming, so run this really late.
 	}
 
-	protected function load_commands() {
+	public function load_commands() {
 		if ( defined( 'WP_CLI' ) && WP_CLI ) {
 			WP_CLI::add_command( 'vip-es health', __NAMESPACE__ . '\Health_Command' );
```
GaryJones: Considering this is unlikely to work, without the `commands/class-health-command.php` having been required first, why is the require_once in a separate method?

And why are they both `public` methods, when the public `init` method already seems to handle this?

tests/elasticsearch/test-class-elasticsearch.php
nickdaugherty: I don't think so, because we want to `init()` or call class methods independently for each test, so having a globally scoped object that we're using in different ways in the tests could lead to cross-contamination of the state (even though technically we currently run each in a separate process).

tests/elasticsearch/test-class-elasticsearch.php
nickdaugherty: It could be - it seemed better to test just the hook part of it in isolation, though there is value in testing the whole thing.

elasticsearch/class-elasticsearch.php
nickdaugherty: > why is the require_once in a separate method?

We have a `load_dependencies()` method which is intended to gather up all the code needed for the plugin to work - it's a good point about this not working if it's called independently of `load_dependencies()`.

> And why are they both public methods

I switched them all to `public` for unit tests, because I wanted to test parts of it in isolation. Technically this one doesn't have to be `public` since there isn't a test for it.

GaryJones: [Using Reflection for testing private methods](https://42coders.com/how-to-use-reflection-to-test-private-and-protected-methods/) is typically preferred, rather than exposing methods that otherwise wouldn't be exposed if it weren't for unit testing.

nickdaugherty: Ah much better, thanks @GaryJones !

tests/elasticsearch/test-class-elasticsearch.php
nickdaugherty: Decided to go this way b/c it's not really possible to use PHPUnit's mock/spy on non-public methods (so was unable to write a test for `init()` that ensured the various helpers are called).

https://github.com/Automattic/vip-go-mu-plugins/commit/e286964025fdc19131679610a51d5776e49b7a09

mjangda:
Whoops, broke the tests. Fixing.

nickdaugherty:
r727-stacks

PR #1397, opened 2020-01-10
Fix OOM errors on large video uploads

hanifn:
## Description

Uploading large video files(size close to 1GB), results in an Out of memory Fatal error  in the post-upload meta processing stage. Specifically, in the call to `wp_read_video_metadata()`.

Root cause appears to be the use of `file_get_contents()` within the stream wrapper cache class. This PR changes how the local cache works by returning a stream instead of the file content to resolve the OOM error.

## Checklist

## Steps to Test

1. Check out PR.
1. Go to `wp-admin` > `Media` > `Library`
1. Upload a video file that is close to the 1GB limit
1. Verify no errors occured.


files/class-api-client.php
```diff
@@ -152,14 +153,22 @@ private function calculate_upload_timeout( $file_size ) {
...
-			return $file_content;
+		$file = $this->cache->get_file( $file_path );
+		if ( $file ) {
+			return $file;
 		}
 
+		$tmp_file = tempnam( get_temp_dir(), 'vip' );
+		$tmp_handle = fopen( $tmp_file, 'w+' );
+
+		$curl_streamer = new Curl_Download_Streamer( $tmp_handle );
```
mjangda: I just noticed that `WP_Http` has a `stream` option that already does this so the `Curl_Download_Streamer` is kind of unnecessary:

```
'stream'
(bool) Whether to stream to a file. If set to true and no filename was given, it will be droped it in the WP temp dir and its name will be set using the basename of the URL. Default false.
'filename'
(string) Filename of the file to write to when streaming. $stream must be set to true. Default null.
```

https://developer.wordpress.org/reference/classes/wp_http/request/

mjangda:
We'll need to test various other media get / upload workflows (e.g. media library, direct usage of stream path via file_(get|put)_contents, etc.)

files/class-api-client.php
hanifn: It works! The streamer class is now no longer necessary! Changed in a448e535a538d6752d38c434bca0af0001a238b8

mjangda: Awesome!

files/class-api-client.php
```diff
@@ -152,13 +152,20 @@ private function calculate_upload_timeout( $file_size ) {
...
 		// check in cache first
-		$file_content = $this->cache->get_file( $file_path );
-		if ( $file_content ) {
-			return $file_content;
+		$file = $this->cache->get_file( $file_path );
+		if ( $file ) {
+			return $file;
 		}
 
+		$tmp_file = tempnam( get_temp_dir(), 'vip' );
```
nickdaugherty: What's the advantage to moving this cache file creation out of the cache class and into the api client class? It seems to me better to encapsulate all the details inside the cache class.

files/class-api-client.php
```diff
@@ -152,13 +152,20 @@ private function calculate_upload_timeout( $file_size ) {
...
+
+		$request_args = [
+			'stream' => true,
+			'filename' => $tmp_file,
+		];
```
nickdaugherty: What do these args do in `$this->call_api()`? Looks like they are passed through to the files service - a comment explaining what it's doing would be helpful.

files/class-api-client.php
```diff
@@ -173,12 +180,10 @@ public function get_file( $file_path ) {
...
 		}
 
-		$body = wp_remote_retrieve_body( $response );
-
 		// save to cache
-		$this->cache->cache_file( $file_path, $body );
+		$this->cache->cache_file( $file_path, $tmp_file );
 
-		return $body;
+		return fopen( $tmp_file, 'r+' );
```
nickdaugherty: Should this be `$this->cache->get_file( $file_path )`? I worry a bit about duplicating the details of how we're reading temp files from disk.

tests/files/test-api-client.php
```diff
@@ -314,8 +322,19 @@ public function get_test_data__get_file() {
 	public function test__get_file( $mocked_response, $expected_result ) {
 		$this->mock_http_response( $mocked_response );
 
-		$actual_result = $this->api_client->get_file( '/wp-content/uploads/file.jpg' );
+		$file = $this->api_client->get_file( '/wp-content/uploads/get_file.jpg' );
+		
+		if ( is_object( $file ) ) {
```
nickdaugherty: In what scenarios does `get_file()` return an object now - shouldn't the behavior be the same as before (and if not, are there any other spots that need to account for different types)?

mjangda: I think the Api_Client is also used in one of the WP_Filesystem_VIP classes so we should make sure to update those uses as well.

mjangda: Yeah, looks like here:

https://github.com/Automattic/vip-go-mu-plugins/blob/master/files/class-wp-filesystem-vip-uploads.php#L60-L68

We'll need to update that method to read the file resource and return the file contents (it's another point where OOM errors can happen but I we can't really change the method signature there).

files/class-api-client.php
hanifn: This was mainly a side-effect of having to pass the file as the arguments for `call_api()`. I'm not sure if it's better encapsulated in the cache class though. It _should_ be encapsulated but I think the API client class works just as well

files/class-api-client.php
hanifn: It is ultimately used as arguments for `wp_remote_request()` but I agree, some comments would help clear that up.

files/class-api-client.php
hanifn: Agreed about duplication and I think we should move the `fopen()` calls to the Stream wrapper class instead so that it's opened with the correct mode.

tests/files/test-api-client.php
hanifn: Right now it's returning a `WP_Error` object or the file stream. I'm not sure if this is the best way though. Perhaps throwing an error instead so that we have a consistent return type is a better idea.

files/class-api-client.php
nickdaugherty: Does it mean that `wp_remote_request()` will stream the response to the file?

tests/files/test-api-client.php
nickdaugherty: Ah a WP_Error is fine - in that case we should use `is_wp_error()` to detect it.

files/class-api-client.php
hanifn: Yes. If `stream` is `true` then `wp_remote_request()` will stream the response into `filename`

files/class-api-client.php
hanifn: Changed in 77af3de4b107b9a46476a537f6eb42e503987769

tests/files/test-api-client.php
hanifn: @mjangda Updated in 354d9bb812247816b96539f64e785a5db61562b0

hanifn:
r734-stacks

PR #1398, opened 2020-01-10
Add ElasticPress as a submodule and autoload it

chrean:
## Description

IN PROGRESS

This PR is aimed at autoloading ElasticPress as a part of MU Plugins.
The initial committed version is 3.3, which supports ElasticSearch 7.5.

## Checklist

## Steps to Test

TBD


chrean:
@nickdaugherty talked with @mjangda and we thought we can ship this as it is, since it already loads the plugin and managing conflicts is the hardest part of this implementation.
We can do those additional refinements, including making the plugin showing up in the MU list, in the near future. I will add Jira issues for those additional tasks.
Also, we thought that defining an `adapters` directory may be misleading so we thought we could just move `elasticpress` up one level, just below `elasticsearch`.
Should we start working with other elasticsearch plugins, then we will add the `adapters` directory or something similar.

nickdaugherty:
r729-stacks

PR #1402, opened 2020-01-14
Disable JP Search when using VIP ES

nickdaugherty:
## Description

Auto disable JP Search when VIP ES is active, to prevent conflicts.

Branched from #1394 (to prevent conflicts) so that should be merged first.

## Checklist

## Steps to Test

1. Enable JP Search on a site via UI
1. Set `USE_VIP_ELASTICSEARCH` to `true`
1. Ensure JP Search is now disabled
1. Enable JP Search [via code](https://wpvip.com/documentation/vip-go/elasticsearch-on-vip-go/#how-to-activate-the-search-module)
1. Disable `USE_VIP_ELASTICSEARCH` to ensure above filter is working
1. Re-enable the constant, ensure JP Search is now disabled


elasticsearch/class-elasticsearch.php
```diff
@@ -34,6 +34,11 @@ protected function setup_constants() {
 
 	protected function setup_hooks() {
 		add_filter( 'ep_index_name', [ $this, 'filter__ep_index_name' ], PHP_INT_MAX, 3 ); // We want to enforce the naming, so run this really late.
+
+		// Network layer replacement to use VIP helpers (that handle slow/down upstream server)
+		add_filter( 'ep_intercept_remote_request', '__return_true', 9999 );
+		add_filter( 'ep_do_intercept_request', [ $this, 'filter__ep_do_intercept_request' ], 9999, 4 );
+		add_filter( 'jetpack_active_modules', [ $this, 'filter__jetpack_active_modules' ], 9999 );
```
rinatkhaziev: I'd do `PHP_INT_MAX` here just to be absolutely sure we disable Jetpack's module

elasticsearch/class-elasticsearch.php
```diff
@@ -56,4 +61,22 @@ public function filter__ep_index_name( $index_name, $blog_id, $indexables ) {
...
+
+		$request = vip_safe_wp_remote_request( $query['url'], $fallback_error, 3, 1, 20, $args );
+	
+		return $request;
+	}
+
+	public function filter__jetpack_active_modules( $modules ) {
+		// Filter out 'search' from the active modules. We use array_filter() to get _all_ instances, as it could be present multiple times
+		$filtered = array_filter ( $modules, function( $module ) {
+			return 'search' === $module ? false : true;
```
mjangda: Minor: the ternary plus flipped return is a bit hard to read. I would simplify with an `if`:

```suggestion
			if ( 'search' === $module ) {
				return false;
			}
			return true;
```

elasticsearch/class-elasticsearch.php
nickdaugherty: I think `9999` is better for this one, b/c it allows overridding if really needed. For example, maybe JP Search provided a certain piece of functionality that VIP ES doesn't and a client wanted to selectively use that part (disabling other parts of JP Search).

nickdaugherty:
r730-stacks

PR #1403, opened 2020-01-15
Update to Two-Factor 0.5.0

pyronaur:
The Updated Two-Factor version fixes the issue where users who log-in using WordPress.com SSO are logged out every day because of missing  `rememberme` value.

Related Pull Requests:

* https://github.com/WordPress/two-factor/pull/308
* https://github.com/WordPress/two-factor/pull/316

How to test:

**Prepare:**
1. Create a personal test WordPress.com account without any A8C Privileges
2. Add that user to a test site as an admin but without any special A8C Privileges using WP CLI.
3. Make sure you are not using a special IP
4. Force 2FA if not already enabled.

**Test**
1. "Login with WordPress.com" to a site
2. Confirm that the 2FA dialog form has a hidden `rememberme` value and it's set to `0`
2. Enter 2FA Code
3. Login
4. Confirm that auth cookie expiration is set to `Session`

Apply the PR, repeat the steps above, with the following changes:

- `2.` Confirm that the 2FA dialog form has a hidden `rememberme` value and it's set to `1`
- `4.` Confirm that the auth cookie expiration is set to roughly 1 year from now.

### DIFF
--- a/.deployignore
+++ b/.deployignore
@@ -1,12 +1,15 @@
 .deployignore
 
 .ackrc
+.babelrc
+.browserslistrc
 .circleci
 .codeclimate.yml
 .coveralls.yml
 .distignore
 .editorconfig
 .eslines.json
+.eslintignore
 .eslintrc
 .eslintrc.js
 .git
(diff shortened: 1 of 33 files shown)


mjangda:
@pyronaur can you share specific flows that need to be tested?

pyronaur:
@mjangda Updated the description

mjangda:
Will deploy on Monday :)

mjangda:
r744-stacks

PR #1404, opened 2020-01-16
Fix privacy archive generation

### LINKED ISSUE #1399: Export Personal Data broken
## Expected/Desired Behavior

Hitting "Download Personal Data" in the Export Personal Data page of wp-admin returns a download of the data.

## Actual Behavior

Hitting "Download Personal Data" in the Export Personal Data page of wp-admin results in the error message:

```
An error occurred while attempting to export personal data.

Unable to generate export file (archive) for writing: Unable to add data to export file.
```

It doesn't matter whether Zip extension is enabled or not, it seems that the file is not being uploaded properly and fails at this point if Zip is enabled: https://github.com/Automattic/vip-go-mu-plugins/blob/738371872a33f879212f20af35d244971bfdd975/001-core/privacy.php#L264

or at this point if Zip is not enabled: https://github.com/Automattic/vip-go-mu-plugins/blob/738371872a33f879212f20af35d244971bfdd975/001-core/privacy.php#L289

1. Go to /wp-admin/tools.php?page=export_personal_data
1. If there are no data export requesters yet, add yourself as one via:
Screen Shot 2020-01-10 at 4 29 22 PM

3. Hit "Download Personal Data"

mjangda:
ZipArchive and PclZip don't support stream paths. To make our privacy export workaround work, we're now forcing the archive to be generated locally (in the tmp dir) and then uploaded after.

This also generates the HTML file locally as well, which was previously being generated on the Files Service and then attempted to be downloaded when archiving.

Most of that file and workaround is a bit of a hack so this should be considered a short-term fix.

Fixes #1399 

## Checklist

## Steps to Test

1. Go to wp-admin > Tools > Privacy
1. Generate a request
1. Attempt to download the file for the request (it should work)

Alternatively, programatically:

```
wp shell
Automattic\VIP\Core\Privacy\generate_personal_data_export_file( <id-of-request> )
```

That should return a `null` not an error.

### DIFF
--- a/001-core/privacy.php
+++ b/001-core/privacy.php
@@ -60,6 +60,7 @@ function generate_personal_data_export_file( $request_id ) {
 	// Create the exports folder if needed.
 	$exports_dir = wp_privacy_exports_dir();
 	$exports_url = wp_privacy_exports_url();
+	$temp_dir = get_temp_dir();
 
 	$result = wp_mkdir_p( $exports_dir );
 	if ( is_wp_error( $result ) ) {
@@ -73,7 +74,8 @@ function generate_personal_data_export_file( $request_id ) {
 	$obscura              = wp_generate_password( 32, false, false );
 	$file_basename        = 'wp-personal-data-file-' . $stripped_email . '-' . $obscura;
 	$html_report_filename = $file_basename . '.html';
-	$html_report_pathname = wp_normalize_path( $exports_dir . $html_report_filename );
+	// Use temp_dir because we don't want the file generated remotely yet.
+	$html_report_pathname = wp_normalize_path( $temp_dir . $html_report_filename );
 
 	$file = fopen( $html_report_pathname, 'w' );
 	if ( false === $file ) {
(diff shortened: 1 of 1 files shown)


001-core/privacy.php
```diff
@@ -60,6 +60,7 @@ function generate_personal_data_export_file( $request_id ) {
 	// Create the exports folder if needed.
 	$exports_dir = wp_privacy_exports_dir();
 	$exports_url = wp_privacy_exports_url();
+	$temp_dir = get_temp_dir();
 
 	$result = wp_mkdir_p( $exports_dir );
```
hanifn: Is this step still necessary? Or even the `$exports_dir` variable itself necessary?

001-core/privacy.php
```diff
@@ -177,14 +179,26 @@ function generate_personal_data_export_file( $request_id ) {
...
+	$local_archive_pathname = $archive_pathname;
+	// Hack: ZipArchive and PclZip don't support streams.
+	// So, let's force the path to use a local one in the temp dir, which will work.
+	// All other references (meta) will still use the correct stream URL. 
+	if ( 0 === strpos( $local_archive_pathname, 'vip://' ) ) {
+		$local_archive_pathname = get_temp_dir() . substr( $archive_pathname, 6 );
+
+		// Create the folder path
+		$local_archive_dirname = dirname( $local_archive_pathname );
+		wp_mkdir_p( $local_archive_dirname );
```
hanifn: Should we add a check to ensure the directory path was properly created?

001-core/privacy.php
mjangda: Yeah, still needed for sites not using stream wrappers (AMS).

001-core/privacy.php
mjangda: Yep, good catch 👍

mjangda: Good catch; done in 6c5c78a62070d020d994bb2ace17ce2fea5a10d7

mjangda:
r733-stacks

PR #1410, opened 2020-01-18
Fix subsite upload path for privacy exports

mjangda:
When generating an export in a subsite (on a multisite install), the export was always uploaded to the root, rather than the `/sites/%d` path.

This fixes the path of the upload.

This is a quick fix until we can remove most of the workaround hacks in this file.

## Checklist

## Steps to Test
From a multisite main site:

1. Generate a export request (Tools > Export Privacy Data)
1. Click "download export file"
1. Verify the zip file downloads and has the correct export.

Then:

1. Goto a subsite, and repeat steps above.

Verify again with stream wrapper disabled. Both should work.


mjangda:
r736-deploy

PR #1411, opened 2020-01-20
Improvements on ElasticSearch validation command

chrean:
## Description

Fixes [PLAT-798]
Adds following validations to the ES validation command:

- Count non-public posts
- Count terms
- Count users

## Checklist

## Steps to Test

- Pull down PR
- start up your local environment
- `wp vip-es health validate-posts-count`
- `wp vip-es health validate-users-count`
- `wp vip-es health validate-counts`

[PLAT-798]: https://vipjira.atlassian.net/browse/PLAT-798


elasticsearch/class-elasticsearch.php
```diff
@@ -49,6 +54,147 @@ protected function load_commands() {
...
+	 * @param array $query_args Valid WP_Query criteria, mandatory fields as in following example:
+	 * $query_args = [
+	 *		'post_type' => $post_type,
+	 *		'post_status' => array( $post_statuses )
+	 * ];
+	 *
+	 * @param string $type Type (Slug) of the objects to be searched (should be either 'user' or 'post')
+	 * @return WP_Query
+	 */
+	private function queryObjects( array $query_args, string $type ) {
```
nickdaugherty:
```suggestion
	private function query_objects( array $query_args, string $type ) {
```

elasticsearch/class-elasticsearch.php
```diff
@@ -49,6 +54,147 @@ protected function load_commands() {
...
+			$query_args = [
+				'post_type' => $post_type,
+				'post_status' => array_values( $post_statuses ),
+			];
+
+			$result = self::verify_index_entity_count( $query_args, $posts );
+
+			// In case of error skip to the next post type
+			if ( is_wp_error( $result ) ) {
+				WP_CLI::line( ' error while verifying post type: ' . $post_type . ', details: ' . $result->get_error_message() );
```
nickdaugherty: Probably just an oversight - can we return the `WP_Error` in the `$results` array instead of `WP_CLI::line()`?

chrean:
Thanks for the review @nickdaugherty !

> but now I'm thinking we should split it out into Automattic\VIP\Elasticsearch\Health to keep this class lean and organize it better...what do you think?

As I was moving functions into the main class, I felt the code was becoming maybe a bit awkward, mixing validation with the rest. On the other hand, I believed that creating a dedicated class for Health monitoring may have been a bit of an overkill right now.
But, since we are going to extend it soon, it may be reasonable doing it now before we deploy.

elasticsearch/class-elasticsearch.php
chrean: Too much Node.JS last few years :D

elasticsearch/class-elasticsearch.php
chrean: Done. Now returning `WP_Error` as part of `$results`, adding a specific `error` field that is being checked by `render_results()`.

elasticsearch/class-elasticsearch.php
```diff
@@ -49,6 +54,144 @@ protected function load_commands() {
 		}
 	}
 
+	/**
+	 * Return singleton instance of class
+	 *
+	 * @return object
+	 */
+	public static function factory() {
```
GaryJones: Why a Singleton pattern (and the constructor isn't private anyway, so this isn't forcing anything)?

How will you mock this class for unit tests if it's a Singleton?
What benefit does it give?

GaryJones: In terms of its use in the `Health_Command` class, instantiating `Elasticsearch` outside of the command class, and injecting it as a dependency to the constructor (and saving that to a property of `Health_Command` class`) would allow `Elasticsearch` to be mocked, and to follow best (SOLID) practices.

The `WP_CLI::add_command(…)` can accept an instantiated object, so you can construct and inject before handing over to WP-CLI. So instead of [this line](https://github.com/Automattic/vip-go-mu-plugins/blob/add/es-validation-taxonomy/elasticsearch/class-elasticsearch.php#L53), you could do:

```php
$hc = new \Health_Command( new Elasticsearch() );
WP_CLI::add_command 'vip-es health, $hc );
```

That way, instead of having to instantiate a new Elasticsearch each time (via the factory method), you can instantiate it once, and re-use the service object to do the verifications.

chrean: > Why a Singleton pattern (and the constructor isn't private anyway, so this isn't forcing anything)?

Do we really need to have multiple instances of this class?
Also this class has no constructor at all; we may either add a `private` one, or a `public` one which returns the singleton instance.
Singleton was used for the sake of this PR, but assuming that the class will be used by other modules in the future, we may convert to a normal class.
@nickdaugherty thoughts?

> How will you mock this class for unit tests if it's a Singleton?

There are ways to mock the class for unit tests, but perhaps it makes things too complex.

chrean: Thanks @GaryJones for the input, much appreciated! Also please feel free to contribute to the PR if you'd like to write some code!

chrean: Moreover: I used singleton but probably I should have used `static` methods for `validate_X` functions.
Ideally, as Nick noted later on, we could also move those methods into their own class.
I will work at refactoring this code to be more consistent and make more sense.

chrean:
@nickdaugherty  I have made some changes as we discussed yesterday, I think this may be good enough to be pushed out.

elasticsearch/elasticsearch.php
```diff
@@ -0,0 +1,51 @@
...
+			return;
+		}
+
+		$relative_class = substr( $class, $len );
+
+		$file = $base_dir . str_replace( '\\', '/', $relative_class ) . '.php';
+
+		// if the file exists, require it.
+		if ( file_exists( $file ) ) {
+			require $file;
```
nickdaugherty:
```suggestion
			require_once $file;
```

elasticsearch/elasticsearch.php
```diff
@@ -0,0 +1,51 @@
...
+		}
+	}
+);
+
+do_action( 'vip_search_loaded' );
```
nickdaugherty:
```suggestion
do_action( 'vip_search_loaded' );

```

elasticsearch/includes/classes/Elasticsearch.php
```diff
@@ -21,10 +20,10 @@ protected function load_dependencies() {
 		 * Load ES Health command class
 		 */
 		if ( defined( 'WP_CLI' ) && WP_CLI ) {
-			require_once __DIR__ . '/commands/class-health-command.php';
+			require_once __DIR__ . '/commands/HealthCommand.php';
```
nickdaugherty: In general I do prefer this type of naming scheme, but the "WordPress way" is to prefix class files with `class-`, so we should follow that convention as end users will find it most familiar.

elasticsearch/includes/classes/Health.php
```diff
@@ -0,0 +1,136 @@
...
+	 * @param array $query_args Valid WP_Query criteria, mandatory fields as in following example:
+	 * $query_args = [
+	 *		'post_type' => $post_type,
+	 *		'post_status' => array( $post_statuses )
+	 * ];
+	 *
+	 * @param string $type Type (Slug) of the objects to be searched (should be either 'user' or 'post')
+	 * @return WP_Query
+	 */
+	private function query_objects( array $query_args, string $type ) {
```
nickdaugherty: Shouldn't this be static as well?
```suggestion
	private static function query_objects( array $query_args, string $type ) {
```

elasticsearch/includes/classes/Health.php
```diff
@@ -0,0 +1,136 @@
...
+		}
+
+		// Verify actual results
+		$es_total = (int) $es_result[ 'found_documents' ][ 'value' ];
+
+		if ( $db_total !== $es_total ) {
+			$diff = sprintf( ', diff: %d', $es_total - $db_total );
+		}
+
+		return [ 'entity' => $indexable->slug, 'type' => ( array_key_exists( 'post_type', $query_args ) ? $query_args[ 'post_type' ] : 'N/A' ), 'db_total' => $db_total, 'es_total' => $es_total, 'diff' => $diff ];
```
nickdaugherty: For readability, it would be best to format each array item on a new line.

elasticsearch/includes/classes/Health.php
```diff
@@ -0,0 +1,136 @@
...
+
+		$query_args = [
+			'order' => 'asc',
+		];
+
+		$result = self::verify_index_entity_count( $query_args, $users );
+		if ( is_wp_error( $result ) ) {
+			$result = [ 'entity' => $users->slug, 'type' => 'N/A', 'error' => $result->get_error_message() ];
+		}
+		return array( $result );
```
nickdaugherty: `$result` should always be an array anyway, right?

elasticsearch/includes/classes/Health.php
```diff
@@ -0,0 +1,136 @@
...
+	public static function verify_index_users_count( $args, $assoc_args ) {
+		$users = Indexables::factory()->get( 'user' );
+
+		$query_args = [
+			'order' => 'asc',
+		];
+
+		$result = self::verify_index_entity_count( $query_args, $users );
+		if ( is_wp_error( $result ) ) {
+			$result = [ 'entity' => $users->slug, 'type' => 'N/A', 'error' => $result->get_error_message() ];
```
nickdaugherty: Same suggestion about putting these on newlines for readability.

elasticsearch/includes/classes/Health.php
```diff
@@ -0,0 +1,136 @@
...
+			$query_args = [
+				'post_type' => $post_type,
+				'post_status' => array_values( $post_statuses ),
+			];
+
+			$result = self::verify_index_entity_count( $query_args, $posts );
+
+			// In case of error skip to the next post type
+			if ( is_wp_error( $result ) ) {
+				$result = [ 'entity' => $posts->slug, 'type' => $post_type, 'error' => $result->get_error_message() ];
```
nickdaugherty: Newlines here would be best

elasticsearch/includes/classes/Health.php
```diff
@@ -0,0 +1,136 @@
...
+	 *
+	 * @since   1.0.0
+	 * @access  public
+	 * @param array $query_args Valid WP_Query criteria, mandatory fields as in following example:
+	 * $query_args = [
+	 *		'post_type' => $post_type,
+	 *		'post_status' => array( $post_statuses )
+	 * ];
+	 *
+	 * @param mixed $indexable Instance of an ElasticPress Indexable Object to search on
```
nickdaugherty: Instead of `mixed`, we can reference the base class directly
```suggestion
	 * @param \ElasticPress\Indexable $indexable Instance of an ElasticPress Indexable Object to search on
```

elasticsearch/includes/classes/Health.php
```diff
@@ -0,0 +1,136 @@
...
+		if ( $db_total !== $es_total ) {
+			$diff = sprintf( ', diff: %d', $es_total - $db_total );
+		}
+
+		return [ 'entity' => $indexable->slug, 'type' => ( array_key_exists( 'post_type', $query_args ) ? $query_args[ 'post_type' ] : 'N/A' ), 'db_total' => $db_total, 'es_total' => $es_total, 'diff' => $diff ];
+	}
+
+	/**
+	 * Validate DB and ES index users counts
+	 */
```
nickdaugherty: Should we add matching docblock comments for this function?

elasticsearch/includes/classes/Health.php
```diff
@@ -0,0 +1,136 @@
...
+		$result = self::verify_index_entity_count( $query_args, $users );
+		if ( is_wp_error( $result ) ) {
+			$result = [ 'entity' => $users->slug, 'type' => 'N/A', 'error' => $result->get_error_message() ];
+		}
+		return array( $result );
+	}
+
+	/**
+	 * Validate DB and ES index post counts
+	 */
```
nickdaugherty: May as well add docblock here too

elasticsearch/elasticsearch.php
```diff
@@ -0,0 +1,51 @@
...
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+/**
+* PSR-4-ish autoloading
+*
+* @since 1.0
+*/
+spl_autoload_register(
+	function( $class ) {
```
nickdaugherty: For formatting, this can probably go on the line above (then we save one level of indention)

elasticsearch/elasticsearch.php
```diff
@@ -0,0 +1,51 @@
+<?php
+/**
+ * Plugin Name: Elasticsearch
```
nickdaugherty:
```suggestion
 * Plugin Name: VIP Search
```

elasticsearch/elasticsearch.php
```diff
@@ -0,0 +1,51 @@
...
+ * Description: VIP Elasticsearch plugin
```
nickdaugherty:
```suggestion
 * Description: Power your site search and other queries with Elasticsearch
```

elasticsearch/includes/classes/commands/HealthCommand.php
```diff
@@ -0,0 +1,99 @@
+<?php
+
+namespace Automattic\VIP\Elasticsearch;
```
nickdaugherty: Should the namespace be `Automattic\VIP\Elasticsearch\Commands`?

elasticsearch/includes/classes/Health.php
nickdaugherty: The corresponding call to this also needs changed if this is made static.

chrean: Right. Done in 7fa2750d.

elasticsearch/includes/classes/Health.php
chrean: 736d99bd

elasticsearch/elasticsearch.php
chrean: Done.

elasticsearch/includes/classes/commands/HealthCommand.php
chrean: Ooops. c88f5462

elasticsearch/includes/classes/Health.php
chrean: Yah added for return, removed `$args, $assoc_args` from those functions since they aren't used.
Kept them in the calling command, though, to make future extensions easier.

4f2b03a3

elasticsearch/class-elasticsearch.php
GaryJones: > Do we really need to have multiple instances of this class?

Likely not, but then, controlling the number of instantiations AND doing the logic that it contains, is two tasks, and therefore goes against the S in SOLID.

> I will work at refactoring this code to be more consistent and make more sense.

👍

elasticsearch/includes/classes/class-health.php
```diff
@@ -0,0 +1,154 @@
...
+		}
+		return array( $result );
+	}
+
+	/**
+	 * Validate DB and ES index post counts
+	 *
+	 * @return array Array containing entity (post/user), type (N/A), error, ES count, DB count, difference
+	 */
+	public static function verify_index_posts_count() {
```
GaryJones: The WP-CLI commands are called `validate-*`, but the corresponding logic are held in Health static methods called `verify_*` (but with DocBlock using the word Validate). Could these be consistent?

elasticsearch/elasticsearch.php
```diff
@@ -0,0 +1,49 @@
...
+* PSR-4-ish autoloading
+*
+* @since 1.0
+*/
+spl_autoload_register( function( $class ) {
+	// project-specific namespace prefix.
+	$prefix = 'Automattic\\VIP\\Elasticsearch\\';
+
+	// base directory for the namespace prefix.
+	$base_dir = __DIR__ . '/includes/classes/';
```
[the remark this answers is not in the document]
chrean: We don't know what will be there, yet.
But if we start with an extensible structure, we are off to a better start as we will be more flexible in the future.

elasticsearch/includes/classes/class-health.php
```diff
@@ -0,0 +1,154 @@
...
+	 * @return WP_Query
+	 */
+	private static function query_objects( array $query_args, string $type ) {
+		if ( 'user' === $type ) {
+			return new WP_User_Query( $query_args );
+		}
+		return new WP_Query( $query_args );
+	}
+
+}
```
nickdaugherty:
```suggestion
}

```

elasticsearch/elasticsearch.php
```diff
@@ -0,0 +1,49 @@
+<?php
+/**
+ * Plugin Name: VIP Search
+ * Description: Power your site search and other queries with Elasticsearch
+ * Version:     0.1
+ * Author:      Automattic VIP
+ * Author URI:  http://wpvip.com
+ * License:     GPLv2 or later
+ * Text Domain: vip-search
```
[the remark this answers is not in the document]
nickdaugherty: The naming is still up in the air a bit, so this is all subject to change.

nickdaugherty:
Turns out the ES maxes out at 10,000 for the 'total hits' (`found_documents`) result by default, which means that when there are more than 10,000 matching docs, the validator will receive a count of 10,000, which breaks the comparison.

This can be overridden with in the request:

https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking-changes-7.0.html#track-total-hits-10000-default

elasticsearch/includes/classes/class-health.php
nickdaugherty: Looks like this was done in 6aac735

nickdaugherty:
r754-stacks

mjangda:
> Turns out the ES maxes out at 10,000 for the 'total hits'

Fixed in #1437

nickdaugherty:
Thanks @mjangda !

PR #1415, opened 2020-01-22
Send error logs from vip_safe_wp_remote_request() to main logs

nickdaugherty:
## Description

Previously, `vip_safe_wp_remote_get()` called `error_log()`, which doesn't properly format the messages for sending to VIP's error log service. This changes that call to a `trigger_error()` (notice level), and adds another logging call that records when requests have been throttled.

## Checklist

## Steps to Test

Example:

1. Check out PR.
1. Add this to a test site, and ensure the right errors are logged (3 for timeouts, 2 for throttle):
```
for( $i = 0; $i < 5; $i++ ) {
	vip_safe_wp_remote_get( 'http://slowwly.robertomurray.co.uk/delay/2000/url/http://www.google.co.uk' );
}
```

### DIFF
--- a/vip-helpers/vip-utils.php
+++ b/vip-helpers/vip-utils.php
@@ -742,8 +742,9 @@ function vip_safe_wp_remote_request( $url, $fallback_value='', $threshold=3, $ti
 	$cache_key = 'disable_remote_request_' . md5( parse_url( $url, PHP_URL_HOST ) . '_' . $parsed_args[ 'method' ] );
 
 	// valid url
-	if ( empty( $url ) || !parse_url( $url ) )
+	if ( empty( $url ) || !parse_url( $url ) ) {
 		return ( $fallback_value ) ? $fallback_value : new WP_Error('invalid_url', $url );
+	}
 
 	// Ensure positive values
 	$timeout   = abs( $timeout );
@@ -761,8 +762,9 @@ function vip_safe_wp_remote_request( $url, $fallback_value='', $threshold=3, $ti
 
 	// check if the timeout was hit and obey the option and return the fallback value
 	if ( false !== $option && time() - $option['time'] < $retry ) {
-		if ( $option['hits'] >= $threshold )
+		if ( $option['hits'] >= $threshold ) {
 			return ( $fallback_value ) ? $fallback_value : new WP_Error('remote_request_disabled', 'Remote requests disabled: ' . maybe_serialize( $option ) );
+		}
 	}
 
 	$start = microtime( true );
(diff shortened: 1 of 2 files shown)


vip-helpers/vip-utils.php
```diff
@@ -771,24 +775,26 @@ function vip_safe_wp_remote_request( $url, $fallback_value='', $threshold=3, $ti
...
+		} else {
 			wp_cache_delete( $cache_key, $cache_group);
+		}
 	}
 
 	if ( is_wp_error( $response ) ) {
 		// Log errors for internal WP.com debugging
 		if ( ! defined( 'WPCOM_VIP_DISABLE_REMOTE_REQUEST_ERROR_REPORTING' ) || ! WPCOM_VIP_DISABLE_REMOTE_REQUEST_ERROR_REPORTING ) {
-			error_log( "vip_safe_wp_remote_request: Blog ID {$blog_id}: Requesting $url with a timeout of $timeout failed. Result: " . maybe_serialize( $response ) );
+			trigger_error( "vip_safe_wp_remote_request: Blog ID {$blog_id}: Requesting $url with method {$parsed_args[ 'method' ]} and a timeout of $timeout failed. Result: " . maybe_serialize( $response ), E_USER_NOTICE );
```
mjangda: I think by default, WordPress force the error reporting level to ignore notices so I don't think these will actually make it to Logstash.

We will probably want to use `E_USER_WARNING` instead.

```suggestion
			trigger_error( "vip_safe_wp_remote_request: Blog ID {$blog_id}: Requesting $url with method {$parsed_args[ 'method' ]} and a timeout of $timeout failed. Result: " . maybe_serialize( $response ), E_USER_WARNING );
```

vip-helpers/vip-utils.php
```diff
@@ -761,8 +762,11 @@ function vip_safe_wp_remote_request( $url, $fallback_value='', $threshold=3, $ti
 
 	// check if the timeout was hit and obey the option and return the fallback value
 	if ( false !== $option && time() - $option['time'] < $retry ) {
-		if ( $option['hits'] >= $threshold )
+		if ( $option['hits'] >= $threshold ) {
+			trigger_error( "vip_safe_wp_remote_request: Blog ID {$blog_id}: Requesting $url with method {$parsed_args[ 'method' ]} has been throttled after {$option['hits']} attempts. Not reattempting until after $retry seconds", E_USER_NOTICE );
```
mjangda: Same note about notice vs warning:

```suggestion
			trigger_error( "vip_safe_wp_remote_request: Blog ID {$blog_id}: Requesting $url with method {$parsed_args[ 'method' ]} has been throttled after {$option['hits']} attempts. Not reattempting until after $retry seconds", E_USER_WARNING );
```

vip-helpers/vip-utils.php
nickdaugherty: I checked Logstash and found lots of Notices so they do make it, but a warning might be better anyway?

nickdaugherty:
Crud, just realized the tests are broken b/c we're trigger errors now :(

Will have to do something like this: https://stackoverflow.com/a/1227686

vip-helpers/vip-utils.php
nickdaugherty: 👍 1e0e8ad

vip-helpers/vip-utils.php
nickdaugherty: 👍 1e0e8ad

nickdaugherty:
Went a different way to disable the errors - just set the `WPCOM_VIP_DISABLE_REMOTE_REQUEST_ERROR_REPORTING` constant.

vip-helpers/vip-utils.php
nickdaugherty: On second thought, I'm not sure `Warning` is the right level for this. It could go either way, really, I just know we'll cause an increase in global warnings with this change, when there's not really anything actionable (which is more of a notice).

mjangda:
The main downside to the constant suppressing errors is that we're no longer testing that the errors are triggering as expected.

The annotation approach is pretty simple to implement, e.g.

```
 * @expectedException PHPUnit_Framework_Error
or
 * @expectedException PHPUnit_Framework_Warning
or
 * @expectedException PHPUnit_Framework_Notice
```

Anyway, not super critical (and I guess the constant already existed).

nickdaugherty:
The other thing is that even if you handle the errors correctly in PHPUnit, they still print out to the console, which pollutes test output. This is probably something that can be solved with error reporting levels (or our inside our custom error handler), but seemed more trouble that it was worth for this PR.

mjangda:
Makes sense. We can run with this as-is and then iterate if needed.

nickdaugherty:
I'm a little concerned about the extra logging from the early return (where it wasn't logged before)...we currently don't have a good idea how often this is happening in prod and I'm a little leery about blowing up the error logs.

nickdaugherty:
r751-stacks

PR #1424, opened 2020-01-27
Block API requests for users with 2fa enabled

mjangda:
Except for a small subset of sites that are transitioning.

Note: `VIP_2FA_ALLOW_API_LOGIN_ENV_IDS` will be added as part of the deploy.

## Checklist

- n/a This change has relevant unit tests (if applicable) (skipping, since this is transitional)

## Steps to Test

1. Apply patch.
1. Make an API request to a site with a user account that has 2fa enabled (see internal P2 for example). It should fail with a 403 (`Incorrect username or password.`)
1. Add `VIP_2FA_ALLOW_API_LOGIN_ENV_IDS` constant with the ID of the current site.
1. Repeat the request. It should pass.


mjangda:
r745-stacks

PR #1425, opened 2020-01-27
$enable => true

mjangda:
Continued from #1424 

We need to actually allow the requests for now rather than passing back the default which is `false`.


mjangda:
r745-stacks

PR #1426, opened 2020-01-28
Streamwrapper: Upload file to fileservice on fopen

hanifn:
## Description

Upload a new file to the fileservice on `fopen()` as that is the expected behaviour for the `fopen()` function. This was causing problems with some plugins where they were checking the file with `is_file()` or `is_writable()` after `fopen()` was called.

## Checklist

## Steps to Test

1. Check out PR.
1. Deploy changes on a test site
1. `wp shell`
1. `fopen( 'vip://wp-content/uploads/testfile', 'w' );`
1. ensure `is_file()` and `is_writable()` on above file returns true.
1. `fclose( $handle );`
1. Ensure file is uploaded to file service
1. ensure existing filesystem functionality still works


files/class-vip-filesystem-stream-wrapper.php
```diff
@@ -105,6 +114,11 @@ public function __construct( API_Client $client = null, $protocol = null ) {
 		}
 
 		$this->protocol = $protocol ?: self::DEFAULT_PROTOCOL;
+
+		$this->debug_mode = false;
+		if ( defined( 'WP_DEBUG' ) ) {
```
mjangda: Since WP_DEBUG can be set to false:

```suggestion
		if ( defined( 'WP_DEBUG' ) && true === WP_DEBUG ) {
```

files/class-vip-filesystem-stream-wrapper.php
```diff
@@ -105,6 +114,11 @@ public function __construct( API_Client $client = null, $protocol = null ) {
 		}
 
 		$this->protocol = $protocol ?: self::DEFAULT_PROTOCOL;
+
+		$this->debug_mode = false;
+		if ( defined( 'WP_DEBUG' ) && true === WP_DEBUG ) {
+			$this->debug_mode = WP_DEBUG;
```
mjangda:
```suggestion
			$this->debug_mode = true;
```

files/class-vip-filesystem-stream-wrapper.php
```diff
@@ -628,6 +686,8 @@ public function stream_metadata( $path, $option, $value ) {
 	 * @return  resource|bool
 	 */
 	public function stream_cast( $cast_as ) {
+		$this->debug( sprintf( 'stream_stat =>  %s + %s + %s', $cast_as, $this->path, $this->uri ) );
```
mjangda:
```suggestion
		$this->debug( sprintf( 'stream_cast =>  %s + %s + %s', $cast_as, $this->path, $this->uri ) );
```

hanifn:
Thanks @mjangda ! Tested this and it looks like it's working as expected now.

@darthhexx Could use you opinion on this. This would double the `PUT` requests for new file uploads but the first requests would be zero bytes. Would this be problematic from the filesystem perspective?

nickdaugherty:
Thinking aloud - an alternative would be to keep it as a temp file, keep a map of files that are currently only temp files, then when is_file(), etc is called with that filename, bypass the API and check the local temp file.

That approach prevents uploading empty files to the files service, though adds complexity and the potential for edge-cases (maybe they really expected to create a 0 byte file and have it persist, for reasons).

hanifn:
> Thinking aloud - an alternative would be to keep it as a temp file, keep a map of files that are currently only temp files, then when is_file(), etc is called with that filename, bypass the API and check the local temp file.

Maybe we can make use of the `file_stat` cache in the `API_Cache` class. On a new file, instead of uploading an empty file, we just update the `file_stat` and `file` cache so functions like `is_file()` and `is_writable()` would still return true. Although, this cache is valid only per request I'm concerned what happens if they never do anything to that file but still expect it to exist in the next request.

darthhexx:
I would definitely suggest that we keep a local temp copy and update the `file_stat` cache in `API_Cache` in the processing request. At the end of the request you would persist any buffered file contents to the Files Service. That way all the local file operations are low latency and any subsequent request would find the contents, zero or not, present in the files service.

hanifn:
@nickdaugherty @darthhexx I've refactored the code to _not_ upload the file on `fopen()` but instead use the local temp file + `file_stats` cache. The stream wrapper will then flush the file to the File Service, even if it's empty, on `fclose()`.

files/class-vip-filesystem-stream-wrapper.php
```diff
@@ -194,6 +239,14 @@ public function stream_open( $path, $mode, $options, &$opened_path ) {
 	 * @access  public
 	 */
 	public function stream_close() {
+		$this->debug( sprintf( 'stream_close => %s + %s', $this->path, $this->uri ) );
+
+		if ( $this->should_flush_empty ) {
+			$this->stream_flush();
+
+			$this->should_flush_empty = false;
```
darthhexx, line 247: You already have a `$this->should_flush_empty = false;` set on one code path in `stream_flush()` and IMO this _should_ be encapsulated in `stream_flush()` and `stream_write()` and not set here. Unless there is some reason to blanket set it no matter whether `stream_flush()` succeeds?

hanifn: Ah good catch. This was a leftover from a previous attempt to trigger the flush using a destructor method which didn't really work. I'll remove this

hanifn: Removed in 98290455510c6d085003584d3ab452dfea18bc3c

hanifn:
r879-stacks

PR #1427, opened 2020-01-30
Temporarily disable ssl verification on requests to VIP ES server

nickdaugherty:
## Description

ES server is using a self-signed cert during testing phase, will be replaced with real cert in near future then this can be removed.

## Checklist

## Steps to Test

1. Try query operation - should succeed
1. Try indexing operation - should succeed


nickdaugherty:
r749-stacks

PR #1428, opened 2020-01-30
Clean up leftover _encloseme

rebeccahum:
## Description

We disable the hook `do_pings` for performance reasons due to the query it runs on every `publish_post` hook:
https://github.com/Automattic/vip-go-mu-plugins/blob/da7fcd24b773e1957687c6eb56f0b56ddd402d9e/performance/do-pings.php#L12 

Which causes the `do_all_pings()` cleanup function to not run anymore: https://github.com/WordPress/WordPress/blob/1e925a5ae3faccad31c48c810af3954d01a60f0e/wp-includes/comment.php#L2626

Unfortunately, that lead to sites with potentially hundreds of thousands rows with the `_encloseme` key on the postmeta table and causing performance issues on simple meta queries like:

```
SELECT post_id, meta_key, meta_value FROM wp_postmeta WHERE post_id IN (x) ORDER BY meta_id ASC
```

We've fixed the duplicate `_encloseme` postmeta as of 5.3: https://core.trac.wordpress.org/ticket/48014#ticket

However, there are still sites with the duplicate postmeta rows in the DB that should be cleaned up.  

## Checklist

## Steps to Test

1) Set `VIP_CLEANUP_ENV` to `true` in vip-config.
2) Run cron job and viola!


nickdaugherty:
@rebeccahum is this ready for review? If so, can you please add the "Needs Review" label?

rebeccahum:
Done, thanks!

lib/vip-encloseme-cleanup.php
```diff
@@ -0,0 +1,80 @@
+<?php
+
+class VIP_Encloseme_Cleanup {
+    const OPTION_NAME = 'vip_encloseme_cleanup';
+    const CRON_HOOK = 'vip_encloseme_cleanup_hook';
+    const CRON_INTERVAL = 'vip_cleanup_interval';
```
mjangda:
```suggestion
    const CRON_INTERVAL = 'vip_encloseme_cleanup_interval';
```

lib/vip-encloseme-cleanup.php
```diff
@@ -0,0 +1,80 @@
...
+        ];
+        
+        return $schedule;
+    }
+
+    public static function schedule_cleanup() {
+        $already_ran = get_option( self::OPTION_NAME );
+
+        if ( $already_ran && wp_next_scheduled( self::CRON_HOOK ) ) {
+            wp_clear_scheduled_hook( self::CRON_HOOK );
```
mjangda:
```suggestion
            wp_clear_scheduled_hook( self::CRON_HOOK );
            return;
```

lib/vip-encloseme-cleanup.php
```diff
@@ -0,0 +1,80 @@
...
+                delete_post_meta( $pid, '_encloseme' );
+                if ( defined( 'DISABLE_VIP_CLEANUP_ENV' ) && true === DISABLE_VIP_CLEANUP_ENV ) {
+                    break 2;
+                }
+            }
+    
+            sleep( 3 );
+
+            $meta_exists = $wpdb->get_var( $find_meta_query );
+        } while ( ! is_null( $meta_exists ) && false === defined( 'DISABLE_VIP_CLEANUP_ENV' ) );
```
mjangda: Infinite loops are scary :) I'm also not sure we need it with the cron-based approach?

If we do want to process a larger batch size per cron event, it may be better to use a `for` loop:

```
// 5 is arbitrary here
for ( $count = 0; $count < 5; $count++ ) {
    // check constant and bail if not set

    // get 1000 entries

   // if not found, bail

   // cleanup

   // sleep
}
```

rebeccahum: @mjangda What about for sites that range from either 100 rows that need to be deleted versus 500,000+ rows that need to be deleted? I'm not sure what we would set the batch size to with such different scenarios.  Or would we just set it to a very high number?

mjangda: We can set it high enough and frequent enough so that it doesn't take days/months to deal with really large sites but not too high or too frequent to cause issues.

For example, If we process 10K entries per event, we'll be finished a 500K site in ~16 hours (`500000 / 10000 / 3` -- 3 because we're running 3 times an hour). We can drop the frequency to 15 mins and be done in 12.5 hours instead.

lib/class-vip-encloseme-cleanup.php
```diff
@@ -0,0 +1,94 @@
...
+			wp_schedule_event( time(), self::CRON_INTERVAL, self::CRON_HOOK );
+		}
+	}
+
+	public static function cleanup_encloseme_meta() {
+		if ( ! wp_doing_cron() ) {
+			return;
+		}
+
+		if ( defined( 'DISABLE_VIP_CLEANUP_ENV' ) && true === DISABLE_VIP_CLEANUP_ENV ) {
```
mjangda, line 49:
```suggestion
		if ( defined( 'DISABLE_VIP_ENCLOSEME_CLEANUP' ) && true === DISABLE_VIP_ENCLOSEME_CLEANUP ) {
```

mjangda: Also, we should probably move this to the `init` method.

lib/class-vip-encloseme-cleanup.php
```diff
@@ -0,0 +1,92 @@
...
+				break; // Bail, no post IDs have been found.
+			}
+
+			$pids = array_map( function( $pid ) {
+				return $pid[0];
+			}, $pids );
+			
+			foreach ( $pids as $pid ) {
+				delete_post_meta( $pid, '_encloseme' );
+			}
```
mjangda, line 87: Remove unnecessary `array_map`:

```suggestion
			foreach ( $pids as $pid ) {
				delete_post_meta( $pid[0], '_encloseme' );
			}
```

mjangda:
Note that after this is deployed, we can incrementally start enabling on sites.

brandon-m-skinner:
r910-stacks

PR #1430, opened 2020-01-31
vip migration cleanup: Reset user roles before trying to connect Jetpack

rinatkhaziev:
In order for Jetpack to be connected the machine user has to exist and be an administrator.

However, if administrator role is missing `jetpack-start connect` will fail to create the user. 

This commit addresses that by introducing `reset_user_roles` function which calls the `wp role reset --all`

### DIFF
--- a/vip-helpers/vip-migrations.php
+++ b/vip-helpers/vip-migrations.php
@@ -48,6 +48,7 @@ function run_after_data_migration_cleanup() {
 
 	wp_cache_flush();
 
+	reset_user_roles();
 	connect_jetpack();
 	connect_vaultpress();
 	connect_akismet();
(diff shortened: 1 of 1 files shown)


mjangda:
This might cause problems if we import a database with custom roles already defined?

rinatkhaziev:
Not according to my testing, our two custom support roles are not affected when running it. But I can test again, or can add a flag to only run it for WPCOM->VIP Go migrations.

rinatkhaziev:
Actually I think it's fine as-is. The command description explicitly states that this resets only default roles:

https://developer.wordpress.org/cli/commands/role/reset/

mjangda:
> can add a flag to only run it for WPCOM->VIP Go migrations

I think this might be better.

rinatkhaziev:
> 
> I think this might be better.

This will be trickier though, because it'll require changes to the import endpoint, I'm not really sure if it's warranted. Here's the source of the `wp role reset` it definitely only touches the default roles:
https://github.com/wp-cli/role-command/blob/bad54a1b02331ee6460cc6a6f967e37dd91e07a3/src/Role_Command.php

uxcitizen:
> This might cause problems if we import a database with custom roles already defined?

In the past I frequently had to run `wp role reset --all` after imports, as roles went missing. 

Only once did it cause an issue that I am aware of and that was for #vip-client-digital-first-media - they had some custom logic on the default-named roles - 

https://github.com/wpcomvip/digital-first-media/blob/master/themes/wp-mason/inc/cli.php
Ticket: https://wordpressvip.zendesk.com/agent/tickets/82646

mjangda:
Going to close this as we're doing it during the migration process now. Can revisit if it's still causing problems.

PR #1431, opened 2020-01-31
Enforce the jetpack_active_plan option

mjangda:
To a default that we expect for VIP. This will avoid issues with the option getting corrupted and breaking Jetpack features like Search.

props @mikeyarce and @brettshumaker for the initial batch of work on this.

## Checklist

## Steps to Test

1. Check out PR.
1. `get_option( 'jetpack_active_plan' )` should return the correct object.
1. Jetpack functionality like Search and Dashboard should continue to work.

### DIFF
--- a/vip-jetpack/vip-jetpack.php
+++ b/vip-jetpack/vip-jetpack.php
@@ -164,7 +164,9 @@ add_filter( 'jetpack_get_available_modules', function( $modules ) {
  * which can then break features like Jetpack Search.
  */
 add_filter( 'pre_option_jetpack_active_plan', function( $pre_option, $option, $default ) {
-	if ( true === WPCOM_IS_VIP_ENV && defined( 'VIP_JETPACK_DEFAULT_PLAN' ) ) {
+	if ( true === WPCOM_IS_VIP_ENV
+		&& defined( 'VIP_JETPACK_DEFAULT_PLAN' )
+		&& Jetpack::is_active() ) {
 		return VIP_JETPACK_DEFAULT_PLAN;
 	}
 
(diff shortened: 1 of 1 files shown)


vip-jetpack/vip-jetpack.php
```diff
@@ -72,6 +155,22 @@
...
+/**
+ * Prevent the jetpack_active_plan from ever being overridden.
+ *
+ * All sites on VIP Go should always have have a valid VIP plan.
+ *
+ * This will prevent issues from the plan option being corrupted,
+ * which can then break features like Jetpack Search.
+ */
+add_filter( 'pre_option_jetpack_active_plan', function( $pre_option, $option, $default ) {
+	if ( true === WPCOM_IS_VIP && defined( 'VIP_JETPACK_DEFAULT_PLAN' ) ) {
```
mikeyarce: Should this be `WPCOM_IS_VIP_ENV`?

mjangda: Yep, fixed.

mikeyarce:
The only potential issue I see is if a site hasn't connected to Jetpack yet to get the option, then the option would not be in the cache, `alloptions`, or in the site's database.   If there was anything looking for this option that didn't use `get_option`, and relied on cache, `alloptions`, or the database, this would be a problem.   

What about adding a check to see if the option exists, and if it doesn't we, just set the option as well?

mjangda:
>  If there was anything looking for this option that didn't use get_option

That seems like a very much a `_doing_it_wrong` and we likely have bigger problems if that's happening :)

We may need to account for a not-yet-connected site though and pass through the value instead.

mikeyarce:
> That seems like a very much a _doing_it_wrong and we likely have bigger problems if that's happening :)

Agreed!

> We may need to account for a not-yet-connected site though and pass through the value instead.

Also agreed!

mjangda:
r752-deploy

PR #1433, opened 2020-02-03
chore: add note about turning xdebug on in vvv prior to running tests

brandon-m-skinner:
## Description

Added text stating that xdebug must be enabled in VVV prior to running tests if you get an error stating that an xdebug function is undefined.

## Checklist

## Steps to Test

Attempt to run tests as outlined in the README without xdebug enabled on VVV. 

You should get an error of `Error: Call to undefined function Automattic\VIP\Cache\xdebug_get_headers()` or similar.

Turn xdebug on in VVV and then try to run tests again. All tests should pass successfully.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,11 @@ composer install
 vagrant@vvv:/wp-content/mu-plugins$ vendor/bin/phpunit
 ```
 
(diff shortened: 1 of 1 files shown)


mjangda:
Aside: I wonder if we should drop this section? VVV is kind of annoying to use and we should standardize our local dev environments on Docker anyway (we should document that).

PR #1436, opened 2020-02-04
Add Elasticsearch health check cronjob

hanifn:
## Description

Add a health check cronjob for sites using VIP Search

## Checklist

## Steps to Test

TBD

### DIFF
--- a/elasticsearch/includes/classes/class-health-job.php
+++ b/elasticsearch/includes/classes/class-health-job.php
@@ -24,13 +24,25 @@ class HealthJob {
 	const CRON_INTERVAL = 60 * 30;
 
 	/**
+	 * Initialize the job class
+	 *
+	 * @access	public
+	 */
+	public function init() {
+		// Add the custom cron schedule
+		add_filter( 'cron_schedules', [ $this, 'filter_cron_schedules' ], 10, 1 );
+
+		$this->schedule_job();
+	}
+
+	/**
 	 * Schedule health check job
 	 *
 	 * Add the event name to WP cron schedule and then add the action
 	 */
 	public function schedule_job() {
 		if ( ! wp_next_scheduled( self::CRON_EVENT_NAME )  ) {
-			wp_schedule_event( time(), '', self::CRON_EVENT_NAME );
+			wp_schedule_event( time(), self::CRON_INTERVAL_NAME, self::CRON_EVENT_NAME );
 		}
 
 		add_action( self::CRON_EVENT_NAME, [ $this, 'check_health' ] );
--- a/elasticsearch/includes/classes/class-health-job.php
+++ b/elasticsearch/includes/classes/class-health-job.php
@@ -18,10 +18,8 @@ class HealthJob {
 
 	/**
 	 * Custom cron interval value
-	 *
-	 * 30 minutes in seconds
 	 */
-	const CRON_INTERVAL = 60 * 30;
+	const CRON_INTERVAL = 60 * 30; // 30 minutes in seconds
 
 	/**
 	 * Initialize the job class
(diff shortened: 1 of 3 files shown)


elasticsearch/includes/classes/class-health-job.php
```diff
@@ -0,0 +1,121 @@
...
+
+	/**
+	 * The name of the scheduled cron event to run the health check
+	 */
+	const CRON_EVENT_NAME = 'vip_search_healthcheck';
+
+	/**
+	 * Custom cron interval name
+	 */
+	const CRON_INTERVAL_NAME = 'vip_half_hour';
```
mjangda: I wonder if we should call this something generic so that it's easier to change the time, if needed?

```suggestion
	const CRON_INTERVAL_NAME = 'vip_search_healthcheck_interval';
```

hanifn: Shouldn't we also change the interval display text too?

mjangda: Yep, good though 👍

hanifn: Changed in 184041da639857bf542743e1bd729c5d043705d2

elasticsearch/includes/classes/class-health-job.php
```diff
@@ -0,0 +1,121 @@
...
+	/**
+	 * Check ElasticSearch index health
+	 */
+	public function check_health() {
+		$user_results = Health::validate_index_users_count();
+
+
+		$this->process_results( $user_results );
+
+		$post_results = Health::validate_index_posts_count();
```
mjangda: Aside: I wonder if in the `Health` class we should just have a single entry point that runs the various validations (`Health::validate()`)? That way if (and when) we add additional checks (e.g. cluster health, index exists, etc.), they're automatically checked here as well.

Not blocking, but definitely something we should consider.

elasticsearch/includes/classes/class-health-job.php
```diff
@@ -0,0 +1,121 @@
...
+	/**
+	 * Process the health check result
+	 *
+	 * @access	protected
+	 * @param		array		$result		Array of results from Health index validation
+	 */
+	protected function process_results( $result ) {
+		// If there's an error, alert
+		if( array_key_exists( 'error', $results ) ) {
+			wpcom_vip_irc(
```
mjangda: Aside: we may want to consider a few things related to this monitoring (outside this PR):

1) do we want to monitor and alert for non-production sites?
2) do we want a way to turn off alerts globally?
3) How do we monitor that this is still active and running (some sort of heartbeat)?

elasticsearch/includes/classes/class-health-job.php
```diff
@@ -0,0 +1,118 @@
+<?php
+
+namespace Automattic\VIP\Elasticsearch;
+
+use Automattic\VIP\Elasticsearch\Health as Health;
```
mjangda: I'm not sure if this is really needed since they're both in the same namespace?

mjangda:
r759-stacks

elasticsearch/includes/classes/class-health-job.php
hanifn: I added that in because it was giving a class not found error before but you're right, it shouldn't be needed if it's in the same namespace. Must have done something else wrong

elasticsearch/includes/classes/class-health-job.php
hanifn: > do we want to monitor and alert for non-production sites?
I don't think we should. This could generate a lot noise that buries problems in production sites. 

> do we want a way to turn off alerts globally?
You mean a constant that can disable alerts for a site?

> How do we monitor that this is still active and running (some sort of heartbeat)?
We can do that but I am worried about the noise. Perhaps it goes to a different channel?

PR #1437, opened 2020-02-05
Get accurate total count from Elasticsearch

mjangda:
By default, it stops at 10,000. We can ask for an exact number by setting the `track_total_hits` arg.

See https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking-changes-7.0.html#track-total-hits-10000-default

## Checklist

## Steps to Test

- On a test site with indexed documents more than 10K, run `wp vip-es health validate-posts-count`
- Verify that diff doesn't error.


mjangda:
r757-deploy

PR #1438, opened 2020-02-05
ES Health: Cleanup diff output

mjangda:
The `diff` returned by the health class should be an int and the calling class should decide what to do with it.

Right now, we're sending a string which makes the value less useful and makes the output somewhat awkward: `(DB: 5789, ES: 0, Diff: , diff: -5789)`

With this change, we're much cleaner: `(DB: 5789, ES: 0, Diff: -5789)`

## Checklist

## Steps to Test

- Run `wp vip-es health validate-posts-count`
- Verify that the diff output looks clean.


mjangda:
r758-deploy

PR #1439, opened 2020-02-06
IRC/Slack alert when cron control cron option is uncacheable

brandon-m-skinner:
## Description

[PLAT-888]

Added IRC/Slack alert for uncacheable cron options in cron-control. 

Still have to test this properly on a Vip Go sandbox and explore mu-plugins a bit more since it still feels a bit rough to me.

## Checklist

## Steps to Test

Checkout the PR and either: 
- run a  `do_action` on `a8c_cron_control_uncacheable_cron_option`; or
- make the cron option too big(overload the `a8c_cron_control_jobs` database table) or lower the `MAX_CACHE_BUCKET` value.

[PLAT-888]: https://vipjira.atlassian.net/browse/PLAT-888

### DIFF
--- a/001-cron.php
+++ b/001-cron.php
@@ -118,7 +118,7 @@ function wpcom_vip_cron_control_event_object_to_string( $event ) {
  */
 function wpcom_vip_log_cron_control_uncacheable_cron_option( $option_size, $buckets, $option_flat_count ) {
 	$message = sprintf( 'Cron Control Cron Option Uncacheable Alert - home: %s | option size: %d | buckets: %d | option flat count: %d', home_url( '/' ), $option_size, $buckets, $option_flat_count );
-	wpcom_vip_irc( 'vipv2-alerts', $message, 2, '', 5 );
+	wpcom_vip_irc( 'vipv2-alerts', $message, 2, 'cache-control-uncacheable-cron-option', 900 );
 }
 
 /**
--- a/001-cron.php
+++ b/001-cron.php
@@ -112,6 +112,18 @@ function wpcom_vip_cron_control_event_object_to_string( $event ) {
 }
 
 /**
+ * Callback for 'a8c_cron_control_uncacheable_cron_option' action. Send an alert to IRC and Slack in case of cron option being too big.
+ *
+ * @param $event object
+ */
+function wpcom_vip_log_cron_control_uncacheable_cron_option( $option_size, $buckets, $option_flat_count ) {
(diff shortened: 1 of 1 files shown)


brandon-m-skinner:
I think I figured out the IRC message routing and clearly the sandbox doesn't have it. 

```
Error sending IRC message (Cron Control Cron Option Uncacheable Alert - home: [url redacted] | option size: 100 | buckets: 200 | option flat count: 300): cURL error 7: Failed to [url redacted]: Connection refused
```

I'm not 100% on what else I can really do with this one with regards to testing and such. I've got a bit more experience with sandboxes and messing around with mu-plugins, but I don't think it makes a lot of sense to continue banging into a wall on this without any idea as to what I'm doing or what sort of expectations surround this sort of thing. I can't find much on this and the code looks basically the same as when I added it within 5 minutes of looking at the issue and looking up the add/do action WordPress stuff.

brandon-m-skinner:
I'd guess system imposed rate limiting with spam and an accusatory, sharp message from systems.

Judging by the rest of the usages of the `wpcom_vip_irc`, I'm guessing the interval value is what's used to prevent spamming. 

A sensible value would be every 5 seconds I think. It's probably not a super critical alert and pretty prone to spamming if someone crosses the threshold.

001-cron.php
```diff
@@ -111,6 +111,16 @@ function wpcom_vip_cron_control_event_object_to_string( $event ) {
...
 }
 
+/**
+ * Callback for 'a8c_cron_control_uncacheable_cron_option' action. Send an alert to IRC and Slack in case of cron option being too big.
+ *
+ * @param $event object
+ */
+function wpcom_vip_log_cron_control_uncacheable_cron_option( $option_size, $buckets, $option_flat_count ) {
+	$message = sprintf( 'Cron Control Cron Option Uncacheable Alert - home: %s | option size: %d | buckets: %d | option flat count: %d', home_url( '/' ), $option_size, $buckets, $option_flat_count );
+	wpcom_vip_irc( 'vipv2-alerts', $message, 2, '', 5 );
```
nickdaugherty: Should we have a value for the 4th parameter here?

nickdaugherty: Instead of 1 message every 5 seconds, which would still be a _lot_ of alert messages going off, we should dial it back to about once every 15 minutes.

brandon-m-skinner:
r774-stacks

PR #1440, opened 2020-02-06
Add PHPCS Diffing

mjangda:
Based on @cojennin's work on Edit Flow: https://github.com/Automattic/Edit-Flow/pull/559

Generates a PHPCS report and then filters out any issues on code that wasn't touched and does not flag those.

The report is set to not fail for now until we see how well it's working.

## Checklist

CI-only change, so, the checklist isn't applicable.

## Steps to Test

- `composer install`
- `./bin/phpcs-diff.sh`

Should see something like:

```
== Generating diff against master: /tmp/tmp.6hWvZdgTSk
== Generating phpcs report: /tmp/tmp.R5DDdMYQ0U
== Running diff
100.00% Covered
```

Also, check that Circle CI's PHP 7.3 + Multisite runs this successfully (https://circleci.com/gh/Automattic/vip-go-mu-plugins/1946). Should only run for that one build.


chrean:
Thanks for taking care of this @mjangda.

I wanted to test this on actual syntax inconsistencies, so I have edited a file by adding random spaces and blank lines; however, the script did not detect anything.

When running phpcs manually, and specifying the `WordPress` coding standard, errors were correctly detected:

```
./vendor/bin/phpcs --standard=WordPress elasticsearch/elasticsearch.php

FILE: /Users/quasistar/wp-local-docker-sites/quasistar-single-test/wordpress/wp-content/mu-plugins/elasticsearch/elasticsearch.php
----------------------------------------------------------------------------------------------------------------------------------
FOUND 4 ERRORS AFFECTING 3 LINES
----------------------------------------------------------------------------------------------------------------------------------
 25 | ERROR | [x] Space before opening parenthesis of function call prohibited
 25 | ERROR | [x] Opening parenthesis of a multi-line function call must be the last content on the line
 26 | ERROR | [x] Functions must not contain multiple empty lines in a row; found 5 empty lines
 52 | ERROR | [x] Closing parenthesis of a multi-line function call must be on a line by itself
----------------------------------------------------------------------------------------------------------------------------------
PHPCBF CAN FIX THE 4 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------------------------------------------------------------------

Time: 591ms; Memory: 10MB
```

How do you suggest we fix this? Should we explicitly specify a coding standard when we invoke phpcs, or?

mjangda:
We likely need to update the standard we're using in our phpcs.xml.dist file.

chrean:
Added here: c2a9ed93

GaryJones:
The `WordPress` standard should not be used - explicitly use `WordPress-Core` (or `WordPress-Extra` which is `WordPress-Core` plus opinionated extras) + `WordPress-Docs`.

The WPCS package contains about three sniffs that may not be up to the quality of the other sniffs yet, but aren't referenced in the rulesets. If you use `WordPress`, you'll be including them.

chrean:
Good catch @GaryJones , addressed in af9b03a3 .

PR #1442, opened 2020-02-06
Shared plugins: Add defensive code to fix Warning

GaryJones:
## Description

Ensures that if a value stored in `wpcom_vip_active_plugins` is not an array, that the subsequent `foreach` does not run. Without it, "Invalid argument supplied for foreach" PHP Warnings are generated (currently ~3700 per day on VIP Go).

Fix could optionally also include a ` || empty( retired_plugins_option )` in the conditional as well, though the `foreach` should not through a warning for an empty array. 

## Checklist

## Steps to Test

1. `wp option set wpcom_vip_active_plugins notanarray`
1. Load site and check logs for "Invalid argument supplied for foreach" PHP Warnings.
1. Apply PR
1. Load site again, and check for no new PHP Warnings.

### DIFF
--- a/vip-plugins/vip-plugins.php
+++ b/vip-plugins/vip-plugins.php
@@ -273,6 +273,10 @@ add_action( 'admin_enqueue_scripts', 'wpcom_vip_plugins_ui_admin_enqueue_scripts
 function wpcom_vip_include_active_plugins() {
 	$retired_plugins_option = get_option( 'wpcom_vip_active_plugins', array() );
 
+	if ( ! is_array( $retired_plugins_option ) ) {
+		return;
+	}
+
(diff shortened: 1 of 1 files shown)


nickdaugherty:
Thanks for this @GaryJones - do you have a sense of what's causing the non-array value in the first place?

GaryJones:
If `get_option( 'wpcom_vip_active_plugins', array() );` returns a non-array value (either from a value stored in the DB, or from a pre get option filter, then the Warning will be thrown.

hanifn:
r771-stacks

GaryJones:
Confirming the Warnings have stopped:

Screenshot 2020-02-12 at 12 20 28

nickdaugherty:
Thanks @GaryJones !

PR #1445, opened 2020-02-06
Limit calls to `Two_Factor_Core::is_user_using_two_factor()`

### LINKED ISSUE #1444: Invalid User ID when updating user profile fields.
There has been a report of an issue with users being unable to edit any user profile field except 2FA fields on `/wp-admin/profile.php`. Disabling the 2FA plugin with `add_filter( 'wpcom_vip_enable_two_factor', '__return_false' );` did solve the issue and helped us isolate the problem to something dealing with the 2FA plugin.

## Expected/Desired Behavior

1. 2FA plugin is enabled
2. User should be able to successfully edit any field at `/wp-admin/profile.php`

## Actual Behavior

Currently, on some environments, when the 2FA plugin is enabled on VIP Go, users are seeing an `Invalid User ID` error when trying to update any user profile field excluding 2FA fields.

1. Create a new user on our `test-03` environment, and log in.
1. If from VIP, you must **not** be connected to our proxy.
1. Visit your user profile at `/wp-admin/profile.php`
1. Attempt to change your first name.
1. This _shouldn't_ succeed, but if it does, attempt another change.
1. You should see the error text `Invalid User ID`

Screen Shot 2020-02-06 at 9 28 05 AM

brettshumaker:
## Description

There has been a report of an issue with users being unable to edit any user profile field except 2FA fields on /wp-admin/profile.php. While the 2FA plugin is enabled, users were seeing an `Invalid User ID` error while trying to update their `First Name` profile field. We reproduced this on other fields as well although updating one of the 2FA fields always worked. Disabling the 2FA plugin with add_filter( 'wpcom_vip_enable_two_factor', '__return_false' ); did solve the issue and helped us isolate the problem to something dealing with the 2FA plugin. The full steps to reproduce the issue can be found in #1444.

First, a bit of the debugging that took place on this issue:
I traced a profile edit from form submission all the way up to `wp_insert_user()`. If `clean_user_cache()` [here](https://github.com/WordPress/WordPress/blob/c3f9af4f00375d03bf4a4be5185e5c09abca094d/wp-includes/user.php#L1986) is not called, the data gets through just fine. If it is called, the user cache is deleted and when `wp_insert_user()` is called, it’s going to call `get_userdata( $ID )`. That eventually calls `WP_User::get_data_by( $field, $value )` passing in `ID` as the field and the user ID as the value.

`WP_User::get_data_by()` looks in the cache for the user data (which doesn’t exist because `clean_user_cache()` deleted it) with `wp_cache_get( $user_id, 'users' )`. Since nothing is in the cache for that user, it tries to get it directly from the database with `$wpdb->get_row( $wpdb->prepare( "SELECT * FROM $wpdb->users WHERE ID = %s LIMIT 1", $value ) );`. Running that query just once results in `null` being returned. Running it again immediately, however, returns the proper result. This lead us to believe that this could be a db lag issue.

I also noticed that if I disabled the `map_meta_cap` filter [here](https://github.com/Automattic/vip-go-mu-plugins/blob/98ca551386d18f5f200dfe146d7626742b6c2bf8/two-factor.php#L177), profile fields could be updated without issue. I briefly went down the path of this potentially being an issue with how we were restricting capabilities but realized that wasn't the core issue. But I did realize that the function hooked to that filter called `wpcom_vip_is_two_factor_forced()` each time for every capability.

Next, I started looking at each function called by `wpcom_vip_is_two_factor_forced()` and saw that `Two_Factor_Core::is_user_using_two_factor()` results in several user-related WordPress core functions being called. These could be causing a lot of database writes and may have lead to a database replication lag issue. If there was a replication lag, direct queries for a user could be returning zero results and cause profile updates fail with an `Invalid User ID` error.

This change stores the result of `Two_Factor_Core::is_user_using_two_factor()` (by user ID) in a static variable and only calls it once per user. Doing this, user profile updates are always successful because a user is always found in the database.

This issue is **not** reproducible on a VIP Go sandbox (without a few mu-plugins edits) OR while connected to the Automattic proxy.

## Checklist

## Steps to Test

To test this on a VIP sandbox, you must make edits to a few files in `mu-plugins/1/`: 

```
diff --git a/000-vip-init.php b/000-vip-init.php
index 1f9e4211..5463fed2 100644
--- a/000-vip-init.php
+++ b/000-vip-init.php
@@ -72,7 +72,7 @@ if ( ! defined( 'WPCOM_IS_VIP_ENV' ) ) {
 }

-define( 'WPCOM_SANDBOXED', false !== strpos( gethostname(), '_web_dev_' ) );
+define( 'WPCOM_SANDBOXED', false );
 define( 'VIP_GO_IS_CLI_CONTAINER', false !== strpos( gethostname(), '_wpcli_' ) );

 // Used to verify emails sent via our SMTP servers
@@ -143,7 +143,7 @@ define( 'CRON_CONTROL_ADDITIONAL_INTERNAL_EVENTS', $internal_cron_events );
 // Avoids issues with `get_filesystem_method` which attempts to write to `WP_CONTENT_DIR` and fails.
 define( 'FS_METHOD', 'direct' );

-if ( WPCOM_SANDBOXED ) {
+if ( false !== strpos( gethostname(), '_web_dev_' ) ) {
        require __DIR__ . '/vip-helpers/sandbox.php';
 }
```

```
diff --git a/vip-helpers/sandbox.php b/vip-helpers/sandbox.php
index 49935635..be9d8c4c 100644
--- a/vip-helpers/sandbox.php
+++ b/vip-helpers/sandbox.php
@@ -1,8 +1,8 @@
 <?php

-if ( ! WPCOM_SANDBOXED ) {
-	return;
-}
+// if ( ! WPCOM_SANDBOXED ) {
+// 	return;
+// }

 add_action( 'wp_footer', 'wpcom_do_sandbox_bar', 100 );
 add_action( 'admin_footer', 'wpcom_do_sandbox_bar', 100 );
```

1. After you've made the above edit to your sandbox, and verified that you are not connected to proxy, you should be able to reproduce this issue on your VIP Go sandbox.
1. Verify that updating a profile field results in an `Invalid User ID` error. You may find the first attempt is successful, but subsequent submissions should fail.
1. Apply the changes in this PR.
1. Update a user profile field and verify that the change was successful.

Fixes #1444


joshbetz:
@brettshumaker the changes seem OK, but do you have a high level background of what the original issue was? I'm having a hard time following the debugging steps without understanding what problem you were trying to solve.

brettshumaker:
> do you have a high level background of what the original issue was

@joshbetz Sure! Super high level is that users couldn't update their profile fields and were receiving an `Invalid User ID` error.

joshbetz:
Aha! That's super helpful context.

brettshumaker:
Edited the PR description to contain information about the issue we're trying to solve.

joshbetz:
Per Slack, we need to figure out what the root cause is to be sure this is fixing the problem and not introducing any other side effects.

cojennin:
@joshbetz @brettshumaker an in-flight query is getting clobbered by some recursion-y call, resulting in the null.

In the update to the profile, [this get_userdata](https://github.com/WordPress/WordPress/blob/master/wp-includes/user.php#L1542) triggers a DB call which ends up [here](https://github.com/humanmade/hyperdb/blob/master/db.php#L802).

This triggers `wpcom_vip_save_query_callback` in `db-config.php` (which I can't seem to find anywhere on Github or Opengrok for reference).

`wpcom_vip_save_query_callback` has this line:

`$save = current_user_can( 'view_query_monitor' ) || current_user_can( 'manage_options' );`

Which triggers a `map_meta_cap` filter, so now we're [here](https://github.com/Automattic/vip-go-mu-plugins/blob/master/two-factor.php#L163), and that eventually brings us [here](https://github.com/Automattic/vip-go-mu-plugins/blob/master/two-factor.php#L35). 

This winds it's way to another call to [get_userdata](https://github.com/WordPress/two-factor/blob/master/class-two-factor-core.php#L216).

The second `get_userdata` clobbers the first in-flight `get_userdata` resulting in the null.

It's easier to visualize if, in `wp-content/db.php`, you replace [these lines](https://github.com/humanmade/hyperdb/blob/master/db.php#L800-L805) with

```
var_dump( $query );
var_dump( 'first_result ');
var_dump( $this->result );
if ( $this->save_queries ) {
    if ( is_callable($this->save_query_callback) )
        $this->queries[] = call_user_func_array( $this->save_query_callback, array( $query_for_log, $elapsed, $this->save_backtrace ?          debug_backtrace( false ) : null, &$this ) );
     else
         $this->queries[] = array( $query_for_log, $elapsed, $this->get_caller() );
}
var_dump( 'second_result' );
var_dump( $this->result );
```

two-factor.php
```diff
@@ -66,7 +66,7 @@ function wpcom_vip_should_force_two_factor() {
 		return false;
 	}
 
-	if ( Two_Factor_Core::is_user_using_two_factor() ) {
+	if ( wpcom_vip_is_current_user_using_two_factor() ) {
```
cojennin: Think we can follow [this](https://github.com/Automattic/vip-go-mu-plugins/blob/master/two-factor.php#L137) to prevent recursion issue [mentioned here](https://beebee8p2.wordpress.com/2020/02/13/invalid-user-id-error-while-editing-user-profile-fields/#comment-18360)
```suggestion
      if ( apply_filters( 'wpcom_vip_is_current_user_using_two_factor', false ) ) {
          return false;
      }
```

And then [here](https://github.com/Automattic/vip-go-mu-plugins/blob/master/two-factor.php#L141) add:

```
        $is_current_user_using_two_factor = wpcom_vip_is_current_user_using_two_factor();

        add_filter( 'wpcom_vip_is_current_user_using_two_factor', function() use ( $is_current_user_using_two_factor ) {
            return $is_current_user_using_two_factor;
        }, 9 );
```

That way we only calculate this once *and* we avoid calculating during `map_meta_cap`

GaryJones: @cojennin Can you please make these change as an extra commit on the branch that this PR uses?

cojennin: @GaryJones all set

mjangda: @cojennin just to confirm, the main reason we need both the static var and the filter is to avoid calling `wpcom_vip_is_current_user_using_two_factor` in `map_meta_cap`?

cojennin: @mjangda static var can be removed, but figured it wasn't doing any harm to limit the number of calls made by `is_user_using_two_factor`

two-factor.php
```diff
@@ -66,7 +66,7 @@ function wpcom_vip_should_force_two_factor() {
 		return false;
 	}
 
-	if ( Two_Factor_Core::is_user_using_two_factor() ) {
+	if ( apply_filters( 'wpcom_vip_is_current_user_using_two_factor', false ) ) {
```
GaryJones: Does the filter name need to be exactly the same name as the new function? Maybe drop the `current_` bit in the filter name?

cojennin: Sure, all set now

two-factor.php
cojennin: @mjangda removed the static var in the latest push

GaryJones:
@brettshumaker Are you able to re-test the currently proposed PR to confirm it's still good?

brettshumaker:
@GaryJones Yes, this patch seems to still fix the behavior in my testing. 👍

two-factor.php
```diff
@@ -173,6 +173,13 @@ function wpcom_vip_enforce_two_factor_plugin() {
 			return $limited;
 		}, 9 );
 
+		// Calcuate two factor authentication support outside map_meta_cap to avoid callback loop
+		// see: https://github.com/Automattic/vip-go-mu-plugins/pull/1445#issuecomment-592124810
+		$is_user_using_two_factor = Two_Factor_Core::is_user_using_two_factor();
+		add_filter( 'wpcom_vip_is_user_using_two_factor', function() use ( $is_user_using_two_factor ) {
+			return $is_user_using_two_factor;
+		}, 9 );
```
mjangda: Minor thing: can we add a comment noting why we're using a priority of 9 here?

GaryJones: @cojennin ^^^

cojennin: Ah my bad, I just copy and pasted the filter above, I'll remove the 9

two-factor.php
```diff
@@ -142,9 +142,13 @@ function wpcom_vip_enforce_two_factor_plugin() {
...
 		$is_user_using_two_factor = Two_Factor_Core::is_user_using_two_factor();
-		add_filter( 'wpcom_vip_is_user_using_two_factor', function() use ( $is_user_using_two_factor ) {
-			return $is_user_using_two_factor;
-		}, 10 );
+		add_filter( 
+			'wpcom_vip_is_user_using_two_factor',
+			function() use ( $is_user_using_two_factor ) {
+				return $is_user_using_two_factor;
+			}, 
+			10
```
GaryJones: @cojennin Since `10` is the default, let's just leave out that argument.

cojennin: Sure, all set

mjangda:
r832-stacks

PR #1446, opened 2020-02-06
Increase ep_bulk_items_per_page and EP_SYNC_CHUNK_LIMIT to 500

brandon-m-skinner:
## Description

PLAT-885

Increase ep_bulk_items_per_page and EP_SYNC_CHUNK_LIMIT to 500

## Checklist

## Steps to Test

TBD

### DIFF
--- a/elasticsearch/elasticpress/includes/classes/Indexable.php
+++ b/elasticsearch/elasticpress/includes/classes/Indexable.php
@@ -70,7 +70,7 @@ abstract class Indexable {
 		 * @return  {int} New number of items
 		 * @since  3.0
 		 */
-		return apply_filters( 'ep_bulk_items_per_page', 350, $this );
+		return apply_filters( 'ep_bulk_items_per_page', 500, $this );
 	}
 
 	/**
--- a/elasticsearch/elasticpress/includes/classes/Indexable.php
+++ b/elasticsearch/elasticpress/includes/classes/Indexable.php
@@ -70,7 +70,7 @@ abstract class Indexable {
 		 * @return  {int} New number of items
 		 * @since  3.0
 		 */
-		return apply_filters( 'ep_bulk_items_per_page', 500, $this );
+		return apply_filters( 'ep_bulk_items_per_page', 350, $this );
 	}
 
 	/**
--- a/elasticsearch/includes/classes/class-elasticsearch.php
+++ b/elasticsearch/includes/classes/class-elasticsearch.php
@@ -51,7 +51,7 @@ class Elasticsearch {
 		add_filter( 'ep_index_name', [ $this, 'filter__ep_index_name' ], PHP_INT_MAX, 3 ); // We want to enforce the naming, so run this really late.
 
 		// Override default per page value set in elasticsearch/elasticpress/includes/classes/Indexable.php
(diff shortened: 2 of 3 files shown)


elasticsearch/elasticpress/includes/classes/Indexable.php
```diff
@@ -70,7 +70,7 @@ public function get_bulk_items_per_page() {
 		 * @return  {int} New number of items
 		 * @since  3.0
 		 */
-		return apply_filters( 'ep_bulk_items_per_page', 350, $this );
+		return apply_filters( 'ep_bulk_items_per_page', 500, $this );
```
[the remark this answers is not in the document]
brandon-m-skinner: Fair enough. The only thing not in the sub-module is the test, so back to the drawing board.

brandon-m-skinner:
The PR is still missing the `per_page` setting. Unless the setting of the `per_page` doesn't matter? I'll just add this in case it does still matter:

Does it make sense to not bother with` ep_bulk_items_per_page` and instead simply set the `per_page` key to 500 in `query_args` in the health check code to satisfy the:

> The ep_bulk_items_per_page filter controls how many posts are indexed at once during bulk indexing. It defaults to 350, but in my testing 500 is the sweet spot for max throughput.

requirement in PLAT-885? Probably just make a separate constant unless just using `EP_SYNC_CHUNK_LIMIT` here makes sense well if they're always going to be the same?

I've been combing through everything and the filter only seems to be used as a default if the query_args['per_page'] isn't set.

Without doing this, the per_page would still be 350 while the EP_SYNC_CHUNK_LIMIT will be 500.

elasticsearch/includes/classes/class-elasticsearch.php
```diff
@@ -50,6 +50,9 @@ protected function setup_constants() {
 	protected function setup_hooks() {
 		add_filter( 'ep_index_name', [ $this, 'filter__ep_index_name' ], PHP_INT_MAX, 3 ); // We want to enforce the naming, so run this really late.
 
+		// Override default per page value set in elasticsearch/elasticpress/includes/classes/Indexable.php
+		add_filter( 'ep_bulk_items_per_page', function() { return 500; } );
```
nickdaugherty: To keep this aligned with the other filters we have, we should switch this out to a function on the class (I mostly suggested an anonymous function here as a PoC).

elasticsearch/includes/classes/class-elasticsearch.php
```diff
@@ -50,6 +50,9 @@ protected function setup_constants() {
 	protected function setup_hooks() {
 		add_filter( 'ep_index_name', [ $this, 'filter__ep_index_name' ], PHP_INT_MAX, 3 ); // We want to enforce the naming, so run this really late.
 
+		// Override default per page value set in elasticsearch/elasticpress/includes/classes/Indexable.php
+		add_filter( 'ep_bulk_items_per_page', [ $this, 'filter__ep_bulk_items_per_page' ], PHP_INT_MAX, 0 );
```
```suggestion
		add_filter( 'ep_bulk_items_per_page', [ $this, 'filter__ep_bulk_items_per_page' ], PHP_INT_MAX );
```

The "args count" param can be omitted here - it's only needed if your callback is expecting more than 1 argument.

nickdaugherty:
r768-stacks

PR #1447, opened 2020-02-07
Deactivate jetpack widgets when VIP search is enabled

hanifn:
## Description

Deactivate all the automatically registered Jetpack widgets when VIP Search is enabled as the widgets will force the site to keep using Jetpack search.

## Checklist

## Steps to Test

1. Pull down PR
1. Enable JP Search + a widget on a test site
1. Set `USE_VIP_ELASTICSEARCH` to true in wp-config.php
1. Run a query on the frontend - JP Search should not enable itself (check Debug Bar for any Elasticsearch queries from JP)

### DIFF
--- a/elasticsearch/includes/classes/class-elasticsearch.php
+++ b/elasticsearch/includes/classes/class-elasticsearch.php
@@ -110,7 +110,17 @@ class Elasticsearch {
 		return array_values( $filtered );
 	}
 
-	public function filter__jetpack_widgets_to_include( $widget_include ) {
-		return [];
+	public function filter__jetpack_widgets_to_include( $widgets ) {
+		if ( ! is_array( $widgets ) ) {
+			return $widgets;
+		}
+
+		foreach( $widgets as $index => $file ) {
+			if( wp_endswith( $file, '/jetpack/modules/widgets/search.php' ) ) {
+				unset( $widgets[ $index ] );
+			}
+		}
+
+		return $widgets;
 	}
 }
--- a/elasticsearch/includes/classes/class-elasticsearch.php
+++ b/elasticsearch/includes/classes/class-elasticsearch.php
@@ -116,6 +116,8 @@ class Elasticsearch {
 		}
 
 		foreach( $widgets as $index => $file ) {
+			// If the Search widget is included and it's active on a site, it will automatically re-enable the Search module,
+			// even though we filtered it to off earlier, so we need to prevent it from loading
 			if( wp_endswith( $file, '/jetpack/modules/widgets/search.php' ) ) {
 				unset( $widgets[ $index ] );
 			}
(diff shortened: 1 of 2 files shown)


elasticsearch/includes/classes/class-elasticsearch.php
```diff
@@ -106,4 +109,8 @@ public function filter__jetpack_active_modules( $modules ) {
 		// array_filter() preserves keys, so to get a clean / flat array we must pass it through array_values()
 		return array_values( $filtered );
 	}
+
+	public function filter__jetpack_widgets_to_include( $widget_include ) {
+		return [];
```
nickdaugherty: This is going to disable _every_ JP widget - we only need to disable the Search one.

elasticsearch/includes/classes/class-elasticsearch.php
```diff
@@ -54,6 +54,9 @@ protected function setup_hooks() {
 		add_filter( 'ep_intercept_remote_request', '__return_true', 9999 );
 		add_filter( 'ep_do_intercept_request', [ $this, 'filter__ep_do_intercept_request' ], 9999, 4 );
 		add_filter( 'jetpack_active_modules', [ $this, 'filter__jetpack_active_modules' ], 9999 );
+
+		//Filter jetpack widgets
+		add_filter( 'jetpack_widgets_to_include', [ $this, 'filter__jetpack_widgets_to_include' ], 10, 1 );
```
```suggestion
		add_filter( 'jetpack_widgets_to_include', [ $this, 'filter__jetpack_widgets_to_include' ], 10 );
```

Can leave off the args count param if there is just 1.

elasticsearch/includes/classes/class-elasticsearch.php
```diff
@@ -54,6 +54,9 @@ protected function setup_hooks() {
 		add_filter( 'ep_intercept_remote_request', '__return_true', 9999 );
 		add_filter( 'ep_do_intercept_request', [ $this, 'filter__ep_do_intercept_request' ], 9999, 4 );
 		add_filter( 'jetpack_active_modules', [ $this, 'filter__jetpack_active_modules' ], 9999 );
+
+		//Filter jetpack widgets
```
```suggestion
		// Filter jetpack widgets
```

elasticsearch/includes/classes/class-elasticsearch.php
nickdaugherty: Switched over to only disabling search in 2bd452fc

chrean:
Tested on a sandbox, can't see the Jetpack search in the Widgets page. Although, I can't see it if I use `master`, either. I feel I am doing something wrong.

nickdaugherty:
> Tested on a sandbox, can't see the Jetpack search in the Widgets page

It's possible that JP Search wasn't enabled on the site you were testing - the module itself needs to first be enabled, then you need to add a JP Search widget to one of the sidebars on the site - then when you enable JP Search, the widgets should be disabled and the frontend should no longer use JP Search for searches.

nickdaugherty:
I did some additional testing and everything looks good, going to merge + deploy.

nickdaugherty:
r766-stacks

PR #1448, opened 2020-02-07
VIP Search: Increase timeouts for bulk indexing requests

hanifn:
## Description

Increase the timeouts for bulk indexing requests to Elasticsearch to 3 seconds. All other requests remains at 1 second.

## Checklist

## Steps to Test

TBD


elasticsearch/includes/classes/class-elasticsearch.php
```diff
@@ -84,12 +84,19 @@ public function filter__ep_index_name( $index_name, $blog_id, $indexables ) {
...
+		$timeout = 1;
 
 		// TEMP - currently ES server is using a self signed cert during the testing phase...that'll be changed in the near
 		// future, at which time we can remove this
 		$args['sslverify'] = false;
-	
-		$request = vip_safe_wp_remote_request( $query['url'], $fallback_error, 3, 1, 20, $args );
+
+		// If query url ends with '_bulk'
+		if ( 0 === substr_compare( $query[ 'url' ], '_bulk', -5 ) ) {
```
nickdaugherty: This won't account for query params and such - it would be better to use `parse_url()` to get the actual path, then check it for the expected endpoint.

elasticsearch/includes/classes/class-elasticsearch.php
```diff
@@ -84,12 +84,19 @@ public function filter__ep_index_name( $index_name, $blog_id, $indexables ) {
 
 	public function filter__ep_do_intercept_request( $request, $query, $args, $failures ) {
 		$fallback_error = new \WP_Error( 'vip-elasticsearch-upstream-request-failed', 'There was an error connecting to the upstream Elasticsearch server' );
+		$timeout = 1;
```
nickdaugherty: Let's actually increase this to a default of `2` - `1` isn't _quite_ long enough for all cases.
```suggestion
		$timeout = 2;
```

elasticsearch/includes/classes/class-elasticsearch.php
```diff
@@ -84,12 +84,19 @@ public function filter__ep_index_name( $index_name, $blog_id, $indexables ) {
...
 		// TEMP - currently ES server is using a self signed cert during the testing phase...that'll be changed in the near
 		// future, at which time we can remove this
 		$args['sslverify'] = false;
-	
-		$request = vip_safe_wp_remote_request( $query['url'], $fallback_error, 3, 1, 20, $args );
+
+		// If query url ends with '_bulk'
+		if ( 0 === substr_compare( $query[ 'url' ], '_bulk', -5 ) ) {
+			// Bulk index request so increase timeout
+			$timeout = 3;
```
```suggestion
			$timeout = 5;
```

From my testing, `5` seems to be a good upper limit (I've seen timeouts at `3`).

hanifn: We'll need to change `vip_safe_wp_remote_request()` for this too work. It's currently set to limit `timeout` to a max of 3 seconds

elasticsearch/includes/classes/class-elasticsearch.php
hanifn: Good idea!

elasticsearch/includes/classes/class-elasticsearch.php
hanifn: Changed in ad0f04966b6984075dec2f0de02581009339a05e

nickdaugherty: Crud, you're right...let's see how it goes at 3 for now and revisit later.

elasticsearch/includes/classes/class-elasticsearch.php
```diff
@@ -97,12 +97,20 @@ public function filter__ep_index_name( $index_name, $blog_id, $indexables ) {
...
 
 		// TEMP - currently ES server is using a self signed cert during the testing phase...that'll be changed in the near
 		// future, at which time we can remove this
 		$args['sslverify'] = false;
-	
-		$request = vip_safe_wp_remote_request( $query['url'], $fallback_error, 3, 1, 20, $args );
+
+		// If query url ends with '_bulk'
+		$query_path = wp_parse_url( $query[ 'url' ], PHP_URL_PATH );
+		if ( 0 === substr_compare( $query_path, '_bulk', -5 ) ) {
```
nickdaugherty: We could also use `wp_endswith( $query_path, '_bulk' )` for convenience
```suggestion
		if ( wp_endswith( $query_path, '_bulk' ) ) {
```

elasticsearch/includes/classes/class-elasticsearch.php
nickdaugherty: Oh just saw you bumped it up in vip_safe_wp_remote_request) - I think that'll be fine. It'll let sites connect to slower APIs on the frontend, but it's not _that_ much different, and the fallback stuff will still happen when there are multiple requests for a slow resource back to back. cc @mjangda for additional thoughts.

nickdaugherty: @mjangda Do you have any thoughts on ^?

mjangda: Only concern would sites that are actively hitting that 3s or higher timeout. If we bump their sites will slow down even further. However, if that's not a problem we're seeing 🚢 

As as aside, it's probably worth throwing a `_doing_it_wrong` when someone passes in a value that's > 5.

hanifn: > As as aside, it's probably worth throwing a _doing_it_wrong when someone passes in a value that's > 5.

Added in 51e7c4de5cd21e21deb5f49e19885f4182ea40eb

vip-helpers/vip-utils.php
```diff
@@ -751,8 +751,14 @@ function vip_safe_wp_remote_request( $url, $fallback_value='', $threshold=3, $ti
 	$retry     = abs( $retry );
 	$threshold = abs( $threshold );
 
-	// timeouts > 3 seconds are just not reasonable for production usage
-	$timeout = ( (int) $timeout > 3 ) ? 3 : (int) $timeout;
+	// timeouts > 5 seconds are just not reasonable for production usage
+	$timeout = (int) $timeout;
+	if ( $timeout > 5 ) {
+		_doing_it_wrong( __FUNCTION__, 'Using a timeout value of over 5 seconds is strongly discouraged because users have to wait for the remote request to finish before the rest of their page loads.', null );
```
nickdaugherty:
```suggestion
		_doing_it_wrong( __FUNCTION__, 'Remote request timeouts are capped at 5 seconds for performance and stability reasons.', null );
```

vip-helpers/vip-utils.php
```diff
@@ -751,8 +751,14 @@ function vip_safe_wp_remote_request( $url, $fallback_value='', $threshold=3, $ti
 	$retry     = abs( $retry );
 	$threshold = abs( $threshold );
 
-	// timeouts > 3 seconds are just not reasonable for production usage
-	$timeout = ( (int) $timeout > 3 ) ? 3 : (int) $timeout;
+	// timeouts > 5 seconds are just not reasonable for production usage
+	$timeout = (int) $timeout;
+	if ( $timeout > 5 ) {
```
nickdaugherty: So much easier to read this way!

hanifn:
r795-stacks

nickdaugherty:
Thanks for fixing the tests!

PR #1449, opened 2020-02-07
Fix fatal error in validate-health command

chrean:
## Description

Fixes [PLAT-911].
Also adds a missing check on the result of `Indexables::factory()->get()` that may return `boolean|Array`.
In case of failure, that function returns `false` so we must take that into account.

## Checklist

## Steps to Test

1. Check out PR.
1. Delete ES indexes on test site: `dcwp elasticpress delete-index`.
1. Run the validation command: `dcwp vip-es health validate-counts`.
1. Verify that no stacktrace gets printed out.

[PLAT-911]: https://vipjira.atlassian.net/browse/PLAT-911

### DIFF
--- a/elasticsearch/includes/classes/class-health.php
+++ b/elasticsearch/includes/classes/class-health.php
@@ -51,7 +51,6 @@ class Health {
 		// There is not other useful information out of query_es(): it just returns false in case of failure
 		if ( ! $es_result ) {
 			$es_total = 'N/A';
-			$msg = 'error while querying ElasticSearch.';
 			return new WP_Error( 'es_query_error', 'failure querying ES. Hint: verify arguments format. #vip-go-elasticsearch' );
 		}
 
--- a/elasticsearch/includes/classes/class-health.php
+++ b/elasticsearch/includes/classes/class-health.php
@@ -79,6 +79,15 @@ class Health {
 	 */
 	public static function validate_index_users_count() {
 		$users = Indexables::factory()->get( 'user' );
+		// Indexables::factory()->get() returns boolean|array
+		// False is returned in case of error
+		if ( ! $users ) {
+			$result = [
+				'entity' => 'user',
+				'type' => 'N/A',
+				'error' => 'Error retrieving users indexables from Elasticsearch',
+			];
+		}
 
 		$query_args = [
 			'order' => 'asc',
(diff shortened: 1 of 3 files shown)


elasticsearch/includes/classes/class-health.php
```diff
@@ -79,6 +79,15 @@ public static function validate_index_entity_count( array $query_args, \ElasticP
 	 */
 	public static function validate_index_users_count() {
 		$users = Indexables::factory()->get( 'user' );
+		// Indexables::factory()->get() returns boolean|array
+		// False is returned in case of error
+		if ( ! $users ) {
+			$result = [
```
nickdaugherty: What do you think about returning a new WP_Error here instead? That would be more in-line with other error handling here (instead of having 2 different types of error responses).

chrean: I am returning a `result` here instead of `WP_Error` because that's what the function already did: see a few lines below where in case `self::validate_entity_count` returns an error, we are returning the error message inside of the result.
I suggest we change the behavior in both cases then?

chrean: I have made these changes in 408cab96.

elasticsearch/includes/classes/commands/class-healthcommand.php
```diff
@@ -80,7 +84,7 @@ private function render_results( array $results ) {
 		foreach( $results as $result ) {
 			// If it's an error, print out a warning and go to the next iteration
 			if ( array_key_exists( 'error', $result ) ) {
-				WP_CLI::warning( 'Error while validating count:' . $result->get_error_message() );
+				WP_CLI::warning( 'Error while validating count: ' . $result[ 'error' ]);
```
nickdaugherty: With the Health helper functions returning WP_Error's now, does this need to be reworked to check `is_wp_error()`?

chrean: Hmmm... No, because `validate_index_posts_count()` does not return `WP_Error`.
And in the case of `validate_users_count()`, I am checking the return value _before_ calling `render_results()`.

nickdaugherty: Gotcha, LGTM then.

chrean:
r767-stacks

PR #1450, opened 2020-02-07
Add debug-bar-elasticpress submodule

chrean:
## Description

Fixes [PLAT-924].

Adds the debug bar plugin for ElasticPress and loads it if ES is enabled and EP is loaded.

## Checklist

## Steps to Test

Clone the PR on a sandbox host.
Sandbox a site and activate VIP Search; verify that the Debug bar contains an "Elastic Press" item.
Query monitor should not contain anything related to EP, instead.

[PLAT-924]: https://vipjira.atlassian.net/browse/PLAT-924

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -46,4 +46,6 @@
 [submodule "elasticsearch/elasticpress"]
 	path = elasticsearch/elasticpress
 	url = https://github.com/Automattic/ElasticPress.git
-	branch = develop
+[submodule "elasticsearch/debug-bar-elasticpress"]
+	path = elasticsearch/debug-bar-elasticpress
+	url = https://github.com/10up/debug-bar-elasticpress.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -46,6 +46,7 @@
 [submodule "elasticsearch/elasticpress"]
 	path = elasticsearch/elasticpress
 	url = https://github.com/Automattic/ElasticPress.git
+	branch = develop
 [submodule "elasticsearch/debug-bar-elasticpress"]
 	path = elasticsearch/debug-bar-elasticpress
 	url = https://github.com/10up/debug-bar-elasticpress.git
--- a/elasticsearch/includes/classes/class-elasticsearch.php
+++ b/elasticsearch/includes/classes/class-elasticsearch.php
@@ -23,10 +23,6 @@ class Elasticsearch {
 		if ( defined( 'WP_CLI' ) && WP_CLI ) {
 			require_once __DIR__ . '/commands/class-healthcommand.php';
 		}
-		// Load ElasticPress
-		require_once __DIR__ . '/../../elasticpress/elasticpress.php';
-		// Load ElasticPress Debug Bar
-		require_once __DIR__ . '/../../debug-bar-elasticpress/debug-bar-elasticpress.php';
 
 		// Load health check cron job
 		require_once __DIR__ . '/class-health-job.php';
--- a/elasticsearch/includes/classes/class-elasticsearch.php
+++ b/elasticsearch/includes/classes/class-elasticsearch.php
@@ -88,7 +88,7 @@ class Elasticsearch {
 
 	public function action__plugins_loaded() {
 		// Conditionally load only if either/both Query Monitor and Debug Bar are loaded and enabled
-		// NOTE - must hook in here b/c the functions required for checking if debug bar is enabled aren't loaded earlier
+		// NOTE - must hook in here b/c the wp_get_current_user function required for checking if debug bar is enabled isn't loaded earlier
 		if ( apply_filters( 'debug_bar_enable', false ) || apply_filters( 'wpcom_vip_qm_enable', false ) ) {
 			// Load ElasticPress Debug Bar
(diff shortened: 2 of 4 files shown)


joshbetz:
Looks like we just need to make sure the code is loaded and then this will work.

chrean:
@joshbetz this has been tested and the code loads correctly.
As described in the related Jira issue, this is working as expected with both Query Monitor and Debug Bar; we should just make sure that the code is loaded only when either of those are loaded and enabled, otherwise code will be loaded in any case and that is suboptimal. Working on this currently.

chrean:
Reverted commit 93c1a423d0bf2bfbd4451c926a4306d537af0624 .
As discussed on Slack, conditionally loading the plugin is universal and should be done by the plugin itself rather than inside of MU Plugins.
We will follow with a PR contributed to 10up's repo.

elasticsearch/includes/classes/class-elasticsearch.php
```diff
@@ -25,6 +25,8 @@ protected function load_dependencies() {
 		}
 		// Load ElasticPress
 		require_once __DIR__ . '/../../elasticpress/elasticpress.php';
+		// Load ElasticPress Debug Bar
+		require_once __DIR__ . '/../../debug-bar-elasticpress/debug-bar-elasticpress.php';
```
joshbetz: Minor: What do you think about loading this from `elasticsearch/elasticsearch.php`?

My thinking is that this class more related to the implementation details of our ES integration and loading a separate plugin doesn't really fall into that.

chrean: Then I would load both EP + Debug Bar from there.
I have added the loading of Debug Bar here because we were already loading EP here, but perhaps it may be better moving both of them?

nickdaugherty: I don't see the advantage to moving the loading of dependencies. Right now we have the dependency loading nicely encapsulated inside `load_dependencies()` on the class, rather than scattered around inside elasticsearch.php. I would very much prefer keeping dependency loading as it is.

joshbetz: I guess it depends on how you define a dependency. `debug-bar-elasticpress` doesn't really seem like a dependency for anything in `class-elasticsearch` (or any of our other ES related code).

chrean: > I guess it depends on how you define a dependency. debug-bar-elasticpress doesn't really seem like a dependency for anything in class-elasticsearch (or any of our other ES related code).

This is a fair point. debug-bar-elasticpress looks more like a module to me, rather than a actual dependency, because VIP search does not _depend_ on it.
Should we then add a `load_modules()` function or something similar? Are there examples we can follow from existing plugins?

chrean:
@nickdaugherty  @joshbetz  I think we should keep the implementation as easy as possible for now, so I reverted last couple changes.
We could rename `load_dependencies()` to something less confusing if we feel it's confusing.
Moreover, as discussed with @mjangda on Slack, I don't think that conditional loading for the debug bar is such a universal concept to be proposed to 10up. Especially if we are going to add filters in order to properly do it.
The solution we have added to this PR is one of the least intrusive ones and I think we should go for it.
We can come back and iterate more should if it need be.

.gitmodules
```diff
@@ -46,4 +46,6 @@
 [submodule "elasticsearch/elasticpress"]
 	path = elasticsearch/elasticpress
 	url = https://github.com/Automattic/ElasticPress.git
-	branch = develop
```
nickdaugherty: @chrean Looks like this PR is also changing the branch of the EP submodule?

chrean: Yeah we should change the branch to `master` as discussed with 10up. Our production environment should point to `master` which is the release branch.
We can make this change in a separate PR anyway, does not need to happen here.

nickdaugherty: That would be better, so we can more easily detect / test any changes between `develop` and `master` in that submodule, independently.

chrean: Done here: 5664f5b9

Based on yesterday's conversation on Slack, I think this PR can be shipped now?

chrean: Follow-up PR here: https://github.com/Automattic/vip-go-mu-plugins/pull/1460

nickdaugherty: > I think this PR can be shipped now

I found a problem - it's not actually loading correctly because the filters aren't yet setup (b/c this is loading before we setup debug bar stuff). It was working in your testing b/c that test site had the plugin installed separately. Fixed in https://github.com/Automattic/vip-go-mu-plugins/pull/1450/commits/e78ed1482a9c239afbdfcbe1c66458d316e55944 (and I'm going to add some tests).

nickdaugherty:
I reworked how we load it slightly - now it's hooked into `plugins_loaded`. EP Debug Bar also hooks into `plugins_loaded` to setup the its query logging, but this still works here b/c that hook gets registered while `plugins_loaded` is happening, so it ends up being fired still.

nickdaugherty:
r779-stacks

PR #1452, opened 2020-02-07
Separate ElasticPress indexing from using the index

joshbetz:
By default, we'll skip the query integration. This means even when
ElasticPress is enabled, the index won't be used for searches.

In order to test, we can set the `es` query argument.

When we're ready to fully enable ES, we can set the
`vip_enable_elasticsearch` option to `true`.

### Deploy

Before we deploy this, we'll want to ensure the `vip_enable_elasticsearch` option is set on any sites that are using ES.


elasticsearch/includes/classes/class-elasticsearch-integration.php
```diff
@@ -0,0 +1,31 @@
...
+	 * The index can be tested at any time by setting an `es` query argument.
+	 * When we're ready to use the index in production, the `vip_enable_elasticsearch`
+	 * option will be set to `true`, which will enable querying for everyone.
+	 */
+	function ep_skip_query_integration() {
+		if ( isset( $_GET[ 'es' ] ) ) {
+			return false;
+		}
+
+		if ( get_option( 'vip_enable_elasticsearch' ) ) {
```
nickdaugherty: I think an option is too flimsy for this - a constant would be better, because an option can be deleted / changed w/o an audit trail, or the caching could be broken, etc, which could have significant repercussions in production.

Related - we should rename the `USE_VIP_ELASTICSEARCH` constant too, to complement whatever we decide on here.

simonwheatley: > I think an option is too flimsy for this - a constant would be better, because an option can be deleted / changed w/o an audit trail, or the caching could be broken, etc, which could have significant repercussions in production.

This seems like it would also work well when we have a UI on the VIP Dashboard for Environment Variables, what do you think?

joshbetz:
r782-vip-go-stacks

PR #1453, opened 2020-02-07
Remove ElasticPress dashboard

joshbetz:
We don't need most of the EP dashboard. We'll provide a custom
dashboard for things like health info and we don't want it to
be possible to trigger an index operation from the dashboard.

Note: This only hides the dashboard so far. We'll also want to
explicitly block indexing from the dashboard.

### DIFF
--- a/elasticsearch/includes/classes/class-dashboard.php
+++ b/elasticsearch/includes/classes/class-dashboard.php
@@ -1,5 +1,19 @@
 <?php
 
-add_action( 'admin_init', function() {
-	remove_menu_page( 'elasticpress' );
-} );
+namespace Automattic\VIP\Elasticsearch;
+
+class Dashboard {
+
+	function __construct() {
+		add_action( 'admin_init', array( __CLASS__, 'admin_init' ) );
+	}
+
+	/**
+	 * Disable the Elasticpress dashboard
+	 */
+	function admin_init() {
+		remove_menu_page( 'elasticpress' );
+	}
+}
+
+new Dashboard;
--- a/elasticsearch/includes/classes/class-dashboard.php
+++ b/elasticsearch/includes/classes/class-dashboard.php
@@ -11,7 +11,7 @@ class Dashboard {
 	/**
 	 * Disable the Elasticpress dashboard
 	 */
-	function admin_init() {
+	static function admin_init() {
 		if ( ! is_automattician() ) {
 			remove_menu_page( 'elasticpress' );
 		}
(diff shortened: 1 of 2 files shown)


nickdaugherty:
Should this code go into the main class-elasticsearch.php class? We're currently not separating out ElasticPress specific functionality (though we may in the future).

joshbetz:
@nickdaugherty The longer term vision is to replace the EP dashboard with a simpler VIP dashboard. Some things like index health will go there. It's with that in mind that we started this class. Happy to organize it according to any guidelines you all have been following though.

nickdaugherty:
Ya makes sense - I've found the dashboard pretty useful when setting up ES on sites so far - maybe we can leave it enabled for Automatticians while we work on a replacement?

nickdaugherty:
The EP dashboard hooks into a _lot_ of stuff, it's a shame there isn't a better way to prevent it from running:

https://github.com/10up/ElasticPress/blob/80ae059e58eb9cff2a2fb8de78544c27787ed7b7/includes/dashboard.php#L29-L56

elazzabi:
I still see the dashboard, is there anything else I need to do to test this PR? (other than checking out the code)

Screen Shot 2020-02-13 at 14 47 34

elazzabi:
Tested locally with the help of @joshbetz, works as expected!

elasticsearch/includes/classes/class-dashboard.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace Automattic\VIP\Elasticsearch;
+
+class Dashboard {
+
+	function __construct() {
+		add_action( 'admin_init', array( __CLASS__, 'admin_init' ) );
```
nickdaugherty: As discussed in the comments, lets be sure that it's left enabled for Automatticians at this point.

joshbetz: Yep, the dashboard still works. We're just hiding the navigation.

nickdaugherty: I mean we should probably just leave the navigation - I've been using it pretty much constantly while setting up sites + testing.

joshbetz: Ok, we can probably just wait to deploy this until we have something to replace it with then.

nickdaugherty: @joshbetz what do you think about 02d31907?

joshbetz: Sweet, that works. I didn't realize it would be so easy 🙃

joshbetz:
r783-vip-go-stacks

PR #1454, opened 2020-02-11
Add option to disable Elasticsearch healthcheck job

hanifn:
## Description

Add option to disable the Elasticsearch healthcheck cronjob by setting a constant.

## Checklist

## Steps to Test

Example:

1. Check out PR.
1. Sandbox a site with VIP search setup
1. `define( 'DISABLE_VIP_SEARCH_HEALTHCHECKS', true )` in `vip-config.php`
1. Verify health check job is no longer in cron events list


elasticsearch/includes/classes/class-elasticsearch.php
```diff
@@ -69,6 +69,13 @@ protected function load_commands() {
 	}
 
 	protected function setup_healthchecks() {
+		// Disable healthcheck job
+		if ( defined( 'DISABLE_VIP_SEARCH_HEALTHCHECKS' ) && true === DISABLE_VIP_SEARCH_HEALTHCHECKS ) {
```
nickdaugherty: Should we just be using the filter that we added? `enable_vip_search_healthchecks`

https://github.com/Automattic/vip-go-mu-plugins/blob/2b03eec117a34fe3f76941d4bab8932688a314d1/elasticsearch/includes/classes/class-elasticsearch.php#L74

hanifn: I think using a constant that's checked earlier is a simpler solution to ensure the job is actually disabled. With the filter, we have to worry about the order and such

hanifn: > I sandboxed a site using this branch on a platform sandbox host + I set the constant to true. Did I do something wrong? 🤔

@chrean Sorry that was me doing it wrong. I was testing using `wp shell` but it any other scenario, the cron-control wouldn't have been loaded yet so it doesn't work. I've changed it to hook into the `init` action so it should work now.

hanifn:
> I sandboxed a site using this branch on a platform sandbox host + I set the constant to true. Did I do something wrong? 🤔

@chrean Sorry that was me doing it wrong. I was testing using wp shell but it any other scenario, the cron-control wouldn't have been loaded yet so it doesn't work. I've changed it to hook into the init action so it should work now.

chrean:
Hmmm... Still not working for me @hanifn .
Did your tests succeed? 🤔

```bash
wp cron event list --allow-root|grep -i health
vip_search_healthcheck	2020-02-13 15:43:37	25 minutes 9 seconds	30 minutes
```

hanifn:
@chrean Yes. My tests shows that it works
```
user@host:~$ wp cron event list --allow-root|grep -i health
user@host:~$
```

Did you add `define( 'DISABLE_VIP_SEARCH_HEALTHCHECKS', true );` to the `vip-config.php` file?

elasticsearch/includes/classes/class-elasticsearch.php
nickdaugherty: I guess it just depends on how sure we want to be that it's disabled - how critical it is to be disabled. Usually having a constant that we define somewhere on behalf of a site is useful when we _really_ need to be sure there is no other way to do something...I don't know if this case warrants having two different ways to disable the health checks? What are the use-cases?

elasticsearch/includes/classes/class-elasticsearch.php
```diff
@@ -85,15 +85,23 @@ protected function load_commands() {
 	}
 
 	protected function setup_healthchecks() {
+		$healhcheck = new HealthJob();
+
+		// Disable healthcheck job
+		if ( defined( 'DISABLE_VIP_SEARCH_HEALTHCHECKS' ) && true === DISABLE_VIP_SEARCH_HEALTHCHECKS ) {
+			// Hook into init action to ensure cron-control has already been loaded
+			add_action( 'init', [ $healhcheck, 'disable_job' ] );
```
nickdaugherty: Instead of scheduling this on `init` (which will run on every request), what if we move it into the job itself? So when the job runs, we check the constant, and if it's `true`, we disable the job and return early. So worst case, if it's already scheduled then the constant is added, the job runs 1 additional time but only unregisters itself and returns.

I'm just wary of adding additional code / processing to every request when it only ever needs to happen at most once.

elasticsearch/includes/classes/class-elasticsearch.php
nickdaugherty: I think this will be fine as-is...

elasticsearch/includes/classes/class-elasticsearch.php
nickdaugherty: For completeness we should wrap the "is enabled" logic in a reusable function that gets called here and once again when the job is running, so that we can bail early.

Currently, disabling the job via the filter won't actually do anything if it's already been scheduled.

hanifn: > Instead of scheduling this on `init` (which will run on every request), what if we move it into the job itself? So when the job runs, we check the constant, and if it's true, we disable the job and return early.

So I tried this and it doesn't seem to be working. It's not removing the existing event. It doesn't run the job though

nickdaugherty:
@hanifn I figured out why the job wasn't unregistering itself - it was because it never got hooked in to run because of the `is_enabled()` check in class-elasticsearch.php - I've shuffled it around a bit in https://github.com/Automattic/vip-go-mu-plugins/pull/1454/commits/493a0310bb11694d240c883649f1a4da68bc25f9 and it's fixed.

chrean:
@nickdaugherty tests are failing now. Maybe last changeset broke it for some reasons?

hanifn:
The failing test is this one: https://github.com/Automattic/vip-go-mu-plugins/blob/6edcd423ff47d78220260a1db3467c6a1a2c5c2a/tests/elasticsearch/test-class-elasticsearch.php#L456-L464

The code has changed such that `$this->healthcheck` is always set. I'm actually not sure if that's the best way to test that the healthcheck job is enabled

hanifn:
@nickdaugherty @chrean I have updated the failing test in 382ae1a7022538db88088940afbcfddd56be6d80

It should pass now

nickdaugherty:
Thanks for fixing that test @hanifn ! I meant to yesterday but got distracted on other stuff.

nickdaugherty:
r806-stacks

PR #1458, opened 2020-02-12
Block all API requests for all 2fa users

mjangda:
We've given enough notice for this.

Note: we can remove `VIP_2FA_ALLOW_API_LOGIN_ENV_IDS` as part of this deploy.

## Checklist

- - This change has relevant unit tests (Will look at adding some unit tests in an upcoming PR)
- - This change has relevant documentation additions / updates (if applicable).

## Steps to Test

- Apply patch.
- Make an API request to a site with a user account that has 2fa enabled (see internal P2 for example). It should fail with a 403 (Incorrect username or password.)

### DIFF
--- a/two-factor.php
+++ b/two-factor.php
@@ -12,42 +12,8 @@ require_once __DIR__ . '/wpcom-vip-two-factor/set-providers.php';
 // Detect if the current user is logged in via Jetpack SSO
 require_once __DIR__ . '/wpcom-vip-two-factor/is-jetpack-sso.php';
 
-// Handle 2fa for API requests
-add_filter( 'two_factor_user_api_login_enable', function( $allow_2fa_bypass, $user_id ) {
-	// Allow API requests for a subset of environments for now.
-	if ( defined( 'VIP_2FA_ALLOW_API_LOGIN_ENV_IDS' )
-		&& is_array( VIP_2FA_ALLOW_API_LOGIN_ENV_IDS )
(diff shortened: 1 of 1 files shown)


two-factor.php
```diff
@@ -12,42 +12,8 @@
...
-	Automattic\VIP\Stats\send_pixel( [
-		'vip-go-2fa-api-blocked-by-site' => FILES_CLIENT_SITE_ID,
-		'vip-go-2fa-api-blocked-by-ua' => sanitize_key( $_SERVER['HTTP_USER_AGENT'] ), 
-	] );
-
-	// Do not allow API requests for users with 2fa enabled.
-	return false;
-}, 1, 2 ); // Allow overrides at later priorities
+// Do not allow API requests from 2fa users.
+add_filter( 'two_factor_user_api_login_enable', '__return_false', 1 ); // Hook in early to allow overrides
```
joshbetz: The default is `false`. I think we can also just remove this line with basically the same effect?

https://github.com/WordPress/two-factor/blob/a5a4ad7b6bd99e57b31d9ec64712af1dea072de1/class-two-factor-core.php#L331

mjangda: That's a good point. I'd say it's probably worth keeping in just to make it an explicit decision on our part just (in case things change in the upstream plugin). I think having tests would help as well.

joshbetz: Yeah, makes sense.

mjangda:
r772-stacks
r773-stacks

PR #1466, opened 2020-02-17
Add Opsgenie alert helpers

hanifn:
## Description

Add a new Alerts class to hold IRC and Opsgenie alert helper functions.

## Checklist

## Steps to Test

TBD


lib/utils/class-alerts.php
```diff
@@ -0,0 +1,270 @@
...
+
+	/**
+	 * Contructor
+	 *
+	 * Set to protected to prevent direct instantiation.
+	 * Use Alerts::get_instance() to get an instance of Alerts
+	 */
+	protected function __construct() {
+		if ( ! defined( 'ALERT_SERVICE_ADDRESS' ) || ! ALERT_SERVICE_ADDRESS ) {
+			throw new Exception( 'Missing alerts service host configuration in ALERT_SERVICE_ADDRESS constant' );
```
nickdaugherty: @chrean also mentioned the exception throwing in some other comments, but an uncaught thrown exception is a *fatal error*. That would be drastically different behavior than what we currently do, which is log an error message and return false without trying to contact the service - https://github.com/Automattic/vip-go-mu-plugins/blob/b81b054081e5eebc3b9c054c9cb058b11fe951e7/vip-helpers/vip-utils.php#L1457

This would at least break local development.

lib/utils/class-alerts.php
```diff
@@ -0,0 +1,270 @@
...
+
+	/**
+	 * Send the alert
+	 *
+	 * @param array $body The alert message body
+	 *
+	 * @return array Response details from wp_remote_post
+	 */
+	protected function send( $body ) {
+		$response = wp_remote_post( $this->service_url, [
```
nickdaugherty: Can we use the `vip_safe_wp_remote_request()` function instead?

lib/utils/class-alerts.php
```diff
@@ -0,0 +1,270 @@
...
+			'timeout' => 0.1,
+			'body' => json_encode( $body ),
+		] );
+
+		if ( is_wp_error( $response ) ) {
+			throw new Exception( sprintf( 'Error sending alert: %s', $response->get_error_message() ) );
```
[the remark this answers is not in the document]
nickdaugherty: We should definitely not throw an exception here - we need to return the WP_Error.

lib/utils/class-alerts.php
```diff
@@ -0,0 +1,270 @@
...
+
+		return $details;
+	}
+
+	/**
+	 * Get an instance of this Alerts class
+	 *
+	 * @return Alerts
+	 */
+	public static function get_instance() {
```
nickdaugherty: For singletons, the common convention is to name the function as `instance()`.
```suggestion
	public static function instance() {
```

lib/utils/class-alerts.php
```diff
@@ -0,0 +1,270 @@
...
+	 *
+	 * @return bool	True if successful. Else, will return false
+	 */
+	public static function opsgenie( $message, $details, $kind = '', $interval = 0 ) {
+		try {
+			$alerts = self::get_instance();
+
+			if ( $kind && $interval ) {
+				if ( ! $alerts->add_cache( $kind, $interval ) ) {
+					return false;
```
nickdaugherty: We should log this condition, and probably even send it to IRC/Slack...b/c a missing OpsGenie alert could be a big problem.

lib/utils/class-alerts.php
```diff
@@ -0,0 +1,270 @@
...
+	 *
+	 * @return bool	True if successful. Else, will return false
+	 */
+	public static function chat( $channel_or_user, $message, $level = 0, $kind = '', $interval = 0 ) {
+		try {
+			$alerts = self::get_instance();
+
+			if ( $kind && $interval ) {
+				if ( ! $alerts->add_cache( $kind, $interval ) ) {
+					return false;
```
nickdaugherty: We should log this case

lib/utils/class-alerts.php
nickdaugherty: I guess everywhere that *currently* calls the constructor is wrapped in a try/catch - but in general we should not use Exceptions and instead pass WP_Errors around, which is the "WordPress way" (so we should strive to match that for consistency with the rest of the WP ecosystem).

lib/utils/class-alerts.php
```diff
@@ -0,0 +1,270 @@
...
+
+namespace Automattic\VIP\Utils;
+
+class Alerts {
+	/**
+	 * The alerts service address
+	 *
+	 * @var string
+	 */
+	protected $service_address;
```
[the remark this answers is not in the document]
hanifn: I typically start off with `protected` members unless it's very clear that `public` is required but you are correct. We should be explicitly stating which members we want to be extensible or not.

lib/utils/class-alerts.php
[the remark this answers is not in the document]
hanifn: My thinking is to have the main `Alerts::chat()` method to catch and handle all the exceptions just so that we have a more convenient and consistent way of logging them.

lib/utils/class-alerts.php
```diff
@@ -0,0 +1,270 @@
...
+	 * Validate alert message
+	 *
+	 * Ensure the message is a valid string
+	 *
+	 * @param $message mixed
+	 *
+	 * @return string Validated and trimmed message
+	 */
+	protected function validate_message( $message ) {
+		if ( is_array( $message ) || is_object( $message ) ) {
```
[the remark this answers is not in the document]
hanifn: Makes sense. I basically copied this part from `wpvom_vip_irc()`

lib/utils/class-alerts.php
```diff
@@ -0,0 +1,270 @@
...
+	}
+
+	/**
+	 * Validate Opsgenie details array
+	 *
+	 * @param $details array
+	 *
+	 * @return array
+	 */
+	protected function validate_details( $details) {
```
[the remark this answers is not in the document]
hanifn: What kind of sanitisation are you thinking off here? I didn't put that in because it looks like Opsgenie already sanitise the alerts data strings

lib/utils/class-alerts.php
```diff
@@ -0,0 +1,270 @@
...
+	}
+
+	/**
+	 * Validate IRC channel or user name
+	 *
+	 * @param $channel_or_user string
+	 *
+	 * @return string Validated and cleaned channel or user name
+	 */
+	protected function validate_channel_or_user( $channel_or_user ) {
```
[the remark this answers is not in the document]
hanifn: No benefit really. I'm not sure if it would make much of a difference since we will need an instance of `Alerts`  anyway in the main static functions like `chat()` and `opsgenie()`

lib/utils/class-alerts.php
hanifn: This is a little tricky since constructors doesn't really return anything. Adding `return new WP_Error()` wouldn't do anything and might even cause an error. This is the primary reason I went with exceptions

lib/utils/class-alerts.php
hanifn: Added in 0625bb47df5c87019355abd8cd35c3769f6461f5

lib/utils/class-alerts.php
hanifn: Added in 0625bb47df5c87019355abd8cd35c3769f6461f5

lib/utils/class-alerts.php
hanifn: Changed in a08e0d406d6623be0ca464b6cd40289c7072dc96

lib/utils/class-alerts.php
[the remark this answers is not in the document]
hanifn: Changed in 88820cb2ee4d10adc3295473d7adb6d92bc6e0df

lib/utils/class-alerts.php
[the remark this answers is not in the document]
chrean: OK then we can probably live with it.

lib/utils/class-alerts.php
chrean: Then maybe we should not either `throw` nor `return WP_Error` inside of the constructor? Maybe only functions like `init()` should return errors? How is that normally done in WP?

lib/utils/class-alerts.php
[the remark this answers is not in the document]
chrean: OK but the point is, this function could - in theory - be called by functions other than `Alerts::chat()` in a future, and when that time comes, we would be breaking WP coding standards and introducing additional headaches.
I would prefer sticking to the normal way to do things even though it does not seem vital for the sake of the current PR: it could become an issue in the future.

lib/utils/class-alerts.php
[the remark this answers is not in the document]
chrean: If we are sure that OpsGenie will sanitise and not blow up if somehting weird gets passed through, I am fine with this. :)

lib/utils/class-alerts.php
hanifn: > Maybe only functions like init() should return errors? How is that normally done in WP?

I'm not sure what's the "proper" way in WP but I think using the `init()` method could work.

hanifn:
@chrean @nickdaugherty I've replaced all the exceptions with `WP_Error`s in 6a2b90db8e9385c94c3ca56dbb2b402061227bde

lib/utils/class-alerts.php
hanifn: I've moved the object config setup code to the `instance()` method in f49de002aae092673c9c394129520f7c3c0585c9
What do you think?

tests/lib/utils/test-alerts.php
```diff
@@ -0,0 +1,258 @@
...
+
+		$result = $validate_details_method->invokeArgs( $alerts, [ $details ] );
+
+		$this->assertWPError( $result );
+		$this->assertEquals( 'invalid-opsgenie-details', $result->get_error_code() );
+		$this->assertEquals( 'test.host', $alerts->service_address, 'Wrong alerts service address' );
+		$this->assertEquals( 9999, $alerts->service_port, 'Wrong alerts service port' );
+		$this->assertEquals( 'http://test.host:9999/v1.0/alerts', $alerts->service_url, 'Wrong alerts service URL' );
+	}
+}
```
nickdaugherty: It would be worth adding a test case for the rate limiting too to ensure that's working as expected.

lib/utils/class-alerts.php
```diff
@@ -0,0 +1,300 @@
...
+	 *
+	 * @param $target (string) Channel or Username.  Usernames prefixed with an @, channel optionally prefixed by #.
+	 * @param $message (string) Message
+	 * @param $level (int) Level The severity level of the message
+	 * @param $kind string Cache slug
+	 * @param $interval integer Interval in seconds between two messages sent from one DC
+	 *
+	 * @return bool	True if successful. Else, will return false
+	 */
+	public static function chat( $channel_or_user, $message, $level = 0, $kind = '', $interval = 0 ) {
```
nickdaugherty: This is coming from existing code, but I'm not sure why it defaults the `$interval` to `0`...I think we need to change that for this new class...to at least `1`. There is no situation where we need to allow more than 1 message per second of any type.

nickdaugherty: The `$kind` defaulting to an empty string is also weird...it means that the rate limiting won't work (even if interval was set) by default, which to me feels pretty broken. We don't necessarily need to fix it in this PR, but a a default `$kind` that was the MD5 of the `$channel_or_user . $message` would at least provide some safeguards to runaway code.

tests/lib/utils/test-alerts.php
```diff
@@ -0,0 +1,258 @@
...
+
+	/**
+	 * @runInSeparateProcess
+	 * @preserveGlobalState disabled
+	 */
+	public function test__instance__missing_config() {
+		$alerts = Alerts::instance();
+
+		$this->assertWPError( $alerts );
+		$this->assertEquals( 'missing-service-address', $alerts->get_error_code(), 'Wrong error message' );
```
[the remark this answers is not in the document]
hanifn: Changed in a6f5e145ae89cb4c4a527884829dc5ba512d74c6

tests/lib/utils/test-alerts.php
```diff
@@ -0,0 +1,258 @@
...
+
+		$validate_channel_or_user_method = self::get_alerts_method( 'validate_channel_or_user' );
+
+		$result = $validate_channel_or_user_method->invokeArgs( $alerts, [ $channel_or_user ] );
+
+		$this->assertWPError( $result );
+		$this->assertEquals( 'invalid-channel-or-user', $result->get_error_code() );
+		$this->assertEquals( 'test.host', $alerts->service_address, 'Wrong alerts service address' );
+		$this->assertEquals( 9999, $alerts->service_port, 'Wrong alerts service port' );
+		$this->assertEquals( 'http://test.host:9999/v1.0/alerts', $alerts->service_url, 'Wrong alerts service URL' );
```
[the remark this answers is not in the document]
hanifn: Make sense. Removed in e2ada45e020991e9d53b5b17b0113efb540686ce

lib/utils/class-alerts.php
```diff
@@ -0,0 +1,300 @@
...
+	}
+
+	/**
+	 * Validate Opsgenie details array
+	 *
+	 * @param $details array
+	 *
+	 * @return array|WP_Error
+	 */
+	private function validate_details( $details) {
```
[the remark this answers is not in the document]
hanifn: Renamed in a3f95284bb85fa6277068db8143a752a3c29575c

lib/utils/class-alerts.php
```diff
@@ -0,0 +1,300 @@
...
+	public static function opsgenie( $message, $details, $kind = '', $interval = 0 ) {
+		$alerts = self::instance();
+
+		if ( is_wp_error( $alerts ) ) {
+			error_log( $alerts->get_error_message() );
+
+			return false;
+		}
+
+		if ( $kind && $interval ) {
```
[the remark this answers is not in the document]
hanifn: Well, if `$kind` is also set, then it will add a new cache entry with an expiry of `0` seconds

lib/utils/class-alerts.php
hanifn: I think the idea of the original code was to make it optional. Perhaps to allow some alerts to skip rate limiting? I agree with the defaults for `$interval` though. Setting it to `0` makes no sense.

lib/utils/class-alerts.php
chrean: I like it better! 👍

lib/utils/class-alerts.php
```diff
@@ -0,0 +1,270 @@
...
+	 *
+	 * Set to protected to prevent direct instantiation.
+	 * Use Alerts::get_instance() to get an instance of Alerts
+	 */
+	protected function __construct() {
+		if ( ! defined( 'ALERT_SERVICE_ADDRESS' ) || ! ALERT_SERVICE_ADDRESS ) {
+			throw new Exception( 'Missing alerts service host configuration in ALERT_SERVICE_ADDRESS constant' );
+		}
+
+		if ( ! defined( 'ALERT_SERVICE_PORT' ) || ! ALERT_SERVICE_PORT ) {
```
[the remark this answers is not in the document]
chrean: Added with f445087b

lib/utils/class-alerts.php
nickdaugherty: I believe it's like that so the site itself can send more than 1 message per second across all message types (if $kind is not specified), which is a little dangerous b/c that's a _lot_ of messages from one site. The alerts service endpoint does its own rate limiting so we're not directly at risk of maxing out IRC or Slack, but it would be possible to overwhelm the rate limit service if runaway code was blasting alerts at it.

lib/utils/class-alerts.php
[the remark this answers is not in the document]
chrean: Looks to m that if `$interval` equals `0`, then the condition `if ( $kind && $interval )` will be false, and `add_cache` won't be called... Or?

lib/utils/class-alerts.php
nickdaugherty: I'm ok keeping the rate limiting as-is to get this PR across the line, just thought it would be a good opportunity to revisit it since this is not used anywhere.

lib/utils/class-alerts.php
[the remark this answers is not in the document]
hanifn: Oh you're right. Both needs to be set to a truthy value for it to move on so most likely nothing will happen

lib/utils/class-alerts.php
hanifn: @nickdaugherty I liked your suggestion so made it so that there's a default `$kind` value making the rate limiting no longer optional. Added in bd2a2b4b87600ea38d87ddcf6beb48b8451f75fa

lib/utils/class-alerts.php
```diff
@@ -0,0 +1,270 @@
...
+	protected static $instance = null;
+
+	/**
+	 * Contructor
+	 *
+	 * Set to protected to prevent direct instantiation.
+	 * Use Alerts::get_instance() to get an instance of Alerts
+	 */
+	protected function __construct() {
+		if ( ! defined( 'ALERT_SERVICE_ADDRESS' ) || ! ALERT_SERVICE_ADDRESS ) {
```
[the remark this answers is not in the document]
chrean: Dismissing this as validating a domain is hard and requires very complex regexes. Probably not worth the effort.

tests/lib/utils/test-alerts.php
chrean: I think this does not apply anymore after last refactoring.

lib/utils/class-alerts.php
```diff
@@ -0,0 +1,300 @@
...
+	 *
+	 * @return bool	True if successful. Else, will return false
+	 */
+	public static function opsgenie( $message, $details, $kind = '', $interval = 0 ) {
+		$alerts = self::instance();
+
+		if ( is_wp_error( $alerts ) ) {
+			error_log( $alerts->get_error_message() );
+
+			return false;
```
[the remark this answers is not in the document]
chrean: Are you suggesting we convert all the `return false` to `return new WP_Error...`?
That's a good number of changes we have to make and I am unsure about the convention here. My feeling is that we should not return false, but as long as we log the error somewhere that may be fine. 🤷‍♂️
However, letting the caller know that there has been an error sounds always good to me.

We can merge this PR, then come back and make the above changes if it makes sense.

As a sidenote I have seen this `error_log` + `return false` approach in other spots throughout MU plugins, too. Not sure if those are remainders of old days, or what.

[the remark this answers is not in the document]
nickdaugherty: Yep that's what I was thinking...b/c `false` is just a boolean which limits the actions that the calling code can take in response to an error. We can leave as-is for now.

tests/lib/utils/test-alerts.php
nickdaugherty: Well we still don't have any test that proves that if you try to send more than 1 message of `$kind` within the `$interval` time, it fails, so we don't _know_ that the rate limiting is working and won't regress in the future. But I don't want to hold this up any longer so let's ship what we have.

nickdaugherty:
r807-stacks

nickdaugherty:
@hanifn I went ahead and deployed this - it's not actually included anywhere (so not yet useable), but we can do that in a separate PR.

PR #1469, opened 2020-02-18
Revert "Revert "Only alert when diff is not 0""

nickdaugherty:
Reverts Automattic/vip-go-mu-plugins#1468 to unrevert the revert for "only alert when diff is not 0" fix for ES health checking 😄


nickdaugherty:
r788-stacks

PR #1470, opened 2020-02-20
Manually setup EP Debug Bar logging due to plugins_loaded already firing

nickdaugherty:
## Description

This code runs on plugins_loaded, but the EP debug bar also hooks into plugins_loaded, so we need to explicitly call the setup function ourselves to ensure it registers itself.

## Checklist

## Steps to Test

Example:

1. Check out PR.
1. Run some search queries
1. Ensure queries show up in EP Debug Bar

### DIFF
--- a/elasticsearch/includes/classes/class-elasticsearch.php
+++ b/elasticsearch/includes/classes/class-elasticsearch.php
@@ -108,6 +108,11 @@ class Elasticsearch {
 		if ( apply_filters( 'debug_bar_enable', false ) || apply_filters( 'wpcom_vip_qm_enable', false ) ) {
 			// Load ElasticPress Debug Bar
 			require_once __DIR__ . '/../../debug-bar-elasticpress/debug-bar-elasticpress.php';
+
+			// And ensure the logging has been setup (since it also hooks on plugins_loaded)
+			if ( function_exists( 'ep_setup_query_log' ) ) {
+				ep_setup_query_log();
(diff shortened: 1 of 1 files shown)


elasticsearch/includes/classes/class-elasticsearch.php
```diff
@@ -108,6 +108,11 @@ public function action__plugins_loaded() {
 		if ( apply_filters( 'debug_bar_enable', false ) || apply_filters( 'wpcom_vip_qm_enable', false ) ) {
 			// Load ElasticPress Debug Bar
 			require_once __DIR__ . '/../../debug-bar-elasticpress/debug-bar-elasticpress.php';
+
+			// And ensure the logging has been setup (since it also hooks on plugins_loaded)
```
chrean: That's more or less what I did, except for the `function_exists` check, which I had not thought of. 🤭

However, I just tested it and could not see any queries being logged.

chrean:
@nickdaugherty I have tested this by sandboxing site 1120 with this PR, ran some searches but could not see any query being logged. Am I doing something wrong?

elasticsearch/includes/classes/class-elasticsearch.php
chrean: As discussed on Slack, working as intended.

nickdaugherty:
r797-stacks

PR #1471, opened 2020-02-24
Disable Jetpack Search offload when ElasticPress is being used

joshbetz:
When ElasticPress query integration is enabled, we want to ensure queries are no longer offloaded to Jetpack Search.

Needs testing


nickdaugherty:
VIP Search already disables JP Search when enabled:

https://github.com/Automattic/vip-go-mu-plugins/blob/8c74507c8a81b0c538ae988f99263b73e7c7bc9b/elasticsearch/includes/classes/class-elasticsearch.php#L165-L195

Though given that activation and integration are separate things, maybe that needs to change to only run if integration is enabled?

joshbetz:
Yeah, we probably _don't_ want to disable Jetpack Search when VIP Search is enabled. Ideally we can enable VIP Search, index all the content, and then flip from Jetpack Search to VIP Search. Otherwise there will be a "downtime" period where the site doesn't have ES while we wait for the VIP Search index to build.

joshbetz:
@nickdaugherty I wonder if we should automatically disable Jetpack Search once we've switched over to VIP Search (i.e. when `VIP_ENABLE_ELASTICSEARCH_QUERY_INTEGRATION`) is set.

I was originally thinking we'd just remove the module filter you mentioned, but I can see a case for still disabling it after we've switch queries over to VIP Search. If we're going to do that, I think we should use the `VIP_ENABLE_ELASTICSEARCH_QUERY_INTEGRATION` constant directly to avoid disabling Jetpack Search during testing.

The only case I can make against automatically disabling once everything is switched is potentially needing to revert. Are we going to cause any problems if we turn the module off and then back on?

nickdaugherty:
I'm not sure about leaving JP Search enabled and just bypassing it in the query filter - JP Search does more than just take over search (like registering widgets). IMO VIP Search and JP Search need to be considered incompatible and not loaded at the same time, to prevent bugs.

I do agree that we should only do the disabling if the `VIP_ENABLE_ELASTICSEARCH_QUERY_INTEGRATION` constant is set, rather than how it was originally (disabled immediately if VIP Search is enabled) to accommodate a transition period.

joshbetz:
Yeah, after thinking about this some more, I agree. There's not much benefit in having Jetpack Search active after you've flipped the switch to VIP Search. There will be some overlap during the testing period, which should be sufficient to get everything migrated before going live with VIP Search.

brandon-m-skinner:
Github being janky. Closing and remaking PR on the same branch

PR #1473, opened 2020-02-24
Term change integration for Search

nickdaugherty:
## Description

IN PROGRESS

Hooks into WP term changes and schedules all associated posts to be re-indexed.

## Checklist

## Steps to Test

TBD

### DIFF
--- a/elasticsearch/includes/classes/class-elasticsearch.php
+++ b/elasticsearch/includes/classes/class-elasticsearch.php
@@ -136,6 +136,8 @@ class Elasticsearch {
 	}
 
 	public function action__edited_term( $term_id, $tt_id, $taxonomy ) {
+		global $wpdb;
+
 		// Find ID of all attached posts (query lifted from wp_delete_term())
 		$object_ids = (array) $wpdb->get_col( $wpdb->prepare( "SELECT object_id FROM $wpdb->term_relationships WHERE term_taxonomy_id = %d", $tt_id ) );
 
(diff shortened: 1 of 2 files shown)


elasticsearch/includes/classes/class-elasticsearch.php
```diff
@@ -116,6 +124,20 @@ public function action__plugins_loaded() {
...
 
+	public function action__edited_term( $term_id, $tt_id, $taxonomy ) {
+		// Find ID of all attached posts (query lifted from wp_delete_term())
+		$object_ids = (array) $wpdb->get_col( $wpdb->prepare( "SELECT object_id FROM $wpdb->term_relationships WHERE term_taxonomy_id = %d", $tt_id ) );
+
+		// Bulk index them all
+		\ElasticPress\Indexables::factory()->get( 'post' )->bulk_index( $object_ids );
+	}
+
+	public function action__set_object_terms( $object_id, $terms, $tt_ids, $taxonomy, $append, $old_tt_ids ) {
```
mjangda: Would this lead to double indexing in the case of a post update?

nickdaugherty: Good point, yes it would. I've reworked this to use EP's SyncManager queue, which de-dupes on the same request (so can only queue the same post once).

nickdaugherty:
This code would be better off in EP directly...I've opened a PR to move it there and add additional indexing checks that are missing from here to match the indexing behavior of regular post updates (there are a variety of things that need checked before we queue a post for indexing).

PR #1475, opened 2020-02-26
Add custom headers to all ElasticSearch requests

chrean:
## Description

Fixes [PLAT-932]

Adds following custom header to all requests sent to ElasticSearch

```
X-Client-Site-ID
X-Client-Env
```

## Checklist

## Steps to Test

1. Check out PR.
1. Sandbox a site with VIP Search enabled (e.g. 1120)
1. How can we see whether those headers are really sent out or not? `var_dump()`ing shows they are actually used, but I don't know how we can be 100% sure they are making all their way to ES. 🤔

[PLAT-932]: https://vipjira.atlassian.net/browse/PLAT-932

### DIFF
--- a/elasticsearch/includes/classes/class-elasticsearch.php
+++ b/elasticsearch/includes/classes/class-elasticsearch.php
@@ -108,6 +108,11 @@ class Elasticsearch {
 		if ( apply_filters( 'debug_bar_enable', false ) || apply_filters( 'wpcom_vip_qm_enable', false ) ) {
 			// Load ElasticPress Debug Bar
 			require_once __DIR__ . '/../../debug-bar-elasticpress/debug-bar-elasticpress.php';
+
+			// And ensure the logging has been setup (since it also hooks on plugins_loaded)
+			if ( function_exists( 'ep_setup_query_log' ) ) {
+				ep_setup_query_log();
+			}
 		}
 	}
 
--- a/elasticsearch/includes/classes/class-elasticsearch.php
+++ b/elasticsearch/includes/classes/class-elasticsearch.php
@@ -144,6 +144,9 @@ class Elasticsearch {
 		$timeout = $this->get_http_timeout_for_query( $query );
 
 		// Add custom headers to identify authorized traffic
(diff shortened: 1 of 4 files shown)


elasticsearch/includes/classes/class-elasticsearch.php
```diff
@@ -143,6 +143,8 @@ public function filter__ep_do_intercept_request( $request, $query, $args, $failu
 
 		$timeout = $this->get_http_timeout_for_query( $query );
 
+		// Add custom headers to identify authorized traffic
+		$args['headers'] = array_merge( $args['headers'], array( 'X-Client-Site-ID' => get_current_blog_id(), 'X-Client-Env' => VIP_GO_ENV ) );
```
mjangda: Site ID should be the Go site ID not the WordPress site ID.

chrean: 7bb95bc2 if I got that right

mjangda: Oh, actually, I guess we should technically check that `$args['headers']` is an array first.

elasticsearch/includes/classes/class-elasticsearch.php
```diff
@@ -143,6 +143,8 @@ public function filter__ep_do_intercept_request( $request, $query, $args, $failu
 
 		$timeout = $this->get_http_timeout_for_query( $query );
 
+		// Add custom headers to identify authorized traffic
+		$args['headers'] = array_merge( $args['headers'], array( 'X-Client-Site-ID' => FILES_CLIENT_SITE_ID, 'X-Client-Env' => VIP_GO_ENV ) );
```
mjangda:
```suggestion
		if ( ! isset( $args['headers'] ) || ! is_array( $args['headers'] ) ) {
			$args['headers'] = [];
		}
		$args['headers'] = array_merge( $args['headers'], array( 'X-Client-Site-ID' => FILES_CLIENT_SITE_ID, 'X-Client-Env' => VIP_GO_ENV ) );
```

mjangda:
r802-stacks

PR #1476, opened 2020-02-28
Standardize two factor authentication SMS messages

pandah3:
## Description

Standardizing the message format for SMS messages used for two factor authentication to help enhance security practices.

Read more about the standardization proposal by Webkit here: https://github.com/WebKit/explainers/tree/master/sms-one-time-code-format

Old format:
`The verification code you requested is: 123456`

New format:
```
123456 is your FooBar verification code.

@foobar.com #123456
```

## Checklist

## Steps to Test

1. Check out PR.
1. Go to `wp-admin` > `Edit My Profile` > `Two-Factor Options` > `SMS`
3. Enter your SMS number and click Submit
4. Check your phone and verify the message format. It should look like:
```
123456 is your FooBar verification code.

@foobar.com #123456
```
5. PHPunit tests
Run `./bin/phpunit-docker.sh tests/wpcom-vip-two-factor/sms-provider.php`

This has been tested on a Go Sandbox using a test site 👍


pandah3:
I refactored the code to isolate the message formatting logic to make it easier to test, but it turned out way too simple(?)

This was my initial testing logic:
- Call `generate_and_send_token()`
- Isolate `$message` somehow b/c `generate_and_send_token()` returns a function that sends the message, it doesn’t return the message itself
- That function also calls `generate_token()`, which generates a random token, making it hard to test with a dummy token
- Test that the actual message matches the expected message

Not sure if I need to mock `send_sms()` to simulate sending the actual SMS message?

wpcom-vip-two-factor/sms-provider.php
```diff
@@ -79,14 +79,25 @@ public function delete_token( $user_id ) {
 		delete_user_meta( $user_id, self::TOKEN_META_KEY );
 	}
 
+	/**
+	 * Standardize the format of the SMS message
+	 *
+	 * @param string $token User token.
+	 */
+	public function format_sms_message( $token ) {
+		$message = $token . " is your " . $_SERVER['HTTP_HOST'] . " verification code.\r\n" . "@" . $_SERVER['HTTP_HOST'] . " #" . $token;
```
dchymko: You could break this across 2 lines

dchymko: Might be better/easier to use sprintf like in here:
https://github.com/Automattic/vip-go-mu-plugins/blob/master/cache/class-vary-cache.php#L48

Also, not sure if this needs it's own function, but I think maybe you did this to make testing easier?

pandah3: Broke the lines up!
Yes, it was my attempt to isolate the code for testing, but I opted to recombine the code to spy on the function instead like we discussed!

tests/wpcom-vip-two-factor/sms-provider.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+class Two_Factor_SMS__Test extends \WP_UnitTestCase {
+ public function setUp() {
+		require_once __DIR__ . '/../wpcom-vip-two-factor/sms-provider.php';
```
dchymko: This path needs one additional directory traversal.
```suggestion
		require_once __DIR__ . '/../../wpcom-vip-two-factor/sms-provider.php';
```

pandah3: Argh, didn't realize it's loaded as a plugin :bangs-head:. Ahh okay, you need to require both paths b/c `sms-provider.php` relies on `two-factor.php` in the shared plugins.

---
Just noting the original issue I had with testing for reference:

Following: https://phpunit.readthedocs.io/en/9.0/test-doubles.html#test-doubles

`Error: Class 'Automattic\VIP\TwoFactor\Two_Factor_Provider' not found`

or some variation of that. I tried adding a namespace (`Automattic\VIP\TwoFactor`) in the original class, then calling that, calling the regular class name, etc., but it wont recognize it.

--

I changed the code to use `createMock()` instead of `getMockBuilder` but I get:
`Error: Call to undefined method Mock_Two_Factor_SMS_383faa33::process_results()`

or

`Cannot stub or mock class or interface "Two_Factor_SMS" which does not exist
I don’t know why it’s not recognizing the class`

pandah3: Also, regarding the directory traversal, I initially had it set to
`/../../wpcom-vip-two-factor/sms-provider.php` (the suggested change),
but I noticed in this code, the traversal is missing one step:

https://github.com/Automattic/vip-go-mu-plugins/blob/master/tests/elasticsearch/includes/classes/test-class-healthjob.php#L7

If you try to run it with the correct path, it errors out. Not sure what that's about

pandah3:
Scrapping the tests for this change because the change is within `generate_and_send_token()` which calls `send_sms()`, which in-turn calls the Twillio API to actually send a message out.

Mocking `send_sms()` then trying to call `generate_and_send_token()` to see if `send_sms()` was called doesn't work out because `send_sms()` is not in the same class as `generate_and_send_token()` (`send_sms()` is actually [not in a class](https://github.com/Automattic/vip-go-mu-plugins/blob/master/lib/sms.php) at all).

Requiring the class for `generate_and_send_token()` (`Two_Factor_SMS`) + the namespace for the file that `send_sms()` is in (`namespace Automattic\VIP\SMS`) doesn't work.

Isolating the message formatting function and testing that alone wouldn't really be testing much.

tests/wpcom-vip-two-factor/sms-provider.php
dchymko: As we went through  today, this is because we were never actually calling any methods, just setting up the mock.

wpcom-vip-two-factor/sms-provider.php
dchymko: This looks good, in context of our testing challenges today, but any thoughts on using `sprintf` here?

wpcom-vip-two-factor/sms-provider.php
```diff
@@ -87,8 +87,9 @@ public function delete_token( $user_id ) {
 	public function generate_and_send_token( $user ) {
 		require_once( WPMU_PLUGIN_DIR . '/lib/sms.php' );
 		$token = $this->generate_token( $user->ID ); // Store in variable to prevent generating code twice
-		$message = $token . ' is your ' . $_SERVER['HTTP_HOST'] . " verification code.\r\n" .
-		'@' . $_SERVER['HTTP_HOST'] . ' #' . $token;
+		$hostname = isset( $_SERVER['HTTP_HOST'] ) ? $_SERVER['HTTP_HOST'] : ''; 
+		$message = $token . ' is your ' . $hostname . " verification code.\r\n" .
+		'@' . $hostname . ' #' . $token;
```
mjangda: `$hostname` should probably be the value of `home_url()` here (minus the protocol).

wpcom-vip-two-factor/sms-provider.php
```diff
@@ -87,8 +87,9 @@ public function delete_token( $user_id ) {
 	public function generate_and_send_token( $user ) {
 		require_once( WPMU_PLUGIN_DIR . '/lib/sms.php' );
 		$token = $this->generate_token( $user->ID ); // Store in variable to prevent generating code twice
-		$message = $token . ' is your ' . $_SERVER['HTTP_HOST'] . " verification code.\r\n" .
-		'@' . $_SERVER['HTTP_HOST'] . ' #' . $token;
+		$hostname = isset( $_SERVER['HTTP_HOST'] ) ? $_SERVER['HTTP_HOST'] : ''; 
+		$message = $token . ' is your ' . $hostname . " verification code.\r\n" .
```
mjangda: `$hostname` should probably be the site title here.

wpcom-vip-two-factor/sms-provider.php
pandah3: Ah, forgot that the home url could be different! Thank you for the catch!

wpcom-vip-two-factor/sms-provider.php
pandah3: Fixed!

wpcom-vip-two-factor/sms-provider.php
pandah3: Changed to using `sprintf()` for standardization across the codebase :)

pandah3:
Any ideas on how to get around calling the `format_sms_message()` method in the test without [instantiating a new class](https://github.com/Automattic/vip-go-mu-plugins/pull/1476/files#diff-85ccafbc3f85d6023259436c0ccf1f52R10)?

Instantiating the new class keeps giving me:

`Error: Call to protected Two_Factor_SMS::__construct() from context 'Two_Factor_SMS__Test'`

because of the `protected` function [here](https://github.com/Automattic/vip-go-mu-plugins/blob/master/wpcom-vip-two-factor/sms-provider.php#L30).

I tried calling the classes via a namespace, mocking the function so I won't have to call the real one, and simply just requiring the path alone, but no luck with anything so far.

---
Also, really not sure what's up with the failing Circle CI lint issues.
My phpcs isn't catching anything and `make lint` isn't spitting out anything either.
Will look into it more

tests/wpcom-vip-two-factor/sms-provider.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+class Two_Factor_SMS_Test extends WP_UnitTestCase {
+	public function setUp() {
+		require_once __DIR__ . '/../../shared-plugins/two-factor/two-factor.php';
+		require_once __DIR__ . '/../../wpcom-vip-two-factor/sms-provider.php';
+	}
+
+	public function test__two_factor_sms_formatting() {
+		$two_factor = new Two_Factor_SMS();
```
dchymko: Because of the protected constructor (following the pattern from the `Two_Factor_Provider` abstract class/interface), we can't use the `new` keyword here. Instead we have to call `get_instance` 
```suggestion
		$two_factor = Two_Factor_SMS::get_instance();
```

pandah3:
r816-stacks

PR #1483, opened 2020-03-04
Add header to indicate when search results come from Elasticsearch

dchymko:
## Description
This adds a custom HTTP header to indicate when the search results came from Elasticsearch.  This is necessary in cases where we want to ensure the search didn't fall back to using the wpdb query.

## Checklist

## Steps to Test

1. Check out PR on a site with ElasticPress enabled
1. Enter a search term and click "Search"
1. Verify that the custom header was sent in the HTTP response for the results request.

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -87,7 +87,7 @@ class Search {
 		// Round-robin retry hosts if connection to a host fails
 		add_filter( 'ep_pre_request_host', array( $this, 'filter__ep_pre_request_host' ), PHP_INT_MAX, 4 );
 		
-		add_filter( 'ep_valid_response', array( $this, 'filter__ep_valid_response' ), 10 );
+		add_filter( 'ep_valid_response', array( $this, 'filter__ep_valid_response' ), 10, 4 );
 	}
 
 	protected function load_commands() {
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -86,6 +86,8 @@ class Search {
 
 		// Round-robin retry hosts if connection to a host fails
 		add_filter( 'ep_pre_request_host', array( $this, 'filter__ep_pre_request_host' ), PHP_INT_MAX, 4 );
+		
+		add_filter( 'ep_valid_response', [ $this, 'filter__ep_valid_response' ], 10 );
 	}
 
 	protected function load_commands() {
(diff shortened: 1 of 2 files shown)


dchymko:
Going to try and add a test for the headers similar to https://github.com/Automattic/vip-go-mu-plugins/blob/master/tests/cache/test-vary-cache.php#L483

search/includes/classes/class-search.php
```diff
@@ -237,4 +240,14 @@ static function ep_skip_query_integration( $skip ) {
 		// The filter is checking if we should _skip_ query integration
 		return ! ( $query_integration_enabled || $query_integration_enabled_legacy );
 	}
+
+	public function filter__ep_valid_response( $response, $query, $query_args, $query_object ) {
+		if ( ! headers_sent() ) {
+			/**
+			 * Manually set a header to indicate the search results are from Elasticsearch
+			 */
+			header( 'X-ElasticPress-Search-Valid-Response: true' );
```
mjangda: Should we name this to be more specific to VIP Search? Also ElasticPress already sends a header which means we now have two headers with this change?

brandon-m-skinner: ElasticPress sends a header when it's involved in the process in any capacity, even if the query never hits ES. 

The idea of this header is to give an indication of when ES is actually involved. See [this Slack thread](https://a8c.slack.com/archives/C03CU7X3U/p1583338418088500) for context.

dchymko: Yes, there are two headers, but we mainly want to know "did the search results actually come from ElasticPress"
The baked in header seems kind of.. less useful, but I don't think we want to swap it out or remove it.

dchymko:
Wondering if maybe we should only output this header when a special querystring param (or header) is passed in?

elazzabi:
@dchymko I guess it would be good to have this only when a param/header is present so we don't pollute the headers.

Should we take into consideration caching when taking the param/header decision? I think requests with different params are considered different (cache wise), but not sure about the header one 🤔

dchymko:
Hmm, what if we trigger it on the `WP_DEBUG` constat _or_ a `ep_debug` param on the querystring?

We use the constant approach in the vary cache work here:
https://github.com/Automattic/vip-go-mu-plugins/blob/master/cache/class-vary-cache.php#L593-L595

cc @nickdaugherty for thoughts/input

search/includes/classes/class-search.php
mjangda: If the original EP header is not very useful, we should definitely consider either removing it or making it more useful. (Although, I guess it may be impossible to do if a EP query/request is triggered halfway through a page-render).

My main concern is additional unnecessary bytes we're adding to each pageload. At a micro-level it's not a big deal, but globally, it adds up to a lot :)

mjangda:
> Hmm, what if we trigger it on the WP_DEBUG constat or a ep_debug param on the querystring?

Either of those could work. Since our primary use case right now is for e2e tests, which approach would be better/preferred?

dchymko:
>Either of those could work. Since our primary use case right now is for e2e tests, which approach would be better/preferred?

Going to move this forward by using the querystring approach for now. That way it can easily be used on multiple sites without having to do anything to the site config.

We should definitely consider replacing that header down the road though.

dchymko:
r831-stacks

PR #1484, opened 2020-03-04
Add retries and random load balancing for EP_HOST

brandon-m-skinner:
## Description
[PLAT-863]

Added a filter for `ep_pre_request_host` to round-robin retry our endpoints.

Added a random load balance implementation for setting EP_HOST from our endpoints.

## Checklist

## Steps to Test

1. Check out PR.

[PLAT-863]: https://vipjira.atlassian.net/browse/PLAT-863

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -81,6 +81,9 @@ class Search {
 
 		// Disable certain EP Features
 		add_filter( 'ep_feature_active', array( $this, 'filter__ep_feature_active' ), PHP_INT_MAX, 3 );
+
+		// Round-robin retry hosts if connection to a host fails
+		add_filter( 'ep_pre_request_host', array( $this, 'filter__ep_pre_request_host' ), PHP_INT_MAX, 4 );
 	}
 
 	protected function load_commands() {
@@ -237,6 +240,25 @@ class Search {
 		// The filter is checking if we should _skip_ query integration
 		return ! ( $query_integration_enabled || $query_integration_enabled_legacy );
 	}
+
+	static function filter__ep_pre_request_host( $host, $failures, $path, $args ) {
+		if( ! defined( 'VIP_ELASTICSEARCH_ENDPOINTS' ) ) { 
+			return $host;
+		}
+
+		return self::round_robin_hosts( $host, VIP_ELASTICSEARCH_ENDPOINTS );
+	}
+
+	static function round_robin_hosts( $host, $hostList ) {
+		$cur_index = array_search( $host, $hostList );
+		$max_index = count( $hostList ) - 1;
+
+		if ( $cur_index === $max_index or is_null( $cur_index ) ) {
+			return $hostList[ 0 ];
+		}
+
+		return $hostList[ $cur_index + 1 ];
+	} 
 
 	/*
 	 * Given a list of endpoints, randomly select one for load balancing purposes.
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -241,6 +241,11 @@ class Search {
 		return ! ( $query_integration_enabled || $query_integration_enabled_legacy );
 	}
 
+	/**
+	 * Filter for ep_pre_request_host
+	 *
+	 * Return the next host in our enpoint list if it's defined. Otherwise, return the last host.
+	 */
 	static function filter__ep_pre_request_host( $host, $failures, $path, $args ) {
 		if( ! defined( 'VIP_ELASTICSEARCH_ENDPOINTS' ) ) { 
 			return $host;
@@ -249,6 +254,9 @@ class Search {
 		return self::round_robin_hosts( $host, VIP_ELASTICSEARCH_ENDPOINTS );
 	}
 
+	/**
+	 * Return the next host in the list based on the current host
+	 */
 	static function round_robin_hosts( $host, $hostList ) {
 		$cur_index = array_search( $host, $hostList );
 		$max_index = count( $hostList ) - 1;
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -43,7 +43,7 @@ class Search {
 		}
 
 		if ( ! defined( 'EP_HOST' ) && defined( 'VIP_ELASTICSEARCH_ENDPOINTS' ) && is_array( VIP_ELASTICSEARCH_ENDPOINTS ) ) {
-			$host = VIP_ELASTICSEARCH_ENDPOINTS[ 0 ];
+			$host = self::load_balance_endpoints( VIP_ELASTICSEARCH_ENDPOINTS );
 
 			define( 'EP_HOST', $host );
 		}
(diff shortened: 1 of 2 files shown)


search/includes/classes/class-search.php
```diff
@@ -237,4 +240,44 @@ static function ep_skip_query_integration( $skip ) {
 		// The filter is checking if we should _skip_ query integration
 		return ! ( $query_integration_enabled || $query_integration_enabled_legacy );
 	}
+
+	/**
+	 * Filter for ep_pre_request_host
+	 *
+	 * Return the next host in our enpoint list if it's defined. Otherwise, return the last host.
+	 */
+	static function filter__ep_pre_request_host( $host, $failures, $path, $args ) {
```
nickdaugherty: This shouldn't actually be `static` (since we call it on the `$this` instance from the `add_filter()` call).

search/includes/classes/class-search.php
```diff
@@ -43,7 +43,7 @@ protected function setup_constants() {
 		}
 
 		if ( ! defined( 'EP_HOST' ) && defined( 'VIP_ELASTICSEARCH_ENDPOINTS' ) && is_array( VIP_ELASTICSEARCH_ENDPOINTS ) ) {
-			$host = VIP_ELASTICSEARCH_ENDPOINTS[ 0 ];
+			$host = self::load_balance_endpoints( VIP_ELASTICSEARCH_ENDPOINTS );
```
nickdaugherty: Thoughts on making this and the others an instance method (vs. static)? They don't technically need anything related to the instance, but also don't need to be called statically, so keeping them as instance methods is probably more consistent with the rest of the class.

search/includes/classes/class-search.php
```diff
@@ -237,4 +240,44 @@ static function ep_skip_query_integration( $skip ) {
...
+			return $host;
+		}
+
+		return self::round_robin_hosts( $host, VIP_ELASTICSEARCH_ENDPOINTS );
+	}
+
+	/**
+	 * Return the next host in the list based on the current host
+	 */
+	static function round_robin_hosts( $host, $hostList ) {
```
nickdaugherty: I would suggest naming the functions with the action that they perform for clarity...so this could be something along these lines:

```suggestion
	static function get_next_host_from_list( $host, $hostList ) {
```

search/includes/classes/class-search.php
```diff
@@ -237,4 +240,44 @@ static function ep_skip_query_integration( $skip ) {
...
+			return $hostList[ 0 ];
+		}
+
+		return $hostList[ $cur_index + 1 ];
+	} 
+
+	/**
+	 * Given a list of endpoints, randomly select one for load balancing purposes.
+	 */
+	static function load_balance_endpoints( $endpoints ) {
```
nickdaugherty: Similar note - this could be called something like this, which indicates what the function actually does (it gets a random host from the list)

```suggestion
	static function get_random_host( $endpoints ) {
```

search/includes/classes/class-search.php
```diff
@@ -237,4 +240,44 @@ static function ep_skip_query_integration( $skip ) {
...
+		if( ! is_array( $endpoints ) ) {
+			return $endpoints;
+		}
+
+		return $endpoints[  
+			rand( 0, count( $endpoints ) - 1 )
```
nickdaugherty: This can be simplified with `array_rand()`:

```suggestion
			array_rand( $endpoints )
```

tests/search/test-class-search.php
```diff
@@ -127,7 +127,7 @@ public function test__vip_search_connection_constants() {
 		$es = new \Automattic\VIP\Search\Search();
 		$es->init();
 
-		$this->assertEquals( EP_HOST, 'https://es-endpoint1' );
+		$this->assertEquals( true, 'https://es-endpoint1' === EP_HOST or 'https://es-endpoint2' === EP_HOST );
```
nickdaugherty: How about using `assertContains()` here?
```suggestion
		$this->assertContains( EP_HOST, VIP_ELASTICSEARCH_ENDPOINTS );
```

tests/search/test-class-search.php
```diff
@@ -488,4 +488,54 @@ public function test__vip_search_setup_healthchecks_disabled_in_non_production_e
...
+		$this->assertEquals( 'test', $es::filter__ep_pre_request_host( 'test', 0, '', [] ) );
+
+		define( 'VIP_ELASTICSEARCH_ENDPOINTS', array(
+			'endpoint1',
+			'endpoint2',
+			'endpoint3',
+			'endpoint4',
+		) );
+
+		$this->assertEquals( true, in_array( $es::filter__ep_pre_request_host( 'endpoint1', 0, '', [] ), VIP_ELASTICSEARCH_ENDPOINTS ) );
```
nickdaugherty: Can use `assertContains` here too.

tests/search/test-class-search.php
```diff
@@ -488,4 +488,54 @@ public function test__vip_search_setup_healthchecks_disabled_in_non_production_e
...
+	public function test__vip_search_load_balance_endpoints() {
+		$endpoints = [
+			'endpoint1',
+			'endpoint2',
+			'endpoint3',
+			'endpoint4',
+		];
+		$es = new \Automattic\VIP\Search\Search();
+
+		$this->assertEquals( true, in_array( $es::load_balance_endpoints( $endpoints ), $endpoints ) );
```
nickdaugherty: Same note about `assertContains()`

search/includes/classes/class-search.php
brandon-m-skinner: Sure, makes sense.

brandon-m-skinner:
I was planning to go through and review all this before submitting for review, but you beat me to it. 

You're a mind reader 🙂.

tests/search/test-class-search.php
```diff
@@ -488,4 +488,57 @@ public function test__vip_search_setup_healthchecks_disabled_in_non_production_e
...
+
+	/**
+	 * Test that checks both single and multi-host retries
+	 */
+	public function test__vip_search_filter__ep_pre_request_host() {
+		$es = new \Automattic\VIP\Search\Search();
+		$es->init();
+
+		// If VIP_ELASTICSEARCH_ENDPOINTS is not defined, just hand the last host back
+		$this->assertEquals( 'test', $es->filter__ep_pre_request_host( 'test', 0, '', array() ) );
```
mjangda: I realize that we haven't been following this on the existing tests in this file, but as a general rule, whenever you have multiple assertions in the same test, it's helpful to add the `$message` param to indicate which assertion failed and why.

```suggestion
		$this->assertEquals( 'test', $es->filter__ep_pre_request_host( 'test', 0, '', array() ), 'pre_request_host filter did not return passed in host without constant defined.' );
```

This way, when a test fails, we get more obvious details around which assertion was the problem.

search/includes/classes/class-search.php
```diff
@@ -237,4 +240,42 @@ static function ep_skip_query_integration( $skip ) {
...
+	 * Filter for ep_pre_request_host
+	 *
+	 * Return the next host in our enpoint list if it's defined. Otherwise, return the last host.
+	 */
+	public function filter__ep_pre_request_host( $host, $failures, $path, $args ) {
+		if ( ! defined( 'VIP_ELASTICSEARCH_ENDPOINTS' ) ) { 
+			return $host;
+		}
+
+		return $this->get_next_host( $host, VIP_ELASTICSEARCH_ENDPOINTS );
```
nickdaugherty: I just realized, this will end up trying the "next" host for every request, because this filter always runs before any request. What that means is that if one of the hosts is down, it will keep being retried every Nth time (if there are enough ES requests) instead of it settling on a working host after one failed request.

nickdaugherty: I think the ideal solution is that it uses the first one from `EP_HOST` by default (which is itself randomized)...then if and only if the request fails, it moves on to the next host in the list - otherwise it keeps hitting the host that it knows is working.

brandon-m-skinner:
Scheduling for deployment tomorrow since JP deployment is today.

brandon-m-skinner:
r817-stacks

PR #1486, opened 2020-03-05
Add husky to enforce phpcs on pre-commit

mjangda:
Uses husky to automatically add a pre-commit hook that auto-runs `phpcs:changed:local`  automatically on commit (assuming `npm install` has been run).

```
$ git ci
husky > pre-commit (node v10.15.2)

> @automattic/vip-go-mu-plugins@1.0.0 phpcs:changed:local /Users/mjangda/Sites/vipgo/mu-plugins-docker/mu-plugins
> vendor/bin/phpcs --report=bin/phpcs-eslint-report.php | eslines --quiet -d index

[update/husky-pre-commit ff53fd2f] Update README with details about PHPCS
 1 file changed, 19 insertions(+)
```

## Checklist

n/a

## Steps to Test

1. `composer install`
1. `npm install`
1. Edit a file and add a PHPCS issue (e.g. `$a     = [  ];`)
1. `git add` + `git commit`

The commit should fail.

### DIFF
--- a/.huskyrc.json
+++ b/.huskyrc.json
@@ -1,5 +1,5 @@
 {
   "hooks": {
-    "pre-commit": "npm run phpcs:changed:local"
+    "pre-push": "npm run phpcs:changed:local"
   }
 }
--- a/.huskyrc.json
+++ b/.huskyrc.json
@@ -1,0 +1,5 @@
+{
+  "hooks": {
+    "pre-commit": "npm run phpcs:changed:local"
+  }
+}
--- a/README.md
+++ b/README.md
@@ -18,6 +18,25 @@ These are generated via CI by the [`generate-docs.sh`]() script.
 make lint
 ```
 
+#### PHPCS
+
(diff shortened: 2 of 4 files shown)


brandon-m-skinner:
Having this on commits is going to be potentially annoying and slow down local development, especially if you do a bunch of small contained commits.

Would it make more sense to use the pre-push hook instead to mitigate this? Same net effect without the drawback of slowing down workflows locally. I personally prefer to lint before I push, so I might be biased.

mjangda:
Yeah, that makes sense; definitely open to switching to pre-push instead.

.huskyrc.json
```diff
@@ -0,0 +1,5 @@
+{
+  "hooks": {
+    "pre-commit": "npm run phpcs:changed:local"
```
brandon-m-skinner:
```suggestion
    "pre-push": "npm run phpcs:changed:local"
```

PR #1493, opened 2020-03-09
Add index content validation

nickdaugherty:
## Description

Adds in-depth ES index content validation helpers

Requires: https://github.com/Automattic/ElasticPress/pull/12

## Checklist

## Steps to Test

1. Pull down PR
1. Index all content with `wp elasticpress index`
1. Disable VIP Search in the config (remove the constant)
1. Edit a post title and close that page in WP Admin (to prevent auto saves from interfering)
1. Re-enable search
1. Run the validation command - `wp vip-search health validate-contents`
1. Inconsistencies should be found and reported (post title or whatever was changed)
1. Re-index with `wp elasticpress index`
1. Re-validate, no inconsistencies should be reported

### DIFF
--- a/search/includes/classes/class-health.php
+++ b/search/includes/classes/class-health.php
@@ -265,7 +265,7 @@ class Health {
 		$missing_from_index = array_diff( $expected_post_ids, $found_document_ids );
 
 		// If anything is missing from index, record it
-		if ( count( $missing_from_index ) ) {
+		if ( 0 < count( $missing_from_index ) ) {
 			foreach( $missing_from_index as $post_id ) {
 				$diffs[ 'post_' . $post_id ] = array( 
 					'expected' => sprintf( 'Post %d not to be indexed', $post_id ),
@@ -278,7 +278,7 @@ class Health {
 		$extra_in_index = array_diff( $found_document_ids, $expected_post_ids );
 
 		// If anything is in the index that shouldn't be, record it
-		if ( count( $extra_in_index ) ) {
+		if ( 0 < count( $extra_in_index ) ) {
 			foreach( $extra_in_index as $document_id ) {
 				// Grab the actual doc from 
 				$diffs[ 'post_' . $document_id ] = array( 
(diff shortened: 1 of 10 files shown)


search/includes/classes/class-health.php
```diff
@@ -141,6 +144,215 @@ public static function validate_index_posts_count() {
...
+		$expected_post_ids = wp_list_pluck( $expected_post_rows, 'ID' );
+		$found_document_ids = wp_list_pluck( $documents, 'ID' );
+
+		$diffs = [];
+
+		// What's missing in ES?
+		$missing_from_index = array_diff( $expected_post_ids, $found_document_ids );
+
+		// If anything is missing from index, record it
+		if ( count( $missing_from_index ) ) {
```
brandon-m-skinner: Should probably do an explicit value check for clarity rather than rely on 0 being implicitly falsey.

search/includes/classes/class-health.php
```diff
@@ -141,6 +144,215 @@ public static function validate_index_posts_count() {
...
+			foreach( $missing_from_index as $post_id ) {
+				$diffs[ 'post_' . $post_id ] = array( 
+					'expected' => sprintf( 'Post %d not to be indexed', $post_id ),
```
brandon-m-skinner: Shouldn't this be `'Post %d to be indexed'`?

search/includes/classes/class-health.php
nickdaugherty: 👍 updated in 54e17c34

search/includes/classes/class-health.php
nickdaugherty: Good catch!

nickdaugherty: Fixed in 39d3b4e2 (there was a conflict with another change I just pushed)

nickdaugherty:
I've added tests for some of the helper functions - the main looping itself isn't covered by unit tests, but the complication of setting up actual posts and an index may not be worth it at this point.

nickdaugherty:
r829-stacks

Issue #453, opened 2017-02-05
Alloptions: Ability to switch autoload value for an option


mjangda:
It would be nice to have command that lets us easily swap the `autoload` value for any given option on a site.

It should:

- [ ] Update the value in the database
- [ ] Flush the necessary caches

Might be good to upstream to wp-cli and/or core.

nickdaugherty:
@trepmal noticed the other day that WP-CLI includes a `--autoload` parameter to `option update` out of the box, so I think we can just use that.

https://developer.wordpress.org/cli/commands/option/update/#options

trepmal:
After some experimenting, this may not work as smoothly as we'd like. A separate direct command may be desirable.

Like `update_option()`, `autoload` is only updated if the value is changed.

---
given this db row:

option_name | option_value | autoload
------------- | ------------ | ---------
greeting | hi | yes

❌ `update_option( 'greeting', 'hi', 'no' );` // value = old value, bails early
❌ `wp option update greeting hi --autoload=no` // value = old value, bails early
❌ `wp option update greeting --autoload=no` *it appears to think, but it actually is a prompt for you to type a new value and hit ctrl+D. if same as old value, guess what, bails early

It seems that the only way to make this work is 1) change the value or 2) use SQL.

nickdaugherty:
Ah, yes that would greatly limit the utility for us then, thanks for looking into it. Ok, let's go ahead with a separate command.

trepmal:
https://github.com/trepmal/option-autoload-cli

GaryJones:
Fixed in #1496.

PR #1496, opened 2020-03-10
Add WP-CLI command to manage option autoload

trepmal:
## Description

Add command for managing option autoload. See #453

Also https://github.com/trepmal/option-autoload-cli

## Checklist

- [x] This change has relevant documentation additions / updates (if applicable). **has inline docs**

## Steps to Test

With PR applied, can test with:

```
wp option autoload get home # should be 'yes' at start
wp option autoload set home no
wp option autoload get home # should be 'no'
wp option autoload set home yes # restore
wp option autoload list no # should list starting with `moderation_keys`, `recently_edited`
```


mjangda:
@trepmal nice work!

Are there differences between this and the command on your repo? Ideally they'd be the same.

trepmal:
The only diffs between this and my repo are 
1) this version is condensed. This class and add_command line are in this one file. 
2) this uses WPCOM_VIP_CLI_Command

emrikol:
Cool!  This is only a temporary thing, right?  Once the option is `update_option()`'d again without autoload, will it revert to autoloading again?

WPprodigy:
> Cool! This is only a temporary thing, right? Once the option is update_option()'d again without autoload, will it revert to autoloading again?

Sort of. If the option isn't technically updated/changed, then the next `update_option()` won't make any difference to the autoloading - afaik.

But secondly, I'm pretty positive that the third `$autoload` arg is rarely used. If the code specifically asks for autoloading, then it makes sense to respect that. In cases where we need to flip the switch, it's quite likely there is no preference to autoloading in the code. <2¢ />

trepmal:
> Cool! This is only a temporary thing, right? Once the option is update_option()'d again without autoload, will it revert to autoloading again?

If the `autoload` arg is not set, then changing the value in the db will stick.
If `autoload` is set, it will be applied the next time the option is updated _with a new value_

```
$ wp option update TESTING test      # prep option
Success: Updated 'TESTING' option.

$ wp option autoload get TESTING     # default autoload
yes

$ wp option autoload set TESTING no  # change it
Success: Autoload changed. Cache flushed.

$ wp option update TESTING test2     # update value
Success: Updated 'TESTING' option.

$ wp option autoload get TESTING     # autoload unchanged
no

$ wp option update TESTING test3 --autoload=yes # update value with new autoload
Success: Updated 'TESTING' option.

$ wp option autoload get TESTING     # autoload changed
yes

$ wp option update TESTING test3 --autoload=no # same value with new autoload
Error: Could not update option 'TESTING'.

$ wp option autoload get TESTING     # autoload unchanged
yes
```

I feel like once upon a time `autoload` could only be set with `add_option()`? It seems like I rarely see the autoload arg with `update_option()`.

I think we'd only run into a self-unsolving issue is if something is forcibly setting autoload to 'yes' , which seems quite unlikely.

rogertheriault:
Are our style requirements that strict? Wow. https://circleci.com/gh/Automattic/vip-go-mu-plugins/3392

mjangda:
> Are our style requirements that strict?

We're just inheriting from the WordPress + VIP Go standards. Also open to tweaks on standards we don't agree with.

mjangda:
> The only diffs between this and my repo are

It would be good to minimize the differences so it's easier to port changes across. I think `1` is probably fine but I would drop `2` unless there are strong reasons for inheriting from the WPCOM class.

wp-cli/vip-option-autoload.php
```diff
@@ -0,0 +1,219 @@
...
+				$cache_success = ( isset( $alloptions_before[ $option ] ) ) && ( ! isset( $alloptions_after[ $option ] ) );
+			break;
+			case 'yes' :
+				$cache_success = ( ! isset( $alloptions_before[ $option ] ) ) && ( isset( $alloptions_after[ $option ] ) );
+			break;
+		}
+
+		WP_CLI::success( sprintf(
+			'Autoload changed.%s',
+			WP_CLI::colorize( $cache_success ? ' Cache flushed.' : ' %rCache flush failed.%n' )
```
hanifn: The spacing looks a little odd here.

Maybe something like this is better?
```suggestion
			'Autoload changed. %s',
			WP_CLI::colorize( $cache_success ? 'Cache flushed.' : '%rCache flush failed.%n' )
```

brandon-m-skinner: It's the difference between:

`Success: Autoload changed. Cache flushed.`

and

`Success: Autoload changed.Cache flushed.`

The space should stay for formatting purposes.

brandon-m-skinner: It's needed for formatting. Otherwise there's no space between the cache statement and the rest of the output.

brandon-m-skinner:
@trepmal I just sorted out the lint issues since I've been dealing with them a lot lately. If this all looks good to you, we'll approve and deploy :+1:

wp-cli/class-option-autoload.php
```diff
@@ -0,0 +1,226 @@
...
+		// option list uses on/off ¯\_(ツ)_/¯
+		$yn = 'yes' === $yn ? 'on' : 'off';
+
+		WP_CLI::run_command(
+			[ 'option', 'list' ],
+			[
+				'autoload' => $yn,
+				'fields' => 'option_name',
+				'format' => $assoc_args['format'],
+			],
```
trepmal, line 181: This comma is valid, but I hate it :)

trepmal: Well, it's only valid on PHP 7.3+ So I'm not keeping it upstream.

trepmal:
I've ported most of the changes upstream and tests are passing there. See inline comment for info

wp-cli/vip-option-autoload.php
hanifn: Yes but when using something like `sprintf()`, you would usually put the spaces in the format string instead of the parameters being inserted. That way, the formatting is handled by the formatting string alone and is not a concern when inserting the parameters. Still, it's minor matter.

brandon-m-skinner: Whoops, sorry I'm blind 🤦‍♂

wp-cli/class-option-autoload.php
brandon-m-skinner: SVN pre-commit hook didn't like it either. It's been removed in master.

brandon-m-skinner:
r868-stacks

PR #1499, opened 2020-03-11
Add logstash support

hanifn:
## Description

Add a helper class and function to support logging to logstash

## Checklist

## Steps to Test

TBD

### DIFF
--- a/logstash/class-logger.php
+++ b/logstash/class-logger.php
@@ -350,29 +350,33 @@ class Logger {
 		if ( static::$processed_entries ) {
 			return; // Already done.
 		}
+
+		$fallback_error = new \WP_Error( 'logstash-send-failed', 'There was an error connecting to the logstash endpoint' );
+
 		static::$processed_entries = true;
 		$endpoint = 'https://public-api.wordpress.com/rest/v1.1/logstash/bulk';
 
 		$entry_chunks = array_chunk( static::$entries, static::BULK_ENTRIES_COUNT );
+
 		// Process all entries.
 		foreach ( $entry_chunks as $entries ) {
 			if ( ! defined( 'VIP_GO_ENV' ) || ! VIP_GO_ENV ) {
 				static::maybe_wp_debug_log_entries( $entries );
 				continue; // Bypassing REST API below in this case.
 			}
-			$json_data     = wp_json_encode( $entries );
+
+			$json_data = wp_json_encode( $entries );
+
 			// Send to logstash via REST API endpoint with payload containing log entry details.
-			$_wp_remote_response         = wp_remote_post(
-				$endpoint,
-				[
-					'timeout'     => 2,
+			$_wp_remote_response = vip_safe_wp_remote_request( $endpoint, $fallback_error, 3, 2, 5, [
+					'method' => 'POST',
 					'redirection' => 0,
-					'blocking'    => false,
-					'body'        => [
+					'blocking' => false,
+					'body' => [
 						'params' => $json_data,
 					],
-				]
-			);
+				] );
+
 			$_wp_remote_response_code    = wp_remote_retrieve_response_code( $_wp_remote_response );
 			$_wp_remote_response_message = wp_remote_retrieve_response_message( $_wp_remote_response );
 
--- a/logstash/class-logger.php
+++ b/logstash/class-logger.php
@@ -337,7 +337,7 @@ class Logger {
 
 		// Sends data to logstash on shutdown.
 		if ( ! has_action( 'shutdown', [ static::class, 'process_entries_on_shutdown' ] ) ) {
-			add_action( 'shutdown', [ static::class, 'process_entries_on_shutdown' ] );
+			add_action( 'shutdown', [ static::class, 'process_entries_on_shutdown' ], 9999 );
 		}
 	}
 
@@ -354,6 +354,12 @@ class Logger {
 		$fallback_error = new \WP_Error( 'logstash-send-failed', 'There was an error connecting to the logstash endpoint' );
 
 		static::$processed_entries = true;
+
+		if ( function_exists( 'fastcgi_finish_request' ) ) {
+			// Flush content to client first to prevent slow page load
+			fastcgi_finish_request();
+		}
+
 		$endpoint = 'https://public-api.wordpress.com/rest/v1.1/logstash/bulk';
 
 		$entry_chunks = array_chunk( static::$entries, static::BULK_ENTRIES_COUNT );
--- a/logstash/class-logger.php
+++ b/logstash/class-logger.php
@@ -375,13 +375,13 @@ class Logger {
 
 			// Send to logstash via REST API endpoint with payload containing log entry details.
 			$_wp_remote_response = vip_safe_wp_remote_request( $endpoint, $fallback_error, 3, 5, 5, [
-					'method' => 'POST',
-					'redirection' => 0,
-					'blocking' => false,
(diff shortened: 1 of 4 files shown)


logstash/logstash.php
```diff
@@ -0,0 +1,22 @@
+<?php
+/**
+ * Plugin Name: VIP Logstash integration
+ * Description: Helper functions and classes for logstash integration.
+ * Author: Automattic
+ * License: GPL version 2 or later - http://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+ */
+
+require_once __DIR__ . '/class-logger.php';
```
mjangda: I think we're missing `namespace Automattic\VIP\Logstash;` here?

hanifn: Added in 23d6080b108b26d6c052af4f5a2622f331ef2a84

brandon-m-skinner:
The `wp_debug_log` stuff works well. Testing text included:

💗⛵  Ťє𝕊ｔ𝒾𝔫𝐠  👊♬
 😎♬  ℍᵉｌｌ𝔬 т𝐄ᔕt  🍔😳

logstash/class-logger.php
```diff
@@ -0,0 +1,426 @@
...
+	}
+
+	/**
+	 * Save message to log file
+	 *
+	 * @since 2020-01-10
+	 *
+	 * @param array $entry Data.
+	 */
+	public static function wp_debug_log( array $entry ) : void {
```
nickdaugherty: We need extra protections here against this running on the Go infrastructure. Since it's a public function, any code can call it, which will attempt (and hopefully fail) to write to `WP_CONTENT_DIR . '/debug.log'`.

I don't think there is a valid case for this to run on Go (only on sandboxes or locally), so we should check the constants to ensure it's only allowed in those spots.

logstash/class-logger.php
```diff
@@ -0,0 +1,426 @@
...
+		$entry_chunks = array_chunk( static::$entries, static::BULK_ENTRIES_COUNT );
+		// Process all entries.
+		foreach ( $entry_chunks as $entries ) {
+			if ( ! defined( 'VIP_GO_ENV' ) || ! VIP_GO_ENV ) {
+				static::maybe_wp_debug_log_entries( $entries );
+				continue; // Bypassing REST API below in this case.
+			}
+			$json_data     = wp_json_encode( $entries );
+			// Send to logstash via REST API endpoint with payload containing log entry details.
+			$_wp_remote_response         = wp_remote_post(
```
nickdaugherty: Let's switch this to use `vip_safe_wp_remote_request()` instead.

logstash/class-logger.php
```diff
@@ -0,0 +1,426 @@
...
+	/**
+	 * Sends data to logstash via REST API on shutdown, viewable in Kibana.
+	 *
+	 * @since 2020-01-10
+	 */
+	public static function process_entries_on_shutdown() : void {
+		if ( static::$processed_entries ) {
+			return; // Already done.
+		}
+		static::$processed_entries = true;
```
denho: To ensure we don't slow down page load by the time it takes for the API requests, we could send the response data to the client with `fastcgi_finish_request()` before delivering the logs.
```suggestion
		static::$processed_entries = true;

		if ( function_exists( 'fastcgi_finish_request' ) ) {
			fastcgi_finish_request();
		}
```

This will probably need to set a high number as a priority for the `shutdown` hook to ensure this code is executed at last. Or we could try utilizing the native `register_shutdown_function()` instead of the WP hook.

logstash/class-logger.php
```diff
@@ -0,0 +1,426 @@
...
+			trigger_error( 'Invalid `extra` in call to ' . esc_html( __METHOD__ ) . '(). Must be an ojbect, array, or scalar value.', E_USER_WARNING );
+			return; // Failed validation.
+
+		} elseif ( isset( $data['extra'] ) && strlen( wp_json_encode( $data['extra'] ) ) > static::MAX_ENTRY_EXTRA_SIZE ) {
+			trigger_error( 'Invalid `extra` in call to ' . esc_html( __METHOD__ ) . '(). Must be ' . esc_html( static::MAX_ENTRY_EXTRA_SIZE ) . ' bytes or less.', E_USER_WARNING );
+			return; // Failed validation.
+		}
+
+		// Adjust `feature` and JSON-encode pretty-print data.
+		$data['feature']        = 'a8c_vip_' . $data['feature'];
```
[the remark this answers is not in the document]
hanifn: Fixed in 338422ea998b4274ac2f0b43e5acebdfb8a60449

logstash/class-logger.php
```diff
@@ -0,0 +1,426 @@
...
+			'site_id'  => get_current_network_id(),                  // Required.
+			'blog_id'  => get_current_blog_id(),                     // Required.
+			'host'     => strtolower( $_SERVER['HTTP_HOST'] ?? '' ), // phpcs:ignore -- Optional.
+
+			'severity' => '',                                        // Optional.
+			'feature'  => '',                                        // Required.
+			'message'  => '',                                        // Required.
+
+			'user_id'  => get_current_user_id(),                     // Optional.
+			'user_ua' => $_SERVER['HTTP_USER_AGENT'] ?? '',          // phpcs:ignore -- Optional.
```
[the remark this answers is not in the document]
hanifn: Fixed in 338422ea998b4274ac2f0b43e5acebdfb8a60449

logstash/class-logger.php
hanifn: Nice catch! Added in 51df58d0e420dcc8590d687a967f5f12fcb4be04

logstash/class-logger.php
hanifn: Changed in 24fbbbb92f0bb45b52fcaf26893a67b6e0952e3d

logstash/class-logger.php
hanifn: Good idea! Changed in 9d673368bd6ca699ed23e3ee230f9e77730e60df

logstash/class-logger.php
```diff
@@ -0,0 +1,439 @@
...
+		foreach ( $entry_chunks as $entries ) {
+			if ( ! defined( 'VIP_GO_ENV' ) || ! VIP_GO_ENV ) {
+				static::maybe_wp_debug_log_entries( $entries );
+				continue; // Bypassing REST API below in this case.
+			}
+
+			$json_data = wp_json_encode( $entries );
+
+			// Send to logstash via REST API endpoint with payload containing log entry details.
+			$_wp_remote_response = vip_safe_wp_remote_request( $endpoint, $fallback_error, 3, 2, 5, [
```
nickdaugherty: Now that we switched to using `fastcgi_finish_request()`, we can bump up the timeouts here since this won't be blocking the client request. ~10 seconds is probably ok here.

hanifn: Good point but `vip_safe_wp_remote_request()` maxes out the timeout value at 5 seconds

hanifn: Changed in 33fcad1bcf9ca21ee9ef1772481d14b09c8eb31d

nickdaugherty: Oh right - ok 5 should be fine...

hanifn:
r837-stacks

bgrgicak:
Just to make sure, was the `logstash/bulk` endpoint implemented? [The diff](https://code.a8c.com/D39990) that should create this endpoint still isn't deployed.

PR #1504, opened 2020-03-13
Allow queries during indexing

nickdaugherty:
## Description

The index isn't dropped or anything, docs are just re-indexed in-place, so it's safe to allow queries against the index while a bulk index is being performed.

NOTE - the filter doesn't yet exist, but will as of https://github.com/Automattic/ElasticPress/pull/15 (safe to deploy without it though)

## Checklist

## Steps to Test

1. First have to be sure you're running the new filter from https://github.com/Automattic/ElasticPress/pull/15
1. Then start a bulk index with `wp elasticpress index`
1. While the index is running, perform some searches (`?s=test`)
1. EP Debug Bar should report queries were performed while the indexing was occurring


nickdaugherty:
r833-stacks

PR #1505, opened 2020-03-13
Revert "Limit calls to `Two_Factor_Core::is_user_using_two_factor()`"

nickdaugherty:
Reverts Automattic/vip-go-mu-plugins#1445

### DIFF
--- a/two-factor.php
+++ b/two-factor.php
@@ -32,7 +32,7 @@ function wpcom_vip_should_force_two_factor() {
 		return false;
 	}
 
-	if ( apply_filters( 'wpcom_vip_is_user_using_two_factor', false ) ) {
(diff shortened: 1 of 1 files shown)


nickdaugherty:
r834-stacks

mjangda:
Reverted due to some fatals caused when `--skip-plugins` was set for WP-CLI calls:

> PHP Fatal error:  Uncaught Error: Class 'Two_Factor_Core' not found in wp-content/mu-plugins/two-factor.php:144

We load the two-factor plugin using `wpcom_vip_load_plugin` but that skips when skip-plugins is set. However, we were still adding the hook `enforce` assuming the plugin was correctly loaded which then fataled.

PR #1510, opened 2020-03-18
Add constant to use core cron

davidsword:
## Description

Add constant to use core cron, instead of cron-control.

May be useful for huge multi sites that run only a few smaller jobs.

Bailing as early as possible. Using a constant instead of filters due to how early this file is.

## Checklist

## Steps to Test

1. Put some xdebug break points in cron-control/cron-control.php plugin
1. load site, notice cron-control.php loads
1. add `define( 'WPCOM_VIP_USE_CORE_CRON', true );` to vip-config
1. reload site, notice cron-control.php doesn't load
1. check the cron queue, note upcoming or due events
1. wait a few minutes, refresh site a few times (reminder to bust the cache)
1. check cron queue again, that jobs moved from core cron working

### DIFF
--- a/001-cron.php
+++ b/001-cron.php
@@ -16,6 +16,11 @@
  * @return bool
  */
 function wpcom_vip_use_core_cron() {
+	
+	if ( defined('WPCOM_VIP_USE_CORE_CRON') && WPCOM_VIP_USE_CORE_CRON ) {
+		return true;
+	}
+
 	// Do not load outside of VIP environments, unless explicitly requested
 	if ( false === WPCOM_IS_VIP_ENV && ( ! defined( 'WPCOM_VIP_LOAD_CRON_CONTROL_LOCALLY' ) || ! WPCOM_VIP_LOAD_CRON_CONTROL_LOCALLY ) ) {
 		return true;
--- a/001-cron.php
+++ b/001-cron.php
@@ -17,7 +17,9 @@
  */
 function wpcom_vip_use_core_cron() {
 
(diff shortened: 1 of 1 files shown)


davidsword:
Setting as draft until I write a test for it, and check if any docs should be updated about this.

davidsword:
No tests as `wpcom_vip_use_core_cron()` is `true` for local dev (phpunit tests) anyways.

imho we don't need any front-facing documentation because CronControl is the correct solution the vast majority of the time. Having reverting to core in docs may lead to disabling instead of properly troubleshooting.

001-cron.php
```diff
@@ -16,6 +16,11 @@
  * @return bool
  */
 function wpcom_vip_use_core_cron() {
+
+	if ( defined( 'WPCOM_VIP_USE_CORE_CRON' ) && WPCOM_VIP_USE_CORE_CRON ) {
```
mjangda: We've started to drop the `WPCOM_` prefix. Plus, add a disclaimer.

```suggestion
	// For some applications, Cron Control is not a good fit. Allow those apps to use core cron instead. 
	// Note: please check with the VIP team before using this constant as it's not always guaranteed to work.
	if ( defined( 'VIP_USE_CORE_CRON' ) && VIP_USE_CORE_CRON ) {
```

mjangda:
@davidsword Actually, on testing this out, I realized that switching between Cron Control and core cron will result in any one-time scheduled events (like scheduled posts) to be lost, since those are stored in the Cron Control table and not options. That seems like a pretty big blocker for this. I don't have any good answers or workaround either right now but will think on it.

joshbetz:
Most likely will be fixed by https://github.com/Automattic/Cron-Control/pull/188 when it's deployed.

PR #1512, opened 2020-03-20
Don't nerf $wp_object_cache->stats in stop_the_insanity()

joshbetz:
When we nerf the stats object, it causes errors because the object cache dropin assumes the array exists in a certain format. The value of deleting this object is limited because there are only a handful of integers. It won't actually reclaim that much memory.


hanifn:
r839-stacks

PR #1513, opened 2020-03-23
Check type is numeric first

hanifn:
## Description

Checks error type is a numeric value first before checking if the error type is in `error_reporting`.
This is to prevent the error:
`PHP Warning: A non-numeric value encountered in /chroot/var/www/wp-content/mu-plugins/lib/wpcom-error-handler/wpcom-error-handler.php on line 85`

## Checklist

## Steps to Test

TBD

### DIFF
--- a/lib/wpcom-error-handler/wpcom-error-handler.php
+++ b/lib/wpcom-error-handler/wpcom-error-handler.php
@@ -82,7 +82,7 @@ function wpcom_get_error_backtrace( $last_error_file, $last_error_type, $for_irc
  * When we run at shutdown we must not die as then the pretty printing of the Error doesn't happen which is lame sauce.
  */
 function wpcom_custom_error_handler( $whether_i_may_die, $type, $message, $file, $line ) {
-	if ( ! is_numeric( $type ) || ! ( $type & ini_get( 'error_reporting' ) ) ) {
+	if ( ! is_numeric( $type ) || ! ( $type & error_reporting() ) ) {
 		return true;
 	}
 
--- a/lib/wpcom-error-handler/wpcom-error-handler.php
+++ b/lib/wpcom-error-handler/wpcom-error-handler.php
@@ -83,7 +83,7 @@ function wpcom_get_error_backtrace( $last_error_file, $last_error_type, $for_irc
  */
 function wpcom_custom_error_handler( $whether_i_may_die, $type, $message, $file, $line ) {
 	// phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.runtime_configuration_error_reporting
-	if ( ! is_numeric( $type ) || ! ( $type & error_reporting() ) ) { 
+	if ( ! is_int( $type ) || ! ( $type & error_reporting() ) ) { 
 		return true;
 	}
 
--- a/lib/wpcom-error-handler/wpcom-error-handler.php
+++ b/lib/wpcom-error-handler/wpcom-error-handler.php
@@ -82,7 +82,8 @@ function wpcom_get_error_backtrace( $last_error_file, $last_error_type, $for_irc
  * When we run at shutdown we must not die as then the pretty printing of the Error doesn't happen which is lame sauce.
  */
 function wpcom_custom_error_handler( $whether_i_may_die, $type, $message, $file, $line ) {
-	if ( ! is_numeric( $type ) || ! ( $type & error_reporting() ) ) {
+	// phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.runtime_configuration_error_reporting
+	if ( ! is_numeric( $type ) || ! ( $type & error_reporting() ) ) { 
 		return true;
 	}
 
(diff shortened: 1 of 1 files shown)


mjangda:
Do we know how to reproduce this? It would be good to determine that so we can verify how/why we're passing an invalid `$type` into this function in the first place.

hanifn:
@mjangda I was manually calling `wpcom_custom_error_handler()` to test this but you're correct, `$type` shouldn't be a non-numeric value in the first place. My gut is telling me that there's something wrong in `wpcom_error_shutdown()` which is called on exit via `register_shutdown_function()` but I'll need to see the exact error logs to be sure.

nickdaugherty:
This should be reproducible with `trigger_error( 'message', 'string' )` or some other non-numeric second parameter. Somewhere there is code that triggers errors with an invalid error level, which results in this error.

brandon-m-skinner:
> This should be reproducible with `trigger_error( 'message', 'string' )` or some other non-numeric second parameter. Somewhere there is code that triggers errors with an invalid error level, which results in this error.

I couldn't replicate that locally. trigger_error:  https://www.php.net/manual/en/function.trigger-error.php

Even overriding it with my own function via set_error_handler without types didn't let `trigger_error( 'message', 'string' );` work:

```
Warning: trigger_error() expects parameter 2 to be int, string given in /app/wp/wp-content/mu-plugins/000-vip-init.php on line 208
```
Testing the changes, they work well. Still would like to know the origin of the warning though.

hanifn:
> Even overriding it with my own function via set_error_handler without types didn't let trigger_error( 'message', 'string' ); work:

@brandon-m-skinner I think the type is set in `trigger_error()` itself. This further reinforces my idea that this error is occurring on shutdown via `wpcom_error_shutdown()`

hanifn:
Another thing we can do to further safeguard against non-numeric values is to use `error_reporting()` instead of `ini_get( 'error_reporting' )`.

`ini_get()` will always return a string whereas `error_reporting()` will always return an integer. What do you think @mjangda @brandon-m-skinner @nickdaugherty ?

lib/wpcom-error-handler/wpcom-error-handler.php
```diff
@@ -82,7 +82,8 @@ function wpcom_get_error_backtrace( $last_error_file, $last_error_type, $for_irc
  * When we run at shutdown we must not die as then the pretty printing of the Error doesn't happen which is lame sauce.
  */
 function wpcom_custom_error_handler( $whether_i_may_die, $type, $message, $file, $line ) {
-	if ( ! ( $type & ini_get( 'error_reporting' ) ) ) {
+	// phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.runtime_configuration_error_reporting
+	if ( ! is_numeric( $type ) || ! ( $type & error_reporting() ) ) {
```
GaryJones: I think`$type` here will be one of the [PHP Error constants](https://www.php.net/manual/en/errorfunc.constants.php), all of which have an integer value - so would `is_int()` be more accurate here than `is_numeric()`?

brandon-m-skinner:
> > Even overriding it with my own function via set_error_handler without types didn't let trigger_error( 'message', 'string' ); work:
> 
> @brandon-m-skinner I think the type is set in `trigger_error()` itself. This further reinforces my idea that this error is occurring on shutdown via `wpcom_error_shutdown()`

That's definitely the case. The type declarations as seem in the PHP reference stay the same for trigger_error no matter what. At least right now.

lib/wpcom-error-handler/wpcom-error-handler.php
brandon-m-skinner: @GaryJones makes a good point. `is_numeric()` returns true even for numeric strings like "1", "3.55444", etc... whereas `is_int()` only returns true for numbers. See [is_int#Examples](https://www.php.net/manual/en/function.is-int.php#refsect1-function.is-int-examples).
```suggestion
	if ( ! is_int( $type ) || ! ( $type & error_reporting() ) ) { 
```

nickdaugherty:
> Another thing we can do to further safeguard against non-numeric values is to use error_reporting()

That sounds like a good idea.

lib/wpcom-error-handler/wpcom-error-handler.php
hanifn: @brandon-m-skinner @GaryJones I actually went with `is_numeric()` because bitwise operations still work with numeric strings but I agree, being more specific here is probably a better idea

hanifn:
> That sounds like a good idea.

@nickdaugherty  Already added in 031002ce673fa34099b0bd34ce7df1e76390394c

lib/wpcom-error-handler/wpcom-error-handler.php
hanifn: Agreed. Changed in 9edf056b85eb86366691ce32d11a2b7237cf964c

hanifn:
r847-stacks

PR #1515, opened 2020-03-24
Update two-factor to calculate `is_using_two_factor` outside of map_m…

cojennin:
Same thing as #1445, but this time add a check to make sure `Two_Factor_Core` exists before using it (see: #1505)

### DIFF
--- a/two-factor.php
+++ b/two-factor.php
@@ -168,7 +168,7 @@ function wpcom_enable_two_factor_plugin() {
 		return;	
 	}
 
-	require_once( 'two-factor' );
+	require_once( WPMU_PLUGIN_DIR . '/shared-plugins/two-factor/two-factor.php' );
 	add_action( 'set_current_user', 'wpcom_vip_enforce_two_factor_plugin' );
 }
 
--- a/two-factor.php
+++ b/two-factor.php
@@ -141,13 +141,7 @@ function wpcom_vip_enforce_two_factor_plugin() {
 
 		// Calcuate two factor authentication support outside map_meta_cap to avoid callback loop
 		// see: https://github.com/Automattic/vip-go-mu-plugins/pull/1445#issuecomment-592124810
-		$is_user_using_two_factor = false;
-		
-		// We load the two-factor plugin using wpcom_vip_load_plugin but that skips when skip-plugins is set.
-		// This filter still runs even if skip-plugins is set, so add a check to make sure Two_Factor_Core exists
-		if ( class_exists( 'Two_Factor_Core' ) ) {
-			$is_user_using_two_factor = Two_Factor_Core::is_user_using_two_factor();
-		}
+		$is_user_using_two_factor = Two_Factor_Core::is_user_using_two_factor();
 
 		add_filter( 
 			'wpcom_vip_is_user_using_two_factor',
@@ -168,6 +162,8 @@ function wpcom_enable_two_factor_plugin() {
 		return;	
 	}
 
+	// We loaded the two-factor plugin using wpcom_vip_load_plugin but that skips when skip-plugins is set.
+	// Switching to require_once so it no longer gets skipped
 	require_once( WPMU_PLUGIN_DIR . '/shared-plugins/two-factor/two-factor.php' );
 	add_action( 'set_current_user', 'wpcom_vip_enforce_two_factor_plugin' );
 }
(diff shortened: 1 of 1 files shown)


joshbetz:
We should also change the `wpcom_vip_load_plugin` to `require_once`

https://github.com/Automattic/vip-go-mu-plugins/blob/master/two-factor.php#L154

cojennin:
> We should also change the `wpcom_vip_load_plugin` to `require_once`
> 
> https://github.com/Automattic/vip-go-mu-plugins/blob/master/two-factor.php#L154

Updated. Is there a reason not to use `wpcom_vip_load_plugin` here? I can include it as a comment

joshbetz:
The reason this failed last time is because we skip themes in plugins in some WP-CLI actions as a way to ensure code that's out of our control doesn't interfere. `wpcom_vip_load_plugin` honors `--skip-plugins`, but `require_once` does not.

cojennin:
> The reason this failed last time is because we skip themes in plugins in some WP-CLI actions as a way to ensure code that's out of our control doesn't interfere. `wpcom_vip_load_plugin` honors `--skip-plugins`, but `require_once` does not.

Gotcha, with that do we need a `class_exists` check anymore then? Seems like it could be removed

two-factor.php
```diff
@@ -151,7 +168,7 @@ function wpcom_enable_two_factor_plugin() {
 		return;	
 	}
 
-	wpcom_vip_load_plugin( 'two-factor' );
+	require_once( 'two-factor' );
```
mjangda: This will fatal :)

Should probably be:

```suggestion
	require_once( WPMU_PLUGIN_DIR . '/shared-plugins/two-factor/two-factor.php' );
```

two-factor.php
```diff
@@ -139,6 +139,23 @@ function wpcom_vip_enforce_two_factor_plugin() {
...
 
+		// Calcuate two factor authentication support outside map_meta_cap to avoid callback loop
+		// see: https://github.com/Automattic/vip-go-mu-plugins/pull/1445#issuecomment-592124810
+		$is_user_using_two_factor = false;
+		
+		// We load the two-factor plugin using wpcom_vip_load_plugin but that skips when skip-plugins is set.
+		// This filter still runs even if skip-plugins is set, so add a check to make sure Two_Factor_Core exists
+		if ( class_exists( 'Two_Factor_Core' ) ) {
+			$is_user_using_two_factor = Two_Factor_Core::is_user_using_two_factor();
+		}
```
mjangda: Don't need the class_exists as noted in another comment:

```suggestion
		$is_user_using_two_factor = Two_Factor_Core::is_user_using_two_factor();
```

Especially since if it doesn't fatal here, it will fatal further down the line.

cojennin: Ah yea, figured it was only needed there cause it hadn't been fataling where it was before. Updated

two-factor.php
```diff
@@ -139,6 +139,17 @@ function wpcom_vip_enforce_two_factor_plugin() {
 			return $limited;
 		}, 9 );
 
+		// Calcuate two factor authentication support outside map_meta_cap to avoid callback loop
+		// see: https://github.com/Automattic/vip-go-mu-plugins/pull/1445#issuecomment-592124810
+		$is_user_using_two_factor = Two_Factor_Core::is_user_using_two_factor();
```
GaryJones: If someone tries to stop 2FA, by doing `remove_action( 'muplugins_loaded', 'wpcom_enable_two_factor_plugin' );`, then this `Two_Factor_Core::...` call is likely to cause a fatal error. It may be worth being defensive and doing a `class_exists()` check anyway.

mjangda: That would be a `_doing_it_wrong`. Plus, there are other pathways in this codebase that would likely also trigger fatals.

cojennin: If that's the case, probably can do away with [this `class_exists` check](https://github.com/Automattic/vip-go-mu-plugins/blob/b8baa5ec0a4bc18aa58990b24eaf74c0aab4735d/two-factor.php#L31) as well, that's what was keeping the previous placement of `Two_Factor_Core` from throwing a fatal

mjangda:
r842-stacks

mjangda:
Verified that `wp --skip-plugins` was not impacted this time around.

PR #1516, opened 2020-03-24
Add facet query filters to search

brandon-m-skinner:
## Description

Disable facet queries in VIP Search but still set the default facet taxonomies size just in case it gets enabled at some point so it doesn't error out.

## Checklist

## Steps to Test
1. Do a search with VIP Search enabled.
2. Check the Query body in QM/debug bar, there will be multiple size attributes.
3. Check out PR.
4. Do a search.
5. There should now only be one size attribute.

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -91,6 +91,13 @@ class Search {
 
 		// Allow querying while a bulk index is running
 		add_filter( 'ep_enable_query_integration_during_indexing', '__return_true' );
+
+		// Set facet taxonomies size. Shouldn't currently be used, but it makes sense to have it set to a sensible
+		// default just in case it ends up in use so that the application doesn't error
+		add_filter( 'ep_facet_taxonomies_size', array( $this, 'filter__ep_facet_taxonomies_size' ), PHP_INT_MAX );
+
+		// Disable facet queries
+		add_filter( 'ep_facet_include_taxonomies', '__return_empty_array' );
 	}
 
 	protected function load_commands() {
@@ -302,5 +309,12 @@ class Search {
 			}
 		}
 		return $response;
+	}
+
+	/*
+	 * Given the current facet taxonomies size and a taxonomy, determine the facet taxonomy size
+	 */
+	public function filter__ep_facet_taxonomies_size( $size, $taxonomy ) {
+		return 5;
 	}
 }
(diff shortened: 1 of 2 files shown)


search/includes/classes/class-search.php
```diff
@@ -91,6 +91,13 @@ protected function setup_hooks() {
 
 		// Allow querying while a bulk index is running
 		add_filter( 'ep_enable_query_integration_during_indexing', '__return_true' );
+
+		// Set facet taxonomies size. Shouldn't currently be used, but it makes sense to have it set to a sensible
+		// default just in case it ends up in use so that the application doesn't error
+		add_filter( 'ep_facet_taxonomies_size', array( $this, 'filter__ep_facet_taxonomies_size' ), PHP_INT_MAX );
```
hanifn: Hmmm.. strange. I can't find any reference to `ep_facet_taxonomies_size` in their code or in their docs. Are you sure this is in use?

hanifn: NVM. Just saw your PR on our ElasticPress fork

nickdaugherty: We don't have to set a priority on this filter - it's ok if it's overridden, it just can't be 10k by default.

search/includes/classes/class-search.php
```diff
@@ -91,6 +91,13 @@ protected function setup_hooks() {
...
+
+		// Disable facet queries
+		add_filter( 'ep_facet_include_taxonomies', '__return_empty_array' );
```
nickdaugherty: This is probably ok - I was originally thinking that the default taxonomy would be `category`...but that does put us back to running the aggregations for every request, even when they aren't used (no widget displaying)...so probably better to let sites define this as-needed, until EP is modified to only request aggregations when used.

Also this will only run when the Facets feature is enabled. Do we know if regularly offloaded queries (non-search) also run the aggregations? Or is it just search?

brandon-m-skinner: @nickdaugherty

Facets seem to be enabled by default. 

Not sure about query offloading. With this PR enabled, I don't get any ElasticPress debug panel stuff on anything but searches when going through wpvip.com or locally. So either this breaks regular offloading, regular offloading isn't included in the debug bar, or offloading isn't enabled on these sites.

brandon-m-skinner: You do get the occasional calls to /_plugins while navigating around instead of /_search though.

nickdaugherty: Alright, let's go with this for now, and can tweak as needed.

brandon-m-skinner:
r843-stacks

PR #1517, opened 2020-03-25
Remove test from the Site Health tool page

pandah3:
## Description

The Site Health tool page is showing some critical/failed issues for tests that aren't relevant to VIP:
site health

This PR removes the unnecessary `background_updates` test.

Here's a list of all tests run for the Site Health tool in WP core:
https://core.trac.wordpress.org/browser/tags/5.3/src/wp-admin/includes/class-wp-site-health.php#L1842

## Checklist

## Steps to Test

1. Check out PR.
1. Go to `wp-admin` > `Tools` > `Site Health`
1. Verify `Background updates` error/issue isn't present:
site health after

### DIFF
--- a/000-debug/0-load.php
+++ b/000-debug/0-load.php
@@ -4,9 +4,3 @@ namespace Automattic\VIP\Debug;
 
 require_once( __DIR__ . '/logger.php' );
 require_once( __DIR__ . '/debug-mode.php' );
-
-function remove_background_updates_test( $tests ) {
- unset( $tests['async']['background_updates'] );
- return $tests;
-}
-add_filter( 'site_status_tests', 'remove_background_updates_test' );
--- a/000-debug/0-load.php
+++ b/000-debug/0-load.php
@@ -4,3 +4,9 @@ namespace Automattic\VIP\Debug;
 
 require_once( __DIR__ . '/logger.php' );
 require_once( __DIR__ . '/debug-mode.php' );
+
+function remove_background_updates_test( $tests ) {
+    unset( $tests['async']['background_updates'] );
+    return $tests;
+}
+add_filter( 'site_status_tests', 'remove_background_updates_test' );
--- a/000-debug/0-load.php
+++ b/000-debug/0-load.php
@@ -6,7 +6,7 @@ require_once( __DIR__ . '/logger.php' );
 require_once( __DIR__ . '/debug-mode.php' );
 
 function remove_background_updates_test( $tests ) {
-    unset( $tests['async']['background_updates'] );
-    return $tests;
+ unset( $tests['async']['background_updates'] );
+ return $tests;
 }
 add_filter( 'site_status_tests', 'remove_background_updates_test' );
--- a/001-core.php
+++ b/001-core.php
@@ -41,7 +41,7 @@ function wpcom_vip_disable_core_update_cap( $caps, $cap ) {
 }
 
 /**
-	* Diable tests in the Site Health (AKA site status) tool page
+	* Disable tests in the Site Health (AKA site status) tool page
 	*
 	* By default, WordPress runs a series of tests on the Site Health tool
 	* page in wp-admin. This disables all irrelevant or unnecessary tests.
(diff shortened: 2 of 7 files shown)


000-debug/0-load.php
```diff
@@ -4,3 +4,9 @@
 
 require_once( __DIR__ . '/logger.php' );
 require_once( __DIR__ . '/debug-mode.php' );
+
+function remove_background_updates_test( $tests ) {
+ unset( $tests['async']['background_updates'] );
+ return $tests;
+}
+add_filter( 'site_status_tests', 'remove_background_updates_test' );
```
mjangda: Let's move this to `001-core.php` and add a comment explaining why we're disabling the test.

000-debug/0-load.php
```diff
@@ -4,3 +4,9 @@
 
 require_once( __DIR__ . '/logger.php' );
 require_once( __DIR__ . '/debug-mode.php' );
+
+function remove_background_updates_test( $tests ) {
+ unset( $tests['async']['background_updates'] );
+ return $tests;
```
mjangda: Tabs over spaces :)

```suggestion
	unset( $tests['async']['background_updates'] );
	return $tests;
```

000-debug/0-load.php
pandah3: Moved and added! I guess it was the where the code was located (`000-debug/0-load.php`) that was returning the incorrect success message, not a bug in the code:
site health tests

001-core.php
```diff
@@ -39,3 +39,17 @@ function wpcom_vip_disable_core_update_cap( $caps, $cap ) {
...
+/**
+ * Diable tests in the Site Health (AKA site status) tool page
+ *
+ * By default, WordPress runs a series of tests on the Site Health tool
+	* page in wp-admin. Some tests, such as the Background Updates test, are
+	* irrelevant because the updates are managed by the VIP team.
+	* This disables all unnecessary tests.
+ */
+function remove_background_updates_test( $tests ) {
+	unset( $tests['async']['background_updates'] );
```
mjangda: A few suggested changes here:

- Fix the indents in the docblock
- Prefix the function name and make it more generic (since we may remove other tests too?)
- Add `isset` checks to be a bit defensive.
- Move the comment about Background Updates inline to make it more contextual (this way, if we change the the behavior, we don't need to remember to update the docblock too).

```suggestion
 * By default, WordPress runs a series of tests on the Site Health tool
 * page in wp-admin. This disables all irrelevant or unnecessary tests.
 */
function vip_disable_unnecessary_site_health_tests( $tests ) {
	// Disable "Background Updates" test.
	// WordPress updates are managed by the VIP team.
	if ( isset( $tests['async'] ) && isset( $tests['async']['background_updates'] ) ) {
		unset( $tests['async']['background_updates'] );
	}
```

001-core.php
```diff
@@ -39,3 +39,17 @@ function wpcom_vip_disable_core_update_cap( $caps, $cap ) {
...
+	return $tests;
+}
+add_filter( 'site_status_tests', 'remove_background_updates_test' );
```
mjangda: Update to use prefixed and generic function name:

```suggestion
add_filter( 'site_status_tests', 'vip_disable_unnecessary_site_health_tests' );
```

000-debug/0-load.php
pandah3: Argh, this always gives me issues on this repo. I turned off auto-indent and manually tabbed myself, but it still gives me issues! My PHPCS never works either -.-. I just committed your suggestion and it seemed to work!

001-core.php
pandah3: Ah got it! I thought we were having each test removal be its own function (for separation purposes) but I guess that could get quite messy!

Fixed + added all the points 👍

001-core.php
pandah3: Changed 👍

hanifn:
r844-stacks

PR #1519, opened 2020-03-26
Disable Jetpack Search offload when ElasticPress is being used

brandon-m-skinner:
## Description

#1471 

> When ElasticPress query integration is enabled, we want to ensure queries are no longer offloaded to Jetpack Search.
> 
> Needs testing
>
>\- @joshbetz


brandon-m-skinner:
r846-stacks

PR #1523, opened 2020-03-27
Add track_total hits as true to formatted args

brandon-m-skinner:
## Description
Add track_total hits as true to formatted args.

Only when the value isn't already set.

## Checklist

## Steps to Test
1. Do a search with VIP Search enabled. The query body won't have 'track_total_hits' set.
2. Checkout the PR.
3. Do a search with VIP Search enabled. The query body will have 'track_total_hits' set to true.

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -218,8 +218,16 @@ class Search {
 
 		$query_integration_enabled = defined( 'VIP_ENABLE_VIP_SEARCH_QUERY_INTEGRATION' ) && true === VIP_ENABLE_VIP_SEARCH_QUERY_INTEGRATION;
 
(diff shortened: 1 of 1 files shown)


search/includes/classes/class-search.php
```diff
@@ -314,4 +317,17 @@ public function filter__jetpack_widgets_to_include( $widgets ) {
 		// Flatten the array back down now that may have removed values from the middle (to keep indexes correct)
 		return array_values( $widgets );
 	}
+
+	/*
+	 * Add track track_total_hits to formatted_args
```
nickdaugherty: Minor / non-blocking - I know we'll end up using this filter for other stuff, so maybe we can make the comment more generic to start with?

PR #1531, opened 2020-03-31
Remove fuzziness from EP search queries

brandon-m-skinner:
## Description

Fuzziness is enabled by default for queries. This results in some pretty unexpected search results since something like 'March' would match 'arch' and anything else within a Levenshtein distance of 1. Not great for search relevancy or performance.

## Checklist

## Steps to Test

1. Perform a search with VIP Search and query monitory/debug bar enabled.
2. Note that fuzziness is 1 in places.
3. Apply PR.
4. Perform another search.
5. Fuzziness is 0 across the board.

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -98,8 +98,8 @@ class Search {
 		// Enable track_total_hits for all queries for proper result sets if track_total_hits isn't already set
 		add_filter( 'ep_post_formatted_args', array( $this, 'filter__ep_post_formatted_args' ), 10, 3 );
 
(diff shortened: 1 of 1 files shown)


nickdaugherty:
This will also disable fuzziness for the "Autosuggest" feature. We haven't really looked at that too much, but it sets a fuzziness of 'auto' when enabled via the same filter (priority 10).

That appears to be global when that Feature is enabled, which seems not great, so I wonder if it's better to disable Autosuggest entirely for now.

brandon-m-skinner:
@nickdaugherty Filter change applied.

PR #1540, opened 2020-04-03
Disable search index validation until index is built

nickdaugherty:
## Description

Only run index validation when it's expected to be consistent - so not when an index is being built, or no index yet exists.

Depends on https://github.com/Automattic/ElasticPress/pull/26

## Checklist

## Steps to Test

Tricky to actually test b/c healthchecks don't run outside production, but can try this:

1. Check out PR.
1. `define( 'VIP_GO_ENV', 'production' );`
1. Start a bulk index - `wp elasticpress index`
1. In a separate `wp shell`, try `\Automattic\VIP\Search\HealthJob::is_enabled()`

### DIFF
--- a/tests/search/includes/classes/test-class-healthjob.php
+++ b/tests/search/includes/classes/test-class-healthjob.php
@@ -44,6 +44,8 @@ class HealthJob_Test extends \WP_UnitTestCase {
 			->method( 'process_results' );
 
 		$job->check_health();
+
+		remove_filter( 'enable_vip_search_healthchecks', '__return_true' );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


tests/search/includes/classes/test-class-healthjob.php
```diff
@@ -153,4 +155,64 @@ public function test__vip_search_healthjob_process_results_with_wp_error() {
...
+		$this->assertTrue( $enabled );
+
+		remove_filter( 'ep_is_indexing', '__return_false' );
+		remove_filter( 'ep_last_sync', '__return_true' );
+		remove_filter( 'vip_search_healthchecks_enabled_environments', $enabled_environments );
+	}
+
+	/**
+	 * NOTE - needs to come last for the "is_enabled()" tests, b/c it sets the constant
+	 * which causes it to bail early
```
mjangda, line 207: This is bound to trip us up in the future. I'd suggest running this job in a separate process to avoid that.

```suggestion
	 * which causes it to bail early
	 *
	 * @runInSeparateProcess
	 * @preserveGlobalState disabled
	 *
```

nickdaugherty: Oh maybe it's the `preserveGlobalState disabled` that I was missing, b/c when I tried to run this in a separate process I was getting errors about serializing closures.

nickdaugherty: Switched in 783b48aa

nickdaugherty:
r866-stacks

PR #1549, opened 2020-04-09
Fixes an issue with `rename()` breaking files.

emrikol:
Previously, the temporary file name was used for the file contents for the _new_ file.  This will pull the actual file contents and use it instead.

## Description

Currently, when a file goes through PHP's `rename()` the stream wrapper will download a copy, and try to upload a new copy with the new name.  Unfortunately, instead of copying the file it just uploads a new file with a `tempnam()` as its contents.

## Checklist

## Steps to Test

1. Grab an uploaded URL from a site.
1. `wp shell` in
1. Do a `rename()` on the file, and then `curl` the _new_ file to see its contents is broken
1. Check out PR and merge the fixes, etc...
1. Do a _new_ `rename()` and `curl` the _new_ file again to see that it's now correct.

### DIFF
--- a/files/class-vip-filesystem-stream-wrapper.php
+++ b/files/class-vip-filesystem-stream-wrapper.php
@@ -543,7 +543,7 @@ class VIP_Filesystem_Stream_Wrapper {
 			}
 
 			// Convert to actual file to upload to new path
-			$file     = $this->string_to_resource( $result, 'r' );
+			$file     = $this->string_to_resource( file_get_contents( $result ), 'r' );
 			$meta     = stream_get_meta_data( $file );
 			$filePath = $meta['uri'];
 
--- a/files/class-vip-filesystem-stream-wrapper.php
+++ b/files/class-vip-filesystem-stream-wrapper.php
@@ -543,7 +543,7 @@ class VIP_Filesystem_Stream_Wrapper {
 			}
 
 			// Convert to actual file to upload to new path
-			$file     = $this->string_to_resource( file_get_contents( $result ), 'r' );
+			$file     = fopen( $result, 'r' );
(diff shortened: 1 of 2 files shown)


files/class-vip-filesystem-stream-wrapper.php
```diff
@@ -543,7 +543,7 @@ public function rename( $path_from, $path_to ) {
 			}
 
 			// Convert to actual file to upload to new path
-			$file     = $this->string_to_resource( $result, 'r' );
+			$file     = $this->string_to_resource( file_get_contents( $result ), 'r' );
```
mjangda, line 546: `Api_Client:: get_file` returns a `tempnam` path, so we need to `fopen` it:

```suggestion
			$file     = fopen( $result, 'r' );
```

Regression from #1397

parkcityj:
@mjangda - I tested updated fix on a sandbox, and it works as expected.

mjangda:
Introduced in #1397 

I've added a unit test that would have caught this issue (props @nickdaugherty for the nudge to start adding them).

Without this PR, we fail:

```
There was 1 error:

1) Automattic\VIP\Files\VIP_Filesystem_Stream_Wrapper_Test::test__rename__sucess
rename/delete_file/from failed for wp-content/uploads/old.txt with error: Expectation failed for method name is equal to "upload_file" when invoked 1 time(s)
Parameter 0 for invocation Automattic\VIP\Files\API_Client::upload_file('/tmp/phpBJMnhE', 'wp-content/uploads/new.txt') does not match expected value.
Failed asserting that two strings are equal. #vip-go-streams
```

mjangda:
r878-stacks

nickdaugherty:
Nice work on the tests @mjangda !

PR #1551, opened 2020-04-09
Offload bulk indexing to queue

nickdaugherty:
## Description

Adds cron jobs to process the index queue and offloads bulk operations to the queue

## Checklist

## Steps to Test

1. Enable async indexing with `define( 'VIP_SEARCH_ENABLE_ASYNC_INDEXING', true );`
1. Edit a category that is attached to multiple posts
1. Look in the `wp_vip_search_index_queue` table
1. Note 1 job per post from the category, marked as `scheduled`
1. Ensure WP cron is running
1. Once the scheduled cron event runs, the posts will be removed from the queue table (b/c they have been processed)


search/includes/classes/Queue/class-cron.php
```diff
@@ -0,0 +1,152 @@
...
+		// Find jobs to process
+		$jobs = $this->queue->get_batch_jobs( self::CRON_PROCESSOR_MAX_OBJECTS_PER_JOB );
+
+		if ( ! count( $jobs ) ) {
+			return;
+		}
+
+		$job_ids = wp_list_pluck( $jobs, 'id' );
+
+		wp_schedule_single_event( time(), self::CRON_PROCESSOR_EVENT_NAME, $job_ids );
```
mjangda, line 141: What's the thinking behind scheduling separate jobs rather than just doing the processing here?

If there's a queue backlog, we'll end up delaying the indexing. Possibly not an issue if we make sure to allow the jobs to run concurrently. But worth considering.

Also, it creates a bit of a disconnect in the states:

1. add to queue => `status = queue`
1. create job => `status = running`
1. run job => `status = running`

Between 2-3 there may be a bit of a wait but we're claiming the queued item is `running` but it's actually not. It's not a huge deal but again debugging becomes challenging when the states aren't an accurate representation.

Aside: we may also need a a process to catch items that are stuck in `running` for a long time for whatever reason.

search/includes/classes/Queue/class-cron.php
```diff
@@ -0,0 +1,152 @@
...
+	 * Process a batch of jobs via cron
+	 * 
+	 * This is the cron hook for indexing a batch of objects
+	 * 
+	 * @param {array} $job_ids Array of job ids to process
+	 */
+	public function process_jobs( $job_ids ) {
+		$jobs = $this->queue->get_jobs( $job_ids );
+
+		if ( ! count( $jobs ) ) {
```
mjangda, line 112:
```suggestion
		if ( empty( $jobs ) ) {
```

search/includes/classes/Queue/class-cron.php
nickdaugherty: Scheduling separate jobs lets us run multiple workers in parallel for more throughput, where if we tied it into this scheduled task, we'd be running "single threaded" on a 1 minute cadence.

The other alternative I've considered is to schedule the jobs themselves immediately, when objects are added to the queue. This has a few downsides:

* Tightly couples processing of the queue with inserting of jobs
* Not as fault tolerant
* Won't have an even distribution of jobs to workers - one cron event could have a single job, the next could have 1000, which could be inefficient

However to reduce latency as far as possible, I think we'll need to schedule the jobs immediately, and then have this other cron job running on a slower interval, looking for anything that was missed (in case there was a fatal or something).

nickdaugherty: > Also, it creates a bit of a disconnect in the states

Good eye :) It certainly does...this was the simplest way to implement locking so that the same jobs don't get processed by multiple workers...we could use an intermediate state, like `scheduled` or similar...the downside is increased write traffic to the DB (which must be replicated). Probably not _too_ big of a deal there, but could get expensive if there are tens of thousands of posts queued up and we're transitioning all the rows from `queued -> scheduled -> running` then deleting them.

Alternatively we could add a 1 byte int column to flag if the row has been scheduled as a cron job or not...that would cut down on data on the wire while not confusing the state.

nickdaugherty: > Aside: we may also need a a process to catch items that are stuck in running for a long time for whatever reason.

Definitely, it's on the list for the next round of improvements.

nickdaugherty: I've switched over to scheduling the cron job immediately (instead of via a secondary cron job). There's still a "sweeper" cron job that runs every 5 minutes, looking for anything that hasn't been scheduled to a cron event (but should be), and in the future, this will handle any deadlock situations.

I also added the intermediate "scheduled" state to represent when objects have been assigned to a cron job, but not yet processed.

nickdaugherty:
Have some failing tests that I'll look at in the morning (works locally, but they're kinda flaky b/c they rely on timestamps).

nickdaugherty:
r891-stacks

PR #1557, opened 2020-04-16
Uploads: Add length check for file name/path

mjangda:
To provide clearer errors when the full path (`/wp-content/uploads/path/to/.../file.png`) is too long.

Runs the checks early in the process (`wp_handle_upload_prefilter`) since the `wp_check_filetype_and_ext` hook is late and doesn't allow the ability to customize errors returned to the user.

See PLAT-1162

## TODO

- [x] Remove tests for removed functions
- [x] Add new tests
- [ ] Check if we need to run through unique filename endpoint?

## Checklist

## Steps to Test

1. Create a file with a name that's 231 characters or more.
1. Try to upload.
1. It should fail.
1. Upload a file with a shorter name; it should upload fine.


hanifn:
Regarding using the unique filename endpoint on the file service, I'm really not sure either way. I guess the question is: what advantage does using the endpoint give us?

One point I can think of is that we can automatically support future file type support.

files/class-vip-filesystem.php
```diff
@@ -154,33 +161,45 @@ public function filter_upload_dir( $params ) {
...
-		$filename = sanitize_file_name( $filename );
-
-		// Setting `ext` and `type` to empty will fail the upload because Go doesn't allow unfiltered uploads
-		// See `_wp_handle_upload()`
-		if ( ! $this->check_filetype_with_backend( $filename ) ) {
-			$filetype_data['ext']             = '';
-			$filetype_data['type']            = '';
-			// Never set this true, which leaves filename changing to dedicated methods in this class
-			$filetype_data['proper_filename'] = false;
+	protected function validate_file_length( $file_path ) {
```
chrean, line 197: Minor thought: the name of this function may be a bit misleading because we are not checking the length of the file, but the length of its name instead.

files/class-vip-filesystem.php
```diff
@@ -154,33 +161,45 @@ public function filter_upload_dir( $params ) {
...
+		// TODO: run through unique filename?
+
+		$check_type = $this->validate_file_type( $file_path );
+		if ( is_wp_error( $check_type ) ) {
+			$file['error'] = $check_type->get_error_message();
+
+			return $file;
+		}
+
+		$check_length = $this->validate_file_length( $file_path );
```
brandon-m-skinner, line 182:
```suggestion
		$check_length = $this->validate_file_path_length( $file_path );
```

files/class-vip-filesystem.php
```diff
@@ -154,33 +161,45 @@ public function filter_upload_dir( $params ) {
...
-
-		// Setting `ext` and `type` to empty will fail the upload because Go doesn't allow unfiltered uploads
-		// See `_wp_handle_upload()`
-		if ( ! $this->check_filetype_with_backend( $filename ) ) {
-			$filetype_data['ext']             = '';
-			$filetype_data['type']            = '';
-			// Never set this true, which leaves filename changing to dedicated methods in this class
-			$filetype_data['proper_filename'] = false;
+	protected function validate_file_length( $file_path ) {
+		if ( mb_strlen( $file_path ) > self::MAX_FILE_PATH_LENGTH ) {
```
brandon-m-skinner, line 198:
```suggestion
	protected function validate_file_path_length( $file_path ) {
		if ( mb_strlen( $file_path ) > self::MAX_FILE_PATH_LENGTH ) {
```

brandon-m-skinner:
@mjangda Yay or nay on the naming change? I'll deploy tomorrow as-is regardless since it's non-blocking.

mjangda:
New name makes sense 👍 Thanks @chrean and @brandon-m-skinner !

hanifn:
r900-stacks

PR #1559, opened 2020-04-16
Offload queries to database if RPS too high in VIP Search

brandon-m-skinner:
## Description

Have an object cache value that expires in a reasonable time frame(currently set to 5 minutes).

Increment the value for each request that passes through VIP Search.

If this value gets higher than the define limits, disable query integration for roughly half of the requests.

## Checklist

## Steps to Test

1. Checkout PR
2. Change `private static $max_query_count = 3000 + 1;` to `private static $max_query_count = 1;`
3. Run a search with QM/Debug Bar open for ElasticPress.
4. Keep refreshing the page while watching the queries.
5. Roughly half shouldn't go through ElasticPress.
6. Reset `private static $max_query_count` back to normal.

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -634,7 +634,7 @@ class Search {
 	 * Increment the number of queries that have been passed through VIP Search
 	 */
 	private static function query_count_incr() {
-		if ( ! wp_cache_get( self::$query_count_cache_key ) ) {
+		if ( false === wp_cache_get( self::$query_count_cache_key ) ) {
 			wp_cache_set( self::$query_count_cache_key, 0, 'default', self::$query_count_ttl );
 		}
 
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -636,7 +636,7 @@ class Search {
 	 */
 	private static function query_count_incr() {
 		if ( false === wp_cache_get( self::$query_count_cache_key ) ) {
-			wp_cache_set( self::$query_count_cache_key, 0, 'default', self::$query_count_ttl );
+			wp_cache_set( self::$query_count_cache_key, 0, self::$query_count_cache_group, self::$query_count_ttl );
 		}
 
 		return wp_cache_incr( self::$query_count_cache_key );
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -9,6 +9,11 @@ class Search {
 	public $queue;
 	private $current_host_index;
 
+	private static $query_count_cache_key = 'vip_search_query_count';
+	private static $query_count_ttl = 300; // 5 minutes in seconds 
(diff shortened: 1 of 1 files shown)


search/includes/classes/class-search.php
```diff
@@ -615,4 +629,15 @@ public function filter__epwr_score_mode( $score_mode, $formatted_args, $args ) {
 	public function filter__epwr_boost_mode( $boost_mode, $formatted_args, $args ) {
 		return 'multiply';
 	}
+
+	/*
+	 * Increment the number of queries that have been passed through VIP Search
+	 */
+	private static function query_count_incr() {
+		if ( ! wp_cache_get( self::$query_count_cache_key ) ) {
```
mjangda, line 637: Explicit not-exists check:

```suggestion
		if ( false === wp_cache_get( self::$query_count_cache_key ) ) {
```

Otherwise, if for some reason the value is already `0` we end up calling an unnecessary `set`.

search/includes/classes/class-search.php
```diff
@@ -9,6 +9,11 @@ class Search {
 	public $queue;
 	private $current_host_index;
 
+	private static $query_count_cache_key = 'vip_search_query_count';
```
mjangda, line 12:
```suggestion
	private static $query_count_cache_key = 'query_count';
	private static $query_count_cache_group = 'vip_search';
```

search/includes/classes/class-search.php
```diff
@@ -9,6 +9,11 @@ class Search {
...
+	private static $query_count_ttl = 300; // 5 minutes in seconds 
+	private static $max_query_count = 3000 + 1; // 10 requests per second plus one for cleanliness of comparing with Search::query_count_incr
+	private static $query_rand_comparison = 5; // Value to compare >= against rand( 1, 10 ). 5 should result in roughly half being true.
```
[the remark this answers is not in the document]
brandon-m-skinner: Yes, since they're just settings.

search/includes/classes/class-search.php
```diff
@@ -615,4 +630,15 @@ public function filter__epwr_score_mode( $score_mode, $formatted_args, $args ) {
 	public function filter__epwr_boost_mode( $boost_mode, $formatted_args, $args ) {
 		return 'multiply';
 	}
+
+	/*
+	 * Increment the number of queries that have been passed through VIP Search
+	 */
+	private static function query_count_incr() {
+		if ( false === wp_cache_get( self::QUERY_COUNT_CACHE_KEY, self::QUERY_COUNT_CACHE_GROUP ) ) {
```
nickdaugherty, line 638: Does `wp_cache_incr()` fail if it doesn't yet exist?

nickdaugherty: Oh we need to do this to set the TTL, got it.

brandon-m-skinner: Yep.

brandon-m-skinner:
r892-stacks

PR #1560, opened 2020-04-17
Added tests for ep_skip_query_integration and edited ep_skip_query_integration from tests/feedback

brandon-m-skinner:
## Description

We were missing some important tests for query integration. So I added them.

## Checklist

## Steps to Test

1. Check out PR.

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -100,6 +100,7 @@ class Search {
 		// Disable query integration by default
 		add_filter( 'ep_skip_query_integration', array( __CLASS__, 'ep_skip_query_integration' ), 5, 2 );
 		add_filter( 'ep_skip_user_query_integration', array( __CLASS__, 'ep_skip_query_integration' ), 5 );
+		add_filter( 'ep_skip_query_integration', array( __CLASS__, 'rate_limit_ep_query_integration' ), PHP_INT_MAX );
 
 		// Disable certain EP Features
 		add_filter( 'ep_feature_active', array( $this, 'filter__ep_feature_active' ), PHP_INT_MAX, 3 );
@@ -350,15 +351,6 @@ class Search {
 		if ( $skip ) {
 			return true;
 		}
-
-		// If the query count has exceeded the maximum
-		//     Only allow half of the queries to use VIP Search
-		if ( self::query_count_incr() > self::$max_query_count ) {
-			// Should be roughly half over time
-			if ( self::$query_db_fallback_value >= rand( 1, 10 ) ) {
-				return true;
-			}
-		}
 		
 		if ( isset( $_GET['es'] ) ) {
 			return false;
@@ -395,6 +387,21 @@ class Search {
 		$skipped = ! ( $enabled_by_constant || $enabled_by_option );
 	
 		return $skipped;
+	}
+
+	public static function rate_limit_ep_query_integration( $skip ) {
+		if ( $skip ) {
+			return true;
+		}
+
+		// If the query count has exceeded the maximum
+		//     Only allow half of the queries to use VIP Search
+		if ( self::query_count_incr() > self::$max_query_count ) {
+			// Should be roughly half over time
+			if ( self::$query_db_fallback_value >= rand( 1, 10 ) ) {
+				return true;
+			}
+		}
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


tests/search/test-class-search.php
```diff
@@ -862,4 +862,61 @@ public function test__vip_search_filter__ep_allow_post_content_filtered_index()
 
 		$this->assertFalse( $enabled );
 	}
+
+	/*
+	 * Ensure the various options, constants, and rate limiting affect query integration in the right ways
+	 */
+	public function test__ep_skip_query_integration() {
```
mjangda, line 869: Minor: Should we split this off into individual tests? I personally find large tests like this hard to follow and they also tend to be fragile (if we change one aspect of how the query integration stuff works, the test may need to ba completely re-written).

tests/search/test-class-search.php
```diff
@@ -862,4 +862,61 @@ public function test__vip_search_filter__ep_allow_post_content_filtered_index()
...
+	}
+
+	/**
+	 * Ensure the incrementor for tracking request counts behaves properly
+	 */
+	public function test__query_count_incr() {
+		$es = new \Automattic\VIP\Search\Search();
+
+		// Reset cache key
+		wp_cache_delete( $es::QUERY_COUNT_CACHE_KEY, $es::QUERY_COUNT_CACHE_GROUP );
```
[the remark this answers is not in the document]
brandon-m-skinner: Probably not right now, but it's the safe bet. Nothing is worse than having tests you didn't touch fail when you unintentionally cause side effects.

search/includes/classes/class-search.php
```diff
@@ -664,7 +665,7 @@ public function filter__epwr_boost_mode( $boost_mode, $formatted_args, $args ) {
 	/*
 	 * Increment the number of queries that have been passed through VIP Search
 	 */
-	private static function query_count_incr() {
+	public static function query_count_incr() {
```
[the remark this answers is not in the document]
brandon-m-skinner: Interesting, I'll give that a shot.

search/includes/classes/class-search.php
```diff
@@ -339,6 +339,15 @@ public function filter__ep_feature_active( $active, $feature_settings, $feature
 	 * constant should be set to `true`, which will enable query integration for all requests
 	 */
 	public static function ep_skip_query_integration( $skip, $query = null ) {
+		// If the query count has exceeded the maximum
+		//     Only allow half of the queries to use VIP Search
+		if ( self::query_count_incr() > self::$max_query_count ) {
```
mjangda, line 344: One thought I had: should we only trigger this when we determine `$skip === false`? In contexts where we return `true`, and ES doesn't kick in, we're doing unnecessary memcache operations and incorrectly incrementing rate limit counts.

brandon-m-skinner: True, the skip check should be at the top 👍

nickdaugherty: I think what Mo was getting at is it would be better to run all the logic in this function first (remove the early return on the `$_GET['es']` check), and _then_ only rate limit if `$skip === false`.

Currently, we check the counter even if the rest of the filter ultimately returns `true` and skips the integration (if the option or constant is not set).

nickdaugherty: We could add the rate limit as a separate late running (PHP_INT_MAX) filter on `es_skip_query_integration` to solve those concerns - let this filter callback run normally, then our rate limit filter callback runs at the very end, disabling the query integration if we're rate limited (and query integration had been enabled via other filters).

mjangda: > We could add the rate limit as a separate late running (PHP_INT_MAX) filter on es_skip_query_integration to solve those concerns

Ah, that a great idea. Much cleaner and easier to disable / detangle if we ever need to.

brandon-m-skinner: Sounds good to me :+1:

search/includes/classes/class-search.php
```diff
@@ -395,6 +390,34 @@ public static function ep_skip_query_integration( $skip, $query = null ) {
...
+	 * @return bool new value of ep_skip_query_integration
+	 */
+	public static function rate_limit_ep_query_integration( $skip ) {
+		// Honor previous filters that skip query integration
+		if ( $skip ) {
+			return true;
+		}
+
+		// If the query count has exceeded the maximum
+		//     Only allow half of the queries to use VIP Search
```
nickdaugherty, line 410:
```suggestion
		// Only allow half of the queries to use VIP Search
```

brandon-m-skinner:
r913-stacks

Issue #1565, opened 2020-04-21
Fix wpcom_vip_get_post_pageviews() logical error


WPprodigy:
There is a bug here: https://github.com/Automattic/vip-go-mu-plugins/blob/da7fcd24b773e1957687c6eb56f0b56ddd402d9e/vip-helpers/vip-stats.php#L73-L102

The `stats_get_csv()` function returns like this:

```
stats_get_csv( 'postviews', [ 'post_id'=> 123 , 'num_days' => 90 ] );
> array(1) { [0] => array(1) { 'views' => "10123" } }
```

Returning `absint( $posts[0] )` will always just be `1` I suppose. We need to go a level deeper to actually return the view count.

PR #1566, opened 2020-04-21
Return the correct view count from wpcom_vip_get_post_pageviews()

WPprodigy:
Fixes https://github.com/Automattic/vip-go-mu-plugins/issues/1565

More info in the issue, but essentially `stats_get_csv()` returns like this:

```
stats_get_csv( 'postviews', [ 'post_id'=> 123 , 'num_days' => 90 ] );
> array(1) { [0] => array(1) { 'views' => "10123" } }
```

To get the view number, it needs to go a level deeper. I also tested with invalid IDs and the logic continues to work (views is still set as a key, just to NULL). The function calls `absint()` at the end as well, so this function will always return a number. 

I took a look over existing usage and it looks mainly in use by those coming from WPcom migrations, specifically ones migrating right now/soon.

### DIFF
--- a/vip-helpers/vip-stats.php
+++ b/vip-helpers/vip-stats.php
@@ -68,7 +68,7 @@ function wpcom_vip_top_posts_array( $num_days = 30, $limit = 10, $end_date = fal
  * @param int          $post_id   Optional. The post ID to fetch stats for. Defaults to the $post global's value.
  * @param int          $num_days  Optional. How many days to go back to include in the stats. Default is 1. Maximum 90 days.
  * @param string|bool  $end_data  Optional. The last day of the desired time frame. Format is 'Y-m-d' (e.g. 2007-05-01) and default is today's UTC date.
- * @return int Number of pageviews, 0 if empty/error.
+ * @return int|false Number of pageviews or false on error.
  */
 function wpcom_vip_get_post_pageviews( $post_id = null, $num_days = 1, $end_date = false ) {
 	// Check Jetpack is present and active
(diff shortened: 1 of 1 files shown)


vip-helpers/vip-stats.php
```diff
@@ -92,13 +92,11 @@ function wpcom_vip_get_post_pageviews( $post_id = null, $num_days = 1, $end_date
 		$args['num_days'] = max( 1, min( 90, absint( $args['num_days'] ) ) );
 
 		$posts = stats_get_csv( 'postviews', $args );
-		$views = $posts[0];
+		$views = $posts[0]['views'];
```
rinatkhaziev, line 95:
```suggestion
		$views = $posts[0]['views'] ?? 0;
```
Maybe null coalesce here just to be extra safe.

vip-helpers/vip-stats.php
```diff
@@ -68,7 +68,7 @@ function wpcom_vip_top_posts_array( $num_days = 30, $limit = 10, $end_date = fal
  * @param int          $post_id   Optional. The post ID to fetch stats for. Defaults to the $post global's value.
  * @param int          $num_days  Optional. How many days to go back to include in the stats. Default is 1. Maximum 90 days.
  * @param string|bool  $end_data  Optional. The last day of the desired time frame. Format is 'Y-m-d' (e.g. 2007-05-01) and default is today's UTC date.
- * @return int|false Number of pageviews or false on error.
+ * @return int Number of pageviews, 0 if empty/error.
```
rinatkhaziev, line 71: The annotation changes but #L87-89 still `return false;`

WPprodigy: Ah, nice catch. Do think it would have been better to just return `0` on error, but a bit late now I guess.

vip-helpers/vip-stats.php
WPprodigy: Guess it can't hurt, though I did test lots of failed scenarios and the key was always there.

PR #1572, opened 2020-04-29
Akismet 4.1.5

mjangda:
https://blog.akismet.com/2020/04/29/version-4-1-5-of-the-akismet-wordpress-plugin-is-now-available/


mjangda:
Good thought. I checked the removed functions and filters and I don't see any instances of those in use in any client code (just copies of the Akismet plugin across various client repos).

brandon-m-skinner:
r937-stacks

PR #1575, opened 2020-05-01
Two Factor 0.5.2

mjangda:
https://github.com/WordPress/two-factor/releases/tag/0.5.2

## To Test

- Verify log in flows work with and without 2fa
- Verify enabling / disabling 2fa works
- Verify admins without 2fa do not have privileged access


brandon-m-skinner:
The first two seem to work fine. 

Locally, having 2fa enabled or not didn't affect any admin I made. Hit my time box, so didn't go any deeper.

dartiss:
Before a final merge, can you let AJ know in ticket #109103, as doing this will break their site due to a temporary fix that they put in place. Thanks.

mjangda:
Replaced by #1582

PR #1576, opened 2020-05-01
Filter `woocommerce_prevent_admin_access` to not block valid users from wp-admin

### LINKED ISSUE #1413: Two-factor auth downgrading user role to customer if WooCommerce is enabled.
## Expected/Desired Behavior
On a site with WooCommerce, a new admin account without 2FA set up would still have access to the WP Dashboard to edit his profile and set up 2FA.

## Actual Behavior
The two-factor module is downgrading the user role to a customer when they don't have 2FA enabled and thus locking new users out of the admin. Users get redirected to WooCommerce "My Account" page. 

1. Install WooCommerce
1. Create a new **admin** account
1. Try to log with that user and you will get redirected to "My Account page"

I believe the desired behavior would be to downgrade the user role to "subscriber" instead of "customer".

brettshumaker:
## Description

To determine if `wp-admin` access should be blocked, WooCommerce is checking the capabilities on the user and not the user's actual role. Since the VIP 2FA plugin alters the user's capabilities so early, when WooCommerce checks them to see if the user should be able to access `wp-admin`, valid users who should have `wp-admin` access are being blocked and redirected to the front end of the site.

I don't think it's unreasonable for WooCommerce to assume that a user's capabilities _should_ match the capabilites of their role. We can use the `woocommerce_prevent_admin_access` filter to introduce a strict check to see that a user's _role_ has the capabilities that WooCommerce checks for and determine whether or not the user should have access to `wp-admin`.

This commit adds that filter, returns early if the VIP 2FA plugin is not enabled OR if WooCommerce has already determined that the user should have `wp-admin` access. It also introduces a helper function that takes an array of capabilites, looks through all roles for that capability, and returns an array of role slugs for _any_ role that contains _any_ of the passed capabilities. If the current user's role matches any role in the returned array, we grant that user access to `wp-admin`.

Fixes #1413 

## Checklist

## Steps to Test

1. On a VIP Go Environment, create a user with the `administrator` role without setting up 2FA.
1. Install and activate WooCommerce.
1. Try to login with the `administrator` user.
1. Verify the user can access `wp-admin`.

To make sure this doesn't break existing WooCommerce logic for users that should _always_ be blocked from `wp-admin`:

1. Create a user with the `subscriber` role on the above VIP Go environment.
1. Try to login with the `subscriber` role.
1. Verify that the user is redirected away from `wp-admin`.


two-factor.php
```diff
@@ -236,3 +236,74 @@ function wpcom_vip_two_factor_admin_notice() {
...
+
+// Filter `woocommerce_prevent_admin_access` to allow users to set up 2FA on sites running WooCommerce.
+add_filter( 'woocommerce_prevent_admin_access', 'wpcom_vip_should_woocommerce_prevent_admin_access', 10, 1 );
+
+function wpcom_vip_should_woocommerce_prevent_admin_access( $prevent_access ) {
+
+	// If the 2FA plugin won't be enabled or $prevent access is already `false`, do nothing and return early.
+	$vip_two_factor_enabled = apply_filters( 'wpcom_vip_enable_two_factor', true );
+	if ( true !== $vip_two_factor_enabled || ! $prevent_access ) {
+		return $prevent_access;
```
brandon-m-skinner, line 248: Minor: Return false rather than return $prevent_access for clarity of intention.
```suggestion
		return false;
```

mjangda:
> Since the VIP 2FA plugin alters the user's capabilities so early, when WooCommerce checks them to see if the user should be able to access wp-admin, valid users who should have wp-admin access are being blocked and redirected to the front end of the site.

Would it be simpler to just track when we downngrade a user (in a static var) and then have the callback for `woocommerce_prevent_admin_access` just check that and act accordingly?

brettshumaker:
> Would it be simpler to just track when we downngrade a user (in a static var)

Yes, yes it is. Addressed this in 44b589a. Let me know your thoughts @mjangda 🙂

brandon-m-skinner:
r944-stacks

PR #1577, opened 2020-05-04
Added ratelimiting for indexing operations

brandon-m-skinner:
## Description

Added ratelimiting for indexing operations. A bit more succinct and focused than initial thoughts/test implementations.

Still needs tests and verification.

## Checklist

## Steps to Test

1. Checkout PR.
2. `./bin/phpunit-docker.sh tests/search/includes/classes/test-class-queue.php`

-----

1. Checkout PR.
2. Ensure Automattic/Elasticpress is working.
3. Set `$max_indexing_op_count` in `search/includes/classes/class-queue.php` to 0.
4. Add a new post in wp-admin.
5. The post should be added to the queue(`wp_vip_search_index_queue` database table)

### DIFF
--- a/search/includes/classes/class-queue.php
+++ b/search/includes/classes/class-queue.php
@@ -23,7 +23,7 @@ class Queue {
 	public const INDEX_QUEUEING_ENABLED_KEY = 'index_queueing_enabled';
 	public static $max_indexing_op_count = 3000 + 1; // 10 requests per second plus one for clealiness of comparing with Search::index_count_incr
 	private const INDEX_COUNT_TTL = 5 * MINUTE_IN_SECONDS; // Period for indexing operations
-	private const INDEX_QUEUEING_TTL = 1 * HOUR_IN_SECONDS; // Keep indexing op queueing for an hour once ratelimiting is triggered
+	private const INDEX_QUEUEING_TTL = 15 * MINUTE_IN_SECONDS; // Keep indexing op queueing for 15 minutes once ratelimiting is triggered
 
 	public function init() {
 		if ( ! $this->is_enabled() ) {
@@ -446,6 +446,9 @@ class Queue {
 
 			$ids = wp_list_pluck( $jobs, 'object_id' );
 
+			// Increment first to prevent overrunning ratelimiting
+			self::index_count_incr( count( $ids ) );
+
 			$indexable->bulk_index( $ids );
 
 			// TODO handle errors
@@ -480,9 +483,6 @@ class Queue {
 		foreach ( array_keys( $sync_manager->sync_queue ) as $object_id ) {
 			$this->queue_object( $object_id, $indexable_slug );
 		}
-
-		// Queue up a cron event to process these immediately
-		$this->cron->schedule_batch_job();
 
 		// Empty out the queue now that we've queued those items up
 		$sync_manager->sync_queue = [];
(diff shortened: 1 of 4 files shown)


search/includes/classes/class-queue.php
```diff
@@ -499,4 +508,54 @@ public function action__ep_after_bulk_index( $document_ids, $slug, $return ) {
...
+		}
+
+		if ( empty( $sync_manager->sync_queue ) ) {
+			return $bail;
+		}
+
+		// Determine whether indexing is ratelimited currently and if this is an indexing operation
+		$is_indexing_ratelimited = false !== wp_cache_get( self::INDEX_QUEUEING_ENABLED_KEY, self::INDEX_COUNT_CACHE_GROUP );
+		
+		if ( $this->is_enabled() ) {
```
nickdaugherty, line 534: Do we actually need this extra check? The filter won't be added if the queue is not enabled.

search/includes/classes/class-queue.php
```diff
@@ -499,4 +508,54 @@ public function action__ep_after_bulk_index( $document_ids, $slug, $return ) {
...
+			// If indexing operation ratelimiting is hit, queue index operations
+			$index_count_in_period = self::index_count_incr( count( $sync_manager->sync_queue ) );
```
nickdaugherty, line 536: "Pre" incrementing like this is smart...

search/includes/classes/class-queue.php
```diff
@@ -18,6 +18,13 @@ class Queue {
 
 	public $schema;
 
+	public const INDEX_COUNT_CACHE_GROUP = 'vip_search';
+	public const INDEX_COUNT_CACHE_KEY = 'index_op_count';
+	public const INDEX_QUEUEING_ENABLED_KEY = 'index_queueing_enabled';
+	public static $max_indexing_op_count = 3000 + 1; // 10 requests per second plus one for clealiness of comparing with Search::index_count_incr
+	private const INDEX_COUNT_TTL = 5 * MINUTE_IN_SECONDS; // Period for indexing operations
+	private const INDEX_QUEUEING_TTL = 1 * HOUR_IN_SECONDS; // Keep indexing op queueing for an hour once ratelimiting is triggered
```
nickdaugherty, line 26: This cool-down period seems a little long...something like 15m would still even things out while allowing a faster return to normalcy.

brandon-m-skinner:
@nickdaugherty 

I thought I did address those two items in the last change set. 

Don't schedule right off the bat:

https://github.com/Automattic/vip-go-mu-plugins/pull/1577/files#diff-3d380ee38a5fb9d99c67a47c87d1830cL476

Increment in the queue processor:

https://github.com/Automattic/vip-go-mu-plugins/pull/1577/files#diff-3d380ee38a5fb9d99c67a47c87d1830cR450

Is there more to it?

nickdaugherty:
@brandon-m-skinner whoops, missed those, probably b/c I didn't see a relevant commit message. The rate limit counter looks right, but the immediate scheduling still needs to happen, just not when we're currently rate limited.

brandon-m-skinner:
:+1:

Most of it is hidden behind the collapsable stuff. Not a big volume of changes but quite a few things happened. Should have probably broken it up more:

```
Applied feedback …
* Don't schedule jobs added to the queue, let the sweeper do it
* Add stat tracking for index ratelimiting
* Queue processing operations should count toward ratelimiting
* Reorganized how we get Elasticsearch hosts
* Removed unnecessary is_enabled check
* Adjusted tests for changes
```

nickdaugherty:
No worries, I could/should have looked closer (only just skimmed the titles of the new commits and looked through the code).

search/includes/classes/class-queue.php
```diff
@@ -472,9 +484,11 @@ public function intercept_ep_sync_manager_indexing( $bail, $sync_manager, $index
 			$this->queue_object( $object_id, $indexable_slug );
 		}
 
-		// Queue up a cron event to process these immediately
-		$this->cron->schedule_batch_job();
-
+		// If indexing operations are NOT currently ratelimited, queue up a cron event to process these immediately.
+		if ( false === wp_cache_get( self::INDEX_QUEUEING_ENABLED_KEY, self::INDEX_COUNT_CACHE_GROUP ) ) {
```
nickdaugherty, line 488: If the cache entry doesn't exist, `wp_cache_result()` will return `null`, which will not pass this conditional. What do you think about a simple helper function to check rate limit status? So that we're not duplicating the "is rate limited" logic in N spots. `Queue::is_rate_limited()`

brandon-m-skinner: @nickdaugherty 

wp_cache_get returns false if the cache entry doesn't exist in both the [docs](https://developer.wordpress.org/reference/functions/wp_cache_get/#return) and our [wp-memcached implementation](https://github.com/Automattic/wp-memcached/blob/master/object-cache.php#L283-L331).

It makes sense to move it to a function for readability though and to handle the possibility of future changes to index rate limiting detection.

nickdaugherty: Are you sure?

```
wp> $result = wp_cache_get( 'nothingblahblahblah' );
=> bool(false)
wp> wp> var_dump( $result );
bool(false)
=> NULL
```

nickdaugherty: Oh damnit, I misread the output, you're right 😄

brandon-m-skinner: @nickdaugherty Pretty sure.I think wp is just being weird:

```
wp> var_dump( false );
bool(false)
=> NULL
```

nickdaugherty: Yeah it prints the result of the line also, I mixed up the stdout from `var_dump()` and that part, need more coffee...

brandon-m-skinner:
r933-stacks

PR #1578, opened 2020-05-05
Support offloading WP_Query to ES via the 'es' query var

nickdaugherty:
## Description

Adds compatibility with es-wp-query, which uses `es` as the query var to enable offloading. ElasticPress uses `ep_integrate`, so this PR just lets you use either.

## Checklist

## Steps to Test

1. Pull PR
1. In functions.php (or anywhere), add this to enable query offloading on the main query:

```
add_filter( 'pre_get_posts', function( $query ) {
	if ( ! $query->is_main_query() ) { 
		return $query;
	}

	if ( ! $query->is_category() && ! $query->is_home() ) {
		return $query;
	}

	$query->set( 'es', true );
} );
```

1. Visit several frontend pages (hp, category, etc)
1. Check debug bar - should be recording 1 ES query per page (with arguments matching the normal WP_Query)


brandon-m-skinner:
r926-stacks

PR #1579, opened 2020-05-05
Add support for query mirroring for offloaded queries

nickdaugherty:
## Description

Adds "query mirroring" functionality for transparently testing the offloading of queries on VIP Search.

## Checklist

## Steps to Test

1. Pull PR
1. Make sure the site is enabling query mirroring - `define( 'VIP_ENABLE_SEARCH_QUERY_MIRRORING', true );`
1. Try some queries and check debug bar - there should be _no_ ES queries yet
1. Enable query mirroring by adding some rules to `should_mirror_wp_query()`:

```
		// Is this one of the targeted queries?
		if ( $query->is_home() || $query->is_category() || $query->is_archive() ) {
			return true;
		}
```

1. Request the HP, a category page, and an archive page
1. Each should run the regular DB query, but also the ES version of it (visible the EP tab of debug bar)
1. (Optional) Add some logging to `do_mirror_wp_query()` to show the diffs:

```
		error_log( 'did mirror' );
		error_log( 'diff is ' . var_export( $diff, true ) );
```

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -158,6 +158,9 @@ class Search {
 
 		// Allow query offloading with the 'es' query var (in addition to default 'ep_integrate')
 		add_filter( 'ep_elasticpress_enabled', array( $this, 'filter__ep_elasticpress_enabled' ), 10, 2 );
+
+		// For testing, mirror certain WP_Query's on certain sites
+		add_filter( 'the_posts', array( $this, 'filter__the_posts' ), 10, 2 );
 	}
 
 	protected function load_commands() {
@@ -233,6 +236,21 @@ class Search {
 			// Also not necessary with blocking=>false, but just in case.
 			'timeout' => 3,
 		] );
+	}
+
+	/**
+	 * Filter that runs at the end of WP_Query, used to transparently mirror certain WP_Query's on certain sites
+	 * for testing / evaluation
+	 */
+	public function filter__the_posts( &$posts, $query ) {
+		// If this query is one that should be transparently mirrored, run the mirroring
+		$should_mirror = $this->should_mirror_wp_query( $query );
+
+		if ( $should_mirror ) {
+			$this->mirror_wp_query( $query );
+		}
+
+		return $posts;
 	}
 
 	public function should_mirror_wp_query( $query ) {
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -327,7 +327,7 @@ class Search {
 			'extra' => array(
 				'query_vars' => $query->query_vars,
 				'diff' => $diff,
-				'uri' => $_SERVER[ 'REQUEST_URI' ],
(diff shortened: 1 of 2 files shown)


search/includes/classes/class-search.php
```diff
@@ -229,6 +238,100 @@ public function do_mirror_search_request() {
...
+
+		// If mirroring is not enabled at all, skip
+		if ( ! $this->is_query_mirroring_enabled() ) {
+			return false;
+		}
+
+		// Is this one of the targeted queries?
+	
+		return false;
+	}
```
brandon-m-skinner, line 270: It looks like you're missing some logic here.

tests/search/test-class-search.php
```diff
@@ -1117,6 +1147,336 @@ public function test__query_count_incr() {
...
+		define( 'VIP_ENABLE_SEARCH_QUERY_MIRRORING', false );
+
+		$es = new \Automattic\VIP\Search\Search();
+
+		$query = new \WP_Query();
+
+		$should_mirror = $es->should_mirror_wp_query( $query );
+
+		$this->assertFalse( $should_mirror );
+	}
```
brandon-m-skinner, line 1232: Which makes all the `should_mirror_wp_query` tests pass because you only `assertFalse` against them.

tests/search/test-class-search.php
```diff
@@ -1117,6 +1147,336 @@ public function test__query_count_incr() {
...
+	 * @runInSeparateProcess
+	 * @preserveGlobalState disabled
+	 */
+	public function test__should_mirror_wp_query_when_query_already_offloaded( $query ) {
+		$es = new \Automattic\VIP\Search\Search();
+		
+		$should_mirror = $es->should_mirror_wp_query( $query );
+
+		$this->assertFalse( $should_mirror );
+	}
```
brandon-m-skinner, line 1279: Same thing here.

search/includes/classes/class-search.php
nickdaugherty: That's intentional - I want to get this shipped with no queries offloaded, then in subsequent PRs, start enabling it on a per-site/per-query basis using rules in here.

brandon-m-skinner: Cool, sounds good.

nickdaugherty:
Something I just realized that we'll need to do for this PR - we should run the actual mirrored queries on `shutdown` instead of directly inside `the_posts`, so that the response can be sent before we perform additional work (to prevent these from slowing down the site).

nickdaugherty:
Moved the the mirrored queries to `shutdown` in c2f16dd - I'm going to add stats for mirrored queries + diffs now, then this is ready.

brandon-m-skinner:
@nickdaugherty Cool, I'll be here to approve and/or deploy if needed so we can ship this ASAP.

nickdaugherty:
Added stats in 1235b19e @brandon-m-skinner

brandon-m-skinner:
r934-stacks

PR #1582, opened 2020-05-07
Two Factor 0.6

mjangda:
Previously https://github.com/Automattic/vip-go-mu-plugins/pull/1575

https://github.com/WordPress/two-factor/releases/tag/0.6.0

## To Test

- Verify log in flows work with and without 2fa
- Verify enabling / disabling 2fa works
- Verify admins without 2fa do not have privileged access


mjangda:
r929-stacks

PR #1589, opened 2020-05-12
Add filter to allow VIP SMTP Bypass

mjangda:
By default, all VIP environments are automatically configured to send transactional emails (e.g. password reset, new comment) generated by WordPreess using VIP SMTP servers.

For cases, where a third party provider needs to be used for all emails from the site, including both marketing and transaction emails, the new filter (`vip_smtp_enabled`) can be use to disable loading of the VIP SMTP mailer and loading a custom one.

Note that loading a custom one is important otherwise no emails will get sent.

## Checklist

## Steps to Test

Need to test this on a sandbox, since local falls back to default mail.

1. Apply patch.
1. Trigger password reset and verify you get email.
1. Add `add_filter( 'vip_smtp_enabled', '__return_false' );`.
1. Trigger password reset and verify you do not get an email.


hanifn:
Tested this on a sandboxed site and seems to be working as expected
Screenshot 2020-05-12 14 40 58
Tests seems to be failing though

tests/test-vip-mail.php
```diff
@@ -6,6 +6,21 @@ public function setUp() {
 		reset_phpmailer_instance();
 	}
 
+	public function test__disable_vip_smtp() {
+		$GLOBALS['all_smtp_servers'] = [ 'server1', 'server2' ];
+
+		add_filter( 'vip_smtp_enabled', '__return_false' );
```
mjangda, line 12: This runs too late for the test to actually work. Need to work to run before `init` somehow. Or re-initialize the VIP_SMTP class.

hanifn: Can we just do another `( new VIP_SMTP() )->init()` after that filter was set?

hanifn: Ok tested that out and it didn't work

brandon-m-skinner:
@mjangda Is this still being worked on?

mjangda:
This would be nice, but is not critical. Closing.

PR #1592, opened 2020-05-14
Offload term edits to cron

brandon-m-skinner:
## Description

Depends on https://github.com/Automattic/ElasticPress/pull/42

There's a limit to how many posts you can queue in-place for reindexing. In light of some term circumstances, term edits should be offloaded to a cronjob that then gathers and queues all the posts that need to be reindexed.

## Checklist

## Steps to Test

1. Pull down PR.
2. Pull down https://github.com/Automattic/ElasticPress/pull/42
2. Find the term taxonomy id of a taxonomy with posts.
3. Schedule the cron job `wp cron event schedule 'vip_search_term_update' --term=<term taxonomy id you chose>`
4. Run the cron job `wp cron event run 'vip_search_term_update'`
5. Navigate around the main page to trigger it.
6. Monitor the ` wp_vip_search_index_queue` db table for the post ids of the term taxonomy id you used.
7. All posts from the term taxonomy id should have been queued.


search/includes/classes/class-queue.php
```diff
@@ -468,6 +468,31 @@ public function offload_indexing_to_queue() {
...
+		$term_taxonomy_id = $this->get_term_taxonomy_id( $term_id );
+
+		if ( is_null( $term_taxonomy_id ) ) {
+			return;
+		}
+
+		$this->cron->schedule_queue_posts_for_term_taxonomy_id( $term_taxonomy_id );
+	}
+
+	public function get_term_taxonomy_id( $term_id ) {
```
nickdaugherty, line 484: This same info can be loaded via `get_term( $term_id )`, which returns a WP_Term, with a `term_taxonomy_id`: https://developer.wordpress.org/reference/classes/wp_term/

I think we should queue the cron job with just the $term_id, then let the cron job itself handle it from there.

search/includes/classes/queue/class-cron.php
```diff
@@ -114,6 +120,32 @@ public function process_jobs( $job_ids ) {
 		$this->queue->process_jobs( $jobs );
 	}
 
+	/**
+	 * Given a term id, queue all posts for reindexing that match it
+	 *
+	 * @param {int} $term_taxonomy_id The term id you want to fetch/index
+	 */
+	public function queue_posts_for_term_taxonomy_id( $term_taxonomy_id ) {
+		error_log( 'Taxonomy='.$term_taxonomy_id );
```
nickdaugherty, line 129: Is this intentional?

search/includes/classes/queue/class-cron.php
```diff
@@ -114,6 +120,32 @@ public function process_jobs( $job_ids ) {
...
+		if ( ! is_int( $term_taxonomy_id ) ) {
+			return;
+		}
+
+		global $wpdb;
+
+		$query = $wpdb->prepare( 'SELECT `object_id` FROM `' . $wpdb->term_relationships . '` WHERE `term_taxonomy_id` = %d', $term_taxonomy_id );
```
nickdaugherty, line 136: Two things - we should batch this so we process more manageable chunks at a time (loading 275k+ rows at once probably won't work). We can also just use a WP_Query() to do this (vs direct DB query).

brandon-m-skinner: Made the switch, still have to figure out batching with WP_Query.

search/includes/classes/class-queue.php
```diff
@@ -468,6 +470,25 @@ public function offload_indexing_to_queue() {
...
+	 * Offload term indexing to the queue
+	 */
+	public function offload_term_indexing_to_queue( $term_id ) {
+		$term = \get_term( $term_id );
+
+		if ( is_wp_error( $term ) || ! is_object( $term ) ) {
+			return;
+		}
+
+		if ( $term->count <= self::MAX_SYNC_INDEXING_COUNT ) {
```
nickdaugherty, line 483: Is this conditional correct? We want to offload indexing to the queue when the term's object count is _over_ the max sync indexing limit.

search/includes/classes/queue/class-cron.php
```diff
@@ -114,9 +125,47 @@ public function process_jobs( $job_ids ) {
...
+		);
+
+		$posts = new \WP_Query( $args );
+
+		if ( ! $posts->have_posts() ) {
+			return;
+		}
+
+		while ( $args['paged'] <= $posts->max_num_pages ) {
+			while ( $posts->have_posts() ) {
```
nickdaugherty, line 152: `$posts->posts` will also just be a plain array here, so we can `foreach()` it like normal, which is a little cleaner.

search/includes/classes/queue/class-cron.php
```diff
@@ -30,6 +30,16 @@ class Cron {
...
 
+	/**
+	 * The name of the cron event for processing term updates
+	 */
+	const TERM_UPDATE_CRON_EVENT_NAME = 'vip_search_term_update';
+
+	/**
+	 * The number of objects queued per batch for term updates
+	 */
+	const TERM_UPDATE_BATCH_SIZE = 25000;
```
nickdaugherty, line 41: I think this batch size is too high...something in the 1k - 5k range would be safer.

search/includes/classes/queue/class-cron.php
```diff
@@ -114,9 +125,47 @@ public function process_jobs( $job_ids ) {
...
+	public function queue_posts_for_term_taxonomy_id( $term_taxonomy_id ) {
+		$args = array(
+			'posts_per_page' => self::TERM_UPDATE_BATCH_SIZE,
+			'paged' => 1,
+			'tax_query' => array(
+				array(
+					'field' => 'term_taxonomy_id',
+					'terms' => $term_taxonomy_id,
+				),
+			),
```
nickdaugherty, line 142: It's kind of a pain to use WP_Query here instead of a direct query, but we _have_ to because we need to run any filters that have been added, and this helps ensure we return only indexable posts.

Speaking of, are we sure this query will only return posts that are indexable? (either by status, or post type, etc). Would be bad if we accidentally indexed otherwise hidden posts.

search/includes/classes/queue/class-cron.php
```diff
@@ -114,9 +125,47 @@ public function process_jobs( $job_ids ) {
...
 
+	/**
+	 * Given a term taxonomy id, queue all posts for reindexing that match it
+	 *
+	 * @param {int} $term_taxonomy_id The term taxonomy id you want to index
+	 */
+	public function queue_posts_for_term_taxonomy_id( $term_taxonomy_id ) {
+		$args = array(
+			'posts_per_page' => self::TERM_UPDATE_BATCH_SIZE,
+			'paged' => 1,
```
nickdaugherty, line 136: We can optimize this query a little...

```suggestion
			'paged' => 1,
			'fields' => 'ids',
			'update_post_meta_cache' => false,
			'update_post_term_cache' => false,
```

search/includes/classes/class-queue.php
brandon-m-skinner: Yes it's correct. The function called is our original non-cron offloading functionality that works great.

Might need to be renamed or add a comment or something to make it clearer.

search/includes/classes/queue/class-cron.php
brandon-m-skinner: I've used it in excess of 100k for the speed gain. 1-5k took over an hour to add all the posts to the queue. It went down to less than 20 minutes by increasing it super high.

It feels almost like each page has a lengthy sleep at the end of it with how it was working. The batch size shouldn't have that much of an effect with no other changes.

search/includes/classes/queue/class-cron.php
brandon-m-skinner: I'm not sure if EP adds them or not. If not, we'll have to used the indexable/hooks to get the correct filters applied.

brandon-m-skinner:
@rinatkhaziev Yes, it is deliberate. 

Not much in the way of reasoning. I did a bit of discovery and found that term taxonomy id is an easy way to get all posts based on a term. I then adapted the original `$wpdb` stuff I wrote into a `WP_Query`. Since it worked well on both, I just kept it as is.

If there's a better way with the same speed and results, I'm open to changing it :+1:.

rinatkhaziev:
I was just thinking about this particular case ([per docs](https://developer.wordpress.org/reference/classes/wp_term/get_instance/)):

> $taxonomy
(string) (Optional) Limit matched terms to those matching $taxonomy. Only used for disambiguating potentially shared terms.

> (WP_Term|WP_Error|false) Term object, if found. **WP_Error if $term_id is shared between taxonomies and there's insufficient data to distinguish which term is intended. False for other failures**

After this point, we silently bail, which could introduce potential unobvious issues. I'll send in suggestions in a second, they shouldn't make any differences in terms of perf.

search/includes/classes/class-queue.php
```diff
@@ -53,7 +55,7 @@ public function is_enabled() {
 	}
 
 	public function setup_hooks() {
-		add_action( 'edit_terms', [ $this, 'offload_indexing_to_queue' ] );
+		add_action( 'edited_terms', [ $this, 'offload_term_indexing_to_queue' ], 0 );
```
rinatkhaziev, line 58:
```suggestion
		add_action( 'edited_terms', [ $this, 'offload_term_indexing_to_queue' ], 0, 3 );
```

search/includes/classes/class-queue.php
```diff
@@ -53,7 +55,7 @@ public function is_enabled() {
...
 		add_action( 'pre_delete_term', [ $this, 'offload_indexing_to_queue' ] );
```
rinatkhaziev, line 59:
```suggestion
		add_action( 'pre_delete_term', [ $this, 'offload_indexing_to_queue' ], 0, 2 );
```
(zero just for consistency with the `edited_term`, unless default `10` is important

search/includes/classes/class-queue.php
brandon-m-skinner: [edited_terms](https://developer.wordpress.org/reference/hooks/edited_terms/) only has 2 args

search/includes/classes/class-queue.php
```diff
@@ -468,6 +470,26 @@ public function offload_indexing_to_queue() {
 		}
 	}
```
rinatkhaziev, line 472:
```suggestion
	/**
	 * Wrapper to pass the correct taxonomy after a term has been edited
	 *
	 * @param int $term_id
	 * @param int $term_taxonomy_id
	 * @param string $taxonomy
	 * @return void
	 */
	public function disambiguate_and_offload_edited_term( $term_id, $term_taxonomy_id, $taxonomy ) {
		return $this->offload_term_indexing_to_queue( $term_id, $taxonomy );
	}

	/**
	 * Wrapper to pass the correct taxonomy before a term is deleted
	 *
	 * @param int $term_id
	 * @param string $taxonomy
	 * @return void
	 */
	public function disambiguate_and_offload_pre_delete_term( $term_id, $taxonomy ) {
		return $this->offload_term_indexing_to_queue( $term_id, $taxonomy );
	}

```

search/includes/classes/class-queue.php
```diff
@@ -468,6 +470,26 @@ public function offload_indexing_to_queue() {
...
 
+	/**
+	 * Offload term indexing to the queue
+	 */
+	public function offload_term_indexing_to_queue( $term_id ) {
```
rinatkhaziev, line 476:
```suggestion
	public function offload_term_indexing_to_queue( $term_id, $taxonomy ) {
```

brandon-m-skinner:
@rinatkhazie [edited_terms](https://developer.wordpress.org/reference/hooks/edited_terms/) has term id and taxonomy and we can just hand them both into [get_term](https://developer.wordpress.org/reference/functions/get_term/) to avoid the silent bailing?

rinatkhaziev:
Silly me confused `edited_terms` vs `edited_term` :) Yeah, let's do just that. (I'll try to remove the suggestions now)

search/includes/classes/class-queue.php
```diff
@@ -468,6 +470,26 @@ public function offload_indexing_to_queue() {
...
+		$term = \get_term( $term_id );
```
rinatkhaziev, line 477:
```suggestion
		$term = \get_term( $term_id, $taxonomy );
```

search/includes/classes/class-queue.php
brandon-m-skinner: Can't do that with deletes since offloading to cron creates a race condition between WordPress deleting the stuff and whatever delay there is until the cron job runs.

search/includes/classes/class-queue.php
rinatkhaziev: Yeah, I can't read :) `edited_term` vs `edited_terms`

search/includes/classes/class-queue.php
rinatkhaziev: Ah, I see

PR #1593, opened 2020-05-15
Add Search::query_es() helper

nickdaugherty:
## Description

For running a generic ES query (by $type) against VIP Search.

This will be used to provide 3rd party plugins (like es-wp-query) the ability to query ES directly without needing to worry about the internals of VIP Search.

This is a simple wrapper around the EP Indexable of $type, to hide the implementation details.

## Checklist

## Steps to Test

1. Check out PR.
1. In code, run a query (need ES query) - `Automattic\VIP\Search\Search::instance()->query_es( 'post', $es_args );
1. Verify response is as expected

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -181,6 +181,16 @@ class Search {
 		add_action( 'init', [ $this->healthcheck, 'init' ] );
 	}
 
+	public function query_es( $type, $es_args = array(), $wp_query_args = array(), $index_name = null ) {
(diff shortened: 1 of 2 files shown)


brandon-m-skinner:
Tested by `json_encode`ing formatted args via the filter we use in VIP Search. Then `json_decode`ing it as an assoc array and handing that into the `query_es` function. 

Exact same results as the search via the interface on the test site.

nickdaugherty:
r942-stacks

PR #1594, opened 2020-05-15
Add es-wp-query subtree for Search

nickdaugherty:
## Description

Adds automattic/es-wp-query as a subtree under `search` so we can use it for query offloading if it hasn't been included already.

## Checklist

## Steps to Test

Nothing to test yet as this code isn't hooked in anywhere, but worth doing a checkout to ensure the subtree is correctly populated.


nickdaugherty:
For this one, we need to be sure we omit "dev" files in SVN (like `bin` dir, etc) - cc @brandon-m-skinner .

nickdaugherty:
r946-stacks

PR #1602, opened 2020-05-22
Fix broken Jetpack connection during import cleanup

rinatkhaziev:
## Description

We have a situation where Jetpack connection gets broken and requires a manual reconnection during the import when the site is locked (i.e. the import is run without `--skip-prepare` flag passed.

`vip_is_jetpack_request` changes:

The list of IP ranges for Jetpack we were using was not current, updated the list from https://jetpack.com/ips-v4.json, added a comment about it being the canonical source.

The logic that was checking `$_SERVER[ 'REMOTE_ADDR' ]`, which in our case won't ever have the correct IP because requests are being proxied through LB. To address that the check was amended to also include `$_SERVER[ 'HTTP_X_FORWARDED_FOR' ]`

Finally, during VIP Init we have a check for defined & truthy `WPCOM_VIP_SITE_MAINTENANCE_MODE`, and in case it is, the request is terminated early after displaying "Site Unavailable" page. This also doesn't play nice with Jetpack, so we run `vip_is_jetpack_request` there to allow the request to proceed.

## Checklist

- [ ] This change works and has been tested locally (or has an appropriate fallback). - NA

## Testing

Testing is a bit tricky, here are the steps:

First off, lock a test site for import by calling `/import/end` 

Then, to verify the block exists:

```
curl --request POST \
  --url 'https://go-movers-test-ms.go-vip.net/test2/xmlrpc.php?for=jetpack' \
  --header 'user-agent: Jetpack by WordPress.com'
```

You should get `This site is temporarily undergoing maintenance, please try again in a few moments.` template (errors/site-maintenance.html)

Then, sandbox the site and apply the changes from this PR. 
Locally, add the `/etc/hosts` entry

Re-run the curl request above (adding the `--socks5` proxy arg and the special sandbox header), with the changes, you should get

```
<?xml version="1.0" encoding="UTF-8"?>
<methodResponse>
  <fault>
    <value>
      <struct>
        <member>
          <name>faultCode</name>
          <value><int>-32700</int></value>
        </member>
        <member>
          <name>faultString</name>
          <value><string>parse error. not well formed</string></value>
        </member>
      </struct>
    </value>
  </fault>
</methodResponse>
```

To test this unproxied (i.e. omitting the `--socks5` arg in curl call) you'd need to add your IP in `$jetpack_ips` array.


brandon-m-skinner:
@rinatkhaziev What are the test steps for this PR?

rinatkhaziev:
@brandon-m-skinner I updated the PR description to include the test steps, we can follow up internally to discuss further.

brandon-m-skinner:
@rinatkhaziev My proxy IP wasn't in the list so I had to add it. Once I did I got the expected results. 

Is this expected? If so, we're good to go.

rinatkhaziev:
Yes, I believe that's expected, the ranged intersect somewhat, but not 100%, I just happened to be proxied via one of those that fall into the range.

brandon-m-skinner:
r954-stacks

PR #1603, opened 2020-05-29
Add allow list for indexing post meta

brandon-m-skinner:
## Description

Need a way to not index all post meta due to field count restrictions. An allow list set by the user seems to be a good way to about it so that only the meta they actually need are included.

## Checklist

## Steps to Test

1. Pull down PR
2. `/bin/phpunit-docker.sh tests/search/test-class-search.php`

-----

1. Pull down PR.
2. Modify the post meta allow list via the`vip_search_post_meta_allow_list` filter in code.
3. Index posts using those meta and some others.
4. Query the freshly indexed posts and check that the only post meta indexed were the ones indicated in the allow list.

------

1. Pull down PR.
2. Set `FILES_CLIENT_SITE_ID` to one of the values in `Search::DISABLE_POST_META_ALLOW_LIST`
3. Index posts with any meta.
4. Query the freshly indexed posts and check that all post meta has been indexed, even if they aren't indicated on the allow list.

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -1054,14 +1054,43 @@ class Search {
 		 * Filters the allow list used for post meta indexing
 		 * 
 		 * @hook vip_search_post_meta_allow_list
-		 * @param {array} $current_allow_list The current allow list for post meta indexing
+		 * @param {array} $current_allow_list The current allow list for post meta indexing either as a list of post meta keys or as an associative array( e.g.: array( 'key' => true ); )
 		 * @param {WP_Post} $post The post whose meta data is being prepared
 		 * @return {array} $new_allow_list The new allow list for post_meta_indexing
 		 */
 		$client_post_meta_allow_list = apply_filters( 'vip_search_post_meta_allow_list', self::POST_META_DEFAULT_ALLOW_LIST, $post );
 
+		// If the array is empty, an array_intersect will result in an empty array. If the allow list isn't an array, assume no post meta is allow listed
+		if ( empty( $client_post_meta_allow_list ) || ! is_array( $client_post_meta_allow_list ) ) {
+			return array();
+		}
+
+		// If client meta allow list is an associative array
+		if ( array_keys( $client_post_meta_allow_list ) !== range( 0, count( $client_post_meta_allow_list ) - 1 ) ) {
+			/* 
+			 * Filter out values not set to true since the current format of the allow list as an associative array is:
+			 * 
+			 * array (
+			 * 		'key' => true,
+			 * );
+			 *
+			 * which means that anything besides true should logically be discarded
+			 */
+			$client_post_meta_allow_list = array_filter(
+				$client_post_meta_allow_list,
+				function( $value ) {
+					return true === $value;
+				}
+			);
+
+			$client_post_meta_allow_list = array_keys( $client_post_meta_allow_list );
+		}
+
+		// Since we're comparing result of get_post_meta(as $current_meta), we need to do an array_intersect_key since $current_meta should be an assoc array
+		$client_post_meta_allow_list_assoc = array_flip( $client_post_meta_allow_list );
+
 		// Only include meta that matches the allow list
-		$new_meta = array_intersect( $current_meta, $client_post_meta_allow_list );
+		$new_meta = array_intersect_key( $current_meta, $client_post_meta_allow_list_assoc );
 
 		return $new_meta;
 	}
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -1006,13 +1006,13 @@ class Search {
 	/**
 	 * Filter for reducing post meta for indexing to only the allow list and for
 	 * respecting the maximum field count gracefully
-	 */	
+	 */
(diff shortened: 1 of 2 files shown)


search/includes/classes/class-search.php
```diff
@@ -18,6 +18,51 @@ class Search {
 
 	private const MAX_SEARCH_LENGTH = 255;
 
+	private const DISABLE_POST_META_ALLOW_LIST = array(
+		2341,
+	);
+
+	// From https://github.com/Automattic/jetpack/blob/c36432aa890dc24cafee4c4362711ffcafb9c983/packages/sync/src/class-defaults.php#L689-L732
+	public const POST_META_DEFAULT_ALLOW_LIST = array(
```
rinatkhaziev, line 26: I'm not sure we should copy-pasta from Jetpack like that, some of it is not necessarily useful for our use cases.

search/includes/classes/class-search.php
```diff
@@ -1030,4 +1143,23 @@ private static function query_count_incr() {
...
+	 * @return {int} The maximum field count
+	 */
+	private function get_maximum_field_count() {
+		$field_count  = apply_filters( 'ep_total_field_limit', 5000 );
+
+		if ( ! is_int( $field_count ) ) {
+			$field_count = intval( $field_count );
+		}
+
+		if ( 20000 < $field_count ) {
```
rinatkhaziev, line 1159: Maybe `trigger_error` or `_doing_it_wrong` to inform the user instead of just silently enforcing max value here?

search/includes/classes/class-search.php
brandon-m-skinner: We needed a default and decided to just use what Jetpack was using so the behaviour matches up a bit better for our migration path.

search/includes/classes/class-search.php
brandon-m-skinner: `_doing_it_wrong` is probably a good way to go :+1:

brandon-m-skinner:
Couldn't figure out how to make the `@expectedIncorrectUsage` annotation work so I just turned off the `trigger_error` via the filter. 

I tried every possible variation of the function, namespace, visibility, etc... I could think of, but no dice. Just didn't want to work for me.

search/includes/classes/class-search.php
rinatkhaziev: I don't think it's the correct approach, I struggle to imagine a client searching for the data pertaining to a Restaraunt module or `_jetpack_post_thumbnail` which is only relevant to WP.com.

Once the list is set in stone it's unlikely to change, and that's how we get cruft that no one wants to touch and nobody is sure why it's there. 

My vote is to only allow Core WP meta and not bother with Jetpack-specific keys.

brandon-m-skinner: We went with a default of empty and then we'll flesh it out as people use it and when we start thinking about migration path.

search/includes/classes/class-search.php
```diff
@@ -1018,6 +1029,71 @@ public function get_current_host() {
...
+		}
+		
+		/**
+		 * Filters the allow list used for post meta indexing
+		 * 
+		 * @hook vip_search_post_meta_allow_list
+		 * @param {array} $current_allow_list The current allow list for post meta indexing
+		 * @return {array} $new_allow_list The new allow list for post_meta_indexing
+		 */
+		$client_post_meta_allow_list = apply_filters( 'vip_search_post_meta_allow_list', self::POST_META_DEFAULT_ALLOW_LIST );
```
nickdaugherty, line 1058: Let's pass the `$post` into this filter as well, so sites can index conditionally index meta (such as by post type).

search/includes/classes/class-search.php
```diff
@@ -1018,6 +1029,71 @@ public function get_current_host() {
...
+
+		// Only include meta that matches the allow list
+		$new_meta = array_intersect( $current_meta, $client_post_meta_allow_list );
+
+		$taxonomies = \get_taxonomies( '', 'names' );
+		$terms = \wp_get_object_terms( $post->ID, $taxonomies );
```
nickdaugherty, line 1064: If we want to keep a running total like this, we should move these two queries out and into `init` or similar, otherwise they'll run for every post being indexed (which is a ton on a bulk index).

brandon-m-skinner:
Decided to remove all the field counting stuff and use a statsd + Grafana alert approach to tracking/handling max field count stuff. 

More accurate, more performant, and easier.

search/includes/classes/class-search.php
```diff
@@ -1018,13 +1032,65 @@ public function get_current_host() {
...
+
+		if ( \is_wp_error( $post ) || ! is_object( $post ) ) {
+			return $current_meta;
+		}
+		
+		/**
+		 * Filters the allow list used for post meta indexing
+		 * 
+		 * @hook vip_search_post_meta_allow_list
+		 * @param {array} $current_allow_list The current allow list for post meta indexing
```
nickdaugherty, line 1057: We need to update the docblock for the new `$post` param as well.

brandon-m-skinner:
r995-stacks

PR #1606, opened 2020-06-01
Group purge requests in batches and send them to the `purge-batch` endpoint

alexcriss:
## Description

Group purge requests in batches of `CACHE_PURGE_BATCH_SIZE` and send the
batched purge requests to the `purge-batch` endpoint.

Batched requests are used only if the `PURGE_BATCH_SERVER_URL` constant
is defined in `wp-config.php`, otherwise the code falls back to the old
approach of sending individual requests to the `purge` endpoint.

When batched requests are supported, the number of URLs that can be
purged is increased to 4000, while it remains 100 in the fallback case.

## Checklist

- [x] This change works and has been tested locally (or has an appropriate fallback) - _not applicable_.
- [x] This change works and has been tested on a Go sandbox - _see test steps below_.
- [x] This change has relevant unit tests (if applicable) - _not applicable_.
- [x] This change has relevant documentation additions / updates (if applicable) - _not applicable_.

## Steps to test

1. Define `PURGE_BATCH_SERVER_URL` in your sandbox and point to a local HTTP server that dumps the body of the request.
2. Set the purge server type constant on you sandbox to `mangle`.
3. Edit a post on the test site and observe the output of the server running at `PURGE_BATCH_SERVER_URL`.
4. Compare the URLs listed to be purged to those sent as individual requests when `PURGE_BATCH_SERVER_URL` is not defined. The two lists should contain the same URLs.

### DIFF
--- a/vip-cache-manager/vip-cache-manager.php
+++ b/vip-cache-manager/vip-cache-manager.php
@@ -294,7 +294,7 @@ class WPCOM_VIP_Cache_Manager {
 
 		$max_purge_urls = defined( 'PURGE_BATCH_SERVER_URL' ) ? self::MAX_PURGE_BATCH_URLS : self::MAX_PURGE_URLS;
 		if ( $num_purge_urls > $max_purge_urls ) {
-			trigger_error( sprintf( 'vip-cache-manager: Trying to PURGE too many URLs (total count %s); limiting count to %d', number_format( $num_purge_urls ), $max_purge_urls ), E_USER_WARNING );
+			trigger_error( sprintf( 'vip-cache-manager: Trying to PURGE too many URLs (total count %s); limiting count to %d', number_format( $num_purge_urls ), number_format( $max_purge_urls ) ), E_USER_WARNING );
 			array_splice( $this->purge_urls, $max_purge_urls );
 		}
 
--- a/vip-cache-manager/vip-cache-manager.php
+++ b/vip-cache-manager/vip-cache-manager.php
@@ -111,7 +111,7 @@ class WPCOM_VIP_Cache_Manager {
 				if ( 500 < strlen( $data ) ) {
 					$compressed_data = gzencode( $data );
 					curl_setopt( $curl, CURLOPT_ENCODING, 'gzip' );
-					curl_setopt( $curl, CURLOPT_POSTFIELDS, $data );
+					curl_setopt( $curl, CURLOPT_POSTFIELDS, $compressed_data );
 				} else {
 					curl_setopt( $curl, CURLOPT_POSTFIELDS, $data );
 				}
--- a/vip-cache-manager/vip-cache-manager.php
+++ b/vip-cache-manager/vip-cache-manager.php
@@ -11,8 +11,10 @@ License: GPL version 2 or later - http://www.gnu.org/licenses/old-licenses/gpl-2
 require_once( __DIR__ . '/api.php' );
 
 class WPCOM_VIP_Cache_Manager {
-	const MAX_PURGE_URLS = 100;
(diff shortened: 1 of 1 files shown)


vip-cache-manager/vip-cache-manager.php
```diff
@@ -84,8 +86,40 @@ public function manual_purge_message() {
 	public function curl_multi( $requests ) {
 		$curl_multi = curl_multi_init();
 
-		foreach ( $requests as $req ) {
-			if ( defined( 'PURGE_SERVER_TYPE' ) && 'mangle' == PURGE_SERVER_TYPE ) {
+		if ( defined( 'PURGE_BATCH_SERVER_URL' ) && defined( 'PURGE_SERVER_TYPE' ) && 'mangle' == PURGE_SERVER_TYPE ) {
```
brandon-m-skinner, line 89: Should we use `===` here vs `==`?

vip-cache-manager/vip-cache-manager.php
```diff
@@ -84,8 +86,40 @@ public function manual_purge_message() {
...
+					curl_setopt( $curl, CURLOPT_HTTPHEADER, array( 'Content-Type: application/json', 'Content-Encoding: gzip' ) );
+					curl_setopt( $curl, CURLOPT_POSTFIELDS, $compressed_data );
+				} else {
+					curl_setopt( $curl, CURLOPT_HTTPHEADER, array( 'Content-Type: application/json' ) );
+					curl_setopt( $curl, CURLOPT_POSTFIELDS, $data );
+				}
+
+				curl_multi_add_handle( $curl_multi, $curl );
+			}
+		} elseif ( defined( 'PURGE_SERVER_TYPE' ) && 'mangle' == PURGE_SERVER_TYPE ) {
```
brandon-m-skinner, line 121: Should we use `===` here vs `==`?

vip-cache-manager/vip-cache-manager.php
```diff
@@ -84,8 +86,40 @@ public function manual_purge_message() {
...
+				$data = json_encode( $req_array );
+
+				$curl = curl_init( constant( 'PURGE_BATCH_SERVER_URL' ) );
+
+				curl_setopt( $curl, CURLOPT_HEADER, false );
+				curl_setopt( $curl, CURLOPT_RETURNTRANSFER, true );
+				curl_setopt( $curl, CURLOPT_TIMEOUT, 5 );
+				curl_setopt( $curl, CURLOPT_POST, true );
+
+				if ( 500 < strlen( $data ) ) {
```
rinatkhaziev, line 110: Is the concern here the potential overhead of gzipping something? With this data structure, it only takes about 5-6 URLs to go over 500. Here's my unscientific experiment: https://repl.it/@rinatkhaziev/KnowledgeableRoastedParentheses#main.php

Maybe it's not worth having its own `if/else` block, and we could just gzip regardless of the size of the batch.

vip-cache-manager/vip-cache-manager.php
alexcriss: Thanks for noticing this, it's fixed in d1828191.

vip-cache-manager/vip-cache-manager.php
alexcriss: Also fixed in d182819.

vip-cache-manager/vip-cache-manager.php
alexcriss: Yes, the idea was avoiding overhead from gzip operations. Compressing a small amount of data may increase the size of what is sent on the wire, and it will also incur a latency penalty due to the gzip operations.

alexcriss:
r973-stacks

PR #1609, opened 2020-06-02
Recommend vip-go-mu-dev for development

mjangda:
According to @brandon-m-skinner, "The lando env is magical".

Let's make it official.


mjangda:
Moved the `j8` init to https://github.com/Automattic/vip-go-mu-dev/pull/17

PR #1611, opened 2020-06-03
Display 'Not Proxied' flag when not proxied.

gudmdharalds:
It can be hard to see if one is really considered to be proxied when accessing `wp-admin`, especially when sites are set up with a frontend caching service. Using such services will cause the calling user being considered not proxied by the backend servers, even if the caller is using the A8C proxy.
 
This mu-plugin will make `wp-admin` display a small banner when not proxied and logged in as Automattican, indicating that the caller is not proxied.

## Checklist

## Steps to Test

1. Log in to `wp-admin` on a site as an Automattician
2. Make sure you are not using an A8C proxy.
3. Notice that small flag saying that you are not proxied in the left hand corner.


002-not-proxied-flag.php
```diff
@@ -0,0 +1,74 @@
...
+
+function maybe_show_not_proxied_flag() {
+	if ( ! \is_admin() ) {
+		return;
+	}
+
+	if (
+		( ! function_exists( '\is_user_logged_in' ) ) ||
+		( ! \is_user_logged_in() )
+	) {
```
joshbetz, line 34: Minor: Thoughts on making this all a single line? IMO multi-line `if` conditions aren't super readable, so I tend to avoid them unless absolutely necessary.

joshbetz:
I love the idea. I'm a bit skeptical about adding more number-prefixed files. Do you think it makes sense to put this in the `000-debug` directory and include it from there?

002-not-proxied-flag.php
gudmdharalds: > Minor: Thoughts on making this all a single line? IMO multi-line if conditions aren't super readable, so I tend to avoid them unless absolutely necessary.

I find multi-line if conditions more readable, but I generally don't mind. I can make the adjustment.

gudmdharalds:
> I love the idea. I'm a bit skeptical about adding more number-prefixed files. Do you think it makes sense to put this in the `000-debug` directory and include it from there?

I think it makes sense. I will make the adjustment :)

rinatkhaziev:
This is an amazing idea! 

One minor styling-related suggestion: 
to make it match the `Sandboxed` banner better, I've copy-pasted styles from there and removed the fixed `width` and `A8c` from  the text:

```
    letter-spacing: 0.2em;
    text-shadow: none;
    font-size: 9px;
    font-weight: bold;
    text-transform: uppercase;
    padding: 0 10px;
```

image

gudmdharalds:
Thanks for the suggestion! I altered the CSS and the string.

rinatkhaziev:
@brandon-m-skinner I'll leave the privilege to merge and deploy to you :)

000-debug/not-proxied-flag.php
```diff
@@ -0,0 +1,69 @@
...
+function enable_not_proxied_flag() {
+	add_action( 'admin_footer', __NAMESPACE__ . '\maybe_show_not_proxied_flag', 9999 ); // output later in the page
+}
+
+function maybe_show_not_proxied_flag() {
+	if ( ! \is_admin() ) {
+		return;
+	}
+
+	if ( ( ! function_exists( '\is_user_logged_in' ) ) || ( ! \is_user_logged_in() ) ) {
```
mjangda, line 31: Minor: `is_admin` and `is_user_logged_in` are redundant checks.

000-debug/not-proxied-flag.php
```diff
@@ -0,0 +1,69 @@
...
+		return;
+	}
+
+	if ( ( ! function_exists( '\is_automattician' ) ) || ( ! \is_automattician() ) ) {
+		return;
+	}
+
+	if ( ( ! function_exists( '\is_proxied_request' ) ) || ( \is_proxied_request() ) ) {
+		return;
+	}
```
mjangda, line 41: Here's a much cleaner and easier to read way to write this:

```suggestion
	$is_automattician = function_exists( '\is_automattician' ) && \is_automattician();
	$is_proxied = function_exists( '\is_proxied_request' ) && \is_proxied_request();

	if ( $is_automattician && ! $is_proxied ) {
		return;
	}
```

Right now, the if condition is overly complex and very easy to make an incorrect change (which is bad considering we don't have tests for this).

000-debug/not-proxied-flag.php
gudmdharalds: They are, will be removing them. They are leftovers from when a different action hook was used.

brandon-m-skinner:
r977-stacks

PR #1616, opened 2020-06-08
Two Factor: Better check for proxied Automattician

GaryJones:
## Description

Instead of just checking for the value of the `A8C_PROXIED_REQUEST` constant, we can also add the weaker check that the apparent VIP Support User does appear to be an Automattician. In combination with #1615, this will also check that the constant is also defined.

If the weaker check is not desired, then switching the comparison to `is_proxied_request()` would save checking against the constant directly.

## Checklist

- [ ] ~This change has relevant unit tests (if applicable).~

## Steps to Test

Example:

1. Check out PR.
1. Turn off proxy and ensure that Two Factor is requested when logging in as an administrator on a site that requires 2FA for administrators (default).
2. Turn on proxy again, but change VIP Support User to either have a non-Automattic email address, or delete the user meta that marks it as verified. See [here](https://github.com/Automattic/vip-support/blob/f613ab745501a3469b266e5447c694f5d2260931/class-vip-support-user.php#L740-L753).


joshbetz:
This will change the functionality in a way that may not be desired. Right now we can test certain things with other user roles.

GaryJones:
> Right now we can test certain things with other user roles.

Do those user roles still use an account which has a verified A8C address? Any manually added user or VIP Support user can still be used, irrespective of the user role assigned to it.

brandon-m-skinner:
LGTM, but needs more testing/considering RE: @joshbetz concerns

GaryJones:
@joshbetz 

> This will change the functionality in a way that may not be desired.

What were you thinking of here?

joshbetz:
I'm fine with moving forward here

pauarge (approved):
LGTM

PR #1619, opened 2020-06-11
Update es-wp-query subtree with fix

nickdaugherty:
## Description

Small fix to the vip-search mapping in es-wp-query

## Checklist

## Steps to Test

Nothing really needed here, but would need to offload a WP tag query - previously would return no results, now returns correct results.


nickdaugherty:
r976-stacks

PR #1624, opened 2020-06-12
Add queue size stat to statsd after processing a batch

brandon-m-skinner:
## Description

Add a gauge to keep track of queue size as it's processed.

## Checklist

### Setup statsd

1. Run StatsD locally

### If you don't have the `wp_vip_search_index_queue` table

1. Open wp shell and execute 2 and 3:
2. ` \Automattic\VIP\Search\Search::instance()->init();`
3. ` \Automattic\VIP\Search\Search::instance()->queue->schema->prepare_table();`

### Continuing...

1. Pull down PR.
2. Open wp shell and execute 3 and 4:
3. ` \Automattic\VIP\Search\Search::instance()->init();`
4. ` \Automattic\VIP\Search\Search::instance()->queue->count_jobs( 'all', 'all' );`
5. Insert a number of rows into `wp_vip_search_index_queue`(make sure some have the 'post' object type) and execute 6:
6. `\Automattic\VIP\Search\Search::instance()->queue->count_jobs( 'all', 'all' );`
7. The result of 6 should match the inserted number in 5.
8. `\Automattic\VIP\Search\Search::instance()->queue->record_queue_count_stat( \ElasticPress\Indexables::factory()->get( 'post' ) );`
9. There should be a new statsd gauge in the output for queue size where the stat matches the number of rows inserted in 5 that have an indexable.
10. Change the queue size, try different object types, and recheck.

### DIFF
--- a/search/includes/classes/class-queue.php
+++ b/search/includes/classes/class-queue.php
@@ -288,6 +288,16 @@ class Queue {
 		);
 	}
 
+	public function get_total_queue_size() {
+		global $wpdb;
+
+		$table_name = $this->schema->get_table_name();
+
+		$queue_size = $wpdb->get_var( "SELECT COUNT(*) FROM {$table_name} WHERE 1" ); // Cannot prepare table name. @codingStandardsIgnoreLine
+
+		return intval( $queue_size );
+	}
+
 	public function get_next_job_for_object( $object_id, $object_type ) {
 		global $wpdb;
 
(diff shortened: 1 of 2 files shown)


search/includes/classes/class-queue.php
```diff
@@ -288,6 +288,16 @@ public function count_jobs_due_now( $object_type = 'post' ) {
 		);
 	}
 
+	public function get_total_queue_size() {
```
nickdaugherty, line 291: We already have a `count_jobs()` function, maybe we can alter that to support all object types (currently defaults to `post`, but if we allow `null` or `any`, we can conditionally alter the query to return all. Same with the status.

nickdaugherty: Or, since our stats are per-index, which is technically per-type, we could count each type individually...hmm...

nickdaugherty: Would it be useful to track the queue statuses individually?

brandon-m-skinner: I'll make `count_jobs` handle 'all' statuses and 'all' object types.

Would probably be a good idea to pass in the object type when setting the stat just so we don't set both user and post queue gauges to the same value.

search/includes/classes/class-queue.php
```diff
@@ -266,13 +266,26 @@ public function count_jobs( $status, $object_type = 'post' ) {
...
-				$status,
-				$object_type
-			)
+		$query = $wpdb->prepare(
+			"SELECT COUNT(*) FROM {$table_name} WHERE `status` = %s AND `object_type` = %s", // Cannot prepare table name. @codingStandardsIgnoreLine
+			$status,
+			$object_type
 		);
+
+		if ( 'all' === strtolower( $status ) ) {
```
nickdaugherty, line 275: Will there be a case where we want to query all object types, but one specific status? Such as to see how many queued items there are across all types.

nickdaugherty: Not necessarily worth changing for this, just noting it :)

rinatkhaziev:
r984-stacks

PR #1625, opened 2020-06-12
Truncate search strings to a reasonable length

brandon-m-skinner:
## Description

This issue was initially because of a bug that resulted in hanging when search strings were longer than a certain length.

That's been resolved, but we should still limit search string lengths. 

## Checklist

## Steps to Test

1. Search with a string longer than 255. It shouldn't be truncated.
2. Pull down PR.
3. Search with a string longer than 255. It should be truncated to 255 in general and also in ElasticPress query monitor/debug bar.

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -907,13 +907,13 @@ class Search {
 		return $formatted_args;
 	}
 
-	public function truncate_search_string_length( $query_object ) {
-		if ( $query_object->is_search() ) {
-			$search = $query_object->query['s'];
+	public function truncate_search_string_length( $query ) {
+		if ( $query->is_search() ) {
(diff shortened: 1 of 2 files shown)


search/includes/classes/class-search.php
```diff
@@ -901,6 +907,16 @@ public function filter__ep_formatted_args( $formatted_args, $args ) {
 		return $formatted_args;
 	}
 
+	public function truncate_search_string_length( $query_object ) {
+		if ( $query_object->is_search() ) {
+			$search = $query_object->query['s'];
```
nickdaugherty, line 912: Minor - `WP_Query` also has a `get()`, so we could use that for consistency with the `set()` down below.

brandon-m-skinner:
r986-stacks

PR #1627, opened 2020-06-13
Alias elasticpress commands to vip-search

brandon-m-skinner:
## Description

Having multiple commands for search management doesn't make a lot of sense. 

Since the ElasticPress dashboard is also hidden, it makes sense to just alias the commands to vip-search.

## Checklist

## Steps to Test

1. Pull down PR.
2. `wp` shouldn't list elasticpress as a command.
3. `wp vip-search` should list all former elasticpress commands in addition to vip-search specific commands.


rinatkhaziev:
r989-stacks

PR #1629, opened 2020-06-16
Only mirror tax query if it's also not a search query

nickdaugherty:
## Description

Our automatic query mirroring for testing es-wp-query has a bug in that it offloads search queries that also have a taxonomy component. Since search queries will always differ somewhat, we can't reliably validate the results and they should be omitted.

## Checklist

## Steps to Test

1. Apply PR
1. Run a taxonomy query on a site with offloading enabled (like `?s&tag=something`)
1. Everything should work fine

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -304,7 +304,7 @@ class Search {
 				929,
 			);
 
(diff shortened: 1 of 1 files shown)


nickdaugherty:
r990-stacks

PR #1633, opened 2020-06-18
Stub Cache class, implement Advanced Post Cache compatibility.

rinatkhaziev:
## Description

To better handle and tweak any potential cache improvements it was decided to introduce `Automattic\VIP\Search\Cache` class. Currently, it doesn't do much except the Advanced Post Cache compatibility. 

That is, it disables APC for any EP-enabled query. APC disabling is pretty much a copy-pasta from es-wp-query adapters.

## Checklist

## Steps to Test

1. Check out PR.
1. Add a breakpoint or a good old `exit('here')` in`Advanced_Post_Cache::posts_request` (advanced-post-cache/advanced-post-cache.php:109)
1. Load the test site with `?s=hello`
1. Verify that breakpoint/exit doesn't fire for the search query.
1. Verify that breakpoint/exit fires for any non-search query.

### DIFF
--- a/search/includes/classes/class-cache.php
+++ b/search/includes/classes/class-cache.php
@@ -28,7 +28,7 @@ class Cache {
 			return;
 		}
 
-		if ( ! is_a( $query, 'WP_Query' ) ) {
+		if ( 'WP_Query' !== get_class( $query ) ) {
 			return;
 		}
 
--- a/search/includes/classes/class-cache.php
+++ b/search/includes/classes/class-cache.php
@@ -1,0 +1,71 @@
+<?php
+namespace Automattic\VIP\Search;
+
+class Cache {
+	function __construct() {
+		add_action( 'pre_get_posts', array( $this, 'disable_apc_for_ep_enabled_requests' ), 0 );
+	}
+
+	/**
+	 * Advanced Post Cache and ElasticPress do not work well together.
+	 * APC caches post IDs populates the $wp_query->posts by running `get_post` on each cached post id during `posts_request`,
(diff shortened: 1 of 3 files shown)


search/includes/classes/class-cache.php
```diff
@@ -0,0 +1,71 @@
...
+	 */
+	public function disable_apc_for_ep_enabled_requests( &$query ) {
+		global $advanced_post_cache_object;
+
+		static $disabled_apc = false;
+
+		if ( ! is_a( $advanced_post_cache_object, 'Advanced_Post_Cache' ) ) {
+			return;
+		}
+
```
nickdaugherty, line 30: Should we also check that `$query` is an instance of `WP_Query`, just to be safe?

rinatkhaziev: I thought about it, but ultimately figured that `\ElasticPress\Indexables::factory()->get( 'post' )->elasticpress_enabled( $query )` check wouldn't pass for any other type of query, so a separate check would be redundant? (That may be not true but just wishful thinking.)

nickdaugherty: It looks like `Indexable::elasticpress_enabled()` doesn't do any checking like that, sadly. Best to be defensive in our code anyway, in case EP were to change the implementation on us.

nickdaugherty: The check's would probably still return `false` and not _break_ at least.

rinatkhaziev: Fair point.

brandon-m-skinner:
r992-stacks

PR #1635, opened 2020-06-19
Split the ep_skip_query_integration filter into a reusable function

nickdaugherty:
## Description

So we can consistently check if query integration is enabled at all from multiple places.

This takes the logic from the filter, puts it in a new static function `is_query_integration_enabled()`, then updates the filter to instead use it.

The tests diff is pretty messy b/c the existing code got moved around.

No change in behavior is expected. Worth noting that I pulled out the code that was preventing query skipping when `vip_search_mirrored` was set on the WP_Query - that was old unused code from when we were trying to use EP for the query offloading via `ep_integrate`, and is no longer needed.

## Checklist

## Steps to Test

1. Check out PR.
1. Leave query integration disabled - constant undefined and option `vip_enable_vip_search_query_integration` is false (or not defined)
1. Run a regular search - `?s=test` - no ES query should be performed
1. Run same search but with `?s=test&es=true` - query _should_ go to ES
1. Repeat but set the `vip_enable_vip_search_query_integration` option to true instead of using the `es` query param
1. Repeat with the `VIP_ENABLE_VIP_SEARCH_QUERY_INTEGRATION` constant to `true` (and option set to false)

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -552,12 +552,42 @@ class Search {
 	}
 
 	/**
-	 * Separate plugin enabled and querying the index
+	 * Separate plugin enabling from querying the index
 	 *
-	 * The index can be tested at any time by setting an `es` query argument.
+	 * This function determines if VIP Search should take over queries (search, 'ep_integrate' => true, and 'es' => true)
+	 *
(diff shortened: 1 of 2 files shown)


nickdaugherty:
r994-stacks

PR #1636, opened 2020-06-22
Only load es wp query when needed

nickdaugherty:
## Description

Modifies how we load es-wp-query so that we only load it if query integration or query mirroring are enabled. This lets us enable Search on sites that are still using es-wp-query with other adapters, to get content indexed and evaluated before disabling the other adapter.

This refactors the es-wp-query loading logic into a standalone function with tests.

## Checklist

## Steps to Test

1. Check out PR.
1. Run some regular queries, like `?s=test` (should not break anything)
1. With query mirroring and integration disabled, try offloading a WP_Query with `'es' => true` in the args...it should not hit ES
1. Enable query mirroring or integration and repeat, it _should_ now hit ES

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -216,17 +216,7 @@ class Search {
 			require_once __DIR__ . '/../../debug-bar-elasticpress/debug-bar-elasticpress.php';
 		}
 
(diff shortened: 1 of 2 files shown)


brandon-m-skinner:
r996-stacks

PR #1637, opened 2020-06-23
Only send inconsistency alert in IRC in case it spans over two healthchecks.

rinatkhaziev:
## Description

To decrease alerting noise we should only alert when an index inconsistency persists across at least two intervals.

### DIFF
--- a/search/includes/classes/class-health-job.php
+++ b/search/includes/classes/class-health-job.php
@@ -150,23 +150,35 @@ class HealthJob {
 					$result[ 'diff' ]
 				);
 
-				$this->send_alert( '#vip-go-es-alerts', $message, 2 );	
+				$this->send_alert( '#vip-go-es-alerts', $message, 2, $result['type'] );
 			}
 		}
 	}
 
 	/**
 	 * Send an alert
-	 * 
+	 *
 	 * @see wpcom_vip_irc()
-	 * 
+	 *
 	 * @param string $channel IRC / Slack channel to send message to
 	 * @param string $message The message to send
 	 * @param int $level Alert level
-	 * 
-	 * @return bool Bool indicating if sending succeeded or failed
+	 * @param string $type content type
+	 *
+	 * @return bool Bool indicating if sending succeeded, failed or skipped
 	 */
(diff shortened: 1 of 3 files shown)


search/includes/classes/class-health-job.php
```diff
@@ -150,23 +150,35 @@ public function process_results( $results ) {
...
+	 * @param string $type content type
+	 *
+	 * @return bool Bool indicating if sending succeeded, failed or skipped
 	 */
-	public function send_alert( $channel, $message, $level ) {
+	public function send_alert( $channel, $message, $level, $type = '' ) {
+		// We only want to send an alert if a consistency check didn't correct itself in two intervals.
+		if ( $type ) {
+			$cache_key = "healthcheck_alert_seen:{$type}";
+			if ( false === wp_cache_get( $cache_key, 'vip-search' ) ) {
```
nickdaugherty, line 174:
```suggestion
			if ( false === wp_cache_get( $cache_key, Cache::CACHE_GROUP_KEY ) ) {
```

search/includes/classes/class-health-job.php
```diff
@@ -150,23 +150,35 @@ public function process_results( $results ) {
 					$result[ 'diff' ]
 				);
 
-				$this->send_alert( '#vip-go-es-alerts', $message, 2 );	
+				$this->send_alert( '#vip-go-es-alerts', $message, 2, $result['type'] );
```
nickdaugherty, line 153: I just realized there's an edge case here - if the if two different "entities" (like Post, User, or Term) have the same "type" registered here, they'll conflict. In practice this likely won't happen (so maybe not worth fixing). If there was a CPT called `user` or something it might happen.

I guess for completeness, we could pass through the entity + type to unique identify it.
```suggestion
				$this->send_alert( '#vip-go-es-alerts', $message, 2, "{$result['entity']}:{$result['type']}" );
```

search/includes/classes/class-health-job.php
rinatkhaziev: Oops, missed that one.

brandon-m-skinner:
r999-stacks

PR #1640, opened 2020-06-23
Fixes filtering the photon domain with a scheme mismatch

joshbetz:
If the home_url is `https://...` and the site loads on `http://...`, we offload to Photon erroneously. The goal here is just to ensure that the image is indeed hosted on this site, so a simple host match should suffice.


joshbetz:
Another way this can happen is if a post has an `http://...` URL hard-coded in the post-content, but the home_url is `https://...`. Basically anything that causes the image URL scheme to be different from the home/site URLs.

nickdaugherty:
Good find

joshbetz:
r1000-vip-go-stacks

PR #1642, opened 2020-06-24
Prevent Woocommerce logs from being written to the Files Service

joshbetz:
By default the WC log location is in the uploads directory, which is copied to the Files Service on VIP Go. The Files Service is not an appropriate location for logs because files are public and each new log line creates a new copy of the previous file along with the new line. We'll need to follow up with a way to clean up these files at end up in the /tmp directory.


joshbetz:
r1001-vip-go-stacks

PR #1646, opened 2020-06-26
Search: Add base index versioning support

nickdaugherty:
## Description

Adds the foundational index versioning code for tracking and managing versions.

Versions are stored inside an option (or a site option if in network mode), which has the following format:

```
array(
    'post' => array(
        array(
            'number' => 2,
            'active' => true,
            'created_time => 123456789,
            'activated_time' => 123456789,
        ),
        array(
            'number' => 3,
            'active' => false,
            'created_time => 123456789,
        )
    ),
    'user' => array(

    ),
    ...
)
```

This PR adds basic management of index versions and adds the filtering to the index name so that when versions are activated, the new index is used by default. For existing indexes (without any versions), the version is always 1, and we do _not_ add v1 to the index name (for backwards compatibility).

Previous index names:

```
vip-123-post-1
```

New index names (for versions > 2):

```
vip-123-post-1-v2
```

To run queries against non-active indexes (for example, to build up a new index in-place while another is active), simply override the current index number with:

```
$search->versioning->set_current_index_number( $indexable, 42 );
```

From then on, all queries will go to index version `42`, until the index number is overridden again, reset via `reset_current_index_number()`, or the request completes.

## Checklist

## Steps to Test

1. Setup a site using VIP Search prior to switching to PR - make sure index exists
1. Pull down PR
1. Perform some queries - everything should work the same as before
1. Re-index the content - everything should be the same (and index name is like `vip-123-post-1` (no version)
1. Create a new version...in wp shell: 
```
\Automattic\VIP\Search\Search::instance()->versioning->add_version( $indexable );
\Automattic\VIP\Search\Search::instance()->versioning->activate_version( $indexable, 2 );
```
1. Re-index content normally - `wp vip-search index --setup`
1. Perform some new queries - should now be hitting v2 of the index - `vip-123-post-1-v2`
1. Try to activate a non-existent index: 
```
\Automattic\VIP\Search\Search::instance()->versioning->activate_version( $indexable, 3 );
```
1. That should fail with an error b/c that index hasn't yet been registered via `add_version`

### DIFF
--- a/alloptions-limit.php
+++ b/alloptions-limit.php
@@ -129,7 +129,7 @@ function wpcom_vip_sanity_check_alloptions_notify( $size, $blocked = false ) {
 					array( 
 						'alias'       => 'alloptions/' . $site_id,
 						'description' => 'The size of AllOptions is reaching the max limit of 1 MB.',
-						'entity'      => $site_id,
+						'entity'      => (string) $site_id,
 						'priority'    => $opsgenie_alert_level,
 						'source'      => 'sites/alloptions-size',
 					),
--- a/alloptions-limit.php
+++ b/alloptions-limit.php
@@ -6,7 +6,7 @@
  * Author: Automattic
  * License: GPL version 2 or later - http://www.gnu.org/licenses/old-licenses/gpl-2.0.html
  */
-require_once( WP_CONTENT_DIR . '/mu-plugins/lib/utils/class-alerts.php' );
+require_once( __DIR__ . '/lib/utils/class-alerts.php' );
 use Automattic\VIP\Utils\Alerts;
 
 add_action( 'plugins_loaded', 'wpcom_vip_sanity_check_alloptions' );
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,6 @@
 {
     "require-dev": {
-        "phpunit/phpunit": "~9.2",
+        "phpunit/phpunit": "~6.5",
         "automattic/vipwpcs": "^2.0",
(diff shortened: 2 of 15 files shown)


brandon-m-skinner:
@nickdaugherty After adding another index and swapping versions, I couldn't reactivated v1. Is this expected?

> For existing indexes (without any versions), the version is always 1, and we do not add v1 to the index name (for backwards compatibility).

```
wp> \Automattic\VIP\Search\Search::instance()->versioning->activate_version( $indexable, 1 );
class WP_Error#2736 (2) {
  public $errors =>
  array(1) {
    'invalid-index-version' =>
    array(1) {
      [0] =>
      string(33) "The index version 1 was not found"
    }
  }
  public $error_data =>
  array(0) {
  }
}
wp> \Automattic\VIP\Search\Search::instance()->versioning->activate_version( $indexable, 2 );
bool(true)
```

Swapping between v2, v3, and v4 works well. Overall, really cool :+1:.

nickdaugherty:
Good find, no that's not intentional. It happens because we don't record v1 in the option, but we use the option values to check if the target version is valid.

Maybe we should add v1 into the array the first time you add a new version?

brandon-m-skinner:
Sounds good to me.

nickdaugherty:
@brandon-m-skinner Fixed the "activate v1" bug in e0d942cf2, then came up with a better way in 2c163b141

nickdaugherty:
r1021-stacks

PR #1647, opened 2020-06-27
Added statsd gauge for post index current field count

brandon-m-skinner:
## Description

Need to keep track of field counts in Elasticsearch so we can add an alert if someone exceeds it.

Added a custom interval so it can be adjusted easily. Chose to update each day right now as it shouldn't change too frequently with the post meta allow list for indexing in place.

## Checklist

- N/A This change has relevant unit tests (if applicable). _Don't see anything unit test would be able to test right now aside from invalid indexables, which is already tested to death_

## Steps to Test

1. Run StatsD locally
3. Apply PR.
4. Get value of `\Automattic\VIP\Config\Sync::CRON_EVENT_NAME`, currently `vip_config_sync_cron`
4. `wp cron event schedule vip_config_sync_cron`
5. `wp cron event run vip_config_sync_cron`
6. Compare 7 and 8. They should be equal. 
7. StatsD gauge, something like `'com.wordpress.elasticsearch.unknown.unknown.field_count.200508.vip-200508-post-1': 183` 
8. Directly check against Elasticsearch, something like `curl -s -XGET localhost:9200/index/_mapping?pretty | grep -w 'type' | wc -l`

### DIFF
--- a/search/includes/classes/class-fieldcountgaugejob.php
+++ b/search/includes/classes/class-fieldcountgaugejob.php
@@ -2,6 +2,8 @@
 
 namespace Automattic\VIP\Search;
 
+require_once( __DIR__ . '/../../../config/class-sync.php' );
+
 class FieldCountGaugeJob {
 	const CRON_EVENT_NAME = 'vip_config_sync_cron';
 
--- a/search/includes/classes/class-fieldcountgaugejob.php
+++ b/search/includes/classes/class-fieldcountgaugejob.php
@@ -1,0 +1,125 @@
+<?php
+
+namespace Automattic\VIP\Search;
+
+class FieldCountGaugeJob {
+	/**
+	 * The name of the scheduled cron event to run the gauge setting
+	 */
+	const CRON_EVENT_NAME = 'vip_search_field_count_gauge_set';
+
+	/**
+	 * Custom cron interval name
+	 */
+	const CRON_INTERVAL_NAME = 'vip_search_field_count_gauge_interval';
+
+	/**
+	 * Custom cron interval value
+	 */
+	const CRON_INTERVAL = 1 * \DAY_IN_SECONDS; // 30 minutes in seconds
+
+	const FIELD_COUNT_GAUGE_DISABLED_SITES = array();
+
+	/**
+	 * Initialize the job class
+	 *
+	 * @access public
+	 */
+	public function init() {
+		// We always add this action so that the job can unregister itself if it no longer should be running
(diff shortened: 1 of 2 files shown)


search/includes/classes/class-fieldcountgaugejob.php
```diff
@@ -0,0 +1,21 @@
...
+namespace Automattic\VIP\Search;
+
+class FieldCountGaugeJob {
+	const CRON_EVENT_NAME = 'vip_config_sync_cron';
+
+	const FIELD_COUNT_GAUGE_DISABLED_SITES = array();
+
+	public function init() {
+		// We always add this action so that the job can unregister itself if it no longer should be running
+		add_action( self::CRON_EVENT_NAME, [ $this, 'set_field_count_gauge' ] );
```
nickdaugherty, line 12: We can also reference the cron event directly from the Sync class (rather than redefine it):

```suggestion
		add_action( \Automattic\VIP\Config\Sync::CRON_EVENT_NAME, [ $this, 'set_field_count_gauge' ] );
```

search/includes/classes/class-search.php
```diff
@@ -208,6 +213,13 @@ protected function setup_healthchecks() {
 		add_action( 'init', [ $this->healthcheck, 'init' ] );
 	}
 
+	protected function setup_regular_stat_collection() {
+		$this->field_count_gauge = new FieldCountGaugeJob();
+
+		// Hook into init action to ensure cron-control has already been loaded
+		add_action( 'init', array( $this->field_count_gauge, 'init' ) );
```
nickdaugherty, line 220: Since we're not actually registering any jobs any more (piggy backing on the existing one), we can skip this delay and just init right away.

search/includes/classes/class-fieldcountgaugejob.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace Automattic\VIP\Search;
+
+class FieldCountGaugeJob {
+	const CRON_EVENT_NAME = 'vip_config_sync_cron';
+
+	const FIELD_COUNT_GAUGE_DISABLED_SITES = array();
```
nickdaugherty, line 8: Is this constant used?

nickdaugherty:
@brandon-m-skinner Looks like we ended up with some merge conflicts on this branch

brandon-m-skinner:
@nickdaugherty Expected either this one or the queue wait time one to conflict, depending on which one got merged first.

brandon-m-skinner:
r1016-stacks

PR #1648, opened 2020-06-29
Reduce _encloseme meta cleanup interval time

hanifn:
## Description

Reduce the interval between jobs from 20 minutes to 10 minutes.
During testing, we've found that 10000 post meta gets deleted in just about 1 minute so setting the interval to 20 minutes appears a little overkill.

## Checklist

## Steps to Test

1. Check out PR.
1. Upload changed files to a sandboxed site
1. ensure `define( 'ENABLE_VIP_ENCLOSEME_CLEANUP_ENV', true );` is in `vip-config.php`
1. `wp cron-control events list`
1. Ensure job is in events list and interval is set to 10 minutes


hanifn:
r1008-stacks

PR #1626, opened 2020-06-12
Add opsgenie alerts for AllOptions

mikeyarce:
## Description
Add OpsGenie Alerts for AllOptions! 😄

## Checklist

Please make sure the items below have been covered before requesting a review:

- [ ] This change works and has been tested locally (or has an appropriate fallback).
- [ ] This change works and has been tested on a Go sandbox.
- [ ] This change has relevant unit tests (if applicable).
- [ ] This change has relevant documentation additions / updates (if applicable).

## Steps to Test

I tried testing this on my sandbox but there was a problem with this:
https://github.com/Automattic/vip-go-mu-plugins/blob/9c293f8325d1b4d51a15e947591e6cae4cc67a68/lib/utils/class-alerts.php#L280

The `is_int` check wasn't passing because it was a string.  

If I bypass that, I get this error (I removed the URL, IP_ADDRESS and PORT):

```
[...]

PR #1649, opened 2020-06-29
Fix alloptions require

mikeyarce:
See: https://github.com/Automattic/vip-go-mu-plugins/pull/1626#issuecomment-650831789

There are some CircleCI failures due to how this is being loaded.  This should fix that.

cc @rinatkhaziev


brandon-m-skinner:
r1010-stacks

PR #1651, opened 2020-06-29
Record processed_from_queue stat

brandon-m-skinner:
## Description

Added a stat to keep track of when things are processed from the queue for an index.

## Checklist

## Steps to Test

1. Run StatsD locally
4. Apply PR.
5. Ensure VIP Search is enabled and the `wp_vip_search_index_queue` DB table and related functionality works locally.
6. Add a bunch of entries to the `wp_vip_search_index_queue` DB table.
7. `wp shell`
8. Build an array of assoc arrays where 'job_id' is equal to the object ids of the entries you added to `wp_vip_search_index_queue`.
9. Call the `Queue::process_jobs` function on the array of assoc arrays you built in `wp shell` previously.
10. The number of jobs you processed should equal the stat from the StatsD output. Should be something lilke `com.wordpress.elasticsearch.unknown.unknown.processed_from_queue.200508.vip-200508-post-1`.

### DIFF
--- a/search/includes/classes/class-queue.php
+++ b/search/includes/classes/class-queue.php
@@ -471,8 +471,34 @@ class Queue {
 			// Mark them as done in queue
 			$this->delete_jobs( $jobs );
 
+			$this->record_processed_from_queue_stat( count( $jobs ), $indexable );
+
 			$this->record_queue_count_stat( $indexable );
 		}
+	}
+
+	public function record_processed_from_queue_stat( $count, $indexable ) {
+		if ( ! $indexable ) {
+			return;
+		}
+
+		if ( ! is_int( $count ) ) {
+			$count = intval( $count );
+		}
+
+		$statsd_mode = 'processed_from_queue';
+
+		// Pull index name using the indexable slug from the EP indexable singleton
+		$statsd_index_name = $indexable->get_index_name();
+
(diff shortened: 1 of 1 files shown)


search/includes/classes/class-queue.php
```diff
@@ -471,10 +471,36 @@ public function process_jobs( $jobs ) {
...
 		}
 	}
 
+	public function record_processed_from_queue_stat( $count, $indexable ) {
+		if ( ! $indexable ) {
+			return;
+		}
+
+		if ( ! is_int( $count ) ) {
+			$count = intval( $count );
```
nickdaugherty, line 486: Should this instead return? If a non-int came in, we probably don't want to rely on PHP typecasting for stat values.

brandon-m-skinner: I don't think it makes much of a difference. Last time we had the same conversation on different sides 😄.

[intval on invalid values return 0](https://www.php.net/manual/en/function.intval.php#refsect1-function.intval-returnvalues). Incrementing by 0 doesn't do anything, so I think it's fine.

nickdaugherty: Alright, LGTM then

brandon-m-skinner:
r1017-stacks

PR #1654, opened 2020-07-01
Fix index healthchecks during indexing

nickdaugherty:
## Description

Previously, later filters (such as from client code) would override our code that prevented checks from running during indexing. We're not actually _disabling_ checking entirely, just temporarily skipping while indexing is happening, so this code path should not be filterable.

## Checklist

## Steps to Test

Tricky to test...need to think about that.


nickdaugherty:
Working on the broken tests.

brandon-m-skinner:
r1018-stacks

PR #1658, opened 2020-07-03
Reindex post if post meta is in allow list

brandon-m-skinner:
## Description

Depends upon https://github.com/Automattic/ElasticPress/pull/44

Since the post meta allow list has been added, posts should not be re-indexed if meta not on that list are changed.

## Checklist

## Steps to Test

1. `./bin/phpunit-docker.sh tests/search/test-class-search.php`

------

1. Apply PR. Ensure https://github.com/Automattic/ElasticPress/pull/44 is applied as well.
2. Ensure VIP Search in enabled and you have a properly consistent index.
3. Add some post meta to a post that isn't on the allow list. The post should not have been re-indexed
4. Change the value of that meta. The post should not have been re-indexed.
5. Delete the meta. The post should not have been re-indexed.
6. Add some post meta to a post that is on the allow list. The post should be re-indexed.
7. Change the value of that meta. The post should be re-indexed.
8. Delete the meta. The post should be re-indexed.

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -204,7 +204,7 @@ class Search {
 		add_filter( 'ep_total_field_limit', array( $this, 'limit_field_limit' ), PHP_INT_MAX );
 	
 		// Check if meta is on allow list. If not, don't re-index
-		add_filter( 'ep_skip_post_meta_sync', array( $this, 'filter__ep_skip_post_meta_sync' ), PHP_INT_MAX, 4 );
+		add_filter( 'ep_skip_post_meta_sync', array( $this, 'filter__ep_skip_post_meta_sync' ), PHP_INT_MAX, 5 );
 	}
 
 	protected function load_commands() {
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -204,10 +204,7 @@ class Search {
 		add_filter( 'ep_total_field_limit', array( $this, 'limit_field_limit' ), PHP_INT_MAX );
 	
 		// Check if meta is on allow list. If not, don't re-index
-		// Try to get these actions to run first
-		add_action( 'updated_post_meta', array( $this, 'maybe_ep_action_queue_meta_sync' ), PHP_INT_MIN, 4 );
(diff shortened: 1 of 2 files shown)


brandon-m-skinner:
I see. So when a filter/action is hooked, it never unhooks unless that's explicitly done. Gotcha.

search/includes/classes/class-search.php
```diff
@@ -1309,6 +1312,71 @@ public function limit_field_limit( $field_limit ) {
...
+				return $skip_sync;
+			}
+		}
+
+		// If post is invalid, respect current sync value
+		if ( is_null( $post ) ) {
+			return $skip_sync;
+		}
+
+		$post_meta_allow_list = \apply_filters( 'vip_search_post_meta_allow_list', self::POST_META_DEFAULT_ALLOW_LIST, $post );
```
nickdaugherty, line 1343: Would it make sense to wrap this `apply_filters()` call into a reusable function in the Search class, to encapsulate (and de-duplicate) the logic?

nickdaugherty: Better yet, all the code below should be de-duplicated too. If we can extract this out into something that both this function and `filter__ep_prepare_meta_data()` can use, it'll keep the core meta allow list logic DRY.

tests/search/test-class-search.php
```diff
@@ -1923,6 +1923,66 @@ function() {
...
+					'random_key',
+				);
+			}
+		);
+
+		$es = \Automattic\VIP\Search\Search::instance();
+
+		$this->assertTrue( $es->filter__ep_skip_post_meta_sync( true, $post, 40, 'random_key', 'random_value' ) );
+	}
+
```
nickdaugherty, line 1985: These unit tests are good for verifying the behavior of `filter__ep_skip_post_meta_sync()`, but we should also verify that our actual filter is being hooked and run correctly. So a similar test, but one that just runs `apply_filters( 'ep_skip_post_meta_sync', ... )` - that'll prove that we are correctly hooking into EP.

brandon-m-skinner:
r1029-stacks

PR #1662, opened 2020-07-08
Search: Add index_version column to queue table

nickdaugherty:
## Description

Now that we have index version support, we need to add versioning capabilities to the queue also. This is the first step of that, simply adding the `index_version column` (an int) and making it part of the unique index.

## Checklist

## Steps to Test

This needs some robust testing, as it's modifying the table schema. I'm not worried about the performance of adding a new index and dropping the old one (b/c most queue tables are small under normal conditions), but I am worried about all the different possible scenarios that exist with regards to the previous table state. There are three that I can think of:

* Site has table already, on schema version 2, empty
* Site has table already, on schema version 2, with existing jobs
* Site does not have the table, and the table is created

To test sites that already have the table:

1. Check out `master`
1. Make sure the table is created...can run a `wp post list` command to trigger. Then verify it exists in the DB with the `unique_object_and_status` index
1. Check out this PR
1. Run a CLI command again - this should trigger the db schema update
1. Inspect the table, it should have dropped the `unique_object_and_status` index and replaced it with `unique_object_status_version`
1. Then need to reset it back to schema version 2 (so checkout master, drop table, delete transient)
1. Then add some rows - could run `wp vip-search queue stress-test`, or edit a taxonomy term, to add rows to the queue
1. Check out the PR again
1. Run a WP CLI command to trigger schema upgrade
1. Schema upgrade should succeed and table should have new index that includes version

To test sites that don't have the table at all:

1. Check out this PR
1. Delete the `wp_vip_search_index_queue` table
1. Delete the schema version transient - `wp transient delete vip_search_queue_db_version`
1. Run another WP CLI command to trigger schema creation - `wp post list` - this should not output any errors
1. Inspect the table - it should have the `index_version` column and an index called `unique_object_status_version`

Also worth pointing out is that the `IF EXISTS` support for `DROP INDEX` was added in MariaDB 10.1.4 (and MySQL doesn't have it). AFAICT we're running either `10.1.40` or `10.2.x` on VIP, so we _should_ be ok there.


brandon-m-skinner:
Blocked due to risk. Needs a bit more testing and more eyes on.

nickdaugherty:
r1023-stacks

PR #1664, opened 2020-07-08
Use update_stats directly

brandon-m-skinner:
## Description
 
StatsD::increment only increases by one. Need to use StatsD::update_stats directly to increase a stat by a different amount.

## Checklist


brandon-m-skinner:
r1027-stacks

PR #1666, opened 2020-07-10
Schedule multiple processor jobs on shorter interval

brandon-m-skinner:
## Description

The sweeper needs to run more often and also process more jobs per run so that when a massive amount of indexing operations are offloaded to be processed asynchronously, they are processed in a timely manner.

## Checklist

## Steps to Test

Before doing any testing, set `PROCESSOR_MAX_OBJECTS_PER_CRON_EVENT` to 10

---

1. Ensure cron-control isn't active.
2. Apply PR
3. Ensure you have lots of posts in your test instance and `wp vip-search queue stress test` to populate `wp_vip_search_index_queue`.
2. `wp shell` and run `Automattic\VIP\Search\Search::instance()->queue->cron->sweep_jobs()`
3. Only one `vip_search_queue_processor` event should be scheduled(can be checked via `wp option get cron` or other method).

-----

1. Ensure cron-control is active and running properly.
2. Apply PR
3. Ensure you have lots of posts in your test instance and `wp vip-search queue stress test` to populate `wp_vip_search_index_queue`
3. Only five `vip_search_queue_processor` events should be scheduled(can be checked via `SELECT COUNT(*) FROM wp_a8c_cron_control_jobs WHERE action = 'vip_search_queue_processor' AND status != 'complete';` if you're fast enough, otherwise fine the general timestamp window in the table and compare).
4. Repeat 2 a few times and check again. Should still only be five events at any time.

-----

1. Ensure cron-control is active and running properly.
2. Apply PR.
3. Ensure `vip_search_queue_sweeper_interval` is now 1 minute instead of 5.


search/includes/classes/queue/class-cron.php
```diff
@@ -201,15 +206,47 @@ public function sweep_jobs() {
...
+
+		global $wpdb;
+
+		$table_name = \Automattic\WP\Cron_Control\Events_Store::instance()->get_table_name();
+
+		$current_processor_job_count = $wpdb->get_var( "SELECT COUNT(*) FROM $table_name WHERE action = 'vip_search_queue_processor' AND status != 'complete'" ); // Cannot prepare table name. @codingStandardsIgnoreLine
+
+		// If null, only schedule one job
+		if ( is_null( $current_processor_job_count ) ) {
+			return self::MAX_PROCESSOR_JOB_COUNT;
```
nickdaugherty, line 246: It would be more predictable to return a `0` here, then let the other code handle this situation, since the function is `get_processor_job_count()`. Thoughts?

brandon-m-skinner: If we're going that route it would make more sense to just return null IMO. 0 implies the cron-control is active and running. 

Getting a null from the processor count means that the table is likely missing meaning cron-control isn't active. Which means we should schedule only one job per round as needed like it did before IMO.

nickdaugherty: Ah ok, I think I understand this a bit better now, and it should be good 👍

nickdaugherty:
Are there any tests we can add that verify the backoff behavior that prevents too many jobs being scheduled?

brandon-m-skinner:
> Are there any tests we can add that verify the backoff behavior that prevents too many jobs being scheduled?

I thought about it but it's kind of hard to do with unit tests. Would need integration testing with cron-control to make something that wasn't super janky.

brandon-m-skinner:
r1033-stacks

PR #1667, opened 2020-07-11
Remove wp-cli-to-cron

mjangda:
No longer necessary since WP-CLI access is now available.

Need to remove the relevant code in the SVN repo as part of the deploy.


brandon-m-skinner:
r1031-stacks

PR #1668, opened 2020-07-13
Search: update query integration argument from 'es' to 'vip_search_enabled' to eliminate the chances of conflicts with client code

rinatkhaziev:
There was at least one instance where 'es' GET argument conflicted with client code, this PR should eliminate that by changing the argument to more specific `vip_search_enabled`

## Checklist

- [NA] This change has relevant documentation additions / updates (if applicable).

## Steps to Test

1. Check out PR.
1. Make sure query integration is not enabled either via the constant or the option.
1. Go to a site, pass `?vip_search_enabled=true&s={SEARCH_TERM}`
1. Verify in QM that the query is offloaded

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -627,7 +627,7 @@ class Search {
 	 *
 	 * This function determines if VIP Search should take over queries (search, 'ep_integrate' => true, and 'es' => true)
 	 *
-	 * The integration can be tested at any time by setting an `es` query argument (?es=true).
+	 * The integration can be tested at any time by setting an `es` query argument (?vip_search_enabled=true).
 	 * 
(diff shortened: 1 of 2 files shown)


search/includes/classes/class-search.php
```diff
@@ -632,7 +633,7 @@ public function filter__ep_feature_active( $active, $feature_settings, $feature
 	 * constant should be set to `true`, which will enable query integration for all requests
 	 */
 	public static function is_query_integration_enabled() {
-		if ( isset( $_GET['es'] ) ) {
+		if ( isset( $_GET[ self::QUERY_INTEGRATION_FORCE_ENABLE_KEY ] ) ) {
```
nickdaugherty, line 636: Just noting we also need to update the comment for this function.

rinatkhaziev: Updated

rinatkhaziev:
r1035-stacks

PR #1671, opened 2020-07-13
Search: Add version support to queue (continued)

nickdaugherty:
## Description

Now that we have the index_version in the queue schema, this PR implements support in the underlying queue functions.

Most queue related functions now have an `$options` argument that lets us specify the index version to use. Internally, when processing jobs, the jobs are first organized by index version, then indexable, then processed in a loop, calling `\Automattic\VIP\Search\Search::instance()->versioning->set_current_version_number()` for each unique index version / indexable combo.

All stats will be correct (use the correct index name) but note that for the "added to queue" stats we don't yet queue up anything for non-active indexes (this will be in a future version), and we'll need to call `set_current_version_number()` when queueing up jobs for those other indexes (otherwise the filtered index name won't be right).

As noted above, this doesn't attempt to queue up anything for non-active index versions, it just adds version support to the queueing functions, so that we can add that feature in the next PR.

## Checklist

## Steps to Test

1. Pull down PR
1. Change search/includes/classes/class-queue.php to set `MAX_SYNC_INDEXING_COUNT` to something low, like `1`.
1. Edit a taxonomy term
1. Run cron jobs due now, which will queue up all posts that were changed - `lando wp cron event run --due-now`
1. View the contents of the `wp_vip_search_index_queue` table - should have 1 job per post from that category, with `index_version` set to the currently active index version (probably `1`)
1. Add a new index version `lando wp vip-search index-versions add post` (probably will be `2`)
1. Setup the index - `lando wp vip-search index --setup --version=2 --indexables=post`
1. Activate the new index - `lando wp vip-search index-versions activate post 2`
1. Re-edit the taxonomy term
1. Re-run cron jobs due now - `lando wp cron event run --due-now`
1. Check the queue table again - should now have queued items for both version 1 and 2
1. Run the cron events due now again, until all items are processed and removed from the queue


nickdaugherty:
r1039-stacks

PR #1683, opened 2020-07-23
Remove Query Monitor submodule

mjangda:
And replace it with a raw implementation from the .org directory. We're generally trending towards not relying too much on submodules and this will help simplify the process of updating the version of QM moving forward.

No deploy should be necessary since there's no functional changes (we're continuing to use 3.1.1).

## Checklist

Please make sure the items below have been covered before requesting a review:

- [ ] This change works and has been tested locally (or has an appropriate fallback).
- [ ] This change works and has been tested on a Go sandbox.
- [ ] This change has relevant unit tests (if applicable).
- [ ] This change has relevant documentation additions / updates (if applicable).

## Steps to Test

1. Make sure site doesn't explode with and without QM active.

PR #1685, opened 2020-07-27
QM: Re-add previously ignored output directory

### LINKED ISSUE #1684: Query Monitor missing files on new VIP Go local installs
## Expected/Desired Behavior

Access Query Monitor while logged in to local environment.

## Actual Behavior

Either Query Monitor has no output ( admin ) or a critical error is thrown ( front-end ).

The error is complaining about the missing file `mu-plugins/query-monitor/output/Headers.php`

That file is not checked in to this repo as of #1683 when the submodule was removed in favor of checking in the files. 

You can see that there's no output folder on master here: https://github.com/Automattic/vip-go-mu-plugins/tree/master/query-monitor

1. Start a new VIP Go local environment
1. Log in to WP Admin
1. Try to open Query Monitor

I was able to fix this locally by replacing the `mu-plugins/query-monitor` folder with a freshly downloaded copy of Query Monitor version 3.1.1 from WordPress.org

https://wordpress.org/plugins/query-monitor/advanced/

On old installs that previously used the submodule version of Query Monitor, no file changes where shown in the local git repository.

GaryJones:
## Description

In #1683, the Query Monitor plugin was changed from being a git submodule to a standard set of files. The version of the plugin was not changed.

However, the `output` directory was missing from #1683, due to a rule in the `.gitignore` file.

This PR adds an exception to that rule for the `query-monitor/output` directory, and adds the missing directory itself.

Without this fix, the original report (#1683) says that "Query Monitor has no output ( admin ) or a critical error is thrown ( front-end )."

This didn't show-up at the time of the merge three days ago, as (fortunately) there was [no deploy](https://github.com/Automattic/vip-go-mu-plugins/pull/1683#issuecomment-663731121).

Fixes #1684.

## Checklist

## Steps to Test

1. Compare https://github.com/Automattic/vip-go-mu-plugins/tree/master@%7B2020-07-27T19:00:06Z%7D/query-monitor to the download of [QM 3.1.1 from wp.org](https://wordpress.org/plugins/query-monitor/advanced/)
1. Note the missing `output` directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,8 @@ phpdoc
 # Composer
 /vendor/
 output
+# Allow the QM output directory, as it's needed.
+!query-monitor/output
 
 # npm
 node_modules
--- a/.gitignore
+++ b/.gitignore
@@ -13,9 +13,7 @@ phpdoc
 
 # Composer
 /vendor/
-output
-# Allow the QM output directory, as it's needed.
-!query-monitor/output
+/output/
 
 # npm
 node_modules
--- a/query-monitor/output/Headers.php
+++ b/query-monitor/output/Headers.php
@@ -1,0 +1,24 @@
+<?php
+/**
+ * Abstract output class for HTTP headers.
+ *
+ * @package query-monitor
+ */
+
+abstract class QM_Output_Headers extends QM_Output {
+
+	public function output() {
+
+		$id = $this->collector->id;
+
+		foreach ( $this->get_output() as $key => $value ) {
+			if ( is_scalar( $value ) ) {
+				header( sprintf( 'X-QM-%s-%s: %s', $id, $key, $value ) );
+			} else {
+				header( sprintf( 'X-QM-%s-%s: %s', $id, $key, json_encode( $value ) ) );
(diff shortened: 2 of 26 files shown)


.gitignore
```diff
@@ -14,6 +14,8 @@ phpdoc
 # Composer
 /vendor/
 output
+# Allow the QM output directory, as it's needed.
+!query-monitor/output
```
mjangda, line 18: Would it be better to make the rule above more explicit? i.e. `/output/`? As it's likely we could run into similar problems with other paths in the future?

GaryJones: Could do. 

`output` was [added](https://github.com/Automattic/vip-go-mu-plugins/commit/037bdde9e0c3506708576c6f2d0ba0ca1724246b) four years ago, when adding PHPDoc-generation, and I would presume that this is the first instance which has arisen from it.

Someone let me know (or just go ahead) if this change to `/output/` is definitely wanted.

mjangda:
Thanks @GaryJones and @douglas-johnson!

PR #1689, opened 2020-07-28
Use full file path for file service caching

hanifn:
## Description

The current file service API cache uses the file base name as the cache key but this causes problem when a file is uploaded with the same name but on a different path. Functions like `file_exist` will return `true` even though the file is on a different path.

## Checklist

## Steps to Test

1. Check out PR.
1. Upload changes to a sandboxed site
1. `wp --allow-root shell`
1. Upload a new file: `file_put_contents( 'vip://wp-content/uploads/path1/test.txt', 'test' );`
1. `file_exist( 'vip://wp-content/uploads/path2/test.txt' );`
1. Verify that `file_exist` call returns `false`.

### DIFF
--- a/files/class-api-cache.php
+++ b/files/class-api-cache.php
@@ -59,69 +59,55 @@ class API_Cache {
 	}
 
 	public function get_file( $filepath ) {
-		$file_name = basename( $filepath );
-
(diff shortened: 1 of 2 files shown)


hanifn:
r1045-stacks

PR #1692, opened 2020-07-30
Search: Create the new index in ES when a version is registered

nickdaugherty:
## Description

Previously, adding a new index version only registered it with VIP Search - it didn't actually create it in Elasticsearch until you ran the `index` command. This changes that behavior so that the index (with mapping) is created when the new version is registered.

This will simplify things and prevent user error/confusion. It also ensures the index is there so we can start syncing changes to it immediately without causing problems.

A full indexing is still required - content is not automatically indexed.

## Checklist

## Steps to Test

1. Check out PR.
1. Make sure you are running https://github.com/Automattic/vip-go-mu-dev/pull/34 and have rebuilt the env (`lando rebuild --yes`) to expose ES to your local machine.
1. `curl 127.0.0.1:19200/_aliases?pretty=true` to list all existing indexes in ES
1. `lando wp vip-search index-versions add post` to add a new version
1. `curl 127.0.0.1:19200/_aliases?pretty=true` - the new index version should show up
1. `curl 127.0.0.1:19200/vip-200508-post-1-v3/_mapping?pretty=true` (use appropriate index name) - the mapping should be present and have the right stuff (from the ElasticPress mapping file)

### DIFF
--- a/search/includes/classes/class-versioning.php
+++ b/search/includes/classes/class-versioning.php
@@ -220,7 +220,26 @@ class Versioning {
 			return $result;
 		}
 
+		// Setup the index + mapping so that it's available for immediate use (as changes will start getting replicated here)
+		$this->create_versioned_index_with_mapping( $indexable, $new_version_number );
+
 		return $new_version;
+	}
+
+	/**
+	 * Create the index in ES and put the mapping
+	 * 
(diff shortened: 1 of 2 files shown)


nickdaugherty:
r1053-stacks

PR #1693, opened 2020-07-30
Add unified lint command

nickdaugherty:
## Description

We currently have several lint commands, primarily 1 for checking PHP syntax, and another for checking PHPCS rules for the diff. This combines those into an easier-to-type single command - `npm run lint`.

## Checklist

- n/a This change works and has been tested on a Go sandbox.

## Steps to Test

1. `npm run lint` - should run both `./bin/php-lint.sh` and `npm run phpcs:changed:remote`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,8 +47,7 @@ job-references:
           name: "Run Tests"
           command: |
             if [[ "$RUN_EXTRA" == "1" ]]; then
-              npm run phplint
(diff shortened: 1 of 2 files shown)


nickdaugherty:
Should also update the command in the CircleCI config so it runs the same thing that we run in dev

nickdaugherty:
> The indenting seems off

We had a mix of tabs and spaces in package.json 🤔 ...fixed now.

PR #1697, opened 2020-08-03
Search: Add version support to document count validation job and command

nickdaugherty:
## Description

Now that index versioning is supported, we need to monitor the consistency of the various versions. This PR adds version support to `wp vip-search health validate-counts` and the cron job that monitors counts.

## Checklist

## Steps to Test

1. Pull down PR
1. Make sure you have at least an additional index version (`lando wp vip-search index-versions list post`)
1. Run `lando wp vip-search health validate-counts` - if there are no inconsistencies, there is no difference
1. Introduce an inconsistency by adding a new version (but not running the index command for it) - `lando wp vip-search index-versions add post`
1. `lando wp vip-search health validate-counts` again - should get inconsistencies as the new index is not built. Like:

```
❌  inconsistencies found when counting entity: post, type: post, index_version: 5 - (DB: 1001, ES: 1, Diff: -1000)
❌  inconsistencies found when counting entity: post, type: page, index_version: 5 - (DB: 51, ES: 1, Diff: -50)
```

1. Test the cron job - not sure this is worth doing manually, I did it myself, but basically you have to add an error_log() to `wpcom_vip_irc()`, edit the `HealthJob::is_enabled()` function to always return true, visit a page so that the cron job is registered, manually run cron with `wp cron event run --due-now`, then check the `wp-content/wp-debug.log` file to see if the message went through.
1. Test users - steps TBD (not sure this works at all)

### DIFF
--- a/search/includes/classes/class-health-job.php
+++ b/search/includes/classes/class-health-job.php
@@ -112,9 +112,18 @@ class HealthJob {
 			$this->process_results( $user_results );
 		}
 
-		$post_results = Health::validate_index_posts_count();
+		// Check all versions
+		$search = \Automattic\VIP\Search\Search::instance();
 
-		$this->process_results( $post_results );
+		$post_indexable = \ElasticPress\Indexables::factory()->get( 'post' );
+
+		$versions = $search->versioning->get_versions( $post_indexable );
+
+		foreach( $versions as $version ) {
+			$post_results = Health::validate_index_posts_count( $version['number'] );
+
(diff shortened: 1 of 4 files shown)


nickdaugherty:
Just thought of something - it would be good to designate active vs non-active indexes in the alerts, so folks viewing them can prioritize accordingly. Will add...

nickdaugherty:
r1058-stacks

PR #1698, opened 2020-08-04
Search: Improve memory management for validate_contents

netsuso:
## Description

Improve memory management for the command `wp vip-search health validate-contents`.

On very large sites with hundreds of thousands of documents, this command would take increasingly amounts of memory (e.g, around 4 GB for 100k documents). This is mostly due to non-flushing cache and logs, so this PR tackles them:

- Disable Elasticpress query log
- Disable WP DB query log
- Flush WP object cache after every validation block

## Checklist

n/a This change has relevant unit tests (if applicable).
n/a This change has relevant documentation additions / updates (if applicable).

## Steps to Test

1. Check out this PR
1. Check out the required PR for ElasticPress submodule: https://github.com/Automattic/ElasticPress/pull/48
1. Run `wp vip-search health validate-contents` for a large site and check that memory usage doesn't increase (or increases slowly)

### DIFF
--- a/search/includes/classes/class-health.php
+++ b/search/includes/classes/class-health.php
@@ -240,9 +240,7 @@ class Health {
 			}
 
 			// Cleanup WordPress object cache to keep memory usage under control
-			global $wp_object_cache;
-			$wp_object_cache->flush();
-			$wp_object_cache->group_ops = array();
+			wpcom_vip_flush_object_cache();
 
 			if ( $is_cli && ! $silent ) {
 				echo sprintf( "...%s\n", empty( $result ) ? '✅' : '❌' ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
--- a/search/includes/classes/class-health.php
+++ b/search/includes/classes/class-health.php
@@ -240,7 +240,7 @@ class Health {
 			}
 
 			// Cleanup WordPress object cache to keep memory usage under control
-			wpcom_vip_flush_object_cache();
+			vip_reset_local_object_cache();
 
 			if ( $is_cli && ! $silent ) {
 				echo sprintf( "...%s\n", empty( $result ) ? '✅' : '❌' ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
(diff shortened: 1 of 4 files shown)


search/includes/classes/class-health.php
```diff
@@ -239,6 +239,11 @@ public static function validate_index_posts_content( $start_post_id = 1, $last_p
 				$last_post_id = self::get_last_post_id();
 			}
 
+			// Cleanup WordPress object cache to keep memory usage under control
+			global $wp_object_cache;
+			$wp_object_cache->flush();
+			$wp_object_cache->group_ops = array();
```
netsuso, line 245: The array `group_ops` is not emptied by `flush()`, but it also increases in size during the command execution. That's why I'm emptying it manually.

nickdaugherty: There's also a function provided by `\WPCOM_VIP_CLI_Command` that does this (and a few other things) called `\WPCOM_VIP_CLI_Command::stop_the_insanity()` - we should use that one to handle these.

netsuso: Oh! Sure, that's a better option, thanks! 😊

vip-helpers/vip-wp-cli.php
```diff
@@ -2,23 +2,12 @@
...
-			return;
-
-		$wp_object_cache->group_ops = array();
-		$wp_object_cache->memcache_debug = array();
-		$wp_object_cache->cache = array();
-
-		if ( is_callable( $wp_object_cache, '__remoteset' ) )
-			$wp_object_cache->__remoteset(); // important
+		wpcom_vip_flush_db_query_log();
+		wpcom_vip_flush_object_cache();
```
netsuso, line 10: Moved this logic to `vip-caching.php`, as it's easier to use there (as this is only for classes inheriting `WPCOM_VIP_CLI_Command`) and because it makes more sense.

netsuso:
> Since `wp_cache_flush()` invalidates everything in memcache ("flush" has a special meaning), can we name these functions something else? Maybe `vip_reset_db_query_log()` and `vip_reset_local_object_cache()`? (We can also drop the `wpcom_` prefix for new stuff).

Sounds quite reasonable! I've changed them following your proposals.

netsuso:
r1061-stacks

PR #1703, opened 2020-08-10
Add support for WordPress 5.5 environment type

netsuso:
## Description

Add support for [WordPress 5.5 environment type](https://make.wordpress.org/core/2020/07/24/new-wp_get_environment_type-function-in-wordpress-5-5/). We set the environment type to exactly the same value as the VIP Go environment name.

WordPress expects the environment name to be one of "production", "staging" or "development". When the VIP Go environment cannot be mapped to any of these standard types, it's added to the list of accepted environments for that site.

## Checklist

## Steps to Test

1. Check out PR and setup dev environment using WordPress 5.5
1. Get the result of `wp_get_environment_type()`
1. Verify that it matches the definition of `VIP_GO_APP_ENVIRONMENT`


000-vip-init.php
```diff
@@ -198,6 +198,41 @@
...
+		$environment_types = array(
+			'development',
+			'testing',
+			'staging',
+			'preprod',
+			'production',
+		);
+
+		// Add the environment to the environment type list if it's a non standard name
+		if ( ! in_array( $environment_name, $environment_types ) ) {
```
nickdaugherty, line 228: `in_array()` is kinda slow in PHP since it traverses the array. This is a pretty short array, but given that this runs on _every_ request, I wonder if there's a more performant way to do it. Two thoughts come to mind (I haven't verified that they're faster) - either just add the env to the array, then call `array_unique()` later, or flip it to an associative array like `development => true` so each only appears once, then convert to the WP format with `array_keys()`.

nickdaugherty: It's also unclear from https://make.wordpress.org/core/2020/07/24/new-wp_get_environment_type-function-in-wordpress-5-5/ what the plan for `WP_ENVIRONMENT_TYPES` is. The current implementation here won't take into account the available environment types of other child environments, just the current one, which _may_ (or may not) be a problem at some point.

My worry is that if we're defining this array but not including all the environment types for a given site and it's child environments, things can get confusing. As an example, say we have a production site with `develop`, `uat`, `uat2`, `testing` environments. Each of those will only know about itself, but the others are all "valid" environment types for that site. Thoughts?

000-vip-init.php
```diff
@@ -198,6 +198,41 @@
 	}
 }
 
+// Set WordPress >5.5 environment name
+if ( defined( 'VIP_GO_APP_ENVIRONMENT' ) && ! defined( 'WP_ENVIRONMENT_TYPE' ) ) {
```
nickdaugherty, line 202: Can we make this code more testable, maybe by moving it into one or two functions?

000-vip-init.php
nickdaugherty: For the types constant, it's probably good enough to ship what we have without trying to account for all env types of a site, and improve it later by moving it to a constant generated in wp-config.php, if needed.

netsuso: > I wonder if there's a more performant way to do it.

Certainly, I also considered that. Checking the key of an array would be less expensive, but when you include `array_keys()` it becomes more expensive than `in_array()`. It's also possible to declare both arrays (one to search, and the other one to apply to the define), but it makes the code not very clean.

I've been doing [some tests](https://gist.github.com/netsuso/0a557e2f6f64bc3e77ffde53aeb4c41a), and in summary there are three options:

- in_array (Code 1): 15 nanoseconds
- two arrays (Code 3): : 9 nanoseconds
- if with shortcut for production (Code 5): 4 nanoseconds

Not that Code 5 is very clear, but at least looks nicer than Code 3 with its two arrays. At least it's the fastest solution for production environments (which are the most sensitive). In any case, we are talking about very short times...

netsuso: > The current implementation here won't take into account the available environment types

Agree with that, it seems somewhat incomplete... And also agree that this may be revisited in the future if needed.

000-vip-init.php
netsuso: I thought about declaring a function, but there weren't any other in this file. This leads me to think that maybe this is not the best place to put this. Any other suggestion? Would there be a better place to initialize this, or should I keep it in this file inside a couple of functions?

johnbillion:
My understanding of this new feature in WordPress core is to allow code to target environment "types", not specific environments. We already have means to do environment-specific testing via platform-specific constants, environment variables, host names, etc.

The core list is restricted to `development`, `staging`, and `production` as that covers the available "types" of environments. Actual environments should fit within one of those types (eg. a UAT site can be classified as either development or staging, as you see fit), which allows for consistency when performing logic for checking whether certain features should be enabled or certain code should run.

Imagine a popular plugin such as Yoast SEO which needs to disable a feature on development sites, it can do this by targeting the `development` environment type. With these changes which add extra environment types, that becomes unreliable.

nickdaugherty:
That makes sense, thanks @johnbillion! @netsuso in light of that, what do you think about leaving `WP_ENVIRONMENT_TYPES` as-is, and mapping `production` => `production`, `develop` => `develop`, then everything else to `staging`?

000-vip-init.php
nickdaugherty: > but there weren't any other in this file

I guess the tricky bit is that it's still hard to test even with a function, b/c our test bootstrap file includes 000-vip-init.php, which will define the constants and change the behavior of tests. We should rethink how that works long term...I guess for now it won't make much of a difference in practice.

netsuso:
It seems a reasonable assumption. I suppose all our production environments are called `production`, in which case any mapping is safe (I wouldn't like a production environment having a non standard name and being incorrectly mapped to `staging` or `development`). I'll change that

netsuso:
We have finally decided that we are setting the current VIP Go environment name as the WP environment type, even if it doesn't match any of the standard types.

This feature is quite open for the moment, and it will need time to settle and for users to decide how to use it. In VIP we have defined its usage as a way for developers to obtain the environment name, exactly as it's defined in the VIP Go Platform.

brandon-m-skinner:
r1076-stacks

PR #1704, opened 2020-08-12
Disable core sitemap xml

brandon-m-skinner:
## Description

Sitemap changes are causing issues in WordPress 5.5. 

We'll turn them off globally for now.


brandon-m-skinner:
r1063-stacks

PR #1706, opened 2020-08-12
Block new `_encloseme` post metadata from being added

hanifn:
## Description

Blocks the addition of new `_encloseme` post metadata. We have a cleanup job that deletes existing `_encloseme` post meta but we neglected to prevent new ones from being added. This lead to some sites looping forever on the cleanup job.

## Checklist

## Steps to Test

TBD


nickdaugherty:
cc @rebeccahum for review as you've worked on this extensively.

rebeccahum:
It should already be blocked on core as of 5.3 (https://core.trac.wordpress.org/ticket/48014): https://github.com/WordPress/WordPress/blob/1f85e7484f154912d8869886ee3e5b68b06c0bc4/wp-includes/post.php#L7048

Are you seeing it being added where it should not be?

mjangda:
If the meta isn't being used, there's no point in adding it. Which means that we'll run our cleanup once but the meta will keep getting added and we'll need to re-run the cleanup routine again in the future.

There's also another problem on some sites where they have an import routine that runs in between the cleanup which means we are constantly in a race to add posts + meta and cleanup meta (which is a routine that will run forever).

tests/performance/test-do-pings.php
```diff
@@ -0,0 +1,55 @@
...
+		$this->assertFalse( $result, 'block_encloseme_metadata_filter should return false since the meta key is _encloseme' );
+		$this->assertFalse( \apply_filters( 'add_post_metadata', $should_update, $object_id, $meta_key, $meta_value, $unique ), 'add_post_metadata should return false since the meta key is _encloseme' );
+
+		$should_update = false;
+
+		$result = block_encloseme_metadata_filter( $should_update, $object_id, $meta_key, $meta_value, $unique );
+
+		$this->assertFalse( $result, 'block_encloseme_metadata_filter should return false' );
+		$this->assertFalse( \apply_filters( 'add_post_metadata', $should_update, $object_id, $meta_key, $meta_value, $unique ), 'add_post_metadata should return false' );
+
```
brandon-m-skinner, line 52:
```suggestion
```

tests/performance/test-do-pings.php
```diff
@@ -0,0 +1,53 @@
...
+		
+		$this->assertTrue( $result, 'block_encloseme_metadata_filter should return true' );
+		$this->assertTrue( \apply_filters( 'add_post_metadata', $should_update, $object_id, $meta_key, $meta_value, $unique ), 'add_post_metadata should return true' );
+
+		$should_update = false;
+
+		$result = block_encloseme_metadata_filter( $should_update, $object_id, $meta_key, $meta_value, $unique );
+
+		$this->assertFalse( $result, 'block_encloseme_metadata_filter should return false' );
+		$this->assertFalse( \apply_filters( 'add_post_metadata', $should_update, $object_id, $meta_key, $meta_value, $unique ), 'add_post_metadata should return false' );
```
mjangda, line 31: I would suggest making this more of an integration-y test where we call `add_post_meta` and verify it works or not works (depending on the meta key). I think, in theory, the unit test makes sense, but what's more interesting and useful is whether it actually gets applied in practice (i.e. via add/update post meta).

Thoughts?

brandon-m-skinner: Sounds good to me.

brandon-m-skinner: Added the new tests in https://github.com/Automattic/vip-go-mu-plugins/pull/1706/commits/d857a662fc4e6d8d4c5e87aa73385f8aef5f38fb

performance/do-pings.php
```diff
@@ -16,3 +16,13 @@ function disable_pings( $event ) {
 	return $event;
 }
 add_action( 'schedule_event', __NAMESPACE__ . '\disable_pings' );
+
+// Don't allow new _encloseme metas
+function block_encloseme_metadata_filter( $should_update, $object_id, $meta_key, $meta_value, $unique ) {
+	if ( '_encloseme' === $meta_key ) {
+		$should_update = false;
```
nickdaugherty, line 23: Noting that `add_metadata()` will return this value, so the calling code will ultimately receive `false`. Just want to be sure that's accounted for with the `_encloseme` stuff - will it perpetually keep trying to add it or similar, if it receives `false`?

brandon-m-skinner:
@rebeccahum @hanifn 

> Noting that add_metadata() will return this value, so the calling code will ultimately receive false. It bails if the return is not null. Just want to be sure that's accounted for with the _encloseme stuff - will it perpetually keep trying to add it or similar, if it receives false?

Thoughts?

mjangda:
> It bails if the return is not null. Just want to be sure that's accounted for with the _encloseme stuff - will it perpetually keep trying to add it or similar, if it receives false?

https://github.com/WordPress/WordPress/blob/master/wp-includes/post.php#L7048

There's no checks there; just fire and forget so we should be fine.

brandon-m-skinner:
r1084-stacks

PR #1709, opened 2020-08-12
Fix parse_str() missing parameter deprecation

GaryJones:
## Description

PHPCompatibility flagged the original instance of the code as:

> The "result" parameter for function parse_str() is missing. Passing this parameter is no longer optional. The optional nature of the parameter is deprecated since PHP 7.2 (PHPCompatibility.FunctionUse.OptionalToRequiredFunctionParameters.parse_str_resultSoftRequired)

As per the `parse_str()` [documentation](php.net/manual/en/function.parse-str.php), "using this function without the result parameter is highly DISCOURAGED and DEPRECATED as of PHP 7.2."

To ensure that the fix didn't change the behaviour, I wrote a bunch of unit tests before making the change, and then made the change, and the unit tests still passed.

As an aside, the unit tests revealed some non-ideal values being returned, to which I've left comments on the tests, but fixing them is out the scope of this PR.

## Checklist

## Steps to Test

1. Check out the first commit, https://github.com/Automattic/vip-go-mu-plugins/commit/71b00588712841ff205c491a5168cefcefb07ee7.
1. `lando ssh` then `phpunit tests/vip-helpers/test-wpcom-vip-set-image-quality-for-url.php` to just run this unit test.
1. Confirm tests pass, and give a visual inspection of the suitability.
1. Check out this PR / the second commit (https://github.com/Automattic/vip-go-mu-plugins/commit/76796c561097f11e1f060dcaa51b411d08b6775b).
1. Re-run the unit tests and confirm they still pass.

### DIFF
--- a/tests/vip-helpers/test-wpcom-vip-set-image-quality-for-url.php
+++ b/tests/vip-helpers/test-wpcom-vip-set-image-quality-for-url.php
@@ -1,0 +1,116 @@
+<?php
+
+namespace Automattic\VIP\Tests;
+
+class WPCOM_VIP_Set_Image_Quality_For_Url_Test extends \WP_UnitTestCase {
+
+	private const ATTACHMENT_URL = 'https://example.com/wp-contents/uploads/2020/08/foo.jpg';
+
+	public function get_data_for_testing_args() {
+		return [
+			'Valid args'                    => [
+				null,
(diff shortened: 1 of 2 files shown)


brandon-m-skinner:
r1069-stacks

PR #1711, opened 2020-08-13
Add Feature flag lib

nickdaugherty:
## Description

Add a feature flag library to gate rollouts of features

## Checklist

## Steps to Test

Currently not used, so test suite is sufficient, but should test CLI + regular requests just to ensure the file is required from `000-vip-init.php` correctly.

### DIFF
--- a/lib/feature/class-feature.php
+++ b/lib/feature/class-feature.php
@@ -18,8 +18,9 @@ class Feature {
 
 		$percentage = static::$feature_percentages[ $feature ];
 
-		// Which bucket is the site in - 100 possibilites, one for each percentage
-		$bucket = static::$site_id % 100;
+		// Which bucket is the site in - 100 possibilites, one for each percentage. We run this through crc32 with
+		// the feature name so that the same sites aren't always the canaries
+		$bucket = crc32( $feature . '-' . static::$site_id ) % 100;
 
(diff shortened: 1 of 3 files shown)


lib/feature/class-feature.php
```diff
@@ -0,0 +1,29 @@
...
+
+	public static function is_enabled_by_percentage( $feature ) {
+		if ( ! isset( static::$feature_percentages[ $feature ] ) ) {
+			return false;
+		}
+
+		$percentage = static::$feature_percentages[ $feature ];
+
+		// Which bucket is the site in - 100 possibilites, one for each percentage
+		$bucket = static::$site_id % 100;
```
netsuso, line 22: In order to avoid the same sites being always the first ones where features are enabled, the feature name can be included to compute the bucket. For example:

`$bucket = crc32( $feature . static::$site_id ) % 100;`

I know this makes tests more difficult to understand, so you can leave this as a suggestion to be implemented if this library ends up being very used.

nickdaugherty: I like this idea, it seems worth implementing now.

nickdaugherty: Added the hash in df7d735

nickdaugherty:
r1073-stacks

PR #1712, opened 2020-08-14
Move `clearstatcache` call into the client class

hanifn:
## Description

It was highlighted that there's potential for the files stats to be inaccurate when uploading files due to PHP's built in file stat cache. We had a call to `clearstatcache()` added but that really only affected new files. This ensures that the stat cache is cleared for _every_ upload.

## Checklist

## Steps to Test

Not really sure how to effectively test this but I basically sandboxed a site and ensured that files are uploaded properly to the file service.


brandon-m-skinner:
r1077-stacks

PR #1718, opened 2020-08-17
Search: Support --network-wide in validate-counts commands

netsuso:
## Description

Support `--network-wide` parameter for `vip-search health validate-counts` commands. If the parameter is provided, these commands will show the stats for all sites of the network, instead of only the main site.

This is not compatible with `--version`, as every subsite can have a different index version.

## Checklist

## Steps to Test

1. Check out PR
1. Execute on a multisite site: `wp vip-search health validate-counts --network-side`
1. Ensure stats for all subsites are shown

### DIFF
--- a/search/includes/classes/commands/class-healthcommand.php
+++ b/search/includes/classes/commands/class-healthcommand.php
@@ -111,6 +111,8 @@ class HealthCommand extends \WPCOM_VIP_CLI_Command {
 				switch_to_blog( $site['blog_id'] );
 
 				$this->validate_indexable_count_for_site( $indexable_slug, $version );
+				
+				restore_current_blog();
 			}
 		} else {
 			WP_CLI::line( "Validating $indexable_slug count\n" );
--- a/search/includes/classes/commands/class-healthcommand.php
+++ b/search/includes/classes/commands/class-healthcommand.php
@@ -47,41 +47,16 @@ class HealthCommand extends \WPCOM_VIP_CLI_Command {
 	 * [--version=<int>]
 	 * : Index version to validate - defaults to all
 	 *
+	 * [--network-wide]
+	 * : Validate all sites in a multisite network
+	 *
 	 * ## EXAMPLES
 	 *     wp vip-es health validate-users-count
 	 *
 	 * @subcommand validate-users-count
 	 */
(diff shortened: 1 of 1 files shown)


search/includes/classes/commands/class-healthcommand.php
```diff
@@ -47,82 +47,124 @@ public function validate_counts( $args, $assoc_args ) {
...
+				}
 
-			if ( is_wp_error( $users_results ) ) {
-				WP_CLI::warning( $users_results->get_error_message() );
-				return;
+				WP_CLI::line( "\nValidating $indexable_slug count for site " . $site['blog_id'] . ' (' . $site['domain'] . $site['path'] . ')\n' );
+
+				switch_to_blog( $site['blog_id'] );
+
+				$this->validate_indexable_count_for_site( $indexable_slug, $version );
```
nickdaugherty, line 113: We should also call `restore_current_blog()` after this, to "reset" it.

search/includes/classes/commands/class-healthcommand.php
```diff
@@ -47,82 +47,124 @@ public function validate_counts( $args, $assoc_args ) {
...
-			$versions[] = intval( $assoc_args['version'] );
+			$version = intval( $assoc_args['version'] );
 		} else {
-			// Defaults to all versions
-			$indexable = \ElasticPress\Indexables::factory()->get( 'user' );
+			$version = null;
+		}
 
-			$version_objects = $search->versioning->get_versions( $indexable );
+		if ( isset( $assoc_args['network-wide'] ) && is_multisite() && $search->is_network_mode() ) {
```
nickdaugherty, line 96: We don't actually need the `$search->is_network_mode()` check - we want to be able to re-index all sites even if we're not using EP's "network mode", which allows sites to cross-search eachother.
```suggestion
		if ( isset( $assoc_args['network-wide'] ) && is_multisite() ) {
```

search/includes/classes/commands/class-healthcommand.php
nickdaugherty, line 113:
```suggestion
				$this->validate_indexable_count_for_site( $indexable_slug, $version );
				
				restore_current_blog();
```

nickdaugherty:
r1074-stacks

PR #1722, opened 2020-08-18
Don't set sites as staging in Jetpack regardless of env type

brandon-m-skinner:
## Description

VIP sites should not be staging in Jetpack by default.


brandon-m-skinner:
r1079-stacks

PR #1723, opened 2020-08-19
Implement a simple async scheduler to add a cron event to be executed some time later.

rinatkhaziev:
## Description

Currently, we lack the ability to quickly schedule an arbitrary long-running command to run across the environments.

This PR provides this missing piece. 

NB: Cron Control is a hard dependency for this, mostly for consistency and minifying the chances of any object cache related surprises. 

#### To schedule a command:

```
wp vip cmd-scheduler schedule --cmd='wp some command-to-run --with-argument' --when=now
```

`--cmd` is the command to run.

`--when` accepts either a unix timestamp or 'now'

#### To check the status of the command

```
wp vip cmd-scheduler check-status --cmd='wp some command-to-run --with-argument'
```
=> 

```
Success: Found event, status: pending; created: 2020-08-19 02:51:59; last modified: 2020-08-19 02:51:59
```

## Checklist

## Steps to Test

Example:

1. Check out PR.
1. Schedule and check some commands as per "example usage" section.


wp-cli/class-vip-go-async-scheduler-command.php
```diff
@@ -0,0 +1,183 @@
...
+	 * @return array 
+	 */
+	public static function get_commands( string $command ): array {
+		global $wpdb;
+
+		$command      = self::normalize_command( $command );
+		$events_store = Events_Store::instance();
+
+		// phpcs:ignore 
+		$events = $wpdb->get_results( $wpdb->prepare( "SELECT * FROM {$events_store->get_table_name()} WHERE action = %s AND args = %s LIMIT 10", self::SCHEDULE_EVENT_KEY, maybe_serialize( [ $command ] ) ) );
```
netsuso, line 138: Would it make sense to move this to the Events_Store class? Just to keep the single responsibility. I know there are already a few methods there to get jobs, and none of them fits here, but even if we have to add another one, at least it might get reused (or refactored) in the future.

wp-cli/class-vip-go-async-scheduler-command.php
```diff
@@ -0,0 +1,183 @@
+<?php
+/**
+ * Thin helper to assist with scheduling long-running CLI commands asynchronously.
+ */
+
+use \Automattic\WP\Cron_Control\Events_Store;
+
+class VIP_Go_Async_Scheduler_Command extends WPCOM_VIP_CLI_Command {
```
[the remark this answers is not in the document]
rinatkhaziev: Yeah I was conflicted on that one. Personally I prefer PSR over the classic WP Convention too, but decided to conform to what's in repo. Will make the change.

netsuso (commented):
Great idea, this will be really helpful for the mass reindexing!

I just don't get the full idea about storing the timestamp in the object cache. If I've understood correctly, this is the workflow (I've not tested it, this is just by looking at the code):

- A command is scheduled (for the first time). The cache key for this command is set to the scheduled timestamp
- If the same command is attempted to be scheduled again, it will fail because there's another execution pending. But this is checked against the DB, so the cache is not useful here (maybe because we don't trust the cached value to exist)
- If we run a check, it will retrieve all events from DB and the cached key:
    - If the cached key exists, it will use its value to find an event with that timestamp. But this will be simply the only pending/running execution, as we are rejecting multiple scheduled executions. So the cached key doesn't give us a lot of information here.
    - If the key doesn't exist, it will show all previous executions stating that "an exact match couldn't be found", which is confusing becase these previous executions are from the exact same command
        - If the key was removed by wp_cache_delete after the command finished running, this will happen when there are no pending/running executions, which is an expected situation.
        - If the key was evicted from cache, there might be a pending/running execution (and only one), which is what should be shown, instead of the list of all previous executions.

wp-cli/class-vip-go-async-scheduler-command.php
rinatkhaziev: Just wanted to limit the scope of changes and, since it's more of a temporary thing, introducing potential cruft in the Cron Control didn't seem worth it.

netsuso: If it's something temporary, then it certainly makes sense :)

rinatkhaziev:
@netsuso 

> I just don't get the full idea about storing the timestamp in the object cache. If I've understood correctly, this is the workflow (I've not tested it, this is just by looking at the code):

Frankly speaking, that may have been just over-engineering on my part 😅 

The idea was to pinpoint the exact event that happened at a specific time. It's possible that there will be events that are already completed but they weren't cleared from the events table yet, so it's possible there will be multiple events for the same command.

I've reworked some of the confusing logic plus introduced the `--when` argument for the `check-status` command. Additionally, the results are printed as a table now, which looks neat and less confusing:

```
Success: Found matching events
+----+---------+---------------------+---------------------+------------+-----------------+
| ID | status  | created             | modified            | timestamp  | timestamp_match |
+----+---------+---------------------+---------------------+------------+-----------------+
| 49 | pending | 2020-08-20 22:46:43 | 2020-08-20 22:46:43 | 1597963604 | yes             |
+----+---------+---------------------+---------------------+------------+-----------------+
```

rinatkhaziev:
r1086-stacks

PR #1727, opened 2020-08-20
Disable above compat admin notice for EP

brandon-m-skinner:
## Description

Once we go over Elasticsearch 7.5, admin notices about being over the compatible version will show.

We should hide these since we should never be above the version we support since we control the full stack.

## Checklist

## Steps to Test

1. Open wp-admin in Lando. You should see the above compat admin notice.
2. Apply PR.
3. The notice should no longer appear.


nickdaugherty:
Works great

brandon-m-skinner:
r1082-stacks

PR #1728, opened 2020-08-21
Add index version alias utilities

nickdaugherty:
## Description

Adds the ability to target index versions using the aliases `next` and `previous`. This will be most useful when doing `--network-wide` operations on multisites that have subsites with different index versions.

This code is not yet used anywhere, but will be utilized in a subsequent PR.

## Checklist

## Steps to Test

Code is not yet active anywhere, so no manual testing needed

### DIFF
--- a/tests/search/includes/classes/test-class-versioning.php
+++ b/tests/search/includes/classes/test-class-versioning.php
@@ -279,6 +279,182 @@ class Versioning_Test extends \WP_UnitTestCase {
 		$this->assertEquals( $expected_inactive_versions, $inactive_versions );
 	}
 
+	public function normalize_version_number_data() {
+		return array(
+			// Regular, 'next'
+			array(
+				// Input array of versions
+				array(
+					1 => array( 
+						'number' => 1,
+						'active' => false,
+					),
+					2 => array(
+						'number' => 2,
+						'active' => true,
+					),
+					3 => array(
+						'number' => 3,
+						'active' => false,
+					),
+				),
+				// Indexable slug
+				'post',
+				// Version string to be normalized
+				'next',
+				// Expected normalized version number
+				3,
+			),
+
+			// Regular, 'previous'
+			array(
+				// Input array of versions
+				array(
+					1 => array( 
+						'number' => 1,
+						'active' => false,
+					),
+					2 => array(
+						'number' => 2,
+						'active' => true,
+					),
+					3 => array(
+						'number' => 3,
+						'active' => false,
+					),
+				),
+				// Indexable slug
+				'post',
+				// Version string to be normalized
+				'previous',
+				// Expected normalized version number
+				1,
+			),
+
+			// Regular, 'active'
+			array(
+				// Input array of versions
+				array(
+					1 => array( 
+						'number' => 1,
+						'active' => false,
+					),
+					2 => array(
+						'number' => 2,
+						'active' => true,
+					),
+					3 => array(
+						'number' => 3,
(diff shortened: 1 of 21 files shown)


tests/search/includes/classes/test-class-versioning.php
```diff
@@ -514,6 +690,82 @@ public function activate_version_data() {
...
+					),
+					4 => array(
+						'number' => 4,
+						'active' => false,
+					),
+				),
+				// Indexable slug
+				'post',
+				// Version to activate
+				'next',
```
netsuso, line 714: This test and the one below are failing because the version number is not normalized. Maybe it's not worth doing it in `activate_version()` (leaving normalization to upper layers), but it has to be done for sure in `test_activate_version()`, because apart from checking the expected new versions array, it also checks that the actually activated version (which is always a number) matches the version we want (and here it's not normalized):

```
$version_to_activate = self::$version_instance->normalize_version_number( $indexable, $version_to_activate );
```

tests/search/includes/classes/test-class-versioning.php
```diff
@@ -279,6 +279,182 @@ public function test_get_inactive_versions( $versions, $indexable_slug, $expecte
...
+					),
+					3 => array(
+						'number' => 3,
+						'active' => false,
+					),
+				),
+				// Indexable slug
+				'post',
+				// Version string to be normalized
+				'active',
```
netsuso, line 417: This one is failing because `get_active_version_number()` returns `1` by default if it doesn't find any version with the active flag.

I understand that `1` is a safety net when there are no versions created, but if there are existing versions and none of them is flagged as "active", then it should return null. For `get_active_version_number()`:

```
        if ( ! $active_version ) {
            $versions = $this->get_versions( $indexable );
            if ( count( $versions ) ) {
                return null;
            } else {
                return 1;
            }
        }
```

(This is not optimal, as it's requesting the full list of versions again, but it was the minimal change I needed to make tests work)

nickdaugherty: Crud, I actually reverted my tests for this last night (since this PR doesn't actually make use of the normalization stuff yet), but I guess I didn't push that commit 😞

nickdaugherty: Rather than returning null, IMO it should return _something_ - perhaps the lowest index version that exists in ES. We need to have some automated way to recover from corruption or deletion of the index versions option. My thought was when that is detected, we set a lock (so this only happens once) then query ES for all indexes, parse out the indexable + version from the index names, then recreate the option and mark the lowest number as active. 

We could handle when no index is marked as active in the same way, by using the lowest (first) index version. What do you think about that idea?

netsuso: I think that makes sense, it all depends on what are the odds for corruption on a wp_option. If that's something that can happen (albeit not frequently), then it's worth asking ES about it, it should be a fast query and you can then store the result in the option, which would be used in the following queries.

tests/search/includes/classes/test-class-versioning.php
nickdaugherty: Ok I've fixed up the tests - once we actually integrate this into the other functions, we'll have to come back to these.

nickdaugherty:
Changes made, ready for another look.

netsuso (approved):
:ship:

nickdaugherty:
r1126-stacks

PR #1731, opened 2020-08-24
Updates VIP GO Privacy Exports for 5.4

emrikol:
## Description

WordPress 5.4 changed how privacy exports managed the post meta data for generated ZIP files.  This was done in https://core.trac.wordpress.org/changeset/48127 and it broke VIP Go's overrides.

This PR should bring the VIP Go functionality up to date with the changes in the core functionality.

## Checklist

## Steps to Test

Example:

1. Visit a site's export `/wp-admin/export-personal-data.php` on a WP 5.4 or higher install
1. Try to export data, and see it fails.
1. Patch
1. Try again and see if a ZIP is downloaded.

### DIFF
--- a/001-core/privacy.php
+++ b/001-core/privacy.php
@@ -47,6 +47,10 @@ function init_privacy_compat_cleanup() {
 function generate_personal_data_export_file( $request_id ) {
 	$request = wp_get_user_request_data( $request_id );
 
+	// Get the export file URL.
+	$exports_url      = wp_privacy_exports_url();
+	$export_file_name = get_post_meta( $request_id, '_export_file_name', true );
+
 	if ( ! $request || 'export_personal_data' !== $request->action_name ) {
 		wp_send_json_error( __( 'Invalid request ID when generating export file.' ) );
 	}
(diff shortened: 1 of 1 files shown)


mjangda:
We should probably get this merged and deployed but wanted to note that our hackaround can probably be removed completely thanks to the `vip://` stream wrapper and because we now have the zip extension enabled by default

emrikol:
> We should probably get this merged and deployed but wanted to note that our hackaround can probably be removed completely thanks to the `vip://` stream wrapper and because we now have the zip extension enabled by default

I've just checked by commenting out the core privacy mu-plugin and it looks like it's not working on 5.5:

```
An error occurred while attempting to export personal data.

Unable to add data to HTML file.
```

image

emrikol:
The builds seem to be failing due to coding standard errors in the core functionality.  Happy to fix it if we need to, but I wonder if diverging from the core function more may make it harder in the future to diff and update?

mjangda:
I think we can ignore those lint errors.

rinatkhaziev:
@emrikol @mjangda Maybe instead of ignoring them ourselves, we can instruct phpcs to ignore them? It seems there's only 1 instance. Just putting `// //phpcs:disable WordPress.Classes.ClassInstantiation.MissingParenthesis` at the top of the file should do.

emrikol:
Since it's only one line, how about we just add a `// phpcs:ignore` to that one spot?

rinatkhaziev:
@emrikol that should work too.

rinatkhaziev:
r1088-stacks

PR #1732, opened 2020-08-24
Hotfix for async scheduler command: don't use wp_schedule_single_event, instead use Cron Control's schedule_event

rinatkhaziev:
## Description

This is a follow-up to #1723 

As it turns out, `wp_schedule_single_event` doesn't actually schedule properly with Cron Control on our infrastructure (It does so in local dev). 

## Checklist

## Steps to Test

Example:

1. Check out PR.
1. Schedule the command `wp vip cmd-scheduler check-status --cmd="my cmd to-run" --when="1598297946"
1. `wp cron-control event list` to make sure it got scheduled.
1. Rejoice.


rinatkhaziev:
The bug actually didn't exist.

PR #1735, opened 2020-08-26
Search: Add Health::reconcile_diff method to allow for the correction of index inconsistencies.

rinatkhaziev:
## Description

Currently, the only way to address index inconsistencies is to re-index the content. Re-indexing also doesn't resolve the issue where something is in the indexed, but the object was deleted in a way that our hooks didn't catch.

This PR addresses that by introducing a new static method Health::reconcile_diff which addresses a batch on inconsistencies (either add to the index queue or delete from the index.

By default, try to correct the inconsistencies, but allow to skip that step via `--do-not-heal` flag.

Here's how it looks:

First run:
```
$ wp vip-search health validate-contents
Validating posts 1 - 500...❌
Warning: Inconsistencies found!
Reconciling the differences!
type,id,issue
post,1,extra_in_index
```

Second run:
```
$ wp vip-search health validate-contents 
Validating posts 1 - 500...✅
Success: No inconsistencies found!
```

## Checklist

## Steps to Test

1. Check out PR.
1. Add or remove a post directly in the DB (i.e. not using any core methods)
1. Run `wp vip-search health validate-contents --heal` and verify that inconsistencies are indeed found
1. Run the queue via `wp cron event run vip_search_queue_processor`
1. Run `wp vip-search health validate-contents --heal` again and verify that there are no inconsistencies anymore.

### DIFF
--- a/search/includes/classes/class-health.php
+++ b/search/includes/classes/class-health.php
@@ -196,7 +196,7 @@ class Health {
 	 *
 	 * @return array Array containing counts and ids of posts with inconsistent content
 	 */
-	public static function validate_index_posts_content( $start_post_id = 1, $last_post_id = null, $batch_size, $max_diff_size, $silent, $inspect = false ) {
+	public static function validate_index_posts_content( $start_post_id = 1, $last_post_id = null, $batch_size, $max_diff_size, $silent, $inspect = false, $do_not_heal = false ) {
 		// If batch size value NOT a numeric value over 0 but less than or equal to PHP_INT_MAX, reset to default
 		//     Otherwise, turn it into an int
 		if ( ! is_numeric( $batch_size ) || 0 >= $batch_size || $batch_size > PHP_INT_MAX ) {
@@ -253,6 +253,8 @@ class Health {
 
 			if ( is_wp_error( $result ) ) {
 				$result['errors'] = array( sprintf( 'batch %d - %d (entity: %s) error: %s', $start_post_id, $next_batch_post_id - 1, $indexable->slug, $result->get_error_message() ) );
+			} elseif ( count( $result ) && ! $do_not_heal ) {
+					self::reconcile_diff( $result );
 			}
 
 			$results = array_merge( $results, $result );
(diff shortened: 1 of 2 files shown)


search/includes/classes/class-health.php
```diff
@@ -460,6 +460,28 @@ public static function diff_document_and_prepared_document( $document, $prepared
...
+	 *
+	 * @param array $diff array of inconsistenices in the following shape: [ id => string, type => string (Indexable), issue => <missing_from_index|extra_in_index> ].
+	 */
+	public static function reconcile_diff( array $diff ) {
+		foreach ( $diff as $key => $obj_to_reconcile ) {
+			switch ( $obj_to_reconcile['issue'] ) {
+				case 'missing_from_index':
+					\Automattic\VIP\Search\Search::instance()->queue->queue_object( $obj_to_reconcile['id'], $obj_to_reconcile['type'] );
+					break;
+				case 'extra_in_index':
```
rinatkhaziev, line 478: @nickdaugherty @brandon-m-skinner did I miss the "inconsistent" case, which may happen when `--inspect` is passed.

rinatkhaziev:
The more I look at it, the call to `reconcile_diff` should be made inside `Health::validate_index_posts_content()` instead of CLI itself. The reason being CLI outputs the results AFTER the whole thing is done, and we should try to heal as quickly as possible - the command might die, leaving us with nothing. 

So I'm thinking adding **7th** argument to `validate_index_posts_content()` so that the signature would become:

```
public static function validate_index_posts_content( $start_post_id = 1, $last_post_id = null, $batch_size, $max_diff_size, $silent, $inspect = false, $should_heal = false )
```

Or, more ambitious, just try to do the healing by default. The reason for that is that validation alone has limited usefulness. 
"Ok, I have those inconsistent posts, and I have the CSV of them, and now I have to go and manually add those ids into indexing" would be reduced "Ok, these were inconsistent posts that were corrected".

On a second run, if there are inconsistencies still that might indicate a problem, but it's going to be easier to inspect and correct.

search/includes/classes/class-health.php
brandon-m-skinner: @rinatkhaziev I think so: https://github.com/Automattic/vip-go-mu-plugins/blob/ac59632bb576b7962fc27e2cfaa46342c0393d3c/search/includes/classes/class-health.php#L319-L333

Inconsistent should be treated like missing_from_index since it just needs a reindex to be fixed.

brandon-m-skinner:
I think healing by default is fine as long as we record the inconsistencies.

rinatkhaziev:
Ok, switching the healing to in-batch (as opposed to after everything is validated), accounted for "inconsistent" and switched from emoji to ASCII counterparts to account for TTYs not supporting emoji. Namely `screen`. Less colorful now, but more reliable in usefulness.

rinatkhaziev:
r1095-stacks

PR #1736, opened 2020-08-26
Features: Use a site_id() function instead of defaulting to FILES_CLIENT_SITE_ID

### LINKED ISSUE #1730: Undefined constant FILES_CLIENT_SITE_ID in lib/feature/class-feature.php
It has been reported that `lib/feature/class-feature.php#L8` is responsible for throwing a warning for `Use of undefined constant FILES_CLIENT_SITE_ID`.

https://github.com/Automattic/vip-go-mu-plugins/blob/8fc26811faef8ede5032359ca85fbf67d4227cd4/lib/feature/class-feature.php#L8

Could we update this to account for that constant not being set in a local development context?

cc: @nickdaugherty since it looks like you worked on this.

nickdaugherty:
## Description

The `FILES_CLIENT_SITE_ID` constant is not necessarily defined, such as in custom local dev environments, so can't use it directly...needs to be wrapped.

Fixes #1730

## Checklist

## Steps to Test

1. Check out PR.
1. Remove any reference to `define( 'FILES_CLIENT_SITE_ID', ... )` anywhere
1. Run a check in the theme's functions.php - `\Automattic\VIP\Feature::is_enabled( 'foo' )` - this should not throw any undefined constant warnings

### DIFF
--- a/lib/feature/class-feature.php
+++ b/lib/feature/class-feature.php
@@ -5,7 +5,7 @@ namespace Automattic\VIP;
 class Feature {
 	public static $feature_percentages = array();
 
-	public static $site_id = FILES_CLIENT_SITE_ID;
+	public static $site_id = false;
 
 	public static function is_enabled( $feature ) {
 		return static::is_enabled_by_percentage( $feature );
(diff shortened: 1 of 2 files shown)


lib/feature/class-feature.php
```diff
@@ -20,11 +20,19 @@ public static function is_enabled_by_percentage( $feature ) {
...
+		$bucket = crc32( $feature . '-' . static::site_id() ) % 100;
 
 		// Is the bucket enabled?
 		$threshold = $percentage * 100; // $percentage is decimal
 
 		return $bucket < $threshold; // If our 0-based bucket is inside our threshold, it's enabled
 	}
+
+	public static function site_id() {
+		if ( static::$site_id ) {
```
netsuso, line 32:
```suggestion
		// This is used in tests, so the site id can be easily changed
		if ( static::$site_id ) {
```

netsuso (dismissed):
Just added a comment to clarify why `static::$site_id` is needed

netsuso:
stacks-r1089

PR #1737, opened 2020-08-27
Update two-factor to 0.7.0

rinatkhaziev:
Version 0.7.0

[Release notes](https://github.com/WordPress/two-factor/releases/tag/0.7.0
)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,8 +6,8 @@ workflows:
       - php73-build-singlesite
       - php74-build-multisite
       - php74-build-singlesite
-      - php73-build-multisite-nightly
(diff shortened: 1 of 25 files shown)


brandon-m-skinner:
Tested on Lando. TOTP works and email works.

rinatkhaziev:
Noting that this probably needs a bit more testing, because it's such an integral part.

mjangda:
> Noting that this probably needs a bit more testing, because it's such an integral part.

I've tested various pathways and all looks good on my end.

🚢

rinatkhaziev:
r1161-stacks

PR #1741, opened 2020-08-27
Fix file stats cache reset on upload

mjangda:
We store the path in the cache without the leading slash but the API returns the path with it. We need to remove the leading slash, otherwise the cache doesn't get cleared and we end up with stale data (e.g. filesize is incorrect).

## Checklist

## Steps to Test

1. `wp shell`
1. `file_put_contents( 'vip://wp-content/uploads/yay1.txt', 'hello' )`
1. `filesize( 'vip://wp-content/uploads/yay1.txt' )` <= should return `5`; without the fix, returns `0`.

### DIFF
--- a/files/class-api-client.php
+++ b/files/class-api-client.php
@@ -145,8 +145,9 @@ class API_Client {
 		// save to cache
 		$this->cache->copy_to_cache( $response_data->filename, $local_path );
 
-		// reset file stats cache if any
-		$this->cache->remove_stats( $response_data->filename );
+		// Reset file stats cache, if any.
+		// Note: the ltrim is because we store the path without the leading slash but the API returns the path with it.
+		$this->cache->remove_stats( ltrim( $response_data->filename, '/' ) );
 
 		return $response_data->filename;
 	}
--- a/tests/files/test-api-client.php
+++ b/tests/files/test-api-client.php
@@ -486,9 +486,22 @@ class API_Client_Test extends \WP_UnitTestCase {
 		$file_path = __DIR__ . '/../fixtures/files/upload.jpg';
 		$upload_path = '/wp-content/uploads/file.txt';
 
+		$cache = self::get_property( $this->api_client, 'cache' )->getValue( $this->api_client );
+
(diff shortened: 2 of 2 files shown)


nickdaugherty (dismissed):

mjangda:
r1092-stacks

files/class-api-client.php
```diff
@@ -145,8 +145,9 @@ public function upload_file( $local_path, $upload_path ) {
 		// save to cache
 		$this->cache->copy_to_cache( $response_data->filename, $local_path );
 
-		// reset file stats cache if any
-		$this->cache->remove_stats( $response_data->filename );
+		// Reset file stats cache, if any.
+		// Note: the ltrim is because we store the path without the leading slash but the API returns the path with it.
+		$this->cache->remove_stats( ltrim( $response_data->filename, '/' ) );
```
netsuso, line 150: Maybe it could have been easier to just use the `$upload_path`. In the end, that's what we generate internally in the wrapper for everything, so it's better that the cache matches it exactly.

files/class-api-client.php
```diff
@@ -145,8 +145,9 @@ public function upload_file( $local_path, $upload_path ) {
 		// save to cache
 		$this->cache->copy_to_cache( $response_data->filename, $local_path );
```
netsuso, line 146: We should also change this here, as we are also copying the file to the wrong cache key. Not as critical as the file stats cache, but at least that the cached file can be used.

netsuso (commented):
Sorry, I didn't notice this PR and I was also looking into it... I'm going to update the BB8

Thanks for solving it!  I've left a couple of comments.

files/class-api-client.php
mjangda: Good thinking! Definitely makes sense to do here.

I'm wondering thought if we should just account for the leading slash in the Cache class because the same problem is bound to come up again.

files/class-api-client.php
mjangda: Oh, yes, that would definitely be much cleaner :)

PR #1743, opened 2020-08-31
Adjust environment type to the ones allowed in WP 5.5.1

netsuso:
## Description

WordPress 5.5.1 [won't allow to set custom environment types](https://make.wordpress.org/core/2020/08/27/wordpress-environment-types/), so our previous code to set it matching the VIP environment name won't work.

This PR recovers a previous implementation ( #1703 ) where we map "production" and "develop" VIP environments to "production" and "development" environment types, assigning the "staging" type to any other.

## Checklist

## Steps to Test

1. Check out PR.
1. Execute `wp_get_environment_type()` in a wp shell
1. For production sites it should return "production", for develop sites it should return "development", and for the rest "staging"

### DIFF
--- a/000-vip-init.php
+++ b/000-vip-init.php
@@ -201,21 +201,23 @@ if ( ( defined( 'USE_VIP_ELASTICSEARCH' ) && USE_VIP_ELASTICSEARCH ) || // legac
 	}
 }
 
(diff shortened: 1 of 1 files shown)


netsuso:
stacks-r1097

PR #1749, opened 2020-09-05
Support multiple Jetpack versions

netsuso:
## Description

Keep multiple jetpack versions inside mu-plugins and allow to pin to one of them with a new define `VIP_JETPACK_PINNED_VERSION`.

Although it supports any value, as long as it matches the directory name, we are going to base it on minor versions, not including patches. The currently available ones are `8.8` and `8.9`. Support for next/beta versions is still work in progress.

You can also skip loading Jetpack altogether by defining `VIP_JETPACK_SKIP_LOAD` to true.

## Reviewing changes

As full Jetpack 8.8.2 and 8.9.1 have been committed, you can check this commit for the actual code changes: https://github.com/Automattic/vip-go-mu-plugins/pull/1749/commits/792bd65a22d052092b62a93f5eb05d5a63f632cf

## Checklist

- (pending) This change has relevant documentation additions / updates (if applicable).

## Steps to Test

1. Check out PR.
1. Check in the Jetpack dashboard that the active version is 8.9.1
1. Add this define: `define( 'VIP_JETPACK_PINNED_VERSION', '8.8' );`
1. Check in the Jetpack dashboard that the active version is 8.8.2
1. Add this define: `define( 'VIP_JETPACK_SKIP_LOAD', true );`
1. Check that the Jetpack dashboard is not accessible
1. Add this define: `define( 'VIP_JETPACK_PINNED_VERSION', '9.5' );`
1. Check in the Jetpack dashboard that the active version is 8.9.1 (current default version)


mjangda:
Thanks for getting this started! Before we get too deep into the implementation, It would be good to plan out how our upgrade process changes with this and how to handle exceptions. For example, do we release minor releases as separate folders or do they go into the major folder? What about betas? When do we remove old versions? What happens when a site is pinned to a version that is removed?

I would also consider moving away from submodules. Adding and removing them is a huge pain and I think we'd be much better off just downloading a built release and committing that (similar to how we handle VaultPress and other plugins). I think that will be easier to automate as well. Open to arguments against this though.

netsuso:
Sure! I just wanted to test this implementation before publishing the PR and opening it to discussion. There are certainly several options available.

Regarding the minor/major versions, I also thought about the option of keeping a major version folder (e.g., 8.7, 8.8, 8.9) and then minor versions would be pushed into their corresponding major version. But then I realized that one of the reasons for not upgrading a site is that the customer is risk averse and doesn't want any update at all, even a minor release.But maybe this is a extreme case and it's not worth keeping so many versions just for this case.

About how and when to remove old versions, that's an important point and not easy to solve. We need to track all sites that are pinned to an old version (Have I heard about site indexing?). I suppose we can support two major versions behind the current one, and this would allow to set deprecation dates to customers.

Regarding submodules, I don't have a strong opinion about them. In general I find them uncomfortable to work with, but at the same time they save a lot of space in the repo. If we are going to add 30-40 versions a year, committing them directly to mu-plugins would mean ~300 MB added every year to the repository. That's 300 MB you will need to download on every clone, and there's no way to strip them out once they have been added, even if you remove old versions. The good thing about this is that we can choose any option and even change it in the future, as it won't affect the implementation.

netsuso:
We also have the fixed plugin version as defined in the header of `jetpack.php`. Maybe we could just remove it, and although the Jetpack version wouldn't be shown in the plugin list, at least it would not lie if it's pinned to an older one.

netsuso:
Some thoughts on submodules after internal discussion:
- Using submodules with numeric versioned folders is discarded, as adding and removing submodules frequently is not nice, because both operations require manual intervention from developers.
- Git stores a file blob as an object that is referenced by a hash of its contents. So if you duplicate a file it doesn't duplicate the disk space. It means that if we add several Jetpack versions without submodules, as they will share most of the files, there will be little overhead and the git repository size won't increase too much.

Regarding the options for the directory structure:
1. Separate directory for every version: This way we would have `jetpack-8.8`, `jetpack-8.8.1`, `jetpack-8.9`, `jetpack-8.9.1`. PInned sites would point to a specific version, and would never be upgraded while pinned.
1. Variant of the previous one: Keeping a directory just for major versions: `jetpack-8.7`, `jetpack-8.8`, `jetpack-8.9`. Minor versions (normally bugfixes or security upgrades) would be upgraded automatically even for pinned sites (they would be pinned to major versions, not minors)
1. To reduce the number of directories that we add and remove, and to allow using submodules, we can have aliased versions: `jetpack-current`, `jetpack-previous1`, `jetpack-previous2` (and also `jetpack-next` for testing). These would match major versions, as in option 2, and there would be a mapping in the loader code ('current' => '8.9', 'previous1' => '8.8', 'previous2' => '8.7'), so sites would still be pinned to specific major versions (this will be very helpful when we have the site detail index, as it's better to know a site is pinned to '8.8' than to 'previous1'). When a new major version is released, the mappings will be shifted ('current' => '8.10', 'previous1' => '8.9', 'previous2' => '8.8')

For all cases, we need to define how much time can a site be pinned to an old version, and decide what happens when an old version is removed and it's still in use (maybe set it to the lowest available version?). For option 3 the amount of time would be determined by the number of previous versions we would keep. For option 1 or 2 it would be more flexible.

nickdaugherty:
The consensus is that option 2 is the way forward.

jetpack.php
```diff
@@ -12,6 +12,10 @@
  * Domain Path: /languages/
  */
 
+if ( ! defined( 'VIP_JETPACK_DEFAULT_VERSION' ) ) {
+	define( 'VIP_JETPACK_DEFAULT_VERSION', '8.9' );
```
netsuso: As there won't exist a `jetpack` directory, we need to set the default version here. This will need to be updated on every upgrade to a new minor version.

nickdaugherty:
@netsuso the size of the diff is making it difficult to review. What do you think about committing the `jetpack-x.x` directories in a different PR(s)? That will also keep the deploy small (Jetpack is quite large and takes a long time to commit in SVN) and get everything in place for a small switch-over PR.

Screen Shot 2020-09-30 at 3 47 50 PM

netsuso:
@nickdaugherty I was about to do it, but then I pointed out in the description that you could just review the latest commit: https://github.com/Automattic/vip-go-mu-plugins/commit/792bd65a22d052092b62a93f5eb05d5a63f632cf

But it will be probably better to just split it to avoid dealing with a so large review. Stay tuned!

netsuso:
This has been splitted in 3+1 PRs:

- #1787 (just the code)
- #1788 (jetpack-8.9)
- #1789 (jetpack-8.8)
- #1790 (jetpack-9.0)

PR #1754, opened 2020-09-11
Search: Add --network-wide support to index version commands

nickdaugherty:
## Description

Adds `--network-wide` support to the `wp vip-search index-versions` `add` and `activate` commands, for working with index versions on multisites. This lets us create a new index version, index the content, then activate it across all subsites.

```
wp vip-search index-versions add post --network-wide
wp vip-search index --network-wide --version=next --indexables=post
wp vip-search index-versions activate post next --network-wide
```

There are other commands that don't yet support `--network-wide` (`delete`, `get`, `list`, `get-active`), but those are lower priority so I don't want to hold up this PR for them.

## Deployment

This makes a change to how index versions are stored - instead of using a site-wide option on multisites in network mode, each index version list is stored per-site, in regular blog options.

This will affect any production multisites sites that have multiple index versions, so we need to check for those and copy over the option manually there.

## Checklist

## Steps to Test

1. Pull down PR
1. Make sure your local environment is a multisite - `lando setup-multisite` if not, then add a site or two via WP Admin
1. `lando wp vip-search index-versions add post --network-wide` to add a new index version on all subsites
1. `lando wp vip-search index-versions list --url=<subsite_url>` to verify the versions exist
1. Index content on all subsites with `lando wp vip-search index --network-wide --version=next --indexables=post`
1. Activate the new indexes with `lando wp vip-search index-versions activate post next --network-wide`
1. Verify that the new index versions are active with `lando wp vip-search index-versions list --url=<subsite_url>`
1. Try the same steps again, but without `--network-wide` (so just single site) to ensure it works as expected
1. Then can add another new version and see how the `next` alias makes it work regardless of the current version for a specific site (site 1 can be on version 2, while site 2 can be on version 3, for example).

### DIFF
--- a/search/includes/classes/commands/class-corecommand.php
+++ b/search/includes/classes/commands/class-corecommand.php
@@ -16,7 +16,7 @@ class CoreCommand extends \ElasticPress\Command {
 
 	protected function _maybe_setup_index_version( &$assoc_args ) {
 		if ( $assoc_args['version'] ) {
-			$desired_version_number = $assoc_args['version'];
+			$version_number = $assoc_args['version'];
 
 			// If version is specified, the indexable must also be specified, as different indexables can have different versions
 			if ( ! isset( $assoc_args['indexables'] ) ) {
@@ -35,13 +35,7 @@ class CoreCommand extends \ElasticPress\Command {
 					return WP_CLI::error( sprintf( 'Indexable %s not found - is the feature active?' ) );
 				}
 
-				$new_version_number = $search->versioning->normalize_version_number( $indexable, $desired_version_number );
-
-				if ( is_wp_error( $new_version_number ) ) {
-					return WP_CLI::error( sprintf( 'Index version %s is not valid: %s', $desired_version_number, $new_version_number->get_error_message() ) );
-				}
-
-				$result = $search->versioning->set_current_version_number( $indexable, $new_version_number );
+				$result = $search->versioning->set_current_version_number( $indexable, $version_number );
 
 				if ( is_wp_error( $result ) ) {
 					return WP_CLI::error( sprintf( 'Error setting version number: %s', $result->get_error_message() ) );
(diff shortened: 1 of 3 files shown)


search/includes/classes/commands/class-corecommand.php
```diff
@@ -40,7 +35,13 @@ protected function _maybe_setup_index_version( &$assoc_args ) {
 					return WP_CLI::error( sprintf( 'Indexable %s not found - is the feature active?' ) );
 				}
 
-				$result = $search->versioning->set_current_version_number( $indexable, $version );
+				$new_version_number = $search->versioning->normalize_version_number( $indexable, $desired_version_number );
```
netsuso, line 38: This is normalizing the version only for the main site, so if `next` is 2 for the main site, it will index version 2 for all sites. It should be done inside the `foreach ( $sites as $site )` loop. The problem is that loop is in `Command::index()` and I guess you were trying to avoid tampering with versions in that file... 😕

netsuso (changes requested):
I love having `add`/`activate` commands with `--network-wide`!  As a suggestion for a future PR, once we have support for the, it would be great to have also `--network-wide` support for the `list` command :)

search/includes/classes/commands/class-corecommand.php
nickdaugherty: You're totally right, whoops. Good catch!

nickdaugherty: @netsuso I believe I have addressed this in ad99b5200, do you mind checking that? Basically what I did was _not_ normalize the index version here and inside Versioning::set_current_version_number(), so that we can store the alias there (`next`) and dynamically translate that to an actual version number at runtime.

I also remove the conditional that used `get_site_option()` in network mode - that will affect prod multisites that have index versions, so we'll need to proceed carefully for deployment.

netsuso: Fixed the `index --network-wide` command, thanks! I'm glad `current_index_version_by_type` existed, it made things easier than I expected 😊

As for the `get_site_option()`, I think I'll wait until your morning to deploy this, as I'm not 100% sure about its implications for multisites and what can happen if no version option is available. IIRC it returns 1 by default, but that could be a problem if another version had been previously generated... Although I suppose we would have noticed earlier in this case.

netsuso (approved):

nickdaugherty:
r1162-stacks

PR #1757, opened 2020-09-14
Adds WP CLI command to get a specific document

pschoffer:
## Description

Adds WP CLI command to get a specific document. This helps us debug VIP search functionality.
```
wp vip-search documents get post 1
```
## Checklist

## Steps to Test

Example:

1. Check out PR.
1. `lando wp vip-search documents get post 1`
1. Verify a document is returned
1. Add a new post and get it's ID
1. `lando wp vip-search documents get post <new-post-id>`
1. Verify a document populated by the new data is returned.

### DIFF
--- a/search/includes/classes/commands/class-documentcommand.php
+++ b/search/includes/classes/commands/class-documentcommand.php
@@ -43,12 +43,12 @@ class DocumentCommand extends \WPCOM_VIP_CLI_Command {
 			return WP_CLI::error( sprintf( 'Indexable %s not found. Is the feature active?', $type ) );
 		}
 
-		$document = $indexable->get($object_id);
+		$document = $indexable->get( $object_id );
 	
 		if ( ! $document ) {
 			return WP_CLI::error( sprintf( 'Document with ID %s of type %s was not found.', $object_id, $type ) );
 		}
 
-		\WP_CLI::success(print_r($document, true));
+		\WP_CLI::success( print_r( $document, true ) );
 	}
 }
--- a/search/includes/classes/commands/class-documentcommand.php
+++ b/search/includes/classes/commands/class-documentcommand.php
@@ -26,6 +26,9 @@ class DocumentCommand extends \WPCOM_VIP_CLI_Command {
 	 * <object_id>
 	 * : The ID of the object
 	 * 
+	 * [--format=<string>]
+	 * : Optional one of: table json csv yaml ids count
+	 * 
 	 * ## EXAMPLES
 	 *     wp vip-search documents get post 2
 	 *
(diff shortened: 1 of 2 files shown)


search/includes/classes/commands/class-documentcommand.php
```diff
@@ -0,0 +1,54 @@
...
+			return WP_CLI::error( sprintf( 'Indexable %s not found. Is the feature active?', $type ) );
+		}
+
+		$document = $indexable->get($object_id);
+	
+		if ( ! $document ) {
+			return WP_CLI::error( sprintf( 'Document with ID %s of type %s was not found.', $object_id, $type ) );
+		}
+
+		\WP_CLI::success(print_r($document, true));
```
pschoffer, line 52: I am not sure if this is the best way to format it?

search/includes/classes/commands/class-documentcommand.php
```diff
@@ -0,0 +1,54 @@
...
+	 * ## EXAMPLES
+	 *     wp vip-search documents get post 2
+	 *
+	 * @subcommand get
+	 */
+	public function get( $args, $assoc_args ) {
+		$type = $args[0];
+		$object_id = $args[1];
+	
+		$search = \Automattic\VIP\Search\Search::instance();
```
pschoffer, line 38: Should this logic be placed in some business logic class, or is it ok to keep here?

Is there a way to unit-test this?

brandon-m-skinner: It looks okay to me. 

Unit testing WPCLI commands is difficult. If the individual components are unit tested, it's fine for now.

brandon-m-skinner (changes requested):
For formatting in this case your should use [\WP_CLI\Utils\format_items](https://make.wordpress.org/cli/handbook/references/internal-api/wp-cli-utils-format-items/). We already use it in the [validate-contents diffs](https://github.com/Automattic/vip-go-mu-plugins/blob/master/search/includes/classes/commands/class-healthcommand.php#L283-L315).

pschoffer:
I did check the ` \WP_CLI\Utils\format_items`, but that only seems to work for collection of multiple items + you need to know all the fields.

I guess I could make it work by getting all the keys of the single object I have and put that in array so that it is a collection with one item... let me try that.

pschoffer:
Great, `\WP_CLI\Utils\format_items` actually works quite well. In `json` format it evens return multiple depth levels which I thought it won't be able to do.

e.g.:
```
[ {
  "terms": {
      "category": [
        {
          "term_id": 1,
          "slug": "uncategorized",
          "name": "Uncategorized",
          "parent": 0,
          "term_taxonomy_id": 1,
          "term_order": 0
        }
      ]
    },
   ...
}]
```

netsuso:
Hum... I think we are showing too much information on a document to show it in a format_items table... I actually preferred the first way of showing the document. Or maybe we can show in a table just the most relevant fields, and add a flag that shows the full document without the table. Any opinions, @brandon-m-skinner ?

pschoffer:
You can already do `lando wp vip-search --format=json documents get post 1` which returns json, and then pipe it to jq or something to have readable `lando wp vip-search --format=json documents get post 1 | jq`. I think this build in --format flag gives it more versatility, but yeah the print_r dump looks better in some cases.

netsuso:
Ok, we can have a similar output with:

```
lando wp vip-search documents get --format=json post 1 | json_pp
```

So I'll go ahead and merge this and we can decide later

netsuso:
stacks-r1145

PR #1758, opened 2020-09-15
Disable do_pings on sites that use `pre_schedule_event`

joshbetz:
The new (pending) version of Cron Control utilizes the `pre_*` hooks to hook into cron, allowing for a much cleaner way to move cron to its own table. When that plugin is active, we hook on `pre_schedule_event` which short-circuits the schedule_event function before we disable do_pings.

Here we're disabling on both pre_schedule_event as well as schedule_event for backwards compatibility.

### DIFF
--- a/performance/do-pings.php
+++ b/performance/do-pings.php
@@ -15,6 +15,7 @@ function disable_pings( $event ) {
 
 	return $event;
 }
+// Hooking on priority 20 to ensure they run after Cron Control (or anything else that hooks on the default priority)
 add_action( 'pre_schedule_event', __NAMESPACE__ . '\disable_pings', 20 );
 add_action( 'schedule_event', __NAMESPACE__ . '\disable_pings', 20 );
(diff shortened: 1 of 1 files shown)


performance/do-pings.php
```diff
@@ -15,7 +15,8 @@ function disable_pings( $event ) {
 
 	return $event;
 }
-add_action( 'schedule_event', __NAMESPACE__ . '\disable_pings' );
+add_action( 'pre_schedule_event', __NAMESPACE__ . '\disable_pings', 20 );
+add_action( 'schedule_event', __NAMESPACE__ . '\disable_pings', 20 );
```
mjangda, line 19: Can we add a comment indicating why we're setting priority to `20`?

joshbetz:
r1149-vip-go-stacks

joshbetz:
Fixed in #1760

PR #1759, opened 2020-09-16
Search: Change vip_search_enabled QS to use dashes

GaryJones:
## Description

There are a few other behaviour-affecting querystring values that customers and Automatticians can use, and they all use hyphens.

For consistency with them, and a general understanding that hyphens are more frequently used in URLs than underscores anyway, this change updates the `vip_search_enabled` querystring key to `vip-search-enabled`.

The GitHub search didn't reveal any other files using it, so I'm not sure if there are any tests for this querystring feature.

## Checklist

## Steps to Test

Unknown.

## Aside

I noticed that the QS value is not checked; the presence of the key is enough to enable the ES queries, so if someone did `vip-search-enabled=false` (as might be reasonably expected given the `=true` part of the docs), then they would get an incorrect situation.

Would be better to update the check to either look for `true` value, or change the DocBlock to not include the `=true` part. I've not amended that here.

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -15,7 +15,7 @@ class Search {
 
 	public const QUERY_COUNT_CACHE_KEY = 'query_count';
 	public const QUERY_COUNT_CACHE_GROUP = 'vip_search';
-	public const QUERY_INTEGRATION_FORCE_ENABLE_KEY = 'vip_search_enabled';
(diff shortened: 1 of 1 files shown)


mjangda:
Only caution here would be if there are customers actively testing that might be relying on this querystring and would be caught off-guard once we deploy thi.

Otherwise, 💯 for consistency.

brandon-m-skinner (approved):

pschoffer:
r1166

PR #1760, opened 2020-09-16
Disable do_pings via pre_schedule_event

joshbetz:
## Description

Fixes the disabling of do_pings via pre_schedule_event. The new filter
takes an extra argument, so we can't use the exact same function. Here
we're wrapping the disable_pings function in a helper to normalize
the arguments.

## Checklist

## Steps to Test

Example:

1. Check out PR.
1. Ensure cron-control-next is enabled on the test site
1. Publish a new post
1. Confirm do_pings event was not scheduled

The easiest way I've found to verify this is to dump the values to the error log with something like `error_log( var_export( $event, true ) )`.

cron-control-next is important because that's where we're using the pre_schedule_event filter.

### DIFF
--- a/performance/do-pings.php
+++ b/performance/do-pings.php
@@ -15,9 +15,13 @@ function disable_pings( $event ) {
 
 	return $event;
 }
-// Hooking at 0 to get in before cron control on pre_schedule_event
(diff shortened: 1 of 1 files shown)


joshbetz:
r1154-vip-go-stacks

PR #1764, opened 2020-09-18
Improve security

### LINKED ISSUE #1763: Lock out password reset attempts by username as well.
## Expected/Desired Behavior

I think we can do a better job of protecting sites from password reset spam when the reset requests are spread out over IP addresses. We've seen an instance where a site receives 20-30 password reset emails over the span of 15 minutes from different IP addresses.

## Actual Behavior

Currently, we're only locking out `IP|username` and `IP` so if someone is able to spam requests from multiple IP addresses they can bypass this protection.

I nearly have a PR ready for this so we can discuss specific implementation.

brettshumaker:
Improve security


brettshumaker:
Is this an acceptable way to handle this scenario?

Are there any potential downsides to doing this?

Currently, I'm limiting this to only password resets. Would it be beneficial to prevent brute-force password attempts coming in from multiple IP addresses?

brettshumaker:
Closing to get a cleaner PR

PR #1768, opened 2020-09-22
Added site details data generation functionality

brandon-m-skinner:
## Description

Add a filter to generate the site details index. 

Hook it to add the environment and core information.

## Checklist

## Steps to Test

1. Apply PR.
2. `lando test tests/config/test-sitedetailsindex.php `

-------------

1. Apply PR.
2. Add ` require_once( 'config/class-site-details-index.php' );` to the end of `000-vip-init.php`.
3. `lando wp shell`
4. Execute `Automattic\VIP\Config\Site_Details_Index::instance()->set_env_and_core( array() );`
5. The output should match what you'd expect for your environment.

------------

1. Apply PR.
2. `lando wp plugin install bbpress --activate`
3.  Add ` require_once( 'config/class-site-details-index.php' );` to the end of `000-vip-init.php`.
4. Execute `$site_details = Automattic\VIP\Config\Site_Details_Index::instance()->get_plugin_info();`
5. Verify that bbpress has `enabled_via_code` set to false.
6. Add `wpcom_vip_load_plugin( 'bbpress' );` to the end of `000-vip-init.php`
7. Execute `$site_details = Automattic\VIP\Config\Site_Details_Index::instance()->get_plugin_info();`
8. Verify that bbpress has `enabled_via_code` set to true.

### DIFF
--- a/config/class-site-details-index.php
+++ b/config/class-site-details-index.php
@@ -11,7 +11,7 @@ class Site_Details_Index {
 	 * Standard singleton except accept a timestamp for mocking purposes
 	 *
 	 * @param mixed $timestamp A fixed point in time to use for mocking.
-	 * @return {Site_Details_Index} A Site_Details_Index object
+	 * @return Site_Details_Index A Site_Details_Index object
 	 */
 	public static function instance( $timestamp = null ) {
 		if ( ! ( static::$_instance instanceof Site_Details_Index ) ) {
@@ -30,8 +30,8 @@ class Site_Details_Index {
 	/**
 	 * Given the site details from the vip_site_details_index_data filter, set the environment and core values
 	 *
-	 * @param {array} $site_details The current site details
-	 * @return {array} The new site details
+	 * @param array $site_details The current site details
+	 * @return array The new site details
 	 */
 	public function set_env_and_core( $site_details ) {
 		global $wp_version;
@@ -140,7 +140,7 @@ class Site_Details_Index {
 	 *
 	 * Used for mocking in tests.
 	 *
-	 * @return {int} The current timestamp or the value of $this->timestamp
+	 * @return int The current timestamp or the value of $this->timestamp
 	 */
 	public function get_current_timestamp() {
 		return $this->timestamp ?? time();
--- a/config/class-site-details-index.php
+++ b/config/class-site-details-index.php
@@ -34,6 +34,7 @@ class Site_Details_Index {
 	 * @return array The new site details.
 	 */
 	public function set_env_and_core( $site_details ) {
+		// If site details isn't an array, treat it as a blank array
 		if ( ! is_array( $site_details ) ) {
 			$site_details = array();
 		}
@@ -43,24 +44,13 @@ class Site_Details_Index {
 		$site_id = 0;
 		if ( defined( 'VIP_GO_APP_ID' ) && VIP_GO_APP_ID ) {
 			$site_id = VIP_GO_APP_ID;
-		} else {
-			if ( defined( 'FILES_CLIENT_SITE_ID' ) && FILES_CLIENT_SITE_ID ) {
-				$site_id = FILES_CLIENT_SITE_ID;
-			}
 		}
 
 		$environment_name = '';
 		if ( defined( 'VIP_GO_APP_ENVIRONMENT' ) && VIP_GO_APP_ENVIRONMENT ) {
 			$environment_name = strval( VIP_GO_APP_ENVIRONMENT );
-		} else {
-			if ( defined( 'VIP_GO_ENV' ) && VIP_GO_ENV ) {
-				$environment_name = strval( VIP_GO_ENV );
-			}
 		}
 
-		// log2logstash has it's own timestamp field as well. However, if/when we ever decide
-		// to swap indexes, we'll need our own. Added here so that the migration will be easier
-		// and all the fields we need are included in the actual site details index data.
 		$site_details['timestamp'] = $this->get_current_timestamp();
 		$site_details['client_site_id'] = $site_id;
 		$site_details['environment_id'] = $site_id;
--- a/config/class-site-details-index.php
+++ b/config/class-site-details-index.php
@@ -103,7 +103,8 @@ class Site_Details_Index {
 	 * Get the site details for the site the code is running on
 	 */
 	public function get_site_details() {
-		/* Filter that can be hooked on to provide information to the Site Details Index
+		/**
+		 * Filter that can be hooked on to provide information to the Site Details Index.
 		 *
 		 * Example:
 		 *
--- a/config/class-site-details-index.php
+++ b/config/class-site-details-index.php
@@ -34,7 +34,6 @@ class Site_Details_Index {
 	 * @return array The new site details.
 	 */
 	public function set_env_and_core( $site_details ) {
-		// If site details isn't an array, treat it as a blank array
 		if ( ! is_array( $site_details ) ) {
 			$site_details = array();
 		}
@@ -42,8 +41,8 @@ class Site_Details_Index {
 		global $wp_version;
 
 		$site_id = 0;
-		if ( defined( 'VIP_GO_APP_ID' ) && VIP_GO_APP_ID ) {
-			$site_id = VIP_GO_APP_ID;
+		if ( defined( 'FILES_CLIENT_SITE_ID' ) && FILES_CLIENT_SITE_ID ) {
+			$site_id = FILES_CLIENT_SITE_ID;
 		}
 
 		$environment_name = '';
--- a/config/class-site-details-index.php
+++ b/config/class-site-details-index.php
@@ -2,14 +2,14 @@
 
 namespace Automattic\VIP\Config;
 
-class SiteDetailsIndex {
+class Site_Details_Index {
 	private static $_instance;
 
 	private const LOG_FEATURE_NAME = 'site_details_index';
 
 	public static function instance() {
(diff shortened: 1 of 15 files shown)


config/class-sitedetailsindex.php
```diff
@@ -0,0 +1,134 @@
...
+
+		$plugin_info = array();
+
+		foreach ( $all_plugins as $key => $value ) {
+			$active = in_array( $key, $active_plugins, true );
+
+			$plugin_info[] = array(
+				'name' => $value['Name'],
+				'version' => $value['Version'],
+				'active' => $active,
```
mjangda, line 76: Aside: one other data point that would be interesting is whether the plugin was enabled via code or UI. I think there's a helper function we can call to check this?

config/class-sitedetailsindex.php
```diff
@@ -0,0 +1,134 @@
...
+			}
+		}
+
+		// log2logstash has it's own timestamp field as well. However, if/when we ever decide
+		// to swap indexes, we'll need our own. Added here so that the migration will be easier
+		// and all the fields we need are included in the actual site details index data.
+		$site_details['timestamp'] = time();
+		$site_details['client_site_id'] = $site_id;
+		$site_details['environment_id'] = $site_id;
+		$site_details['environment_name'] = strval( get_bloginfo( 'name' ) );
```
mjangda, line 47: `get_bloginfo( 'name' )` returns the title for the site which is different from what we refer to as the "environment name" so this will get confusing. I think we probably want `VIP_GO_APP_ENVIRONMENT`?

config/class-sitedetailsindex.php
```diff
@@ -0,0 +1,134 @@
+<?php
+
+namespace Automattic\VIP\Config;
+
+class SiteDetailsIndex {
```
mjangda, line 5: Minor: underscores

```suggestion
class Site_Details_Index {
```

All references below and elsewhere would need to be updated as well. And we should rename the file to `class-site-details-index.php`.

config/class-sync.php
```diff
@@ -65,6 +65,7 @@ public function filter_cron_schedules( $schedule ) {
 
 	public function do_cron() {
 		$this->maybe_sync_jetpack_privacy_settings();
+		$this->put_site_details();
```
mjangda, line 68: Should we restrict this to a one (or a few sites) to start to make sure everything's working correctly?

tests/config/test-sitedetailsindex.php
```diff
@@ -0,0 +1,109 @@
...
+		$sdi = SiteDetailsIndex::instance();
+		
+		$this->assertTrue( is_integer( has_filter( 'vip_site_details_index_data', [ $sdi, 'set_env_and_core' ] ) ) );
+	}
+
+	/**
+	 * @runInSeparateProcess
+	 * @preserveGlobalState disabled
+	 */
+	public function test__set_env_and_core() {
```
mjangda, line 42: For the ones where we can (e.g. multisite, blog_id, client_site_id, plugins), we should verify the values as well.

mjangda (commented):
Noted a couple of small suggestions.

tests/config/test-sitedetailsindex.php
GaryJones: `time()` (as used [here](https://github.com/Automattic/vip-go-mu-plugins/pull/1768/files?w=1#diff-60017c90bd0b751e35b1cf0a2e47709dR44)) is an external dependency (it uses the system clock). If the timestamp value was injected into the SiteDetailsIndex class, then it could be mocked and thus unit tested to verify the expected value as well. See https://matthiasnoback.nl/2018/02/mocking-at-architectural-boundaries-persistence-and-time/.

brandon-m-skinner: Sure. 

Probably shouldn't do plugins since it varies depending on environment. For example I have bbpress show up when I run the tests in lando since that plugin is installed when I was testing stuff.

I added the timestamp mocking on a per-instance basis.

config/class-sitedetailsindex.php
brandon-m-skinner: Sure, easy enough.

tests/config/test-sitedetailsindex.php
brandon-m-skinner: Actually, scratch that on plugins. Can just mock by setting the cache.

config/class-site-details-index.php
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	/**
+	 * Given a value, set the current timestamp to provided value if it's an integer
+	 *
+	 * Used for mocking in tests.
+	 */
+	public function set_current_timestamp( $timestamp ) {
+		if ( ! is_int( $timestamp ) ) {
+			return;
```
rinatkhaziev, line 155: Should this set the current timestamp instead of just silently bailing? Additionally, the `instance` method calls it with `null` making the logic a bit confusing.

config/class-site-details-index.php
```diff
@@ -0,0 +1,160 @@
...
+		add_filter( 'vip_site_details_index_data', [ $this, 'set_env_and_core' ] );
+	}
+
+	/**
+	 * Given the site details from the vip_site_details_index_data filter, set the environment and core values
+	 *
+	 * @param {array} $site_details The current site details
+	 * @return {array} The new site details
+	 */
+	public function set_env_and_core( $site_details ) {
```
rinatkhaziev, line 36: We should either add a type annotation here or check the `$site_details` for `is_array` to be on a safer side.

config/class-site-details-index.php
```diff
@@ -0,0 +1,160 @@
...
+			if ( defined( 'FILES_CLIENT_SITE_ID' ) && FILES_CLIENT_SITE_ID ) {
+				$site_id = FILES_CLIENT_SITE_ID;
+			}
+		}
+
+		$environment_name = '';
+		if ( defined( 'VIP_GO_APP_ENVIRONMENT' ) && VIP_GO_APP_ENVIRONMENT ) {
+			$environment_name = strval( VIP_GO_APP_ENVIRONMENT );
+		} else {
+			if ( defined( 'VIP_GO_ENV' ) && VIP_GO_ENV ) {
```
rinatkhaziev, line 52: We should probably drop `VIP_GO_ENV`, it's a back-compat constant that shouldn't be relied on. There won't be a case where `VIP_GO_APP_ENVIRONMENT` is not defined but `VIP_GO_ENV` is.

config/class-site-details-index.php
```diff
@@ -0,0 +1,160 @@
...
+	 * Given the site details from the vip_site_details_index_data filter, set the environment and core values
+	 *
+	 * @param {array} $site_details The current site details
+	 * @return {array} The new site details
+	 */
+	public function set_env_and_core( $site_details ) {
+		global $wp_version;
+
+		$site_id = 0;
+		if ( defined( 'VIP_GO_APP_ID' ) && VIP_GO_APP_ID ) {
```
rinatkhaziev, line 40: SImilarly to the below, feels like using either `VIP_GO_APP_ID` or `FILES_CLIENT_SITE_ID` is sufficient, since they both use the same value on config generation and there wouldn't be a case where one was defined and the other wouldn't.

rinatkhaziev (changes requested):
Looks good, left a few notes inline.

config/class-site-details-index.php
brandon-m-skinner: Setting the timestamp to anything else will means the timestamp for that instance is always gonna be what you set it as.

Since it's done with a getter/setter setup, keep the value null means that

```
return $this->timestamp ?? time();
```

Will always be time();

config/class-site-details-index.php
```diff
@@ -0,0 +1,160 @@
...
+
+class Site_Details_Index {
+	private static $_instance;
+
+	private $timestamp = null;
+
+	/**
+	 * Standard singleton except accept a timestamp for mocking purposes
+	 *
+	 * @param {mixed} $timestamp A fixed point in time to use for mocking
```
GaryJones, line 13: Unlike JSDoc, the use of `{}` around `@param` and `@return` types is not standard for PHPDoc.

https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md#abnf

For WPCS, summary and tag descriptions should also end with a full stop.

https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#1-functions-class-methods

```suggestion
	 * @param mixed $timestamp A fixed point in time to use for mocking.
```

config/class-site-details-index.php
```diff
@@ -0,0 +1,160 @@
...
+		}
+
+		return $plugin_info;
+	}
+
+	/**
+	 * Get the site details for the site the code is running on
+	 */
+	public function get_site_details() {
+		/* Filter that can be hooked on to provide information to the Site Details Index
```
GaryJones, line 106: I really appreciate the investment in documenting this filter.

The [WPCS standard for documenting filters](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#4-hooks-actions-and-filters) is to use a DocBlock, rather than a multi-line comment. That's what WP-Parser and other compatible tools will look for.

```suggestion
		/**
		 * Filter that can be hooked on to provide information to the Site Details Index.
```

brandon-m-skinner:
Great points @GaryJones, got them sorted out.

rinatkhaziev (approved):
🚢

brandon-m-skinner:
r1173-stacks

PR #1770, opened 2020-09-22
Log Elasticsearch warnings and errors

netsuso:
## Description

Detect errors (HTTP code error + error section in the response) and warnings (HTTP header) in Elasticsearch responses and log them to logstash, so they get more visibility and we don't depend on someone looking at Elasticsearch logs.

## Checklist

## Steps to Test

1. Check out PR.
1. Enforce some error/warning in Elasticsearch queries (e.g., adding `orderby` on some text field, or adding `'_source' => [ 'exclude' => 'meta.*' ]`.
1. In the local dev environment it should appear in `debug.log`. In a sandbox it should appear in logstash under features `vip_search_query_error` or `vip_search_es_warning`

### DIFF
--- a/logstash/class-logger.php
+++ b/logstash/class-logger.php
@@ -516,7 +516,7 @@ class Logger {
 	 * @param array $entry Data.
 	 */
 	public static function wp_debug_log( array $entry ) : void {
-		if ( ! defined( 'VIP_GO_ENV' ) || ! VIP_GO_ENV ) {
+		if ( defined( 'VIP_GO_ENV' ) && VIP_GO_ENV ) {
 			// Don't run this on VIP Go
 			return;
 		}
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -605,6 +605,7 @@ class Search {
 			// Check for the 'Warning' header and log it
 			if ( isset( $response_headers['warning'] ) ) {
 				$warning_message = $response_headers['warning'];
+				trigger_error( $warning_message, \E_USER_WARNING );
 				\Automattic\VIP\Logstash\log2logstash( array(
 					'severity' => 'warning',
 					'feature' => 'vip_search_es_warning',
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -582,6 +582,34 @@ class Search {
 
 			$statsd->timing( $statsd_prefix . '.total', $duration );
 			$statsd->timing( $statsd_per_site_prefix . '.total', $duration );
+
+			// Check for errors in the response and log them
+			$response_code = (int) wp_remote_retrieve_response_code( $request );
+			if ( $response_code >= 400 ) {
+				$response_error = $response['error'];
+				$error_message = $response_error['reason'] ?? 'Unknown Elasticsearch query error';
+				\Automattic\VIP\Logstash\log2logstash( array(
+					'severity' => 'error',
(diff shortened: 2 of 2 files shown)


logstash/class-logger.php
```diff
@@ -516,7 +516,7 @@ public static function maybe_wp_debug_log_entries( array $entry ) : void {
 	 * @param array $entry Data.
 	 */
 	public static function wp_debug_log( array $entry ) : void {
-		if ( ! defined( 'VIP_GO_ENV' ) || ! VIP_GO_ENV ) {
+		if ( defined( 'VIP_GO_ENV' ) && VIP_GO_ENV ) {
```
netsuso, line 519: The check didn't match the description ("Don't run this on VIP Go"). The intention is to avoid VIP Go sites to attempt to write to `wp-content/debug.log`

search/includes/classes/class-search.php
```diff
@@ -550,14 +550,14 @@ public function filter__ep_do_intercept_request( $request, $query, $args, $failu
 
 		$timeout = $this->get_http_timeout_for_query( $query, $args );
 
-		$request = vip_safe_wp_remote_request( $query['url'], $fallback_error, 3, $timeout, 20, $args );
+		$response = vip_safe_wp_remote_request( $query['url'], $fallback_error, 3, $timeout, 20, $args );
```
mjangda, line 553: 💯 on renaming the variables!

search/includes/classes/class-search.php
```diff
@@ -572,19 +572,48 @@ public function filter__ep_do_intercept_request( $request, $query, $args, $failu
...
 			$statsd->timing( $statsd_prefix . '.total', $duration );
 			$statsd->timing( $statsd_per_site_prefix . '.total', $duration );
+
+			$response_code = (int) wp_remote_retrieve_response_code( $response );
+
+			// Check for an error HTTP code and log the Elasticsearch error
+			if ( $response_code >= 400 ) {
+				$response_error = $response_body['error'];
+				$error_message = $response_error['reason'] ?? 'Unknown Elasticsearch query error';
+				\Automattic\VIP\Logstash\log2logstash( array(
```
mjangda, line 592: For this and the `Warning` header below, should we also call `trigger_error` with `E_USER_WARNING` to give customers visibility into these errors as well? (With enough volume, it would get surfaced in New Relic.)

netsuso: Oh, it certainly makes sense! One day I'll need to investigate what are all the paths for error logging, as this is still a bit obscure for me :sweat_smile:

search/includes/classes/class-search.php
netsuso: It took me more time to understand the code due to this `$request` variable name, so I thought it would be a good idea to rename it for the next one who gets here 😊

search/includes/classes/class-search.php
netsuso: For the moment I've only added this to the Warning header condition, as in that case I can trust more the message. For the 400/500 error responses, there's more diversity, and often the message just becomes "all shards failed". It means that all of them failed to execute the query because it was wrong, but sending an "all shards failed" warning to New Relic can be very misleading.

I'll monitor what we receive in logstash (with the full detail of the error) and will decide what's the best way to trigger a useful warning here.

mjangda (approved):
LGTM

netsuso:
stacks-r1172

PR #1772, opened 2020-09-23
Add APC Cache Interceptor

joshbetz:
The APC Cache Interceptor defines hot cache keys to be intercepted and persisted in APCu.

The idea here is to create some short-term persistent caches to reduce load on memcached.

Our first two examples here are

* `userlogins:wpcomvip` which is requested on every page and just returns a user ID that never changes
* `sites:<id>` which returns some basic information about a subsite and very rarely changes

The caveat is that changes are not propagated to other PHP instances, so cached data is potentially
stale as soon as it is cached. It is important that we don't use stale data to overwrite the database. This
is why it's notable that the values above don't (or very rarely) change and why we're only caching the site
data for 10 seconds.

Depends on https://github.com/Automattic/wp-memcached/pull/66


rinatkhaziev:
@joshbetz Is this just a copy of WP.com implementation? (plus 3904593e1d821d885e7de2e1b179cf19adf0656d change)

joshbetz:
This bit is also different

https://github.com/Automattic/vip-go-mu-plugins/pull/1772/files#diff-53160506b927020bde2e91131a96a4e6R173-R178

We don't prepend the salt or flush version on wp.com, so this isn't a problem there.

rinatkhaziev (dismissed):

lib/class-apc-cache-interceptor.php
```diff
@@ -0,0 +1,494 @@
...
+
+		public function __construct() {
+			// Allow for the possibility that this is already defined
+			if ( ! defined( 'WP_OBJ_CACHE_HOOKS' ) ) {
+				define( 'WP_OBJ_CACHE_HOOKS', true );
+			}
+			// If we should not be hooking then there's no reason to add overhead to actions and filters
+			if ( defined( 'WP_OBJ_CACHE_HOOKS' ) && WP_OBJ_CACHE_HOOKS ) {
+				do_action( 'apcu_hot_cache_init', $this );
+				add_filter( 'wp_cache_get_pre', array( $this, 'wp_cache_get' ), 1, 4 );
```
rinatkhaziev, line 317: Now that we changed the filter in https://github.com/Automattic/wp-memcached/pull/66 we should also update its usage here.
```suggestion
				add_filter( 'pre_wp_cache_get', array( $this, 'wp_cache_get' ), 1, 4 );
```

nickdaugherty (dismissed):

joshbetz:
r1181-vip-go-stacks

PR #1773, opened 2020-09-23
Search: code cleanup - remove query mirroring code and re-organize the order of constants and properties in Search class.

rinatkhaziev:
## Description

This PR is a bit of housekeeping and cruft removal:

- Query mirroring was used during the early days of the project to ensure there are no discrepancies between offloaded and the SQL queries, as well as gauging the stability of the solution. None of it is needed anymore.
- Additionally, reorganize class constants and properties: first define public and private constants, then define the class and instance properties.

## Checklist

## Steps to Test

Not much to test, the functionality and accompanying tests are removed completely.


nickdaugherty (approved):
Looks great, gotta love any PR that removes nearly 600 lines 😄

GaryJones:
I'd rather see the commit for re-organising the class properties and constants be left for a separate PR, and not be bundled with a change that is about removing a specific un-needed feature.

Why?

Apart from the general case of potentially needing to revert the removal of a feature, independently of re-organising a class, the public properties are not typed declared (in a `@var` for now), have no defaults, and are not documented with a description of their purpose and intended usage, so there's precisely no indication how other code is supposed to interact with these properties.

- Is `$healthcheck` a boolean, an iterative array or collection of healthcheck items, or an instance of an object?
- What unit is `$queue_wait_time` in? Integer seconds, or microseconds, or maybe a `\DateInterval`?
- What does `$field_count_gauge` hold? What is its purpose?
- What gets held in the `$queue`? Is it an array? An instance of a collection class? Some other object? What keys or properties does it have? How can I inspect what's in the queue? How can I add to it or remove from it?

It should be possible to answer (rhetorical) questions like those from just looking at the properties themselves (do they all really need to be public?), but I can't, and would need to look through the whole class to understand why they exist and how I could use them.

Keeping the changes in this PR in separate PRs would allow for a focus on making the code more easier to read and understand, which will be useful if non-Platform devs (in our out of VIP) are to have a better understanding of how the code actually works.

(Public properties are naturally also ripe for customer devs also accidentally being able to make unintended / unexpected changes to the values, like `$max_query_count`, and any future handling of special cases can't easily be done - getters are a little less fragile, but changing them is further outside the main point I'm making here.)

rinatkhaziev:
@GaryJones I generally agree on all the things you said and 150% agree on annotations and general docs which are lacking at the moment, it's definitely something that we're aware of and are making efforts to address.

The original intent of this was to remove the query mirroring, at which point I also noticed that constants and properties could use some tidying up, seemed like a quick win.

I've updated the PR title in description to better describe the scope of the changes.

brandon-m-skinner:
r1174-stacks

PR #1776, opened 2020-09-24
Update elasticpress module to fix deprecated fields

netsuso:
## Description

Update ElasticPress to include [this change](https://github.com/Automattic/ElasticPress/pull/56), to avoid warnings from Elasticsearch about deprecated fields.

## Checklist


.gitmodules
```diff
@@ -40,9 +40,9 @@
 [submodule "cron-control-next"]
 	path = cron-control-next
 	url = https://github.com/Automattic/Cron-Control.git
-[submodule "elasticsearch/elasticpress"]
+[submodule "search/elasticpress"]
```
netsuso, line 43: I had to change the submodule name to match the path to avoid CircleCI failing with `submodule not found`

netsuso: It seems that for the job `php73-core-tests`, as the docker image doesn't contain `git`, CircleCI has to run its own git implementation, which I think it's based on [go-git](https://github.com/go-git/go-git). It seems that it looks for changes in submodules, and then takes the path and looks it up in the list of modules. But in this case the path didn't match the submodule name, so it failed to find it.

brandon-m-skinner (approved):

brandon-m-skinner:
r1177-stacks

PR #1778, opened 2020-09-25
Support multiple Warning headers on Elasticsearch responses

netsuso:
## Description

In #1770 we added logging for `Warning` headers in Elasticsearch responses. It turns out that if there are several of them, then `wp_remote_retrieve_headers()` will return an array with all of them instead of a string. This PR adds support for that case.

## Checklist

## Steps to Test

1. Check out PR.
1. Run `wp vip-search index-versions add post` (which triggers several warnings, at least in the local dev environment)
1. Verify that all warnings are shown, and there are no errors about array->string conversion

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -413,15 +413,21 @@ class Search {
 
 			$response_headers = wp_remote_retrieve_headers( $response );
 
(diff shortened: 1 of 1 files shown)


brandon-m-skinner (approved):

netsuso:
stacks-r1183

PR #1779, opened 2020-09-25
Add --network-wide support for pending search commands

netsuso:
## Description

There are still four vip-search commands that don't support `--network-wide` parameter (get, get-active, list, delete). This PR addresses them.

## Checklist

## Steps to Test

1. Check out PR.
1. Create a multisite testing site
1. Add a couple of index versions with `vip-search index-versions add post --network-wide`
1. Run `vip-search index-versions get post active --network-wide` and verify that the active version is returned for all sites
1. Run `vip-search index-versions list post --network-wide` and verify that all index versions are returned for all sites

### DIFF
--- a/search/includes/classes/commands/class-versioncommand.php
+++ b/search/includes/classes/commands/class-versioncommand.php
@@ -354,6 +354,8 @@ class VersionCommand extends \WPCOM_VIP_CLI_Command {
 		if ( ! $indexable ) {
 			return WP_CLI::error( sprintf( 'Indexable %s not found. Is the feature active?', $type ) );
 		}
+
+		CoreCommand::confirm_destructive_operation( $assoc_args );
 	
 		if ( isset( $assoc_args['network-wide'] ) && is_multisite() ) {
 			if ( ! is_numeric( $assoc_args['network-wide'] ) ) {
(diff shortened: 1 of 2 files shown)


search/includes/classes/commands/class-versioncommand.php
```diff
@@ -288,67 +355,68 @@ public function delete( $args, $assoc_args ) {
...
+				$assoc_args['network-wide'] = 0;
+			}
 
-		if ( is_wp_error( $version_number ) ) {
-			return WP_CLI::error( sprintf( 'Index version %s is not valid: %s', $args[1], $version_number->get_error_message() ) );
-		}
+			$sites = \ElasticPress\Utils\get_sites( $assoc_args['network-wide'] );
 
-		$active_version_number = $search->versioning->get_active_version_number( $indexable );
+			CoreCommand::confirm_destructive_operation( $assoc_args );
```
nickdaugherty, line 365: Minor - to keep things more DRY, can the confirmations go outside the network-wide conditional?

nickdaugherty (approved):
This is really great! Awesome work, everything works as expected.

I left one small suggestion, not a blocker though.

search/includes/classes/commands/class-versioncommand.php
netsuso: Sure! I moved it here because for the single site mode it asks for confirmation after validating the version. But in network mode it has to be done early. So for simplicity it's better to move it outside as you suggest

netsuso:
stacks-r1185

PR #1780, opened 2020-09-25
Search: switch vip-search index --network-wide to spawn child processes instead of using switch_to_blog

rinatkhaziev:
## Description

The current implementation may cause the code not to be loaded - `switch_to_blog` doesn't do that, leading to issues like post types missing from the index for a specific site. 

E.g. if a subsite has a plugin that the main site is not loading, a post type registered by that plugin will be missing from the index. 

This PR addresses the issue by utilizing `WP_CLI::runcommand` which spawns a child process and switches execution to it, thus ensuring the proper codebase is loaded.

We will likely need to extend the approach to other `--network-wide` commands.

## Checklist

## Steps to Test

1. Check out PR.
1. Run `wp vip-search index --network-wide` on multisite
1. Verify that indexing runs through without issues
1. Run `wp vip-search index` and verify that only the first site is indexed


nickdaugherty (approved):
This is awesome! Such an elegant solution for such a big bug. Tested and works great!

nickdaugherty:
To prove that --network-wide is loading each subsite's themes, I did this:

1. `lando wp theme install twentynineteen`
1. `lando wp theme activate twentynineteen --url=sometest.vip-go-dev.lndo.site`
1. Add `die( 'in twentynineteen' ); to `wp-content/themes/twentynineteen/functions.php`
1. `lando wp vip-search index --network-wide` - runs fine on first site, dies with above message on site 2, as expected

search/includes/classes/commands/class-corecommand.php
```diff
@@ -68,9 +68,33 @@ public function index( $args, $assoc_args ) {
...
+				$assoc_args['url'] = home_url();
+
+				WP_CLI::line( 'Indexing ' . $assoc_args['url'] );
+				WP_CLI::runcommand( 'vip-search index ' . Utils\assoc_args_to_str( $assoc_args ) );
+				Utils\wp_clear_object_cache();
+			}
 
-		WP_CLI::run_command( $args, $assoc_args );
+			WP_CLI::line( WP_CLI::colorize( '%CNetwork-wide run took: ' . ( round( microtime( true ) - $start, 3 ) ) . '%n' ) );
+			restore_current_blog();
```
GaryJones, line 93: This `restore_current_blog()` should be matched to the `switch_to_blog()` (i.e. inside the `foreach`) so that the global array that tracks the stack of switched sites, and the global `switched` is correct.

GaryJones (dismissed):

search/includes/classes/commands/class-corecommand.php
rinatkhaziev: Nice catch, thanks!

GaryJones (approved):
I've not tested, but nothing else jumps out at me when looking at the code.

rinatkhaziev:
r1203-stacks

PR #1781, opened 2020-09-25
APC Cache Interceptor: Don't offload subsite information in network admin

joshbetz:
Offloading the subsite information in the network admin has the side effect of showing stale data after clicking save on a site. (i.e. from /wp-admin/network/site-info.php?id=<id>). The form still saves properly and if you refresh the page, you see the correct thing. It's kind of suboptimal though.

To counteract that, we're disabling apcu for these keys when you're in the network admin. That way, you still get the data directly from memcached in the screen when you could be updating this data, but in all other cases, you get the apcu offloaded data.

## Checklist

## Steps to Test

Example:

1. Check out PR.
1. Go to wp-admin > My Sites > Network Admin -> Sites
1. Click Edit on a site record
1. Update a Site Address
1. Observe that the address is the updated version when the page reloads

### DIFF
--- a/lib/class-apc-cache-interceptor.php
+++ b/lib/class-apc-cache-interceptor.php
@@ -17,8 +17,11 @@ function do_apcu_hot_cache_init( $hc ) {
 	// Cache the wpcomvip user ID
 	$hc->add_passive_rx( 'userlogins', '/userlogins:wpcomvip$/', 300 );
 
-	// Cache subsite information for 10 seconds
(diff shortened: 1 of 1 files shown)


mjangda:
Could the same problem happen in REST API / WP-CLI contexts as well?

joshbetz:
@mjangda good call - fixed

nickdaugherty (approved):

PR #1782, opened 2020-09-26
Add login limit threshold filters

joshbetz:
Allow client sites to customize the login limit thresholds

## Checklist

## Testing

1. Checkout this PR
1. Filter `wpcom_vip_ip_login_threshold` to return `0`
1. Confirm trying to login gets the rate limited error message
1. Filter `wpcom_vip_ip_username_login_threshold` to return `0`
1. Confirm trying to login gets the rate limited error message
1. Remove filters
1. Confirm trying to login gets no rate limited error message

You can also experiment with various other values to show that a given username is rate limited, but other usernames from the same IP have a different rate limit.

### DIFF
--- a/security.php
+++ b/security.php
@@ -36,8 +36,8 @@ function wpcom_vip_is_restricted_username( $username ) {
  */
 function wpcom_vip_track_auth_attempt( $username, $cache_group, $cache_expiry ) {
 	$ip   = preg_replace( '/[^0-9a-fA-F:., ]/', '', $_SERVER['REMOTE_ADDR'] );
-	$key1 = $ip . '|' . $username; // IP + username
-	$key2 = $ip; // IP only
+	$ip_username_cache_key = $ip . '|' . $username; // IP + username
+	$ip_cache_key = $ip; // IP only
 
 	// Longer TTL when logging in as admin, which we don't allow on WP.com
 	$is_restricted_username = wpcom_vip_is_restricted_username( $username );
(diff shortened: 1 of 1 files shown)


security.php
```diff
@@ -155,23 +155,23 @@ function wpcom_vip_lost_password_limit( $errors ) {
...
-	$key1                   = $ip . '|' . $username;
-	$key2                   = $ip;
-	$count1                 = wp_cache_get( $key1, $cache_group );
-	$is_restricted_username = wpcom_vip_is_restricted_username( $username );
-	$threshold2             = 50;
+	$key1 = $ip . '|' . $username;
+	$key2 = $ip;
+	
+	$threshold1 = apply_filters( 'wpcom_vip_ip_username_login_threshold', 5 );
+	$threshold2 = apply_filters( 'wpcom_vip_ip_login_threshold', 50 );
```
GaryJones, line 162: Might be useful to some clients to make the `$ip` and the requested host / URL available in the filter, so that they could set values based on requests from an IP range, or for particular subsites in a multisite install.

joshbetz: Yeah, that would be great.

```
$ip = preg_replace( '/[^0-9a-fA-F:., ]/', '', $_SERVER['REMOTE_ADDR'] );
```

Unfortunately this is probably not that useful to just pass as an argument to the filter. Happy to do whatever makes sense though.

> or for particular subsites in a multisite install.

I'm thinking it may be better to just run `get_current_blog_id()` for this. Depending on how the multisite is setup, it may or may not matter what the blog ID is. (For subdirectory installs, logging in to any site logs you into every site.) Because of that, I think it could be misleading to pass the blog ID here. Does that seem reasonable?

security.php
```diff
@@ -155,24 +155,24 @@ function wpcom_vip_lost_password_limit( $errors ) {
...
-	$key1                   = $ip . '|' . $username;
-	$key2                   = $ip;
-	$count1                 = wp_cache_get( $key1, $cache_group );
-	$is_restricted_username = wpcom_vip_is_restricted_username( $username );
-	$threshold2             = 50;
+	$ip_username_cache_key = $ip . '|' . $username;
+	$ip_cache_key = $ip;
+	
+	$ip_username_threshold = apply_filters( 'wpcom_vip_ip_username_login_threshold', 5 );
+	$ip_threshold = apply_filters( 'wpcom_vip_ip_login_threshold', 50 );
```
mjangda, line 162: We should add a docblock for these new filters.

security.php
```diff
@@ -36,8 +36,8 @@ function wpcom_vip_is_restricted_username( $username ) {
  */
 function wpcom_vip_track_auth_attempt( $username, $cache_group, $cache_expiry ) {
 	$ip   = preg_replace( '/[^0-9a-fA-F:., ]/', '', $_SERVER['REMOTE_ADDR'] );
-	$key1 = $ip . '|' . $username; // IP + username
-	$key2 = $ip; // IP only
+	$ip_username_cache_key = $ip . '|' . $username; // IP + username
+	$ip_cache_key = $ip; // IP only
```
mjangda, line 40: Thanks for renaming these; so much better!

security.php
joshbetz: Actually, reading that regex again, it's just removing invalid characters. I'll pass the IP to the filters.

mjangda (approved):
LGTM.

mjangda:
r1190-stacks

PR #1774, opened 2020-09-24
Adds core tests job while including the mu-plugins

pschoffer:
## Description

We wanted to be able to test mu-plugins against core WordPress tests.

## Steps to Test

Check the CircleCI jobs linked to this PR.

PR #1791, opened 2020-10-01
Update/core tests with mu plugins

pschoffer:
## Description

We wanted to be able to test mu-plugins against core WordPress tests.

This includes indentation fix that broke the previous PR :) - https://github.com/Automattic/vip-go-mu-plugins/pull/1774

## Steps to Test

Check the CircleCI jobs linked to this PR.


netsuso (approved):
All good! I think we can have a look at the Core tests checkout time, but it can be in another PR :)

PR #1792, opened 2020-10-01
Speeding up git checkout for circleCI jobs

pschoffer:
## Description
Speeding up git checkout for circleCI jobs. 

EDIT: turns out avoiding to use CircleCI to use it's default git+ssh, but installing one locally is way faster.

## Steps to Test

Check core tests jobs in circle CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -87,6 +87,7 @@ job-references:
 
   php_core_job: &php_core_job
     steps:
+      - run: sudo apt-get update && sudo apt-get -y install git ssh
       - checkout
       - run: *install_core_tests_dependencies
       - run: *prepare_repo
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -87,6 +87,7 @@ job-references:
 
   php_core_job: &php_core_job
     steps:
+      - run: sudo apt-get update && sudo apt-get -y install git ssh
```
netsuso, line 90: Maybe we can make this cleaner in the CircleCI job output:

```suggestion
      - run:
          name: Add missing packages
          command: sudo apt-get update && sudo apt-get -y install git ssh
```

netsuso (changes requested):
Great job execution time savings! I have just a cosmetic suggestion :)

netsuso:
No need to deploy this one as it only affects `.circleci`, which is not included in the svn stacks repo

PR #1795, opened 2020-10-02
Tracks indexing job lenght

pschoffer:
## Description

First step towards storing statsD metrics around VIP search. This PR adds indexing time metric under:
`stats.timers.com.wordpress.elasticsearch.$dc.[[node]]_[[cluster]].bulk_index.[total|engine]` key

## Checklist

## Steps to Test

Example:

1. in local environement run ` lando wp vip-search index` (with --setup if ran for the first time)
1. Then you can do `lando logs -s stats` to see the recorded stats


pschoffer:
I am not sure if the new stat can be merged one by one, so please wait with merging unless you know that it is ok:)

brandon-m-skinner:
Does this only record for a successful full index?

pschoffer:
> I am not sure if the new stat can be merged one by one, so please wait with merging unless you know that it is ok:)

I got confirmation from Brandon, that it is OK to merge only one stat at a time (although new keys might not be recorded untill a request to SysTeam will be made).

> Does this only record for a successful full index?

Nop, this will record any operation when we call `wp vip-search index ...`, we kinda only clean the parameters and pass along to the elasticpress code for processing.

There is no feedback from EP about if it succeeded or not back to the caller (it only makes STDOUT prints). We could of course move the time logic inside EP, but that seems to me as unnecessary more forking of the project.

I am not sure about what do you mean by full index? Do we want to split this metric into multiple based on arguments like type or `--setup` flag?

brandon-m-skinner:
> I am not sure about what do you mean by full index? Do we want to split this metric into multiple based on arguments like type or --setup flag?

All of whatever indexable is indexed. Partial indexes or recording stats from quick failures are just junk data for this stat.

>Nop, this will record any operation when we call wp vip-search index ..., we kinda only clean the parameters and pass along to the elasticpress code for processing.
>
>There is no feedback from EP about if it succeeded or not back to the caller (it only makes STDOUT prints). We could of course move the time logic inside EP, but that seems to me as unnecessary more forking of the project.

In that case, the stats collection logic doesn't belong in the command logic. It should be in the actual request logic. See the [request interrupt](https://github.com/Automattic/vip-go-mu-plugins/blob/master/search/includes/classes/class-search.php#L342-L435) and [the detection logic](https://github.com/Automattic/vip-go-mu-plugins/blob/82308873535a775d263daa3f2cc24a36d8774ccd/search/includes/classes/class-search.php#L838-L893) for an idea. Basically we need to detect:

- an indexing operation
- whether there is an offset or [something else set](https://github.com/Automattic/ElasticPress/pull/59/files#diff-3045e7e36c99cf10ba4593970093bb39R503) that would restrict the index to less than a full index
- whether the operation succeeded

and only record a stat for full, successful indexing operations so we have the most accurate full indexing stats possible.

pschoffer:
I see. Is it possible to do on request level though?

From my understanding, if you would like to do an index of all types that have to be at least one request per type (e.g. `https://foo.com/index/posts/_bulk`, `https://foo.com/index/users/_bulk`).

In the current version of this PR I at least differentiate between `index` vs `bulk_index`, but there is no easy way (like offset parameter) to know if the body includes all the documents for a given type. I guess I could parse the body and count records and then count records for that type somehow (as the check count CLI tool does). Is that something we want?

brandon-m-skinner:
Most indexing, even on a small scale, is done via the bulk endpoint in ElasticPress. 

>I guess I could parse the body and count records and then count records for that type somehow (as the check count CLI tool does). Is that something we want?

If it works, yes. It's pretty quick to do count comparisons and if it's within say 100 or something it's safe to call it a full index. Otherwise the stats won't be useful since we'll have lots of low-count indexes running in milliseconds skewing our data way down.

pschoffer:
Alright, I will look into that tmrw. Do you have any feedback for the rest of it? What about keeping the `index.total` and `index.engine` stats. Do we want those, or should I drop them as well (if not full index)? I think the more stats the better, but there might be an issue with too many stats.

brandon-m-skinner:
If they're currently there, we should just leave them for now and keep this PR focused on adding the indexing metric tracking.
The stats can be tidied up in subsequent PRs.

search/includes/classes/class-search.php
```diff
@@ -885,13 +891,41 @@ public function get_statsd_request_mode_for_request( $url, $args ) {
...
+		$body = $args['body'] ?? '';
+		$body_lines = explode( "\n", $body );
+
+		if ( $body_lines && is_array( $body_lines ) ) {
+			$doc_count = floor( count( $body_lines ) / self::LINES_PER_DOCUMENT_IN_BULK_INDEX );
+			$indexable_type_path = $path[ count( $path ) - 2 ];
+
+			$all_indexables = $this->indexables->get_all();
+			foreach ( $all_indexables as $indexable ) {
+				if ( strpos( $indexable_type_path, $indexable->slug ) !== false ) {
```
pschoffer, line 916: I am not sure about this:
`strpos( $indexable_type_path, $indexable->slug ) !== false `

We basically check that the url has some slug in it. Any opinions? e.g. `vip-200508-post-1` contains `post`

lib/statsd/class-statsd.php
```diff
@@ -15,8 +15,25 @@ class StatsD {
...
+		$this->update_stats( $stats, $time, 1, 'ms' );
+	}
+
+	/**
+	 * Tracks time of callback and then sets one or more timing values
+	 *
+	 * @param string|array $stats The metric(s) to set.
+	 * @param callback $callback to be tracked
+	 **/
+	public function time( $stats, $callback ) {
```
nickdaugherty, line 28: It doesn't look like this function is actually used anywhere, so probably better to leave it out of this PR (until it's actually needed).

lib/statsd/class-statsd.php
```diff
@@ -15,8 +15,25 @@ class StatsD {
...
+		$start_time = microtime( true );
+
+		call_user_func( $callback );
```
nickdaugherty, line 31: This won't pass any arguments to the function, so could only be on certain functions (or would cause errors).

search/includes/classes/class-search.php
```diff
@@ -885,13 +891,41 @@ public function get_statsd_request_mode_for_request( $url, $args ) {
...
 		// Unknown
 		return 'other';
 	}
 
+	/**
+	 * From the path will parse what type of indaxable is it. From the body of the request how many documents we are indexing.
+	 *
+	 * If the number of documents matches the number of corresponding objects in DB, will return true, false otherwise.
+	 */
+	private function is_full_index( array $path, array $args ) {
```
nickdaugherty, line 906: The logic rules for this function seem very fragile, and it could end up being fairly expensive to run because it has to do a DB lookup for every call.

If we're trying to detect the duration of a full bulk index, for every post, we should put our stat tracking into the CLI command. Without having the start time, we'd only ever have the duration of the last request anyway (unless I'm totally misunderstanding the purpose of this code).

nickdaugherty (changes requested):
Hey @pschoffer this is a good start, I left some notes inline. I like that the StatsD and Indexables instances have been changed to instance properties of the Search class to make testing easier, that's awesome!

Looking at this code, I'm realizing that we've been calling StatsD wrong - since the methods were `static`, we shouldn't have been calling them on the instance, so it's nice that that has been fixed. I think switching those methods to non-static makes sense, but it could go either way really, since the class itself is stateless. But this way does make it easier to mock the object so that's a win. In general, changes like that are better off in a separate PR to keep the scope small and all changes isolated (so we don't have to review and deploy multiple unrelated changes at the same time), but this PR is pretty small anyway so it's all good.

lib/statsd/class-statsd.php
pschoffer: Good point.

search/includes/classes/class-search.php
pschoffer: Yeah that is difficult. My first implementation was in the command, but there we can't differentiate between full index and/or errors (because we don't get the EP command result within our vip command) - so @brandon-m-skinner  helped me by pointing this section out. 

But Yeah, I am not sure how those bulk requests are created and if they can be split into batches on bigger sets of documents, it would be a problem as well.

nickdaugherty (approved):

nickdaugherty (approved):

pschoffer:
stacks-r1223

PR #1796, opened 2020-10-02
Clear queue of related jobs on index version delete

brandon-m-skinner:
## Description

Presently, deleting an index version doesn't stop the queued job from continuing to be processed. This is pretty inefficient, so we should just delete all jobs for the indexable slug / index version that was deleted.

## Checklist

## Steps to Test

1. Apply PR
2. `lando test --stop-on-failure tests/search/includes/classes/test-class-queue.php`

----------------------

1. Populate the `wp_vip_search_index_queue` table with various jobs with different index versions.
2. Apply PR.
3. `lando wp shell`
4. Execute `\Automattic\VIP\Search\Search::instance()->queue->delete_jobs_for_index_version( <indexable slug>, <index version> );
5. Verify that all the jobs that should be removed are gone and no others.

### DIFF
--- a/search/includes/classes/class-queue.php
+++ b/search/includes/classes/class-queue.php
@@ -819,6 +819,42 @@ class Queue {
 		return intval( $average_wait_time );
 	}
 
+	/**
+	 * Given an indexable slug and an index version, delete all matching jobs from the queue.
+	 *
+	 * Used to clean up after an index version deletion and prevent processing jobs that don't need to be processed.
+	 *
+	 * @param string $indexable_slug An indexable slug.
+	 * @param int | string $index_version An index version.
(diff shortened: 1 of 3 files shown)


nickdaugherty (approved):
This looks good - two other things to consider:

* What happens to already scheduled cron jobs for these now-deleted items?
* What happens during in-flight cron jobs that are in the middle of processing the deleted items?

We just need to be sure the cron job handles any missing jobs gracefully (doesn't throw errors, etc).

brandon-m-skinner:
It seems to handle the missing IDs fine. So in that sense it seems okay.

The issue with version deleting is that it messes up the index version sets and causes errors to be thrown. I've addressed that in a separate PR since it's distinct from just clearing the queue of extraneous jobs.

https://github.com/Automattic/vip-go-mu-plugins/pull/1799

brandon-m-skinner:
r1206-stacks

PR #1799, opened 2020-10-02
Skip indexing job from queue if index version not found

brandon-m-skinner:
## Description

Presently there's no handling for if the index version has been deleted before the jobs are processed from the queue.

They should not be re-indexed and they should not count towards index rate limiting.

## Checklist

## Steps to Test

1. Apply PR.
2. Schedule a process jobs cron job for the queue that contains a list of valid IDs with a non-existent index version number.
3. Run the cron job.
4. There should be no warnings or errors from the process and the index rate limiting count shouldn't increase.

### DIFF
--- a/search/includes/classes/class-queue.php
+++ b/search/includes/classes/class-queue.php
@@ -502,21 +502,25 @@ class Queue {
 			foreach ( $jobs_by_type as $type => $jobs ) {
 				$indexable = $indexables->get( $type );
 
-				\Automattic\VIP\Search\Search::instance()->versioning->set_current_version_number( $indexable, $index_version );
-
-				$ids = wp_list_pluck( $jobs, 'object_id' );
-
-				// If the index version no longer exists, skip incrementing the ratelimiting counter and don't index the ids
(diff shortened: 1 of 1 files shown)


pschoffer:
Would a test for this be a good idea maybe?

It seems there already is a test file for this, just needs one or two more test cases:
tests/search/includes/classes/test-class-queue.php

Buut:
```
public function test_process_jobs() {
	// TODO
}
```

brandon-m-skinner:
Generally I'd agree. Most of my PRs are at least half tests. 

However in this case, adding tests for `process_jobs` should be a separate PR. 

The reasoning for that is:

- The change set would balloon the PR and make the actual point of the PR harder to discern. Small, focused PRs are the goal. If the tests already existed, expanding them for this change would make sense.
- It's part of an already approved change, blocking two PRs.
- The PR is minor and easily testable and based on extensively used production code meaning the risk is low.

nickdaugherty (approved):
Looks correct to me.

brandon-m-skinner:
r1207-stacks

PR #1801, opened 2020-10-05
Update es-wp-query

nickdaugherty:
## Description

Syncs the es-wp-query subtree with Automattic/es-wp-query#master. This is a fix for `is_main_query()` inside the query wrapper class (props @rebeccahum for bringing over this fix from the upstream repo).

## Checklist

## Steps to Test

Example:

1. Check out PR.
1. In code, add logic inside a `pre_get_posts` hook that will only run if the query's `$query->is_main_query()` returns true
1. Set `'es' => true` on the main query
1. Ensure the logic from step 2 now runs (where it didn' before)


nickdaugherty:
r1208-stacks

PR #1806, opened 2020-10-08
Jetpack 9.0 as new default

pschoffer:
## Description

Prepare the PR to roll out the new default Jetpack version

## Checklist

## Steps to Test

* Make sure you dont have pinned Jetpack version - VIP_JETPACK_PINNED_VERSION
* Check Jetpeck version
* change the code to the pr
* See the jetpack version is updated

### DIFF
--- a/jetpack.php
+++ b/jetpack.php
@@ -5,7 +5,7 @@
  * Plugin URI: https://jetpack.com
  * Description: Bring the power of the WordPress.com cloud to your self-hosted WordPress. Jetpack enables you to connect your blog to a WordPress.com account to use the powerful features normally only available to WordPress.com users.
  * Author: Automattic
- * Version: 9.0
+ * Version: 9.0.1
  * Author URI: https://jetpack.com
  * License: GPL2+
  * Text Domain: jetpack
--- a/jetpack.php
+++ b/jetpack.php
@@ -5,7 +5,7 @@
  * Plugin URI: https://jetpack.com
  * Description: Bring the power of the WordPress.com cloud to your self-hosted WordPress. Jetpack enables you to connect your blog to a WordPress.com account to use the powerful features normally only available to WordPress.com users.
  * Author: Automattic
- * Version: 8.9.1
+ * Version: 9.0
  * Author URI: https://jetpack.com
  * License: GPL2+
  * Text Domain: jetpack
@@ -13,7 +13,7 @@
  */
 
 if ( ! defined( 'VIP_JETPACK_DEFAULT_VERSION' ) ) {
-	define( 'VIP_JETPACK_DEFAULT_VERSION', '8.9' );
+	define( 'VIP_JETPACK_DEFAULT_VERSION', '9.0' );
 }
 
 // Bump up the batch size to reduce the number of queries run to build a Jetpack sitemap.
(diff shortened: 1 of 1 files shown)


pschoffer:
NOT to be merged until 15 UTC

jetpack.php
```diff
@@ -5,15 +5,15 @@
  * Plugin URI: https://jetpack.com
  * Description: Bring the power of the WordPress.com cloud to your self-hosted WordPress. Jetpack enables you to connect your blog to a WordPress.com account to use the powerful features normally only available to WordPress.com users.
  * Author: Automattic
- * Version: 8.9.1
+ * Version: 9.0
```
netsuso, line 8: We should put 9.0.1 here, as it's the actual version we are going to deploy

netsuso (changes requested):

jetpack.php
pschoffer: Yeah I was thinking about it, but then when we will deploy 9.0.2, will we update this file as well?

GaryJones: If someone uses a constant to pin to a different version, and they did `wp plugin list`, would it be the version value on line 8 here, or the version number from the version they've pinned to?

i.e. now that vip-go-mu-plugins includes more versions of Jetpack, should this file be treater as a "Jetpack Loader for VIP" plugin, and then have it's own version (`1.0.0`)?

pschoffer: Good point. I tried it out and it shows version from this file, seems regardless of what version is loaded (be it as default or pinned). 

```
wp plugin list --allow-root
+-------------------------+----------+-----------+---------+
| name                    | status   | update    | version |
+-------------------------+----------+-----------+---------+
| jetpack                 | must-use | none      | 9.0     |
```

pschoffer: > i.e. now that vip-go-mu-plugins includes more versions of Jetpack, should this file be treater as a "Jetpack Loader for VIP" plugin, and then have it's own version (1.0.0)?

We should also move the file to something like `jetpack-loader.php` to show it as jetpack-loader in the plugin list rather than jetpack, or?

pschoffer, line 8:
```suggestion
 * Version: 9.0.1
```

pschoffer: I have updated the version to be `9.0.1` for now, as maybe we would like to do the jetpack-loader change as a separate change. Maybe discuss more broadly?

GaryJones: Agreed, should be handled outside of this PR. I just wanted to draw attention to the potential issue as I saw it and thought about it, and get it in your collective radar.

> We should also move the file to something like jetpack-loader.php to show it as jetpack-loader in the plugin list rather than jetpack, or?

Shouldn't be necessary to rename the file - just the "Plugin Name: Jetpack by WordPress.com" header, though if changing that name, it would likely then make sense to rename the file as well.

netsuso: @GaryJones I agree with you, I also raised this concern before. We could even just leave it as plugin `jetpack` but with no version. It's a pity that WordPress doesn't allow to set the version number dynamically...

netsuso (approved):

pschoffer:
stacks-r1220

PR #1807, opened 2020-10-08
Set a max limit on posts_per_page for a JP infinite scroll request.

WPprodigy:
See https://github.com/Automattic/jetpack/blob/ad8ec79d5b67572e4b6d10ac8e08a542186c4032/modules/infinite-scroll/infinity.php#L249-L273

Setting reasonable limits to how many posts we fetch with infinite scroll.


vip-jetpack/vip-jetpack.php
```diff
@@ -372,3 +372,10 @@ function vip_jetpack_admin_enqueue_scripts() {
 }
 
 add_action( 'admin_enqueue_scripts', 'vip_jetpack_admin_enqueue_scripts' );
+
+/**
+ * Set a max limit on posts_per_page for an infinite scroll request. Prevent client-side abuse.
+ */
+add_filter( 'infinite_scroll_posts_per_page', function ( $posts_per_page ) {
+	return (int) $posts_per_page > 5000 ? 5000 : $posts_per_page;
```
rinatkhaziev, line 380:
```suggestion
	return min( (int) $posts_per_page, 5000 );
```

rinatkhaziev (approved):
Looks good, left a minor suggestion, but it's not-blocking though.

PR #1809, opened 2020-10-12
Update/request counters

pschoffer:
## Description

Follow up on - https://github.com/Automattic/vip-go-mu-plugins/pull/1795 - should be ONLY merged after 1795

It adds statsd counter metric for each kind of request.

## Checklist

This is kinda tough to test on sandbox.

in the local environment you can do:
1. `lando wp vip-search index`
1. `lando logs -s statsd ` and see the new counter collected.


pschoffer:
stacks-r1225

PR #1814, opened 2020-10-15
Search: Update ElasticPress submodule to f0ffd49

rinatkhaziev:
This PR combines several performance improvements when doing bulk indexing:

- Implement ID ranges instead of relying on OFFSET when doing pagination. OFFSET gets progressively slower on larger datasets, using ID ranges provides consistent and fast results.
- Term preparation refactoring
- date_18in usage improvements (reduce the number of total calls)

https://github.com/Automattic/ElasticPress/pull/66


nickdaugherty (approved):

rinatkhaziev:
r1233-stacks

PR #1815, opened 2020-10-15
Filter the global alias for cross site searching

nickdaugherty:
##  Description

We filter index names to match our expected naming convention, but not the global (network) alias (which is used to search across all subsites). This PR adds filtering for the alias so that it matches our naming conventions.

Depends on https://github.com/Automattic/ElasticPress/pull/67

## Checklist

## Steps to Test

1. Setup a multisite test environment
1. Configure EP to be in network mode - `define( 'EP_IS_NETWORK', true );`
1. Tell EP to search across all subsites in the network: `add_filter( 'ep_search_scope', function() { return 'all'; } );`
1. Perform a search - `?s=anything`
1. In the Debug Bar panel, check the queried index name - it should look like `vip-<site_id>-post-all`


nickdaugherty:
Hmm @netsuso there are unexpected changes in the EP submodule diff - it looks like the new indexing improvements are included, but I thought those were already merged? Either way, we should update the EP submodule in a separate PR, then merge `master` into this.

netsuso (approved):
Tested and working! I've also updated elasticpress, so this is ready to merge

PR #1816, opened 2020-10-19
Search: Consider stuck `running` jobs as deadlocked too

nickdaugherty:
## Description

`scheduled` jobs are those that have been sent to a cron job, but jobs can also get stuck `running` if they are picked up but never completed (like when process is interrupted), so we need to check for both.

## Checklist

- n/a This change has relevant documentation a`dditions / updates (if applicable).

## Steps to Test

There is test coverage, so manual testing not strictly necessary. But could be done like this:

1. Check out PR.
1. Insert a row into the `wp_vip_search_index_queue` table with a status of `running` and a `scheduled_time` in the distant past
1. Let the sweeper cron job run
1. Ensure the job has now been processed

### DIFF
--- a/search/includes/classes/class-queue.php
+++ b/search/includes/classes/class-queue.php
@@ -458,7 +458,7 @@ class Queue {
 		$deadlocked_time = time() - self::DEADLOCK_TIME;
 
 		$jobs = $wpdb->get_results( $wpdb->prepare(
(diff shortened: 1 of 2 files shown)


brandon-m-skinner (approved):

nickdaugherty:
r1239-stacks

PR #1819, opened 2020-10-19
Add VIP Request Block utility library

rinatkhaziev:
## Description

Previously we've lacked a simple way to block by IP or User-Agent on the application level, resulting in us committing snippets akin to what's in `block_and_log` method. It works fine, but to make it easier to perform this type of a block we decided to concoct this little utility class.

It has 3 main methods:

`VIP_Request_Block::ip( string $ip )` blocks a request based on IP address. First, we try to block by `True-Client-IP` which has to be set in a correct reverse proxy configuration. If that doesn't match, we try `X-Forwarded-For`.

```
VIP_Request_Block::ip( '13.37.13.37' );
```

`VIP_Request_Block::ua( string $ip )` blocks based on User-Agent.

```
VIP_Request_Block::ua( 'Fuzz Faster U Fool v1.337' );
```

`VIP_Request_Block::header( string $header, string $value )` blocks based on a value of any header. 

```
VIP_Request_Block::header( 'x-my-header', 'my-header-value' );
```

This should cover the majority of the cases.

NB: for the best results the requests should be blocked in `vip-config/vip-config.php` which is loaded before the core, thus minimizing the number of wasted CPU cycles.

There are two other places where we need to implement changes in order to get a complete solution:

1. Add the following call in wp-config:

```
// Load VIP_Request_Block utility class, if available
if ( file_exists( ABSPATH . '/wp-content/mu-plugins/lib/class-vip-request-block.php' ) ) {
	require_once ABSPATH . '/wp-content/mu-plugins/lib/class-vip-request-block.php';
}
```

2. Add example usage in [vip-go-skeleton](https://github.com/Automattic/vip-go-skeleton) template repo.

## Checklist

## Steps to Test

Testing locally is fairly straightforward: 

1. Check out PR.
1. Add the above `require_once` block somewhere in local wp-config.php AFTER the definition of `ABSPATH`
1. Set the `$_SERVER['HTTP_X_FORWARDED_FOR'] = '1.2.3.4'` somewhere in `vip-config.php`
1. Visit the test site and verify that you get a 410


pschoffer:
Works great. I tested with `VIP_Request_Block::ip( myIP );` and `VIP_Request_Block::header( 'x-my-header', 'my-header-value' );`.

pschoffer (approved):

pschoffer:
Stacks-r1244

PR #1824, opened 2020-10-21
Solve a memory leak in ElasticPress when WP_DEBUG is enabled

netsuso:
##  Description

This is not exactly a memory leak, it's just a log of all Elasticsearch queries that ElasticPress stores for debugging purposes. If not explicitly defined with `EP_QUERY_LOG`, it gets automatically enabled when `WP_DEBUG` is set to true.

But for long executing CLI commands it doesn't make sense to store this log even if debugging is enabled, as it takes a lot of memory and it's not being used for anything.

## Checklist

## Steps to Test

1. Check out master
1. Try to index a big site on a Go sandbox (WP_DEBUG will be set to true, don't enforce it to false). It will leak memory
1. Check out this PR.
1. Index the same site again. It won't leak any memory (unless it's a problem with the site code)

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -135,6 +135,17 @@ class Search {
 		if ( ! defined( 'EP_DASHBOARD_SYNC' ) ) {
 			define( 'EP_DASHBOARD_SYNC', false );
 		}
+
+		// Disable DB and ES query logs for CLI commands to keep memory under control
+		if ( defined( 'WP_CLI' ) && WP_CLI ) {
+			if ( ! defined( 'SAVEQUERIES' ) ) {
+				define( 'SAVEQUERIES', false );
+			}
+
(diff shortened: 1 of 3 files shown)


rinatkhaziev (dismissed):
Thanks again for the catch!

I think it's better to be a bit more explicit and define the constant in [Search::setup_constants](https://github.com/Automattic/vip-go-mu-plugins/blob/master/search/includes/classes/class-search.php#L116)  (and remove the same constant from the Health command) and wrap it in a `defined( 'WP_CLI' ) && WP_CLI` check, this way we'll solve it once and for all and won't be at a risk of introducing the same issue again in the future.

netsuso:
@rinatkhaziev Totally agree! Much better as you proposed, I've just pushed the change 😊

brandon-m-skinner (approved):

PR #1827, opened 2020-10-24
Add wp vip cache purge-url subcommand

rinatkhaziev:
## Description

Currently, we have a rather nuclear option to purge the whole site cache from Varnish. It is needed sometimes, but more often than not all is really needed is to clear just one or two URLs.

This PR addresses that by adding `wp vip cache purge-url` command that does exactly that.

Additionally, this PR introduces a confirmation prompt when purging the cache for the whole site.

## Checklist

## Steps to Test

1. Checkout the PR
2. Sandbox a site
3. Hit a URL on sandbox, make sure it got cached.
4. Run `wp vip cache purge-url <URL>`
5.  Hit the URL again, make sure `x-cache: miss` header is present.


wp-cli/vip-cache.php
```diff
@@ -13,6 +13,32 @@ function purge( $args, $assoc_args ) {
...
+	 */
+	public function purge_url( $args, $assoc_args ) {
+		if ( ! wp_http_validate_url( $args[0] ) ) {
+			WP_CLI::error( 'Invalid URL:' . $args[0] );
+		}
+
+		// Queue a specific URL purge
+		wpcom_vip_purge_edge_cache_for_url( $args[0] );
+		if ( defined( 'WPCOM_SANDBOXED' ) && WPCOM_SANDBOXED ) {
+			WP_CLI::line( 'Because you are sandboxed this has only purged the Varnish cache on your sandbox host, un-sandbox and use the button in the site admin area if you need to purge the VIP Go edge cache.' );
```
WPprodigy, line 38:
```suggestion
			WP_CLI::line( 'This has only purged the Varnish cache on your sandbox host, un-sandbox if you need to purge the production VIP Go edge cache.' );
```

wp-cli/vip-cache.php
```diff
@@ -13,6 +13,32 @@ function purge( $args, $assoc_args ) {
...
+		if ( ! wp_http_validate_url( $args[0] ) ) {
+			WP_CLI::error( 'Invalid URL:' . $args[0] );
+		}
+
+		// Queue a specific URL purge
+		wpcom_vip_purge_edge_cache_for_url( $args[0] );
+		if ( defined( 'WPCOM_SANDBOXED' ) && WPCOM_SANDBOXED ) {
+			WP_CLI::line( 'This has only purged the Varnish cache on your sandbox. If you need to purge the production caches, please un-sandbox first and then run this command again..' );
+		}
+		WP_CLI::success( 'URL purged from Varnish!' );
```
mjangda, line 40:
```suggestion
		WP_CLI::success( 'URL purged from the VIP page cache,' );
```

mjangda (approved):

GaryJones:
It seems that someone may accidentally use `wp cache purge https://someurl` instead of `wp cache purge-url https://someurl`. The result may then be the page cache for the whole site may get purged instead of just a single URL.

Worth adding in some protection to the existing `purge` command that means that `purge-url` command is actually called instead, if a URL arg is passed?

rinatkhaziev:
@GaryJones - good points, I've added the confirmation prompt and made the command silently pass through to `purge_url` if there's an argument present and it's a valid URL.

brandon-m-skinner (approved):

WPprodigy:
> This has only purged the Page Cache on your sandbox. If you need to purge the production caches, please un-sandbox first and then run this command again.

This means we'll never be able to use the CLI from an actual sandbox, but only via VIP CLI.

Would it possible to forward the request through to the production varnish caches even from a sandbox?

rinatkhaziev:
> Would it possible to forward the request through to the production varnish caches even from a sandbox (maybe with a flag)?

I think it's a Systems thing, not much we can do about it at the moment.

brandon-m-skinner:
r1272-stacks

PR #1828, opened 2020-10-26
Remove curly braces syntax for array access

trepmal:
The array and string offset access syntax using curly braces is deprecated as of PHP 7.4

https://www.php.net/manual/en/migration74.deprecated.php#migration74.deprecated.core.array-string-access-curly-brace

This will begin throwing a fatal error in PHP 8

https://github.com/php/php-src/blob/107962208a19d8b6dc1a190cb25fc37614411e71/UPGRADING#L198


mjangda (approved):

mjangda:
r1251-stacks

PR #1831, opened 2020-10-29
Debug Bar 1.1

mjangda:
https://plugins.trac.wordpress.org/changeset/2407732/debug-bar

- Add a new panel for HTTP requests using the native WP methods.
- AMP dev mode compatibility.
- Change cache busters to use filemtime instead of hardcoded values.
- Minor CSS updates, so theems don't override the `<pre>` styles.


brandon-m-skinner:
Skipped this version and jumped straight to 1.1.2

PR #1833, opened 2020-11-02
Add/test jquery updates

pschoffer:
## Description

Adds [Test JQuery Updates ](https://wordpress.org/plugins/wp-jquery-update-test/) plugin into `mu-plugins`. The plugin is enabled if `VIP_ENABLE_TEST_JQUERY_UPDATES` constant is set.

Note that the current version is meant for WP >5.6.X, so it won't show anything on lower versions.

To get 5.5.x version you can run to get back to that revision (note that this will create git commits in your repo so please revert those afterwards):
`./bin/update-test-jquery-updates.sh b9e54c55bbadcb21fb606fa185146f314262b3c8` 
 

## Checklist

## Steps to Test

1. checkout the pr
1. run `./bin/update-test-jquery-updates.sh b9e54c55bbadcb21fb606fa185146f314262b3c8`
1. See the new admin menu under `plugins`


bin/update-test-jquery-updates.sh
```diff
@@ -0,0 +1,39 @@
...
+revision="trunk"
+if [ "$1" ]; then
+  revision=$1
+fi
+
+directory="test-jquery-updates"
+
+if [ -d "$directory" ]; then
+  print_heading "Removing $directory directory before update"
+  rm -rf $directory
```
pschoffer, line 27: Because we need to mess with the code (see lower about `install_plugins`), the git history gets all messed up. So I end up removing the folder and recreating on every update

bin/update-test-jquery-updates.sh
```diff
@@ -0,0 +1,39 @@
...
+  git commit -avm "Removing $directory directory before update"
+fi
+
+print_heading "Adding test-jquery-updates to $directory from revision $revision"
+git subtree add --squash -P $directory git@github.com:WordPress/wp-jquery-update-test.git "$revision"
+
+
+print_heading "Change capability required from install_plugins (not allowed on VIP platform) to activate_plugins"
+sed -i 's/install_plugins/activate_plugins/g' "$directory/class_wp_jquery_update_test.php"
```
pschoffer, line 36: The plugin requires install_plugins capability which from what I found is not allowed on our platform. I replaced all the checks to activate_plugin instead. Happy to hear any inputs on this. I am not really sure if this is the best way to approach it or even if I understood the issue well.

pschoffer: More info in slack conversation: 
https://a8c.slack.com/archives/C03CU7X3U/p1604306651004800

nickdaugherty (approved):
I haven't tested this, but the code looks good. Let's hold off merging until code freeze is complete, though.

mjangda:
Thanks for working on this @pschoffer!

Three thoughts:

1) I wonder if it's worth putting the plugin in a subfolder like `debug` or something (we could start to move other debug/testing tools there too). The repo is getting pretty cluttered and messy to navigate with so many top-level files and folder and it'd be nice to work towards some cleanliness.

2)

> The plugin is enabled if VIP_ENABLE_TEST_JQUERY_UPDATES constant is set.

I'm wondering if there's a way we also can allow for testing via querystring. For example, adding something like `?vip_test_jquery=true` would load the plugin and toggle the necessary changes so you can quickly verify if things will break.

My thinking is that the easier we can make this to use, the more likely customers are to benefit from it.

3)

> Note that the current version is meant for WP >5.6.X, so it won't show anything on lower versions. 

Does this mean that for customers to use the plugin, they would need their sites updated to the beta? Is there a way to make it work on both 5.5 and 5.6?

pschoffer:
HI @mjangda thanks for checking it out!

1)
> I wonder if it's worth putting the plugin in a subfolder like debug or something (we could start to move other debug/testing tools there too). The repo is getting pretty cluttered and messy to navigate with so many top-level files and folder and it'd be nice to work towards some cleanliness.

Sure I don't really have a preference here. Both sound good.

2)
> I'm wondering if there's a way we also can allow for testing via query string. For example, adding something like ?vip_test_jquery=true would load the plugin and toggle the necessary changes so you can quickly verify if things will break.
My thinking is that the easier we can make this to use, the more likely customers are to benefit from it.

Yeah It was mentioned in the ticket but I thought of maybe trying it out in this simple form first and then build on top of it? I will add it to the ticket so we can get back to it (there is already project skeleton + docs to do.) - https://vipjira.atlassian.net/browse/CANTINA-93.

Also what do you mean by by toggle changes? Looking at the Test JQuery Updates dashboard seems the defaults are sensible. Buuut I would lie to say I understand how the JQuery updates work :) 

image


3) 
> Does this mean that for customers to use the plugin, they would need their sites updated to the beta? Is there a way to make it work on both 5.5 and 5.6?

To be honest I don't fully understand how the plugin works. But it seems it comes packaged with Jquery that is valid only for a single given WP version at any time. What we could do though is have multiple versions-folders of the plugin similar to what Suso did with Jetpack? And then load the correct one based on `$GLOBALS['wp_version']` It is slightly more difficult because the plugin doesn't really have any specific tags per WP version and release cycles it is just going through the commit history to see when the requirements changed. Then putting it to `debug` folder makes a lot more sense I think. What do you think?

pschoffer:
I tried out the smarter logic in choosing the correct version of the plugin mentioned above.

What do you think? @mjangda @nickdaugherty @brandon-m-skinner

debug.php
```diff
@@ -0,0 +1,28 @@
...
+	$avaiable_options = array_diff( scandir( $base_folder ), [ '.', '..' ] );
+
+	$best_fit;
+	foreach ( $avaiable_options as $candidate ) {
+		if ( version_compare( $GLOBALS['wp_version'], $candidate, '>=' ) ) {
+			if ( ! $best_fit || version_compare( $candidate, $best_fit, '>' ) ) {
+				$best_fit = $candidate;
+			}
+		}
+	}
```
mjangda, line 26: What do you think about just having a hard-coded mapping and avoiding the added overhead from scandir + version_compare?

debug.php
```diff
@@ -0,0 +1,28 @@
...
+ * Description: A set of debug tools and plugins
+ */
+
+
+/**
+ * Test Jquery Updates
+ *
+ * There are multiple versions per WordpressVersion the loader tries to pick the most recent one
+ */
+if ( ( defined( 'VIP_ENABLE_TEST_JQUERY_UPDATES' ) && VIP_ENABLE_TEST_JQUERY_UPDATES ) ) {
```
mjangda, line 13: We can add the querystring check here to dynamically load as needed:

```suggestion
$has_test_jquery_param = isset( $_GET['vip_load_test_jquery_updates'] );
$has_test_jquery_constant = defined( 'VIP_ENABLE_TEST_JQUERY_UPDATES' ) && true === VIP_ENABLE_TEST_JQUERY_UPDATES;
if ( $has_test_jquery_param || $has_test_jquery_constant ) {
```

debug.php
```diff
@@ -0,0 +1,28 @@
...
+	$best_fit;
+	foreach ( $avaiable_options as $candidate ) {
+		if ( version_compare( $GLOBALS['wp_version'], $candidate, '>=' ) ) {
+			if ( ! $best_fit || version_compare( $candidate, $best_fit, '>' ) ) {
+				$best_fit = $candidate;
+			}
+		}
+	}
+	require_once( "$base_folder/$best_fit/$entrypoint_script" );
+}
```
mjangda, line 28: Cleanup the global vars:

```suggestion
}

unset( $has_test_jquery_param, $has_test_jquery_constant );
```

mjangda (commented):
Looking good; left a few suggestions.

debug.php
pschoffer: I kinda like that this doesn't need any interaction if we would add a version like 5.5.3 >, and so on. Especially considering the test-jquery-plugins is not strictly versioned.

The downside being performance and a bit more complex code. For complex code, I think the variables naming is alright to make it easier to reason about.

Performance I think we only gonna use this in debug sessions where the performance is not the priority. Did I get that correctly?

debug.php
pschoffer: Sweet, thanks I haven't look into how to access the request, but this seems super simple:) (not sure what I was worried about). I will test it locally and with CLI (where there is no request).

pschoffer:
Thank @mjangda  for the query change. It works great.

The only issue I noticed is that it is not persisting when following links. That is expected. Is there any feature of `mu-plugins` or `core` to register sort of pass-through parameters? E.g. in wp-admin when you click to go the Test Jquery Updates page, it breaks as the plugin is no longer loaded. If not then it is probably not worth worrying about:)

pschoffer:
@mjangda  I added the default settings. It gets a bit tricky as the default settings we would want to differ by the WP and plugin versions. But the current code seems to get things working as we would like.

bin/update-test-jquery-updates.sh
```diff
@@ -0,0 +1,58 @@
...
```
brandon-m-skinner, line 55: Minor change so it works with BSD sed as well as GNU sed.

```suggestion
sed -i.bak 's/install_plugins/activate_plugins/g' "$directory/class_wp_jquery_update_test.php"
```

brandon-m-skinner (changes requested):
Looking good overall, just a minor change to make sed work locally for the Mac users in the house.

bin/update-test-jquery-updates.sh
pschoffer: Perfect thanks for checkint it out! I tried on my machine and it works as well. It creates the backup file though, which we probably don't want to checked in so I added a .gitignore file in that folder to ignore those.

brandon-m-skinner (approved):

brandon-m-skinner:
r1273-stacks

PR #1834, opened 2020-11-02
Fix Failing core tests

pschoffer:
## Description

Looking into new core test failure:
```
in ests/phpunit/tests/xmlrpc/wp/newComment.php

Cannot modify header information - headers already sent by (output started at /tmp/wordpress-develop/tests/phpunit/includes/bootstrap.php:122)
```

Our code adds a header on login failure, but the test isn't setup to work with headers (or http requests in general). The authentication fails even without our plugin and is expected in the test. So I am adding that test class to ignored ones. 

## Checklist

## Steps to Test

Check the circleCI build pipelines

### DIFF
--- a/bin/utils.sh
+++ b/bin/utils.sh
@@ -92,6 +92,7 @@ exclude_core_tests() {
 		"tests/xmlrpc/wp/getMediaItem.php"
 		"tests/xmlrpc/wp/getComments.php"
 		"tests/xmlrpc/wp/getComment.php"
+		"tests/xmlrpc/wp/newComment.php"
 		"tests/xmlrpc/wp/editTerm.php"
 		"tests/xmlrpc/wp/editProfile.php"
 		"tests/xmlrpc/wp/editPost.php"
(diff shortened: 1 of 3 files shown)


nickdaugherty:
@pschoffer did this just start failing recently? If so, what was the changeset where it started failing?

pschoffer:
There were new tests added in the core in recent releases that started to break. 

Some more info is in - https://vipjira.atlassian.net/secure/RapidBoard.jspa?rapidView=82&modal=detail&selectedIssue=CANTINA-166 that I failed to link:)

nickdaugherty (approved):
Discussed with @pschoffer on Slack and we couldn't come up with a good solution for the `header()` issue, so we can just disable for the time being. He's going to investigate a few ideas, otherwise we can go forward with this.

bin/utils.sh
```diff
@@ -112,5 +112,31 @@ exclude_core_tests() {
...
+		echo "WP_TESTS_DIR was not passed in";
+		exit 1;
+	fi
+
+
+	local header_handling_locations=(
+		"$wp_tests_dir/tests/xmlrpc/wp/newComment.php:Tests_XMLRPC_wp_newComment"
+	)
+
+	local header_handling='\/**\n * @runTestsInSeparateProcesses\n * @preserveGlobalState disabled\n *\/'
```
pschoffer, line 128: I tried to add `@runInSeparateProcess` per each test that needs it, but it hanged forever untill it timed-out. This seems to work better

nickdaugherty (approved):
Nice! I think this will work for us.

PR #1835, opened 2020-11-02
Bump phpcodesniffer-composer-installer to 0.7

mjangda:
Which supports both Composer 1 and 2.

Fixes this:

```
Problem 1
    - dealerdirect/phpcodesniffer-composer-installer is locked to version v0.6.2 and an update of this package was not requested.
    - dealerdirect/phpcodesniffer-composer-installer v0.6.2 requires composer-plugin-api ^1.0 -> found composer-plugin-api[2.0.0] but it does not match the constraint.
```

See https://github.com/Dealerdirect/phpcodesniffer-composer-installer/issues/119

Previously #1829

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,8 +57,6 @@ job-references:
       sudo apt-get update && sudo apt-get install subversion libgcc-8-dev default-mysql-client
       sudo -E docker-php-ext-install mysqli
       npm ci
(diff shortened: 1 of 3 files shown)


brandon-m-skinner (approved):

mjangda:
Failing core tests are unrelated. Also no deploy needed for this.

PR #1836, opened 2020-11-03
Show plugin update notices on the plugins page.

mjangda:
Core restricts showing plugin update notices to users with update_plugins, which is not granted to anyone on VIP.

Instead, copy core's wp_plugin_update_rows() logic and use the manage_options cap instead.

There are other checks further down the stack to prevent users without the update_plugins cap to actually trigger updates.

## Checklist

## Steps to Test

1. Check out PR.
1. Go to `wp-admin/plugins.php`
1. Install an old version of a plugin (e.g. goto https://en-ca.wordpress.org/plugins/edit-flow/advanced/ -- then scroll to bottom and download an old version)
1. Reload plugins page.

The newly added plugin should now have a notice saying a new version is available.


001-core/updates.php
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Show plugin update notices on the plugins page.
+ *
+ * This is basically a copy of Core's wp_plugin_update_rows() function.
+ * The only difference is that we use the manage_options cap instead of update_plugins.
+ * Because no one on VIP has update_plugins, core's notices are never displayed.
+ * With this change, we make them visible.
+ */
+function show_plugin_update_notices() {
+	if ( ! current_user_can( 'manage_options' ) ) {
```
mjangda, line 14: For correctness, should use the `activate_plugins` cap.

```suggestion
 * The only difference is that we use the activate_plugins cap instead of update_plugins.
 * Because no one on VIP has update_plugins, core's notices are never displayed.
 * With this change, we make them visible.
 */
function show_plugin_update_notices() {
	if ( ! current_user_can( 'activate_plugins' ) ) {
```

GaryJones: Wouldn't `update_plugins` ([noted here](https://github.com/WordPress/WordPress/blob/3da046e1bce8408a1b6cd08e03eaf69a6bb506cd/wp-includes/capabilities.php#L439)) by more accurate - i.e. only show notices about plugin updates to those who have the capability update plugins? Activating !== Updating.

GaryJones:
Instead of using a different (technically un-related) capability, would it be better to give suitable roles the `update_plugin` capability right before these notices would be shown, and then remove it straight after?

mjangda:
> Instead of using a different (technically un-related) capability, would it be better to give suitable roles the update_plugin capability right before these notices would be shown, and then remove it straight after?

Good thought, but we'd have to do it in multiple spots and would likely get very complex (e.g. we still need to restore the number bubble in the admin menu).

Also to be clear, this is just about providing a notice about the update (not allowing the update since that happens via GitHub) so it's debatable whether `update_plugin` would be the correct cap to use. Also arguably, this is a bug in core since `wp_plugin_update_rows` should probably drop the cap check (since there are more appropriate cap checks further down the code stack that remove actions; e.g. if you don't have update_plugin, it just shows a notice; if you have update_plugin, it shows an update button).

GaryJones:
> it's debatable whether update_plugin would be the correct cap to use

That's true.

Maybe add a custom cap (`'see_plugin_updates_notice'`) to Admin roles then, and use that here? That gives customers more flexibility about who should see these notices.

WPprodigy:
Closing in favor of https://github.com/Automattic/vip-go-mu-plugins/pull/3078

PR #1839, opened 2020-11-05
Adds admin notice plugin

pschoffer:
#  Description

First draft of the Notice plugin, that will enable us to add notices to wp-admin page

## Checklist

## Steps to Test

Update the code and add a new notice.

Check the admin page and see the new notice.


admin-notice/class-admin-notice-controller.php
```diff
@@ -0,0 +1,52 @@
...
+		];
+	}
+
+	public function display_notices() {
+		$filtered_notices = $this->filter_notices_by_time( $this->all_notices );
+
+		$html_notices = array_map( [ $this, 'convert_notice_to_html' ], $filtered_notices );
+
+		foreach ( $html_notices as $html_notice ) {
+			echo( $html_notice . "\n" );
```
pschoffer, line 37: Not sure how to correctly escape this? `esc_html` will strip the html code, but that is not what I want. The input variables were escaped in  `convert_notice_to_html`.

admin-notice/class-admin-notice-controller.php
```diff
@@ -0,0 +1,52 @@
...
+	private $notice_class = 'notice notice-info';
+	public $all_notices = [];
+
+	public static function init() {
+		$instance = new Admin_Notice_Controller();
+
+		add_action( 'admin_notices', [ $instance, 'display_notices' ] );
+
+		$instance->all_notices = [
+			new Admin_Notice( 'WordPress 5.5.2 will be released on Friday, October 30th', '01-07-2020', '30-10-2020 15:00' ),
```
pschoffer, line 27: Not sure if this is the best way of defining the notices? 

I was thinking about having them in separate file, but I am not sure if I should split it either. What do you think?

admin-notice/admin-notice.php
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Plugin Name: Admin Notice
+ * Description: Adds dismisable notice in admin area
+ * Version:     0.1.0
+ * Author:      Automattic VIP
+ * Author URI:  https://wpvip.com
+ * License:     GPLv2 or later
+ *
+ * @package Automattic\VIP\Admin_Notice
+ */
```
GaryJones, line 11: Why does these plugin headers need to be duplicated here?

admin-notice/class-admin-notice-controller.php
```diff
@@ -0,0 +1,52 @@
...
+
+namespace Automattic\VIP\Admin_Notice;
+
+// function sample_admin_notice__error() {
+//     $class = 'notice notice-info';
+//     $message = __( 'Irks! An error has occurred.', 'sample-text-domain' );
+
+//     printf( '<div class="%1$s"><p>%2$s</p></div>', esc_attr( $class ), esc_html( $message ) );
+// }
+// add_action( 'admin_notices', 'sample_admin_notice__error' );
```
GaryJones, line 11: If this is meant to be an example of how to use it, then either put it as a proper DocBlock, or move it to a readme.

admin-notice/class-admin-notice-controller.php
```diff
@@ -0,0 +1,52 @@
...
+
+
+require_once __DIR__ . '/class-admin-notice.php';
```
GaryJones, line 14: Keep class definition and side-effect code to their own files. Move this require into the admin-notices/admin-notices.php file. The responsibility of this file is to define the class, and not to ensure other files are loaded.

admin-notice/class-admin-notice-controller.php
```diff
@@ -0,0 +1,52 @@
...
+
+
+class Admin_Notice_Controller {
+	private $notice_class = 'notice notice-info';
+	public $all_notices = [];
+
+	public static function init() {
+		$instance = new Admin_Notice_Controller();
```
GaryJones, line 22:
```suggestion
		$instance = new static();
```

No need to repeat the class name. `new self()` wouldn't support classes that extend this one, and since there's no `final` on the class, you're presumably OK with allowing extended classes.

admin-notice/class-admin-notice-controller.php
GaryJones: Seems like the list of notices should really be defined elsewhere, and not buried in a class of what should be re-usable logic.

admin-notice/class-admin-notice-controller.php
GaryJones: `echo wp_kses_post( $html_notice );`

admin-notice/class-admin-notice-controller.php
GaryJones: Instead of `Admin_Notices::init()` in the base file, you could do something like:

```
$admin_notices = new Admin_Notices();
$admin_notices->add( new Admin_Notice( 'WordPress 5.5.2 will be released on Friday, October 30th', '01-07-2020', '30-10-2020 15:00' ) );
// ... add more notices as needed ...
$admin_notices->render();
```

That would move the data about admin notices into the root of the plugin, avoid having to use a a static init factory method, uses composition instead of highly coupling Admin_Notice and Admin_Notices classes (which then allows for easier mocking in tests), and allows for easier use of decorators to limit who should be shown certain admin messages:

```
$admin_notices = new Admin_Notices();
$admin_notices->add( new Editor_Notice( new Admin_Notice( 'WordPress 5.5.2 will be released on Friday, October 30th', '01-07-2020', '30-10-2020 15:00' ) ) );
// ... add more notices as needed ...
$admin_notices->render();

...

// Should be called Admin_Notice, but using suffix here to make it clearer for the existing code. Concrete class could be Info_Notice, to then allow for Error_Notice or Warning_Notice as well.
interface Admin_Notice_Interface {}

...

// Base Decorator - so override the individual Admin_Notice
class Admin_Notice_Decorator {
	protected $notice;
	public function __construct( Admin_Notice_Interface $notice ) {
		$this->notice = $notice;
	}
}

...

class Editor_Notice extends Admin_Notice_Decorator {
    public function render() {
		if ( current_user_can( 'edit_posts' ) {
			$this->notice->render();
		}
    }
}

```

admin-notice/class-admin-notice.php
```diff
@@ -0,0 +1,24 @@
...
+	public $message;
+	public $start_date;
+	public $end_date;
+
+	/**
+	 * Create AdminNotice
+	 *
+	 * $message - the text to be displayed
+	 * $start_date - the date* since the notice should start to show up
+	 * $end_date - the date* till the notice should show up
```
GaryJones, line 15: Use proper `@param` tags.

admin-notice/class-admin-notice.php
```diff
@@ -0,0 +1,24 @@
...
+	 *
+	 * * Date Format should follow 'Day-Month-Year Hour:Minute' format.
+	 */
+	public function __construct( $message, $start_date, $end_date ) {
```
GaryJones, line 19: Add type declarations to this parameters.

GaryJones (changes requested):

pschoffer:
@GaryJones thanks for the great input! I will implement some changes and get back to you on the rest:)

admin-notice/class-admin-notice-controller.php
pschoffer: Great Thanks! Works like a charm.

admin-notice/class-admin-notice-controller.php
pschoffer: Great! I loved the idea of putting it to the main file!

About the naming, I called the main class `Admin_Notice_Controller`. It seems more fitting than `Admin_Notices` as it is not just a collection, but rather has some more logic.But that could be my Java backage background talking:) Do you think `Admin_Notices` would be a better name?

admin-notice/admin-notice.php
pschoffer: Oh I just thought that is the usual way of doing it when looking into the other plugins. But keeping it in one place is definitely better. I removed it.

admin-notice/class-admin-notice-controller.php
pschoffer: Nah, sorry about that. That was just for me to learn how to do it and I forgot to remove it. Removed now.

admin-notice/class-admin-notice-controller.php
pschoffer: Done

admin-notice/class-admin-notice-controller.php
pschoffer: As per your advice, I am no longer creating an instance here, so I will resolve this comment.

admin-notice/class-admin-notice.php
pschoffer: I updated it. From what I googled this should be now the correct proper form, but feel free to point out more specifically what is wrong (this is the first time I am using those tags).

admin-notice/class-admin-notice.php
pschoffer: Done.

pschoffer:
@GaryJones I pushed an update, addressing most of your comments. It would be great if you could give it another pass.

Regarding the different capabilities needed to see a notice. I was planning to introduce a follow-up PR that would introduce that, but rather than using inheritance I was planning to just add another constructor parameter with a sensible default (what would that be?). That way there is no need to create a new class every time we would like to limit somehow differently, just pass in another capability name - string in the constructor. What do you think?

admin-notice/class-admin-notice-controller.php
GaryJones: You could have both - Notices is the collection, and Controller (if still needed) does things with the collection (like adding and removing single notices).

If the logic is moved to the root file (roughly per my example), then that's really the controlling  logic / composition root there, so may not need a separate controller class.

admin-notice/class-admin-notice.php
GaryJones: Have a read through https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#docblock-formatting for how it should be in WP, and https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/param.html#param (and maybe the [draft PSR-19](https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc-tags.md#59-param)) for general usage.

admin-notice/class-admin-notice.php
```diff
@@ -0,0 +1,24 @@
...
+	/**
+	 * Create AdminNotice
+	 *
+	 * @param {string} $message - the text to be displayed
+	 * @param {string} $start_date - the date* since the notice should start to show up
+	 * @param {string} $end_date - the date* till the notice should show up
+	 *
+	 * * Date Format should follow 'Day-Month-Year Hour:Minute' format.
+	 */
+	public function __construct( string $message, string $start_date, string $end_date ) {
```
GaryJones, line 19: Are there cases where it wouldn't make sense to constrain a notice by time? Think about the notice we show when an account billing is not up do date - we don't know what the end time should be, so while one could enter a date with a far-distant year, that's really a hack of poor class design.

I'm thinking whether it would not be overkill to have Admin_Notice just take in a message and a notice type (make them class constants somewhere) as args, and then have a `Time_Constrained_Notice()` decorator that handled that, along with my previous `Editor_Notice()` (I guess `Role_Constrained_Notice()` or `Capability_Constrained_Notice()`) constraint suggestion.

```php
// Show to everyone, but only for a limited time.

$admin_notice_controller->add(
	new Time_Constrained_Notice(
		new Admin_Notice( 'Happy New Year!', Admin_Notice_Type::INFO ),
		'01-01-2021 00:00',
		'02-01-2021 00:00'
	)
);

// Plugin updates from a different PR. Wrap this in it's own logic of when to show.

$admin_notice_controller->add(
	new Capability_Constrained_Notice(
		new Admin_Notice( 'Plugin updates should be done through GitHub', Admin_Notice_Type::Warning ),
		'update_plugins'
	)
);

// Lockout Error (example).

$admin_notice_controller->add(
	new Capability_Constrained_Notice(
		new Admin_Notice( 'You have not paid your bill.', Admin_Notice_Type::Error ),
		'manage_options'
	)
);
```

admin-notice/class-admin-notice-controller.php
```diff
@@ -0,0 +1,38 @@
...
+			echo( wp_kses_post( $html_notice ) . "\n" );
+		}
+	}
+
+	public function convert_notice_to_html( $notice ) {
+		return sprintf( '<div class="%s"><p>%s</p></div>', esc_attr( $this->notice_class ), esc_html( $notice->message ) );
+	}
+
+	public function filter_notices_by_time( $input_notices ) {
+		$now = new \DateTime( 'now', new \DateTimeZone( 'UTC' ) );
```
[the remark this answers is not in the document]
pschoffer: The tradeoff is the relative increased difficulty to use. When you as a caller have to create DateTime object as a caller and also make sure both decision date and the interval is in the same Timezone.

What do you think about the current code where the dates moved to `Date_Condition` and I provided string parameter for the decision date (, but keeping the DateTime creation and TimeZone management encapsulated)?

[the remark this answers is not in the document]
GaryJones: > When you as a caller have to create DateTime object as a caller and also make sure both decision date and the interval is in the same Timezone.

Good point.

So, maybe have a factory method for the instantiation of the Admin_Notices_Controller then (a bit like a Singleton).

Accepting a `string` on the start / end dates makes sense.

admin-notice.php
```diff
@@ -0,0 +1,12 @@
+<?php
+/**
+ * Plugin Name: Admin Notice
+ * Description: Adds dismisable notice in admin area
```
GaryJones, line 4:
```suggestion
 * Description: Adds dismissible notice in admin area
```

admin-notice.php
```diff
@@ -0,0 +1,12 @@
...
+ * Version:     0.1.0
+ * Author:      Automattic VIP
```
GaryJones, line 6: Either `Automattic` or `WordPress VIP`.

admin-notice/admin-notice.php
```diff
@@ -0,0 +1,21 @@
...
+	exit; // Exit if accessed directly.
+}
+
+require_once __DIR__ . '/class-admin-notice.php';
+require_once __DIR__ . '/class-admin-notice-controller.php';
+require_once __DIR__ . '/conditions/interface-condition.php';
+require_once __DIR__ . '/conditions/class-date-condition.php';
+
+$admin_notice_controller = new Admin_Notice_Controller();
+$admin_notice_controller->init();
```
GaryJones, line 15: What advantage is it to call `init()` here, rather than moving the `add_action()` call here and removing all WordPress-defined function calls out of it, to just have a plain PHP class?

It's a controller - it shouldn't care what application it is part of as it is just moving data around.

admin-notice/class-admin-notice-controller.php
```diff
@@ -0,0 +1,29 @@
...
+		$filtered_notices = array_filter( $this->all_notices, function ( $notice ) {
+			return $notice->should_render();
+		});
+
+		$html_notices = array_map( function ( $notice ) {
+			return $notice->to_html();
+		}, $filtered_notices );
+
+		foreach ( $html_notices as $html_notice ) {
+			echo( wp_kses_post( $html_notice ) . "\n" );
```
GaryJones, line 26: The presence of the WordPress escaping _here_ bugs me.

Rather than having this controller tied to the escaping, consider doing:

```php
public function display_notices() {
	// Get $filtered_notices as currently.

	foreach( $filtered_notices as $notice ) {
		$notice->display();
	}
}
```

That way, we can handle all of the escaping as the HTML markup is being generated and displayed, from within the Admin_Notice itself (`display()` method), so that any class that extends Admin_Notice can adjust it's escaping as needed (perhaps to allow `wp_kses()` with a custom set of elements -> attributes that allow for particular aria-* attributes).

tests/admin-notice/conditions/test-class-admin-notice-time-condition.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+namespace Automattic\VIP\Admin_Notice;
+
+require_once __DIR__ . '/../../../admin-notice/conditions/interface-condition.php';
+require_once __DIR__ . '/../../../admin-notice/conditions/class-date-condition.php';
+
+class Date_Contition_Test extends \WP_UnitTestCase {
```
GaryJones, line 8:
```suggestion
class Date_Condition_Test extends \WP_UnitTestCase {
```

I don't think this codebase does it yet, but if you ever split tests up into pure unit tests vs WP unit tests (pure unit tests don't need to spin up WP so much faster), then this one could just be a pure unit test i.e. it doesn't call or reference any WP-defined code.

GaryJones (changes requested):
Just a couple of changes to the controller to thin it even more, then I think this is good.

I like the implementation of the Date_Condition :-)

admin-notice/class-admin-notice-controller.php
pschoffer: I am still not sold on this as I don't like the side effect of printing to STDOUT. buut you have good points as well and I don't have strong opinion on this one. So I changed it as you suggested.

admin-notice/admin-notice.php
pschoffer: done

admin-notice.php
pschoffer: done

admin-notice/class-admin-notice.php
```diff
@@ -17,10 +17,11 @@ public function __construct( string $message, array $conditions = [] ) {
 		$this->conditions = $conditions;
 	}
 
-	public function to_html() {
-		return sprintf( '<div class="notice notice-info"><p>%s</p></div>', esc_html( $this->message ) );
+	public function display() {
+		return printf( '<div class="notice notice-info"><p>%s</p></div>', esc_html( $this->message ) );
```
GaryJones, line 21: If it's printing immediately, then no need for a `return` here.

admin-notice/class-admin-notice.php
```diff
@@ -0,0 +1,34 @@
...
+		$this->message = $message;
+		$this->conditions = $conditions;
+	}
+
+	public function display() {
+		return printf( '<div class="notice notice-info"><p>%s</p></div>', esc_html( $this->message ) );
+	}
+
+
+	public function should_render() {
```
GaryJones, line 25: DocBock missing.

tests/admin-notice/test-admin-notice-controller.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace Automattic\VIP\Admin_Notice;
+
+class Admin_Notice_Test extends \WP_UnitTestCase {
```
GaryJones, line 5: How about tests when there are multiple Date_Conditions specified in the Admin_Notice arg?

tests/admin-notice/test-class-admin-notice.php
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	public function should_render_data() {
+
+		return [
+			[ [], true ],
+			[ [ false ], false ],
+			[ [ true ], true ],
+			[ [ true, true ], true ],
+			[ [ true, false ], false ],
```
GaryJones, line 27: How about a `[ false, true ]`? Or `[ false, false ]`?

GaryJones (approved):
LGTM :-)

pschoffer:
Stacks-r1260.

PR #1840, opened 2020-11-06
IP Blocks: Match any IP in X-Forwarded-For

joshbetz:
The X-Forwarded-For header can be a list of IPs, separated by commas. In order to catch those cases, we need to parse that list of IPs and evaluate them individually.


lib/class-vip-request-block.php
```diff
@@ -32,8 +32,15 @@ public static function ip( string $value ) {
 		}
 
 		// phpcs:ignore WordPressVIPMinimum.Variables.ServerVariables.UserControlledHeaders, WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___SERVER__HTTP_USER_AGENT__
```
GaryJones, line 34:
```suggestion
		// phpcs:ignore WordPressVIPMinimum.Variables.ServerVariables.UserControlledHeaders
```

lib/class-vip-request-block.php
```diff
@@ -32,8 +32,15 @@ public static function ip( string $value ) {
...
-		if ( isset( $_SERVER['HTTP_X_FORWARDED_FOR'] ) && $value === $_SERVER['HTTP_X_FORWARDED_FOR'] ) {
-			return self::block_and_log( $value, 'x-forwarded-for' );
+		if ( isset( $_SERVER['HTTP_X_FORWARDED_FOR'] ) ) {
+			$ips = explode( ',', $value );
```
GaryJones, line 36: You're checking for whether a heading has been set, but then there's an assumption that what has been passed into this method (`$value`) is indeed the value of that header. That seems broken.

GaryJones (changes requested):

lib/class-vip-request-block.php
rinatkhaziev: I think maybe a simple `stripos` would do in this case? Seems faster than splitting and then comparing each individual IP.

rinatkhaziev: @joshbetz do you want to take another swing at this?

rinatkhaziev:
Superseded by #2080

PR #1845, opened 2020-11-11
Add/admin notice conditions

pschoffer:
## Description

Adds support for `wp-version` and `capabaility` restrictions for admin notices.

## Checklist

## Steps to Test

1. Check out PR
1. Add admin notice in admin-notice/admin-notice.php
```
$admin_notice_controller->add(
	new Admin_Notice(
		'Test',
		[
			... add conditions here
		]
) );
```
1. Add conditions
1. See the notice is shown if all conditions are true doesn show otherwise.


admin-notice/admin-notice.php
```diff
@@ -18,5 +20,8 @@
 $admin_notice_controller->add(
 	new Admin_Notice(
 		'WordPress 5.5.2 will be released on Friday, October 30th',
-		[ new Date_Condition( '01-07-2020', '30-10-2020 15:00' ) ]
+		[
+			new Date_Condition( '01-07-2020', '30-10-2020 15:00' ),
```
netsuso, line 24: Not related to this PR, but I think we should change the date format. In the US they use `mm-dd-yyyy` instead of `dd-mm-yyyy`, so this format will lead to a human error sooner or later.  I suggest `yyyy-mm-dd`, which is much less confusing (and nicer to computers :D)

admin-notice/conditions/class-wp-version-condition.php
```diff
@@ -0,0 +1,35 @@
...
+	 */
+	public function __construct( string $minimal_version, string $maximum_version = null ) {
+		$this->minimal_version = $minimal_version;
+		$this->maximum_version = $maximum_version;
+	}
+
+	public function evaluate() {
+		$current_version = $GLOBALS['wp_version'];
+
+		if ( version_compare( $current_version, $this->minimal_version, '<' ) ) {
```
netsuso, line 25: I think we should also allow this to be null, as a typical case will be "show this notice to all WP versions before x.y". By the way, probably "minimum" is a better name than "minimal", although native speakers will know better 😉

netsuso (changes requested):

admin-notice/conditions/class-wp-version-condition.php
pschoffer: Updated.

admin-notice/admin-notice.php
pschoffer: You are right yyyy-mm-dd is more universal. Both formats are accepted (it is just passed to DateTime constructor), I switched the examples though.

netsuso: Oh, that's right, I hadn't looked at the code. Maybe some phpdoc in the `Date_Condition` constructor would help on that. But whenever possible, if examples are `yyyy-mm-dd` it will be better, thanks! 😊

netsuso (approved):

pschoffer:
Stacks-r1262.

PR #1847, opened 2020-11-12
Add/admin notice dismiss

pschoffer:
## Description

We have recently added admin-notice plugin. This change adds the ability to dismiss them.

## Checklist

## Steps to Test

1. Check out PR
1. Add a notice that shows up with dismiss identifier
1. Click dissmiss buton
1. See that notice disapeared even after reload


admin-notice/class-admin-notice.php
```diff
@@ -5,20 +5,35 @@
...
 	 *
 	 * @param string $message The text to be displayed.
 	 * @param Admin_Notice_Condition[] $conditions Conditions to pass in order to show the notice.
+	 * @param string $dismiss_identifier if provided the dismiss will become dissmisible
 	 */
-	public function __construct( string $message, array $conditions = [] ) {
+	public function __construct( string $message, array $conditions = [], string $dismiss_identifier = '' ) {
+		$position = strpos( $dismiss_identifier, self::COOKIE_DELIMETER );
+		if ( false !== $position ) {
+			\trigger_error( esc_html( "Admin Notice dissmiss identifier - $dismiss_identifier can't contain delimeter - " . self::COOKIE_DELIMETER ) );
```
pschoffer, line 22: Not sure what to do here? Should I log error in another way?

admin-notice/js/script.js
```diff
@@ -0,0 +1,72 @@
...
+}
+
+function registerDismissHooks() {
+    const dismissButtons = document.getElementsByClassName("vip-notice-dismiss");
+    for (const button of dismissButtons) {
+        button.onclick = onDismissClicked;
+    }
+}
+
+function hideDismissedNotices() {
```
pschoffer, line 55: This was originally in server-side php code, but our standard warn me against using cookies on server-side due to caching. So I moved it here

admin-notice/class-admin-notice.php
GaryJones: To log it into Kibana, it needs the second arg passed to bump the default error level.

admin-notice/js/script.js
GaryJones: This code will only be shown when logged in, and requests with logged-in cookies get a `PASS` at the edge, so there wouldn't be any page cache involved here.

pschoffer: Oh I see, Is there a comment I should add to pass linting then? Is that a common practice?

admin-notice/class-admin-notice.php
GaryJones, line 22:
```suggestion
			\trigger_error( esc_html( "Admin Notice dismiss identifier - $dismiss_identifier can't contain delimeter - " . self::COOKIE_DELIMETER ) );
```

admin-notice/class-admin-notice.php
```diff
@@ -5,20 +5,35 @@
...
+		}
+
 		$this->message = $message;
 		$this->conditions = $conditions;
+		$this->dismiss_identifier = $dismiss_identifier ? $dismiss_identifier : '';
```
GaryJones, line 27:
```suggestion
		$this->dismiss_identifier = $dismiss_identifier ?: '';
```

In a ternary condition, if the value being checked is the same as the truthy result, then since  PHP 5.3, the truthy result (lefthand operand) can be skipped.

admin-notice/class-admin-notice.php
```diff
@@ -5,20 +5,35 @@
...
 	}
 
 	public function display() {
-		printf( '<div class="notice notice-info"><p>%s</p></div>', esc_html( $this->message ) );
+		printf( '<div data-vip-admin-notice="%s" class="notice notice-info vip-notice">', esc_html( $this->dismiss_identifier ) );
+		printf( '<p>%s</p>', esc_html( $this->message ) );
+		if ( $this->dismiss_identifier ) {
+			printf( '<button type="button" class="notice-dismiss vip-notice-dismiss"></button>' );
```
GaryJones, line 34: Where is the content of the button? There should be something here for accessibility purposes, even if it is then visually styled differently.

admin-notice/class-admin-notice.php
pschoffer: What should be the correct second arg or how do I find out?

admin-notice/class-admin-notice.php
pschoffer: uuuh that is nice! However, this code is no longer needed it can be just a simple assignment. I forgot to remove it from the previous attempt.

admin-notice/class-admin-notice.php
pschoffer: E_USER_ERROR?

admin-notice/js/script.js
pschoffer: BTW what is your preference? I kinda like that both setting and reading cookie logic is in Javascript in the same file as opposed to having it split between the backend and frontend and there is no need to avoid linting issues. On the other hand, having notice to pop-up before js starts up to hide is less optimal.

admin-notice/class-admin-notice.php
pschoffer: Added content

admin-notice/class-admin-notice.php
GaryJones: It's in our public docs on WPVIP.com.

admin-notice/js/script.js
GaryJones: If someone dismisses a notice, and then loads when JS is not available, does the current solution means that it would show up again for them?

Is there a user meta / site meta field that is set which would avoid a notice from displaying from the back-end? i.e. store the fact a notice is dismissed in the DB, and not in a cookie (which could be wiped at any time).

How does WP core, or other popular plugins (Yoast SEO?) handle dismissible notices?

admin-notice/class-admin-notice.php
pschoffer: Found it thanks! - https://wpvip.com/documentation/manually-logging-errors-in-new-relic/ => `E_USER_WARNING`

admin-notice/js/script.js
pschoffer: I couldn't find a standard way to do this. [Some plugin](https://github.com/Automattic/vip-go-mu-plugins/blob/2f4d90e41975b252425d85e2fd7e5b518c83fb5f/performance/bulk-edit.php#L82) seems to be leveraging `dismissed_wp_pointers` user meta (it has ajax action that can be used to add dismissed pointers). Seems nice but a bit hacky.

Some articles I read mentioned cookies being used in most cases which is in line with the request for this change: `Dismissable (persistently...can be cookie based)` so I went with cookie.

tests/admin-notice/test-class-admin-notice.php
```diff
@@ -17,12 +17,12 @@ public function test__display() {
 		$notice->display();
 	}
 
-	public function test__display_with_button() {
+	public function test__display_dissmissible() {
```
GaryJones, line 20: Typo in test method name.

GaryJones (dismissed):
One small typo, but otherwise LGTM.

tests/admin-notice/test-class-admin-notice.php
pschoffer: Uh my Czenglish betrays me again. What is the correct form? When I googled I found `dismissible - subject to dismissal` which seems to be what I typed, but sometimes I am really blind on these things.

pschoffer: Oh it is the single 's' now I see it:)

pschoffer:
@GaryJones thanks for the patience! I checked some more options and using ajax call with user meta seems to work better than the cookie.

Could you please check again the implementation using the meta instead of cookie?

admin-notice/class-admin-notice-controller.php
```diff
@@ -5,17 +5,44 @@
...
+			'data_attribute' => Admin_Notice::DISMISS_DATA_ATTRIBUTE,
+			'identifier_key' => self::DISMISS_IDENTIFIER_KEY,
+		] );
+	}
+
+	public function dismiss_vip_notice() {
+		check_ajax_referer( self::DISMISS_NONCE_ACTION );
+
+		if ( isset( $_POST[ self::DISMISS_IDENTIFIER_KEY ] ) ) {
+			update_user_meta( get_current_user_id(), self::DISMISS_USER_META, esc_html( $_POST[ self::DISMISS_IDENTIFIER_KEY ] ) );
```
GaryJones, line 44: What is `esc_html()` doing here? Use `sanitize_text_field()` instead.

admin-notice/class-admin-notice.php
```diff
@@ -5,20 +5,33 @@
...
 
 	public function display() {
-		printf( '<div class="notice notice-info"><p>%s</p></div>', esc_html( $this->message ) );
+		$notice_class = 'notice notice-info vip-notice';
+
+		if ( $this->dismiss_identifier ) {
+			$notice_class .= ' is-dismissible';
+		}
+
+		printf( '<div %s="%s" class="%s">', esc_html( self::DISMISS_DATA_ATTRIBUTE ), esc_html( $this->dismiss_identifier ), esc_html( $notice_class ) );
```
GaryJones, line 32:
```suggestion
		printf( '<div %s="%s" class="%s">', esc_html( self::DISMISS_DATA_ATTRIBUTE ), esc_attr( $this->dismiss_identifier ), esc_attr( $notice_class ) );
```

GaryJones (changes requested):

pschoffer:
Thanks @GaryJones , I have applied your suggestions.

pschoffer:
Stacks-r1265.

admin-notice/js/script.js
```diff
@@ -0,0 +1,44 @@
+function tryGetNoticeContainer(currentElement) {
+    if (currentElement === null || currentElement.hasAttribute(dismissal_data.data_attribute)) {
+        return currentElement;
+    }
+    return tryGetNoticeContainer(currentElement.parentElement);
+}
+
+function persistDismissedNotice(containerElement) {
+    const dismissIdentifier = containerElement?.getAttribute(dismissal_data.data_attribute);
```
hrkhal, line 9: @pschoffer @GaryJones  This has introduced regressions into VIP sites. The file isn't being transpiled. It's breaking the admin for some users on older chrome.

https://github.com/Automattic/vip-go-mu-plugins-built/blob/master/admin-notice/js/script.js#L9

PR #1857, opened 2020-11-13
vip-plugins: link to post->permalink instead of post->meta->plugin_url

curtisbelt:
This PR is backwards compatible and will continue to use `meta->plugin_url` if `permalink` is unavailable.

For this PR to work properly, `permalink` value must be added to response that comes from wpvip.com REST API (handled by a separate PR for that site).

### DIFF
--- a/vip-plugins/vip-plugins.php
+++ b/vip-plugins/vip-plugins.php
@@ -1,4 +1,4 @@
-<?php 
+<?php
 
 /**
  * Retrieve featured plugins from the wpvip.com API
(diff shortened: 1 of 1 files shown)


curtisbelt:
The `permalink` value used in this PR is now available on the API (https://wpvip.com/wp-json/vip/v0/plugins?type=technology)

brandon-m-skinner (approved):

brandon-m-skinner:
r1268-stacks

PR #1870, opened 2020-11-18
Dissmiss Notice Event by Onclick

pschoffer:
## Description

Previously the hook was on `DOMNodeRemoved` which cause some other operations (e.g. page unloading) to fire it even if the button was not clicked.

## Checklist

## Steps to Test

1. Create a notice
1. Navigate to wp-admin
1. Refresh few times
1. Notice is still there
1. Dismiss Notice
1. Refresh few times
1. Notice is gone

### DIFF
--- a/admin-notice/js/script.js
+++ b/admin-notice/js/script.js
@@ -1,31 +1,31 @@
 function tryGetNoticeContainer(currentElement) {
-    if (currentElement === null || currentElement.hasAttribute(dismissal_data.data_attribute)) {
-        return currentElement;
-    }
-    return tryGetNoticeContainer(currentElement.parentElement);
+	if (currentElement === null || currentElement.hasAttribute(dismissal_data.data_attribute)) {
+		return currentElement;
+	}
+	return tryGetNoticeContainer(currentElement.parentElement);
 }
 
 function persistDismissedNotice(containerElement) {
-    const dismissIdentifier = containerElement && containerElement.getAttribute(dismissal_data.data_attribute);
-    if (dismissIdentifier) {
-        var formData = new FormData();
+	const dismissIdentifier = containerElement && containerElement.getAttribute(dismissal_data.data_attribute);
+	if (dismissIdentifier) {
+		var formData = new FormData();
 
-        formData.append('_ajax_nonce', dismissal_data.nonce);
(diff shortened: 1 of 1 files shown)


GaryJones:
Does this work if you tab to the button and activate it (Enter or Space)?

pschoffer:
Good question. Haven't thought about it. But now I tested it and it does work.

I found [this article](https://webaim.org/techniques/javascript/eventhandlers#onclick) implying it works on `buttons` (such as in this case) but wouldn't on `divs` etc..

admin-notice/js/script.js
```diff
@@ -29,10 +29,9 @@ function onDismissed(event) {
 }
 
 function registerDismissHooks() {
-    const notices = document.getElementsByClassName("vip-notice");
-    for (const notice of notices) {
-        // Hooking up on the event caused by the core implementation of is-dismissible class, which generates dismiss button.
-        notice.addEventListener("DOMNodeRemoved", onDismissed);
+    const buttons = document.getElementsByClassName("notice-dismiss");
+    for (const button of buttons) {
+        button.onclick = onDismissed;
```
nickdaugherty, line 34: Is this going to overwrite other events that are registered on these elements? Can we just use jQuery?

brandon-m-skinner: It will. Would have the use addEventListener to append click events without removing the rest at the minimum.

Should use jQuery though since that way the compatibility exactly matches WordPress so we don't end up with any     surprises like our dismiss notice not working while all other JS does or the opposite.

pschoffer: From my earlier conversation with @GaryJones use of JQuery is discourage.

Good point with overwriting non-vip-admin notices buttons. Maybe we should add back a `vip-admin-notice` on the div, then get all those diffs. And then get all the buttons in those gifts to avoid rewriting some other buttons.

nickdaugherty: 👍 I implemented this in 1e284b42e

nickdaugherty:
@GaryJones I can confirm that Spacebar works correctly to dismiss this when focused 👍

brandon-m-skinner (approved):

netsuso:
stacks-r1278

Issue #1859, opened 2020-11-16
Deprecated instagram_cache_oembed_api_response_body hook in vip-caching.php


gmazzap:
## Expected Behavior

No deprecated notices.

## Actual Behavior

Deprecated notice shown:

> Deprecated: instagram_cache_oembed_api_response_body used for __return_true is deprecated since version with no alternative available. in /var/www/wp-includes/functions.php on line 4784

## Steps to Reproduce the Problem

1. Ensures WP_DEBUG is true
2. Ensure VIP GO MU plugins are installed
3. Ensure JetPack 9.1 is installed
4. Open any website page in browser

Note: 1 to 3 are true in any current VIP GO environment right now.

## Proposed solution

Remove this line: https://github.com/Automattic/vip-go-mu-plugins/blob/ac59632bb576b7962fc27e2cfaa46342c0393d3c/vip-helpers/vip-caching.php#L774

nickdaugherty:
Thanks for the report @gmazzap - we've triaged this internally and will have a look.

GaryJones:
Fixed in #1873.

PR #1872, opened 2020-11-19
Fix Jetpack warning for instagram oembed cache

netsuso:
## Description

As described in #1859, in Jetpack 9.1 the instagram oembed cache has been deprecated. As we are adding a filter to enable this cache, it results in a warning when Jetpack 9.1 or above is in use.

This PR solves this by checking the Jetpack version before adding the filter. It had to be moved out from `vip-helpers/vip-cache.php`, as Jetpack was still not loaded, so we couldn't check the Jetpack version there. Instead, it's now directly in the mu-plugins Jetpack loader.

## Checklist

## Steps to Test

1. Check out PR.
1. Ensure that the warning no longer appears if you are using Jetpack 9.1
1. If you pin Jetpack to an older version (e.g. 9.0), the filter is correctly applied.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,9 +7,6 @@
 [submodule "advanced-post-cache"]
 	path = advanced-post-cache
 	url = https://github.com/Automattic/advanced-post-cache.git
-[submodule "jetpack"]
-	path = jetpack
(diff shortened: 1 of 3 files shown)


netsuso:
Added two branches by mistake. Closing this one

PR #1874, opened 2020-11-19
Remove legacy jetpack submodule

netsuso:
## Description

The Jetpack plugin in `mu-plugins` is now based on versioned subtrees. This branch removes the legacy submodule, as we are no longer using it (or we should be no longer using it)

Concerns:
- If any code (especially on site repos) is directly requiring any Jetpack class from `mu-plugins/jetpack`, they will fail. To solve this, we could add a `jetpack` symlink pointing to the active versioned subtree.
- Removing a submodule is a painful process, as merges will fail for everyone in their local clone, and it will require manual intervention. Adding the symlink explained above would only exacerbate the problem. If anyone has any idea on how to do this without needing manual intervention, please comment.

## Checklist

## Steps to Test

1. Check out PR.
1. Ensure everything keeps working, especially Jetpack


pschoffer (approved):
Looks good, no crashes when tested locally.

I have found one code that expects `.../jetpack/...` path, but that is an exclusion code so I don't think it has an impact on this PR (more details in slack). 

Why have you marked it as blocked @netsuso ?

WPprodigy:
Probably going to require a bit of a wide effort. Can probably re-approach this later when we maybe change where JPs are located.

PR #1879, opened 2020-11-23
Use filters for Search ratelimiting to allow per-site adjustments

brandon-m-skinner:
## Description

Not all sites need the same settings and rate limits. Some need more, some need less. Therefore, we should allow the various settings to be set via filters.

Needs validation as well as sensible upper / lower bounds for all settings to prevent issues. Also needs related unit tests.

## Checklist

## Steps to Test

1. Set rate limiting thresholds for querying and indexing low.
2. Do searches and indexing operations to trigger ratelimiting.
3. Note the behaviour.
4. Apply PR.
5. Set rate limiting thresholds for querying and indexing low.
6. Do searches and indexing operations to trigger ratelimiting.
7. The behaviour should be the exact same as before the PR.


pschoffer (changes requested):
Looks great overall, just a minor note about default values

nickdaugherty (changes requested):
Overall looks good; left a note about an upper bound on the count limits.

nickdaugherty (dismissed):
LGTM

netsuso:
stacks-r1289

PR #1880, opened 2020-11-24
VIP Cache Manager: implement "Purge page and its assets" admin bar button

rinatkhaziev:
## Description

This PR adds an Admin Bar button that allows a user to purge the current page and all its external (as opposed to inline) styles and scripts.

Screen Shot 2020-11-23 at 8 24 01 PM

Screen Shot 2020-11-23 at 8 24 09 PM

## Checklist

## Steps to Test

1. Check out PR.
1. Make sure you're logged in.
1. Go to any frontend page and verify the button is there and working
1. Verify cookies are delicious.

## Considerations

- Do we also want to display the button in the admin to allow purging the assets for Dashboard requests?
- Do we want to add a filter to allow users to adjust the needed capability? Currently, it's hardcoded for `manage_options` which may limit the usefulness.


vip-cache-manager/js/admin-bar.js
```diff
@@ -0,0 +1,69 @@
...
+	// Stores the ref to the DOM node
+	let btn;
+
+	/**
+	 * Grab all the necesary URLs (incl. scripts and CSS) for the purge.
+	 */
+	const getURLsToPurge = () => {
+		return [document.location.toString()].concat(
+			Array.from(document.querySelectorAll('script[src]')).map(({ src }) => src),
+			Array.from(document.querySelectorAll('link[rel=stylesheet]')).map(({ href }) => href)
```
mjangda, line 26: I was wondering if we should include images as well, but probably not necessary since they're not impacted by deploys, which is when this is most needed.

mjangda:
Nice work! Quick thought: we should add some stats for how often this is used and by which sites (probably easiest using `Automattic\VIP\Stats\send_pixel()`.

vip-cache-manager/js/admin-bar.js
rinatkhaziev: I'm wondering whether that `filter` is needed here, maybe just leave the validation to the backend altogether?

GaryJones: I would love to have images included. We've got Media customers who have had to fully delete an image due to legal reasons, and while they could delete the image from the media library, the image was cached at the VIP edge and it was a pain to try and clear it.

If we're not going to provide a free-form text field where URLs could be entered to purge (image) URLs one by one, then I think a "Flush page cache" button should flush _everything_ on that page, including scripts, styles, images, uploaded videos and PDFs and any other assets referenced on the page.

vip-cache-manager/js/admin-bar.js
```diff
@@ -0,0 +1,69 @@
...
+		try {
+			const res = await postData(`${ajaxurl}`, { nonce, urls });
+			const { success, data } = res;
+
+			btn.textContent = data.result || 'Success';
+			btn.disabled = true;
+			btn.removeEventListener('click', onClickHandler);
+		} catch (err) {
+			purgeQueried = false;
+			btn.textContent = '❌ Cache Purge Failed';
```
GaryJones, line 60: Is there the option to use https://make.wordpress.org/accessibility/handbook/markup/wp-a11y-speak/ here, so that screen reader users get notified on the success / failure of the action?

vip-cache-manager/vip-cache-manager.php
```diff
@@ -61,6 +64,79 @@ public function clear_queued_purge_urls() {
...
+			return;
+		}
+
+		$admin_bar->add_menu(
+			[
+				'id'     => 'vip-purge-page',
+				'parent' => null,
+				'group'  => null,
+				'title'  => 'Flush Cache for Page',
+				'href'   => '#',
```
GaryJones, line 83: Could the `meta` -> `html` value be set, so that this becomes a `<button type="button">` rather than an empty `<a>` link (well, link pointing to `#`), for accessibility purposes?

vip-cache-manager/js/admin-bar.js
mjangda: @GaryJones An image delete from the media library should trigger a purge; was that not the case?

GaryJones: I guess it may have been fixed now, but it was previously broken - deleted media library item didn't delete. I guess it could also have been down to https://github.com/Automattic/vip-go-mu-plugins/blob/5fd43e123c0cc0f2da925201fe2484bcf14b5712/vip-cache-manager/vip-cache-manager.php#L346-L349.

Let's skip over my concern if it's believed to be handled elsewhere then.

vip-cache-manager/vip-cache-manager.php
rinatkhaziev: @GaryJones unfortunately, I wasn't able to find an example where it works out of the box, setting `html` property does allow to add _extra_ HTML, but it doesn't replace the node.

GaryJones: Oh, yes it [adds extra html](https://github.com/WordPress/WordPress/blob/d7c0343d5f7e0efdad5b7dce64981e74f5d0b7a0/wp-includes/class-wp-admin-bar.php#L575-L577). TIL. Thanks for looking!

rinatkhaziev:
@mjangda I've added the send_pixel in 2fa146f347040875773a1c011a14d373370397d7, do you think that's sufficient or overengineered?

vip-cache-manager/vip-cache-manager.php
```diff
@@ -61,6 +64,96 @@ public function clear_queued_purge_urls() {
...
+	 *
+	 * @return void
+	 */
+	public function ajax_vip_purge_page_cache() {
+		$req = json_decode( file_get_contents( 'php://input' ) );
+
+		if ( json_last_error() ) {
+			\Automattic\VIP\Stats\send_pixel( [
+				'vip-go-cache-manager-action' => 'user-url-purge',
+				'vip-go-cache-manager-site'   => VIP_GO_APP_ID,
```
mjangda, line 120: Since the PURGE didn't complete, we should probably not bump these:

```suggestion
```

vip-cache-manager/vip-cache-manager.php
```diff
@@ -61,6 +64,96 @@ public function clear_queued_purge_urls() {
...
+				'vip-go-cache-manager-status' => 'bad-payload',
+			] );
+			wp_send_json_error( [ 'error' => 'Malformed payload' ], 400 );
+		}
+
+		if ( ! ( current_user_can( 'manage_options' ) && isset( $req->nonce ) && wp_verify_nonce( $req->nonce, 'purge-page' ) ) ) {
+			\Automattic\VIP\Stats\send_pixel( [
+				'vip-go-cache-manager-action' => 'user-url-purge',
+				'vip-go-cache-manager-site'   => VIP_GO_APP_ID,
```
mjangda, line 129: Since the PURGE didn't complete, we should probably not bump these:

```suggestion
```

vip-cache-manager/vip-cache-manager.php
```diff
@@ -61,6 +64,96 @@ public function clear_queued_purge_urls() {
...
+
+		// URLs are validated in queue_purge_url.
+		foreach ( $urls as $url_to_purge ) {
+			$this->queue_purge_url( $url_to_purge );
+		}
+
+		\Automattic\VIP\Stats\send_pixel( [
+			'vip-go-cache-manager-action' => 'user-url-purge',
+			'vip-go-cache-manager-site'   => VIP_GO_APP_ID,
+			'vip-go-cache-manager-status' => 'success',
```
mjangda, line 146: A bit more specific naming for the stats:

```suggestion
			'vip-go-cache-manager-url-purge-by-site'   => VIP_GO_APP_ID,
			'vip-go-cache-manager-url-purge-status' => 'success',
```

rinatkhaziev:
@mjangda I've recreated your suggestions via a separate commit.

vip-cache-manager/js/admin-bar.js
```diff
@@ -0,0 +1,69 @@
...
+		if (!(nonce && ajaxurl)) {
+			alert('VIP Cache Manager: page cache purging disabled');
+		}
+
+		purgeQueried = true;
+
+		const urls = getURLsToPurge();
+
+		try {
+			const res = await postData(`${ajaxurl}`, { nonce, urls });
```
mjangda, line 52: Unclear why we need a string literal there:

```suggestion
			const res = await postData(ajaxurl, { nonce, urls });
```

vip-cache-manager/vip-cache-manager.php
```diff
@@ -61,6 +64,92 @@ public function clear_queued_purge_urls() {
...
+		$req = json_decode( file_get_contents( 'php://input' ) );
+
+		if ( json_last_error() ) {
+			\Automattic\VIP\Stats\send_pixel( [
+				'vip-go-cache-manager-url-purge-status' => 'bad-payload',
+			] );
+			wp_send_json_error( [ 'error' => 'Malformed payload' ], 400 );
+		}
+
+		if ( ! ( current_user_can( 'manage_options' ) && isset( $req->nonce ) && wp_verify_nonce( $req->nonce, 'purge-page' ) ) ) {
```
mjangda, line 124: I would split these up because it's very easy to introduce a mistake with a long conditional like this. We can then also separate the deny stats into `deny-permissions` and `deny-nonce`.

vip-cache-manager/js/admin-bar.js
```diff
@@ -0,0 +1,69 @@
+(async () => {
```
mjangda, line 1: This javascript file needs more `( nacin spacin )` :)

vip-cache-manager/js/admin-bar.js
```diff
@@ -0,0 +1,69 @@
...
+			headers: {
+				'Content-Type': 'application/json'
+			},
+			body: JSON.stringify({ ...data })
+		});
+		return response.json();
+	}
+
+	// Indicates whether there's a purge request happening
+	let purgeQueried = false;
```
mjangda, line 16: Minor: This would be better named as `purgeInProgress`.

mjangda (commented):
Overall this looks pretty solid. Spotted a couple of small things.

vip-cache-manager/js/admin-bar.js
rinatkhaziev: Ha, that was supposed to be `purgeQueued` :)

vip-cache-manager/vip-cache-manager.php
rinatkhaziev: Good point, will address.

vip-cache-manager/js/admin-bar.js
rinatkhaziev: `Prettier` doesn't like  `( nacin spacin )` but let me see what I can do short of fixing each occurrence manually :)

vip-cache-manager/vip-cache-manager.php
rinatkhaziev: Ok, I've split the check into two, as proposed.

vip-cache-manager/js/admin-bar.js
rinatkhaziev: Applied `eslint-config-wpvip` to the file (will work on adding it to the repo in a separate PR)

mjangda:
Props @emrikol for the initial pass at this!

mjangda (approved):
Tested and working great!

netsuso:
stacks-r1307

PR #1881, opened 2020-11-24
Adds corrupted version detection

pschoffer:
## Description

Self-healing versioning options

## Checklist

## Steps to Test

1. Break your versioning options by running - `lando wp eval "update_option( 'vip_search_index_versions', 'malformed' );"`
1. Run `wp vip-search index-versions list post` shows correct versions (with lowest version active)
1. Add version 
1. Remove option  `lando wp eval "delete_option('vip_search_index_versions');"`
1. Run `wp vip-search index-versions list post` shows correct versions (with lowest version active)

Note:
To clean cache lock in order to test:
`lando wp eval "wp_cache_set('index_versions_self_heal_lock', 0, 'vip_search');"`

## Changelog 

###Search Corrupted Index Version Detection

Adding a feature that detects if index versions on given site are in incorrect state or missing.

Next step is to automatically recreate the versioning information by scanning existing ES indexes. 

At this point the change only detects an issue and logs the state it would recreate (without actually doing it). We will review those changes first and then enable self-healing.

https://github.com/Automattic/vip-go-mu-plugins/pull/1881


search/includes/classes/class-versioning.php
```diff
@@ -745,4 +755,128 @@ public function action__ep_delete_indexable( $object_id, $indexable_slug ) {
...
+		if ( ! is_array( $indicies ) ) {
+			return [];
+		}
+
+		$versioning = [];
+
+		foreach ( $indicies as $index ) {
+			$index_parts = explode( '-', $index );
+
+			// Proper index is `vip-<env_id>-<indexable-slug>(-<blog_id>)(-v<version>)`
```
pschoffer, line 841: How should we use blog_id if in any way? Any other concerns?

search/includes/classes/class-versioning.php
```diff
@@ -745,4 +755,128 @@ public function action__ep_delete_indexable( $object_id, $indexable_slug ) {
...
 	}
+
+	/**
+	 * Check if the versions are persisted correctly. If not recreate them.
+	 */
+	public function maybe_self_heal() {
+		if ( $this->is_versioning_valid() ) {
+			return;
+		}
+
```
pschoffer, line 766: Needs memcache lock + handle requests while healing.

Any tips on how to do those things?

nickdaugherty: For the memcache lock, the idea is to prevent multiple instances of this "repair" routine from running in parallel, which can quickly spiral out of control on high traffic sites.

The easiest way to do this is to just set a cache entry like `wp_cache_set( $key, 1, $group, $ttl )` when the routine starts, and check if it's already locked prior to starting, with `1 === wp_cache_get( $key )`. Oh and we should clean it up after the routine finishes, in case it gets corrupted again - `wp_cache_delete( $key, $group )`.

For handling requests during this...that's a tricky one. One one hand, we could check this key prior to every request and just tell it to fall back to the DB...but at the same time, adding the overhead of a memcache roundtrip to each request that hits ES isn't ideal, especially since this case will be super rare. The healing routine should also be pretty fast, so special handling for it might be an overoptimization that we can skip for now.

search/includes/classes/class-versioning.php
```diff
@@ -745,4 +755,128 @@ public function action__ep_delete_indexable( $object_id, $indexable_slug ) {
...
+		if ( ! is_array( $all_versions ) || count( $all_versions ) == 0 ) {
+			return false;
+		}
+
+		$valid_slugs = \ElasticPress\Indexables::factory()->get_all( null, true );
+
+
+		foreach ( $all_versions as $slug => $versions ) {
+			if ( ! in_array( $slug, $valid_slugs ) ) {
+				return false;
```
nickdaugherty, line 785: I think we can omit this check because this will end up triggering a rebuild when features are disabled, which is probably a worse user experience (since the active index could end up being different after the rebuild, if the previously active one wasn't the oldest) than letting the data stay there and possibly be stale if that feature was re-activated. Thoughts?

search/includes/classes/class-versioning.php
```diff
@@ -745,4 +755,128 @@ public function action__ep_delete_indexable( $object_id, $indexable_slug ) {
...
+
+	public function get_all_accesible_indicies() {
+		$found_indices = [];
+
+		$response = $this->elastic_search_instance->remote_request( '_cat/indices?format=json' );
+
+		$response_code = (int) wp_remote_retrieve_response_code( $response );
+
+		if ( is_wp_error( $response ) || $response_code >= 400 ) {
+			return $found_indices;
```
nickdaugherty, line 812: What do you think about returning the WP_Error here instead of the empty array? That gives the calling code more options on how to handle the error condition.

search/includes/classes/class-versioning.php
```diff
@@ -745,4 +755,128 @@ public function action__ep_delete_indexable( $object_id, $indexable_slug ) {
...
+		foreach ( $response_body as $index_obj ) {
+			if ( is_array( $index_obj ) && isset( $index_obj['index'] ) ) {
+				$found_indices[] = $index_obj['index'];
+			}
+		}
+
+		return $found_indices;
+	}
+
+	public function reconstruct_versioning( $indicies ) {
```
nickdaugherty, line 831:
```suggestion
	public function reconstruct_versioning_option( $indicies ) {
```

Small tweak - I'd suggest naming this so that it's more clear what exactly is being reconstructed.

search/includes/classes/class-versioning.php
nickdaugherty: The versioning option is stored separately on every subsite - so if we're on a multisite, we need to only use the indexes that match the current blog id (`get_current_blog_id()`) and ignore the indexes for all other subsites.

search/includes/classes/class-versioning.php
```diff
@@ -745,4 +755,128 @@ public function action__ep_delete_indexable( $object_id, $indexable_slug ) {
...
+			return [];
+		}
+
+		$versioning = [];
+
+		foreach ( $indicies as $index ) {
+			$index_parts = explode( '-', $index );
+
+			// Proper index is `vip-<env_id>-<indexable-slug>(-<blog_id>)(-v<version>)`
+			if ( count( $index_parts ) >= 3 ) {
```
nickdaugherty, line 842: Suggestion - it would be easier to test and maintain if the index name parsing was split out into a separate function that returned the data as an associative array.

nickdaugherty (changes requested):
This is looking good! I left some feedback inline. On the multisite stuff, we should also add tests that ensure we only reconstruct the version option for the current site, even if many indexes are returned from ES.

search/includes/classes/class-versioning.php
pschoffer: Oh alright makes sense.

What about non-blog specific indexes? I read [somewhere](https://github.com/Automattic/vip-go-mu-plugins/blob/4da0dc0963530c03d0cf331eb10d9fcfc399b8ae/search/includes/classes/class-search.php#L358) that `user` is example of such

search/includes/classes/class-versioning.php
pschoffer: Sounds good, I dont see any harm in keeping unmatched ones there.

search/includes/classes/class-versioning.php
pschoffer: Sure I can do that. I didn't want to log anything as ur request filter is already catching it and logging it. But I can return an error here and silently handle it in caller.

search/includes/classes/class-versioning.php
pschoffer: Great thanks for the directions! I put it to dedicated methods to hopefully be more readable.

search/includes/classes/class-versioning.php
pschoffer: Fixed

search/includes/classes/class-versioning.php
nickdaugherty: Good point - there are some global indexes. I believe EP has a utility for detecting if a given index is global, though that may not help us for this.

search/includes/classes/class-versioning.php
```diff
@@ -745,4 +759,176 @@ public function action__ep_delete_indexable( $object_id, $indexable_slug ) {
...
+
+	public function is_versioning_valid() {
+		$all_versions = get_option( self::INDEX_VERSIONS_OPTION, array() );
+
+		if ( ! is_array( $all_versions ) || count( $all_versions ) == 0 ) {
+			return false;
+		}
+
+		foreach ( $all_versions as $versions ) {
+			if ( ! is_array( $versions ) || count( $versions ) == 0 ) {
```
nickdaugherty, line 809: Having a count of 0 doesn't _necessarily_ mean that the option is invalid, because index version 1 will get automatically added via `get_versions()` if the array is empty. That was done for backwards compatibility of sites that had indexes set up before versioning was added.

In practice though, this might be ok 🤔 I guess my only worry is that when deployed, this will end up "fixing" a bunch of sites that don't yet have any additional index versions, which is _technically_ unnecessary. I'm not sure how that could be solved though...

search/includes/classes/class-versioning.php
```diff
@@ -745,4 +759,176 @@ public function action__ep_delete_indexable( $object_id, $indexable_slug ) {
...
+				$found_indices[] = $index_obj['index'];
+			}
+		}
+
+		return $found_indices;
+	}
+
+	public function reconstruct_versioning_option( $indicies ) {
+		if ( ! is_array( $indicies ) ) {
+			return [];
```
nickdaugherty, line 860: Will this cause an infinite loop of reconstruction attempts? Since we return an empty array, that will get saved into the option, then `is_versioning_valid()` will think it's invalid (empty) and try to re-construct it.

I think we need to set a longer lock when we fail to properly re-construct the option. So if we fail, we can set the lock for say 5 minutes so we don't keep hammering ES with `_cat/indices` requests on every request. We also need to log that condition to the PHP log and probably the log2logstash log so we can surface the problem.

search/includes/classes/class-versioning.php
```diff
@@ -745,4 +759,176 @@ public function action__ep_delete_indexable( $object_id, $indexable_slug ) {
...
+			$versioning[ $slug ][] = $index_info['version'];
+		}
+
+		foreach ( $versioning as $slug => $versions ) {
+			sort( $versions );
+			$version_objects = array_map( function( $version ) {
+				return [
+					'number' => $version,
+					'active' => false,
+				];
```
nickdaugherty, line 890: Can we reuse existing functionality here somehow? This format doesn't match the existing format (missing some fields), so it would be good to keep things DRY and consistent.

nickdaugherty (changes requested):
There's a few new things I noticed (comments inline)...this is trickier than I first imagined 😄

search/includes/classes/class-versioning.php
```diff
@@ -745,4 +759,176 @@ public function action__ep_delete_indexable( $object_id, $indexable_slug ) {
...
+	 * Check if the versions are persisted correctly. If not recreate them.
+	 */
+	public function maybe_self_heal() {
+		if ( $this->is_self_heal_ongoing() ) {
+			return;
+		}
+		$this->mark_self_heal_ongoing();
+
+		if ( $this->is_versioning_valid() ) {
+			$this->mark_self_heal_finished();
```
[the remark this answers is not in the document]
pschoffer: fixed

search/includes/classes/class-versioning.php
```diff
@@ -9,11 +9,20 @@
 
 class Versioning {
 	const INDEX_VERSIONS_OPTION = 'vip_search_index_versions';
+	const INDEX_VERSIONS_SELF_HEAL_LOCK_CACHE_KEY = 'index_versions_self_heal_lock';
+	const INDEX_VERSIONS_SELF_HEAL_LOCK_CACHE_GROUP = 'vip_search';
+	const INDEX_VERSIONS_SELF_HEAL_LOCK_CACHE_TTL = 300; // 5 minutes in seconds.
```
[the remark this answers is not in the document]
pschoffer: fixed

pschoffer:
About the way we are persisting it. Some parts of code are using `site_option` where others (most) are just using `option`. Form what I [read](https://developer.wordpress.org/reference/functions/get_site_option/) it is the same thing in singlesite instalation, while in multisite using site_option would result in having one global option where the using only `option` would be one per site.

Seems one-per site is what we want or? (with setting/switching current blog_id). Or maybe just having one global but then having more dimensions (per site).

Any input on that @nickdaugherty ?

nickdaugherty:
> Seems one-per site is what we want or? (with setting/switching current blog_id).

Yeah that's what we want, at least for the non-global indexes. That's why I think there's a bug, b/c we're using the `*_site_option()` when we want `*_option()`. It's tricky though, because some indexes are global like `users`...each individual site can't have it's own version of that index b/c it's shared across all of them, so that index's versions need to be stored in `*_site_option()`. I don't think I thought about that when first writing this...

search/includes/classes/class-versioning.php
nickdaugherty: @pschoffer I think this is still a problem - an empty array will cause line 813 to return false, marking it as invalid, so the process will just keep going over and over on every request. In theory this only happens if ES is down or we can't load the indexes for some other reason, but we should be defensive here and have a "circuit breaker" that prevents this from trying to rebuild over and over.

search/includes/classes/class-versioning.php
pschoffer: That I think might be ok. First time this code will run it will reconstruct the defaults and on every other hit it will just use those and not reconstruct anything right?

search/includes/classes/class-versioning.php
pschoffer: I added the longer (10 m) backoff and slack alert, but not sure how to handle that differently.

search/includes/classes/class-versioning.php
pschoffer: There is that normalize_version method that runs when getting versions. But as it runs on getting versions I assumed it is fine to keep. I can add it here as well though. Do you think that would be good?

search/includes/classes/class-versioning.php
```diff
@@ -40,17 +61,23 @@ public function __construct() {
 		// the Queue to take over EP's queue, at which point we don't need to insert them here, as they are handled during Queue::queue_object())
 		add_filter( 'pre_ep_index_sync_queue', [ $this, 'filter__pre_ep_index_sync_queue' ], 100, 3 );
 
-		add_action( 'plugins_loaded', [ $this, 'action__plugins_loaded' ] );
+		add_action( 'init', [ $this, 'action__elasticpress_loaded' ], PHP_INT_MAX );
```
pschoffer, line 64: Check out this change. Seems the deletion only ever worked for post type as the features are enabled on [init](https://github.com/10up/ElasticPress/blob/a2545cbb96740694ff9c5067a5f634fafef16932/includes/classes/Features.php#L36)

Which happens after plugins_loaded.

nickdaugherty: Oh good catch - what do you think about splitting this out into a separate PR, since it's mostly unrelated? We can get this merged/deployed independently.

search/includes/classes/class-versioning.php
nickdaugherty: 👍 that sounds good and should work for this to keep it consistent.

search/includes/classes/class-versioning.php
pschoffer: I created this PR: https://github.com/Automattic/vip-go-mu-plugins/pull/1903

However I need that change here as well, otherwise you can't test partial self-healing (e.g. only of post, but leaving users intact).

We can then squash this and rebase when the 1903 gets merged.

search/includes/classes/class-versioning.php
pschoffer: added

nickdaugherty:
@pschoffer I just had a thought on this - since it's a reasonably risky update (all sites without version option will get version options written), what if we deploy it in a "dry run" mode first, to see how it behaves? Specifically, we run everything normally except for the `update_versions()` call that persists it - so let the detection + locking all run, but instead of updating the option it just uses log2logstash to record what it will do (recording the new option value in the `extras` of the log entry).

That way we can monitor to see exactly what changes and on what sites it will do, before we make it live. I guess one thing we should consider is we'll get a log message every 10 seconds...we could bump up the lock to say 10 minutes for the time being, to reduce the amount of data we're collecting.

nickdaugherty (dismissed):

nickdaugherty:
Whoops we have a conflict now. @pschoffer while in our dry-run mode, i think we should bump the time between attempts so say 1 hour (or even more really), b/c every site that only has the default versions will end up sending an alert (currently every 10m)

pschoffer:
@nickdaugherty  I resolved the conflicts and pushed the backup interval to 2 hours.

search/includes/classes/class-versioning.php
nickdaugherty: We could return `false` or a `WP_Error` here and check for that value in the calling code - if it gets an error, it sets a longer lock to prevent this from running over and over (and failing). Not necessary for this PR but IMO we should do it before we take this out of dry run mode.

nickdaugherty (approved):

pschoffer:
Stcks-r1312

PR #1888, opened 2020-11-27
Cleans_stale_notice

pschoffer:
## Description

Adds removing stale dismissed identifiers in every ~ 1 in 100 admin requests.

Also sneaked in the fix for adding multiple dismissibility keys.

## Checklist

## Steps to Test

1. Define few (at least 3) notices that are dismissible
2. Dismiss two of them
3 On refresh only see one remaining


admin-notice/class-admin-notice-controller.php
```diff
@@ -41,8 +43,35 @@ public function dismiss_vip_notice() {
 		check_ajax_referer( self::DISMISS_NONCE_ACTION );
 
 		if ( isset( $_POST[ self::DISMISS_IDENTIFIER_KEY ] ) ) {
-			update_user_meta( get_current_user_id(), self::DISMISS_USER_META, sanitize_text_field( $_POST[ self::DISMISS_IDENTIFIER_KEY ] ) );
+			add_user_meta( get_current_user_id(), self::DISMISS_USER_META, sanitize_text_field( $_POST[ self::DISMISS_IDENTIFIER_KEY ] ) );
```
pschoffer, line 46: Fix dismissing multiple notices bug.

netsuso (approved):
Everything looks good! I've tested both the fix for multiple dismisses and the cleanup of stale dismissed meta, and it worked as expected :)

netsuso:
stacks-r1287

admin-notice/class-admin-notice-controller.php
GaryJones: Won't this only persist the first dismissed notice, and not any more?

PR #1895, opened 2020-12-03
Add PHP 8 tests

mjangda:
With PHP 8 out, we should start running lint and tests against it so we don't get caught off-guard in the future.

Although, looks like there might be a mess of composer dependency issues 🙀


brandon-m-skinner:
Slightly better with a `composer update`. 

```
  Problem 1
    - phar-io/manifest is locked to version 1.0.3 and an update of this package was not requested.
    - phar-io/manifest 1.0.3 requires php ^5.6 || ^7.0 -> your php version (8.0.0) does not satisfy that requirement.
  Problem 2
    - phar-io/version is locked to version 2.0.1 and an update of this package was not requested.
    - phar-io/version 2.0.1 requires php ^5.6 || ^7.0 -> your php version (8.0.0) does not satisfy that requirement.
  Problem 3
    - phpunit/php-code-coverage is locked to version 6.1.4 and an update of this package was not requested.
    - phpunit/php-code-coverage 6.1.4 requires php ^7.1 -> your php version (8.0.0) does not satisfy that requirement.
  Problem 4
    - phpunit/phpunit is locked to version 7.5.20 and an update of this package was not requested.
    - phpunit/phpunit 7.5.20 requires php ^7.1 -> your php version (8.0.0) does not satisfy that requirement.
  Problem 5
    - sebastian/global-state is locked to version 2.0.0 and an update of this package was not requested.
    - sebastian/global-state 2.0.0 requires php ^7.0 -> your php version (8.0.0) does not satisfy that requirement.
```

brandon-m-skinner:
The issue seems to be related to dependencies of our dependencies, namely phpunit using an old phar-io/manifest version that doesn't allow anything above PHP 7.

However, WordPress tests aren't compatible with phpunit versions above 7.x.

brandon-m-skinner:
https://core.trac.wordpress.org/ticket/50902

brandon-m-skinner:
With the latest changes, all the jobs run. Just need to sort out the phpcs issues, deprecations, and errors in tests now.

pschoffer:
Oh nice hack with the PHPUnit fork:)

GaryJones:
Many of the changes for PHP 8 are related to types. To properly check them, then tests need to be checking both the _happy path_ and the _unhappy path_ (hopefully most of that is happening already).

Where it breaks down though, is knowing what level of coverage there is for our code - and to really get an accurate sense of what is truly testing all the happy and unhappy paths for a bit of code, and what is a function that was called incidentally as part of testing another function (and therefore not testing the unhappy path), then `@covers` annotations need to be added. See https://core.trac.wordpress.org/ticket/39265 for the effort this is taking for WP Core, and an explanation from @jrfnl of why adding these are valuable. Right now, only two test files includes them.

rinatkhaziev:
I'm going to close this so it doesn't sit collecting dust, once we start the PHP8 compat work we can resurrect or start fresh.

rinatkhaziev:
@GaryJones Feel free to reach out in Slack if you wish to discuss this further.

PR #1897, opened 2020-12-03
Ensure that a proper Jetpack version is loaded depending on WordPress version

netsuso:
## Description

As there are still VIP Go sites running on WordPress <5.5, they are failing to load Jetpack 9.2, as it's only compatible with newer WordPress versions.

With this change, we ensure that sites running older WP versions are using Jetpack 9.1 instead of 9.2.

Once they are upgraded to a newer WordPress version, they will immediately start using Jetpack 9.2.

## Checklist

## Steps to Test

1. Check out PR.
1. Set up a sandbox/dev env with WordPress 5.4.2
1. Ensure that Jetpack works and version 9.1 is loaded
1. Go back to WordPress 5.5
1. Ensure that Jetpack 9.2 is loaded

### DIFF
--- a/jetpack.php
+++ b/jetpack.php
@@ -12,6 +12,8 @@
  * Domain Path: /languages/
  */
 
+// Choose an appropriate default Jetpack version, ensuring that older WordPress versions
+// are not using a too modern Jetpack version that is not compatible with it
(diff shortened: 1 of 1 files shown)


brandon-m-skinner (approved):

netsuso:
stacks-r1300

PR #1899, opened 2020-12-04
Improve logging for failed HTTP requests to ElasticSearch

netsuso:
## Description

When there's a problem with the HTTP connection to ElasticSearch, we are masking it with a fallback error that prevents the actual error (e.g, a curl error) to be logged.

This branch removes that fallback error, and also creates a new logstash feature (`vip_search_http_error`) to log all those connection errors.

## Checklist

## Steps to Test

1. Check out PR.
1. Just before the call to `vip_safe_wp_remote_request` in `class-search.php`, add a line to create a bad ES query. E.g: `$args['body'] = '{';`
1. Run a search on the testing site. The usual logstash error should appear in the log (including `error_type` and `root_cause` in the extra field)
1. Replace that added line for another one to provoke a HTTP error. E.g: `$query['url'] = 'http://10.0.0.1/';`
1. Run a search again. The new logstash error, including the cURL timeout error, should appear in the log

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -592,7 +592,12 @@ class Search {
 			}
 		}
 
-		return $response;
+		if ( is_wp_error( $response ) ) {
+			// Return a generic VIP Search WP_Error instead of the one from wp_remote_request
+			return new \WP_Error( 'vip-search-upstream-request-failed', 'There was an error connecting to the upstream search server' );
+		} else {
+			return $response;
+		}
 	}
 
 	public function ep_handle_failed_request( $response, $statsd_prefix ) {
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -531,8 +531,6 @@ class Search {
 	}
 
 	public function filter__ep_do_intercept_request( $request, $query, $args, $failures ) {
-		$fallback_error = new \WP_Error( 'vip-search-upstream-request-failed', 'There was an error connecting to the upstream search server' );
-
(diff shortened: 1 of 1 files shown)


search/includes/classes/class-search.php
```diff
@@ -531,8 +531,6 @@ public function filter__ep_bulk_items_per_page() {
 	}
 
 	public function filter__ep_do_intercept_request( $request, $query, $args, $failures ) {
-		$fallback_error = new \WP_Error( 'vip-search-upstream-request-failed', 'There was an error connecting to the upstream search server' );
```
netsuso, line 534: @nickdaugherty I'm not sure why this is here, and whether removing it could lead to any unexpected problem

nickdaugherty: @netsuso this is the value that is returned to the calling code from `vip_safe_wp_remote_request()` if the request is not successful. We need to keep it so that we correctly bubble up a WP_Error() to calling code, to match the behavior of the WP_HTTP API, as lots of code checks for errors with `is_wp_error()`.

nickdaugherty: Discussed more on Slack - the intent of the question was on if we can use the default WP_Errors that get returned, so we have more granularity in the logging, which would be good. Then we can turn those into a more user-friendly error that gets returned for surfacing higher up.

netsuso: I've updated the PR according to that discussion

nickdaugherty (approved):
LGTM

netsuso:
stacks-r1302

PR #1902, opened 2020-12-10
Initial implementation of Files ACL Endpoint

mjangda:
This will allow us to restrict access to files based on different signals from the request (e.g. inherited post status, user permissions, etc.)

See this comment for more details about this functionality: https://github.com/Automattic/vip-go-mu-plugins/pull/1902#issuecomment-755390156

## TODO

- [ ] Finalize endpoint path / name
- [x] Caching
- [x] More tests
- [x] Swap X-Reason with trigger_error

## Checklist

## Steps to Test

Since the nginx pieces aren't in place, we can test by calling the PHP script/endpoint directly:

```
curl 'http://vip-go-dev.lndo.site/wp-content/mu-plugins/endpoints/internal-check-file-acl.php' -H 'X-Original-Uri: /testin/wp-content/uploads/path/to/file.png'
```

Here's a script to quickly scaffold some test data using `vip-go-mu-dev` (`vip-dev-env` is missing GD and doesn't work :():

```
WP="lando wp"

IMAGE_ID=`$WP media import --title="Attachment" --porcelain "http://s0.wp.com/i/void.gif"`
IMAGE_URL=`$WP post get $IMAGE_ID --field=guid`

echo "== Attachment URL"
echo $IMAGE_URL
```

You can then pass that into the `X-Original-Uri` header. All requests should return a 202.

#### Invalid requests

If you make a request with invalid data (e.g. missing header, missing wp-content path in header, etc.) it should fail with a 500

If the filter returns an invalid value, it will also fail with a 500.

#### Restricted Files

If you add the following filter it should return a 403 + `X-Private: true`:

```
add_filter( 'vip_files_acl_file_visibility', function( $file_visibility, $sanitized_file_path ) {
        return Automattic\VIP\Files\Acl\FILE_IS_PRIVATE_AND_DENIED;
}, 10, 2 );
```

#### Restricted (But Allowed) Files

If you add the following filter it should return a 202 + `X-Private: true`:

```
add_filter( 'vip_files_acl_file_visibility', function( $file_visibility, $sanitized_file_path ) {
        return Automattic\VIP\Files\Acl\FILE_IS_PRIVATE_AND_ALLOWED;
}, 10, 2 );
```

#### Not Found

```
add_filter( 'vip_files_acl_file_visibility', function( $file_visibility, $sanitized_file_path ) {
        return Automattic\VIP\Files\Acl\FILE_NOT_FOUND;
}, 10, 2 );
```

Should return a 202 as well but the request itself will fail at the nginx level.

#### Subsites

If you create a multisite subsite:

```
lando wp core multisite-convert

lando wp site create --slug=testing
```

You can repeat the earlier scaffolding script with `WP="lando wp --url=--url=http://vip-go-dev.lndo.site/testing/"`

Then, add the `/testing` prefix to `X-Original-Uri`. It should load only the subsite context. You can verify this by adding a subsite check to the filter or `var_dump( get_blog_details() );`


endpoints/internal-check-file-acl.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+namespace Automattic\VIP\Files\Acl;
+
+require_once __DIR__ . '/../files/acl.php';
+
+$file_request_uri = $_SERVER['HTTP_X_ORIGINAL_URI'] ?? null;
+
+$validated_file_path = pre_wp_validate_request( $file_request_uri );
```
rinatkhaziev, line 9: This part is kind of unobvious:
1. We expect the file path to be a string.
2. But then the validation can turn it into an array if the file is invalid.
3. But if it's valid we also normalize and sanitize the string in the validation function.

This introduces a bunch of ambiguity, which can make it hard to follow the logic.

I propose splitting `pre_wp_validate_request` in two functions:
1.  `pre_wp_validate_request` would always return a tuple with code and message, 200 means valid)
2. `sanitize_file_path` (or `normalize_file_path`) would then do `Strip off subdirectory` part.

This way we can get a consistent return signature and make it easier to follow.

mjangda: I originally had it returning a `WP_Error` object but then realized that WP is not loaded 🤦 

I was a bit unsure about doing both validation and sanitization in the same function but ended up going that way for simplicity by minimizing the number of functions + branches (they're not written for re-use, just for encapsulation and testability) and to avoid duplicating work (we'd have to call `parse_url` twice, once in the validation function and once in the sanitization one).

I could rename the function (`pre_wp_validate_and_sanitize_original_uri`?) to add more clarity if that would help?

files/acl.php
```diff
@@ -0,0 +1,135 @@
...
+ *
+ * @return (int) The found attachment ID, or 0 on failure.
+ */
+function file_path_to_attachment_id( $path ) {
+	global $wpdb;
+
+	$attachment_id = 0;
+
+	// TODO: caching?
+	$sql = $wpdb->prepare(
```
mjangda, line 113: Caching is a big blocker here :)

As is a meta_key index.

mjangda:
Note that the endpoint doesn't currently do anything beyond some light validation, loading WordPress, triggering the `vip_files_acl_file_visibility` filter and returning with the correct status code.

The idea here is that we can have built-in ways to enable different functionality that sites need:

- Restricting attachments for unpublished content (would run a reverse lookup for any media path to attachment, check the status of the parent post, and allow access if `current_user_can`): https://gist.github.com/mjangda/8657cfde2b2a4cafd95f509dd1d3dde9#file-vip-private-files-unpublished-media-php
- Fully private site (would hook in and verify that the request has a logged in user and `current_user_can`): https://gist.github.com/mjangda/8657cfde2b2a4cafd95f509dd1d3dde9#file-vip-private-files-all-media-php

And users can extend the filter in whatever way they chose.

endpoints/internal-check-file-acl.php
```diff
@@ -0,0 +1,34 @@
...
+	pre_wp_send_error_headers( $validated_file_path );
+
+	exit;
+}
+
+// TODO: make sure HTTP HOST and path have subdir set for multisite
+// TODO: handle resized files (e.g. file-200x200.jpg)
+
+// Bootstap WordPress
+require __DIR__ . '/../../../wp-load.php';
```
GaryJones, line 20: I think you can define `SHORTINIT` as `true` before here, and still have the `apply_filters()` work, though that would then limit what functions are available to be used in the filter callbacks.

files/acl.php
```diff
@@ -0,0 +1,135 @@
...
+ * Called before WordPress is loaded.
+ *
+ * @param array $error Array containing [ (int) status code, (string) error message ]
+ */
+function pre_wp_send_error_headers( $error ) {
+	list( $error_code, $error_message ) = $error;
+
+	http_response_code( $error_code );
+
+	header( 'X-Reason: ' . $error_message );
```
GaryJones, line 58: The `X-` prefix is [deprecated](https://tools.ietf.org/html/rfc6648), and I think we could give a bit more context to the header name by calling it `Acl-Reason` or similar.

mjangda: Interesting; TIL. The intent of that seems more geared toward protocols that may one day become public as opposed to private ones (https://specs.openstack.org/openstack/api-wg/guidelines/headers.html).

In any case, going with a more specific prefix (like `VIP-`) is probably better. Also, this header will be going away :)

endpoints/internal-check-file-acl.php
mjangda: I ended up separating them and moving both to a separate namespace so their use and limitations are more obvious.

files/acl.php
mjangda: X-Reason was removed.

mjangda:
@Automattic/vip-platform-cantina (specifically calling out @nickdaugherty @rinatkhaziev for your WP knowledge) and @alexcriss this is ready for a look.

Not ready to deploy since we'll want the nginx pieces in place first.

## Notes

### Flow

Due to performance costs of running the checks for each file, this feature will be selectively enabled per environment.

When the feature is enabled:

- nginx gets a request for a media item.
- The request is passed through to the PHP script included in this PR. Specifically, the script is called with all auth cookies and the original URI for the media item in the `X-Original-Uri` header.
- The script does some light validation and sanitization of the `X-Original-Uri` value. If the value begins with a subdirectory path, we set that in the script context so that WordPress loads up the correct subsite.
- We load WordPress via `wp-load.php`
- We call a filter with the sanitized path and the value returned by the filter determines what our response is (see below).
- Nginx uses the response to handle the request (whether to allow access or restrict it).

Currently the filter doesn't do anything except return that the file is public. We'll have separate PRs to support different behaviours (see https://gist.github.com/mjangda/8657cfde2b2a4cafd95f509dd1d3dde9):

1. Preventing access to media attached to unpublished posts.
2. Preventing access to all media on a site (or subsite).

This is intentional since you only want one behaviour at any given time. We'll need to think about how we allow customers to enable and disable these on a per-env / per-subsite basis.

The filter is marked as private because we want to control and limit it's usage for now. We can consider opening it up in the future. 

### Filter Return Values

The filter must return one of the following otherwise we error out (500).

The names should make their behaviours obvious :)

- `FILE_IS_PUBLIC`
- `FILE_IS_PRIVATE_AND_ALLOWED`
- `FILE_IS_PRIVATE_AND_DENIED`
- `FILE_NOT_FOUND`

### Endpoint

- Instead of integrating into the REST API or directly as a rewrite rule via WP, we're using a separate PHP script which will be called directly was chosen so we can properly load the subsite context, do some early validation, and we get a slight performance boost.
- We also load all of WordPress to ensure customizations from themes/plugins are respected.
- The endpoint will not be publicly accessible, and only callable internally via nginx.
- "Success" responses are intentionally not returning a 200. This is to avoid problems with a fatal returning a 200 (thanks PHP!) and accidentally exposing a private file.

### Tests

- There are a bunch of unit tests included.
- We'll need some automated e2e tests or similar to verify the endpoint itself works and also in production once the nginx pieces are wired up.

Will add some testing notes to the PR description.

### Looking for Feedback

Would love thoughts on:

- The name and location of the PHP script / endpoint (I put it in `endpoints` to make its purpose obvious) but I'm not sure I love that.
- I opted for namespaces + functions instead of classes since classes felt unnecessary but happy to switch that around for consistency / cleanliness.
- Anything else I haven't thought of.

files/acl.php
```diff
@@ -0,0 +1,91 @@
...
+}
+
+/**
+ * Variant of core's attachment_url_to_postid() function
+ *
+ * @param $path (string) The path to resolve.
+ *
+ * @return (int) The found attachment ID, or 0 on failure.
+ */
+function get_file_path_from_attachment_id( $path ) {
```
mjangda, line 56: It might make sense to move this out and include it with the "restrict media on unpublished content" PR since it's only going to be used there.

files/acl.php
mjangda: Caching was added.

files/acl/pre-wp-utils.php
```diff
@@ -0,0 +1,61 @@
...
+}
+
+/**
+ * Sanitize the path by stripping off the wp-content/uploads bits.
+ *
+ * @param array $file_paths Indexed array with two entries: 0 is the path before `/wp-content/uploads/` and 1 is the path + file after.
+ */
+function sanitize_and_split_path( $file_path ) {
+	// Strip off subdirectory (i.e. /en/wp-content/ <= remove `/en` )
+	$wp_content_index = strpos( $file_path, '/wp-content/uploads/' );
```
nickdaugherty, line 52: Just a thought - this parsing could also be done with `explode( '/wp-content/uploads/', $file_path, 2 );` - then there's no need for `substr` parsing (where it's easy to have off-by-one errors).

endpoints/internal-check-file-acl.php
```diff
@@ -0,0 +1,56 @@
...
+	http_response_code( 500 );
+
+	exit;
+}
+
+$file_path = parse_url( $file_request_uri, PHP_URL_PATH );
+
+$is_valid_path = Pre_WP_Utils\validate_path( $file_path );
+if ( ! $is_valid_path ) {
+	http_response_code( 500 );
```
nickdaugherty, line 21: Should this be an HTTP `400` instead? Here, this state represents a bad request from the client. Above, where we check if there's a URI at all makes sense to be a 500, b/c that's an error on our end.

endpoints/internal-check-file-acl.php
```diff
@@ -0,0 +1,56 @@
...
+
+	exit;
+}
+
+list( $subdirectory, $sanitized_file_path ) = Pre_WP_Utils\sanitize_and_split_path( $file_path );
+
+if ( $subdirectory ) {
+	$_SERVER['REQUEST_URI'] = $subdirectory . ( $_SERVER['REQUEST_URI'] ?? '' );
```
nickdaugherty, line 29: I'm not 100% sure this is safe. What if `$subdirectory` was a full url, like `//foo.com/path` (which will pass the "starts with `/` check"), or included path traversal characters (`../`)? A hypothetical request would be: `http://bar.com/%2F%2Ffoo.com/path/wp-content/uploads/baz` (though the web server may correctly handle that). Code elsewhere may not correctly (or safely) handle weird `REQUEST_URI` values. We should do more sanitization inside `sanitize_and_split_path()`.

endpoints/internal-check-file-acl.php
```diff
@@ -0,0 +1,56 @@
...
+}
+
+// Unset vars we no longer need, so they don't leak into global scope.
+// Should probably move the logic above into a function :)
```
nickdaugherty, line 33: +1 on encapsulating that into a function. Not only removes the potential for global pollution, but makes the setup testable.

files/acl.php
```diff
@@ -0,0 +1,91 @@
...
+
+		case FILE_IS_PRIVATE_AND_DENIED:
+			$status_code = 403;
+			$header = 'X-Private: true';
+			break;
+
+		case FILE_NOT_FOUND:
+			// This is a signal for the server to pass the request through to the Files Service.
+			// Even though the attachment was not found within WordPress, it may still be a physical file in the Files Service.
+			$status_code = 202;
```
nickdaugherty, line 33: This behavior seems counterintuitive and may turn into a "gotcha" in the future. If we're giving control to the application to say a file is not found...then shouldn't we honor that and not attempt to serve anything? Otherwise, we do we let the application set this code?

My worry is that folks will try to "hide" the existence of certain files by assuming they can return the FILE_NOT_FOUND code...but that will actually leave the file open to being served by the file service, which will not be expected behavior.

files/acl.php
nickdaugherty, line 56: Is it just me or is the name of this function confusing? `get_file_path_from_attachment_id()` indicates that a file path will be returned when passed in an attachment id, but the behavior is actually the opposite.

files/acl.php
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * @param $path (string) The path to resolve.
+ *
+ * @return (int) The found attachment ID, or 0 on failure.
+ */
+function get_file_path_from_attachment_id( $path ) {
+	global $wpdb;
+
+	$cache_key = 'path_' . md5( $path );
+	$attachment_id = wp_cache_get( $cache_key, 'files-acl' );
```
nickdaugherty, line 60: Can we move the cache group to a constant, and add a `vip-` prefix to it?

files/acl.php
```diff
@@ -0,0 +1,91 @@
...
+
+	http_response_code( $status_code );
+
+	if ( $header ) {
+		header( $header );
+	}
+}
+
+/**
+ * Variant of core's attachment_url_to_postid() function
```
nickdaugherty, line 50: A comment clarifying _why_ a variant is needed would be helpful for understanding.

files/acl.php
```diff
@@ -0,0 +1,91 @@
...
+	$results = $wpdb->get_results(
+		$wpdb->prepare(
+			"SELECT post_id, meta_value FROM $wpdb->postmeta WHERE meta_key = '_wp_attached_file' AND meta_value = %s",
+			$path
+		)
+	);
+
+	if ( $results ) {
+		// Use the first available result, but prefer a case-sensitive match, if exists.
+		$attachment_id = reset( $results )->post_id;
```
nickdaugherty, line 76: Is the reset technically needed, or can we use `$results[0]`?

tests/files/acl/test-pre-wp-utils.php
```diff
@@ -0,0 +1,124 @@
...
+				'/wp-content/uploads/kittens.gif',
+			],
+	
+			'valid-path-nested' => [
+				'/wp-content/uploads/subfolder/2099/12/cats.jpg',
+			],
+	
+			'valid-path-subsite' => [
+				'/subsite/wp-content/uploads/puppies.png',
+			],
```
nickdaugherty, line 50: Since this implementation supports it, can we add a test for a subsite that is several directory levels deep? `/foo/bar/baz/subsite/wp-content/uploads/puppies.png`

tests/files/acl/test-pre-wp-utils.php
```diff
@@ -0,0 +1,124 @@
...
+
+			'valid-path-subsite' => [
+				'/subsite/wp-content/uploads/puppies.png',
+				[
+					'/subsite',
+					'puppies.png',
+				],
+			],
+
```
nickdaugherty, line 106: Same note about a test case with multiple levels of subdirectories.

tests/files/test-acl.php
```diff
@@ -0,0 +1,152 @@
...
+		} else {
+			$this->assertNotContains( 'X-Private: true', xdebug_get_headers(), 'Sent headers include the X-Private header' );
+		}
+	}
+
+	/**
+	 * @runInSeparateProcess
+	 * @preserveGlobalState disabled
+	 */
+	public function test__send_visibility_headers__invaid_visibility() {
```
nickdaugherty, line 71:
```suggestion
	public function test__send_visibility_headers__invalid_visibility() {
```

nickdaugherty (changes requested):
This is really cool, so happy to see this coming together. I've left some inline thoughts...

files/acl.php
mjangda: Ha, the name is definitely backwards :)

tests/files/acl/test-pre-wp-utils.php
nickdaugherty: Another test I just thought of - a URI that includes `wp-content/uploads` twice - `/wp-content/uploads/path/to/wp-content/uploads/otters.png`

mjangda: Added in 0590ad5e4

tests/files/acl/test-pre-wp-utils.php
mjangda: Added in f36a26288

tests/files/test-acl.php
mjangda: Done in 75bd8c0fd

files/acl.php
mjangda: That's a good point.

The particular thinking here was for unpublished media (https://gist.github.com/mjangda/8657cfde2b2a4cafd95f509dd1d3dde9#file-vip-private-files-unpublished-media-php). Specifically, what to do when the file doesn't exist as an attachment in WP. But I think it probably makes sense just to drop this and return this status and return as `FILE_IS_PUBLIC`. We can add the logic later if needed.

mjangda: Removed in 5f2972db4

files/acl.php
mjangda: Moved to https://gist.github.com/mjangda/8657cfde2b2a4cafd95f509dd1d3dde9#file-vip-private-files-unpublished-media-php-L42

files/acl.php
mjangda: Moved out of PR but added here: https://gist.github.com/mjangda/8657cfde2b2a4cafd95f509dd1d3dde9#file-vip-private-files-unpublished-media-php-L45-L48

files/acl.php
mjangda: Fixed here: https://gist.github.com/mjangda/8657cfde2b2a4cafd95f509dd1d3dde9#file-vip-private-files-unpublished-media-php-L54

files/acl.php
mjangda: Moved out of PR, but implemented here: https://gist.github.com/mjangda/8657cfde2b2a4cafd95f509dd1d3dde9#file-vip-private-files-unpublished-files-php-L10

files/acl.php
mjangda: It is not; just a copy-pasta from the core version. Updated.

endpoints/internal-check-file-acl.php
mjangda: Done in e5c8b01.

It felt awkward since it's now a function that does multiple things but I agree that this is a bit cleaner and safer.

endpoints/internal-check-file-acl.php
mjangda: Suggestions on how to sanitize this? We are limited to pure PHP and it should ideally be fast.

I worry that this might become whack-a-mole like since it's pretty open in terms what is allowed in a URL.

endpoints/internal-check-file-acl.php
mjangda: Agreed that 400 is more "correct". I've asked @alexcriss to confirm if that's possible as we're pretty limited in what response codes we can use.

files/acl/pre-wp-utils.php
mjangda: Good call. Much cleaner.

mjangda:
@nickdaugherty All feedback should now be addressed except the response codes and sanitization.

I'll work on the supporting PRs, in the meantime.

mjangda:
Supporting PR to enable the restriction functionality here: https://github.com/Automattic/vip-go-mu-plugins/pull/1948

mjangda:
This is safe to deploy without the nginx changes as we have a restriction in place now to only allow access to the endpoint locally and on sandboxes.

nickdaugherty (approved):
Looking good!

rinatkhaziev:
r1349-stacks

mjangda:
We need to do `git mv endpoints/internal-check-file-acl.php files/endpoints/internal-check-file-acl.php` in a separate PR, per the Slack discussion.

netsuso:

PR #1906, opened 2020-12-11
phpunit-docker: Suffix db and network with unique names

mjangda:
To avoid conflict with other containers, suffix the name of the `tests` network and the `db` database with a timestamp. Also remove the tests config and recreate it since our container names are now different with each run.

## Checklist

n/a

## Steps to Test

1. `./bin/phpunit-docker.sh` should now work properly when you're running `vip-dev-env` at the same time.


bin/download-wp-tests.sh
```diff
@@ -98,6 +98,9 @@ install_test_suite() {
 
 	cd $WP_TESTS_DIR_ACTUAL
 
+	# Always start from a clean config
+	rm -f wp-tests-config.php
+
 	if [ ! -f wp-tests-config.php ]; then
 		download https://develop.svn.wordpress.org/${WP_TESTS_TAG}/wp-tests-config-sample.php "$WP_TESTS_DIR_ACTUAL"/wp-tests-config.php
 		sed $ioption "s:dirname( __FILE__ ) . '/src/':'$WP_CORE_DIR':" "$WP_TESTS_DIR_ACTUAL"/wp-tests-config.php
```
pschoffer, line 106: What do you think about dropping this `if`, when we always delete the file.

(I tried to do github code suggestion, but it seems you can only do that certain amount of lines away from the change :( )

pschoffer (changes requested):
I tested the change and it works great. It worked for me without the change as well though, the db container in my case when using vip-dev-env is `vipdevlocal_database_1` for a `local` env.

I added a small note, but otherwise LGTM.

Cool, that you pinned the mariadb version!

bin/download-wp-tests.sh
mjangda: Yeah, that's a good idea 👍

mjangda:
> I tested the change and it works great. It worked for me without the change as well though, the db container in my case when using vip-dev-env is vipdevlocal_database_1 for a local env.

Thanks for testing. Looking at it again, I think the root problem was that I had an old, stale test config file that was breaking things so the `rm` is the most important thing in here. The container rename is just a bonus :)

bin/download-wp-tests.sh
mjangda: Done.

pschoffer (approved):
LGTM

pschoffer:
Code is not synced to SVN => no Deploy needed.

PR #1909, opened 2020-12-12
Added additional failing tests from WP 5.6 to exclude list.

brandon-m-skinner:
## Description

Some additional core tests started failing after the upgrade to WordPress 5.6. They should be excluded since they're known to fail 100% of the time due to mu-plugins changing assertions in ways the tests don't expect.


bin/utils.sh
```diff
@@ -31,51 +31,55 @@ exclude_core_tests() {
...
-		"tests/admin/includesScreen.php"
-		"tests/filesystem/base.php"
+		"tests/date/getFeedBuildDate.php"
+		"tests/dependencies/scripts.php"
 		"tests/feed/atom.php"
-		"tests/adminbar.php"
-		"tests/xmlrpc/wp/getUser.php"
-		"tests/rest-api/rest-users-controller.php"
+		"tests/feed/rss2.php"
+		"tests/filesystem/base.php"
```
brandon-m-skinner, line 45:
```suggestion
```

PR #1910, opened 2020-12-12
Remove exclusively core test jobs from CircleCi workflow.

brandon-m-skinner:
## Description

Running just core tests without mu-plugins attached is no longer necessary.

They're not actionable and don't provide any indication of the state of this repository.

We will keep the core tests running with our mu-plugins attached though.

Failing core tests are excluded in https://github.com/Automattic/vip-go-mu-plugins/pull/1909

### Remove core WordPress test jobs that don’t include vip-go-mu-plugins from vip-go-mu-plugins CircleCi workflow

Core WordPress tests are already validated as running successfully as part of the WordPress release process. Running them without including vip-go-mu-plugins on every change to vip-go-mu-plugins isn’t necessary.

We will continue to run the WordPress core test jobs in CircleCI that include vip-go-mu-plugins on every change to ensure changes work as expected with WordPress core.

https://github.com/Automattic/vip-go-mu-plugins/pull/1910


rinatkhaziev (approved):

pschoffer:
This change is not synced to SVN => no deply needed

PR #1912, opened 2020-12-14
Bump vip-support to 1c47e4c: Don't add roles on init

mjangda:
add_role can lead to front-end writes which are not ideal. It can lead to undesirable circumstances with other roles being removed if there are intermittent database issues.

We still trigger our upgrade routine on admin_init so new installs will still get the roles (and there's still a chance for the latter issue to trigger but the likelihood is significantly reduced).

Also includes a convenience WP-CLI command to force run the upgrade routine if needed (wp vipsupport reset-roles)

https://github.com/Automattic/vip-support/pull/89


nickdaugherty (approved):

nickdaugherty:
r1305-stacks

PR #1921, opened 2020-12-21
Updates the way vip-search versioning information is stored

pschoffer:
## Description
* Stores global indexables to site_option (one per network)
* Stores local indexables to option (one per site)
* Note: for single-site instalation site_option falls back to option.

### Updates the way vip-search versioning information is stored

The versioning information for vip-search index needed to be updated to address global indexables (e.g. User), and different versioning across sites in a multisite environment.

This change stores global indexable separately - one per network. While making sure non-global versions are stored one per site.

## Checklist

## Steps to Test

* Activate user indexables `lando wp elasticpress activate-feature users`
* Add version `lando wp vip-search index-versions add user`
* Check the GLOBAL option: `lando wp eval "var_dump(get_site_option('vip_search_global_index_versions'));"`
* Should have users versions only
* Check the LOCAL option: `lando wp eval "var_dump(get_option('vip_search_index_versions'));"`
* Should have posts versions only
* Check that retrieving version still works as well e.g. `lando wp vip-search index-versions list post`


search/includes/classes/class-versioning.php
```diff
@@ -200,7 +201,19 @@ public function get_inactive_versions( Indexable $indexable ) {
...
+			'global_versions' => get_site_option( self::INDEX_VERSIONS_OPTION_GLOBAL, array() ),
+		];
+
+		$versions = [];
+
+		foreach ( $sources as $source ) {
+			if ( is_array( $source ) ) {
+				$versions = array_merge( $versions, $source );
+			}
+		}
```
pschoffer, line 216: This I think should handle all the use cases, when mix of deprecated and new versions are on site.  Due to the order the globals have overwrite the local settings (which is important when working with old way of storing user - per site locally vs now one per network globally.).

Please let me know if you can see any issues with this approach!

nickdaugherty: @pschoffer there may be a better way to do this. Firstly, we should be checking `$indexable->global` to see where we should pull the data from. Then, if the indexable _is_ global, we can have the fallback to the deprecated option if the right site option doesn't exist or is empty. When this occurs, we should log it so we can see if it's actually affecting any sites (and if not, remove the fallback).

nickdaugherty (changes requested):

search/includes/classes/class-versioning.php
pschoffer: Good idea with getting it based on the global flag!

The fallback needs to stay for both globals and non-global I think. Because the code before would store the versions on network option on multisite no matter if it is a global or non-global indexable.

I also added the logging and a call to update_versions so that it gets stored in the correct place for the next time. This should ensure that the deprecated location can be phased out eventually.

search/includes/classes/class-versioning.php
```diff
@@ -200,27 +201,68 @@ public function get_inactive_versions( Indexable $indexable ) {
...
+							'activated_time' => null,
+						),
+					);
+				} else {
+					return [];
+				}
 			}
+		} else {
+			// Normalize the versions to ensure consistency (have all fields, etc)
+			return array_map( array( $this, 'normalize_version' ), $versions[ $slug ] );
```
nickdaugherty, line 260: In the current implementation, the version is _always_ normalized (it's not behind a conditional) - is this change intentional?

nickdaugherty: Oh I see, this code has been duplicated from another spot...hmm, can we write this so we don't have to duplicate the normalization behavior in multiple places?

search/includes/classes/class-versioning.php
```diff
@@ -200,27 +201,68 @@ public function get_inactive_versions( Indexable $indexable ) {
...
 		}
+	}
 
-		// Normalize the versions to ensure consistency (have all fields, etc)
-		return array_map( array( $this, 'normalize_version' ), $versions[ $slug ] );
+	private function _get_versions_is_slug_defined( $versions, $slug ) {
```
nickdaugherty, line 264: The naming of this function is a little confusing - what if we called it `versions_array_has_slug()` or similar?

search/includes/classes/class-versioning.php
```diff
@@ -200,27 +201,68 @@ public function get_inactive_versions( Indexable $indexable ) {
...
-				return array(
-					1 => array(
-						'number' => 1,
-						'active' => true,
-						'created_time' => null, // We don't know when it was actually created
-						'activated_time' => null,
-					),
+		if ( ! $this->_get_versions_is_slug_defined( $versions, $slug ) ) {
+			// Check deprecated location
+			$deprecated_versions = get_site_option( self::INDEX_VERSIONS_OPTION, array() );
```
nickdaugherty, line 216: Technically, the deprecated location would have only been used if the site is in network mode. To prevent unnecessary option lookups here, should we put this behind the same network mode check that update_versions() used to use?

search/includes/classes/class-versioning.php
nickdaugherty: Hmm I realize now that it's updating the option with the normalized values (so has to be duplicated). In that case, can we extract that functionality to a new function? Can still remove the unnecessary `else` here.

nickdaugherty (changes requested):

search/includes/classes/class-versioning.php
pschoffer: Good idea. Will do.

search/includes/classes/class-versioning.php
pschoffer: My thinking was that the fallback to deprecated solution is temporary so I shouldn't do too much refactoring that wont make sense after that is removed. Where I felt extracting `array_map( array( $this, 'normalize_version' ), $versions[ $slug ] )` belongs. As it seems already kinda ok to read to me. But I don't have strong opinions on that. Although maybe it makes sense to create a function `normalize_versionS` which would handle that mapping over items. What do you think?

With adding the `else`, it just seemed better to be more explicit, that we only ever end up there if the `if` is false. All the branches of that `if` returns directly. Because it is kinda complex if (and will get worse with the network check I guess) it might be easy to overlook though. I am fine with skipping the else if it feels easier on your eyes:)

search/includes/classes/class-versioning.php
pschoffer: Sounds good will do.

search/includes/classes/class-versioning.php
nickdaugherty: Fair enough, let's leave the normalization stuff. For the `else`, I'd go ahead and remove it since it's redundant right now. If future code makes it more complex, we can add it.

search/includes/classes/class-versioning.php
pschoffer: Done

search/includes/classes/class-versioning.php
pschoffer: Done

search/includes/classes/class-versioning.php
pschoffer: Done

nickdaugherty (approved):
Awesome, I think we're good to go here.

pschoffer:
stacks-r1321.

pschoffer:
Deployed to staging stack now

netsuso:
Release v1.4.0

PR #1923, opened 2020-12-22
Add new fields to the site details data

netsuso:
## Description

Add more fields to the site details. Namely:

- Jetpack information (version, active status, Jetpack id, active modules)
- VIP Search status (enabled, query integration)
- Site url (useful mostly for multisite installations)
- Plugin path (in some situations, easier to handle than the plugin name)

### Add more information to site details

We keep working in the Site Details functionality, that will allow us to have up to date details about sites, such as the list of plugins and their versions, whether VIP Search is enabled, Jetpack connection status...

This adds some more fields, namely the information about Jetpack (version, active status, Jetpack id, active modules). In the future, all these will be stored and be accessible to run actions on them. E.g, easily find all sites that have a certain old Jetpack version that needs upgrading.

## Checklist

## Steps to Test

1. Check out PR
1. Run PHP code that calls `Site_Details_Index::instance()->get_site_details()`
1. Verify it works and has the expected values

### DIFF
--- a/config/class-site-details-index.php
+++ b/config/class-site-details-index.php
@@ -63,12 +63,17 @@ class Site_Details_Index {
 
 		$site_details['timestamp'] = $this->get_current_timestamp();
 		$site_details['client_site_id'] = $site_id;
-		$site_details['environment_id'] = $site_id;
 		$site_details['environment_name'] = $environment_name;
-		$site_details['plugins'] = $this->get_plugin_info();
 		$site_details['core']['wp_version'] = strval( $wp_version );
 		$site_details['core']['blog_id'] = get_current_blog_id();
+		$site_details['core']['site_url'] = get_site_url();
 		$site_details['core']['is_multisite'] = is_multisite();
+
+		$site_details['plugins'] = $this->get_plugin_info();
+
+		if ( class_exists( 'Jetpack' ) ) {
+			$site_details['jetpack'] = $this->get_jetpack_info();
+		}
 		
 		return $site_details;
 	}
@@ -93,6 +98,7 @@ class Site_Details_Index {
 			$enabled_via_code = in_array( $key, $plugins_enabled_via_code, true );
 
 			$plugin_info[] = array(
+				'path' => $key,
 				'name' => $value['Name'],
 				'version' => $value['Version'],
 				'active' => $active,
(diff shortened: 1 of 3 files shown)


config/class-site-details-index.php
```diff
@@ -103,6 +111,38 @@ public function get_plugin_info() {
...
+	/**
+	 * Gather basic information about VIP Search for a site
+	 */
+	public function get_search_info() {
+		$search_info = array();
+
+		$search_info['enabled'] = ( defined( 'USE_VIP_ELASTICSEARCH' ) && USE_VIP_ELASTICSEARCH ) ||
+			( defined( 'VIP_ENABLE_VIP_SEARCH' ) && true === VIP_ENABLE_VIP_SEARCH );
+
+		$search_info['query_integration'] = ( defined( 'VIP_ENABLE_ELASTICSEARCH_QUERY_INTEGRATION' ) && true === VIP_ENABLE_ELASTICSEARCH_QUERY_INTEGRATION ) ||
```
nickdaugherty, line 123: This logic is encapsulated in `Search::is_query_integration_enabled()` so we can re-use that.

nickdaugherty: For maximum flexibility and understanding, what if we named this `query_integration_enabled`? Or made it an array - `'query_integration' => array( 'enabled' => $enabled )`.

nickdaugherty (changes requested):
Just noted a few small things, otherwise 👍

config/class-site-details-index.php
netsuso: Using `Search::is_query_integration_enabled()` looks much better! Taking that into account, I was wondering if it would make sense to detect whether VIP is enabled by checking if the `Search` class exists, as it's loaded from `000-vip-init.php` precisely on that condition.

I like the `query_integration_enabled` option, I'll change it.

netsuso: I'm finally checking for the `Search` class. I've also added a comment in the `mu-plugins` loader so in change the loading logic changes, whoever is in change remembers to update the Site Details generation.

netsuso: And I've also changed the Jetpack details generation. Now there's an 'available' field that identifies if the Jetpack code is loaded at all.

nickdaugherty (approved):

pschoffer:
Stacks-r1316.

PR #1924, opened 2020-12-22
Temporarily store site details data in logstash

netsuso:
## Description

We have recently found some situations where having Site Details already available would be very useful. The final implementation will still take some time, but in the meantime we would benefit of Site Details data being stored somewhere, even if it's not the final place.

This PR recovers a couple of commits from the original PR by @brandon-m-skinner - [commit1](https://github.com/Automattic/vip-go-mu-plugins/pull/1768/commits/63beab34d2df1997d3e62cd0835841bcec87b618#) [commit2](https://github.com/Automattic/vip-go-mu-plugins/pull/1768/commits/1143ba1659a733f98a39fd281f70eafb7f55f192).

Code was originally removed then because with `log2logstash` most of the relevant site details would be stored as a JSON string the `extra` field of the `vip-debug` index, which would have prevented the proposed solution of searching directly over these documents on ElasticSearch.

In the current situation, we won't query ElasticSearch to find certain values, so the JSON string limitation won't be a problem. We will possibly use this to just fetch the full JSON documents and process them with an ad-hoc script.

### Store Site Details in a temporary location

We keep working in the Site Details functionality, that will allow us to have up to date details about sites, such as the list of plugins and their versions, whether VIP Search is enabled, Jetpack connection status...

While the final implementation is still in progress, we are now storing site details in a temporary location in logstash, so that we can start using this data as soon as possible. One example is finding all sites that are running a certain version of a plugin, to prepare for an upgrade to a newer version.

## Checklist

## Steps to Test

1. Check out PR in a VIP Go Sandbox
1. Check if a new logstash entry for feature `a8c_vip_site_details`  has been added to `/chroot/tmp/logstash.log`
1. If not (it should only happen once every eight executions), keep running it until the logstash entry is added

### DIFF
--- a/config/class-sync.php
+++ b/config/class-sync.php
@@ -129,7 +129,7 @@ class Sync {
 		// In order to reduce the amount of stored documents, we increase the execution interval
 		// from 30 minutes to approximately 4 hours
 		if ( 1 === rand( 1, 8 ) ) {
-			require_once( 'class-site-details-index.php' );
+			require_once( __DIR__ . '/class-site-details-index.php' );
 
 			Site_Details_Index::instance()->put_site_details_in_logstash();
 		}
--- a/config/class-sync.php
+++ b/config/class-sync.php
@@ -65,6 +65,7 @@ class Sync {
 
 	public function do_cron() {
 		$this->maybe_sync_jetpack_privacy_settings();
+		$this->put_site_details();
 	}
 
 	public function maybe_sync_jetpack_privacy_settings() {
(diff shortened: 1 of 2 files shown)


config/class-sync.php
```diff
@@ -123,6 +124,17 @@ public function sync_jetpack_privacy_settings( $expected_value, $value_from_opti
 		return $success;
 	}
 
+	public function put_site_details() {
+		// This is a temporary workaround while the full Site Details implementation is finished
+		// In order to reduce the amount of stored documents, we increase the execution interval
+		// from 30 minutes to approximately 4 hours
+		if ( 1 === rand( 1, 8 ) ) {
+			require_once( 'class-site-details-index.php' );
```
nickdaugherty, line 132:
```suggestion
			require_once( __DIR__ .  '/class-site-details-index.php' );
```

In general it's more robust to use an absolute path (and it's what we do in most other spots in mu-plugins).

nickdaugherty (changes requested):
Just noted one minor thing, otherwise looks great.

config/class-sync.php
netsuso: Sure! Changed that

nickdaugherty (approved):

pschoffer:
Stacks-r1317.

PR #1929, opened 2021-01-07
IpUtils - clearing up phpcs errors

pschoffer:
## Description

Same as bellow

#
### Phpcs error cleanup - IpUtils class

As part of the ongoing work on cleaning up legacy we go over files not touched in a while and make sure they still comply with our code standards. This change updates the style of IpUtils library.

## Checklist

## Steps to Test

Not sure really. The code is covered in unit tests pretty well so maybe no need for manual testing.

### DIFF
--- a/lib/proxy/ip-forward.php
+++ b/lib/proxy/ip-forward.php
@@ -22,7 +22,7 @@ function fix_remote_address( $user_ip, $remote_proxy_ip, $proxy_ip_whitelist ) {
 	require_once( __DIR__ . '/ip-utils.php' );
 
 	// Verify that the remote proxy matches our whitelist
-	$is_whitelisted_proxy_ip = IpUtils::checkIp( $remote_proxy_ip, $proxy_ip_whitelist );
+	$is_whitelisted_proxy_ip = IpUtils::check_ip( $remote_proxy_ip, $proxy_ip_whitelist );
 
 	if ( ! $is_whitelisted_proxy_ip ) {
 		return false;
--- a/lib/proxy/ip-utils.php
+++ b/lib/proxy/ip-utils.php
@@ -16,121 +16,117 @@ namespace Automattic\VIP\Proxy;
  *
  * @author Fabien Potencier <fabien@symfony.com>
  */
-class IpUtils
-{
-    /**
-     * This class should not be instantiated.
-     */
-    private function __construct()
-    {
-    }
+class IpUtils {
 
-    /**
-     * Checks if an IPv4 or IPv6 address is contained in the list of given IPs or subnets.
-     *
-     * @param string       $requestIp IP to check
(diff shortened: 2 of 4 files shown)


lib/proxy/class-iputils.php
```diff
@@ -0,0 +1,135 @@
+<?php
```
pschoffer, line 1: Uups this will be harder to review. As part of the rules that needed some fixin the name of the file needed change. Not it looks like whole new blob. Probably it's better to change diff locally with git/your editor.

GaryJones: Could split the changes up into one commit for the file name change (and related reference updates), and the other for the changes in the contents of this file. Then each commit can be viewed individually.

lib/proxy/class-iputils.php
```diff
@@ -0,0 +1,135 @@
...
+
+	/**
+	 * Checks if an IPv4 or IPv6 address is contained in the list of given IPs or subnets.
+	 *
+	 * @param string       $request_ip IP to check
+	 * @param string|array $ips       List of IPs or subnets (can be a string if only a single one)
+	 *
+	 * @return bool Whether the IP is valid
+	 */
+	// phpcs:ignore WordPress.NamingConventions.ValidFunctionName.MethodNameInvalid
```
GaryJones, line 35: The insertion here techincally breaks the association of the DocBlock above with the method.

Could consider moving the ignore line to the end of the next line instead.

Also, how widely used is the method name? Could it be renamed / new `check_ip()` method added and have this one point to it, etc?

lib/proxy/class-iputils.php
pschoffer: Good idea! I thought it would make the git-hook we have there enforcing style not let me push, but it did!

lib/proxy/class-iputils.php
pschoffer: Turns out not that many:) I changed the name now.

GaryJones: As these were public methods, have you checked across the customer code bases to see if anyone is using the old method names?

netsuso: I've grepped through all client site repos and I haven't seen any usage of this class in `mu-plugins`, so the changed should be safe

pschoffer: Thanks Suso! To add to this all the requires of this file are scoped within functions, so the client would have to require the code themselv.

GaryJones: Great, thanks for checking 👍🏻

brandon-m-skinner (approved):

pschoffer:
Stacks-r1322.

netsuso:
Release v1.4.0

PR #1933, opened 2021-01-08
Do not validate counts if Search index doesn't exist

brandon-m-skinner:
## Description

In a multisite context, if one site has VIP Search all sites have access to VIP Search. Even if they don't have an index. This is problematic as it logs false positives in our health checks.

To prevent this from happening, we need to skip trying to validate counts if an index doesn't exist.

### Bug Fix: Skip Elasticsearch index / database count comparison health check if no index exists

In a multisite context, if one site has VIP Search all sites have access to VIP Search. Even if they don't have an index. This is problematic as it logs false positives in our health checks. To prevent this from happening, we now skip site / indexable combos that don't exist.

https://github.com/Automattic/vip-go-mu-plugins/pull/this-pr

## Checklist

## Steps to Test

1. `wp vip-search health validate-counts` in a multisite that has VIP Search but on a subsite that doesn't have an index.
2. You should get an error message.
3. Apply PR.
4. Repeat step 1.
5. You should get a skipped message.

---------------

1. `wp vip-search health validate-counts --network-wide` in a multisite that has VIP Search but has subsite(s) that don't have indexes.
2. You should get an error message for each subsite without indexes.
3. Apply PR.
4. Repeat step 1.
5. You should get skipped messages for each subsite without an index rather than errors.

### DIFF
--- a/search/includes/classes/commands/class-healthcommand.php
+++ b/search/includes/classes/commands/class-healthcommand.php
@@ -85,6 +85,12 @@ class HealthCommand extends \WPCOM_VIP_CLI_Command {
 	 * @param array $assoc_args CLI arguments
 	 */
 	private function validate_indexable_count( $indexable_slug, $assoc_args ) {
+		$indexable = \ElasticPress\Indexables::factory()->get( $indexable_slug );
+		if ( ! $indexable ) {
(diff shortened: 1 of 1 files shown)


nickdaugherty (approved):
Code looks good, I haven't yet tested it.

netsuso:
stacks-r1326

netsuso:
Release v1.4.1

PR #1937, opened 2021-01-12
Renames ip-utils.php to class-iputils.php

pschoffer:
## Description

Follow up on https://github.com/Automattic/vip-go-mu-plugins/pull/1929

### Phpcs error cleanup - IpUtils class renaming 1

As the last part of phpcs cleanup, we plan to rename `iputils.php` file to adhere to our naming standards. To increase the safety of this change only introduce the new file. Next changes will use the new file. This was done as a follow-up to be isolated change from the previous cleanup done on this file.

## Checklist

## Steps to Test
Not sure really. The code is covered in unit tests pretty well so maybe no need for manual testing.


mjangda:
Note: the deploy for this should be done in three separate commits / deploys:

1. Add the new `class-iputils.php` file (commit + deploy)
2. Update the `require` calls (commit + deploy) 
3. Delete the old file (commit + deploy)

pschoffer:
@mjangda Oh, I don't think that would be easily done with the new staging flow. I will break it down some more to 3 PRs then.

Is it because there is a risk of files being cached somehow?

mjangda:
> Is it because there is a risk of files being cached somehow?

It's because deploys are not atomic right now (just runs `svn up`) which means the the rename can happen before the update to the files with the `require` calls. There's a small window where requests hit the app between the rename and updates to the other files which can result in fatals.

pschoffer:
@mjangda  Oh, wouldn't that be an issue to most the multifile changes then? Say you rename a function (or introduce a new one) and use it in another file, don't you get the same issue? Or is that not a fatal error and therefore acceptable risk as oppose to failing to find the file?

pschoffer:
I limited this change to the new file as per @mjangda suggestion above.

netsuso (approved):
Just adding a file that is a copy from another one already reviewed... Looks good to me! ;)

netsuso:
stacks-r1333

netsuso:
Release v1.5.0

PR #1940, opened 2021-01-14
Adds changelog creation automation

pschoffer:
### Automates Changelog Creation

This change adds a circleCI job, that runs on master after each merge. It finds the latest closed PR and create a Changelog record *draft* based on the PR description.

## Checklist

## Steps to Test

Well we need to merge this to see I guess, but this change doesn't touch any production code.


bin/github-changelog.php
```diff
@@ -0,0 +1,93 @@
...
+            }
+            $result = $result . "\n" . $line;
+        }
+    }
+    return $result;
+}
+
+function get_changelog_html( $pr ) {
+    $Parsedown = new Parsedown();
+    $description_html =  $Parsedown->text( $pr['body'] );
```
netsuso, line 42:
```suggestion
    $body = preg_replace( '/<!--(.|\s)*?-->/', '', $pr['body'] );
    $description_html =  $Parsedown->text( $body );
```
I suggest removing HTML comments from the PR description (some PRs are keeping the original comments in the default description), to avoid them reaching the changelog post.

bin/github-changelog.php
```diff
@@ -0,0 +1,93 @@
...
+
+function fetch_last_PR() {
+    $ch = curl_init( GITHUB_ENDPOINT );
+    curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
+    curl_setopt($ch, CURLOPT_HEADER, 0);
+    curl_setopt($ch, CURLOPT_HTTPHEADER, [ 'User-Agent: script' ] );
+    $data = curl_exec($ch);
+    curl_close($ch);
+
+    return json_decode($data, true)[0];
```
netsuso, line 19: I think this will work fine most of the times, but there could be some race conditions (e.g, someone posts a comment to another PR before the CircleCI job is executed). A safer approach would be to look at all fetched PRs for the `merged_at` field, and simply keep the PR with the most recent merge time.

netsuso (changes requested):
It works great, thank you so much! :) :bow:   I left a couple of suggestions as comments.

bin/github-changelog.php
pschoffer: Cool yeah, I noticed though... but didn't trust my regexp skills enough:)

bin/github-changelog.php
pschoffer: Great idea. Will do.

pschoffer: done

netsuso: Hum... I think I didn't explain myself :sweat_smile:   I meant not just filtering by PRs with a `merged_at` field, but also traversing all of them and keeping the one with the most recent `merged_at` (i.e, the last merged PR, instead of the last updated one, that might be due to some comment or changed label)

netsuso: Of course in an extreme case it might happen that the most recently merged PR is not in the list of the 10 most recent updated ones, but I doubt that's a real possibility 😉

pschoffer: Oh, I thought you were worried it would pick some other PR that was just closed without merging... in both cases it seems like really an edge case, right? Somebody would have to comment on one of the merged PRs between the time the last pr was merged and the build pipeline was finished.

I don't think that would happen that often ...and the impact is kinda small. But if you think it is worth doing I will check it out some more tomorrow.

netsuso: More than a comment I would expect some other change (an added/removed label, e.g). But yes, probably the case of a dismissed PR would be the most plausible, and that's covered with your recent change. As any other case would be rare, we can leave it this way, and see how it behaves

netsuso (approved):

PR #1943, opened 2021-01-14
Add attachment to indexable post types when protected content enabled

brandon-m-skinner:
## Description

Presently there's potential for a race condition between setting / unsetting attachment as an indexable post type. To combat this and meet the expectations previously set, we should add attachment as an indexable post type in VIP Search when protected content is enabled. This is done with priority `PHP_INT_MAX - 1` so that `PHP_INT_MAX` can be used to remove attachment from the indexable post types if needed in some situations. 
### Ensure 'attachment' is set as an indexable post type in VIP Search when protected content feature is enabled

Presently there's potential for a race condition between setting / unsetting attachment as an indexable post type. To combat this and meet the expectations previously set, we should add attachment as an indexable post type in VIP Search when protected content is enabled. This is done with priority `PHP_INT_MAX - 1` so that `PHP_INT_MAX` can be used to remove attachment from the indexable post types if needed in some situations. 

https://github.com/Automattic/vip-go-mu-plugins/pull/this-pr

## Checklist

## Steps to Test

1. Apply PR.


search/includes/classes/class-search.php
```diff
@@ -1482,6 +1489,26 @@ public function filter__ep_prepare_meta_allowed_protected_keys( $keys, $post ) {
...
 
+	/**
+	 * Since we've established that enabling the protected content feature causes attachments
+	 * to be indexed, we should ensure that 'attachment' is in the indexable post types if
+	 * protected content is enabled.
+	 *
+	 * @param array $indexable_post_types Current list indexable post types in VIP Search.
+	 * @return array New list of indexable post types in VIP Search.
+	 */
+	public function filter__ep_indexable_post_types( $indexable_post_types ) {
```
pschoffer, line 1500: What do you think about naming this method according to content? Or is that some code style to name following the filter?

I would suggest `filter__ep_indexable_post_types__add_attachment_to_post_types` or something like that. I know this is super long name, but maybe better descriptive?

pschoffer: Otherwise, it looks great!

brandon-m-skinner: Switch it. We're not bound by convention 😆

search/includes/classes/class-search.php
```diff
@@ -1482,6 +1489,26 @@ public function filter__ep_prepare_meta_allowed_protected_keys( $keys, $post ) {
...
+	 * @param array $indexable_post_types Current list indexable post types in VIP Search.
+	 * @return array New list of indexable post types in VIP Search.
+	 */
+	public function add_attachment_to_ep_indexable_post_types( $indexable_post_types ) {
+		if ( ! is_array( $indexable_post_types ) ) {
+			return $indexable_post_types;
+		}
+		
+		if ( ! in_array( 'attachment', $indexable_post_types, true ) ) {
+			$indexable_post_types[] = 'attachment';
```
nickdaugherty, line 1506: This array is actually an associative array, here's an example: https://github.com/Automattic/ElasticPress/blob/4ac57414105483818c2ad6464e50547248f49a20/includes/classes/Feature/Documents/Documents.php#L81

search/includes/classes/class-search.php
```diff
@@ -389,6 +389,13 @@ protected function setup_hooks() {
 
 		// Do not show the above compat notice since VIP Search will support whatever Elasticsearch version we're running
 		add_filter( 'pre_option_ep_hide_es_above_compat_notice', '__return_true' );
+
+		// If protected content is enabled, ensure that the attachment post type is an indexable post type.
+		// Set the priority to PHP_INT_MAX - 1 so customers can use PHP_INT_MAX to unset it if needed.
+		if ( false !== $this->is_protected_content_enabled() ) {
+			$priority = PHP_INT_MAX - 1;
```
nickdaugherty, line 396: PHP_INT_MAX (the priority needed to override this) is pretty final - is there a lower priority we can use that leaves more flexibility for customers to override?

nickdaugherty (changes requested):
Looks good, just noted two things

nickdaugherty (dismissed):

brandon-m-skinner:
Needs some minor reworking to use the Feature functionality in EP rather than using the options directly.

brandon-m-skinner:
r1341-stacks

netsuso:
Release v1.7.0
https://github.com/Automattic/vip-go-mu-plugins/releases/tag/v1.7.0

PR #1948, opened 2021-01-20
Add Files ACL Restriction Functionality

mjangda:
## Description

Builds on top of #1902

Add the ability to leverage the Files ACL framework and restrict access to files. Includes two variants:

**Restrict access to unpublished files**

For example, images on a draft post.

If media is attached to an unpublished post, that media should not be visible. We do a reverse lookup (from path to attachment ID) and then find a parent and make a determination based on the state of the parent.

**Restrict access to all files**

For example, private sites.

Useful for intranets and other private sites that need to restrict access to all content.

### Loader

Includes a loader to determine which restriction to load (since they're not compatible). It requires a constant to be set (via VIPd / ConfigMap / wp-config) and an option set correctly to enable either all file restriction or unpublished file restrictions. We're using options to allow targeting functionality on a per-subsite level. 

### Files ACL 

We partially implemented Files ACL restriction functionality. This code is not enabled in production just yet.

## Checklist

## Steps to Test

**No Constant**

If the constant `VIP_FILES_ACL_ENABLED` is not defined, nothing should happen.

**No Constant but options**

`wp option set vip_files_acl_restrict_all_enabled 1` or `wp option set vip_files_acl_restrict_unpublished_enabled 1` without the constant defined should trigger warnings (`File ACL restrictions are enabled without server configs`).

**Restrict All Files**

- Add `define( 'VIP_FILES_ACL_ENABLED', true );`
- `wp option set vip_files_acl_restrict_all_enabled 1`
- Generate an attachment:

```
WP="lando wp"

IMAGE_ID=`$WP media import --title="Attachment" --porcelain "http://s0.wp.com/i/void.gif"`
IMAGE_URL=`$WP post get $IMAGE_ID --field=guid`

echo "== Attachment URL"
echo $IMAGE_URL
```

- Then test the request:

```
curl 'http://vip-go-dev.lndo.site/wp-content/mu-plugins/files/acl/endpoint-check-file-acl.php' -H 'X-Original-Uri: /wp-content/uploads/path/to/file.png'
```

**Restrict Unpublished Files**

- Add `define( 'VIP_FILES_ACL_ENABLED', true );`
- `wp option set vip_files_acl_restrict_all_enabled 0`
- `wp option set vip_files_acl_restrict_unpublished_enabled 1`
- Generate attachments:

```
WP="lando wp --url=http://vip-go-dev.lndo.site/testing/"

DRAFT_POST_ID=`$WP post create --post_type=post --post_status=draft --post_title="Draft Post" --porcelain`
DRAFT_IMAGE_ID=`$WP media import --post_id=$DRAFT_POST_ID --title="Attachment on Draft Post" --porcelain "http://s0.wp.com/i/void.gif"`
DRAFT_IMAGE_URL=`$WP post get $DRAFT_IMAGE_ID --field=guid`

echo "== Attachment on Draft Post"
echo $DRAFT_IMAGE_URL

PUBLISH_POST_ID=`$WP post create --post_type=post --post_status=publish --post_title="Published Post" --porcelain`
PUBLISH_IMAGE_ID=`$WP media import --post_id=$PUBLISH_POST_ID --title="Attachment on Published Post" --porcelain "http://s0.wp.com/i/void.gif"`
PUBLISH_IMAGE_URL=`$WP post get $PUBLISH_IMAGE_ID --field=guid`

echo "== Attachment on Published Post"
echo $PUBLISH_IMAGE_URL

UNATTACHED_IMAGE_ID=`$WP media import --title="Unattachment Attachment" --porcelain "http://s0.wp.com/i/void.gif"`
UNATTACHED_IMAGE_URL=`$WP post get $UNATTACHED_IMAGE_ID --field=guid`

echo "== Attachment with no post"
echo $UNATTACHED_IMAGE_URL
```

Now try `curl` the files via the endpoint. Publish and un-attached should return 202. Draft should return 403. If you make a request with auth cookies for a logged in user (e.g. editor), all should return 202 and draft should include `X-Private: true`.


files/acl/acl.php
```diff
@@ -1,11 +1,59 @@
...
+	}
+
+	if ( $is_restrict_unpublished_enabled ) {
+		require_once( __DIR__ . '/files/acl/restrict-unpublished-files.php' );
+		return;
+	}
+}
+
+function get_option_as_bool( $option_name, $default = false ) {
+	$value = get_option_as_bool( 'vip_files_acl_restrict_all_enabled', false );
```
rinatkhaziev, line 46: @mjangda This reads like a bad copy-pasta?

mjangda: Yep :) Good catch!

mjangda: Probably need a test for this.

In the future, might be worth abstracting this out to a shared lib because I copied this from VIP Search.

files/acl/acl.php
```diff
@@ -1,11 +1,59 @@
...
+
 namespace Automattic\VIP\Files\Acl;
 
 const FILE_IS_PUBLIC = 'FILE_IS_PUBLIC';
 const FILE_IS_PRIVATE_AND_ALLOWED = 'FILE_IS_PRIVATE_AND_ALLOWED';
 const FILE_IS_PRIVATE_AND_DENIED = 'FILE_IS_PRIVATE_AND_DENIED';
 
+add_action( 'muplugins_loaded', __NAMESPACE__ . '\maybe_load_restrictions' );
+
+function maybe_load_restrictions() {
```
mjangda, line 19: Need tests for this.

files/acl/restrict-all-files.php
```diff
@@ -0,0 +1,25 @@
...
+
+	$user_has_read_permissions = current_user_can( 'read' );
+	if ( ! $user_has_read_permissions ) {
+		return FILE_IS_PRIVATE_AND_DENIED;
+	}
+
+	return FILE_IS_PRIVATE_AND_ALLOWED;
+}
+
+add_filter( 'vip_files_acl_file_visibility', __NAMESPACE__ . '\check_file_visibility', 10, 2 );
```
mjangda, line 25: Should this `add_filter` call be moved to `maybe_load_restrictions`?

files/acl/restrict-unpublished-files.php
```diff
@@ -0,0 +1,97 @@
...
+		}
+		
+		return FILE_IS_PRIVATE_AND_DENIED;
+	}
+
+	return FILE_IS_PUBLIC;
+}
+
+// TODO: move these to loader?
+add_filter( 'vip_files_acl_file_visibility', __NAMESPACE__ . '\check_file_visibility', 10, 2 );
```
mjangda, line 48: Should this `add_filter` call be moved to `maybe_load_restrictions`?

mjangda:
We should ideally merge + deploy + rebase #1955 first.

files/acl/acl.php
mjangda: Done in 98616d4 and 6859d2a

files/acl/acl.php
mjangda: Added in 6859d2a

mjangda:
We should merge #1955 first and then rebase.

files/acl/restrict-all-files.php
mjangda: Moved.

files/acl/restrict-unpublished-files.php
mjangda: Moved.

mjangda:
Rebased against master so should be good for another look.

files/acl/restrict-unpublished-files.php
```diff
@@ -0,0 +1,94 @@
...
+			foreach ( $results as $result ) {
+				if ( $path === $result->meta_value ) {
+					$attachment_id = $result->post_id;
+					break;
+				}
+			}
+		}
+	}
+
+	wp_cache_set( $cache_key, $attachment_id, CACHE_GROUP, 5 * MINUTE_IN_SECONDS );
```
rinatkhaziev, line 91: Not a blocker, all things considered, but we should probably guard against a stampede here. 
```suggestion
	wp_cache_set( $cache_key, $attachment_id, CACHE_GROUP, 5 * MINUTE_IN_SECONDS + rand( 1, 30 );
```

rinatkhaziev (approved):
@mjangda  Thank you, looks great! 

A couple of minor notes/considerations we may want to address later:
1. See inline comment about randomizing cache TTL a bit to prevent stampedes.
2. Maybe we should also consider increasing the TTL in general - how often do we expect the attachment paths to change?
3. Wondering whether using APCu as a local cache if available might make sense - it would in theory allow at least some requests to skip Memcached network request - it may be of marginal use though.

PR #1949, opened 2021-01-21
Updates ElasticPress with the latest code

pschoffer:
## Description

This PR brings over the following PRs for the ElasticPress submodule:
https://github.com/Automattic/ElasticPress/pull/68

We updated ElasticPress dependency for VIP Search to enhance the data format about the post author stored in ElasticSearch.

## Checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -43,8 +43,6 @@ Not a lot of significant changes in this patch release, just bugfixes and compat
 - WordPress.com Toolbar: avoid Fatal errors when the feature is not active.
 - WordPress.com Toolbar: avoid 404 errors when loading the toolbar.
 
-https://github.com/Automattic/vip-go-mu-plugins/pull/1905
-
 Example for a feature change:
 
 ### New Filters: Adjust Brute Force Thresholds
@@ -65,14 +63,10 @@ add_filter( 'wpcom_vip_ip_username_login_threshold', function( $threshold, $ip, 
     return $threshold;
 }, 10, 3 );
 ```
-
-https://github.com/Automattic/vip-go-mu-plugins/pull/1782
 -->
 ### The title of my changelog entry(See examples in the comments above)
 
 Look at my sweet description of my changes with examples, screen shots, etc... that demonstrate the functionality.
-
-https://github.com/Automattic/vip-go-mu-plugins/pull/this-pr
 
 ## Checklist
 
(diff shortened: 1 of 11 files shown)


brandon-m-skinner:
Looks okay. Just some questions:

1. How does this change existing user indexes?
2. Does this require us to re-index all user indexes we currently have?
3. Will this cause some unintended on current customers who are currently making heavy use of the user indexes(yes, there are some)?

pschoffer:
Good questions... @brandon-m-skinner 

1) Firstly this influences Post indexable not User. It changes, what we store for the `post_author.login`. From my understanding the idea here is to be able to search posts using the user nickname as that is a common use-case.

The issue was that the user's nickname is sanitized login, where we used to store non-sanitized login. So that would match unless the login has special characters (that gets dropped in sanitization) or spaces (replaced with `-`).

2) Ideally yes, but I am not sure if that is feasible? If we wouldn't, I guess the results from ES might be inconsistent but not totally wrong + search queries may act up a bit. There is, however, `post_author.raw` that contains the same value `post_author.login` did before this change. So we could advise querying against this field if somebody still wants to query non sanitized field for some reason.

3) Apart from the things mentioned in 2) I don't see other consequences.

@rebeccahum @nickdaugherty  could you please add to/correct my answers above? Also, how do you feel about re-indexing. Could we get away with not doing it? How difficult is it to do?

brandon-m-skinner:
Reinforcing everything is a big lift and leaving everything deliberately in an inconsistent state for a day or two might be problematic.

If it comes to this, how can these risks be mitigated?

rebeccahum:
> There is, however, post_author.raw that contains the same value post_author.login did before this change. So we could advise querying against this field if somebody still wants to query non sanitized field for some reason.

Could we do a risk report of the current VIP Search sites to see how impactful this would be? It might end up only affecting 2-3 authors per site.

brandon-m-skinner:
There's no easy way to query all the indexes unfortunately. Would require some questionable hackery.

Not to mention reindexing everything would 100% hit rate limiting and back up every queue. This could potentially cause widespread content discrepancies.

rebeccahum:
Do we need to re-index everything at once or can it be throttled on a site-basis?

nickdaugherty:
Technically we could get away without re-indexing most sites - it's only sites that are querying on user login that would notice a difference, and only for users that have a login that would be affected by sanitize_title() (which is not the norm AFAIK). 

The full content validation would be failing on any site we don't re-index, but we don't run that automatically anywhere so not a big deal.

netsuso:

PR #1951, opened 2021-01-22
Update/search default meta whitelist

pschoffer:
### Plugin Updated: Search

In order to make future migration from Jetpack search easier, we implemented support for a `jetpack_sync_post_meta_whitelist` filter.

## Checklist

## Steps to Test
1. Create post with some meta - `lando wp post create --post_title="TEst title" --post_content="Content" --post_status="publish" --meta_input='{"foo": 1, "bar": 2}'`
1. Check meta - `lando wp vip-search documents get post <id> --format=json | jq ".[].meta"` - no meta should be present.
1. Add a vip-search allow list filter somewhere to your site `\add_filter('vip_search_post_meta_allow_list', function( $keys ) { return array_merge($keys, [ 'foo' ]); }, 0);`
1. Reindex - `lando wp vip-search index`
1. Check meta - should return **only** the one from your filter (`foo` in exable above)
1. Add jetpack filter - `\add_filter('jetpack_sync_post_meta_whitelist', function( $keys ) { return array_merge($keys, [ 'bar' ]); }, 0);`
1. Reindex
1. Check meta  - should return also the one form jetpack filter (both `foo` and `bar` from example above).

Note: Make sure your `vip_search_post_meta_allow_list` adds new values and does not ignore the defaults or remove it all together.


search/includes/classes/class-search.php
```diff
@@ -11,9 +11,54 @@ class Search {
...
-	// sites.
+	/**
+	 * Empty for now. Will flesh out once migration path discussions are underway and/or the same meta are added to the filter across many
+	 * sites.
+	 */
 	public const POST_META_DEFAULT_ALLOW_LIST = array();
+	/**
+	 * Jetpack Default Allow list to make migration path easier. See usage to see how is jetpack allow_list used.
+	 */
+	public const JETPACK_POST_META_DEFAULT_ALLOW_LIST = array(
```
pschoffer, line 22: I chose to copy this in rather than have external dependency especially considering possible issues with different versions of JP clients can be on. What do you think?

search/includes/classes/class-search.php
```diff
@@ -1485,8 +1547,12 @@ function( $value ) {
 		return $post_meta_allow_list;
 	}
 
+	public function get_post_meta_allow_list( $post ) {
+		return $this->parse_vip_search_post_meta_allow_list_filter( $post );
+	}
+
 	public function filter__ep_prepare_meta_allowed_protected_keys( $keys, $post ) {
-		return \apply_filters( 'vip_search_post_meta_allow_list', $keys, $post );
+		return $this->parse_vip_search_post_meta_allow_list_filter( $post, $keys );
```
pschoffer, line 1555: The previous implementation would fail if the vip_search_post_meta_allow_list would return assoc_array here, so I lifted up that logic to a new function `parse_vip_search_post_meta_allow_list_filter`

tests/search/test-class-search.php
```diff
@@ -2,6 +2,7 @@
 
 namespace Automattic\VIP\Search;
 
+require_once __DIR__ . '/../../search/search.php';
```
pschoffer, line 5: I moved this up so I have access to the Search constants in the data providers (that seems to run before setup).

search/includes/classes/class-search.php
```diff
@@ -11,9 +11,54 @@ class Search {
...
 	public const SEARCH_ALERT_LEVEL = 2; // Level 2 = 'alert'
-	// Empty for now. Will flesh out once migration path discussions are underway and/or the same meta are added to the filter across many
-	// sites.
+	/**
+	 * Empty for now. Will flesh out once migration path discussions are underway and/or the same meta are added to the filter across many
+	 * sites.
+	 */
 	public const POST_META_DEFAULT_ALLOW_LIST = array();
+	/**
+	 * Jetpack Default Allow list to make migration path easier. See usage to see how is jetpack allow_list used.
```
brandon-m-skinner, line 20:
```suggestion
	 * Jetpack default post meta allow list to make migration path easier. See usage of jetpack_sync_post_meta_whitelist to see how this is used.
```

search/includes/classes/class-search.php
```diff
@@ -1445,7 +1490,24 @@ public function filter__ep_skip_post_meta_sync( $skip_sync, $post, $meta_id, $me
 		return false;
 	}
 
-	public function get_post_meta_allow_list( $post ) {
+	/**
+	 * Processes vip_search_post_meta_allow_list. This method serves two purposes. Fistly handles assoc array convertion and secondly handles
```
brandon-m-skinner, line 1494:
```suggestion
	 * Processes vip_search_post_meta_allow_list. This method serves two purposes. Firstly handles assoc array conversion and secondly handles
```

search/includes/classes/class-search.php
```diff
@@ -1445,7 +1490,24 @@ public function filter__ep_skip_post_meta_sync( $skip_sync, $post, $meta_id, $me
...
+	 * default keys creation. To make migration from jetpack search easier this might mean leveraging jetpack_sync_post_meta_whitelist filter.
+	 * In case we already have some default keys defined we skip defualt key generation.
```
brandon-m-skinner, line 1496:
```suggestion
	 * In case we already have some default keys defined we skip default key generation.
```

search/includes/classes/class-search.php
```diff
@@ -1445,7 +1490,24 @@ public function filter__ep_skip_post_meta_sync( $skip_sync, $post, $meta_id, $me
...
+	 *
+	 * @param {WP_Post} $post The post whose meta data is being prepared
```
brandon-m-skinner, line 1498:
```suggestion
	 * @param {WP_Post} $post The post whose meta data is being prepared.
```

search/includes/classes/class-search.php
```diff
@@ -1445,7 +1490,24 @@ public function filter__ep_skip_post_meta_sync( $skip_sync, $post, $meta_id, $me
...
+	 * @param {array | null} $default_keys Optional parameter to skip default keys generation (used as part of ep_prepare_meta_allowed_protected_keys processing).
+	 * @return {array} The new allow list for post_meta_indexing
```
brandon-m-skinner, line 1500:
```suggestion
	 * @return {array} The new allow list for post_meta_indexing.
```

brandon-m-skinner (commented):
Fixed a few typos and added some periods to doc block stuff(part of WordPress coding standards).

I'm not sure that separating them and having multiple streams of defaults in the best choice here. If people want to blank the defaults, replace them, or add on to them they can use the appropriate filter.

The intent is to provide a list of default post meta that's generally useful out of the box in all scenarios. If the meta doesn't exist, it doesn't get indexed anyway. Which means we probably should just combine the Jetpack ones with whatever our defaults end up being. That way we're not managing different streams of defaults or dealing with unexpected behaviour if the "magic" kicks in unexpectedly. We're not the ones who will be handling this sort of thing for the most part so it should be as simple as possible, consistent, and not have any potential for surprises.

pschoffer:
Thanks @brandon-m-skinner for all the fixes!! I am not sure how to answer the conditional defaults answer. What do you think @nickdaugherty ?

search/includes/classes/class-search.php
nickdaugherty: Thinking about this...there's definitely pros and cons to both approach, but since we're applying the `jetpack_sync_post_meta_whitelist` filter later, we already have an external dependency. Instead of hardcoding this list (which will need to be manually updated) and manually applying the filter, what if we just call `Automattic\Jetpack\Sync\Defaults::get_post_meta_whitelist()`?

The downside of this is that upstream changes in Jetpack (changes to default meta) will throw our index consistency off (if checking for full content consistency). At some point in the future, we'll (ideally) have all VIP sites using VIP Search, at which point being tied to JP in this way becomes a liability. 

Need to think this through more, I'll start a discussion thread in Slack.

nickdaugherty: The consensus from the Slack discussion is that we should leave it as-is and not add another direct dependency on Jetpack. These change so infrequently that using JP directly would be an over-optimization, and this way there are no surprises.

nickdaugherty:
I agree with @brandon-m-skinner's thoughts - the conditionals probably make it too complicated, and aren't quite the "WordPress way" that filters work (which is you pass it a value, and the filter simply modifies it or not). It could also be that I'm not fully understanding the need for the conditionals, in which case I may be wrong 🤷

pschoffer:
Oh OK, so are we saying we should drop the condition of having the jetpack_sync_post_meta_whitelist filter added and thereforw always do `$default_allow_list = \apply_filters( 'jetpack_sync_post_meta_whitelist', self::JETPACK_POST_META_DEFAULT_ALLOW_LIST );` making `JETPACK_POST_META_DEFAULT_ALLOW_LIST` our vip search defaults and dropping the `POST_META_DEFAULT_ALLOW_LIST` as that would no longer be used?

nickdaugherty:
@pschoffer there are a few issues with the current conditional logic - firstly, the 2nd and 3rd conditions will never be reached:

The `ep_prepare_meta_allowed_protected_keys` filter gets an empty array as the first argument by default: https://github.com/Automattic/ElasticPress/blob/7169b4bebe4fa69e1c0c9475b0cddb89b7446982/includes/classes/Indexable/Post/Post.php#L653

So the first check will always pass, unless some other filter set it to a non-array value: https://github.com/Automattic/vip-go-mu-plugins/pull/1951/files#diff-dd3f02b793bcc2930a3ba1d70b0520c0865dc4a09d2dbd1f49a1c55af101c40dR1509

I don't think the `\has_filter( 'jetpack_sync_post_meta_whitelist' )` check is what we want - the idea is to match our defaults with that of Jetpack, even if the customer hasn't filtered this to change it. Jetpack hooks into this filter in several spots, so I think this would always be true regardless of if a customer implemented it. So we want to run this filter regardless, then merge the results with our normal defaults, which are in `POST_META_DEFAULT_ALLOW_LIST` (currently empty though). We can probably simplify the two constants to remove the new JP meta list and just use `POST_META_DEFAULT_ALLOW_LIST`. However, that list is also the same list that will come through when running the `jetpack_sync_post_meta_whitelist` filter, from here, which means that it will override the list in our constants: https://github.com/Automattic/jetpack/blob/6f3b04a85f6aa095156975b34b2d1d4872f37d4a/vendor/automattic/jetpack-sync/src/class-defaults.php#L756

There's not an easy way to check if _customer code_ has implemented that filter, which is I think what you were trying to do with the `\has_filter()` check here.

The `ep_prepare_meta_allowed_protected_keys` filter that we're hooking into is only concerned with protected meta (keys that start with an underscore), so we're actually conflating two different concerns here (and this goes beyond this PR) - the `ep_prepare_meta_allowed_protected_keys` filter is incorrectly being used to define _all_ meta that is indexed by default, not just protected meta. The `Search::get_post_meta_allow_list()` function is later used in the `ep_prepare_meta_data()` filter to set which meta is allowed to be indexed, and internally it uses our `vip_search_post_meta_allow_list` filter, which also ends up getting called from our new `Search::parse_vip_search_post_meta_allow_list_filter()` function. There seems to be lots of circular dependencies here.

Finally, we need to be sure to de-duplicate the array so each meta key only shows up once.

So ultimately we need the code to do this: Set the default list of protected meta to the list of protected meta we want to index (taken, from Jetpack's settings), then also use that list of protected meta along with regular meta as the default values of our `vip_search_post_meta_allow_list` filter. Phew, this is complicated...we can discuss on Slack more.

pschoffer:
Great feedback, let's chat in slack in time overlap.

I will just put some response here as I can link code easily here:)

> firstly, the 2nd and 3rd conditions will never be reached:

Note that the method you refer to here has 2 different use, one with **protected** meta where the second parameter is always provided and therefore the 2 and 3 conditions will never be reached as you correctly noted. That is by design as the protected filter already has some posts possibly set so we don't want to overwrite those with ours/jetpack defailts. Or do we?

However the second call is from [get_post_meta_allow_list](https://github.com/Automattic/vip-go-mu-plugins/blob/01848bb8f62750612783f3b609aac1c89551aacb/search/includes/classes/class-search.php#L1557) a.k.a **non-protected**. The second parameter is not provided there so it defaults to null, which is not array and therefore we want to set sensible defaults where the 2nd and 3rd condition comes into play.

> In the [ticket](https://vipjira.atlassian.net/browse/CANTINA-70) you said: `Maybe if the site has a filter registered for jetpack_sync_post_meta_whitelist, then we default to those, else default to nothing?` 

Isn't `has_filter` what you meant here? Or do we want to change this and fallback to JP like defaults even if not registered?

> Set the default list of protected meta to the list of protected meta we want to index.

Do we wanna give users the ability to differentiate between allowing protected vs regular meta? From UX perspective I really like the way we have done that so that there is only one filter `vip_search_post_meta_allow_list`. Then I as a user need to implement just that in order to add any meta (no matter if it is protected or not). On the other hand that might beat the purpose of protected meta in the first place, or? I guess users can always use the EP underlying filters but that is probably not what we want.

pschoffer:
Fixed up and retested based on @nickdaugherty chat:)

search/includes/classes/class-search.php
```diff
@@ -396,6 +441,8 @@ protected function setup_hooks() {
 		if ( false !== $this->is_protected_content_enabled() ) {
 			add_filter( 'ep_indexable_post_types', array( $this, 'add_attachment_to_ep_indexable_post_types' ), 9999 );
 		}
+
+		add_filter( 'vip_search_post_meta_allow_list', array( $this, 'vip_search_post_meta_allow_list_defaults' ) );
```
nickdaugherty, line 445: Do we want to keep with our convention of prefixing these callbacks with `filter__`?

search/includes/classes/class-search.php
```diff
@@ -1451,7 +1498,29 @@ public function filter__ep_skip_post_meta_sync( $skip_sync, $post, $meta_id, $me
...
 	}
 
-	public function get_post_meta_allow_list( $post ) {
+	/**
+	 * Adds default values for vip_search_post_meta_allow_list by combining vip-search defaults with jetpack values (to ease migration).
+	 *
+	 * @param {array} $keys Input post meta keys.
+	 * @return {array} Post meta keys enhanced by default values.
+	 */
+	public function filter__vip_search_post_meta_allow_list_defaults( $keys ) {
```
nickdaugherty, line 1507: @pschoffer I simplified the logic of this function a bit in 7fcde723f

search/includes/classes/class-search.php
nickdaugherty: Changed in 7f719ce

nickdaugherty:
@pschoffer I think this is pretty close - I made some small changes today (if you could look them over please).

How do you think we should handle the rollout? This will change the meta indexed on all sites (which will at least cause the indexes to be inconsistent, content wise). We'll also need to do some testing on existing sites to ensure any meta they have already configured stays indexed after these changes.

search/includes/classes/class-search.php
pschoffer: Great stuff thanks @nickdaugherty !

brandon-m-skinner:
Some internal discussions on this one. 

TL;DR:

There are a fixed amount of migration paths with different default needs. To keep the field counts low and be more responsible with our resources, we'll have defaults based on migration path. That has several benefits including being able to remove the migration code from the platform in the future in a clean way.

nickdaugherty:
I implemented the above-mentioned change, the addition of `VIP_SEARCH_MIGRATION_SOURCE` and the conditional default meta list, in 144a2583d (cc @pschoffer + @brandon-m-skinner).

nickdaugherty (approved):

pschoffer:
Looks awesome thanks @nickdaugherty , I plan to test on BaT a bit and deploy to staging today.

pschoffer:
I tested this and it works great... deploying to stage.

PR #1954, opened 2021-01-22
Include the original ElasticSearch query and backtrace when logging errors

netsuso:
## Description

When an ElasticSearch query fails or it returns a warning, Include the full original query in the message sent to logstash.

### Improve logging for ElasticSearch failed queries

When a VIP Search query fails, it gets logged to Logstash so we can investigate the problems. Now we are also logging the full query sent to ElasticSearch and the PHP backtrace, so we can have much more detailed information about the root cause.

## Checklist

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -595,7 +595,7 @@ class Search {
 		$response_code = (int) wp_remote_retrieve_response_code( $response );
 
 		if ( is_wp_error( $response ) || $response_code >= 400 ) {
-			$this->ep_handle_failed_request( $response, $statsd_prefix );
+			$this->ep_handle_failed_request( $response, $query, $statsd_prefix );
 		} else {
 			// Record engine time (have to parse JSON to get it)
 			$response_body_json = wp_remote_retrieve_body( $response );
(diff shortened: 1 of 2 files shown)


netsuso:
First I had only added the query, but now I've also added the PHP backtrace. I'm not sure if we will obtain a lot of useful info from there, but it's worth trying

rinatkhaziev:
@netsuso since this changes the signature of `ep_handle_failed_request`, are we absolutely sure we updated the calls accordingly everywhere?

netsuso:
> @netsuso since this changes the signature of `ep_handle_failed_request`, are we absolutely sure we updated the calls accordingly everywhere?

This is actually a helper method, the code was initially inside `filter__ep_do_intercept_request`, but it was extracted to avoid having a big method.

It should actually be private, I think it's public so it can be tested :man_shrugging: 

But anyway, I've just checked the full `mu-plugins` code and there aren't any more references

rinatkhaziev (approved):
:shipit:

brandon-m-skinner:
r1350-stacks (staging)

netsuso:

PR #1957, opened 2021-01-26
adds cron job to detect inactive index-versions

pschoffer:
### Plugin Updated: Search

We added inactive index detection. This will detect all the inactive index versions if there was no change in active/non-active index versions for more than a month. This is the first step towards having an automated cleanup of these non-used index versions.

## Checklist

## Steps to Test
I couldn't find a nice way to test the cron task so what I end up doing is:

1. Switch from cron to running the code always by adding `do_action( self::CRON_EVENT_NAME );` in `VersioningCleanupJob::init`.
2. Then you will probably fake some data. You can use this to setup one inactive version`lando wp eval "update_option( 'vip_search_index_versions', [ 'post' => [ 1 => ['number' => 1, 'created_time' => 1610662421 ], 2 => ['number' => 2, 'created_time' => 1610662421, 'activated_time' => 1610662421, 'active' => true ] ]] );"`
3. Finally you will probably not see the logs, so adding something like `var_dump($message);` in `VersioningCleanupJob::send_notification` will make you see what's up.


search/includes/classes/class-versioningcleanupjob.php
```diff
@@ -0,0 +1,146 @@
...
+			return [];
+		}
+
+		$inactive_versions = [];
+		foreach ( $versions as $version ) {
+			if ( $version['active'] ?? false ) {
+				continue;
+			}
+
+			if ( ( $version['created_time'] ?? time() ) > $week_ago_breakpoint ) {
```
pschoffer, line 107: I added this compared to the ticket requirement. This covers cases when we would have active index for a long time but would like to prepare another one to switch to. Without this, we would lose this new version the first time the cron would run.

search/includes/classes/class-versioningcleanupjob.php
```diff
@@ -0,0 +1,138 @@
...
+	/**
+	 * The name of the scheduled cron event to run the versioning cleanup.
+	 */
+	const CRON_EVENT_NAME = 'vip_search_versioning_cleanup';
+
+	const WEEK_IN_SECONDS = 7 * 24 * 60 * 60;
+
+	const SEARCH_ALERT_SLACK_CHAT = '#vip-go-es-alerts';
+
+	public function __construct( $indexables, $versioning ) {
```
pschoffer, line 16: Hmm I am not sure how to pass the dependencies correctly. But doing this felt the most correct. However, I haven't seen this pattern anywhere in our code so there is possibly a better way?

pschoffer:
Docs PR: https://github.com/Automattic/vip-docs/pull/93

mjangda:
Thoughts on bundling this as part of the Healthcheck job rather than adding a separate one?

pschoffer:
@mjangda I guess we could reuse small parts of the code as a benefit of having it in one file, but this way it seems more clean and self-contained to me. Also, test-class is much more focused. We would still probably want two independend schedules, right? Is there any other benefit you had in mind?

Also worth mentioning that we have a total of 4 cron jobs in search now, each in its own file (which I like as per above:)). So it might break code consistency if we would group two of those.

mjangda:
In that case, keeping it separate makes sense 👍

search/includes/classes/class-versioningcleanupjob.php
```diff
@@ -0,0 +1,138 @@
...
+
+	/**
+	 * Initialize the job class.
+	 *
+	 * @access  public
+	 */
+	public function init() {
+		add_action( self::CRON_EVENT_NAME, [ $this, 'versioning_cleanup' ] );
+
+		$this->schedule_job();
```
WPprodigy, line 29: I believe it would be best to attempt/check to schedule the job on `admin_init` instead. Less frontend load for option checking / memcache fetches.

brandon-m-skinner:
Adjusted the hooks, fixed a failing test, and made a minor consistency change(moved the versioning cleanup job to a class property)

search/includes/classes/class-versioningcleanupjob.php
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+
+	/**
+	 * Retrieve inactive versions to be deleted
+	 *
+	 * @param \ElasticPress\Indexable $indexable The Indexable for which to retrieve index versions
+	 * @return array Array of inactive index versions
+	 */
+	public function get_inactive_versions( \ElasticPress\Indexable $indexable ) {
```
nickdaugherty, line 68: Minor - the naming of this function could be more direct - on a first read, this makes me think it will just return all inactive versions. A better name would reflect that it's all inactive versions that are up for deletion, like `get_stale_inactive_versions()` or something.

search/includes/classes/class-versioningcleanupjob.php
```diff
@@ -0,0 +1,138 @@
...
+		$versions = $this->versioning->get_versions( $indexable );
+
+		if ( ! $versions && ! is_array( $versions ) ) {
+			return [];
+		}
+
+		$active_version = null;
```
nickdaugherty, line 75: The `Version` class already has `get_active_version()` - can that be used instead of the loop below?

search/includes/classes/class-versioningcleanupjob.php
nickdaugherty: Do we also want to only delete inactive versions that came _before_ the current active version? We could run into situations where version 2 is active, then version 3 is being indexed or migrated too, but that process takes longer than 1 week, at which point the new index is removed even though we still may want to move to it.

If we only consider indexes that have a index number _less than_ the current active index, we can avoid incorrectly deleting indexes that may still be switched to. Thoughts?

search/includes/classes/class-versioningcleanupjob.php
```diff
@@ -0,0 +1,138 @@
...
+				$active_version = $version;
+				break;
+			}
+		}
+
+		if ( ! $active_version ) {
+			return [];
+		}
+
+		$week_ago_breakpoint            = time() - self::WEEK_IN_SECONDS;
```
nickdaugherty, line 87: We should add a comment to this to remind ourselves why we have this check on the activation time of the active version. It's because we want to keep the old version for a period of time, even if it's older than 1 week, while the new version proves stable.

nickdaugherty:
@pschoffer I forsee some interactions with the "option rebuilding" code that we have. If the versions option was rebuilt by our automated process, it won't have the created_time and activated_time for each version. If I'm reading the code correctly, this would cause all inactive versions to be deleted.

This would be undesirable b/c the option rebuilding code may not have correctly picked the right version to make active - it just does the lowest number. So if you have versions 2 and 3, with 3 being active, and the option gets rebuilt, 2 will be considered active then 3 would be deleted, even though that was the actual version number in use.

We'll need to account for that somehow - we could start with just not deleting anything if there is no created_time recorded, but that leaves us open to many stale indexes. If the active index doesn't have an activated time recorded, maybe we don't delete anything at all? Then if the site switches to a new version later, it'll get a new activation time, then the regular logic can kick in. That's just an idea, tricky problem...

pschoffer:
@nickdaugherty the current code fallbacks to now. Which is later than `week_ago` meaning the one without created time will not be deleted.
```
if ( ( $version['created_time'] ?? time() ) > $week_ago_breakpoint ) {
    // If the version was created within last week OR doen't have created_time defined it is not inactive
    continue;
}
```
I added a note in the code and test case to cover that.

Long term, maybe we could change the self-healing logic to init created_date to time(), when recovering? Meaning they would be deleted a week later.

search/includes/classes/class-versioningcleanupjob.php
pschoffer: done

search/includes/classes/class-versioningcleanupjob.php
pschoffer: done

search/includes/classes/class-versioningcleanupjob.php
pschoffer: done

search/includes/classes/class-versioningcleanupjob.php
nickdaugherty: Good call, this has been switched to `admin_init` or `wp_loaded` (for CLI).

search/includes/classes/class-versioningcleanupjob.php
nickdaugherty: We decided we won't distinguish between older and newer indexes, we'll just bump the threshold to 1 month instead of 1 week to give a bigger buffer.

nickdaugherty (approved):
Looks good @pschoffer. I switched the time-frame to 1 month as discussed.

PR #1958, opened 2021-01-26
Fix a bug with use_block_editor_for_post filter in WP Core

gudmdharalds:
## Description

WordPress Core has a [bug](https://core.trac.wordpress.org/ticket/52363) in the `use_block_editor_for_post` filter, for which a patch is available. The aim of this Pull-Request is to temporarily fix this bug on the VIP Go Platform until it is resolved in WordPress Core. 

When the bug in Core is fixed, this Pull-Request can be reverted and the temporary patch removed. Currently, the WordPress Core bugfix is aimed at WordPress 5.7 release.

This bug was discovered while working on [deprecating Gutenberg Ramp](https://lobby.vip.wordpress.com/2020/11/23/removing-gutenberg-ramp/).

### Bug description

The bug in the `use_block_editor_for_post` filter can lead to Gutenberg Editor being loaded for posts of types that do _not_ support the editor, resulting in a JavaScript error and a white screen. This will happen when using the filter in the following way:

> add_filter( 'use_block_editor_for_post', '__return_true' );

The is because the current implementation of the `use_block_editor_for_post()` function in [WordPress 5.6](https://core.trac.wordpress.org/browser/tags/5.6/src/wp-admin/includes/post.php?rev=49777#L2109) (and [trunk](https://core.trac.wordpress.org/browser/trunk/src/wp-admin/includes/post.php#L2109)) does not check if the post-type in question supports editor. Attempting to edit posts of types that do not support editor, combined with a filter usage like the above, will cause Gutenberg Editor to being loaded for a particular post leading to a white screen. This will for example happen with the `attachment` post type that is built-into WordPress. Note that the `use_block_editor_for_post_type` filter does _not_ have this problem, as it does a check to make sure the post-type supports an editor, and thus it can be used in the same manner with success. 

### Resolving the bug

Both the WordPress Core patch and this Pull-Request will add a check to avoid an editor being loaded for post of type that does not support it, so that the problem will not occur with the `use_block_editor_for_post` filter.

This Pull-Request is not identical to the WordPress Core patch, but it will achieve the same result: The WordPress Core patch will solve this problem within the `use_block_editor_for_post()` function by adding a simple `if` check – this function calls the `use_block_editor_for_post` filter after this check succeeds -- while this Pull-Request will add a filter-callback that adds the check and forces a `false` return result when the check fails. 

The bugfix applied by this Pull-Request and the WordPress Core patch can co-exist on the platform. 

### Steps to replicate the whitescreen

To replicate:

* Set up WordPress 5.6 with no plugins active and default theme activated. 

* Add `add_filter( 'use_block_editor_for_post', '__return_true' );` to a file in `client-mu-plugins/`. 

* Edit any post item of type `post`, verify that Gutenberg loads successfully. 

* Add image to Media Library, note the ID of it. 

* Attempt to edit the newly uploaded image by navigating to: `wp-admin/post.php?post=ID&action=edit` -- replace `ID` with ID of the image. 

* White screen should appear, with JavaScript error: `Uncaught (in promise) TypeError: Cannot read property 'raw' of undefined [...]` 

Expected behavior: Gutenberg is not loaded for post-types not supporting editor, even if the `use_block_editor_for_post` filter is used as described above.

Observed behavior: White screen, unable to do editing.

### Bugfix for `use_block_editor_for_post` WordPress Core filter 

The bugfix will add callback for the `use_block_editor_for_post` filter at a higher level than most usage on the VIP Go Platform, so that if a post of a type that does not support editor is being loaded the Gutenberg Editor will not load.

The filter looks like this:

```
add_filter(
       'use_block_editor_for_post',
        function( $can_edit, $post ) {
                if ( ! isset( $post->post_type ) ) {
                        return $can_edit;
                }

                if ( ! post_type_supports( $post->post_type, 'editor' ) ) {
                        return false;
                }
 
                return $can_edit;
        },
        20,
        2
);
```

Note that priority `20` is set as most usage is at `10` (default), and a number of recently added PRs is at `15`. We do not want to go too high nor too low.

## Checklist

Note that in the checklist below, Gutenberg Ramp has been added, because this is a plugin that influences the end-result of the `use_block_editor_for_post` filter, and we have it enabled on VIP Go. We are currently deprecating it. Some customers however have it disabled and some enabled. Items are added here so that various states of this plugins are tested; disabled, enabled and not configured, and enabled and configured.

These checks have been covered:

- [x] This change works and has been tested locally without Core patch applied.
- [x] This change has been tested locally with the Core patch applied as well.

- With Gutenberg Ramp disabled:
  - [x] This change works and has been tested on a Go sandbox without Core patch applied.
  - [x] This change has been tested on a Go sandbox with the Core patch applied as well.

- With Gutenberg Ramp plugin enabled but not configured:
  - [x] This change works and has been tested on a Go sandbox without Core patch applied.
  - [x] This change has been tested on a Go sandbox with the Core patch applied as well.

- With Gutenberg Ramp plugin enabled and configured (set to load):
  - [x] This change works and has been tested on a Go sandbox without Core patch applied.
  - [x] This change has been tested on a Go sandbox with the Core patch applied as well.

## Steps to Test

1. Add `add_filter( 'use_block_editor_for_post', '__return_true' );` to a file in `client-mu-plugins`.
2. Edit any post item of type `post`, verify that Gutenberg loads successfully. 
3. Add image to Media Library, note the ID of it. 
4. Attempt to edit the newly uploaded image by navigating to: `wp-admin/post.php?post=ID&action=edit` -- replace `ID` with ID of the image. 
5. Gutenberg should not load, and editing of the image should work as normal.


rinatkhaziev (approved):
Thank you for the PR @gudmdharalds, looks good!

gudmdharalds:
Hi @Automattic/vip-platform-cantina,

Can this Pull-Request be merged this week? We will be merging [Gutenberg Ramp related Pull-Requests next week](https://lobby.vip.wordpress.com/2021/01/28/updated-timeline-removing-gutenberg-ramp/) and for that to be successful this will need to be merged.

Thank you!

rinatkhaziev:
r1383-stacks

PR #1961, opened 2021-01-27
adds tag support

pschoffer:
## Description

I first tried to have Github label => changelog tag mapping in the code, which works fine. But then I thought it is a way more flexible if we can maintain that in Github directly. So the automation is looking for `ChangelogID: <ID>` in the label description that we don't really use for anything else.

You can change it here:
https://github.com/Automattic/vip-go-mu-plugins/labels?page=2&sort=name-asc

I also added `[Tag] Jetpack` and `[Tag] Search` as examples.

### Changelog automation enhancement - tags

We updated changelog automation to support tags. Some Github labels will automatically be converted to the changelog tags.

## Checklist

## Steps to Test

CircleCI thingie so no real testing.

### DIFF
--- a/bin/github-changelog.php
+++ b/bin/github-changelog.php
@@ -90,7 +90,7 @@ function get_changelog_tags( $github_labels ) {
     ];
 
     foreach ( $github_labels as $label ) {
-        preg_match('/ChangelogID:\s*(\d+)/', $label['description'], $matches);
+        preg_match('/ChangelogTagID:\s*(\d+)/', $label['description'], $matches);
         if ( $matches ) {
             array_push( $tags, $matches[1] );
         }
--- a/bin/github-changelog.php
+++ b/bin/github-changelog.php
@@ -65,13 +65,13 @@ function parse_changelog_html( $changelog_html ) {
     ];
 }
 
-function create_draft_changelog( $title, $content ) {
+function create_draft_changelog( $title, $content, $tags ) {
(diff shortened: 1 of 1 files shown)


bin/github-changelog.php
```diff
@@ -84,13 +84,29 @@ function create_draft_changelog( $title, $content ) {
     echo $response;
 }
 
+function get_changelog_tags( $github_labels ) {
+    $tags = [
+        '1784989' // mu-plugins
+    ];
+
+    foreach ( $github_labels as $label ) {
+        preg_match('/ChangelogID:\s*(\d+)/', $label['description'], $matches);
```
brandon-m-skinner, line 93: ChangelogID is a bit confusing since it may imply the Changelog post ID. I think ChangelogTagID is more descriptive.
```suggestion
        preg_match('/ChangelogTagID:\s*(\d+)/', $label['description'], $matches);
```

brandon-m-skinner (changes requested):
Looks awesome. 

Only thing is that `ChangelogID` is ambiguous.

Will need to update the existing labels for that as well, but pretty minor since there's only two right now 😄.

What do you think?

pschoffer:
@brandon-m-skinner Sounds good. Done.

nickdaugherty (approved):

brandon-m-skinner (approved):

PR #1963, opened 2021-01-28
Properly handle file paths with `sites/:id`

mjangda:
Those need to be stripped out as well since the files paths are stored in postmeta without it.

### Private Files: Handle `/sites/` paths for Private Files

We've added a fix that properly handles files with a subsite path (e.g. `/wp-content/uploads/sites/4/2021/01/unpublished.jpg`) with our Private Files feature. We now strip the `sites/:id` portion of the path before forwarding the request to WordPress.

## Checklist

## Steps to Test

See unit tests.

### DIFF
--- a/files/acl/restrict-unpublished-files.php
+++ b/files/acl/restrict-unpublished-files.php
@@ -14,7 +14,7 @@ const CACHE_GROUP = 'vip-files-acl';
 
 function check_file_visibility( $file_visibility, $file_path ) {
 	if ( 0 === strpos( $file_path, 'sites/' ) ) {
-		$file_path = preg_replace( '#^sites\/\d+\/#', '', $file_path );
+		$file_path = preg_replace( '#^sites/\d+/#', '', $file_path );
 	}
 
 	// Reverse lookup for the attachment ID
--- a/files/acl/restrict-unpublished-files.php
+++ b/files/acl/restrict-unpublished-files.php
@@ -12,7 +12,16 @@ use const Automattic\VIP\Files\Acl\FILE_IS_PRIVATE_AND_ALLOWED;
 
 const CACHE_GROUP = 'vip-files-acl';
 
+
+/**
+ * Given a path determine whether the file is private or public
+ *
+ * @param string $file_visibility one of Automattic\VIP\Files\Acl\(FILE_IS_PUBLIC | FILE_IS_PRIVATE_AND_ALLOWED | FILE_IS_PRIVATE_AND_DENIED).
+ * @param string $file_path path to file to be checked for visibility.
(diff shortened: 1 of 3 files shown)


files/acl/pre-wp-utils.php
```diff
@@ -81,6 +81,10 @@ function validate_path( $file_path ) {
 function sanitize_and_split_path( $file_path ) {
 	list( $pre_wpcontent_path, $post_wpcontent_path ) = explode( '/wp-content/uploads/', $file_path, 2 );
 
+	if ( 0 === strpos( $post_wpcontent_path, 'sites/' ) ) {
+		$post_wpcontent_path = preg_replace( '#^sites\/\d+\/#', '', $post_wpcontent_path );
```
rinatkhaziev, line 85: Slash escaping is redundant here, because the delimiters are `#`.
```suggestion
		$post_wpcontent_path = preg_replace( '#^sites/\d+/#', '', $post_wpcontent_path );
```

rinatkhaziev (changes requested):
Thank you! There's a minor readability suggestion, otherwise looks good!

files/acl/pre-wp-utils.php
mjangda: Good call!

brandon-m-skinner (dismissed):

mjangda:
This needs to be re-thought because of changes in https://github.com/Automattic/vip-go-mu-plugins/pull/1975

We can probably handle the stripping to unpublished-files instead since that's the only thing impacted right now.

mjangda:
Re-thought and should be all good now.

files/acl/restrict-unpublished-files.php
```diff
@@ -13,6 +13,10 @@
 const CACHE_GROUP = 'vip-files-acl';
 
 function check_file_visibility( $file_visibility, $file_path ) {
+	if ( 0 === strpos( $file_path, 'sites/' ) ) {
+		$file_path = preg_replace( '#^sites\/\d+\/#', '', $file_path );
```
rinatkhaziev, line 17: Redundant forward-slash escaping. Additionally, I think some comments are in order as well as PHPDoc for the function and type hints in the function signature.

rinatkhaziev (changes requested):
@mjangda there's a couple of notes, mostly chores/clerical stuff, but I feel like it's better to address this right away for the sake of our future selves.

files/acl/restrict-unpublished-files.php
rinatkhaziev:
```suggestion
		$file_path = preg_replace( '#^sites/\d+/#', '', $file_path );
```

rinatkhaziev (approved):

rinatkhaziev:
r1379-stacks

PR #1965, opened 2021-01-29
Allow skipping loading VaultPress

netsuso:
## Description

VaultPress is currently always loaded, even if it's not used. With this PR, setting the define `VIP_VAULTPRESS_SKIP_LOAD` to true will avoid loading it entirely.

### Add the option to skip loading VaultPress

There's now the option to skip loading VaultPress code entirely, which can be useful if it's not being used.

Some considerations need to be taken to disable VaultPress if it has been used before. So in case you are interested, please open a support ticket to get more information.

## Checklist

## Steps to Test

1. Check out PR.
1. Ensure that VaultPress is still available by default in the Jetpack dashboard in `wp-admin`
1. Set the define `VIP_VAULTPRESS_SKIP_LOAD` to `true` and check that VaultPress is no longer available in the Jetpack dashboard.


brandon-m-skinner (approved):

GaryJones:
Can we please ensure this new constant makes it into the public docs as a How To?

netsuso:
> Can we please ensure this new constant makes it into the public docs as a How To?

For the moment this is going to be managed by support, as disabling the VaultPress plugin on a site that has used it before may lead to confusing results (in the form of alert emails notifying the owner of the site that certain features have been disabled). @parkcityj was looking into this.

netsuso:
stacks-r1360

PR #1970, opened 2021-02-03
Add VIP NotOptions Mitigation

trepmal:
Detect invalid values in the notoptions cache.

## Description

The notoptions cache should only ever include keys whose value is `true` (bool).
In cases where that value is corrupted (these circumstances not addressed here, this is solely mitigation), flush cache and alert the VIP team.

- only run in admin or CLI context
- only alert in production environments

Historical VIP discussion:

- https://wp.me/p9GBOq-2bh
- https://wp.me/p9GBOq-27U
- 
### Options: Add code to detect invalid values in the notoptions cache (Inactive)

Adds a new file with code that automatically detects corruption of the `notoptions` cache entry, clearing the cache and alerting when it is detected. This code is not yet activated - it will be required by a follow-up PR.

## Checklist

## Steps to Test

With the PR applied, purposefully contaminate notoptions:

```
$notoptions = wp_cache_get( 'notoptions', 'options' );
$notoptions['somekey'] = 'anystring';
wp_cache_set( 'notoptions', $notoptions, 'options' );
```

*must be a `production` environment to run alerting code, else alter that condition*

Then, any CLI command will meet conditions to run the check:

```
wp option get home
```


notoptions-mitigation.php
```diff
@@ -0,0 +1,81 @@
...
+function notoptions_mitigation() {
+	$notoptions = wp_cache_get( 'notoptions', 'options' );
+
+	// filter for any values not equal to (bool)true.
+	$not_trues = array_filter( $notoptions, function( $v ) {
+		return $v !== true;
+	} );
+
+	// if they exist, something's borked
+	if ( $not_trues ) {
```
nickdaugherty, line 26: Can we more explicitly check this?
```suggestion
	if ( 1 >= count( $not_trues ) ) {
```

nickdaugherty: Actually we count it further down with `$total_invalid`, so we can just move that up here and check against it.

notoptions-mitigation.php
```diff
@@ -0,0 +1,81 @@
...
+				$total_invalid,
+				( $flushed ? '' : 'un' )
+			);
+
+			$to_irc = $subject . ' #vipnotoptions';
+
+			// Send to IRC, if we have a host configured
+			if ( defined( 'ALERT_SERVICE_ADDRESS' ) && ALERT_SERVICE_ADDRESS ) {
+				if ( 'production' === $environment ) {
+					wpcom_vip_irc( '#vip-deploy-on-call', $to_irc, $irc_alert_level, 'a8c-notoptions' );
```
nickdaugherty, line 60: The `Alert` class now has a `send_to_chat()` function that can replace this one.

notoptions-mitigation.php
```diff
@@ -0,0 +1,81 @@
...
+				( ( 0 !== $wpdb->blogid ) ? ", blog ID {$wpdb->blogid}" : ''),
+				$total_invalid,
+				( $flushed ? '' : 'un' )
+			);
+
+			$to_irc = $subject . ' #vipnotoptions';
+
+			// Send to IRC, if we have a host configured
+			if ( defined( 'ALERT_SERVICE_ADDRESS' ) && ALERT_SERVICE_ADDRESS ) {
+				if ( 'production' === $environment ) {
```
nickdaugherty, line 59: Do we want to log to Slack/IRC on all environments for better visibility into when this happens? Then only send to Opsgenie for prod.

nickdaugherty (changes requested):
This is looking great - I left some initial thoughts on here. We may also be able to find a home for this in an existing file (vs. adding a new one).

notoptions-mitigation.php
trepmal: seems like a good idea

GaryJones:
> The notoptions cache should only ever include keys whose value is `true` (bool).

For future reader reference, this setting to `true` is defined in [`get_option()`](https://developer.wordpress.org/reference/functions/get_option/).

GaryJones (changes requested):

notoptions-mitigation.php
```diff
@@ -0,0 +1,89 @@
...
+		$environment   = ( ( defined( 'VIP_GO_APP_ENVIRONMENT' ) && VIP_GO_APP_ENVIRONMENT ) ? VIP_GO_APP_ENVIRONMENT : 'unknown' );
+		$site_id       = defined( 'VIP_GO_APP_ID' ) ? VIP_GO_APP_ID : false;
+
+		// Send notices to VIP staff if this is happening on VIP-hosted sites
+		if ( $is_vip_env && $site_id ) {
+
+			$irc_alert_level = 2; // ALERT
+			$opsgenie_alert_level = 'P4';
+
+			$subject = 'NOTOPTIONS: %s (%s VIP Go site ID: %s%s) %s invalid values found in notoptions. Cache flush was %ssuccessful.';
```
rebeccahum, line 49: If it's a multi-site, it might be helpful to know which subsite this occurred on.

notoptions-mitigation.php
```diff
@@ -0,0 +1,89 @@
...
+		return true !== $v;
+	} );
+
+	$total_invalid = count( $not_trues );
+
+	// if they exist, something's borked
+	if ( 1 <= $total_invalid ) {
+
+		// attempt repair
+		$flushed = wp_cache_delete( 'notoptions', 'options' );
```
rebeccahum, line 36: Not sure if this is complicating things, but would it be worthwhile to add a role reset and potentially, reconnecting Jetpack? Maybe triggered via a cron event after.

notoptions-mitigation.php
```diff
@@ -0,0 +1,89 @@
...
+		// begin prepping notification
+		$is_vip_env    = ( defined( 'WPCOM_IS_VIP_ENV' ) && true === WPCOM_IS_VIP_ENV );
+		$environment   = ( ( defined( 'VIP_GO_APP_ENVIRONMENT' ) && VIP_GO_APP_ENVIRONMENT ) ? VIP_GO_APP_ENVIRONMENT : 'unknown' );
+		$site_id       = defined( 'VIP_GO_APP_ID' ) ? VIP_GO_APP_ID : false;
+
+		// Send notices to VIP staff if this is happening on VIP-hosted sites
+		if ( $is_vip_env && $site_id ) {
+
+			$irc_alert_level = 2; // ALERT
+			$opsgenie_alert_level = 'P4';
```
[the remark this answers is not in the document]
trepmal: Needs follow-up either way, in case anything like roles needs to be reset.

notoptions-mitigation.php
```diff
@@ -0,0 +1,89 @@
...
+
+	// filter for any values not equal to (bool)true.
+	$not_trues = array_filter( $notoptions, function( $v ) {
+		return true !== $v;
+	} );
+
+	$total_invalid = count( $not_trues );
+
+	// if they exist, something's borked
+	if ( 1 <= $total_invalid ) {
```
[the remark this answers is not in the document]
trepmal: I fought a long war with the linter and nearly lost...........

notoptions-mitigation.php
trepmal: This might be a v2 when we have more information about frequency etc.

notoptions-mitigation.php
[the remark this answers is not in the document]
rebeccahum: If it needs to be actioned on after, should it be higher than P4 in the case that there's no DoC?

notoptions-mitigation.php
nickdaugherty: +1 for keeping the first iteration simple...we can expand it after.

notoptions-mitigation.php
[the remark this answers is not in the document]
GaryJones: A compromise could be to include it into the Opsgenie description (body?), if it's a production site. Length would be less of an issue there than a Slack message.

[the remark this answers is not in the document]
nickdaugherty: We can also log the exact options in a different way, using `\Automattic\VIP\Logstash\log2logstash()`

notoptions-mitigation.php
```diff
@@ -0,0 +1,89 @@
...
+				( $flushed ? '' : 'un' )
+			);
+
+			$to_irc = $subject . ' #vipnotoptions';
+
+			// Send to IRC, if we have a host configured
+			if ( defined( 'ALERT_SERVICE_ADDRESS' ) && ALERT_SERVICE_ADDRESS ) {
+
+				$alerts = Alerts::instance();
+				$alerts->send_to_chat( '#vip-deploy-on-call', $to_irc, $irc_alert_level, 'a8c-notoptions' );
```
GaryJones, line 67: Strong request to have the Slack alert to include a `#` link to a FG page that contains details and actions of what data to grab, and then how to check whether roles need resetting, how to reset them, check Jetpack connection, how to reconnect that, etc.

Some On Call folks won't have dealt with this issue much before, so having an alert point to all the actionable steps would be extremely helpful.

trepmal: 💡 will do

notoptions-mitigation.php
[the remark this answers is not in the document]
trepmal: minor refactor made this moot

vip-helpers/vip-notoptions-mitigation.php
```diff
@@ -0,0 +1,92 @@
...
+				$alerts = Alerts::instance();
+				$alerts->send_to_chat( '#vip-deploy-on-call', $to_irc, $irc_alert_level, 'a8c-notoptions' );
+
+				if ( 'production' === $environment ) {
+
+					// Send to OpsGenie
+					$alerts->opsgenie(
+						$subject,
+						array(
+							'alias'       => 'notoptions/' . $site_id,
```
nickdaugherty, line 78: Small suggestion - we can make the alias more specific, which gives us more flexibility if we need other notoptions alerts in the future.
```suggestion
							'alias'       => 'notoptions/cache-corruption/' . $site_id,
```

vip-helpers/vip-notoptions-mitigation.php
```diff
@@ -0,0 +1,92 @@
...
+							'description' => 'Invalid values found in notoptions cache',
+							'entity'      => (string) $site_id,
+							'priority'    => $opsgenie_alert_level,
+							'source'      => 'sites/notoptions-value',
```
nickdaugherty, line 82:
```suggestion
							'source'      => 'mu-plugins/notoptions-value-monitor',
```

000-vip-init.php
```diff
@@ -163,6 +163,7 @@
 require_once( __DIR__ . '/lib/utils/class-alerts.php' );
 
 // Load our development and environment helpers
+require_once( __DIR__ . '/vip-helpers/vip-notoptions-mitigation.php' );
```
nickdaugherty, line 166: It would be better to split this out to a separate change that can be deployed after the actual file is committed / deployed, to avoid any problems with ordering in the code checkout process.

nickdaugherty (approved):

brandon-m-skinner (approved):

PR #1972, opened 2021-02-04
Always filter the jetpack_https_test transient to 1

nickdaugherty:
## Description

Outbound SSL is always available on VIP so we can skip the HTTP requests that JP makes to determine this (since JP has to work on many different hosting providers).

props @emrikol for the idea + code

### Jetpack: Skip HTTPS support checking to reduce HTTP requests

Since Jetpack works across many different hosting providers (some which may not support HTTPS / SSL), it attempts to detect support by making an HTTPS request to WordPress.com, which is then cached for 1 day. On VIP, SSL is always supported so we can skip this check by filtering the transient used to store the result.

## Checklist

## Steps to Test

1. Check out PR.
2. In `wp shell`, run `Jetpack::permit_ssl()` - this should return `1`
3. `wp transient get jetpack_https_test` - this should return `1` as well

### DIFF
--- a/tests/test-vip-jetpack.php
+++ b/tests/test-vip-jetpack.php
@@ -160,4 +160,12 @@ class VIP_Go_Jetpack_Test extends WP_UnitTestCase {
 
 		$this->assertEquals( $expected_modules, $actual_modules );
 	}
+
+	public function test__jetpack_https_test__transient_filter() {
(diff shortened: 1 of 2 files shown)


brandon-m-skinner (approved):

pschoffer:
Stacks-r1371.

PR #1974, opened 2021-02-05
Private Files: Always send X-Private header

mjangda:
Even if the file is public, send a value of false to avoid traps when we hit an error

### Private Files: Always send X-Private header

To harden the feature and avoid unexpected issues when the ACL endpoint may hit an error (e.g. fatal), we now will always send the X-Private header when the endpoint successfully processes the file. This way, we can differentiate between successful and unsuccessful requests and handle them appropriately upstream.

## Checklist

## Steps to Test

Covered by unit tests.

### DIFF
--- a/files/acl/acl.php
+++ b/files/acl/acl.php
@@ -71,19 +71,22 @@ function send_visibility_headers( $file_visibility, $file_path ) {
 	// Default to throwing an error so we can catch unexpected problems more easily.
 	$status_code = 500;
 	$header = false;
+	$is_private = null;
 
 	switch ( $file_visibility ) {
 		case FILE_IS_PUBLIC:
 			$status_code = 202;
+			$is_private = false;
 			break;
 
 		case FILE_IS_PRIVATE_AND_ALLOWED:
 			$status_code = 202;
-			$header = 'X-Private: true';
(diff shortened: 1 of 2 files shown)


files/acl/acl.php
```diff
@@ -71,19 +71,22 @@ function send_visibility_headers( $file_visibility, $file_path ) {
...
 		case FILE_IS_PRIVATE_AND_ALLOWED:
 			$status_code = 202;
-			$header = 'X-Private: true';
+			$is_private = true;
 			break;
 
 		case FILE_IS_PRIVATE_AND_DENIED:
 			$status_code = 403;
+			$is_private = true;
 			$header = 'X-Private: true';
```
alexcriss, line 90: I think we can probably get rid of this too :)

alexcriss (changes requested):
Thanks a lot for this, I left one minor silly comment!

files/acl/acl.php
mjangda: Of course! Good catch!

nickdaugherty (approved):
LGTM

rinatkhaziev:
r1380-stacks

PR #1978, opened 2021-02-09
Fix the potential race condition with notoptions

pschoffer:
## Description

There is a race condition described below. 

I thought about performance impact and I think it should be minimal. As we would remove the key from the notoptions [few lines later ](https://github.com/WordPress/WordPress/blob/d28712518ccfde268ab49f4d8dd832e935229610/wp-includes/option.php#L598) anyways. The only change in behavior could be in case that the insert would fail for other than expected reason. In that case we would populate notoptions again next time we would try to get that option. So minimal impact in my opinion, but please challange me on that!

### Update: Handle potential race conditions around notoptions

There is a possibility in the core that if a `notoptions` are being updated at the same time, a change could be rewritten.
This would then result in option being stored in DB but also marked as non-existing in `notoptions` making further attempts to add such option always fail without removing the options from the list of non-existing options.

This change removes the option from notoptions before the DB operation rather than after the DB operation changed at least 1 line in the DB. Currently this is only enabled on non-production environments while it is evaluated.

## Checklist

## Steps to Test

As this is race condition it is really hard to reproduce without changing the core code. But we can try the regular flow isnt affected.
1. Try to get non-existing option - `lando wp eval "var_dump(get_option( 'foo' ));"` (should return false)
2. Verify the `foo` option is in notoptions now - `lando wp eval "var_dump(wp_cache_get( 'notoptions', 'options' ));"` 
3. Update foo option. - lando wp eval "var_dump(update_option( 'foo', 'bar' ));" - This will clear the option from notoptions in most cases, but sometimes during the race condition that would fail
4. Get option to verify it was persisted - `lando wp eval "var_dump(get_option( 'foo' ));"`


nickdaugherty:
We'll want to handle option updates as well (just like we do with the alloptions code that this is based on). Do we want to hook into `added_option` instead of `add_option`? That's what we do for alloptions.

Anyway, we need to thoroughly evaluate and test this as has a big reach.

pschoffer:
Well in this scenario, the issue is that the option seems to not exist (because it is in notoptions), this means even calling `update_option` will call `add_option` so it will end up here anyways and it seems to be better to limit the impact.

Also in this case the `added_option` is not triggered so it would not help. It actually is triggered even after the notoptions would be updated so there would be no benefit I think.

nickdaugherty (approved):

nickdaugherty:
r1402-stacks

PR #1980, opened 2021-02-10
Private Files: PURGE attachment URLs on post update

mjangda:
When a post is updated, we should purge all children attachments. This allows us to propagate changes more quickly to the edges, and also allows us to cache public files for much longer which has performance benefits.

### Private Files: Purge cache for attachments

When the Restrict Unpublished Files module is enabled, a post update will now automatically purge the URLs for any child attachments for that post. This allows us to quickly propagate visibility changes and also cache public files for much longer, which has several performance benefits.

## Checklist

## Steps to Test

1. Check out PR.
1. Add constant: `define( 'VIP_FILES_ACL_ENABLED', true );
1. `wp option set vip_files_acl_restrict_unpublished_enabled 1`
1. Create a post with attachments. Any updates to the post should add attachments to the purge list.

### DIFF
--- a/files/acl/restrict-unpublished-files.php
+++ b/files/acl/restrict-unpublished-files.php
@@ -12,7 +12,6 @@ use const Automattic\VIP\Files\Acl\FILE_IS_PRIVATE_AND_ALLOWED;
 
 const CACHE_GROUP = 'vip-files-acl';
 
-
 /**
  * Given a path determine whether the file is private or public
  *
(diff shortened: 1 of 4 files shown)


files/acl/restrict-unpublished-files.php
```diff
@@ -105,3 +104,43 @@ function get_attachment_id_from_file_path( $path ) {
...
+
+	$attachment_ids = get_posts( [
+		'post_parent'    => $post->ID,
+		'post_type'      => 'attachment',
+		'posts_per_page' => 100, // Set a reasonably high limit (instead of -1 as default)
+		'orderby' => 'ID', // For performance (instead of `date` as default)
+		'order' => 'ASC',
+		'fields' => 'ids',
+		'suppress_filters' => false,
+	] );
```
alexcriss, line 135: This might be a silly comment, what happens if a post has more than 100 attachments, should we make sure to get all of them, possibly in batches?

mjangda: Yes, we can do that. It's a not very likely scenario but I was just being a lazy engineer :)

mjangda: Actually, since we're just fetching IDs, we can probably bump the number much higher. I would say 250 is probably the max. If you have more than that, you have bigger problems :)

alexcriss: Thanks for the reply :)

I know it's very unlikely, but since this will leave files marked as public for very long when a post is updated if there are more than 100 I felt compelled to reply ;)

rinatkhaziev: I wouldn't say it's "very unlikely", a large gallery can easily be over 100 images.

mjangda: Updated to 250.

files/acl/restrict-unpublished-files.php
```diff
@@ -105,3 +104,43 @@ function get_attachment_id_from_file_path( $path ) {
...
+	}
+
+	$attachment_ids = get_posts( [
+		'post_parent'    => $post->ID,
+		'post_type'      => 'attachment',
+		'posts_per_page' => 250, // Set a reasonably high limit (instead of -1 as default)
+		'orderby' => 'ID', // For performance (instead of `date` as default)
+		'order' => 'ASC',
+		'fields' => 'ids',
+		'suppress_filters' => false,
```
rinatkhaziev, line 134: @mjangda I'm kind of conflicted on this one being set to false because we'll be at risk of userland code interfering with our logic here. I think it's ok to potentially sacrifice a tiny bit of performance (I'm assuming this is set to false for APC to be able to cache the query, but, realistically, this particular query would only run in this given context that also won't happen on the frontend (🤞 )
```suggestion
		'suppress_filters' => true,
```

mjangda: Yep, that seems reasonable.

rinatkhaziev (approved):
:shipit:

netsuso:
stacks-r1386

PR #1982, opened 2021-02-11
Add scheduled user-role backup and management CLI

trepmal:
## Description

Adds 1) a cron job to regularly backup user roles to a secondary option and 2) a CLI to manage listing/viewing/restoring those backups to the primary.

Why? In some cases of option corruption, user roles are lost and restoration is complicated and/or slow. This intends to make it quick and easy.

Note: this deals with the `wp_user_roles` option, not the assignments to users.

### Add User Role Backups

[tbd]

## Checklist

## Steps to Test

Begin by backing up current roles, then alter roles and backup again:
```
wp cron event run 'Automattic\VIP-backup_roles'
wp role delete editor
wp role delete author
wp role delete contributor
wp cron event run 'Automattic\VIP-backup_roles'
```

A couple options will now be listed in `wp vip role-backup list`

1. optionally view a backup: `wp vip role-backup view [<time_key>]`
2. and restore: `wp vip role-backup restore [<time_key>]`
3. and confirm: `wp role list`


WPprodigy:
Tested pretty extensively, all went well.

vip-helpers/class-vip-backup-user-role-cli.php
```diff
@@ -0,0 +1,148 @@
...
+		$backup_roles = get_option( 'vip_backup_user_roles', [] );
+		$backup_roles = array_reverse( $backup_roles, true );
+
+		$data = [];
+
+		foreach ( $backup_roles as $time => $roles ) {
+			$data[] = [
+				'time_key'  => $time,
+				'time_h'    => gmdate( 'Y-m-d H:i:s', $time ),
+				'num_roles' => count( $roles ),
```
nickdaugherty, line 39: Would it be useful to show the role names here as well as a CSV string?

nickdaugherty: I realize there's a `view` command as well, but I wonder if it would be more efficient if you can tell at a glance what each backup contains (rather than running `view` for each individually).

vip-helpers/class-vip-backup-user-role-cli.php
```diff
@@ -0,0 +1,148 @@
...
+
+	/**
+	 * Backup Current Roles Now
+	 *
+	 * ## OPTIONS
+	 *
+	 */
+	public function now( $args, $assoc_args ) {
+
+		\WP_CLI::confirm( 'This will remove the oldest backup. Ok?' );
```
nickdaugherty, line 138: This isn't _strictly_ true if the backup option hasn't yet filled up, but should be fine in practice.

vip-helpers/class-vip-backup-user-role-cli.php
```diff
@@ -0,0 +1,148 @@
...
+		if ( 'latest' === $key ) {
+			$backup = array_pop( $backup_roles );
+		} else {
+			if ( isset( $backup_roles[ $key ] ) ) {
+				$backup = $backup_roles[ $key ];
+			} else {
+				\WP_CLI::error( 'Specified backup time_key not found.' );
+			}
+		}
+
```
nickdaugherty, line 123: We probably want a confirmation step here - we could show the current roles count vs the backup's roles count along with the date.

A message like `This will overwrite the current $current_count roles with the $backup_count roles from the backup taken at $backup_time` would help confirm that you have the right one.

vip-helpers/class-vip-backup-user-role-cli.php
trepmal: simple enough, could add value 👍

vip-helpers/class-vip-backup-user-role-cli.php
trepmal: Probably worth clarifying _just in case_. I think adding a `USER_ROLE_BACKUP_LENGTH` constant with the designated history (currently `3`) would be good to ensure that the reference here always matches the backup job

vip-helpers/vip-notoptions-mitigation.php
```diff
@@ -10,9 +10,13 @@
 
 use Automattic\VIP\Utils\Alerts;
 
+define( 'USER_ROLE_BACKUP_LENGTH', 3 );
```
WPprodigy, line 13: Should we prefix this with `VIP_`? And also, just curious if there would be any harm in upping this to say 5 - since we aren't autoloading it shouldn't matter, and will better help protect against maybe a weekend problem that went unnoticed for 3 days?

nickdaugherty: @WPprodigy this file is namespaced so that'll also cover the constant.

nickdaugherty (approved):
Tested it all out and it works great!

nickdaugherty:
r1391-stacks

PR #1983, opened 2021-02-12
Search: Add index allocation settings for datacenter

nickdaugherty:
## Description

This sets the allocation settings of all VIP Search indexes so that they are allocated in the site's origin datacenter, instead of spread throughout the cluster.

NOTE - the tests will be broken until https://github.com/Automattic/ElasticPress/pull/79 is merged into mu-plugins

This needs a coordinated rollout as there is another config change at the cluster level that needs to happen.

### Update: Allocate Search indexes in site's origin datacenter

Rather than allowing index shards to be distributed around the cluster, this change ensures that indexes are placed in a site's origin datacenter to be closer to the web containers.

This will only affect new indexes at first - existing indexes will be modified at a later time.

## Checklist

## Steps to Test

1. Check out PR
2. Define the site origin dc constant in your vip-config.php - `define( 'VIP_ORIGIN_DATACENTER', 'dfw' );`
2. Recreate the indexable mappings - note that this will delete your indexes too! `wp vip-search put-mapping`
3. Check the index settings - this can be done via direct request to Elasticsearch, like `curl http://localhost:19200/vip-200508-post-1/_settings | jq`
4. The settings should now contain `index.routing.allocation.include.dc=dfw`

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -1588,6 +1588,18 @@ class Search {
 		return array_merge( $keys, $vip_search_allow_list_keys );
 	}
 
+	public function get_origin_dc_from_es_host( $host ) {
+		$dc = null;
+
+		$matches = array();
+
+		if ( preg_match( '/^es-ha[-.](.*)\.vipv2\.net$/', $host, $matches ) ) {
+			$dc = $matches[ 1 ];
+		}
+
+		return strtolower( $dc );
+	}
+
 	/**
 	 * Since we've established that enabling the protected content feature causes attachments
 	 * to be indexed, we should ensure that 'attachment' is in the indexable post types if
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -1588,6 +1588,10 @@ class Search {
 		return array_merge( $keys, $vip_search_allow_list_keys );
 	}
 
+	public function get_index_routing_allocation_include_dc() {
+		return defined( 'VIP_ORIGIN_DATACENTER' ) ? VIP_ORIGIN_DATACENTER : $this->get_origin_dc_from_es_host( $this->get_current_host() );
+	}
+
 	public function get_origin_dc_from_es_host( $host ) {
 		$dc = null;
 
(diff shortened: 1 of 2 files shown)


brandon-m-skinner (dismissed):
Fixed the failing test. 

Looks good. I'll leave the roll out to you since you're already coordinating on it.

search/includes/classes/class-search.php
```diff
@@ -1624,6 +1635,57 @@ public function filter__ep_prepare_meta_allowed_protected_keys( $keys, $post ) {
...
+			$mapping['settings']['index.routing.allocation.include.dc'] = $origin_datacenter;
+		}
+
+		return $mapping;
+	}
+
+	public function get_index_routing_allocation_include_dc() {
+		$dc = defined( 'VIP_ORIGIN_DATACENTER' ) ? VIP_ORIGIN_DATACENTER : $this->get_origin_dc_from_es_endpoint( $this->get_current_host() );
+
+		if ( ! in_array( $dc, self::ALLOWED_DATACENTERS ) ) {
```
rinatkhaziev, line 1661:
```suggestion
		if ( ! in_array( $dc, self::ALLOWED_DATACENTERS, true ) ) {
```

search/includes/classes/class-search.php
```diff
@@ -1624,6 +1635,57 @@ public function filter__ep_prepare_meta_allowed_protected_keys( $keys, $post ) {
...
+		if ( $origin_datacenter ) {
+			// We want all indexes to live in the site's origin datacenter
+			$mapping['settings']['index.routing.allocation.include.dc'] = $origin_datacenter;
+		}
+
+		return $mapping;
+	}
+
+	public function get_index_routing_allocation_include_dc() {
+		$dc = defined( 'VIP_ORIGIN_DATACENTER' ) ? VIP_ORIGIN_DATACENTER : $this->get_origin_dc_from_es_endpoint( $this->get_current_host() );
```
rinatkhaziev, line 1659:
```suggestion
		$dc = strtolower( defined( 'VIP_ORIGIN_DATACENTER' ) ? VIP_ORIGIN_DATACENTER : $this->get_origin_dc_from_es_endpoint( $this->get_current_host() ) );
```

Just to be extra safe here, we do usually refer to them in ALLCAPS, so it's not inconceivable that `VIP_ORIGIN_DATACENTER` might get set to ALLCAPS due to a mistake/misconfiguration

nickdaugherty: Good catch - I've updated this in 28b59c1fe (went with a slightly different approach for readability)

rinatkhaziev (approved):

PR #1990, opened 2021-02-17
Adds validate content lock

pschoffer:
## Description
* This makes sure not multiple processes are trying to validate (and
heal) content

### Plugin Updated: Search

Adds a locking mechanism to `wp vip-search health validate-contents` command. This makes sure that there is only one validation happening at one time. This will enable to make this into automated check in the future as well as avoid potential race condition issues with self-healing if differences.

## Checklist

- `n/a` This change has relevant documentation additions / updates (if applicable).

## Steps to Test

Either on a site that takes long time to `wp vip-search health validate-contents` try to run 2 at the same time. It shouldn't be possible anymore.

On smaller sites you can set the lock manually:
`lando wp eval "set_transient( 'vip_search_content_validation_lock', true, 20 );";`


search/includes/classes/class-health.php
```diff
@@ -250,9 +259,9 @@ public static function validate_index_posts_content( $start_post_id = 1, $last_p
 		}
 
 		// Get indexable objects
-		$indexable = Indexables::factory()->get( 'post' );
+		$indexable = $this->indexables->get( 'post' );
 
-		// Indexables::factory()->get() returns boolean|array
+		// Indexables::get() returns boolean|array
```
brandon-m-skinner, line 264:
```suggestion
		// $this->indexables->get() returns boolean|array
```

search/includes/classes/class-health.php
```diff
@@ -322,10 +335,26 @@ public static function validate_index_posts_content( $start_post_id = 1, $last_p
...
 	}
 
-	public static function validate_index_posts_content_batch( $indexable, $start_post_id, $next_batch_post_id, $inspect ) {
+	public function is_validate_content_ongoing(): bool {
+		$is_locked = get_transient( self::CONTENT_VALIDATION_LOCK_NAME, false );
+
+		return (bool) $is_locked;
+	}
+
+	public function reset_validate_content_ongoing() {
```
brandon-m-skinner, line 349: I think this should be `set_validate_content_lock` since all it does is set the transient `vip_search_content_validation_ongoing` to true for 15 minutes on a loop that acts as a lock. 

Reset implies some previous standard state that we're going back to. That isn't the case here.

What do you think?

search/includes/classes/class-health.php
```diff
@@ -322,10 +335,26 @@ public static function validate_index_posts_content( $start_post_id = 1, $last_p
...
+		set_transient( self::CONTENT_VALIDATION_LOCK_NAME, true, self::CONTENT_VALIDATION_LOCK_TIMEOUT );
+	}
+
+	public function delete_validate_content_ongoing() {
```
brandon-m-skinner, line 353: I think this should be `remove_validate_content_lock` or `delete_validate_content_lock`. The reasoning is we're operating on a lock.

What do you think?

search/includes/classes/class-health.php
```diff
@@ -12,6 +12,8 @@
 class Health {
 	const CONTENT_VALIDATION_BATCH_SIZE    = 500;
 	const CONTENT_VALIDATION_MAX_DIFF_SIZE = 1000;
+	const CONTENT_VALIDATION_LOCK_NAME = 'vip_search_content_validation_ongoing';
```
brandon-m-skinner, line 15: I think this should be `vip_search_content_validation_lock` since that's what it is.

What do you think?

search/includes/classes/class-health.php
```diff
@@ -12,6 +12,8 @@
...
+	const CONTENT_VALIDATION_LOCK_TIMEOUT = 900; // 15 min
```
brandon-m-skinner, line 16: I kinda think this should scale with the batch size. If it gets set super high and take 15 mins to run through we could potentially get duplicates.

What do you think?

brandon-m-skinner (changes requested):
Just some minor things for added clarity and potentially scaling the lock TTL based on batch size.

Let me know what you think 🙇 .

search/includes/classes/class-health.php
pschoffer: I am not sure... there seems to be no downside to have the number set to the upper limit in such scaling always. In the happy case scenario, it gets removed when done anyway. If we think 15 mins is too low we can do like 20 or even 30 I guess.

No downside meaning it would block another try for the remainder of that time if the validation would die unexpectedly. But that might be a good thing as well, or?

BTW this whole locking is modeled on the indexing lock including the naming. I don't think we need anything more advanced here.

brandon-m-skinner: Transients can be easily deleted via wp cli, so it's not a huge concern.

I don't think the indexing locks are particularly effective either for what it's worth since if you set the batch sizes really high you can bypass the locking.

My proposal is to just get the average time per-post of a higher batch, add a little to it, and then set the lock based on that value multiplied by batch size. This should allow us to safely handle whatever sized batches are handed in without too much added complexity.

If it doesn't seem necessary, we can ship as is 🤷 .

pschoffer: > just get the average time per-post of a higher batch
Do you have suggestions on doing that? 

My guess would be that it would vary somewhat based on how complex records each site has. So even if we would experimentally measure the time on one site it won't work well for another. Also, this code handles all types of indexables and I am not sure if _posts_ time would map to _users_ for instance. We could build some sort of dynamic mechanism that would keep updating the average per site and store it as an option or something. But as I mentioned before, I feel that is overkill.

brandon-m-skinner: It doesn't need to be perfect or overly complicated. Just run batches on a few sites, take the ones that take the longest per-post, add a bit to account for some sites being slower, and then just use that magic number plus basic multiplication against the batch size to get the lock length. Can always bump the magic number if it's too low over-time.

pschoffer: Did some testing. 
Slow Site:
**Batch Size, Average Time (seconds)**
500, 0.5
1000, 1.3
2000, 10,24

Fast Site:
**Batch Size, Average Time (seconds)**
500, 0.1
2000, 0.3
20000, 3

Sooo, the differences are quite huge. Maybe we could then do 1 second per 100 docs? Does that sound reasonable?

pschoffer: I added the suggested ^^

nickdaugherty:
Discussed on Slack a bit - in my mind there are two questions remaining:

* Do we want this to block CLI commands? My original idea was to only lock cron, so that new cron jobs wouldn't run until any previous job had finished up. We probably want the ability to run this via CLI manually, even if cron is running in the background all day
* For the locking, I like that we're bumping a transient on every batch (I had the same idea before I realized that's what you implemented :smile) - but I think we can just set a TTL of 15m or similar for that transient. If it hasn't been bumped in 15m, we can consider the job to be done, failed, or stuck. As long as it's a value that's well above what any given batch could reasonably be expected to take, it's safe and we don't need to guess how long it _might_ take (which can vary widely between sites).

pschoffer:
@nickdaugherty thanks for the feedback.

- Yeah that it is how it worked before, and as I mentioned before I kinda liked the simplicity and the fact that it works the same way as the indexing lock. @brandon-m-skinner pushed for the dynamic timer there though and I didnt have that strong conviction about it:) But I guess now when you are also for the fix timer, that makes us 2 vs 1.- I reverted the last change with dynamic TTL back to 15 mins.
- I added --force-parallel-execution (couldn't figure out a nicer name, suggestions are welcome) for the CLI so that we can "have the lock and eat it too" :) So by default you only get one execution no matter if cron or CLI, but if you --force it you can execute CLI even if some other process is already doing it.

search/includes/classes/commands/class-healthcommand.php
```diff
@@ -262,20 +262,26 @@ private function render_results( array $results ) {
...
+	 * : Optional Force execution even if the process is already ongoing
+	 *
 	 * ## EXAMPLES
 	 *     wp vip-search health validate-contents
 	 *
 	 * @subcommand validate-contents
 	 */
 	public function validate_contents( $args, $assoc_args ) {
-		$results = \Automattic\VIP\Search\Health::validate_index_posts_content(
+		$health = new \Automattic\VIP\Search\Health();
```
nickdaugherty, line 274: Something I changed recently is the `Health()` class now requires a `Search` instance to be passed to the constructor (for DI), so we'll have to add that here.

tests/search/includes/classes/test-class-health.php
```diff
@@ -438,4 +438,42 @@ public function test_validate_index_entity_count__skipping_non_initialized_index
...
+		$patrtially_mocked_health->indexables = $mocked_indexables;
+
+		$mocked_indexable = $this->getMockBuilder( \ElasticPress\Indexable::class )
+			->setMethods( [ 'query_db', 'prepare_document', 'put_mapping', 'build_mapping', 'build_settings' ] )
+			->getMock();
+
+		$mocked_indexables->method( 'get' )->willReturn( $mocked_indexable );
+
+		$patrtially_mocked_health->expects( $this->once() )->method( 'set_validate_content_lock' );
+		$patrtially_mocked_health->expects( $this->once() )->method( 'remove_validate_content_lock' );
```
nickdaugherty, line 475: Love all the mocking in this test!

search/includes/classes/commands/class-healthcommand.php
```diff
@@ -262,20 +262,26 @@ private function render_results( array $results ) {
 	 * [--silent]
 	 * : Optional silences all non-error output except for the final results
 	 *
+	 * [--force-parallel-execution]
```
nickdaugherty, line 265: I'm still not convinced we want to do locking on the CLI commands (even if it's bypass-able), but we can go with this and see if it ends up being an issue in practice.

nickdaugherty: Along those lines, can we detect the "locked" error (with code `content_validation_already_ongoing`) in the error handler (line 287) and print out a message indicating it can be bypassed with this option?

nickdaugherty (approved):
@pschoffer looks good - I left a few comments but nothing major. There's also a merge conflict now. Other than that, good to go.

search/includes/classes/commands/class-healthcommand.php
pschoffer: Good idea!

pschoffer (commented):
Thanks for the input @nickdaugherty. Back to you:)

nickdaugherty (approved):
Looks good 🚢

nickdaugherty:
Looks good @pschoffer - I ran out of time in my day to deploy this to staging, but it's good to go.

PR #1995, opened 2021-02-19
Move WP_EP_DEBUG defining and EP debug bar panel loading to before dependencies are loaded

brandon-m-skinner:
## Description

We changed the loading order for VIP Search a little while back. This was done because we needed ElasticPress to correctly check if the protected content feature was enabled or not so we could add attachments as a default indexable post type under the right circumstances.

This seemed fine, but apparently broke the ElasticPress debug bar panel when it's loaded into debug bar. If it was loaded into query monitor it worked as expected. It also worked fine if WP_DEBUG or WP_EP_DEBUG or EP_QUERY_LOG were true in the config.

This PR moves the all important defining of WP_EP_DEBUG when debug bar or query monitor are enabled to before dependencies are loaded. Thereby having the constant set appropriately for when ElasticPress is instantiated.

### Plugin Updated: Search

Fixed an issue where the ElasticPress debug bar panel wasn't showing the expected data when used with debug bar without WP_DEBUG or equivalent constant set to true.

## Checklist

## Steps to Test

1. Disable WP_DEBUG.
2. Perform a search query using VIP Search.
3. View the ElasticPress panel in debug bar and note that it has no queries.
4. Apply PR.
5. Repeat 2.
6. The expected data should now be in the ElasticPress panel in debug bar.

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -104,6 +104,7 @@ class Search {
 	public function init() {
 		$this->apply_settings(); // Applies filters for tweakable Search settings and should run first.
 		$this->setup_constants();
+		$this->maybe_enable_ep_query_logging();
 		$this->load_dependencies();
 		$this->setup_hooks();
 		$this->load_commands();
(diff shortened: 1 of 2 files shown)


pschoffer:
Works as expected1

pschoffer (approved):

PR #1997, opened 2021-02-19
Update/phpcs class vip filesystem

pschoffer:
### Code Standards: files/class-vip-filesystem.php

We’re continuing to clean up code that isn’t 100% compliant with our coding standards.

## Checklist

## Steps to Test
Nothing really as it is mostly space/comment change and the only other change should be covered in unit-tests.

Uploading a picture should be good test though.

### DIFF
--- a/files/class-vip-filesystem.php
+++ b/files/class-vip-filesystem.php
@@ -81,7 +81,7 @@ class VIP_Filesystem {
 
 		// Create and register stream
 		$this->stream_wrapper = new VIP_Filesystem_Stream_Wrapper( new_api_client(),
-			self::PROTOCOL );
+		self::PROTOCOL );
 		$this->stream_wrapper->register();
 	}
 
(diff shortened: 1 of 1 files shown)


files/class-vip-filesystem.php
```diff
@@ -221,6 +221,7 @@ protected function validate_file_type( $file_path ) {
 		if ( is_wp_error( $result ) ) {
 			if ( 'invalid-file-type' !== $result->get_error_code() ) {
 				trigger_error(
+					// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
pschoffer, line 224: Not sure if we should escape here, or leave ignored? If escape, which wp escape function would be a good fit here?

brandon-m-skinner: It should be fine without it. Even single quotes passed in via get_error_message would be handled appropriately by sprintf.

brandon-m-skinner (approved):

brandon-m-skinner:
r1399-stacks

PR #1998, opened 2021-02-19
Fix incorrect limit return for wpcom_vip_top_posts_array

mikeyarce:
## Description

Due to how WordPress.com caches these requests, `wpcom_vip_top_posts_array` can return incomplete data.  WordPress.com will cache these by number of days and not by limit.

Example:
```
wpcom_vip_top_posts_array( 5, 10 );   // 10 posts returned
wpcom_vip_top_posts_array( 5, 100 );  // 10 posts returned
```

In this change, we set the `limit` passed to WordPress.com to always be 100, so we'll always get the maximum amount of posts available.  This function will then only return the amount based on the `limit` that it was passed.

### Bug Fix: Fixed an issue where wpcom_vip_top_posts_array() was not always returning the correct amount of posts

Previously, it was possible for our requests to return cached results with the wrong number of posts. This has been rectified.

## Checklist

## Steps to Test

1. Check out PR.
2. Sandbox a VIP Go test site, make sure it has visits!  
3. If you need to spoof visits - open up an Incognito/Private window and just reload a bunch.  Or run an Apache Bench command. 
2. Open up WP_CLI and use `wp shell`
3. Run `wpcom_vip_top_posts_array( 5, 10 );`  // Did you get 10 posts?
4. Now run `wpcom_vip_top_posts_array( 5, 20 );` // Did you get 20 posts?


vip-helpers/vip-stats.php
```diff
@@ -34,7 +34,7 @@ function wpcom_vip_top_posts_array( $num_days = 30, $limit = 10, $end_date = fal
 		
 		$args = array(
 			'days'  => $num_days,
-			'limit' => $limit,
+			'limit' => 100,
```
trepmal, line 37: I think an inline explanation for why $limit isn't used here would be valuable for the future.

mikeyarce: What do you think of this...
`'limit' => 100, // Due to caching, we request max limit and only return requested $limit below`

trepmal: I think that would be fine. Perhaps a reference back to this PR for fuller context.

brandon-m-skinner:
Adjusted changelog description to be in line with existing entries.

nickdaugherty (approved):
Looks great, thanks @mikeyarce !

nickdaugherty:
r1420-stacks

PR #1999, opened 2021-02-22
Update async publish actions

WPprodigy:
## Description

Prevents a large number of cron jobs from being registered, unless they are truly needed.

This PR includes 3 steps:

1) Track the usage of this feature, to better help make future decisions.
2) Remove the two "default" offloads.
3) Only register the cron job when the hooks are being used.

### Some Considerations

For 2, by removing the two default offloads - the [track_publish_post()](https://github.com/Automattic/vip-go-mu-plugins/blob/82308873535a775d263daa3f2cc24a36d8774ccd/stats.php#L24) and [_publish_post_hook()](https://github.com/WordPress/WordPress/blob/c6c1ef3c32ee4bfdc86a7dc2290604aee3bc3de8/wp-includes/post.php#L7289) logic will now happen synchronously when a post is published, the effects of which quite minimal here. One additional note is that `track_publish_post()` would have technically only have been run when the async cron job was truly added based on the `should_offload()` default logic, meaning it was also excluded from being run on new posts created during cron, cli, and xmlrpc. This could be replicated if desired by checking for more than just `WP_IMPORTING` in this function.

For 3, if one of these hooks was for some reason only conditionally registered underneath cli/cron checks specifically - then we'd possibly be missing out that the cron needed to be registered. This is quite unlikely though, and I have searched for all use cases I can find and the few out there do not fall into this trap. But if desired, we could move step 3 into a future PR for after we have more usage information.

### Plugin Updated: Async Publish Actions

This change removes the creation of async post publish cron jobs if they are not being utilized. If the relevant hooks _are_ in use, the cron jobs will continue to be created normally.

## Checklist

Documentation: https://docs.wpvip.com/how-tos/code-quality-and-best-practices/write-asynchronous-publishing-actions/. Will be updated when applicable.

## Steps to Test

1. Check out PR.
1. Create and publish a new post.
1. Quickly run `wp cron-control events list` and notice how no `wpcom_vip_async_transition_post_status` job was added.
1. Now add something like the below, and note how a cron job is registered and run after publishing a post:

```
add_action( 'async_transition_post_status', function() {
	error_log('hello');
} );
```

For phpunit, `phpunit --group async-publish-actions` to skip to the relevant parts.

### DIFF
--- a/async-publish-actions.php
+++ b/async-publish-actions.php
@@ -61,6 +61,13 @@ function _wpcom_do_async_transition_post_status( $post_id, $new_status, $old_sta
 		return;
 	}
 
+	// Gauge how often this feature is used.
+	if ( true === WPCOM_IS_VIP_ENV && false === WPCOM_SANDBOXED ) {
+		\Automattic\VIP\Stats\send_pixel( [
+			'vip-go-async-publish-job' => FILES_CLIENT_SITE_ID,
+		] );
+	}
+
 	/**
 	 * Fires when a post is transitioned from one status to another.
 	 *
@@ -132,6 +139,11 @@ function _queue_async_hooks( $new_status, $old_status, $post ) {
 		return;
 	}
 
+	// If nothing is hooked on, avoid scheduling a cron job.
+	if ( ! has_action( 'async_transition_post_status' ) && ! has_action( "async_{$old_status}_to_{$new_status}" ) && ! has_action( "async_{$new_status}_{$post->post_type}" ) ) {
+		return;
+	}
+
 	if ( false !== wp_next_scheduled( ASYNC_TRANSITION_EVENT, $args ) ) {
 		return;
 	}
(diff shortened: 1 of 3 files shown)


async-publish-actions.php
```diff
@@ -61,6 +61,13 @@ function _wpcom_do_async_transition_post_status( $post_id, $new_status, $old_sta
 		return;
 	}
 
+	// Gauge how often this feature is used.
+	if ( true === WPCOM_IS_VIP_ENV && false === WPCOM_SANDBOXED ) {
+		\Automattic\VIP\Stats\send_pixel( [
+			'vip-go-async-publish-job' => FILES_CLIENT_SITE_ID,
```
nickdaugherty, line 67: Minor note - this handles all post status transitions (not just publish) so maybe we want the stat name to reflect that? `vip-go-async-post-transition-status-job` or similar?

WPprodigy: TIL that technically the job is only registered when going to/from publish: https://github.com/Automattic/vip-go-mu-plugins/blob/12f85b7e170425014854a1b1b1799871a5654f06/async-publish-actions.php#L151

But definitely not oppose to renaming the stat pixel.

nickdaugherty: Oh i didn't realize that - alright in that case this is fine then.

nickdaugherty (approved):

nickdaugherty:
r1404-stacks

PR #2001, opened 2021-02-24
Refresh Jetpack Connection Pilot (A Space Odyssey)

WPprodigy:
## Description

1) By checking against `hashed_site_url`, we can better ensure there won't be a mixup due to search/replaces & migrations/data syncs.
2) Log "last connection info" every time when possible.
3) Only attempt to register the cron job in admin/cron (prevent every FE request from triggering).
4) Remove deprecated `Jetpack_Client::wpcom_json_api_request_as_blog` call. Was deprecated in JP 7.5, and supposed to be removed in 8.1 (but still there).
5) Refresh bash script from JP source. We still need to override though ([more info](https://github.com/Automattic/vip-go-mu-plugins/pull/986#issuecomment-784730476)).

### Refresh the Jetpack Connection Pilot module

In preparation for the rollout for improved Jetpack connection monitoring, the following notable changes have been made:

- Hash and save the existing site_url in the healtcheck so it can be more safely compared in the future before auto-reconnecting.
- Prevent the cron schedule from being checked/registered on `init`, instead only checking during admin_init/cli.
- Update calls to deprecated JP methods to prevent php warnings.

## Checklist

## Steps to Test

Needs to be done on an active sandbox. A few scenarios worth testing (be sure to apply PR on sandbox):

### Active Connections

Ensure `define( 'VIP_JETPACK_CONNECTION_PILOT_SHOULD_RUN', true );` is added in vip-config.

```
# Make sure JP is connected
wp jetpack status
// wp jetpack-start connect

# Delete the heartbeat option
wp option delete vip_jetpack_connection_pilot_heartbeat

# Run the cron job on the sandbox
wp shell
\Automattic\VIP\Jetpack\Connection_Pilot::do_cron();

# Now check that the heartbeat updated correctly
wp option get vip_jetpack_connection_pilot_heartbeat
```

### Inactive Connections

Additionally add `define( 'VIP_JETPACK_CONNECTION_PILOT_SHOULD_RECONNECT', true );` this time.

```
# Kill the connection
// wp jetpack disconnect blog

# You'll want to have a heartbeat option set still.
wp option get vip_jetpack_connection_pilot_heartbeat

# Run the cron job on the sandbox
wp shell
\Automattic\VIP\Jetpack\Connection_Pilot::do_cron();

# Now JP should be reconnected
wp jetpack status
```

You can't really test the full logging on a sandbox, but once deployed we'll be able to test the full thing with similar steps as above. Except to trigger the full logging, you'll run the cron on a "real" container:

```
# Find the ID for wpcom_vip_run_jetpack_connection_pilot, and run it
vip @<name>.<env> -- wp cron-control events list
vip @<name>.<env> -- wp cron-control events run 27
```
Note you'll want the constant(s) actually committed to the site in this case.

### DIFF
--- a/vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
+++ b/vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
@@ -203,13 +203,13 @@ class Connection_Pilot {
 	 *
 	 * @param string   $message optional.
 	 * @param \WP_Error $wp_error optional.
-	 * @param array    $last_heartbeat optional.
 	 *
 	 * @return mixed True if the message was sent to IRC, false if it failed. If sandboxed, will just return the message string.
 	 */
-	protected function send_alert( $message = '', $wp_error = null, $last_heartbeat = null ) {
+	protected function send_alert( $message = '', $wp_error = null ) {
 		$message .= sprintf( ' Site: %s (ID %d).', get_site_url(), defined( 'VIP_GO_APP_ID' ) ? VIP_GO_APP_ID : 0 );
 
+		$last_heartbeat = $this->last_heartbeat;
 		if ( isset( $last_heartbeat['site_url'], $last_heartbeat['cache_site_id'], $last_heartbeat['timestamp'] ) ) {
 			$message .= sprintf(
 				' The last known connection was on %s UTC to Cache Site ID %d (%s).',
--- a/vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
+++ b/vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
@@ -26,11 +26,11 @@ class Connection_Pilot {
 	 * See the a8c_cron_control_clean_legacy_data event for more details.
 	 */
 	const CRON_SCHEDULE = 'hourly';
-	
+
 	/**
 	 * The healtcheck option's current data.
 	 *
-	 * Example: [ 'site_url' => 'https://example.go-vip.co', 'cache_site_id' => 1234, 'timestamp' => 1555124370 ]
+	 * Example: [ 'site_url' => 'https://example.go-vip.co', 'hashed_site_url' => '371a92eb7d5d63007db216dbd3b49187', 'cache_site_id' => 1234, 'timestamp' => 1555124370 ]
 	 *
 	 * @var mixed False if doesn't exist, else an array with the data shown above.
 	 */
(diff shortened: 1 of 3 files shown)


vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
```diff
@@ -220,8 +226,8 @@ protected function send_alert( $message = '', $wp_error = null, $last_heartbeat
 			$message .= sprintf( ' Jetpack connection error: [%s] %s', $wp_error->get_error_code(), $wp_error->get_error_message() );
 		}
 
-		if ( ( defined( 'WPCOM_SANDBOXED' ) && WPCOM_SANDBOXED ) || 
-			( ! defined( 'ALERT_SERVICE_ADDRESS' ) ) || 
+		if ( ( defined( 'WPCOM_SANDBOXED' ) && WPCOM_SANDBOXED ) ||
```
rinatkhaziev, line 229: Let's simplify this whole block:
1) we don't really care about any of these constants because on a sandbox the call to Alerts service will fail regardless.
2)Instead of using `error_log` let's use log2logstash (which also writes to error log)

vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
```diff
@@ -220,8 +226,8 @@ protected function send_alert( $message = '', $wp_error = null, $last_heartbeat
...
+			( ! defined( 'ALERT_SERVICE_ADDRESS' ) ) ||
 			( defined( 'VIP_JETPACK_CONNECTION_PILOT_SILENCE_ALERTS' ) && VIP_JETPACK_CONNECTION_PILOT_SILENCE_ALERTS ) ) {
 			error_log( $message );
```
rinatkhaziev, line 232:
```suggestion
		\Automattic\VIP\Logstash\log2logstash( [
			'severity' => 'error',
			'feature' => 'jetpack-connection-pilot',
			'message' => $message,
			'extra' => [], // can stuff additional info here
		] );
```

vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
WPprodigy: Should be good now.

rinatkhaziev (approved):
:shipit:

rinatkhaziev:
r1412-stacks

PR #2004, opened 2021-03-01
Search: Add result compression

pschoffer:
### Plugin Updated: Search

This change enables compression on HTTP requests towards Elastic Search.

## Checklist

## Steps to Test
1. Add a debug line `var_dump(['Response headers', $response['headers'] ]);`
2. Do any ES request e.g. `wp vip-search documents get post 1`
3. See the headers - `'content-encoding'` is not there
4. Apply the PR
5.  Do any ES request e.g. `wp vip-search documents get post 1`
6. Verify `'content-encoding'` header is present

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -648,7 +648,14 @@ class Search {
 			$args['headers'] = [];
 		}
 
-		$args['headers'] = array_merge( $args['headers'], array( 'X-Client-Site-ID' => FILES_CLIENT_SITE_ID, 'X-Client-Env' => VIP_GO_ENV ) );
+		$args['headers'] = array_merge(
+			$args['headers'],
+			[
+				'X-Client-Site-ID' => FILES_CLIENT_SITE_ID,
+				'X-Client-Env' => VIP_GO_ENV,
+				'Accept-Encoding' => 'gzip, br',
+			]
+		);
 
 		$statsd_mode = $this->get_statsd_request_mode_for_request( $query['url'], $args );
 		$collect_per_doc_metric = $this->is_bulk_url( $query['url'] );
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -653,7 +653,7 @@ class Search {
 			[
 				'X-Client-Site-ID' => FILES_CLIENT_SITE_ID,
 				'X-Client-Env' => VIP_GO_ENV,
-				'Accept-Encoding' => 'gzip, br',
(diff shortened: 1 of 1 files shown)


search/includes/classes/class-search.php
```diff
@@ -648,7 +648,14 @@ public function filter__ep_do_intercept_request( $request, $query, $args, $failu
 			$args['headers'] = [];
 		}
 
-		$args['headers'] = array_merge( $args['headers'], array( 'X-Client-Site-ID' => FILES_CLIENT_SITE_ID, 'X-Client-Env' => VIP_GO_ENV ) );
+		$args['headers'] = array_merge(
+			$args['headers'],
+			[
+				'X-Client-Site-ID' => FILES_CLIENT_SITE_ID,
+				'X-Client-Env' => VIP_GO_ENV,
+				'Accept-Encoding' => 'gzip, br',
```
rinatkhaziev, line 656:
```suggestion
				'Accept-Encoding' => 'gzip, deflate',
```

Brotli is currently not supported by the libcurl version we use:

```
curl --compressed https://httpbin.org/brotli
curl: (61) Unrecognized content encoding type. libcurl understands deflate, gzip content encodings.
```

rinatkhaziev (changes requested):
Thanks for the PR, had no idea we don't do this already :)

Unfortunately `libcurl` we use is not compiled with brotli support, so let's swap that out to what we currently support.

pschoffer:
Cool catch!

rinatkhaziev (approved):
:shipit:

nickdaugherty (approved):

nickdaugherty:
r1416-stacks

nickdaugherty:
@pschoffer I deployed this to staging, but i'm not 100% sure it's working as I'm not seeing the new header in requests when viewed through Debug Bar. It could be that it's a debug bar thing, but worth double checking.

pschoffer:
I double checked today. It does seem to work when I print out headers directly in code. I got:
```
 string 'HTTP/1.1 200 OK
Server: nginx
Date: Tue, 02 Mar 2021 10:33:09 GMT
Content-Type: application/json; charset=UTF-8
Transfer-Encoding: chunked
Connection: close
Vary: Accept-Encoding
Content-Encoding: gzip
```

So that looks good. I still see only a subset of the request headers without the `Accept-Encoding` among others. So I guess it is filtered somewhere in debug bar.

nickdaugherty:
Great, thanks for checking @pschoffer !

PR #2005, opened 2021-03-01
Merge 'add/local-vip-go-env-const'... 

seanosh:
## Description

When setting up a local development environment that utilizes `vip-go-mu-plugins`, it's helpful to enable debug bar and relevant tooling. Setting `WPCOM_IS_VIP_ENV` to `false` is correct for a local environment, but doesn't offer a way to selectively enable something like debug tooling. 

I went with this approach because other logic for enabling mostly revolves around things specific to the platform (employee, proxy status) that would require additional/riskier changes or inclusion of hooks/filters at the onset locally, which isn't as "clean" as defining constant for storing environment type. 

cc: @joshbetz (thanks for being a sounding board on this!)

### Feature: Enable Logged Out Debug Mode for local envs

Enables Debug mode when `WP_ENVIRONMENT_TYPE' is defined and set to `local`:

``` 
define( 'WP_ENVIRONMENT_TYPE', 'local' );
```

## Checklist

## Steps to Test
1. Check out and apply PR in local environment
2. Add to `wp-config.php`:
```
define( 'VIP_GO_ENV', 'local' );
define( 'VIP_BLOCK_WP_MAIL', true );
define( 'FILES_CLIENT_SITE_ID', 1 );
```
3. Visit local environment URL without having logged in and observe debug panel presence.


mjangda:
Might be better to set and check `'local' === WP_ENVIRONMENT_TYPE`? (https://make.wordpress.org/core/2020/07/24/new-wp_get_environment_type-function-in-wordpress-5-5/)

A non-false value for VIP_GO_ENV might have other unintended consequences.

seanosh:
Nice, thanks @mjangda - updated the PR accordingly.

rinatkhaziev:
To really make it magical we also need to define the `local` in this block: 

https://github.com/Automattic/vip-go-mu-plugins/blob/02729f0e4f23c72dee12d2b2208d9b2384e42fd7/000-vip-init.php#L208-L224

000-debug/debug-mode.php
```diff
@@ -39,8 +39,9 @@ function is_debug_mode_enabled() {
 	$is_nocache = isset( $_COOKIE['vip-go-cb'] ) && '1' === $_COOKIE['vip-go-cb'];
 	$is_debug = isset( $_COOKIE['a8c-debug'] ) && '1' === $_COOKIE['a8c-debug'];
 	$is_proxied = \is_proxied_request();
+	$is_local = ( ( defined( 'WP_ENVIRONMENT_TYPE' ) && 'local' === WP_ENVIRONMENT_TYPE ) ? true : false );
```
rinatkhaziev, line 42:
```suggestion
	$is_local = defined( 'WP_ENVIRONMENT_TYPE' ) && 'local' === WP_ENVIRONMENT_TYPE;
```

rinatkhaziev: The ternary is redundant here.

rinatkhaziev (changes requested):
Thanks for the addition @seanosh, left a suggestion.

rinatkhaziev (approved):

rinatkhaziev:
Updated changelog entry

nickdaugherty:
r1443-stacks

nickdaugherty:
Release v1.22.0

PR #2007, opened 2021-03-01
Fix image sizes and srcset for intermediate images

### LINKED ISSUE #1391: Filter `sizes` value for intermediate images
## Desired Behavior

Responsive images should have a value on the `sizes` attribute which reflects the width of the chosen intermediate size. 

## Actual Behavior

When images are added to posts using the block editor, the value of the sizes attribute currently reflects the width of the original/full-size image. This can cause these images to display at the incorrect dimensions. 

For more information:

P2: p9zhOE-Qw-p2
Ticket: #102087-zd-wordpressvip

mikeyarce:
## Description
This PR uses a new filter available in WP 5.7 to properly pass image dimensions for the image `srcset` and `sizes`.
This needs to be tested and merged in coordination with the WP 5.7 release (March 9 is the current tentative release date).

New filter added in WP 5.7:  https://core.trac.wordpress.org/browser/trunk/src/wp-includes/media.php?rev=50136#L1643

How the fix works:
- First we hook into `wp_img_tag_add_width_and_height_attr` which is early enough that our image source still has the query arguments available so we know what the image size is intended to be
- We save that image size data and then pass it into `wp_image_src_get_dimensions` where we set image dimensions
- If the image is a full size image, we set the default dimensions
- If the image is not full size, we calculate the image size based on the given width and the calculated aspect ratio.

#### Bug Fix: Fixed an issue where images were not properly sized when using the Block Editor and intermediate image sizes.

Previously, all images added through the Block Editor that were assigned intermediate image sizes ( e.g. Thumbnail ) would appear as full size images.

## Checklist

## Steps to Test
1. Patch your `wp-includes/media.php` file to include the changes to `wp_image_src_get_dimensions` found here: https://core.trac.wordpress.org/browser/trunk/src/wp-includes/media.php?rev=50136#L1601
2. Apply this code to your sandbox
3. Make sure your test site has Gutenberg Enabled
4. Create a new Post
5. Select the Image Block, add an image
6. From the Block Properties, select an intermediate image size like Thumbnail
7. Add more image blocks for Full Size, Medium, Large, etc..
8. Verify that the `srcset` and `sizes` attributes output are correct.

Before the change:
before-patch-srcset

After the change:
after-patch-srcset


GaryJones:
Will those who are still on WP versions before 5.7, continue to see the same full size image for each named image size (i.e. the Before)?

a8c-files.php
```diff
@@ -1156,6 +1156,59 @@ function a8c_files_maybe_inject_image_sizes( $data, $attachment_id ) {
...
 }
+/**
+ * Filter dimensions for images to return a proper image size and srcset values
+ *
+ * @param bool   $value         The filtered value, defaults to `true`.
+ * @param string $image         The HTML `img` tag where the attribute should be added.
+ * @param string $context       Additional context about how the function was called or where the img tag is.
+ * @param int    $attachment_id The image attachment ID.
+ */
+function a8c_files_maybe_inject_dimensions( $value, $image, $context, $attachment_id ) {
```
GaryJones, line 1167: Any reason not to use type declarations on this parameters?

a8c-files.php
```diff
@@ -1156,6 +1156,59 @@ function a8c_files_maybe_inject_image_sizes( $data, $attachment_id ) {
...
+				$dimensions = array(
+					(int) $query_args['w'],
+					(int) ceil( $query_args['w'] * $aspect_ratio ),
+				);
+			}	
+		}
+		return $dimensions;
+
+	}, 10, 4 );
+    return true;
```
GaryJones, line 1210: Indented with 4 spaces instead of a tab character.

a8c-files.php
```diff
@@ -1156,6 +1156,59 @@ function a8c_files_maybe_inject_image_sizes( $data, $attachment_id ) {
...
+ * @param bool   $value         The filtered value, defaults to `true`.
+ * @param string $image         The HTML `img` tag where the attribute should be added.
+ * @param string $context       Additional context about how the function was called or where the img tag is.
+ * @param int    $attachment_id The image attachment ID.
+ */
+function a8c_files_maybe_inject_dimensions( $value, $image, $context, $attachment_id ) {
+	$vipgo_img_src = preg_match( '/src="([^"]+)"/', $image, $match_src ) ? $match_src[1] : '';
+
+	/** This filter is documented in wp-admin/includes/media.php */
+	add_filter( 'wp_image_src_get_dimensions', function( $dimensions, $image_src, $image_meta, $attachment_id ) use ( $vipgo_img_src ){
```
GaryJones, line 1171: This closure is like 90% of the parent function.

Consider extracting it a named function, so that it can then be properly tested.

a8c-files.php
```diff
@@ -1156,6 +1156,59 @@ function a8c_files_maybe_inject_image_sizes( $data, $attachment_id ) {
...
+		if ( 
+			! isset( $url['query'] ) &&
+			strpos( $image_src, $image_meta['file'] ) !== false
+		) {
+			$dimensions = array(
+				(int) $image_meta['width'],
+				(int) $image_meta['height'],
+		);
+		// If it's not Full Size, let's use our calculated Aspect Ratio to output proper dimensions according to the size
+		} else if (
```
GaryJones, line 1190:
```suggestion
		} elseif ( 
```

a8c-files.php
```diff
@@ -1156,6 +1156,59 @@ function a8c_files_maybe_inject_image_sizes( $data, $attachment_id ) {
...
+		// Is it a Full Size image?  We're checking for sizes in query arguments
+		if ( 
+			! isset( $url['query'] ) &&
+			strpos( $image_src, $image_meta['file'] ) !== false
+		) {
+			$dimensions = array(
+				(int) $image_meta['width'],
+				(int) $image_meta['height'],
+		);
+		// If it's not Full Size, let's use our calculated Aspect Ratio to output proper dimensions according to the size
```
GaryJones, line 1189:
```suggestion
		// If it's not Full Size, let's use our calculated Aspect Ratio to output proper dimensions according to the size.
```

a8c-files.php
```diff
@@ -1156,6 +1156,59 @@ function a8c_files_maybe_inject_image_sizes( $data, $attachment_id ) {
...
+		} else if ( 
+			isset( $url['query'] ) &&
+			strpos( $image_src, $image_meta['file'] ) !== false 
+		) {
+			if ( isset( $query_args['w'] ) && isset( $query_args['h'] ) ) {
+				$dimensions = array(
+					(int) $query_args['w'],
+					(int) $query_args['h'],
+			);
```
GaryJones, line 1198:
```suggestion
				);
```

a8c-files.php
mikeyarce: Good question!   We're not really using most of these since we're just returning `true` for the value in this filter and we're not returning any of them either, so I wasn't sure if that would be helpful in this case.  Would you suggest still adding them?  I also wasn't sure if I should pass them all even if we're not using them?

a8c-files.php
mikeyarce: I will make the change but I'm curious as to why.  From the PHP docs it looks like `else if` and `elseif` are the same when using curly brackets but not the same when using colons.  Is it just a stylistic choice in this case?

mikeyarce:
Thanks for your review @GaryJones!  I've addressed the issues and separated out the dimension logic into another function.

> Will those who are still on WP versions before 5.7, continue to see the same full size image for each named image size (i.e. the Before)?

I tested this and for versions before 5.7, site's will continue to see the same `before` full size image for each named image.

rinatkhaziev:
We're investigating other avenues, in the meantime, we'll keep it open but with the appropriate status.

a8c-files.php
GaryJones: If you're not using them, then there is no need to support them in the function signature, unless a param to the right is being used.

However, since you need `$attachment_id`, then they all need to be passed.

The non-first types should be known, as they are from core code, but the first type may have been filtered at an earlier priority into a different type, so although it's not being used here, the advantage of using types here is that it becomes an extra check in case some other code is doing something wrong. It's also good for IDEs and static analysis tools, like Psalm or Phpstan

Adding the return type also has no drawbacks.

mchanDev:
Hi @rinatkhaziev 

Is there a work around for this atm?

There is an open ticket here [#128167-z](https://wordpressvip.zendesk.com/agent/tickets/128167) that I would like to reply to

a8c-files.php
GaryJones: @mikeyarce It's a WPCS choice: https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#use-elseif-not-else-if plus, `elseif` is then a unique keyword, should one need to find instances of it.

PR #2014, opened 2021-03-04
Use ambiguous message that does not reveal the type of login error

rebeccahum:
## Description
The default WP login error message can reveal whether a user exists and if the password is incorrect. We should change it to be more ambiguous.

Username doesn't exist:
Screen Shot 2021-03-04 at 4 49 29 PM

Username exists but password is incorrect:
Screen Shot 2021-03-04 at 4 50 07 PM

After the change

## Checklist


rinatkhaziev (dismissed):
Thank you! Can't believe we haven't fixed that before :)

security/login-error.php
```diff
@@ -0,0 +1,23 @@
...
+
+	$err_types = [
+		'invalid_username',
+		'invalid_email',
+		'incorrect_password',
+	];
+
+	foreach ( $err_types as $err ) {
+		if ( in_array( $err, $err_codes, true ) ) {
+			$error = '<strong>ERROR</strong>: Invalid login name or password.';
```
rinatkhaziev, line 16: No biggie, but since we only care about any of those we can just break on the first one?
```suggestion
			$error = '<strong>ERROR</strong>: Invalid login name or password.';
			break;
```

GaryJones:
This is not considered a security issue at the WP core level: https://core.trac.wordpress.org/ticket/3708

As noted, it's also a WCAG requirement to ensure error messages are clear and accurately describe the problem so the user can fix it. Ultimately this makes it harder for valid users to log in as well, which then makes it frustrating. 

Since usernames are not considered to be secret in WP, and email addresses are often guessable anyway, and there are better ways to make accounts secure (2FA, SSO, strong passwords, limited roles, etc.), then I believe the change proposed here would be more negative than positive.

I'd rather the code be put into the public docs, so that customers could choose to use this behaviour themselves.

rinatkhaziev:
Blocked pending internal discussion.

nickdaugherty:
> it's also a WCAG requirement to ensure error messages are clear and accurately describe the problem so the user can fix it

@GaryJones do you think that a generic `Invalid credentials. Please try again.` message is actually less accessible than separate email and password messages? The problem is still described with a clear message in both scenarios - that the credentials are incorrect - so I'm trying to understand how this would be an accessibility issue.

OWASP recommends a generic message that doesn't reveal which part is incorrect. https://cheatsheetseries.owasp.org/cheatsheets/Authentication_Cheat_Sheet.html#authentication-and-error-messages

GaryJones:
@nickdaugherty Thanks for bringing this up again.

I took a deeper look at the WCAG. There are two Success Criterions here:

 - [3.3.1 Error Identification](https://www.w3.org/WAI/WCAG21/Understanding/error-identification.html) (Level A)
   > If an input error is automatically detected, the item that is in error is identified and the error is described to the user in text.
 - [3.3.3 Error Suggestion](https://www.w3.org/WAI/WCAG21/Understanding/error-suggestion.html) (Level AA)
   > If an input error is automatically detected and suggestions for correction are known, then the suggestions are provided to the user, **unless it would jeopardize the security or purpose of the content**.
   
   (my emphasis)

I think I was probably referencing the first SC, but the second SC does have wiggle room to account for the suggestion that OWASP makes.

One suggestion I would want to see though, is to reference the name of the field labels in the generic message, rather than introducing new jargon. i.e. since, the labels say "Username or Email Address" and "Password", that the message would be something like. "The Username / Email Address or Password is incorrect", rather than "Invalid credentials" to avoid the case where someone (perhaps a non-English speaker or someone with cognitive difficulties) doesn't understand what is meant by "credentials".

Would that be a happy middle ground for everyone?

rebeccahum:
What about "Invalid login or password. Please try again."? With login referring to username or e-mail address then..does that work?

GaryJones:
"login" isn't in the label of either field though.

Your initial screenshots show that "username", "email address", and "password" are all mentioned in error message. 

Here is a login error message in Greek (which I don't read):

Screenshot 2021-04-08 at 10 35 37

Can you guess which error message is it?

Other than an uppercase/lowercase first letter, I can see that the error message mentions one of the labels, so I could understand that it might be that which is wrong. 

Screenshot 2021-04-08 at 10 24 30

If the error message was the Greek equivalent of "login", "credentials", "authentication" or some other term which isn't already present on the page, then I'd have fewer hints to know what was wrong. I've used Greek as an example, but it would also apply to those with cognitive difficulties as well. The issue is harder to see when we don't have those difficulties, and we're fluent with the language, and familiar with WordPress.

I know it's a little wordy, but we're not limited in space here (the error message in the screenshots is already longer than what has been proposed), and I believe including the field label names is the way to go.

Options:

- "The username / email address or password is incorrect."
- "The username or email address, or password is incorrect."
- "The username or password is incorrect." + "The email address or password is incorrect." (depending on which they enter)

security/login-error.php
```diff
@@ -0,0 +1,25 @@
...
+	foreach ( $err_types as $err ) {
+		if ( in_array( $err, $err_codes, true ) ) {
+			$error = '<strong>ERROR</strong>: The username/e-mail address or password is incorrect. Please try again.';
+			break;
+		}
+	}
+ 
+	return $error;
+}
+// Use a login message that does not reveal the type of login error in brute-forces.
```
GaryJones, line 24: Why not move this comment into a DocBlock and formally document the function? Add the `@param` (and `@since`?), etc.

security/login-error.php
```diff
@@ -0,0 +1,25 @@
+<?php
+namespace Automattic\VIP\Security;
+
+function use_ambiguous_login_error( $error ) {
```
GaryJones, line 4: Can param and return type declarations be added?

security/login-error.php
```diff
@@ -0,0 +1,25 @@
...
+	$err_types = [
+		'invalid_username',
+		'invalid_email',
+		'incorrect_password',
+		'invalidcombo',
+	];
+
+	foreach ( $err_types as $err ) {
+		if ( in_array( $err, $err_codes, true ) ) {
+			$error = '<strong>ERROR</strong>: The username/e-mail address or password is incorrect. Please try again.';
```
GaryJones, line 17: Why are you shouting? It's already strongly emphasised with HTML - it doesn't need allcaps as well 😄 Besides, your screenshot shows that WordPress doesn't use caps as well.

Also, existing nomenclature uses "email", not "e-mail" (again, as seen in screenshot).

```suggestion
			$error = '<strong>Error</strong>: The username/email address or password is incorrect. Please try again.';
```

GaryJones (changes requested):

GaryJones (approved):

pauarge (approved):

PR #2019, opened 2021-03-08
Ensure hosts not empty before randomly picking one

### LINKED ISSUE #2018: PHP warning due to array_rand in class-search.php
In a VIP Go website where we are testing VIP Search, we are seeing an error like:

>array_rand(): Array is empty in /var/www/wp-content/mu-plugins/search/includes/classes/class-search.php on line 1116

See the line https://github.com/Automattic/vip-go-mu-plugins/blob/master/search/includes/classes/class-search.php#L1116

Using `array_rand` on an empty array was not an issue in PHP < 7.1, but is a warning since PHP 7.1 and will become a fatal error in PHP 8.

Considering the result of `Search::get_random_host()` is expected to be a valid host, is probably worth to return `null` when host is not an array _or_ when that array is empty.

Something like this:

```php
public function get_random_host( $hosts ) {
    if ( ! $hosts || ! is_array( $hosts ) ) {
        return null;
    }

    return $hosts[ array_rand( $hosts ) ];
}
```

gmazzap:
## Description

Check #2018 for context

### Search: Fix the issue raised when trying to randomly determine Elasticsearch host but no hosts available.

PHP `array_rand()` functions require a non-empty array, and it will raise a PHP warning in PHP 7.1-7.4 and a fatal error in PHP 8 in the case the array is empty. 

## Checklist

## Steps to Test
Testing the issue would involve the possibilty to control Elasticsearch server, so that's not easy to test. However, the fact that `array_rand` requires a non-empty array and that `Search::get_random_host()` currently doesn't ensure that is self-evident.

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -1109,8 +1109,8 @@ class Search {
 	 * Given a list of hosts, randomly select one for load balancing purposes.
 	 */
 	public function get_random_host( $hosts ) {
-		if ( ! is_array( $hosts ) ) {
-			return $hosts;
+		if ( ! $hosts || ! is_array( $hosts ) ) {
+			return null;
 		}
 
 		return $hosts[ array_rand( $hosts ) ];
--- a/tests/search/test-class-search.php
+++ b/tests/search/test-class-search.php
@@ -2751,6 +2751,18 @@ class Search_Test extends \WP_UnitTestCase {
 		$this->assertTrue( $es->is_protected_content_enabled() );
 	}
 
+    public function test__get_random_host_return_null_if_no_host() {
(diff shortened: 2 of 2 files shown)


pschoffer:
Perfect thanks @gmazzap . I tested this and it works as expected.

There is a linting issue, that I fixed in this PR back to your branch: https://github.com/gmazzap/vip-go-mu-plugins/pull/1

So after you merge that, I can take it from there.

gmazzap:
Merged there, thanks!

pschoffer (approved):
LGTM

pschoffer:
Great thanks for the PR. It is now deployed to staging environments and gets deployed everywhere within the next few days.

PR #2021, opened 2021-03-08
Update ElasticPress to latest

nickdaugherty:
## Description

Updates the ElasticPress submodule to the latest on Automattic/ElasticPress#develop.

### Search: Update ElasticPress

This updates the ElasticPress submodule to [da9e58e](https://github.com/Automattic/ElasticPress/commit/da9e58e74d08d2e9c4ff5ecdec43cb8144bc019e), which adds utility functions for getting and updating an index's settings.

## Checklist

## Steps to Test

1. Pull down PR
2. Update submodules - `git submodule update --init`
2. Perform a variety of searches and verify that ES requests/responses are working as expected
3. Get an indexable's settings - in `wp shell` - `\ElasticPress\Indexables::factory()->get( 'post' )->get_index_settings()`
4. The above should print out the current settings for the post index


netsuso (approved):
I've also tested `update_index_settings()`, which works, but I'm wondering why we have `flag_settings=true` in `get_index_settings()`, as it returns settings in a format that is not valid for `update_index_settings()`, which needs an unflattened data structure. But we can review this in the next iteration.

netsuso:
stacks-r1432

nickdaugherty:
> but I'm wondering why we have flag_settings=true in get_index_settings()

This is to match the format in the EP mapping template file and to make it easier to check individual keys (so we can specify that `index.settings.allocation.routing.include.dc` should equal a value, rather than trying to handle that as a deeply nested array. 👍 let's address the difference between getting and updating a little later.

PR #2023, opened 2021-03-11
Search: Add index setting automation

nickdaugherty:
## Description

When index settings are determined to be inconsistent with the desired settings, automatically update the (specific) settings to keep it in sync.

### Search: Add auto-updating of index settings

This change adds support for automatically updating an index's settings to match the desired state, enabled on only a subset of sites. This provides the ability to easily control index settings both per-site and across the VIP platform.

## Checklist

## Steps to Test

1. Add a `var_dump()` to `HealthJob::check_all_indexables_settings_health()` - `var_dump( $unhealthy_indexables );` - this is a super hack but this method sends alerts to Slack / IRC which doesn't work in local dev or sandbox (need a better way to do this)
1. `wp eval "(new \Automattic\VIP\Search\HealthJob(\Automattic\VIP\Search\Search::instance()))->check_all_indexables_settings_health();"`
1. Depending on your settings, this may or may not produce any output
1. If it doesn't, you can change a setting. The easiest way is to change the value returned from `Search::filter__ep_default_index_number_of_replicas()` from `2` to anything else. Alternatively you can implement `ep_default_index_number_of_replicas` to return some other number in your theme
1. In `lib/feature/class-feature.php` change `search_indexable_settings_auto_heal` from `0.1` to `1.0` to gurantee it will be triggered on 100% environments including yours.
1. Once there is a discrepancy between what VIP Search thinks the settings should be and what they are, you'll see it in the `var_dump()` output
1. Running `wp eval "(new \Automattic\VIP\Search\HealthJob(\Automattic\VIP\Search\Search::instance()))->check_all_indexables_settings_health();"` should result in no diff.

### DIFF
--- a/search/includes/classes/class-healthjob.php
+++ b/search/includes/classes/class-healthjob.php
@@ -27,20 +27,29 @@ class HealthJob {
 
 	/**
 	 * Instance of the Health class
-	 * 
+	 *
 	 * Useful for overriding in tests via dependency injection
 	 */
 	public $health;
 
 	/**
 	 * Instance of Search class
-	 * 
+	 *
 	 * Useful for overriding (dependency injection) for tests
 	 */
 	public $search;
 
+	/**
+	 * Instance of \ElasticPress\Indexables
+	 *
+	 * Useful for overriding (dependency injection) for tests
+	 */
+	public $indexables;
+
 	public function __construct( \Automattic\VIP\Search\Search $search ) {
+		$this->search = $search;
 		$this->health = new Health( $search );
+		$this->indexables = \ElasticPress\Indexables::factory();
 	}
 
 	/**
@@ -182,7 +191,7 @@ class HealthJob {
 		}
 	}
 
-	public function heal_index_settings( array $unhealthy_indexables ) {
+	public function heal_index_settings( $unhealthy_indexables ) {
 		// If the whole thing failed, error
 		if ( is_wp_error( $unhealthy_indexables ) ) {
 			$message = sprintf( 'Error while attempting to heal index settings for %s: %s', home_url(), $unhealthy_indexables->get_error_message() );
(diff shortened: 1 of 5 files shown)


search/includes/classes/class-healthjob.php
```diff
@@ -182,7 +191,7 @@ public function process_indexables_settings_health_results( $results ) {
 		}
 	}
 
-	public function heal_index_settings( array $unhealthy_indexables ) {
```
pschoffer: array here throws a runtime error if wp_error is passed in instead of an array.

pschoffer (approved):
@nickdaugherty I added few tests and tested the code a bit. It looks fine to me and works (or at least it didn't find change after the fix:), not sure how to test more).

If my changes looks goot to me then I think we can go ahead.

nickdaugherty:
Thanks for adding those tests @pschoffer , they're great! I added another one and tweaked a few things, it should be ready to go now.

For rollout, I think we should actually turn the percentage down to 0 for this (so it doesn't run anywhere) while we ramp up the monitoring to 100%. Once we have all the monitoring in place, we'll know if any sites have inconsistent settings which will help us understand how many sites will be affected when we turn this on. What do you think?

nickdaugherty:
@pschoffer I dialed down the auto healing to 0 so we can ship this with no impact. Would you be able to test it on a sandbox next week, then deploy to staging?

tests/search/includes/classes/test-class-health.php
```diff
@@ -596,6 +596,103 @@ public function test_get_index_settings_diff_for_indexable( $actual, $desired, $
...
+	}
+
+	/**
+	 * @dataProvider heal_index_settings_for_indexable_data
+	 */
+	public function test_heal_index_settings_for_indexable( $desired_settings, $options ) {
+		// Mock search and the versioning instance
+		$mock_search = $this->getMockBuilder( Search::class )
+			->setMethods( [] )
+			->getMock();
```
pschoffer, line 645:
```suggestion
		$mock_search = $this->createMock( Search::class );
```

pschoffer (dismissed):
Looks great. I added one test simplification that I **think** should work the same:)

tests/search/includes/classes/test-class-health.php
nickdaugherty: Nice, that's much cleaner.

pschoffer (approved):

nickdaugherty:
r1438-stacks

nickdaugherty:
Release v1.22.0

PR #2024, opened 2021-03-11
Exclude basic file tests from core run

pschoffer:
## Description

These tests don't really make sense for us as they check among other
things if .md/packache files are correct. Those are core developer aimed
tests.

The current failing one is that SECURITY.md didn't have some expected line in it.

### Development Pipeline Update

We have excluded some of the core WordPress tests from running as part of our mu-plugins build pipeline to avoid false failures.


netsuso (approved):
It makes sense, :shipit: !

pschoffer:
"Deployed", but no changes to prod.

PR #2025, opened 2021-03-11
Site Details: Include plugins that are enabled sitewide

netsuso:
## Description

We are currently fetching the list of enabled plugins with `get_option( 'active_plugins' )`, but this is not taking into account plugins that are "network active" on a multisite. This PR checks also `get_site_option( 'active_sitewide_plugins' )`, and adds an extra flag `enabled_sitewide` to indicate whether a plugin is enabled network wide.

### Include sitewide plugins in site details

We keep improving the site details data we are gathering from all sites to have more information about multiple runtime variables. This time, we are including the list of plugins enabled sitewide on multisite installations, as we were only considering plugins enabled locally for a site.

## Checklist

## Steps to Test

1. Check out PR on a Go Sandbox
1. Check the output in `/chroot/tmp/logstash.log` and ensure the new field `enabled_sitewide` is present


rinatkhaziev (changes requested):
A couple of notes: 

`sitewide` is a bit confusing since in WP terminology it's called "network-wide", I think to avoid confusion we should stick to `network` because it's a well-known term, `sitewide` is ambiguous, because it's called "Multisite network" that consists of subsites.

Also, what do you think about introducing a new field have it as `enabled_via=code|network|local`? 

This would marginally improve filtering (you would only ever need to filter by this field instead of the current two).

netsuso:
> `sitewide` is a bit confusing since in WP terminology it's called "network-wide", I think to avoid confusion we should stick to `network` because it's a well-known term, `sitewide` is ambiguous, because it's called "Multisite network" that consists of subsites.
> 
> Also, what do you think about introducing a new field have it as `enabled_via=code|network|local`?
> 
> This would marginally improve filtering (you would only ever need to filter by this field instead of the current two).

- Renamed "sitewide" to "network"
- After some discussion in Slack, added an `activated_by` field (ah, naming things... :D)

rinatkhaziev (approved):
:shipit:

rinatkhaziev (approved):

netsuso:
stacks-r1439

nickdaugherty:
Release v1.22.0

PR #2035, opened 2021-03-17
Remove Photon srcset filter when Photon is active 

htdat:
## Description

More details about this issue is here p9GBOq-1Tz-p2

One use case for Jetpack Photon (with this constant `WPCOM_VIP_USE_JETPACK_PHOTON`) is to overcome intermediate image links after an import: 

- from https://mikey.go-vip.co/wp-content/uploads/2020/06/chickens34-2-150x150.jpg
- to https://mikey.go-vip.co/wp-content/uploads/2020/06/chickens34-1.jpg?resize=150%2C150

However, it comes with the wrong crop issues for srcset attributes of new images (after the import) as Jetpack Photon just uses `w` parameter like: 

```
srcset="https://mikey.go-vip.co/wp-content/uploads/2020/06/chickens34-2.jpg?w=150 150w, https://mikey.go-vip.co/wp-content/uploads/2020/06/chickens34-2.jpg?w=300 300w, https://mikey.go-vip.co/wp-content/uploads/2020/06/chickens34-2.jpg?w=100 100w, https://mikey.go-vip.co/wp-content/uploads/2020/06/chickens34-2.jpg?zoom=3&amp;resize=150%2C150 450w" sizes="(max-width: 150px) 100vw, 150px">
``` 

**By introducing this PR, we make sure new images still have the expected srcset like this**: 

```
https://mikey.go-vip.co/wp-content/uploads/2020/06/chickens34.jpg?resize=150,150 150w, https://mikey.go-vip.co/wp-content/uploads/2020/06/chickens34.jpg?resize=300,300 300w, https://mikey.go-vip.co/wp-content/uploads/2020/06/chickens34.jpg?resize=100,100 100w
``` 

I introduced `WPCOM_VIP_USE_JETPACK_PHOTON_WITHOUT_SRCSET` to make sure it does not break any other intentions with Jetpack Photon srcset filter. 

## Checklist

## Steps to Test

1. Add this line into vip-config.php. 

```php
define( 'WPCOM_VIP_USE_JETPACK_PHOTON', true );
``` 
2. Add a new thumbnail image. 
3. Check the srcset attribute of this image, it's using `w` parameter - not what we want.
4. Add this line  in step 1.
```php
define( 'WPCOM_VIP_USE_JETPACK_PHOTON_WITHOUT_SRCSET', true );
```
5. Check srcset again, they should use `resize` parameter.

### DIFF
--- a/a8c-files.php
+++ b/a8c-files.php
@@ -174,7 +174,13 @@ class A8C_Files {
 
 		// If Photon isn't active, we need to init the necessary filters.
 		// This takes care of rewriting intermediate images for us.
-		Jetpack_Photon::instance();
+		$jetpack_photon = Jetpack_Photon::instance();
+
+		// Remove this filter from Photon so that we can use VIP filters for srcset
+		// Filter reference: https://github.com/Automattic/jetpack/blob/26db3e436ccca3e16a62d02d95e792a81f38a1e4/projects/plugins/jetpack/class.photon.php#L85
+		if (  defined( 'WPCOM_VIP_USE_JETPACK_PHOTON_WITHOUT_SRCSET' ) && true === WPCOM_VIP_USE_JETPACK_PHOTON_WITHOUT_SRCSET ) {
+			remove_filter( 'wp_calculate_image_srcset', array( $jetpack_photon, 'filter_srcset_array' ), 10, 5 );
+		}
 
 		// Jetpack_Photon's downsize filter doesn't run when is_admin(), so we need to fallback to the Go filters.
 		// This is temporary until Jetpack allows more easily running these filters for is_admin().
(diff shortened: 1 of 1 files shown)


mikeyarce:
This looks good @htdat but it raises some questions:

Instead of fixing the `srcset` with Jetpack Photon active, could we fix VIP Go's ability to return a proper image for something like `https://mikey.go-vip.co/wp-content/uploads/2020/06/chickens34-2-150x150.jpg`?  

Would that eliminate any possible need for Jetpack's Photon to be active?   I think we want to make Jetpack optional down the road so relying less on things like this might be better in the long run.

htdat:
> Instead of fixing the srcset with Jetpack Photon active, could we fix VIP Go's ability to return a proper image for something like https://mikey.go-vip.co/wp-content/uploads/2020/06/chickens34-2-150x150.jpg?

> Would that eliminate any possible need for Jetpack's Photon to be active? I think we want to make Jetpack optional down the road so relying less on things like this might be better in the long run.

Yes, this makes sense. This is more an effort to work around the issue we're having. 
The better way forward is to not reply on Jetpack Photon for these images `chickens34-2-150x150.jpg`. I will explore this path.

htdat:
I am going to close this as I am working on another approach on https://github.com/Automattic/vip-go-mu-plugins/pull/2050

PR #2037, opened 2021-03-19
Adds validate content interruption recovery

pschoffer:
## Description
* When processing validate_contents we store the id we are working of
* IF the process would end unexpectly we would recover that id and start
from where we left of.

### Plugin Updated: Search

Adds an interruption recovery mechanism to the validate content feature. 

## Checklist

## Steps to Test
The hardest part here is to simulate interruption.One way you can do this is by adding `exit()` in the validate-contnets loop.

1. add `exit()` call in `search/includes/classes/class-health.php` somewhere in the loop around line `323`
2. Run validate_contents with some start_id (so that it doesnt get interrupted on index 1 so that we can see it works)
`lando wp vip-search health validate-contents --start_post_id=5`
3. Clear parallel execution lock (as we killed the process it didnt clean up properly) - `lando wp eval "var_dump(delete_transient('vip_search_content_validation_lock'));";`
4. Remove the `exit()` line you added in the first step
5. Run validate_contents **without** start_id (so that we can see the interupted was used instead)
`lando wp vip-search health validate-contents`
6. In the logs you should see something like `Validating posts 5 - X` as oppose to `Validating posts 1 - X`.

### DIFF
--- a/search/includes/classes/class-health.php
+++ b/search/includes/classes/class-health.php
@@ -387,7 +387,7 @@ class Health {
 	}
 
 	public function update_validate_content_process( $next_post_id ) {
-		update_option( self::CONTENT_VALIDATION_PROCESS_OPTION, $next_post_id );
+		update_option( self::CONTENT_VALIDATION_PROCESS_OPTION, $next_post_id, false );
 	}
 
 	public function remove_validate_content_process() {
--- a/search/includes/classes/class-health.php
+++ b/search/includes/classes/class-health.php
@@ -259,11 +259,14 @@ class Health {
 	 */
 	public function validate_index_posts_content( $start_post_id, $last_post_id, $batch_size, $max_diff_size, $silent, $inspect, $do_not_heal, $force_parallel_execution = false ) {
 		$process_parallel_execution_lock = ! $force_parallel_execution;
+		// We only work with process if we can guarantee no parallel execution and user did't pick specific start_post_id to avoid unexpected overwriting of that.
+		$track_process = ( ! $start_post_id || 1 === $start_post_id ) && ! $force_parallel_execution;
+
 		if ( $process_parallel_execution_lock && $this->is_validate_content_ongoing() ) {
 			return new WP_Error( 'content_validation_already_ongoing', 'Content validation is already ongoing' );
 		}
 		$interrupted_start_post_id = $this->get_validate_content_abandoned_process();
-		if ( $process_parallel_execution_lock && $interrupted_start_post_id ) {
+		if ( $track_process && $interrupted_start_post_id ) {
 			$start_post_id = $interrupted_start_post_id;
 		}
 
@@ -311,7 +314,8 @@ class Health {
 		do {
 			if ( $process_parallel_execution_lock ) {
 				$this->set_validate_content_lock();
-				// We only work with process if we can guarantee no parallel execution
+			}
+			if ( $track_process ) {
 				$this->update_validate_content_process( $start_post_id );
 			}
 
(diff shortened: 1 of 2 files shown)


search/includes/classes/class-health.php
```diff
@@ -361,11 +369,31 @@ public function validate_index_posts_content( $start_post_id, $last_post_id, $ba
...
+	 * to potentially pick-up where we left of on long running validate contents that got interrupted.
+	 *
+	 * @return int|bool returns the ID of the first post in a batch that was process when process was updated OR false if no such values is saved.
+	 */
+	public function get_validate_content_abandoned_process() {
+		return get_option( self::CONTENT_VALIDATION_PROCESS_OPTION );
+	}
+
+	public function update_validate_content_process( $next_post_id ) {
+		update_option( self::CONTENT_VALIDATION_PROCESS_OPTION, $next_post_id );
```
nickdaugherty, line 390: We'll need to pass `false` as the 3rd param to prevent this option from being autoloaded.
```suggestion
		update_option( self::CONTENT_VALIDATION_PROCESS_OPTION, $next_post_id, false );
```

search/includes/classes/class-health.php
```diff
@@ -261,6 +262,10 @@ public function validate_index_posts_content( $start_post_id, $last_post_id, $ba
 		if ( $process_parallel_execution_lock && $this->is_validate_content_ongoing() ) {
 			return new WP_Error( 'content_validation_already_ongoing', 'Content validation is already ongoing' );
 		}
+		$interrupted_start_post_id = $this->get_validate_content_abandoned_process();
+		if ( $process_parallel_execution_lock && $interrupted_start_post_id ) {
```
nickdaugherty, line 266: Should we be checking if a `$start_post_id` was actually passed into `validate_index_posts_content()`? Right now you could call the command with this both `--start_post_id` and `--force-parallel-execution` and it would ignore your `--start_post_id` if processing had been interrupted. Is that the desired outcome, or do we want to allow it to be overridden from the command?

nickdaugherty (changes requested):
This is awesome! I just noted two things.

search/includes/classes/class-health.php
pschoffer: Hmm the naming isn't great here really... `$process_parallel_execution_lock = ! $force_parallel_execution`. I didn't want to have the negation everywhere to not be confusing so I tried to introduce this, I thought descriptive, variable.

So actually when you send in `--force-parallel-execution` the `$process_parallel_execution_lock === false` and then this `if` will be false and the `$interrupted_start_post_id` will be ignored.

However when you **don't** send in `--force-parallel-execution`, the  `$interrupted_start_post_id` will take precedence (and `--start_post_id` would be ignored).

I did think about skipping process tracking if `$start_post_id > 1` or `$last_post_id` is send in. But then I thought it would be a valid usecase... say it always dies on post id 350001, so you would say start from 340000 and last_id 350000 (but track and clear process so that we could do more work later). And in general it felt like there might be more complicated combinations of parameters when to track and when not.

So I thought, let's keep it simple. Sending `--force_parallel_execution` ignore the lock and ignores the tracking. Hmm maybe the parameter should be just `--force` then? What do you think?

nickdaugherty: Hmm interesting - IMO the `--start_post_id` flag should take precedence - if it's passed, respect the user's wishes and start there. It would be pretty confusing to pass in the post id you want to start at and have it actually start at another seemingly random location. At least if we're overriding the user's choice, we need to tell them, and tell them how to stop that behavior (by using `--force_parallel_execution`.

pschoffer: I added the change discussed in slack.

nickdaugherty (approved):
LGTM

nickdaugherty:
Release v1.24.0

PR #2040, opened 2021-03-23
Add validate content cron

pschoffer:
### Plugin Updated: Search

Adds automated cron task that will run once a week and will validate all the post contents by comparing data in database against the documents in elasticsearch. It will automatically fix any issues it finds.

## Checklist

- [`n/a`] This change has relevant unit tests (if applicable).

## Steps to Test
1. There are some hacking needed first - in `class-healthjob.php:152` add `var_dump($results);` - to see stuff
1. In  `class-healthjob.php:is_enabled` add `return true;` right at the beginning (to enable it outside of prod)
1. In `lib/feature/class-feature.php` change `'search_content_validation_and_auto_heal_cron_job' => 0.1` to `'search_content_validation_and_auto_heal_cron_job' => 1`
1. `wp eval "var_dump(do_action('vip_search_health_validate_content'));"`
1. Now introduce some issue (like delete post from DB directly)
1. `wp eval "var_dump(do_action('vip_search_health_validate_content'));"` - should show the diff
1. Running `wp eval "var_dump(do_action('vip_search_health_validate_content'));"` again should show no difference as the first run healed the issue.

### DIFF
--- a/search/includes/classes/class-health.php
+++ b/search/includes/classes/class-health.php
@@ -255,9 +255,48 @@ class Health {
 	/**
 	 * Validate DB and ES index post content
 	 *
+	 * ## OPTIONS
+	 *
+	 * [inspect]
+	 * : Optional gives more verbose output for index inconsistencies
+	 *
+	 * [start_post_id=<int>]
+	 * : Optional starting post id (defaults to 1)
+	 *
+	 * [last_post_id=<int>]
+	 * : Optional last post id to check
+	 *
+	 * [batch_size=<int>]
+	 * : Optional batch size
+	 *
+	 * [max_diff_size=<int>]
+	 * : Optional max count of diff before exiting
+	 *
+	 * [do_not_heal]
+	 * : Optional Don't try to correct inconsistencies
+	 *
+	 * [silent]
+	 * : Optional silences all non-error output except for the final results
+	 *
+	 * [force_parallel_execution]
+	 * : Optional Force execution even if the process is already ongoing
+	 *
+	 *
+	 * @param array $options list of options
+	 *
+	 *
 	 * @return array Array containing counts and ids of posts with inconsistent content
 	 */
-	public function validate_index_posts_content( $start_post_id, $last_post_id, $batch_size, $max_diff_size, $silent, $inspect, $do_not_heal, $force_parallel_execution = false ) {
+	public function validate_index_posts_content( $options ) {
+		$start_post_id = $options['start_post_id'] ?? 1;
+		$last_post_id = $options['last_post_id'] ?? null;
+		$batch_size = $options['batch_size'] ?? null;
+		$max_diff_size = $options['max_diff_size'] ?? null;
+		$silent = isset( $options['silent'] );
+		$inspect = isset( $options['inspect'] );
+		$do_not_heal = isset( $options['do_not_heal'] );
+		$force_parallel_execution = isset( $options['force_parallel_execution'] );
+
 		$process_parallel_execution_lock = ! $force_parallel_execution;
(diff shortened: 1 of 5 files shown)


pschoffer:
This is first draft. Do we want to roll it out slowly. I.e. using the feature functionality or some other way?

search/includes/classes/class-healthjob.php
```diff
@@ -115,6 +124,42 @@ public function filter_cron_schedules( $schedule ) {
 		return $schedule;
 	}
 
+	/**
+	 * Check index health
+	 */
+	public function validate_contents() {
+		// Check if job has been disabled
+		if ( ! $this->is_enabled() ) {
+			$this->disable_job();
```
nickdaugherty, line 133: We need to update `disable_job()` to also disable the new event.

search/includes/classes/class-healthjob.php
```diff
@@ -115,6 +124,42 @@ public function filter_cron_schedules( $schedule ) {
...
+		}
+
+		$results = $this->health->validate_index_posts_content(
+			1,
+			null,
+			null,
+			null,
+			true, // silent
+			false,
+			false,
```
nickdaugherty, line 150: 😬 the number of args to this function is too high...I like that there's a comment on one, do you think we should comment all of them here?

nickdaugherty (changes requested):
This looks good, just left a few notes. For rollout, I agree that we should enable this on just a subset of sites (via the Feature class).

With the `weekly` period, if the job fails half way, does that mean we'd have to wait another week for it to start up where it left off? Is `daily` (or something else) a better interval?

search/includes/classes/class-healthjob.php
pschoffer: Good point. I switched the signature to use assoc array instead.

search/includes/classes/class-healthjob.php
pschoffer: Nice catch! Done.

pschoffer:
Thanks @nickdaugherty for the feedback. I tested this manually today. I am not sure if unittests would add much value here. The validate_contents functionality is already tested elsewhere and this change is introducing just some cron WP operations + slack messages that are difficult to test anyways. I might be wrong though. What do you think?

For the interval, I thought that daily would be too often as from my understanding this might run more than a day on some sites, but then again it should be able to handle that with the parallel running lock. I figured we could start with weekly and bump it up if needed. We could also add a filter for that so that we/clients can tweak the interval per case basis. What do you think?

search/includes/classes/class-health.php
```diff
@@ -255,9 +255,48 @@ public static function validate_index_posts_count( $options = array() ) {
...
-	public function validate_index_posts_content( $start_post_id, $last_post_id, $batch_size, $max_diff_size, $silent, $inspect, $do_not_heal, $force_parallel_execution = false ) {
+	public function validate_index_posts_content( $options ) {
+		$start_post_id = $options['start_post_id'] ?? 1;
+		$last_post_id = $options['last_post_id'] ?? null;
+		$batch_size = $options['batch_size'] ?? null;
+		$max_diff_size = $options['max_diff_size'] ?? null;
+		$silent = isset( $options['silent'] );
+		$inspect = isset( $options['inspect'] );
+		$do_not_heal = isset( $options['do_not_heal'] );
+		$force_parallel_execution = isset( $options['force_parallel_execution'] );
```
nickdaugherty, line 298: Doesn't need changed, but wanted to point out for the future that WP has a utility function that can make dealing with defaults easy - `wp_parse_args()` ([#](https://developer.wordpress.org/reference/functions/wp_parse_args/)).

nickdaugherty (approved):
Awesome I think this is ready. Good call on keeping the interval at `weekly` to start.

nickdaugherty:
Release v1.24.0

PR #2046, opened 2021-03-26
Disable search if we don't have ES connectivity

pschoffer:
### Plugin Updated: Search

Disable Search if connectivity to elasticsearch isn't configured. This helps us to avoid having false positive alerts and related issues.

## Checklist

- [`n/a/`] This change has relevant unit tests (if applicable).
- [`n/a`] This change has relevant documentation additions / updates (if applicable).

## Steps to Test
1. Remove `VIP_ELASTICSEARCH_ENDPOINTS` or `VIP_ELASTICSEARCH_USERNAME` or `VIP_ELASTICSEARCH_PASSWORD`
1. Running `wp vip-search` will result in not known command.


tests/search/test-class-search.php
```diff
@@ -2830,6 +2830,41 @@ public function test__maybe_enable_ep_query_logging_debug_bar_and_qm_enabled() {
...
+		$this->assertFalse( $result );
+	}
+
+	public function test__are_es_constants_defined__no_username() {
+		define( 'VIP_ELASTICSEARCH_ENDPOINTS', [] );
+		define( 'VIP_ELASTICSEARCH_PASSWORD', 'bar' );
+
+		$result = \Automattic\VIP\Search\Search::are_es_constants_defined();
+
+		$this->assertTrue( $result );
```
rinatkhaziev, line 2865:
```suggestion
		$this->assertFalse( $result );
```

rinatkhaziev: I think there's a typo here, that causes test to fail

search/includes/classes/class-search.php
```diff
@@ -112,6 +112,16 @@ public function init() {
 		$this->setup_regular_stat_collection();
 	}
 
+	/**
+	 * Check if the constants needed for ElasticSearch connection are defined.
+	 */
+	public static function are_es_constants_defined() {
+		$endpoints_defined = defined( 'VIP_ELASTICSEARCH_ENDPOINTS' );
```
nickdaugherty, line 119: We should also check that this is a non-empty array.

tests/search/test-class-search.php
pschoffer: Perfect thanks!. Fixed.

search/includes/classes/class-search.php
pschoffer: Good point!

000-vip-init.php
```diff
@@ -193,17 +193,19 @@
 // Load elasticsearch helpers
 // Warning: Site Details depends on the existence of class Search.
 // If this changes in the future, please ensure that details for search are correctly extracted
-if ( ( defined( 'USE_VIP_ELASTICSEARCH' ) && USE_VIP_ELASTICSEARCH ) || // legacy constant name
-	defined( 'VIP_ENABLE_VIP_SEARCH' ) && true === VIP_ENABLE_VIP_SEARCH ) {
+$search_enabled = ( defined( 'USE_VIP_ELASTICSEARCH' ) && USE_VIP_ELASTICSEARCH ) || defined( 'VIP_ENABLE_VIP_SEARCH' ) && true === VIP_ENABLE_VIP_SEARCH;
```
nickdaugherty, line 196: @pschoffer This will introduce a new `$search_enabled` global variable which is suboptimal. Even though it's a longer line, it's probably better to leave the conditional as-is.

search/search.php
```diff
@@ -20,4 +20,6 @@
 
 require_once __DIR__ . '/includes/classes/class-search.php';
 
-do_action( 'vip_search_loaded' );
+if ( \Automattic\VIP\Search\Search::are_es_constants_defined() ) {
```
nickdaugherty, line 23: Now that we're having to repeat this "loading" logic in multiple places, it may make sense to move the logic from 000-vip-init.php into here? We could leave the conditional loading of search/search.php, but bring the JP filtering and instance code into here so we have a single conditional?

nickdaugherty (changes requested):
Nice - I noticed two things when looking at this, let me know what you think.

search/search.php
pschoffer: Nice that is good idea, I couldn't figure out how to do it:)

000-vip-init.php
```diff
@@ -193,18 +193,8 @@
 // Load elasticsearch helpers
 // Warning: Site Details depends on the existence of class Search.
 // If this changes in the future, please ensure that details for search are correctly extracted
-if ( ( defined( 'USE_VIP_ELASTICSEARCH' ) && USE_VIP_ELASTICSEARCH ) || // legacy constant name
-	defined( 'VIP_ENABLE_VIP_SEARCH' ) && true === VIP_ENABLE_VIP_SEARCH ) {
+if ( ( defined( 'USE_VIP_ELASTICSEARCH' ) && USE_VIP_ELASTICSEARCH ) || define( 'VIP_ENABLE_VIP_SEARCH' ) && true === VIP_ENABLE_VIP_SEARCH ) {
```
nickdaugherty, line 196: Can we revert this to the original code (which was 2 lines)? This version has a typo (`define` vs `defined`).

search/search.php
```diff
@@ -20,4 +20,15 @@
...
+	do_action( 'vip_search_loaded' );
```
nickdaugherty, line 24: The `do_action()` should actually come at the end, after we've initialized the instance (which is how it worked previously).

nickdaugherty (changes requested):
This is awesome, I love that we've moved logic out of 000-vip-init.php and have it grouped with search code where it belongs. 🥳 I left two easy comments, then this should be ready. We'll definitely want to test this thoroughly 😄

search/search.php
pschoffer: Make sense, that is better.

However the original implementation did trigger that action before the instance if I understand php code correctly?

The `000-vip-init.php` would run  `require_once( __DIR__ . '/search/search.php' );` which would run the whole `/search/search.php` file including `do_action()` statement. After which the `000-vip-init.php` would continue to setup the instance, correct?

nickdaugherty: Ah crap you're right, my bad. Not sure what I was thinking, long day yesterday 😅

nickdaugherty: I did a search and didn't find anything using this action, so should be perfectly safe to move its location slightly like this - makes more sense to do it after instantiation, which is possible now that we moved this code.

search/includes/classes/class-search.php
```diff
@@ -117,6 +117,16 @@ public function init() {
 		$this->setup_regular_stat_collection();
 	}
 
+	/**
+	 * Check if the constants needed for ElasticSearch connection are defined.
+	 */
```
pauarge, line 122: Should we add the return type on the PHPDoc?

pschoffer: Sure, done.

pauarge (approved):

PR #2047, opened 2021-03-26
ALLOPTIONS limit alert/block - account for the fact that pecl-memcache client automagically compresses the values before sending it to the server

rinatkhaziev:
## Description

We're triggering the alert when alloptions size is over 800KB and block the site from loading when it's over 1MB. 

Unfortunately, that approach didn't account for the fact that Memcache client automagically compresses values of they're over 20KB and the savings from storing the compressed value are more than 20% (it's a setting for Memcached pool, we use the default ones and they never change, that's why they're hardcoded.)

The changes may look cryptic, that's why there are so many code comments with links.

### Improvement: Options size monitoring

We have systems in place that alert us when the size of the options grows too big to fit into a Memcached server entry to prevent increased pressure on the Memcached server and limit potential performance degradation. 

We made changes to alerting logic to prevent false positives. 

## Checklist

## Steps to Test

1. Check out PR.
1. Comment out https://github.com/Automattic/vip-go-mu-plugins/blob/6fe27e5e6619484b947d394622cfc77b6b62bb14/alloptions-limit.php#L15-L17
1. Define `define( 'VIP_ALLOPTIONS_NOTIFY_EMAIL', 'some@email.test' );`
1. Spawn a `wp shell`
1. Add a ridiculously large option: `add_option( 'test_compression', range( 1, 1000000 );`
1. Run `wpcom_vip_sanity_check_alloptions()`
1. Verify in mailhog that you got the alert.

### DIFF
--- a/alloptions-limit.php
+++ b/alloptions-limit.php
@@ -17,16 +17,34 @@ function wpcom_vip_sanity_check_alloptions() {
 	}
 
 	// Warn should *always* be =< die
-	$alloptions_size_warn  =  800000;
-	$alloptions_size_die   = 1000000; // 1000000 ~ 1MB, too big for memcache
+	$alloptions_size_warn = 800000;
+	// 950000 is slightly less than ~ 1MB, which is the Memcached entry limit.
+	// The purpose of this limit is to safe-guard against a barrage of requests with cache sets for values that are too large
+	// Because WP would keep trying to set the data to Memcached, potentially resulting in Memcached (and site's) performance degradation.
+	$alloptions_size_die  = 950000;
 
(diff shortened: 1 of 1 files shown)


netsuso (approved):
Nice change! I've tested it with this range:

`add_option( 'test_compression', range( 1, 500000 ) );`

Which is a sweet spot as it's > 1 MB uncompressed, but < 1 MB compressed, so the difference can be appreciated between the old code and this PR.

The 20% limit seems too conservative in general, as this will normally be text or similar (json, serialized PHP data), that is compressed heavily.

PR #2049, opened 2021-03-29
Introduce vip-woocommerce module

WPprodigy:
## Description

Add a module specifically for WooCommerce logic. Is light right now,  but with anticipation for more to follow (and likely to split into more specific files at that point). Notably we'll want to prevent as many WC-specific things from loading as possible when WC isn't on the site.

### WC Logging

For now just covering the logging problem. By default, WC logs into `/uploads/wc-logs/` and logging can occur for various things like payment gateway debugging. The VIP filesystem isn't really intended for constant file writes like this though. So when logging is enabled, it will now use the DB Logger class. And if a plugin happens to (incorrectly) use the file logger specifically, it will continue to go into /tmp/ as [implemented here](https://github.com/Automattic/vip-go-mu-plugins/pull/1642). [More info](https://developer.woocommerce.com/2017/01/26/improved-logging-in-woocommerce-2-7/) on this API.

Documentation changes to follow, with recommendations to leave debug logging off until it's needed.

### WooCommerce Compatibility: Logging

Increase compatibility with WooCommerce's logging system.

- Add a module for WooCommerce-specific logic.
- Use the DB Logger by default, w/ some safeguards in place.

## Checklist


rinatkhaziev:
@WPprodigy in what ways do you think we can detect the presence of Woo early on? I see you opted for the `plugins_loaded` check, but are there any others? 
Thinking in terms of not loading this altogether when Woo is not present.

WPprodigy:
Plugins don't really have to follow strict folder-naming conventions, so it can be a bit tricky to detect earlier than it loading itself.

The constants are a bit of a one-off situation where we need to do them before WC has the chance to load, since this happens right away: https://github.com/woocommerce/woocommerce/blob/trunk/includes/class-woocommerce.php#L166.

Aside from that though, all other logic I can think of currently think of can be hooked in later up after WC is initialized. At which point a cheap `class_exists()` is both safe and performant.

WPprodigy:
Leaving the `file_exists` checks for now since it's still going to be deployed to vipd, but can be removed in the future.

pauarge (approved):
LGTM

rinatkhaziev:
r1833-stacks

PR #2050, opened 2021-03-30
Fix: Handle intermediate images without Jetpack Photon

htdat:
## Description 

This is an improvement from this issue https://github.com/Automattic/vip-go-mu-plugins/pull/2035#issuecomment-803841358

Here is what it will do: 

- If `VIP_INTERMEDIATE_IMAGES_FIX` is set to `true` in vip-config.php, it will trigger this class. Otherwise, it does not work. 
- Hook to `the_content`.
- Use regex and look for all image tags and their `src` values.
- Loop through these images: for ones with intermediate patterns, replace them with the resize parameter. 

## To test 

- Pull this PR or apply this patch.
- Add this line into vip-config.php to trigger the script. 

```php
define( 'VIP_INTERMEDIATE_IMAGES_FIX', true );
``` 

- In your VIP Go site, add an image in a testing post and its file name is using the intermediate patterns like this: https://mikey.go-vip.co/wp-content/uploads/2020/06/chickens34-150x150.jpg
- Check in the HTML source of this post in the front-end you will see something like this https://mikey.go-vip.co/wp-content/uploads/2020/06/chickens34.jpg?resize=150%2C150

## Next steps 

This is more a PoC (Proof of Concept). If it sounds good, I will need to make changes in this file https://github.com/Automattic/vip-go-mu-plugins/blob/518ca1a3f25680ffabf5919480b40d78458beb2a/a8c-files.php 

Changes include (but of course, I will need to think more about the backward compatibility too): 

- Remove or migrate the Photon part. 
- Class `A8C_Files_Utils` is used for Photon part so we may consider removing it too. 
- Maybe more or some other considerations?


mikeyarce:
Thanks for this, @htdat!    

This works for me for showing the intermediate images.  What about making this a CLI command "fixer" to do a replace on the site as an option?  that way we're not hooking into `the_content` all the time?  

I could see it having a dry run to see an output of images that will be changed and then a final command to replace the images in the content with the new `resize` format? 

What are your thoughts on PRing this straight into the `a8c-files`  plugin?

htdat:
> This works for me for showing the intermediate images. What about making this a CLI command "fixer" to do a replace on the site as an option? that way we're not hooking into the_content all the time?

> I could see it having a dry run to see an output of images that will be changed and then a final command to replace the images in the content with the new resize format?

Yes, this is doable. I will continue iterating it. 

> What are your thoughts on PRing this straight into the a8c-files plugin?

I will look into this. As here is more a Proof of Concept, I do not want to make changes to a8c-files.php directly. 
But even so, currently, a8c-files.php has two classes `A8C_Files_Utils` and `A8C_Files`, call [other classes](https://github.com/Automattic/vip-go-mu-plugins/blob/89de8b7708ed6c8a623b99df02c88cec47006f4e/a8c-files.php#L35-L37), and it mainly solves other issues. 

Therefore, I would create another mu-plugin/file for this work. What do you think?

PR #2055, opened 2021-03-31
Search: Temporarily disable index setting auto healing in 1 test

nickdaugherty:
## Description

Needs to be disabled because the notice from https://github.com/10up/ElasticPress/issues/2155 breaks the test. Can't use `$this->expectException()` b/c the notice is converted into an Exception that aborts the rest of the test

## Checklist

- n/a This change works and has been tested on a Go sandbox.
- n/a I've created a changelog description that aligns with the provided examples. 

## Steps to Test

This is a unit-test-only change, no manual testing (or deployment) needed.


nickdaugherty:
This is no longer needed after @pschoffer's fix in https://github.com/Automattic/ElasticPress/pull/85 (and the submodule update in https://github.com/Automattic/vip-go-mu-plugins/pull/2060)

PR #2057, opened 2021-04-01
Allow VIP File System uploads as email attachments

trepmal:
## Description
Extends PHPMailer to allow replacing of the `isPermittedPath()` to accommodate VIP File System

## Changelog Description

tbd

## Checklist

## Steps to Test

```
wp> $media_id = get_posts( [ 'post_type' => 'attachment', 'posts_per_page' => 1 ] )[0]->ID;
int(..)
wp> wp_mail( 'EMAIL_ADDRESS', 'Attachment test', 'Lorem Ipsum', '', [ get_attached_file( $media_id ) ] );
bool(true)
```

_Save yourself a testing headache and verify the media item actually exists._

**Before patch:** see that _Attachment test_ has arrived, without an attachment. 👎 
_Note: `wp_mail()` still sends, hence `true`_

**After patch:** see that _Attachment test_ has arrived, with an attachment. 👍


trepmal:
I'm not sure if this would be better done as a Core hack to ensure that other plugins extending PHPMailer pick up the same fix.

nickdaugherty:
Slack discussion here: https://a8c.slack.com/archives/C013N7TDZAT/p1617645073143900

It's worth thinkingn through a core patch to update PHPMailer to support the `vip://` protocol, as some plugins / sites subclass PHPMailer, which would bypass our override here. The downside there of course is maintaining a core patch indefinitely.

vip-mail.php
```diff
@@ -8,6 +8,33 @@
...
+	 *
+	 * Used to reject URLs and phar files from functions that access local file paths,
+	 * such as addAttachment. Allows VIP File System's `vip` protocol.
+	 *
+	 * @param string $path A relative or absolute path to a file
+	 *
+	 * @return bool
+	 */
+	protected static function isPermittedPath( $path ) {
+		if ( 0 === strpos( $path, 'vip://' ) ) {
```
rinatkhaziev, line 28:
```suggestion
		if ( 0 === strpos( $path, 'vip://wp-content/uploads' ) ) {
```
Stream Wrapper only supports paths that are starting with `wp-content/uploads`, let's match that here.

rinatkhaziev (changes requested):
Looks good, I left one minor suggestion to better match the behavior between Stream Wrapper and this code.

rinatkhaziev (approved):

mjangda:
@trepmal @rinatkhaziev can we add some tests for this please?

rinatkhaziev:
@mjangda I'm going to do manual testing now

rinatkhaziev:
I've tested all basic conditions: no file, file from File service, file from local filesystem, and a not-existing file, everything seems to be operating smoothly.

But I'll open a feature flag PR and then we can slowly roll it out.

PR #2058, opened 2021-04-01
Add new subsite wp-blogs-update CLI command

nlovrencic:
## Description
This PR adds a `wp vip subsites wp-blogs-update` command, purpose of which is updating the `domain` and `path` of wp_blogs. We are adding this command because Search and Replace doesn't cover wp_blogs in a good way:

### wp_blogs search and replace problem

Consider the following wp_blogs structure:

```
+---------+----------+----------+
| blog_id | domain   | path     |
+---------+----------+----------+
|       1 | example.com | subsite1 |
+---------+----------+----------+
|       2 | example.com | subsite2 |
+---------+----------+----------+
```

If we want to change subsite with blog_id 2 from `example.com/subsite2` to `subsite2.example.com`, we cannot just simply use search and replace because:
- if we do search and replace from `example.com/subsite2` to `subsite2.example.com` nothing will happen
- if we do search and replace from `example.com` to `subsite2.example.com` all entries will get changed

The above could also be achieved by using existing `wp db` CLI commands, but we wanted to add this as a `wp vip` CLI command, in case we'll need to extend the functionality or add more commands.

### New VIP CLI command: Update domain and path of a subsite in wp_blogs

We have added a new `wp vip subsites update-wp-blogs` command, which updates the domain and path values for a subsite in wp_blogs.

## Checklist

- n\a This change has relevant unit tests (if applicable).

## Steps to Test

1. Check out PR.
2. Setup a test multisite environment
3. Run the CLI command: `wp vip subsites update-wp-blogs <blog-id> <domain> --subsite-path=<path>`
4. Verify that `domain`, `path` and `last_updated` values are changed in wp_blogs for the provided blog-id


wp-cli/vip-wp-blogs.php
```diff
@@ -0,0 +1,64 @@
...
+	 *
+	 * @subcommand update-wp-blogs <blog-id> <domain>
+	 * @sypnosis [--subsite-path=<path>]
+	 */
+	public function update_wp_blogs( $args, $assoc_args ) {
+        global $wpdb;
+
+		WP_CLI::line( 'Updating wp_blogs...' );
+
+		$path = WP_CLI\Utils\get_flag_value( $assoc_args, 'subsite-path');
```
pschoffer, line 30: Should the default paht be `/` ?

nlovrencic: Good catch!

pschoffer:
Looks good to me. Could you please add a few unittests to cover the new class. Mostly about that path defaulting I guess.

You can then run it using `./bin/phpunit-docker.sh tests/your-test-path` as running the whole test suite takes long time.

pschoffer:
Turns out testing CLI commands is not trivial. So I tested manualy. It works well.

TIL you can do `wp db query "SELECT * FROM wp_blogs;"` to fetch stuff from DB. Nice!

pschoffer (approved):

pschoffer (changes requested):
Oh one more thing.

If you run this on non-multisite it results:
```
Updating wp_blogs...

ARGUMENTS
* blog ID: 1
* domain: test
* path: /
* new subsite URL: test/

WordPress database error Table 'wordpress.wp_blogs' doesn't exist for query SHOW FULL COLUMNS FROM `wp_blogs` made by include('phar:///usr/local/binphp/boot-phar.php'), include('phar:///usr/local/binvendor/wp-cli/wp-cli/php/wp-cli.php'), WP_CLI\bootstrap, WP_CLI\Bootstrap\LaunchRunner->process, WP_CLI\Runner->start, WP_CLI\Runner->run_command_and_exit, WP_CLI\Runner->run_command, WP_CLI\Dispatcher\Subcommand->invoke, call_user_func, WP_CLI\Dispatcher\CommandFactory::WP_CLI\Dispatcher\{closure}, call_user_func, VIP_Subsites_CLI_Command->update_wp_blogs
Error: Table 'wordpress.wp_blogs' doesn't exist
```

pschoffer:
We could add a static function to the `VIP_Subsites_CLI_Command` something like:
```
public static function load_commands() {
   if ( is multisite ) {
       WP_CLI::add_command(....);
   }
}
```

I am not sure how to do `is multisite` :) But I'll try to find out.

pschoffer (approved):
Perfect works great

nickdaugherty:
Release v1.32.0

PR #2067, opened 2021-04-08
Require current password when changing it

pauarge:
## Description

This PR introduces the requirement to input the current user's password when changing it on its profile page in `wp-admin`.  This has been implemented using WP actions. There is some JS code that is being used to reposition the elements in the form.

The new input is only shown when the user clicks `Set New Password`:

Screen Shot 2021-04-08 at 4 56 30 PM

Screen Shot 2021-04-08 at 4 56 23 PM

If the user doesn't input a password, an error is thrown:

Screen Shot 2021-04-08 at 5 01 05 PM

If the password is not correct, there is a different error:

Screen Shot 2021-04-08 at 5 01 33 PM

### Require current password for changing it

When changing its own password in the WP admin, the user will be prompted with their current password.

## Checklist

## Steps to Test

1. Check out PR.
2. Go to `wp-admin` > `Users` > `Profile`
3. Change password by clicking `Set New Password`


001-core/password.php
```diff
@@ -0,0 +1,63 @@
...
+
+function add_current_password_field() { ?>
+	<table id="nojs-current-pass" class="form-table" role="presentation">
+		<tr>
+			<th><label for="current_pass"><?php _e( 'Current Password' ); ?></label></th>
+			<td>
+				<div id="current-password-confirm">
+					<input type="password" name="current_pass" id="current_pass" class="regular-text" value="" autocomplete="off" />
+					<p class="description">
+						<?php _e( 'Please, type your <strong>current password</strong> to change it for a new one' ); ?>.
```
nickdaugherty, line 33:
```suggestion
						<?php _e( 'Please type your <strong>current password</strong> to update it' ); ?>.
```
Just a small wording tweak

001-core/password.php
```diff
@@ -0,0 +1,63 @@
...
+add_action( 'admin_head', 'reposition_current_password_field' );
+
+function reposition_current_password_field() {
+	$screen = get_current_screen();
+	if ( 'profile' != $screen->id ) {
+		return;
+	} 
+	?>
+
+	<script type="text/javascript">
```
nickdaugherty, line 52: Can we move this to a separate `.js` file then enqueue it with `wp_enqueue_script()`? Using WP's enqueue system provides more flexibility (for example, enqueued files get concatenated together on VIP).

001-core/password.php
```diff
@@ -0,0 +1,63 @@
...
+
+function validate_current_password( $errors, $update, $user ) {
+	check_admin_referer( 'update-user_' . $user->ID );
+
+	if ( ! isset( $_POST['pass1'] ) || empty( $_POST['pass1'] ) || ! $update ) {
+		return;
+	}
+
+	if ( ! isset( $_POST['current_pass'] ) || empty( $_POST['current_pass'] ) ) {
+		$errors->add( 'wrong_current_password', __( '<strong>ERROR</strong>: Please enter your current password.' ), array( 'form-field' => 'current_pass' ) );
```
nickdaugherty, line 13: Do we want to have a different error code for this condition? Right now it's the same as when the pw doesn't match, which could be confusing.

tests/001-core/test-password.php
```diff
@@ -0,0 +1,73 @@
+<?php
+
+class Current_Password_Change extends WP_UnitTestCase {
```
nickdaugherty, line 3: Minor - it would be good to name this class so that it's obvious it's a test case.
```suggestion
class Current_Password_Change_Test extends WP_UnitTestCase {
```

001-core/password.php
nickdaugherty: For consistency with WP Core's error messages, can we use `Error`?
```suggestion
		$errors->add( 'wrong_current_password', __( '<strong>Error</strong>: Please enter your current password.' ), array( 'form-field' => 'current_pass' ) );
```
Here's a screenshot of a different profile form error:

Screen Shot 2021-04-08 at 3 16 44 PM

001-core/password.php
```diff
@@ -0,0 +1,63 @@
...
+		return;
+	}
+
+	$error = wp_authenticate( $user->user_login, $_POST['current_pass'] );
+	if ( is_wp_error( $error ) ) {
+		$errors->add( 'wrong_current_password', __( '<strong>ERROR</strong>: The entered current password is not correct.' ), array( 'form-field' => 'current_pass' ) );
```
nickdaugherty, line 19:
```suggestion
		$errors->add( 'wrong_current_password', __( '<strong>Error</strong>: The entered current password is not correct.' ), array( 'form-field' => 'current_pass' ) );
```

nickdaugherty (changes requested):
This is really cool @pauarge! I've left a few comments inline, let me know if anything's not clear.

I noticed a bug - when you try to add (and probably edit) a _different_ user (not yourself), the nonce validation breaks the form. We'll need to be sure our stuff only runs when you're editing your own profile.

Additionally, we have a Feature flag library (`\Automattic\VIP\Feature`) - can we put this functionality behind a feature flag gate (can find examples in the code base) so we can gradually roll this out?

001-core/password.php
```diff
@@ -0,0 +1,63 @@
+<?php
+
+add_action( 'user_profile_update_errors', 'validate_current_password', 1, 3 );
+
+function validate_current_password( $errors, $update, $user ) {
```
rinatkhaziev, line 5: Let's add PHPDoc and type hints for arguments.

001-core/password.php
GaryJones: The text domain is missing on this and other i18n functions. This string isn't in the WordPress core files, so it would need a text domain to have it be loaded from elsewhere. But that hasn't yet been determined for this repo :-)

001-core/password.php
```diff
@@ -0,0 +1,63 @@
...
+		$errors->add( 'wrong_current_password', __( '<strong>ERROR</strong>: The entered current password is not correct.' ), array( 'form-field' => 'current_pass' ) );
+	}
+}
+
+add_action( 'show_user_profile', 'add_current_password_field' );
+
+function add_current_password_field() { ?>
+	<table id="nojs-current-pass" class="form-table" role="presentation">
+		<tr>
+			<th><label for="current_pass"><?php _e( 'Current Password' ); ?></label></th>
```
GaryJones, line 28: Does this repo treat future translated strings as trusted? If not, then `_e` should be `esc_html_e()` and other translated strings need escaping too.

001-core/password.php
```diff
@@ -0,0 +1,63 @@
...
+function reposition_current_password_field() {
+	$screen = get_current_screen();
+	if ( 'profile' != $screen->id ) {
+		return;
+	} 
+	?>
+
+	<script type="text/javascript">
+		(function($) {
+			$(document).ready(function() {
```
GaryJones, line 54: There's an effort to move away from jQuery in WordPress, so if this moving and styling of DOM elements can be done without jQuery, the better.

Also - what happens if someone doesn't have JS available?

001-core.php
```diff
@@ -7,6 +7,7 @@
  * License: GPL version 2 or later - http://www.gnu.org/licenses/old-licenses/gpl-2.0.html
  */
 
+require_once( __DIR__ . '/001-core/password.php' );
```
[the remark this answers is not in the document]
pauarge: I did put into `core` because we were changing Core's behavior, but moving it to `security`.

001-core/password.php
pauarge: Not sure what to do about the text domain if it hasn't been determined for this repo. Should we come up with one now or is it something we want to discuss further?

GaryJones: Probably want to discuss further, so in the meantime, the i18n function is basically redundant.

001-core/password.php
pauarge: Not sure about this one. I only found `esc_html_e` inside the Akismet folder, so I'd like to double check if that is a repo-wide thing.

001-core/password.php
pauarge: In that case, is it ok if I remove it or can we leave it there?

001-core/password.php
pauarge: I'm pushing some changes to not use jQuery in this JS. If JS is not available, the repositioning will not happen and the form will look like this:

Screen Shot 2021-04-07 at 5 36 41 PM

pauarge:
@nickdaugherty Nice catch! There where actually two bugs: the one you point out with the nonce and another one when editing a third user, which would not find the form element and thus report an empty current password. I've added two conditionals in the top of the `validate_current_password` function and the corresponding unit tests.

001-core/password.php
GaryJones: There's going to be a few other strings that will need the i18n strings added, so I would say it can be removed.

001-core/password.php
nickdaugherty: Let's use the escaped versions when possible. It would be difficult, but not impossible for an attack to be carried out through a translation. Thanks for the reminder @GaryJones!

rinatkhaziev (changes requested):
Thanks, looks good! 

Left some feedback but it's mostly semantic and minor stuff. 

I also tested it and once I fixed the type hint (otherwise it fataled) it works great!

security/password.php
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Validate current password in submitted user profile.
+ *
+ * @param WP_Error $errors Error object passed by reference
+ * @param bool $update Whether this is a user update
+ * @param WP_User $user User object (passed by reference)
+ *
+ * @return void
+ */
+function validate_current_password( WP_Error $errors, bool $update, WP_User $user ) {
```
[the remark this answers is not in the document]
pauarge: I initially tried `stdClass` as per the documentation, but as you point out `object` would be a better fit here since it satisfies `WP_User` (which is what I it's being passed in multiple places).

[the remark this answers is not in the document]
pauarge: Regarding passing by reference, I removed it because this IDE suggestion:

Screen Shot 2021-04-12 at 9 26 54 AM

security/password.php
```diff
@@ -0,0 +1,82 @@
...
+	if ( ! isset( $_POST['pass1'] ) || empty( $_POST['pass1'] ) ) {
+		return;
+	}
+
+	if ( ! isset( $_POST['current_pass'] ) || empty( $_POST['current_pass'] ) ) {
+		$errors->add( 'empty_current_password', '<strong>Error</strong>: Please enter your current password.', array( 'form-field' => 'current_pass' ) );
+		return;
+	}
+
+	$error = wp_authenticate( $user->user_login, $_POST['current_pass'] );
```
rinatkhaziev, line 39:
```suggestion
	$auth = wp_authenticate( $user->user_login, $_POST['current_pass'] );
```

security/password.php
```diff
@@ -0,0 +1,82 @@
...
+	if ( is_wp_error( $error ) ) {
```
rinatkhaziev, line 40:
```suggestion
	if ( is_wp_error( $auth ) ) {
```

rinatkhaziev (changes requested):
The code looks good. just one minor naming change (see inline suggestions) and we're good to go!

The reason for that naming suggestion is that `$error` sounds like we're expecting an error, where in reality we're expecting a WP_User object.

rinatkhaziev (approved):

PR #2070, opened 2021-04-12
Handle error response codes in Alerts

pauarge:
## Description

The send function from the alerts library would only fail (i.e. return a `WPError`) if the request would completely fail (could not be sent, there was a problem with the network, etc). That is fine, but it treated as correct requests that received a response with an error code.

This PR introduces additional checks on the response, and raise a `WPError` in case that response returns a non successful HTTP status code (`4xx` or `5xx`). Also, unit tests for this functionality have been added.

### Handle failed status codes on alerts

Raise an error when sending and alert that returns a failed response.

## Checklist


lib/utils/class-alerts.php
```diff
@@ -50,18 +50,24 @@ private function __construct() {
...
-			'body' => json_encode( $body ),
+			'body'   => json_encode( $body ),
 		] );
 
 		if ( is_wp_error( $response ) ) {
 			return new WP_Error( 'alert-send-failed', sprintf( 'Error sending alert: %s', $response->get_error_message() ) );
 		}
 
+		// The request got a response
+		$code = $response->response->code;
```
rinatkhaziev, line 66: Two things here:
1. The result of a request is not an object but an associative array
2. Let's use Core APIs to grab the code, [see](https://developer.wordpress.org/reference/functions/wp_remote_retrieve_response_code/)

lib/utils/class-alerts.php
```diff
@@ -50,18 +50,24 @@ private function __construct() {
...
+		if ( $code >= 400 ) {
+			return new WP_Error( 'alert-send-failed', sprintf( 'The request returned an invalid response: %s', $response->response->message ) );
```
rinatkhaziev, line 68:
```suggestion
			return new WP_Error( 'alert-send-failed', sprintf( 'The request returned an invalid response: %s', $response['response']['message'] ) );
```

rinatkhaziev (changes requested):
Left a couple of notes inline.

rinatkhaziev (approved):
This is great, extra thanks  for adding the tests!

PR #2080, opened 2021-04-26
Support for mutiple IPs in HTTP_X_FORWARDED_FOR

pauarge:
## Description

**This PR basically borrows from https://github.com/Automattic/vip-go-mu-plugins/pull/1840 and [especially this comment](https://github.com/Automattic/vip-go-mu-plugins/pull/1840#discussion_r520287221).**

The `HTTP_X_FORWARDED_FOR` might contain a list of IP addresses. Instead of comparing the passed `$value` to the whole header, this PR changes the behavior to find a match of the provided IP in the header.

Added some tests to check basic functionality of the class.

### Support for mutiple IPs in `HTTP_X_FORWARDED_FOR`

The IP filter now correctly blocks IPs when the requests goes through multiple proxies.

## Checklist


lib/class-vip-request-block.php
```diff
@@ -25,15 +25,24 @@ class VIP_Request_Block {
 	 * @param string $value target IP address to be blocked.
 	 */
 	public static function ip( string $value ) {
+		// Don't run validation if the passed value is not a valid IP
+		if ( ! filter_var( $value, FILTER_VALIDATE_IP ) ) {
+			return;
+		}
+
```
pauarge, line 32: Adding this because now we're matching substrings with `stripos`. If the user would pass `1` as `$value` and the IP is `1.1.1.1`, we would block the request even though `1` is not a valid IP address.

tests/lib/test-vip-request-block.php
```diff
@@ -0,0 +1,38 @@
...
+		parent::setUpBeforeClass();
+
+		require_once __DIR__ . '/../../lib/class-vip-request-block.php';
+	}
+
+	public function test__error_raised_true_client_ip() {
+		VIP_Request_Block::ip( '2.2.2.2' );
+
+		// Asserting that no exception has been raised up to this point
+		$this->assertTrue( true );
```
GaryJones, line 19: Instead of asserting that true is true, consider using https://phpunit.readthedocs.io/en/7.5/annotations.html#doesnotperformassertions annotation instead, to avoid the risky test.

lib/class-vip-request-block.php
```diff
@@ -25,15 +25,24 @@ class VIP_Request_Block {
 	 * @param string $value target IP address to be blocked.
 	 */
 	public static function ip( string $value ) {
+		// Don't run validation if the passed value is not a valid IP
+		if ( ! filter_var( $value, FILTER_VALIDATE_IP ) ) {
+			return;
```
rinatkhaziev, line 30:
```suggestion
			error_log( 'VIP Request Block: The value passed is not a correct IP address: ' . $value );
			return;
```

rinatkhaziev (changes requested):
Thank you, looks good. Another thank you for tests.

Some minor things:
1. Added a suggestion to make the comment a bit more clear.
2. In tests, let's also test for the correct block for True-Client-IP, something along the lines of 
```$_SERVER['HTTP_TRUE_CLIENT_IP']  = '4.4.4.4';```
3. In tests, it's not immediately apparent why `$this->expectException( PHPUnit\Framework\Error\Warning::class );` does constitute a successful block, something like `We're detecting that the block is successful by expecting `Cannot modify header information - headers already sent by` warning`

lib/class-vip-request-block.php
pauarge: I committed this suggestion and the linting fails due to using `error_log` in production. Not sure what to do because the same method is used below in the same file.

rinatkhaziev (approved):
Awesome!

PR #2084, opened 2021-04-27
Add wp-parsely plugin (opt-in with filter) & management script

jblz:
## Description

This change adds the wp-parsely plugin ([github source](https://github.com/Parsely/wp-parsely), [WordPress.org plugin repo](https://wordpress.org/plugins/wp-parsely/)) to the the vip-go-mu-plugins repo with a helper plugin to (conditionally) load it. Also included is a script to help us maintain the version with `git subtree` (inspired by the [jetpack script](https://github.com/Automattic/vip-go-mu-plugins/blob/31fcb103d12b38faa7fbba652e3cfc3da6b40989/bin/add-jetpack-subtree.sh) ).

At present, the helper plugin does nothing unless a particular filtering function returns true. Having this in place will allow us to iterate on the plugin in the context of VIP Go as we work to integrate the service.

Additionally, the helper plugin loads in `after_setup_theme` and bails if it detects a `Parsely` class. This is to prevent fatals if the plugin is already loaded via a "normal" plugin. Older versions of the plugin attempted to define the class unconditionally, so this works around that.

### Plugin Added: wp-parsely 2.5.0

[Parse.ly](https://parse.ly) gives creators, marketers, and developers the tools to understand content performance, prove content value, and deliver tailored content experiences that drive meaningful results.

The wp-parsely plugin makes it a snap to add the required tracking code to enable Parse.ly on your WordPress site. 

This change adds the plugin to the repository, but is disabled unless intentionally enabled for your site via the `wpvip_parsely_load_mu` filter.

## Checklist

~~- [ ] This change has relevant unit tests (if applicable).~~
~~- [ ] This change has relevant documentation additions / updates (if applicable).~~

## Steps to Test

1. Check out PR.
1. Go to `wp-admin`
    * There should be no `Parse.ly` entry under Setttings
1. Enable the function by adding `add_filter( 'wpvip_parsely_load_mu', '__return_true' );` to a sourced theme or plugin file.
1. Go to `wp-admin`
    * There **should** be a `Parse.ly` entry under Setttings
1. Installing the wp-parsely plugin on the site should override this version of the plugin
1. The manage script should be able to pull in any tag from the plugin's repo (e.g. `./bin/manage-wp-parsely-subtree.sh 2.5 2.5.0` should pull in the 2.5.0 release of the plugin to the `wp-parsely-2.5` directory and commit it to your local branch)


brandon-m-skinner:
@jblz Can you add the files not needed for this to work to [.deployignore](https://github.com/Automattic/vip-go-mu-plugins/blob/master/.deployignore)?

The changelog entry could use more context about what this is. The audience for the changelog is customers.

jblz:
Thanks for taking a look @brandon-m-skinner!

I added the patterns to the .deployignore that weren't already covered by top level patterns.

I also updated the changelog copy to be more friendly.  Is the `Changelog Description` in the PR description all I have to do?

brandon-m-skinner:
@jblz Yep, just the changelog description. The entry is generated automatically from that section on merge.

GaryJones:
Looks like `.phpcs.xml.dist` should be in the generic list at the top, and `init.py` should be in the wp-parsely section.

rinatkhaziev (approved):

PR #2094, opened 2021-05-04
Search Dev Tools

rinatkhaziev:
## Description

Search Dev Tools is aiming to be a one-stop-shop for developers integrating Search into their applications.

## Key Features

- See general debug information like which post types and post statuses are indexable, what meta keys are allowed, and is the site being rate-limited right now.
- See the list of the queries that were executed during this request:
	* Tweak a query in an editor with syntax highlighting and see the updated result.
	* See the stack trace for a query.
	* See WP_Query arguments for a query

## Screenshots

search-dev-tools-1
search-dev-tools-2
search-dev-tools-3
search-dev-tools-4
search-dev-tools-5
search-dev-tools-6

### Plugin Released: Search Dev Tools

Search Dev Tools are available under a feature flag.

## Checklist

## Steps to Test

Add the following to `vip-config.php`:

```
define( 'VIP_SEARCH_DEV_TOOLS', true );
define( 'VIP_ENABLE_VIP_SEARCH', true );
```

Please refer to README.md for instructions on how to build and test locally.


search/search-dev-tools/src/components/query/index.js
```diff
@@ -0,0 +1,153 @@
...
+					<CollapsibleList title="Trace" list={ backtrace } />
+				</div>
+			</div>
+			<div className={style.query_res_header}>
+				Response
+			</div>
+			<div className={style.query_src}>
+				<div className={style.query_actions}>
+					{ state.editing || state.result !== txtResult
+						? ( <Fragment>
```
pschoffer, line 98: React has fragment shortcut -  `<>...</>`. Does that work in preact as well?

pschoffer (commented):
It looks nice. However, there are some more frontend specific set of technologies I need to read on:) 

For instance the whole `scss` is a bit confusing for me. I would vote for using plain old css and try to compose more than inherit and using variables coming from scss. But I bet half of it is just my ignorance talking:)

search/search-dev-tools/src/components/query/index.js
rinatkhaziev: @pschoffer It used to be that Preact only supported `<Fragment>` but I just checked the docs and they say it's supported now. Let me try!

rinatkhaziev: It works, but with a caveat that the named import should still use Fragment, removing the import blows the bundle size up 400b (gzipped).

search/search-dev-tools/src/components/overlay/index.js
```diff
@@ -0,0 +1,41 @@
...
+	useEffect( () => {
+		window.addEventListener( 'keydown', closeWithEscape );
+		return () => {
+			// Cleanup after event
+			window.removeEventListener( 'keydown', closeWithEscape );
+		};
+	}, [] );
+
+	return (
+		<div
```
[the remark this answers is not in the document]
rinatkhaziev: heh, I copy-pasted that from Jetpack without giving it too much of a thought. Let me see if works or not.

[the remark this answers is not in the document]
rinatkhaziev: Ok I addressed this, along with a couple of other clean ups

search/search-dev-tools/search-dev-tools.php
```diff
@@ -0,0 +1,252 @@
...
+
+if ( ! ( defined( 'VIP_ENABLE_VIP_SEARCH' ) && VIP_ENABLE_VIP_SEARCH ) ) {
+	return;
+}
+
+define( 'SEARCH_DEV_TOOLS_CAP', 'edit_others_posts' );
+
+add_action(
+	'rest_api_init',
+	function() {
```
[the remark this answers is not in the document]
rogertheriault: +1 - also hard to unhook (our docs contain a few anon functions and I'm working to move all of them to named functions)

rinatkhaziev:
@pschoffer Thanks for the review!  I've addressed all the feedback.

Regarding CSS vs SASS - the latter is quicker to write + it has mixins, most of the variables are already CSS variables, not SASS. There's not that much of inheritance, it's mostly used as a shortcut:

Consider:

```
.component__wrapper {}
.component__content {}
.component__something {}
```
vs

```
.component {
  &__wrapper {}
  &__content {}
  &__something {}
}
```

It's more of a stylistic/convenience choice.

search/search-dev-tools/search-dev-tools.php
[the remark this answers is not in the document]
rinatkhaziev: Fair point, switched every instance to a named function with proper PHPDoc

rogertheriault:
Just noting that on a completely generic site, even with Search already working, the following must be defined to see the admin bar item:

```
define( 'VIP_SEARCH_DEV_TOOLS', true );
define( 'VIP_ENABLE_VIP_SEARCH', true );
```

rinatkhaziev:
@rogertheriault  So `VIP_ENABLE_VIP_SEARCH` is needed to load via code, and the other one is just temporary until we get it released for general availability. I'll update the PR, `VIP_SEARCH_DEV_TOOLS` was a last-minute addition.

search/search-dev-tools/search-dev-tools.php
```diff
@@ -0,0 +1,269 @@
...
+		'vip/v1',
+		'search/dev-tools',
+		[
+			'methods'             => [
+				'POST',
+			],
+			'callback'            => __NAMESPACE__ . '\rest_callback',
+			'permission_callback' => __NAMESPACE__ . '\should_enable_search_dev_tools',
+			// Uncomment this for testing with `npm run dev`
+			// 'permission_callback' => '__return_true',
```
pauarge, line 47: Could we make this conditional to the environment we're in? Maybe using something like `WPCOM_IS_VIP_ENV` or `VIP_GO_APP_ENVIRONMENT` to figure out if we are in a dev machine?

search/search-dev-tools/src/components/overlay/index.js
```diff
@@ -0,0 +1,41 @@
...
+	}, [] );
+
+	return isVisible
+		? (
+			<div
+				className={[
+					'search-dev-tools__overlay',
+					`search-dev-tools__overlay--${ colorTheme }`,
+
+				].join( ' ' )}
```
pauarge, line 31:
```suggestion
				className={`search-dev-tools__overlay search-dev-tools__overlay--${ colorTheme }`} 
```

search/search-dev-tools/search-dev-tools.php
rinatkhaziev: It's better not to introduce any differences in behavior unless absolutely needed, otherwise, we're at risk of not noticing some change that would only happen in a non-local environment. 

I've removed the note, looking back at it, it seems something for a README.

pauarge (approved):

rinatkhaziev:
r1554-stacks

PR #2112, opened 2021-05-07
don't validate count without index

pschoffer:
## Description

This adds another condition to skipping count validation. Before this change count validations would through an alert if the index doesnt exist.

### Plugin Updated: Search

Improve internal index count monitoring job to skip non-initialized indexes.

## Checklist

## Steps to Test

1. Have a multisite environment
2. add subsite (this creates site with index-v1 only in DB, but not in ES)
3. Run `wp vip-search health validate-counts --network-wide`
4. You should get a message about index skipped due to index not being in ES.


search/includes/classes/class-health.php
```diff
@@ -69,11 +69,20 @@ public function validate_index_entity_count( array $query_args, \ElasticPress\In
...
 		];
 
+		if ( ! $indexable->index_exists() ) {
+			// If index doesnt exist and we will skip the rest of the check
+			$result['skipped'] = true;
+			$result['reason'] = 'index-not-found';
+			return $result;
+		}
+
+
```
pauarge, line 85: Non-blocking nit pick: we should only add one final line break

pauarge (approved):

search/includes/classes/class-health.php
pschoffer: Sounds good
```suggestion
```

pauarge (approved):

PR #2113, opened 2021-05-07
Support file related WP-CLI commands in Windows

lipemat:
## Description

The `WP_Filesystem_Base::find_folder` method strips out Windows drive letters and sanitizes Windows paths.
The `\Automattic\VIP\Files\WP_Filesystem_VIP::is_wp_content_subfolder_path` method verifies the path raw
which will never match and Windows environments and results in the following error when running any
update related `wp-cli` command.

```
Warning: Could not create directory.
```
This change matches the sanitization to both methods so the results are reliable and the commands
will function.

https://github.com/WordPress/wordpress-develop/blob/35c1cdee74d294b611221e1395b24b298a71a299/src/wp-admin/includes/class-wp-filesystem-base.php#L222

## Changelog Description
Fix: Support running wp-cli update commands on Windows.

## Checklist

## Steps to Test

On a Windows environment, update any plugin using `wp-cli`. 
Example:
```shell
wp plugin update akismet
```


rinatkhaziev:
Hi @lipemat,

Thanks for the PR! 

We've discussed this internally, and, unfortunately, this is not something we're willing to merge.

We don't run Windows in production nor do we recommend using Windows for local development. Please see our documentation on [setting up local dev environments](https://docs.wpvip.com/how-tos/set-up-a-vip-go-local-development-site/)

I apologize for the inconvenience.

lipemat:
Outside of fixing the issue on Windows, it seems like making the files path handling consistent with WP core would be a good thing. 🤷‍♂️

PR #2119, opened 2021-05-10
Tweak Jetpack Connection Pilot to autoconnect

pauarge:
## Description

This PR aims to run Jetpack Connection Pilot logic earlier so we can automatically connect newly created sites. We also hook to `wp_initialize_site` and `wp_update_site` to immediately run the same logic whenever changes (or creation) happens on the site.

We also make sure that reconnection is only attempted on production sites.

### Jetpack Connection Pilot

Improved automatic Jetpack connection process.

## Checklist

## Steps to Test
1. Check out PR.
2. Set up a new site (or a subsite in a network) on a production environment.
3. Make sure that Jetpack is initially disconnected.
4. In a short time, you should see Jetpack enabled.


vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
```diff
@@ -255,7 +260,12 @@ public static function should_run_connection_pilot() {
 	 * @param $error \WP_Error Optional error thrown by the connection check
 	 * @return bool True if a reconnect should be attempted
 	 */
-	public static function should_attempt_reconnection( $error = null ) {
+	public static function should_attempt_reconnection( \WP_Error $error = null ): bool {
+		// Do not try to reconnect anything else other than production sites
+		if ( 'production' !== VIP_GO_APP_ENVIRONMENT ) {
+			return false;
+		}
```
pauarge, line 267: Forcing non-prod environments to avoid reconnection. Is it ok that we introduce this conditional here, or we prefer to rely solely in the global variable properly set up?

000-vip-init.php
```diff
@@ -124,6 +124,11 @@
 	)
 );
 
+// Enable Jetpack private connection by default
+if ( ! defined( 'VIP_JETPACK_IS_PRIVATE' ) ) {
+	define( 'VIP_JETPACK_IS_PRIVATE', true );
+}
+
```
pauarge, line 131: Potentially breaking change

WPprodigy:
In `filter_vip_jetpack_connection_pilot_should_reconnect()` it filters if we should attempt an automatic reconnection. It defaults to fault, is is enabled on a case-by-case basis there.

I was thinking maybe when a new subsite is created, create an individual cron job with the current timestamp essentially. This individual cron job could pass an arg to do_cron() that signals it's a first-time connection attempt for a new site, letting it pass through for a guarenteed connection attempt. Then we don't need to worry much about load order either, as it will just run on the next cron job run-through.

pauarge:
@WPprodigy Thanks for your feedback! I have changed:

- Removed the explicit check for non-production environments. With the changes in `000-vip-init.php` it should be enough.
- Instead of calling `do_cron` right away when `wp_initialize_site` or `wp_update_site` happen, we schedule a cron job to run it a minute after.
- To evalute if `should_attempt_reconnection`, we are now checking `is_fresh_subsite` to attempt a connection if the site has been recently created.

vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
```diff
@@ -86,6 +94,10 @@ public function schedule_cron() {
 		}
 	}
 
+	public function schedule_immediate_cron() {
+		wp_schedule_event( strtotime( '+1 minutes' ), self::CRON_SCHEDULE, self::CRON_ACTION );
```
WPprodigy, line 98: This will leave a perpetual job, instead can schedule single event: https://developer.wordpress.org/reference/functions/wp_schedule_single_event/

vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
```diff
@@ -252,16 +264,44 @@ public static function should_run_connection_pilot() {
...
-	public static function should_attempt_reconnection( $error = null ) {
+	public static function should_attempt_reconnection( \WP_Error $error = null ): bool {
 		if ( defined( 'VIP_JETPACK_CONNECTION_PILOT_SHOULD_RECONNECT' ) ) {
 			return VIP_JETPACK_CONNECTION_PILOT_SHOULD_RECONNECT;
 		}
 
+		// Attempting connection for fresh sites
+		if ( self::is_fresh_subsite() ) {
+			return true;
+		}
```
WPprodigy, line 279: Just for the sake of consistency, could move this to the top of `filter_vip_jetpack_connection_pilot_should_reconnect()` maybe?

000-vip-init.php
```diff
@@ -139,7 +144,7 @@
 		'schedule'  => 'hourly',
 		'action'    => 'wpcom_vip_run_jetpack_connection_pilot',
 		'callback'  => array( '\Automattic\VIP\Jetpack\Connection_Pilot', 'do_cron' ),
-		'timestamp' => strtotime( sprintf( '+%d minutes', mt_rand( 1, 60 ) ) ),
+		'timestamp' => strtotime( sprintf( '+%d minutes', mt_rand( 1, 30 ) ) ),
```
WPprodigy, line 147: Just noting this change won't have any effect unless https://github.com/Automattic/Cron-Control/pull/177 is live. Generally I think it's not needed so much anymore if we stick with the phased release structure we have planned.

pauarge: If it doesn't work right now because we need to merge that other PR, I reckon it's better to get rid of this line entirely, as it might lead to confusion. As you point out, we should be good to go with the phased release.

pauarge:
@WPprodigy As discussed, I have renamed the constants. We now only have one constant, `VIP_JETPACK_AUTO_MANAGE_CONNECTION` (we can rename if the name is not descriptive enough). If that is enabled, we will try to auto connect and reconnect whenever is possible.

I have left some logic in case someone had been using `VIP_JETPACK_CONNECTION_PILOT_SHOULD_RUN` or `VIP_JETPACK_CONNECTION_PILOT_SHOULD_RECONNECT`.

WPprodigy:
Nice, definitely looking good. I like `VIP_JETPACK_AUTO_MANAGE_CONNECTION` naming, nice job 😀 . And we can maybe still use `VIP_JETPACK_CONNECTION_PILOT_SHOULD_RECONNECT` while rolling out, but then yeah can remove that entirely once fully released.

pauarge:
Marking this PR as blocked, waiting to align on other Jetpack work prior to releasing it.

vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
```diff
@@ -252,16 +269,40 @@ public static function should_run_connection_pilot() {
...
+			return false;
+		}
+
+		$now = new DateTime("now");
+		$time_diff = $now->getTimestamp() - $site_registered->getTimestamp();
+		return $time_diff < 3600;
+	}
 }
 
 Connection_Pilot::instance();
```
WPprodigy, line 308: What do you think about booting this all up on the `init` hook instead of right away here? Essentially:

```
  Connection_Pilot::instance();
 }, 25 );
 ```
 
Benefit being customers can use the filters like `vip_jetpack_connection_pilot_should_run` (and constants) in their theme or plugins that way. Otherwise they are pretty restricted to having to do all this before the mu-plugin is loaded, meaning vip-config.php is the only option. In which place you can't really target subsites very easily.

Alternatively, run the `init_actions()` method on the `init` hook instead maybe? Needs compatibility testing specifically with the cron hook, but pretty sure it would be okay.

pauarge: Absolutely. I didn't want to touch this as it wasn't directly related to the PR, but I agree that it adds value. What's more, I've seen this unconditional instantiations of classes in some other places in mu-plugins. I think it's way safer to do it with a hook.

I've pushed a change with your proposed code.

PR #2131, opened 2021-05-13
Update Query Monitor 3.6.5 => 3.7.1

johnbillion:
## Description

This updates the Query Monitor plugin from 3.6.5 to 3.7.1.

### Query Monitor updated to version 3.7.1.

Full changelog here: https://github.com/johnbillion/query-monitor/releases

## Checklist

## Steps to Test

**Important:** This introduces [a new feature which exposes the SQL query logs to suitably authenticated users during REST API requests](https://querymonitor.com/blog/2021/05/debugging-wordpress-rest-api-requests/). There is no change to the privileges required to see QM's data, but this will need testing on VIP before deploying because AFAICT [VIP uses a non-standard structure for the database queries logged in the `wpdb` class](https://github.com/automattic/vip-go-mu-plugins/blob/9180eab0ff8a2792325e8a5decf59e4a25c28e56/vip-helpers/vip-query-monitor.php).

### DIFF
--- a/query-monitor.php
+++ b/query-monitor.php
@@ -140,6 +140,7 @@ add_action( 'wp', 'wpcom_vip_qm_disable_on_404' );
 
 
 // We are putting dispatchers as last so that QM still can catch other operations in shutdown action
+// See https://github.com/johnbillion/query-monitor/pull/549
 function change_dispatchers_shutdown_priority( array $dispatchers ) {
 	if ( is_array( $dispatchers ) ) {
 		if ( isset( $dispatchers['html'] ) ) {
--- a/query-monitor.php
+++ b/query-monitor.php
@@ -1,20 +1,19 @@
 <?php
-
 /**
  * Query Monitor plugin for WordPress
  *
  * @package   query-monitor
  * @link      https://github.com/johnbillion/query-monitor
  * @author    John Blackbourn <john@johnblackbourn.com>
- * @copyright 2009-2020 John Blackbourn
+ * @copyright 2009-2021 John Blackbourn
  * @license   GPL v2 or later
  *
  * Plugin Name:  Query Monitor
  * Description:  The Developer Tools Panel for WordPress.
- * Version:      3.6.5
- * Plugin URI:   https://github.com/johnbillion/query-monitor
- * Author:       John Blackbourn & contributors
- * Author URI:   https://github.com/johnbillion/query-monitor/graphs/contributors
+ * Version:      3.7.1
+ * Plugin URI:   https://querymonitor.com/
+ * Author:       John Blackbourn
+ * Author URI:   https://querymonitor.com/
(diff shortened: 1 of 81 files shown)


pschoffer:
Thanks for the PR @johnbillion. While testing this I found the issue we have introduced, similar to the discussion in [this PR](https://github.com/johnbillion/query-monitor/pull/549).

That is that currently, we don't capture any REST calls, due to the priority change we have done [here](https://github.com/Automattic/vip-go-mu-plugins/blob/5fecaa13f2ff40ee1bdee6160b40c22f853f1a21/query-monitor.php#L157). I'll fix that first so we can verify more here.

pschoffer:
Hmm, actually @johnbillion I might have been wrong. I am probably setting up the manual testing wrong. Could you help me with some basic scenario when the rest call should have been caught by the query monitor? Or feel free to point me to some resource to read up on.

I tried this (I do get back the response printed on the page, but nothing in the HTTP API calls in the query monitor):
```
function tmp_request() {
	$request = new WP_REST_Request( 'GET', '/wp/v2/posts' );
	$request->set_param( 'per_page', 2 );
	$response = rest_do_request( $request );
	var_dump(['REST TEst', $response]);
}

add_action( 'wp_body_open', 'tmp_request' );
```

johnbillion:
You'll want to test this with a client-side request instead of a server-side request because you need to perform an enveloped request for the full output. Details are in the links above.

The easiest way to do this client side is to set up an application password in your WordPress user profile, then use an HTTP client such as Insomnia or Postman with the username and password in BasicAuth.

pschoffer:
Oh right now I get it.  Thanks for the patience.

I tested using something like:
```
function tmp_request() {
	print '<script>
		let username = "username";
		let password = "password";

		let headers = new Headers();

		headers.set("Authorization", "Basic " + btoa(username + ":" + password));
		fetch("http://vip-local.vipdev.lndo.site//wp-json/wp/v2/posts/1?_envelope", {headers: headers}).then(console.log);
	</script>
	';
}

add_action( 'wp_footer', 'tmp_request' );
```

However I didn't manage to get the BaseAuth to work as expected...so I commented the part of the code that verifies ` self::user_can_view()` in both REST and REST_Envelope dispatchers. After which I did get the `qm` object in the response and the new headers back.

The `gm` headers lack some expected values due to us having our custom Query collector. Most notable `results` aren't collected. But we still get the query sql, stack and the time. 

Everything else I tested works as expected.

pschoffer:
I tested it some more on sandbox + ran our regular test pipeline. Looks all good. Deploying to staging next.

johnbillion:
Do we need to trigger a new deployment in order for an updated `mu-plugins` directory to be deployed to our staging and prod sites?

pauarge:
@johnbillion It should be deployed to staging and prod now.

PR #2133, opened 2021-05-14
Search Dev Tools: address some edge cases

rinatkhaziev:
## Description

During the staging testing we've discovered a couple of issues:

1. Dev Tools were loaded regardless of whether the Search itself is properly initialized or not that could cause a fatal error. I.e. if the Search code is enabled but the Search itself is not configured properly (constants are not defined correctly) we don't initialize Search.
2. Very long strings could break the layout.
3. As a bonus, updating some dependencies.

A hotfix that addressed potential layout issues and an edge case that may cause a fatal error.

## Checklist

## Steps to Test

1. Check out the PR.
2.  Comment out VIP_ELASTICSEARCH_ENDPOINTS or VIP_ELASTICSEARCH_USERNAME 
3. Verify there are no fatals and dev tools code isn't being loaded.


search/search-dev-tools/src/components/query/style.scss
```diff
@@ -116,6 +115,7 @@ pre[class*='language-'] {
 	grid-area: query-src;
 	background-color: #ffffff;
 	position: relative;
+	max-width: calc( 50vw - 32px );
```
pauarge, line 118: We could have `50vw - 32px` as a constant. Also, it could be great if we put `32px` in its `vw` equivalent.

rinatkhaziev: Good point on using a variable. 

> Also, it could be great if we put 32px in its vw equivalent.
`vw` is a relative unit (1% of the viewport width) so it's not directly comparable, that's why we're using `calc` - to explicitly set the dimensions to 50% of a viewport minus 32px (which is a gutter/spacing constant that's used across the design).

pschoffer (approved):
LGTM

PR #2134, opened 2021-05-14
Update/ensure index

pschoffer:
### Plugin Updated: Search

This change ensures the index is correctly initialized in ElasticSearch before operations that could create a default index. Default index doesn't have the correct mapping set and therefore could potentially cause search issues.

## Checklist

## Steps to Test

Replace index and ES cluster in the following:
1. Delete index `curl -X DELETE http://127.0.0.1:49155/vip-200508-post-1`
2. Add a post `dev-env wp -- post create --post_title="Test" --post_content="Content" --post_excerpt="Test" --post_status="publish"`
3. Check post_date `type` to be `date` - `curl http://127.0.0.1:49155/vip-200508-post-1/_mapping | jq | grep -A 5 post_date`


search/includes/classes/class-search.php
```diff
@@ -1316,19 +1378,18 @@ public function is_bulk_url( string $url ) {
 	}
 
 	/**
-	 * Given an ES url, determine the index name of the request for stats purposes
+	 * Given an ES url, determine the index name of the request
 	 */
-	public function get_statsd_index_name_for_url( $url ) {
```
pschoffer, line 1321: The function was not used so I repurposed it.

tests/search/includes/classes/test-class-search.php
```diff
@@ -2,7 +2,9 @@
```
pschoffer, line 2: I finally found courage to move this test file where it belongs:)

search/includes/classes/class-search.php
```diff
@@ -765,6 +770,63 @@ public function filter__ep_do_intercept_request( $request, $query, $args, $failu
...
+	/**
+	 * We want to ensure that index exists before doing certain type of operation on it.
+	 *
+	 * The reason is that indexing a document, before mappings were put, would result in index with default, incorrect mapping.
+	 * This method avoids this issue by creating the index with correct mapping if the index doesn't exist yet.
+	 */
+	public function ensure_index_existance( $url, $args ) {
+		$method = strtoupper( $args['method'] ?? '' );
+		$methods_that_need_index = [ 'POST', 'PUT' ];
+		if ( ! in_array( $method, $methods_that_need_index ) ) {
```
rinatkhaziev, line 782:
```suggestion
		if ( ! in_array( $method, $methods_that_need_index, true ) ) {
```

search/includes/classes/class-search.php
```diff
@@ -765,6 +770,63 @@ public function filter__ep_do_intercept_request( $request, $query, $args, $failu
...
+			return true;
+		}
+
+		$index_name = $this->get_index_name_for_url( $url );
+		if ( ! $index_name ) {
+			return true;
+		}
+
+		$path_parts = explode( '/', trim( $url ) );
+		if ( end( $path_parts ) == $index_name ) {
```
rinatkhaziev, line 793: This can be swapped with `wp_endswith`

pschoffer: This covers both `http://.../index_name/` and `http://.../index_name`, I guess I could do `trim( trim( $url ), '/' )` and then `wp_endswith` on that. Seems kinda similar to me. I don't have strong opinions, if you think I should go with wp_endswith instead, I am fine with that.

rinatkhaziev: as long as we don't need other `$path_parts` let's do the `wp_endswith( trim( $url, '/'), $index_name )`

rinatkhaziev (approved):

search/includes/classes/class-search.php
pschoffer: done

PR #2141, opened 2021-05-17
Fix imports when Jetpack is disabled

pauarge:
## Description

We allow Jetpack to be disabled through `VIP_JETPACK_SKIP_LOAD`. If that is the case, site imports through the CLI will fail, because in their cleanup phase, they try to reconnect Jetpack.

This PR checks if Jetpack is disabled, and if that's the case, it skips that reconnection attempt.

### Fix imports when Jetpack is disabled

Imports through CLI would fail if Jetpack is disabled on the site.

## Checklist

## Steps to Test

1. Check out PR.
2. Disable Jetpack through `VIP_JETPACK_SKIP_LOAD`
3. Perform a site import. It should complete


wp-cli/vip-migrations.php
```diff
@@ -58,12 +58,15 @@ function cleanup( $args, $assoc_args ) {
...
-			'jetpack_private_options',
 			'vaultpress',
 			'wordpress_api_key',
 		];
 
+		if ( ! defined( 'VIP_JETPACK_SKIP_LOAD' ) || ! VIP_JETPACK_SKIP_LOAD ) {
+			$options[] = 'jetpack_options';
+			$options[] = 'jetpack_private_options';
+		}
+
```
rinatkhaziev, line 69: I think we should revert this part as it used to be. I.e., we should keep trying to delete these options, otherwise, we may end up in a situation where Jetpack is reactivated and there are production tokens in non-prod DB.

rinatkhaziev (changes requested):

rinatkhaziev (approved):

PR #2150, opened 2021-05-20
Loading VIP Jetpack conditionally and hiding more messaging

pauarge:
## Description

In this PR:

- We load VIP Jetpack customizations (`vip-jetpack.php`) only if Jetpack is enabled (we move the unconditional `require_once` to inside a function that's only reached if Jetpack is enabled).
- We disable some additional Jetpack messaging through the `jetpack_just_in_time_msgs` constant.

### Plugin Updated: Jetpack 

Better handling of Jetpack when it's disabled. Removal of some extra Jetpack messaging.

## Checklist

## Steps to Test

1. Check out PR.
2. Load wp-admin with Jetpack enabled (Jetpack should function just fine).
3. Load wp-admin with Jetpack disabled (we shouldn't see any errors).

### DIFF
--- a/jetpack.php
+++ b/jetpack.php
@@ -135,8 +135,8 @@ function vip_jetpack_load() {
 	if ( version_compare( JETPACK__VERSION, '9.1', '<' ) ) {
 		add_filter( 'instagram_cache_oembed_api_response_body', '__return_true' );
 	}
+
+	require_once( __DIR__ . '/vip-jetpack/vip-jetpack.php' );
 }
 
 vip_jetpack_load();
-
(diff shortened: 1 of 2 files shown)


rinatkhaziev (approved):
LGTM but I defer to Caleb in case I missed something.

WPprodigy (approved):
I think there's a small risk where a client uploads jetpack incorrectly to `client-mu-plugins/jetpack/jetpack.php` but doesn't have all the right content. Leading to a fatal in `add_jetpack_menu_placeholder()` after these changes.

Pretty extreme scenario though, _I think_.

PR #2151, opened 2021-05-20
ALLOPTIONS handling improvements

rinatkhaziev:
## Description

Simplify ALLOPTIONS alerting/blocking logic: use higher limits for uncompressed values for alerting purposes, and only check the compressed limit before terminating the request. 

The previous iteration did reduce the number of alerts and unnecessary blocking but it may have inadvertently affected the performance during certain traffic scenarios. This should handle things in a more careful fashion.

The next step will be to remove this logic from web context, since we're trying to limit this once a minute, running on cron would allow us to eliminate any potential race conditions due to traffic spikes.

### Improvements for internal alerting

We've reworked some internal alerting regarding very large autoloader options.

## Checklist

## Steps to Test

1. Checkout the PR
2. Add a large autoload option, like `add_option( 'test_compression', range( 1, 500000 ) );`
3. Verify that site is not blocked
4. Increase the value

### DIFF
--- a/alloptions-limit.php
+++ b/alloptions-limit.php
@@ -52,7 +52,7 @@ function wpcom_vip_sanity_check_alloptions() {
 	}
 }
 
-function wpcom_vip_sanity_check_alloptions_die( $alloptions ) {
+function wpcom_vip_sanity_check_alloptions_die( &$alloptions ) {
 	// Do a final check before killing the request if the value is actually more than the value limit.
 	// We're using gzdeflate here because pecl-memcache uses Zlib compression for large values.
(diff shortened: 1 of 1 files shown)


netsuso (approved):
It makes much more sense this way, thank you!

I had a suggestion about where to put the check for the compressed size. It might still be a good fit in the main block from where it has been removed. We could simply check the compressed size only after the uncompressed size threshold has been exceeded. This way the logic is "check the fast and easy uncompressed size, and if it exceeds that limit, then calculate the compressed size to determine more precisely if it fits in memcached", instead of two disconnected checks.

But as we are going to refactor this again soon and put it into a cronjob, I think it's not worth changing anything right now.

rinatkhaziev:
r1568-stacks

PR #2157, opened 2021-05-24
Deadlock Job Cleanup Updateup

pschoffer:
## Description

There is an edge case that can cause the cleanup job of deadlocked jobs ... well to be deadlocked itself :) Please see the comment in the code for more context.

### Plugin Update: Search

Update deadlock cleanup job for the indexing queue to handle correctly multiple jobs trying to reindex the same object.

## Checklist

## Steps to Test

This is quite hard to test. I think the state can happen due to a race condition in EP code.

The scenario you want is:
 * 2 jobs in the queue - wp_vip_search_index_queue db table
 * 1 needs to be in state 'running'
 * 1 needs to be in state 'scheduled'

So I guess inserting those to DB directly works:
```
INSERT into wp_vip_search_index_queue (`object_id`, `object_type`, `status`, `index_version`, `scheduled_time`) VALUES (1, 'post', 'scheduled', 1, '0000-00-00'), (1, 'post', 'running', 1, '0000-00-00');
```

Then run:
```
wp eval "do_action('vip_search_queue_sweeper');"
```

wp_vip_search_index_queue table should now have only one job in state scheduled and scheduled time now-ish

### DIFF
--- a/search/includes/classes/class-queue.php
+++ b/search/includes/classes/class-queue.php
@@ -710,11 +710,12 @@ class Queue {
 		$filtered_deadlocked_jobs = [];
 		$jobs_to_be_deleted = [];
 
-		foreach ($all_deadlocked_jobs as $job) {
+		foreach ( $all_deadlocked_jobs as $job ) {
 			$unique_key = sprintf( '%s_%s_%s',
 				$job['object_id'] ?? '',
 				$job['object_type'] ?? '',
-				$job['index_version'] ?? '');
+				$job['index_version'] ?? ''
+			);
 
 			if ( array_key_exists( $unique_key, $found_objects ) ) {
 				$jobs_to_be_deleted[] = $job;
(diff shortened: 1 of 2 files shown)


search/includes/classes/class-queue.php
```diff
@@ -696,6 +698,41 @@ public function free_deadlocked_jobs() {
 		}
 	}
 
+	/**
+	 * In some cases there could be multiple jobs for the same object stucked in deadlock state.
```
rinatkhaziev, line 702:
```suggestion
	 * In some cases there could be multiple jobs for the same object stuck in a deadlock state.
```

search/includes/classes/class-queue.php
```diff
@@ -696,6 +698,41 @@ public function free_deadlocked_jobs() {
...
+	 * If we would try to update all of them at once we would break the DB contstraint
```
rinatkhaziev, line 703:
```suggestion
	 * If we would try to update all of them at once we would break the DB constraint
```

rinatkhaziev:
Thanks, looks good! Left a couple of minor typo suggestions.

netsuso (approved):

netsuso:
stacks-r1572

PR #2165, opened 2021-05-28
Adds longer timeout to validate counts

pschoffer:
## Description

Sometimes (often on bigger sites) our periodic validate count checks fail due to timeout. This change introduces a way to increase timeout and increases it to 5s (from 2s) for validating counts operation.

### Plugin Updated: Search

This change introduces a way to increase timeout and increases it to 5s (from 2s) for validating counts operation.

## Checklist

## Steps to Test

1. Add debug line: search/includes/classes/class-search.php:722 - `var_dump(['timeout', $timeout]);`
2. `dev-env exec -- wp vip-search health validate-posts-count` See the timeout being 5s


rinatkhaziev (changes requested):
Introducing the filter leaves it open for potential abuse, so

1) we either need to either keep it hardcoded and figure out the value based on the method's arguments (can we tell that this is a search health query?)
2) or have our own filter running on PHP_INT_MAX. 

Option 1 is safer because one can still `remove_filter` and do some damage.

pschoffer:
Thanks @rinatkhaziev for the feedback.

Yeah, this is a bit tough. Here are some issues:

add 1) `we either need to either keep it hardcoded and figure out the value based on the method's arguments (can we tell that this is a search health query?)`
* I couldn't find anything really.
* We add `track_total_hits` but that is added in core elasticpress in some other queries as well. Not to mention users could just add that if they would want a longer timeout.

add 2) `or have our own filter running on PHP_INT_MAX.`
* I couldn't figure out how to do this. 
* Reason being we would want to have it set to 2s and only for the exception of health checks bump it to 5s. So we would need to already do the removing of the default filter and adding a different one I guess. Basically providing the code snippet on how to do this.

extra 3) Another option I considered before is to change the signature of both `Indexable::query_es` and `Elasticpress::query` to pass some extra options all the way to our search request handler. This feels a bit unpleasant though as we would diverge even more from EP code base. Also, this still lets the clients code to do the same.

extra 4) Similar, but maybe slightly better is to have a `request_timeout` field on our Search class instance and increase it from Health instead of using filters. This is only a bit more obscure than filters still accessible by the client's code. Unless there is a way to have functions only callable within the same namespace??

However, one thing to note is that there is a hard limit of how big the timeout can be in [here](https://github.com/Automattic/vip-go-mu-plugins/blob/a6d8e75ca30e4767329de580d55e8ac0366f9614/vip-helpers/vip-utils.php#L797). So the maximum for non-admin non-CLI is 5s.

pschoffer:
BTW you mentioned:

> Introducing the filter leaves it open for potential abuse

Could you say some more on that. It feels that giving clients the ability to increase (or decrease) the timeout might make some parts of the system less responsive if misused, but in the end, it is their site :). So if they would value to have some valuable yet slow query somewhere, it is OK to give them the option, or?

rinatkhaziev:
> Could you say some more on that. It feels that giving clients the ability to increase (or decrease) the timeout might make some parts of the system less responsive if misused, but in the end, it is their site :)

So the longer the timeout the higher the chances of things going wrong. E.g. in a catastrophic scenario where there's severe degradation in response time from ES for some reason, it could cause PHP worker starvation and the site would start 503. The chances are slim, but when running at the scale we do it may become a significant problem on a particularly bad day. It's not theoretical and happened enough times in the past that we should absolutely do everything we can to eliminate the possibility. 
People always try to game the system, sometimes to their own detriment :) and removing filter is as easy as `remove_all_filters( 'vip_search_default_query_timeout' )` or just override with a later priority.

I asked @nickdaugherty's opinion and I think his idea makes sense here: increase the timeout in the CLI context, roughly:

```
$timeout = defined( 'WP_CLI' ) && WP_CLI ? 5 : 2;
```

pschoffer:
Nice, yeah that is simpler and nice sollution.

rinatkhaziev (approved):

rinatkhaziev:
r1602-stacks

PR #2166, opened 2021-05-28
Featured Plugins Fix: Wait for images to load before calculating height

mikeyarce:
Currently if the JS loads prior to the featured plugin images, the box height will not be calculated properly.  This sets the initial calculation to happen after the images have loaded.

Edit:  I'm now trying an approach using CSS Grid so we don't need to calculate anything.

## Description

Currently if the images load after the script calculation happens:
Screen Shot 2021-05-28 at 8 53 57 AM

This PR just moves the initial calculation to be after the window has loaded so that the calculation happens after images are loaded.

## Checklist

## Steps to Test
Example:

1. Check out the PR
2. Make sure to change the version number in the `vip-plugins.php` file to invalidate the cache
3. Throttle your browser to something like "Slow 3G" to get consistent results
4. See if the calculation is accurate


mikeyarce:
@natebot noticed a few more issues with the featured plugins:

They are supposed to display some meta descriptions which are currently empty.  `$plugin->meta->listing_description` is outputting just an empty `<p>` element.  That's due to the endpoint not actually having the data:

https://wpvip.com/wp-json/vip/v0/plugins?type=technology

@natebot - do we actually still need this or should we remove it?

We also talked about potentially changing the `h4` tags as it's not good semantic practice.  Should we change them to paragraphs and add some CSS to style them?

rinatkhaziev (approved):
`+11 −87` LOVE IT

rinatkhaziev:
r1603-stacks

PR #2174, opened 2021-06-01
Adds safe reindexing support

pschoffer:
## Description
This change makes it easier to reindex life sites without impacting returned results.

### Plugin Updated: Search

Adds a `--using-versions` to index subcommand. This enables caller to call for instance:

```
wp vip-search index --using-versions --indexables=post --setup
```

This command will create a new version of index and fully index the content. While this is happening the previous version of index is still active and therefore not impacted and serving correct content. After indexing finishes, the command will activate the newly indexed version and remove the previous version. 

## Checklist

## Steps to Test

1. run `wp vip-search index --using-versions --indexables=post --setup`
1. run `wp vip-search index-versions list post`and see the new version replaced the previous one.


search/includes/classes/commands/class-corecommand.php
```diff
@@ -14,38 +14,105 @@ class CoreCommand extends \ElasticPress\Command {
...
-				$indexable_slugs = explode( ',', str_replace( ' ', '', $assoc_args['indexables'] ) );
-
-				foreach ( $indexable_slugs as $slug ) {
-					$indexable = \ElasticPress\Indexables::factory()->get( $slug );
+				$indexables = $this->_parse_indexables( $assoc_args );
+
+				if ( $using_versions ) {
+					foreach ( $indexables as $indexable ) {
+						$current_versions = $search->versioning->get_versions( $indexable );
+						if ( count( $current_versions ) > 1 ) {
```
rinatkhaziev, line 92: I think I get the intent here, but I'm not sure whether we should be restrictive here. I.e. if there were multiple index versions I would still appreciate the ability to run this logic.  As a user, I'd like the ability to safely re-index regardless of how many old versions I have and I might be cautious enough as to not try messing with other index versions because I may not have enough context to know why they're there.

search/includes/classes/commands/class-corecommand.php
```diff
@@ -57,7 +124,11 @@ protected function _maybe_setup_index_version( &$assoc_args ) {
 	 * [--version]
 	 * : The index version to index into. Used to build up a new index in parallel with the currently active index version
 	 *
-	 * @synopsis [--setup] [--network-wide] [--per-page] [--nobulk] [--show-errors] [--offset] [--start-object-id] [--end-object-id] [--indexables] [--show-bulk-errors] [--show-nobulk-errors] [--post-type] [--include] [--post-ids] [--ep-host] [--ep-prefix] [--version] [--skip-confirm]
+	 * [--using-versions]
```
rinatkhaziev, line 127: Let's add the description here.

search/includes/classes/commands/class-corecommand.php
```diff
@@ -14,38 +14,105 @@ class CoreCommand extends \ElasticPress\Command {
...
+		}
+
+		if ( array_key_exists( 'using-versions', $assoc_args ) && ! isset( $assoc_args['indexables'] ) ) {
+			return WP_CLI::error( 'The --indexables argument is required when specifying --using-versions' );
+		}
+
+		if ( array_key_exists( 'using-versions', $assoc_args ) && isset( $assoc_args['network-wide'] ) ) {
+			return WP_CLI::error( 'The --using-versions argument is not supported together with --network-wide' );
+		}
+	}
```
rinatkhaziev, line 34: WP_CLI::error doesn't return (it either exits or throws an exception, so this is a bit misleading.

rinatkhaziev: There are other instances of this that should be addressed too.

rinatkhaziev (changes requested):
Thanks for this, a significant QoL improvement, I left some comments inline here, but the most important part I think is the no-multiple-versions approach, I'm not sure it's the best one, the reasoning is in the inline comment.

search/includes/classes/commands/class-corecommand.php
pschoffer: It would actually fail without this check.

Firstly because we only allow 2 versions per indexable - https://github.com/Automattic/vip-go-mu-plugins/blob/b515aae5fe64d42f3bd924b37b466dbc20f3efb5/search/includes/classes/class-versioning.php#L21

And even if we would get around this. The current code is using "next" and "previous" to refer to the indexes when choosing which version to index, remove and activate. Which would break, if the user would already have "next" index before we start the process. I guess we could use numbers instead but that makes it slightly more complex. 

(mostly the first reason though)

search/includes/classes/commands/class-corecommand.php
pschoffer: Good point. I removed it from this file.

rinatkhaziev (approved):

PR #2178, opened 2021-06-02
Stricter validation for URLs in Search Dev Tools

rinatkhaziev:
## Description

This PR adds a strict validation for the URL argument for search/dev-tools REST endpoint. 

We used to do a fairly naive check - a valid URL ending in '_search'  would pass it. Theoretically, that can be exploited.

With PR validation is much stricter now.

We've added strict validation rules for the REST endpoint arguments

## Checklist

## Steps to Test

1.  Add `$query['url'] = 'somethingbogus'` around https://github.com/Automattic/vip-go-mu-plugins/blob/92f77df56e3c3a3268b9abd080f14b1cedae8680/search/search-dev-tools/search-dev-tools.php#L134
2. Open Dev Tools and try to edit a query
3. The remote request to the REST endpoint should be a 400 now.

### DIFF
--- a/tests/search/test-search-dev-tools.php
+++ b/tests/search/test-search-dev-tools.php
@@ -26,15 +26,19 @@ class Search_Dev_Tools_Test extends \WP_UnitTestCase {
 	public function data_provider_endpoint_urls() {
 		return [
 			[
-				'input'    => 'http://vip-search:9200/vip-123-v1/_search',
+				'input'    => 'http://vip-search:9200/vip-123-post-1/_search',
 				'expected' => true,
 			],
 			[
-				'input'    => 'http://vip-search:9200/vip-123-v1/_search',
+				'input'    => 'http://vip-search:9200/vip-123-post-1,vip-123-post-post-2-v1/_search',
 				'expected' => true,
 			],
 			[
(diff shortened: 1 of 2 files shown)


tests/search/test-search-dev-tools.php
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	public function data_provider_endpoint_urls() {
+		return [
+			[
+				'input'    => 'http://vip-search:9200/vip-123-v1/_search',
+				'expected' => true,
+			],
+			[
+				'input'    => 'http://vip-search:9200/vip-123-v1/_search',
```
nickdaugherty, line 33: Looks like this input is duplicated - did you mean to use a different endpoint than `_search`?

nickdaugherty (approved):
Nice, thank you! I left one minor note on a test case, otherwise should be good to go.

tests/search/test-search-dev-tools.php
rinatkhaziev: Oh that meant to catch the multiple indexes, another failed copypasta.

Thanks for the catch!

rinatkhaziev:
r1596-stacks

PR #2197, opened 2021-06-08
Adding checks for Jetpack Autoconnection staged rollout

pauarge:
## Description

We want to roll out Jetpack Autoconnection through Connection Pilot slowly. For now, we only attempt to connect a site to Jetpack if it's a new site.

- If it's a new site, we are picking an arbitrary GOOP Site ID and assuming everything after it is _new_.
- If it's a multisite, we can use our `is_fresh_site` logic.

We are safe to remove the changes introduced in this PR after the ramp up is complete.

### Plugin Updated: Jetpack

Preparing Jetpack Autoconnection ramp-up.

## Checklist

## Steps to Test

1. Check out PR.
2. Enable Connection Pilot on a fresh site and see that it autoconnects.
3. Enable Connection Pilot on an old site and see that it doesn't autoconnect.


vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
```diff
@@ -274,7 +274,17 @@ public static function should_run_connection_pilot(): bool {
 	 * @return bool True if a reconnect should be attempted
 	 */
 	public static function should_attempt_reconnection( \WP_Error $error = null ): bool {
-		// The constant is deprecated, but keeping this check for historical reasons
+		// TODO: Only attempting to reconnect on new sites. We can remove this code after ramp-up
+		if ( ! is_multisite() && defined( 'VIP_GO_APP_ID' ) && VIP_GO_APP_ID < 3750 ) {
```
rinatkhaziev, line 278: I think this needs to be the other way around? 
```suggestion
		if ( ! is_multisite() && defined( 'VIP_GO_APP_ID' ) &&  3750 < VIP_GO_APP_ID  ) {
```

rinatkhaziev (changes requested):
There seems to be a typo, in the conditional, other than that looks good!

vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
pauarge: I think it is correct as is. The function returns `false` in case we don't want to attempt reconnection. In this case, we only want to attempt it for "new" sites, that is, sites that have an ID bigger than 3750.

rinatkhaziev: ahh, I misread the next line :) Please disregard.

rinatkhaziev (approved):
🚀

rinatkhaziev (approved):
🚢

PR #1794, opened 2020-10-01
Improve security

brettshumaker:
Improve security of login/password reset limits.

PR #2200, opened 2021-06-09
Improve security

brettshumaker:
Improve security.

Supersedes #1794


security.php
```diff
@@ -202,18 +216,45 @@ function wpcom_vip_username_is_limited( $username, $cache_group ) {
...
+		 * @param string $username Username of the password reset request
+		 */
+		$ip_username_threshold = apply_filters( 'wpcom_vip_ip_username_password_reset_threshold', 3, $ip, $username );
+
+		/**
+		 * Password Reset IP Threshold
+		 *
+		 * @param string $ip IP address of the password reset request
+		 */
+		$ip_threshold = apply_filters( 'wpcom_vip_ip_password_reset_threshold', 30, $ip );
```
pauarge, line 244: Are we changing the default value from `3` to `30`?

pauarge (changes requested):

security.php
brettshumaker: Good catch! I may have changed this to match the 10x difference between `$ip_username_threshold` and `$ip_threshold` when rate-limiting login attempts [here](https://github.com/Automattic/vip-go-mu-plugins/blob/master/security.php#L196-L203).

Since I can't find where I discussed this with anyone else, I'm happy to change this back to `3` since I don't think it has artificially hindered anyone yet and we are providing a filter to change it.

pauarge (approved):
LGTM

PR #2207, opened 2021-06-11
Adding VaultPress autoconnection

pauarge:
This PR extends Jetpack Connection Pilot capability to add autoconnection to VaultPress. If a Jetpack connection is active, the system will try to connect automatically to VaultPress.

## Description

This PR extends Jetpack Connection Pilot capability to add autoconnection to VaultPress. If a Jetpack connection is active, the system will try to connect automatically to VaultPress.

### VaultPress auto connection

Connection Pilot is now capable of autoconnecting to VaultPress.

## Checklist

## Steps to Test

1. Check out PR.
2. Go to a Sandbox and launch a CLI.
3. Run the `reconnect()` routine from Connection Pilot.

### DIFF
--- a/vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
+++ b/vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
@@ -130,12 +130,24 @@ class Controls {
 		return self::jetpack_is_connected();
 	}
 
+	/**
+	 * Connect a site to VaultPress.
+	 *
+	 * Uses VaultPress' function to connect VaultPress using Jetpack. An active Jetpack connection is required on the site.
+	 *
+	 * @return bool|\WP_Error True if site is connected, error otherwise.
+	 */
 	public static function connect_vaultpress() {
 		if ( class_exists( 'VaultPress' ) ) {
-			return \VaultPress::init()->register_via_jetpack( true );
+			$vaultpress = \VaultPress::init();
+			if ( ! $vaultpress->is_registered() ) {
+				return $vaultpress->register_via_jetpack( true );
+			}
+
+			return true;
 		}
 
-		return false;
+		return new \WP_Error( 1, __( 'VaultPress could not be found.' ) );
 	}
 
 	/**
--- a/vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
+++ b/vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
@@ -130,6 +130,10 @@ class Controls {
 		return self::jetpack_is_connected();
 	}
 
+	public static function connect_vaultpress() {
+		return VaultPress::init()->register_via_jetpack( true );
+	}
+
 	/**
 	 * Provision the site with WP.com.
 	 *
--- a/vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
+++ b/vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
@@ -131,7 +131,11 @@ class Controls {
 	}
 
 	public static function connect_vaultpress() {
-		return VaultPress::init()->register_via_jetpack( true );
+		if ( class_exists( 'VaultPress' ) ) {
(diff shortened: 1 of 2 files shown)


vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
```diff
@@ -130,6 +130,10 @@ public static function connect_site( $skip_connection_tests = false, $disconnect
 		return self::jetpack_is_connected();
 	}
 
+	public static function connect_vaultpress() {
+		return VaultPress::init()->register_via_jetpack( true );
```
rinatkhaziev, line 134: Should we add a `class_exists` check here too?

vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
```diff
@@ -149,12 +149,20 @@ public function reconnect() {
...
 				$this->send_alert( 'Alert: Jetpack was automatically reconnected, but the connection may have changed cache sites. Needs manual inspection.' );
-
 				return;
 			}
 
 			$this->send_alert( 'Jetpack was successfully (re)connected!' );
 
+			$vaultpress_connection_attempt = Connection_Pilot\Controls::connect_vaultpress();
+
+			if ( is_wp_error( $vaultpress_connection_attempt ) ) {
```
pschoffer, line 159: Maybe log/alert the error? Or does that get logged already ?

pschoffer (changes requested):

vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
pauarge: We need to rework how logging/alerting in Connection Pilot. Currently, we simply send Slack alerts like this when something succeeds or not. 

You are right that in this specific case we could be logging way more information. I'd rather do it like this in the PR, since everything else in CP works similarly. After that, we can rethink how we want to do it (it should be a combination of Slack and Logstash).

WDYT @WPprodigy

pschoffer: OK sounds good. Otherwise looks good. So I will approve this, and feel free to address it in later PR.

pschoffer (approved):

vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
```diff
@@ -149,12 +149,20 @@ public function reconnect() {
...
+				$this->send_alert( 'Alert: Could not connect VaultPress automatically.' );
```
WPprodigy, line 160: Can pass the wp_error as the second arg here, then the logging will give some more details.

Since it's a VP connection error, can make this line more generic: https://github.com/Automattic/vip-go-mu-plugins/blob/af245ce05ced3135c287f92bd992eb90ce28faf7/vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php#L251

WPprodigy:
Will need to give it some good testing, as I know in the past it does not always play nice if JP was just connected for the first time. Usually takes a ~minute before VP can connect this way.

vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
```diff
@@ -149,12 +149,20 @@ public function reconnect() {
...
+				return;
+			}
+
+			$this->send_alert( 'VaultPress was successfuly connected!' );
```
WPprodigy, line 164: I think we can do without this being an alert. Knowing when it fails should be sufficient.

pauarge:
@WPprodigy I've slightly changed the implementation. With the current one, it should try to connect once JP is already connected. It might take up to an hour, but we make sure it's fine. 

We can gather feedback on whether that it's too slow or not, but I think it's fine given that those components are not as critical as Jetpack.

pschoffer (approved):

PR #2208, opened 2021-06-11
Remove index removal

pschoffer:
## Description

DROP INDEX `IF EXISTS` is not supported on Vitess or mysql. Looking at it closer I think it is safe to remove those lines totally (as oppose to manually check for index existence before removal).

The idea there was that version 2 of vip-search DB queue used and created it, however the current version 3 no longer creates or uses this DB index (switched around a year ago).

As this code ultimately runs on every admin/CLI request it is reasonable to assume the old index is cleaned out already.  

@nickdaugherty could you please check if my assumptions make sense to you?

### Plugin Updated: Search

Removes legacy DB index cleanup that is no longer needed.

## Checklist

## Steps to Test

IF you want to execute this code:
1.  On `search/includes/classes/queue/class-schema.php:111` change `_prepare_table` to be public
2. Run `dev-env exec -- wp eval "(new \Automattic\VIP\Search\Queue\Schema())->_prepare_table();"`

### DIFF
--- a/search/includes/classes/queue/class-schema.php
+++ b/search/includes/classes/queue/class-schema.php
@@ -139,12 +139,6 @@ class Schema {
 		$table_count = count( $wpdb->get_col( $wpdb->prepare( 'SHOW TABLES LIKE %s', $table_name ) ) );
 
 		if ( 1 === $table_count ) {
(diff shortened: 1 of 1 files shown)


nickdaugherty (approved):
Should be safe to remove as all schemas have since been upgraded.

PR #2218, opened 2021-06-14
Tweaking Connection Pilot cron job schedules

pauarge:
## Description

This PR tweaks how cron jobs are scheduled for Jetpack Connection Pilot so connections happen a bit quickly. Two modifications:

- First run goes down from ~1 minute after environment initialization to ~0 minutes after.
- Second run goes from between 1 and 60 minutes to 2 and 30 minutes. Note that subsequent runs are kept in an hourly schedule.

### Jetpack Connection Pilot schedules

Tweaked Jetpack Connection cron jobs schedules to improve auto connection times.

## Checklist

### DIFF
--- a/vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
+++ b/vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
@@ -90,7 +90,7 @@ class Connection_Pilot {
 
 	public function schedule_cron() {
 		if ( ! wp_next_scheduled( self::CRON_ACTION ) ) {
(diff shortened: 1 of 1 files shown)


WPprodigy (approved):
Looks good. Noting that for a new site, the internal crontab job is going to be registered off the bat I believe, here: https://github.com/Automattic/vip-go-mu-plugins/blob/029ef4db12c4468ac8a4f57945db5a2a23b50eb2/000-vip-init.php#L153-L157

So it's likely the `schedule_cron` change won't have an effect. That said, :shipit:

PR #2222, opened 2021-06-16
Decreasing minimum site id for Connection Pilot

pauarge:
## Description

We are performing some tests on Connection Pilot with Akismet and we'd like to test with some sites that have a slightly lower site id.


pschoffer (approved):
LGTM

PR #2223, opened 2021-06-16
Forcing VIP user to perform Akismet autoconnect

pauarge:
## Description

Akismet autoconnection would fail because the Jetpack connection would not be valid. That is because the Jetpack connection is owned by the WP VIP user. This PR forces that user when the connection attempt is performed.

### Jetpack Connection Pilot

Fix bug for Akismet autoconnection.

## Checklist

### DIFF
--- a/vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
+++ b/vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
@@ -141,8 +141,8 @@ class Controls {
 		if ( class_exists( 'Akismet_Admin' ) ) {
 
 			if ( is_akismet_key_invalid() ) {
-				// Getting wpcomvip user, since it's the owner of the Jetpack connection
 				$current_user = wp_get_current_user();
+				// Getting wpcomvip user, since it's the owner of the Jetpack connection
 				$vip_user = get_user_by( 'login', 'wpcomvip' );
 				if ( ! $current_user || ! $vip_user ) {
 					return false;
--- a/vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
+++ b/vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
@@ -138,17 +138,22 @@ class Controls {
 	 * @return bool True if connection worked, false otherwise
 	 */
 	public static function connect_akismet(): bool {
-		$current_user = wp_get_current_user();
-		$vip_user = get_user_by( 'login', 'wpcomvip' );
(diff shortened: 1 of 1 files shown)


vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
```diff
@@ -138,12 +138,20 @@ public static function connect_site( bool $skip_connection_tests = false, bool $
 	 * @return bool True if connection worked, false otherwise
 	 */
 	public static function connect_akismet(): bool {
-		if ( class_exists( 'Akismet_Admin' ) ) {
+		$current_user = wp_get_current_user();
+		$vip_user = get_user_by( 'login', 'wpcomvip' );
```
pschoffer, line 142: This does the user lookup, which might slow down sites that don't have Akismet as you do this lookup before you check the class exists.

Will it keep trying on every request, or is that scenario valid?

I guess we could load the wpcomvip user only when we need it like few lines bellow

pschoffer (changes requested):
Not sure if the get_user_by could not have a performance implications? More info in code comment

vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
pauarge: I think it's indeed better to do the check after `is_akismet_valid`, since it is only in that case where we need to do the user switch.

vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
```diff
@@ -139,11 +139,24 @@ public static function connect_site( bool $skip_connection_tests = false, bool $
 	 */
 	public static function connect_akismet(): bool {
 		if ( class_exists( 'Akismet_Admin' ) ) {
+
 			if ( is_akismet_key_invalid() ) {
-				return \Akismet_Admin::connect_jetpack_user();
+				// Getting wpcomvip user, since it's the owner of the Jetpack connection
+				$current_user = wp_get_current_user();
```
pschoffer, line 145:
```suggestion
                                $current_user = wp_get_current_user();
				// Getting wpcomvip user, since it's the owner of the Jetpack connection
```

pschoffer: Switching the lines to have comment just where you are getting the user:)

pschoffer (changes requested):

pschoffer (approved):
Amazing!

PR #2229, opened 2021-06-18
Update validate content logic to correctly handle missing keys 

pschoffer:
## Changelog Description

Validate content tries to build a Elastic Search document from post and compare it with the one actually stored in Elastic Search.
Before this change, the code would ignore any keys missing in either indexed or the newly build documen. This change makes these differences to be acknowledged and potentially used to trigger auto-healing process.  

An example of this is a new meta key being added to be indexed. After the change, we will recognize that there is a difference for all documents that should have but are missing that meta. And we will slowly reindex them as part of validate-content logic even without explicit re-index call.  

### Plugin Updated: Search

This change upgrades validate contents health check to also trigger on missing keys.

## Checklist

## Steps to Test

1.  (one of way to get inindexed key) register custom taxanomy:
```
register_taxonomy(
	'collections',
	['post', 'page'],
	[
		'show_admin_column' => true,
		'hierarchical' => true,
		'show_in_rest' => true,
		'public' => false,
		'sort' => false,
		'show_ui' => true,
	]
);
```
1. Add this taxanomy to a page/post in wp-admin
2. register taxanomy to be indexed (important to do AFTER you added the taxanomy)
```
add_filter(
	'vip_search_post_taxonomies_allow_list',
	function( $taxonomies ) {
		$taxonomies[] = 'collections';
		return $taxonomies;
	}
);
```
3. validate contents dev-env exec -- wp vip-search health validate-contents --start_post_id=5 --last_post_id=5


pauarge (approved):
LGTM

WPprodigy:
Think we need to make similar adjustments to `simplified_diff_document_and_prepared_document()` as well.

search/includes/classes/class-health.php
```diff
@@ -595,14 +595,27 @@ public static function filter_expected_post_rows( $rows, $post_types, $post_stat
...
+		$checked_keys = [];
+
 		foreach ( $document as $key => $value ) {
+			$checked_keys[ $key ] = true;
 			if ( in_array( $key, self::DOCUMENT_IGNORED_KEYS, true ) ) {
 				continue;
 			}
 
 			if ( is_array( $value ) ) {
-				$recursive_diff = self::simplified_diff_document_and_prepared_document( $document[ $key ], $prepared_document[ $key ] );
```
pschoffer, line 604: Cha turns out the simple one used to be very shallow too :) The `$recursive_diff` was ignored before.

rinatkhaziev:
@pschoffer can we add a little bit more context in the title and description, please? I.e. the "why" and the "how".

rinatkhaziev (approved):
👍

PR #2238, opened 2021-06-21
Specify blocking only applies to origin requests

davidsword:
## Description

This PR clarifies the usage of VIP_Request_Block class by explicitly spelling out that it only applies to origin and not edge.

As I don't believe there's docs on this class (just yet), best to be as explicit as possible here that this blocking does not block edge requests

### Library updated: VIP_Request_Block

We've updated a code comment in VIP_Request_Block to clarify the fact that the library is meant to be used to block abusive clients at the origin and not edge locations.


rinatkhaziev (approved):

pauarge (approved):
LGTM

rinatkhaziev:
r1642-stacks

PR #2239, opened 2021-06-21
Indexables::factory may return a boolean FALSE if an indexable is not  registered, which may result in Error: Uncaught exception 'Error'

rinatkhaziev:
## Description

In a certain edge case related to the load order of userland code Automattic\VIP\Search::disable_apc_for_ep_enabled_requests with message 'Call to a member function elasticpress_enabled() on bool' in Search\Cache::disable_apc_for_ep_enabled_requests.

This PR addresses that by checking for truthiness before trying to call 'elasticpress_enabled' method.

### Plugin Updated: Search

We've addressed an edge case that could potentially result in a fatal error in `Automattic\VIP\Search::disable_apc_for_ep_enabled_requests` pre_get_posts callback

## Checklist

### DIFF
--- a/search/includes/classes/class-cache.php
+++ b/search/includes/classes/class-cache.php
@@ -34,7 +34,9 @@ class Cache {
 			return;
 		}
 
-		if ( \ElasticPress\Indexables::factory()->get( 'post' )->elasticpress_enabled( $query ) && ! apply_filters( 'ep_skip_query_integration', false, $query ) ) {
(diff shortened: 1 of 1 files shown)


kuoko (approved):

kuoko:
@rinatkhaziev Thanks

pauarge (changes requested):
Could we add a unit test to check for this regression?

pauarge (approved):

rinatkhaziev:
r1643-stacks

PR #2240, opened 2021-06-22
Skipping VaultPress reconnect if it's disabled

pauarge:
## Description

We are getting plenty of `VaultPress could not be found` warnings in our Slack channel. Those are technically not an issue, because sites can safely opt-out of using ValutPress with a constant. 

Therefore, we want to reduce the noise produced by those alerts and get rid of them.

### Jetpack Connection Pilot and VaultPress

We are no longer treating a disabled VaultPress instance as a warning that triggers an alert.

## Checklist


vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
```diff
@@ -168,16 +168,12 @@ public static function connect_akismet(): bool {
 	 * @return bool|\WP_Error True if site is connected, error otherwise.
 	 */
 	public static function connect_vaultpress() {
-		if ( class_exists( 'VaultPress' ) ) {
-			$vaultpress = \VaultPress::init();
-			if ( ! $vaultpress->is_registered() ) {
-				return $vaultpress->register_via_jetpack( true );
-			}
-
-			return true;
```
pauarge, line 177: The only acceptable case in which VaultPress is not defined should be when it's skipped, which is already checked in the caller. If not, we'll fatal, but it is fine since we'd be detecting an edge case.

rinatkhaziev (approved):
Looks good.

I had to check whether the underlying methods return an instance of WP_Error in case of an error and it seems that's the case.

PR #2242, opened 2021-06-23
Add a temporary feature flag for VIP_PHPMailer until unit tests are added

rinatkhaziev:
## Description

Out of an abundance of caution and due to the fact the unit tests are missing for the feature we're gating it under `USE_VIP_PHPMAILER` constant.

### Component Updated: Mail 

We've released a compatibility fix for PHPMailer and VIP Filesystem so that VIP Filesystem attachments are allowed to be sent with an email.  

Currently, it requires `define( 'USE_VIP_PHPMAILER', true )` in vip-config.php

## Checklist

Please make sure the items below have been covered before requesting a review:

- [x] This change works and has been tested locally (or has an appropriate fallback).
- [x] This change works and has been tested on a Go sandbox.
- [ ] This change has relevant unit tests (if applicable).
- [ ] This change has relevant documentation additions / updates (if applicable).
- [x] I've created a changelog description that aligns with the provided examples.

## Steps to Test

Define the constant and follow the steps outlined in #2057

PR #2244, opened 2021-06-23
Don't include the VIP_PHPMailer on WordPress < 5.5, add test coverage

rinatkhaziev:
## Description

#2242 and #2057 introduced a regression on 5.4 due to the fact that PHPMailer paths were changed in 5.5, which we didn't catch because of not testing 5.4. 

This should address by checking the WP version and not requiring the new class if WordPress is older than 5.5

### Component updated: Mailer

We've disabled the new VIP_PHPMailer implementation on WordPress installs older than 5.5.

## Checklist

### DIFF
--- a/tests/test-vip-mail.php
+++ b/tests/test-vip-mail.php
@@ -14,7 +14,7 @@ class VIP_Mail_Test extends \WP_UnitTestCase {
 		}
 	}
 
-	protected function tearDown(): void {
+	public function tearDown(): void {
 		reset_phpmailer_instance();
 	}
 
--- a/tests/test-vip-mail.php
+++ b/tests/test-vip-mail.php
@@ -9,6 +9,9 @@ class VIP_Mail_Test extends \WP_UnitTestCase {
 	public function setUp() {
 		parent::setUp();
 		reset_phpmailer_instance();
+		if ( ! defined( 'USE_VIP_PHPMAILER' ) ) {
+			define( 'USE_VIP_PHPMAILER', true );
+		}
 	}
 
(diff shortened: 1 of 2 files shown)


rinatkhaziev:
Weirdly, the test is failing on 5.4 with $wp_version being 5.6 for some inexplicable reason :/ 

```
1) VIP_Mail_Test::test_load_VIP_PHPMailer_gte_55
VIP_PHPMailer should be loaded on >= 5.5. Version: 5.6
Failed asserting that true matches expected false.
```

tests/test-vip-mail.php
```diff
@@ -1,9 +1,17 @@
...
 	public function setUp() {
 		parent::setUp();
 		reset_phpmailer_instance();
+		if ( ! defined( 'USE_VIP_PHPMAILER' ) ) {
+			define( 'USE_VIP_PHPMAILER', true );
+		}
+	}
+
+	public function tearDown(): void {
+		reset_phpmailer_instance();
```
GaryJones, line 14: Why run this at the end of each test, when it's also run at the start of each test?

pschoffer: I tried removing it and it seemed fine for the specific test file. I pushed the change to run all the test pipelines.

pschoffer (approved):
LGTM

tests/test-vip-mail.php
rinatkhaziev: @GaryJones I mostly wanted to make sure the global state is not polluted AFTER the tests are done, is that not the right way to do it?

PR #2245, opened 2021-06-25
Safely update rewrite rule options

WPprodigy:
Long story short, replicating an error here requires a very specific set of events, and specific DB failures along the way with certain DB queries passing in between.

The easiest way to test would be to do a hack inside the patch area, and just trust me that it is possible to have such a WP Error at this point 🙃 :

```
// Safely save the precious
add_filter( "sanitize_option_{$option_key}", 'wpcom_vip_safeguard_rewrite_rule_updates', 999, 2 );

if ( $option_key === 'category_base' ) {
	$option_value = new WP_Error( 'wpdb_get_table_charset_failure' );
}

update_option( $option_key, $option_value );
remove_filter( "sanitize_option_{$option_key}", 'wpcom_vip_safeguard_rewrite_rule_updates', 999, 2 );
```

And if you want to see the bad-case, remove the filters and run again. Can use this to trigger the problem:

```
// Mimic what happens after a failed DB query.
	if ( ! isset( $_GET['viptestingrewrites'] ) || $_GET['viptestingrewrites'] !== 'flush' ) {
		return;
	}

	flush_rewrite_rules();
	do_action( 'rri_flush_rules' );
} );
```

Will also need something like the following in theme/plugin code beforehand to trigger all this:

```
wpcom_vip_load_permastruct( '/%category%/%post_id%/%postname%/' );
```

(Note: after running the failure case, you'll need to directly delete the `category_base` and `rewrite_rules` options from the database. Then comment out [this line](https://github.com/WordPress/WordPress/blob/94c655c89286fdf43e5bab82d4108fb679e46a7a/wp-includes/default-filters.php#L580) in WP core, followed by `wp cache flush`.)

### DIFF
--- a/vip-helpers/vip-permastructs.php
+++ b/vip-helpers/vip-permastructs.php
@@ -164,7 +164,7 @@ function wpcom_vip_refresh_wp_rewrite() {
 
 /**
  * Filter option values to prevent WP Errors from being saved,
- * and generated permanant fatal errors after rewrite rules are flushed.
+ * and generating permanent fatal errors after rewrite rules are flushed.
  *
  * @param string|WP_Error $value The new option value about to be saved.
  * @param string $option The option key.
--- a/vip-helpers/vip-permastructs.php
+++ b/vip-helpers/vip-permastructs.php
@@ -134,8 +134,12 @@ function wpcom_vip_refresh_wp_rewrite() {
 			$option_value = $default_value;
 		else
 			remove_filter( $filter, $callback, 99 );
-		// Save the precious
+
+		// Safely save the precious
+		add_filter( "sanitize_option_{$option_key}", 'wpcom_vip_safeguard_rewrite_rule_updates', 999, 2 );
 		update_option( $option_key, $option_value );
+		remove_filter( "sanitize_option_{$option_key}", 'wpcom_vip_safeguard_rewrite_rule_updates', 999, 2 );
+
 		// Only reapply the filter if it was applied previously
 		// as it overrides the option value with a global variable
 		if ( $reapply )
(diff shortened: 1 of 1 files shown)


WPprodigy:
Just noting here that ultimately this is something we should fix in WP core, the way it handles errors in `sanitize_option()` is faulty, particularly when a WP_Error is returned from core-functionality but has not error message attached to it. I'll follow up with this separately.

In the mean time, this will help mitigate.

vip-helpers/vip-permastructs.php
```diff
@@ -157,3 +161,25 @@ function wpcom_vip_refresh_wp_rewrite() {
...
+/**
+ * Filter option values to prevent WP Errors from being saved,
+ * and generating permanent fatal errors after rewrite rules are flushed.
+ *
+ * @param string|WP_Error $value The new option value about to be saved.
+ * @param string $option The option key.
+ * @return string
+ */
+function wpcom_vip_safeguard_rewrite_rule_updates( $value, $option ) {
+	$defaults = [
```
rinatkhaziev, line 174:
```suggestion
	// Pass through by default.
	if ( ! ( defined( 'VIP_SAFEGUARD_REWRITE_RULES' ) && true === VIP_SAFEGUARD_REWRITE_RULES ) ) {
		return $value;
	}
	
	$defaults = [
```

WPprodigy:
Going to go the route of patching this issue in WP core. Will link to the trac ticket here when available.

WPprodigy:
=> https://core.trac.wordpress.org/ticket/53986#ticket

WPprodigy:
Just noting that the above has been merged into core now.

PR #2254, opened 2021-06-29
Search dev tools: address a potential fatal error 'Uncaught Error: Cannot use object of type WP_Error as array' when an underlying ES query has timed out

rinatkhaziev:
## Description

We didn't account for the fact that if a query times out there will be an instance of WP_Error instead of an associative array. This resulted in a failure in the `array_map` callback when generating the data for the frontend.

This PR addresses by checking type checking the `$query['request']` and handling arrays and WP_Errors appropriately.  `else` condition will handle any unexpected variable types.

Additionally, uses optional chaining on the frontend when displaying the response code.

We've addressed a case where a timeout of the ES query resulted in a fatal error when printing the necessary data during `admin_footer/wp_footer` action callbacks.

## Checklist

## Steps to Test

1. Check out PR.
1. Cause a query to timeout somehow (either adjust the timeout to something really small, like `add_filter( 'http_request_timeout', function() {  return 0.01; } );` 
1. Verify that there's no fatal error and Search Dev Tools UI is rendered properly with an error message in the query result.


pauarge (approved):
LGTM

rinatkhaziev:
r1663-stacks

PR #2255, opened 2021-06-29
Allow to test all files with specific WP version

pauarge:
## Description

The bash script that we use to run unit tests locally would not work if the user wanted to test all files with a specific WordPress configuration. This PR addresses that and allows the user to input multiple optional parameters:

```
./bin/phpunit-docker.sh --file tests/path/to/test --version 5.4.4 --multisite 0
```

Also, we print more explicitly what options are being executed in each run.

Note that there's no need to deploy this PR, since it only contains changes to local development tooling.

### Allow to unit test all files with specific WP version

Updated internal tooling to better unit test the code.

## Checklist

## Steps to Test

1. Check out PR.
2. Run the testing script with multiple files/WP versions.


rinatkhaziev (changes requested):
Nice! 

We should update the readme too though. https://github.com/Automattic/vip-go-mu-plugins/blob/master/README.md#phpunit

README.md
```diff
@@ -38,20 +38,30 @@ If you're using the Lando-based environvment and it's already running, you can r
...
+./bin/phpunit-docker.sh
 ```
 
 You can either pass a version number to test against a specific version, or leave it blank to test against the latest version.
 
+```bash
+./bin/phpunit-docker.sh --version 5.4.4
+```
+
 You can also pass the path to a specific test as well as extra PHPUnit arguments:
```
rinatkhaziev, line 53: I don't think this line is true anymore because we only pass `--stop-on-failure` and the path to test with the new changes?

TBF, I'm not really sure whether arguments passthrough worked though.

rinatkhaziev (changes requested):
Left another minor thing, other than that looks good!

README.md
pauarge: Actually, that command _does_ work. We no longer pass the version as a parameter to the bootstrap, but that doesn't matter. Notice that the version is defined by what's mounted as a volume, which still uses the `$WP_VERSION` variable.

bin/phpunit-docker.sh
```diff
@@ -1,15 +1,51 @@
...
+# Test with explicit WP version and file: ./bin/phpunit-docker.sh --file tests/path/to/test --version 5.4.4
+# Test a specific file: ./bin/phpunit-docker.sh --file tests/path/to/test.php
 # Stop on failures: ./bin/phpunit-docker.sh --stop-on-failure
 # etc.
 
-WP_VERSION=${2-latest}
-WP_MULTISITE=${3-0}
+while test $# -gt 0; do
+  case "$1" in
+  --version)
```
mjangda, line 12: Thoughts on using `--wp` instead since `--version` is more commonly used to find out the version of a given script?

mjangda: Other than that, this looks great 👍  Named params are so much nicer to use :)

mjangda (approved):
🚢

PR #2258, opened 2021-06-30
adds concurency for search indexing queue processing

pschoffer:
### Plugin Updated: Search

There is an async queue for indexing in VIP search. This queue is used primarily on bulk operations over items (e.g. posts). This change enables the processing of this queue to run up to 5 instances concurrently. This will lead to having faster time for processing a bulk change of a large volume of documents. 

## Checklist

## Steps to Test

1. `dev-env exec -- wp eval "var_dump(apply_filters('a8c_cron_control_concurrent_event_whitelist', []));"`


rinatkhaziev (changes requested):
Great catch, one thing - let's follow convention and make it a separate method instead of using a closure.

search/includes/classes/queue/class-cron.php
```diff
@@ -62,6 +62,13 @@ public function init() {
 		add_action( self::SWEEPER_CRON_EVENT_NAME, [ $this, 'sweep_jobs' ] );
 		add_action( self::TERM_UPDATE_CRON_EVENT_NAME, [ $this, 'queue_posts_for_term_taxonomy_id' ] );
 
+		// Make job processing = indexing of documents run concurrently. This should help with handling spikes
+		// of bulk reindexing as well as keep cron's option used for queued jobs small, by processing them faster.
+		add_filter( 'a8c_cron_control_concurrent_event_whitelist', function( $whitelist ) {
+			$whitelist[ self::PROCESSOR_CRON_EVENT_NAME ] = self::MAX_PROCESSOR_JOB_COUNT;
```
WPprodigy, line 68: For some additional context, cron-control limits you to 10 total concurrent events as a default on each site. We should probably not take up half the queue w/ the default of 5 since this indexing can go on for quite a while. Perhaps something like:

```
$allowed_total_concurrency = ceil( Automattic\WP\Cron_Control\JOB_CONCURRENCY_LIMIT / 4 );
$whitelist[ self::PROCESSOR_CRON_EVENT_NAME ] = min( self::MAX_PROCESSOR_JOB_COUNT, $allowed_total_concurrency );
```

So the `MAX_PROCESSOR_JOB_COUNT` will never process much more than 25% of the overall concurrency limit. On larger sites that increase the total concurrency limit, eventually we can safely do all 5 at once. But by default, will do 3.

search/includes/classes/queue/class-cron.php
```diff
@@ -62,6 +62,8 @@ public function init() {
 		add_action( self::SWEEPER_CRON_EVENT_NAME, [ $this, 'sweep_jobs' ] );
 		add_action( self::TERM_UPDATE_CRON_EVENT_NAME, [ $this, 'queue_posts_for_term_taxonomy_id' ] );
 
+		add_filter( 'a8c_cron_control_concurrent_event_whitelist', [ $this, 'configure_concurency' ] );
```
WPprodigy, line 65: The filter has to be registered before  / up-to the `after_setup_theme` hook: https://github.com/Automattic/Cron-Control/blob/cde4f55e1b129fca22e27ad1310f9a2f7021e389/includes/class-events.php#L51

I'm _pretty_ sure we're good here, just noting.

WPprodigy (commented):
After further discussion about cron-control's inner workings, I'm thinking we want to make `MAX_PROCESSOR_JOB_COUNT` the dynamic factor here.

Currently, the sweeper will register say 5 processor jobs. A cron worker will pick up the first and start running. The next worker will pick up the next job, boot up WP, only to find the job is "locked" concurrency-wise. Another worker does the same for the next job, then another worker will try to pick up that 2nd one again, etc etc. Basically a bunch of workers are going to keep trying and trying to run these events rather wastefully.

So to prevent that, let's only ever register "just enough" jobs that it will be allowed to safely process at once, without application-side locks delaying things. The sweeper runs every minute anyways.

In short, `MAX_PROCESSOR_JOB_COUNT` can be made dynamic instead. So perhaps defaulting to 2 and going up to a max of 5 (based on \Automattic\WP\Cron_Control\JOB_CONCURRENCY_LIMIT). Then can whitelist for the same amount of concurrency.

This way we'll never be holding up cron workers needlessly.

pschoffer:
That's great idea @WPprodigy. If I got your input correctly you mean we should use the same limit for concurrency and number of scheduled jobs. I updated the code. Is this implementation that you had in mind?

WPprodigy (dismissed):
Yep - exactly that. Looks good!

rinatkhaziev:
@pschoffer Looks good! Nitpick, but could you please update all instances of `concurency` to `concurrency`?

pschoffer:
Thanks @rinatkhaziev , it's fixed.

pauarge (approved):

PR #2262, opened 2021-07-01
Search-dev-tools fix linting issues

pschoffer:
### Plugin Updated: Search

Fix minor linting issues in search-dev-tools panel.

## Checklist

## Steps to Test

1. Check out the PR
2. Fire up dev-env
3. Click around in the dev tools panel


search/search-dev-tools/src/components/information-pane/index.js
```diff
@@ -27,7 +27,7 @@ export const GeneralInformation = () => {
 /**
  * Represents a single collaplsible info block.
  *
- * @param {Object} props
+ * @param {Object} props including label,value and options
```
pauarge, line 30:
```suggestion
 * @param {Object} props including label, value and options
```

search/search-dev-tools/src/components/query/index.js
```diff
@@ -35,19 +34,19 @@ const Query = ( { args, request, url, query_args, backtrace = [] } ) => {
...
 
 	/**
-	 * 
+	 *
 	 * @param {Object} query the query to Run
-	 * @param {String} url ES url
+	 * @param {String} _url ES url
 	 */
-	const fetchForQuery = async ( query, url ) => {
+	const fetchForQuery = async ( query, _url ) => {
```
pauarge, line 41: What if we rename it to `es_url` or something like that? An underscore prefix might have some added meaning that we don't want

pauarge (changes requested):
Great effort, left a couple of minor things.

search/search-dev-tools/src/components/query/index.js
rinatkhaziev: I'm not even sure it's a good rule in this context TBH. We're not mutating the `url` nor we re-declaring it. My vote would be to skip this change, and instead slap the `eslint-disable-line no-use-before-define no-shadow` on both occurrences .

Additionally, it will break the useEffect callback on L64: https://github.com/Automattic/vip-go-mu-plugins/pull/2262/files#diff-a16e28a93d61fa9558c75f898bba81f5c500e9fccdec1e093a29ffa5027eed9dL65

rinatkhaziev (changes requested):
Thank you for addressing this! 

Left a minor comment.

search/search-dev-tools/src/components/query/index.js
pschoffer: I couldn't decide which way to go ... so I ignored you both and just deleted the code :)

(nah, actually as @rinatkhaziev mentioned we are not actually mutating the argument so there is no reason to not use the variable that is being overwritten I think, let me know what you think)

pauarge: Looks cleaner now, I like it.

pauarge (approved):

PR #2275, opened 2021-07-07
Use Logstash through the API

pauarge:
## Description

The current approach to use a file to send logs to Logstash [doesn't work on Kubernetes](https://vipsyseng.wordpress.com/2021/07/01/active-fixes-and-rollouts-for-kubernetes/#comment-4047). The alternative is to use the Logstash public API.

This PR reuses [some code that was originally implemented for that purpose](https://github.com/Automattic/vip-go-mu-plugins/pull/1499/files#diff-018f88dd9fe487c8446ee053e5da08a7dff7bd3b033d6d8370b5f52d0420f8cdR349). We are also introducing a constant to allow each site to configure which method they want to use.

### Use Logstash through the API

Adding the possibility to send Logstash logs through its API so it works in K8s.

## Checklist

### DIFF
--- a/logstash/class-logger.php
+++ b/logstash/class-logger.php
@@ -467,6 +467,8 @@ class Logger {
 			return; // Already done.
 		}
 
+		$fallback_error = new \WP_Error( 'logstash-send-failed', 'There was an error connecting to the logstash endpoint' );
+
 		static::$processed_entries = true;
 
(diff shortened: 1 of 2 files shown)


pauarge:
Looks like we can make it work with K8s. Closing the PR for now.

PR #2287, opened 2021-07-12
Fix non-existent notify_pilot function in Connection Pilot

pauarge:
## Description

Jetpack Connection Pilot would fail to log an error because it was calling a function that did not exist, `notify_pilot`. This PR replaces that call with the `send_alert` function, that is used elsewhere and does the same thing.

### Fix non-existent notify_pilot function in Connection Pilot

Fixes an edge case in which a log would not be sent due to calling a non-existent method.

## Checklist


WPprodigy (approved):
✅

PR #2292, opened 2021-07-16
Simplifies sweep cron

pschoffer:
## Description
This addresses a bug where we would keep scheduling a processor job on a busy cron if we would fail to get the number of already scheduled processor jobs before timeout.

### Plugin Updated: Search

We simplified logic around queued indexing of documents. The main motivation was to eliminate an edge case in which we could have scheduled more than 5 (the current limit) of indexing jobs. 

## Checklist

## Steps to Test

1. Have multiple index versions (change of post is queued for inactive version so we will take advantage of this to add something to a queue) - `vip dev-env exec -- wp vip-search index-versions add post`
2. Add a category to a post: `vip dev-env exec -- wp post term add 1 category new-category2` (this will index the active category right away, but will queue the inactive vesion, without scheduling processor)
3. Run `dev-env exec -- wp cron event list` to verify there is **no** `vip_search_queue_processor` job
4. Run `dev-env exec -- wp cron event run vip_search_queue_sweeper` (this will create a processor job)
5.  Run `dev-env exec -- wp cron event list` to verify there **is** `vip_search_queue_processor` job
4. Run again `dev-env exec -- wp cron event run vip_search_queue_sweeper` and see it doesnt add more processor jobs


pschoffer:
@WPprodigy this is the fix for the scheduling issue you found the other day. I would love to get your input on this!

search/includes/classes/queue/class-cron.php
```diff
@@ -212,29 +212,27 @@ public function sweep_jobs() {
 
 		$this->queue->free_deadlocked_jobs();
 
-		// If job limit has been surpassed, don't schedule any new jobs
-		if ( $this->get_processor_job_count() > self::MAX_PROCESSOR_JOB_COUNT ) {
-			return;
-		}
+		$job_count = $this->get_processor_job_count();
+		while ( ! is_wp_error( $job_count ) && $job_count < self::MAX_PROCESSOR_JOB_COUNT ) {
```
WPprodigy, line 216: Should this be `$job_count <= self::MAX_PROCESSOR_JOB_COUNT` now? If the max is scheduled, I don't believe we'd want to add another one.

search/includes/classes/queue/class-cron.php
```diff
@@ -212,29 +212,27 @@ public function sweep_jobs() {
...
+			if ( is_wp_error( $schedule_success ) || ! $schedule_success ) {
+				break;
 			}
-			
-			$this->schedule_batch_job();
 
-			$i++;
-		} while ( $this->get_processor_job_count() < self::MAX_PROCESSOR_JOB_COUNT );
-		// Prevent continuing if the current processor job count goes over the max allowed job count
+			$job_count = $this->get_processor_job_count();
```
WPprodigy, line 223: Would $job_count++ work here instead of a new query?

search/includes/classes/queue/class-cron.php
pschoffer: The logic is changed here a bit but this should result in scheduling 5 jobs if MAX_PROCESSOR_JOB_COUNT = 5.

Now it tries to add jobs if this is true (it used to exits here if $job_count > MAX).

IF current $job_count is 4, 4 < 5 => add one more

search/includes/classes/queue/class-cron.php
pschoffer: I thought about it and it probably would. But it would introduce a bigger change of race condition issue. This cron is not the only place the jobs is added so I figure it is safer to keep the query in as it was there before.

(The previous code used a local $i counter for some reason alongside with the `$this->get_processor_job_count() < self::MAX_PROCESSOR_JOB_COUNT` so it would also do the query on each iteration.)

pauarge (approved):

WPprodigy (approved):

PR #2302, opened 2021-07-20
Pre-set the `apikey` portion of empty `parsely` site options to `site_url` (if loaded)

jblz:
## Description

This hooks into the [`option_parsely` filter](https://developer.wordpress.org/reference/hooks/option_option/) when the wp-parsely plugin is loaded by this repo.

If an `apikey` is specified in the option, the filtering function does not alter the value.
If a `_wpvip_init_option` value is set in the option, the filtering function does not alter the value (this is set when this function runs for the first time while an `apikey` is not set).

During this initialization, we reference `site_url()` for a smart default for the `apikey`.

If `site_url()` includes a path beyond a simple `/` (e.g. for a subdirectory install), path parts are prepended to the domain to emulate a "subdomain" as slashes are not permitted in Parse.ly api keys. This is a standard workaround per @hbbtstar & @arhine.

### Plugin Updated: VIP Parse.ly Integration -- Pre-set API Keys

We made some behind-the-scenes changes that will help customers try Parse.ly on their sites more easily.

## Checklist

## Steps to Test

1. Check out PR.
1. Delete the `parsely` site option
1. Browse to WP-Admin > Settings > Parsely
1. Verify an appropriate value is present in the `Parse.ly Site ID` Settings field.


wp-parsely.php
```diff
@@ -46,3 +48,48 @@ function wpvip_load_wp_parsely_plugin() {
...
 add_action( 'after_setup_theme', 'wpvip_load_wp_parsely_plugin' );
+
+/**
+ * Pre-sets the "apikey" portion of the `parsely` site option hash map to the site_url IF it's empty.
+ * This is only hooked into `option_parsely` if we're loading the plugin from this repo.
+ *
+ * @param array $parsely_settings Settings from the database
+ * @return array Settings with the 'apikey' value overridden
+ */
+function wpvip_override_parsely_option_if_empty( $parsely_settings ) {
```
pauarge, line 59:
```suggestion
function wpvip_override_parsely_option_if_empty( array $parsely_settings ): array {
```

wp-parsely.php
```diff
@@ -46,3 +48,48 @@ function wpvip_load_wp_parsely_plugin() {
...
+	// Bail if an apikey is already set
+	if ( isset( $parsely_settings['apikey'] ) && strlen( $parsely_settings['apikey'] ) > 0 ) {
```
pauarge, line 61:
```suggestion
	if ( isset( $parsely_settings['apikey'] ) && ! empty( $parsely_settings['apikey'] ) ) {
```

pauarge (changes requested):
Looks good, left a couple of minor suggestions

wp-parsely.php
jblz: Thanks for the review, @pauarge!

Unless I'm missing something, I don't think we can do the `array` type declaration on the argument as an uninitialized option will result in an uncaught `TypeError` on newer PHP versions:

e.g.

```
# This is in PHP 7.4.21
php -r 'function func_that_takes_an_array( array $maybe_an_array = [] ) {} func_that_takes_an_array( false );'
PHP Fatal error:  Uncaught TypeError: Argument 1 passed to func_that_takes_an_array() must be of the type array, bool given, called in Command line code on line 1 and defined in Command line code:1
```

I think the return type is safe, however.

wp-parsely.php
GaryJones: Empty already allows for it not being set, so the first part of the condition could be dropped as well.

Alternatively, bearing in mind https://beberlei.de/2021/02/19/when_to_use_empty_in_php_i_say_never.html, then it's more prescriptive as to what we specifically want to check for, which is that it's set, and has a string length greater than 0.

wp-parsely.php
GaryJones: In which case, as [per WP core](https://github.com/WordPress/WordPress/blob/39cb388519cc573b9fae6db2c3fe94a66bee0d41/wp-includes/option.php#L221), the `@param` should be `mixed`, not `array` on line 56.

pauarge: If the input type can be `mixed`, should we update the type on the doc comment above?

wp-parsely.php
pauarge: If we can drop the first condition, I think the code would look cleaner and simpler. However, I see your point on the article linked above, so either way works for me.

wp-parsely.php
GaryJones: Yes - that's line 56 :-)

wp-parsely.php
jblz: I'm usually in the `empty` camp, but wrote it this way since I know Gary prefers the more prescriptive approach and I figured he'd have 👀 on this PR 😄.

pauarge (approved):

PR #2293, opened 2021-07-16
Updating Elasticpress to 3.5.6

pauarge:
## Description

This PR fetches the latest version of our Elasticpress fork, which includes version 3.5.6.

**Please test carefully in different environments, this PR might break things.**

### Plugin Updated: Elasticpress 3.5.6

We upgraded Elasticpress to 3.5.6

## Checklist

Please make sure the items below have been covered before requesting a review:

- [ ] This change works and has been tested locally (or has an appropriate fallback).
- [ ] This change works and has been tested on a Go sandbox.
- [ ] This change has relevant unit tests (if applicable).
- [ ] This change has relevant documentation additions / updates (if applicable).
- [x] I've created a changelog description that aligns with the provided examples.

## Steps to Test

1. Check out PR.
2. Make sure Search is enabled.
3. Perform searches, reindexes, check the Elasticpress admin...

PR #2303, opened 2021-07-21
Fix broken EP hook after EP upgrade

pauarge:
## Description

The last EP upgrade (https://github.com/Automattic/vip-go-mu-plugins/pull/2293) introduced a new hook, `ep_delete_post` with a single argument (instead of the two arguments the generic `ep_delete_ . $slug` had): https://github.com/Automattic/ElasticPress/blob/35a7c4a22158b460f55ef684e53a3cac318df66f/includes/classes/Indexable/Post/SyncManager.php#L389 

For this reason, our `action__ep_delete_indexable` would fail to execute, because sometimes the second argument would not be passed. Given that there is no such `ep_delete_term` or `ep_delete_user` with one argument, we can assume that the second argument is `post`. We are adding that as a default parameter to the function to avoid a fatal PHP error.

### Fix broken EP hook after EP upgrade

We upgraded our VIP Search integration with Elasticpress to fix a broken hook.

## Checklist

## Steps to Test

1. Check out PR.
2. Try EP in general: perform an index, add, edit, remove, posts, etc.

### DIFF
--- a/search/includes/classes/class-versioning.php
+++ b/search/includes/classes/class-versioning.php
@@ -786,8 +786,11 @@ class Versioning {
 	 * since deletes are comparatively much less frequent, so the savings of preventing back-and-forth switching between
 	 * index versions is not as great. We also process the deletes immediately, b/c the queue system currently does not
 	 * support deletes (just indexing)
+	 *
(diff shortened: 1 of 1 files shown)


nickohrn:
Thank you for filing this issue as my local development environment was affected this morning and was about to file one. Your description is much better than mine would have been.

brandon-m-skinner (approved):

PR #2305, opened 2021-07-23
Remove Jetpack 9.6

pauarge:
## Description

Removing Jetpack 9.6 from the list of available versions of Jetpack. 9.6 is an old version that is no longer used in our platform.

### Plugin Removed: Jetpack 9.6

Removing Jetpack 9.6 from VIP Go.

## Checklist


GaryJones:
Are there any sites still using 9.5 or below?

Removing a higher level version means that a customer who is behind on updating loses a stepping stone of where to update to, abs being forced to update to an even higher version may also mean needing to increase a WP version as well.

pauarge:
@GaryJones Yes, there (very few) sites running Jetpack versions older below 9.6. 

See that's already the case in the repository that we don't have some intermediate versions (9.2, 9.3). I see your point of 9.6 being a stepping stone, but if that's the case, client sites can always install that specific version themselves. The idea behind removing 9.6 (and ideally, others) is to discourage people from using unsupported versions.

GaryJones:
> client sites can always install that specific version themselves. 

A good point, which I'd forgotten.

rinatkhaziev (approved):

PR #2306, opened 2021-07-23
Fix VIP stats functions when Jetpack stats module is disabled

pauarge:
## Description

Supersedes https://github.com/Automattic/vip-go-mu-plugins/pull/2266.

The functions `wpcom_vip_top_posts_array` and `wpcom_vip_get_post_pageviews` would fail if the Jetpack stats module was disabled. We now check if it is enabled before proceeding.

In the cases in which Jetpack was not found, we would return _fake_ data. We now return empty values and we log the error.

### Fix VIP stats functions when Jetpack stats module is disabled

This functions now return data.

## Checklist

## Steps to Test

_Taken from Raam's PR_

1. Before applying this patch,  programmatically disable the Jetpack Stats module and then use `wpcom_vip_top_posts_array()` or `wpcom_vip_get_post_pageviews()`. The following code snippet can be added to an MU plugin for testing:

```php
<?php
// Disable the Stats Jetpack Module
function disable_jetpack_modules( $modules ) {
	$disabled_modules = ['stats'];

	foreach ( $disabled_modules as $module_slug ) {
		$found = array_search( $module_slug, $modules, true );
		if ( false !== $found ) {
			unset( $modules[ $found ] );
		}
	}
	return $modules;
}
add_filter( 'jetpack_active_modules', 'disable_jetpack_modules', 99, 9 );

// Use wpcom_vip_get_post_pageviews() to show Post Views at the bottom of Single Posts content
function filter_the_content_in_the_main_loop( $content ) {
    if ( is_singular() && in_the_loop() && is_main_query() ) {
        return $content . esc_html__( 'Post Views: ' . wpcom_vip_get_post_pageviews( get_the_ID(), 1 ) );
    }

    return $content;
}
add_filter( 'the_content', 'filter_the_content_in_the_main_loop', 1 );
```
2. Verify that this PHP Fatal error is thrown when viewing a single post:
```
PHP Fatal error:  Uncaught Error: Call to undefined function stats_get_csv()
```
3. Apply the patch in this PR and verify that the error is no longer thrown (the functions silently fall back to using fake data as expected).


raamdev (approved):
This looks good to me. 

I just wanted to leave a note here to clarify that after this change is released, anyone using `wpcom_vip_top_posts_array()` who has Jetpack disabled or the Jetpack Stats module disabled will go from seeing real posts returned (but fake 'top posts' data) to seeing nothing returned, which may be slightly unexpected (e.g., may result in Top Posts widgets being empty). 

That said, I agree that returning nothing and triggering an error to clarify that the function is not intended to work without Jetpack makes more sense than silently returning fake data and having the problem go unnoticed, which is the case before this change.

vip-helpers/vip-stats.php
```diff
@@ -40,20 +40,8 @@ function wpcom_vip_top_posts_array( $num_days = 30, $limit = 10, $end_date = fal
...
-		for ( $i = 0; $i < $limit; $i++ ) {
-			shuffle( $words );
-			$posts[] = array(
-				'post_id' 			=> $i,
-				'post_title' 		=> ucfirst( implode( ' ', array_slice( $words, 2, mt_rand( 2, 5 ) ) ) ),
-				'post_permalink' 	=> add_query_arg( 'p', $i, home_url() ),
-				'views' 			=> mt_rand( 0, 20000 ),
-			);
-		}
+		trigger_error( 'Cannot call wpcom_vip_top_posts_array without both Jetpack and the Jetpack Stats module active.', E_USER_WARNING );
```
GaryJones, line 43:
```suggestion
		trigger_error( 'Cannot call wpcom_vip_top_posts_array() without both Jetpack and the Jetpack Stats module active.', E_USER_WARNING );
```

vip-helpers/vip-stats.php
```diff
@@ -106,8 +94,8 @@ function wpcom_vip_get_post_pageviews( $post_id = null, $num_days = 1, $end_date
 			wp_cache_set( $cache_key, $views, 'vip_stats', 3600 );
 		}
 	} else {
-		// If Jetpack is not present or not active, fake the data returned
-		$views = mt_rand( 0, 20000 );
+		trigger_error( 'Cannot call wpcom_vip_get_post_pageviews without both Jetpack and the Jetpack Stats module active.', E_USER_WARNING );
```
GaryJones, line 97:
```suggestion
		trigger_error( 'Cannot call wpcom_vip_get_post_pageviews() without both Jetpack and the Jetpack Stats module active.', E_USER_WARNING );
```

GaryJones (changes requested):
Minor suggested tweaks of the error message. Good otherwise.

GaryJones (approved):

PR #2332, opened 2021-08-05
Check if the SERVICES_AUTH_TOKEN is empty

netsuso:
## Description

Check if the SERVICES_AUTH_TOKEN is empty before sending data to the site details service.

If the `services_auth_token` meta is not defined for a site, this define will be an empty string. In this case, we don't want to send data to the site details service, as it will fail for sure.

### Preparations for the site details service

We are about to release the site details service to production, and we need to ensure we can do it gradually, allowing to test some sites first. This PR allows that to happen

## Checklist

### DIFF
--- a/config/class-site-details-index.php
+++ b/config/class-site-details-index.php
@@ -202,7 +202,7 @@ class Site_Details_Index {
 	public function put_site_details() {
 		$site_details = $this->get_site_details();
 
-		if ( defined( 'SERVICES_API_URL' ) && defined( 'SERVICES_AUTH_TOKEN' ) ) {
+		if ( defined( 'SERVICES_API_URL' ) && defined( 'SERVICES_AUTH_TOKEN' ) && ! empty( SERVICES_AUTH_TOKEN ) ) {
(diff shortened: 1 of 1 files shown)


config/class-site-details-index.php
```diff
@@ -202,7 +202,7 @@ public function get_site_details() {
 	public function put_site_details() {
 		$site_details = $this->get_site_details();
 
-		if ( defined( 'SERVICES_API_URL' ) && defined( 'SERVICES_AUTH_TOKEN' ) ) {
+		if ( defined( 'SERVICES_API_URL' ) && defined( 'SERVICES_AUTH_TOKEN' ) && ! empty( SERVICES_AUTH_TOKEN ) ) {
```
mjangda, line 205: Just to be extra safe should we check that `SERVICES_API_URL` is not empty as well?

netsuso: I thought about this... but this is a define that will be added to `vip-secrets.php`, so I think it should be quite reliable (i.e, an empty `SERVICES_API_URL` would be as bad as a wrong url on it)

mjangda (approved):

netsuso:
stacks-r1734

PR #2340, opened 2021-08-09
Enables support for enforcing index version

pschoffer:
### Plugin Updated: Search

When working on a new index version it is often useful to be able to run a search against the inactive version before making it active.

This change introduces support for `vip_search_<indexable>_version` query parameter. Using this will let the user choose against which version of index is query ran.

As an example in a scenario where we have version 1 - active we can search against version 2 by visiting:
```
https://my.site.com/?s=<search-term>&&vip_search_post_version=2
``` 
The parameter also supports aliases `active`, `next` and `previous`.

## Checklist

## Steps to Test
Follow the instructions described in changelog


search/includes/classes/class-search.php
```diff
@@ -1955,4 +1958,24 @@ public function log_ep_invalid_response( $request, $query ) {
...
+		$indexables = $this->indexables->get_all();
+
+		foreach ($indexables as $indexable) {
+			$query_force_version_argument = sprintf( self::QUERY_FORCE_VERSION_PATTERN, $indexable->slug );
+
+			if ( ! isset( $_GET[ $query_force_version_argument ] ) ) {
+				continue;
+			}
+
+			$normalized_version = $this->versioning->normalize_version_number( $indexable, $_GET[ $query_force_version_argument ] );
```
pschoffer, line 1972: `set_current_version_number` is doing `normalize_version_number` as well. The reason I explicitly called it here, is that It felt safer with regards to working directly with user input. This way we make sure we always validate the input even if the implementation is changed in `set_current_version_number`. 

But then again this might be a bit of overkill and we could drop these 4 lines making the code simpler.

What do you think?

rinatkhaziev: I think it's fine to keep it explicit like that.

rinatkhaziev (approved):
This is great :shipit:

PR #2343, opened 2021-08-10
Lock search algorithm to 3.5 for future proofing

rebeccahum:
## Description

> Search Algorithm Upgrade Notice: Version 3.5 includes a revamp of the search algorithm. This is a backwards compatibility break. If you'd like to revert to the old search algorithm, you can use the following code: add_filter( 'ep_search_algorithm_version', function() { return '3.4'; } );. The new algorithm offers much more relevant search results and removes fuzziness which results in mostly unwanted results for most people. If you are hooking in and modifying the search query directly, it's possible this code might break and you might need to tweak it.

We should lock down the version to 3.5 for the time being by default for users in case the algorithm gets updated in the future.

## Changelog Description
Enterprise Search: Lock ElasticPress search algorithm at 3.5

## Checklist


search/includes/classes/class-search.php
```diff
@@ -1955,4 +1958,15 @@ public function log_ep_invalid_response( $request, $query ) {
...
+
+	/**
+	 * Filter out default algorithm version to be used.
+	 * 
+	 * @param $default_algorithm_version string Algorithm version.
+	 * 
+	 * @return string
+	 */
+	public function filter__ep_search_algorithm_version( $default_algorithm_version ) {
+		return '3.5';
```
GaryJones, line 1970: The PR description says to lock to 3.4, yet this uses `3.5`?

rebeccahum: @GaryJones I should have updated the PR description (which I will do now!) — after some discussion with @rinatkhaziev, we're going to lock it at 3.5 to futureproof any surprises

rinatkhaziev:
@rebeccahum thanks for working on this, it seems there's a syntax error though:
> Parse error: syntax error, unexpected 'public' (T_PUBLIC) in /home/circleci/project/search/includes/classes/class-search.php on line 1975

search/includes/classes/class-search.php
```diff
@@ -521,6 +521,9 @@ protected function setup_hooks() {
 
 		// Log details of failed requests
 		add_action( 'ep_invalid_response', [ $this, 'log_ep_invalid_response' ], PHP_INT_MAX, 2 );
+
+		// Lock search algorithm to 3.5
+		add_filter( 'ep_search_algorithm_version', [ $this, 'filter__ep_search_algorithm_version' ], 9999 );
```
rinatkhaziev, line 526: What do you think about keeping it at a default priority?

rebeccahum: Makes sense.

pauarge (approved):

pschoffer:
:wave: @rebeccahum,
could you please keep the changelog entry format from the template (the title prefixed by `##` id missing), otherwise the automation picking these up is having a bad time :)

I will do this one manually, just an ask for the future.

PR #2345, opened 2021-08-11
WP-CLI: User Cleanup command

mjangda:
## Description

Adds a new WP-CLI command which will allow us to revoke user permissions by email. This is useful when we need to handle user departure from an internal team or a customer.

TODO:
- [x] Add confirmation step
- [x] Add output to WP-CLI command

### New WP-CLI Command: User Cleanup

We've introduced a new command to make it easier to offboard users from sites. Given one or more email address, permissions (including super admin) can be removed from users matching those email addresses.

```
vip @<app>.<env> -- wp vip user cleanup --emails="test@example.com,also@example.com"
```

## Checklist

## Steps to Test

1. Check out PR.
1. Generate several users using `wp user generate`
1. Run the cleanup `wp vip user cleanup --emails="<comma-separated-emails>"`
1. Verify users are removed.


vip-helpers/class-user-cleanup.php
```diff
@@ -0,0 +1,115 @@
...
+		} else {
+			$emails_raw = [ $emails_string ];
+		}
+
+		foreach ( $emails_raw as $email_raw ) {
+			$email_raw = trim( $email_raw );
+
+			$email_filtered = filter_var( $email_raw, FILTER_VALIDATE_EMAIL );
+
+			if ( empty( $email_filtered ) ) {
```
rinatkhaziev, line 27: Small nitpick, but since the filter_var's return signature is `bool|string` let's check against that.
```suggestion
			if ( false === $email_filtered  ) {
```
Alternatively can flip the logic and make a truthy comparison and push into the array there, two less LoC :)

vip-helpers/class-user-cleanup.php
```diff
@@ -0,0 +1,115 @@
...
+			$emails[] = $email_filtered;
+		}
+
+		return $emails;
+	}
+
+	public static function split_email( $email ) {
+		list( $email_username, $email_hostname ) = explode( '@', $email );
+
+		// Strip off everything after the +, if it exists
```
rinatkhaziev, line 40:
```suggestion
		// Strip off everything after the +, if it exists so that we can run a wildcard match
```

wp-cli/class-vip-user-cli-command.php
```diff
@@ -0,0 +1,76 @@
...
+	 * Remove access for a user from an environment.
+	 *
+	 * Allows lookup via email.
+	 *
+	 * ## OPTIONS
+	 *
+	 * [--emails=<emails>]
+	 * : comma-separated list of emails
+	 */
+	public function cleanup( $args, $assoc_args ) {
```
rinatkhaziev, line 19: I think a confirmation prompt would be nice here, just in case.

vip-helpers/class-user-cleanup.php
```diff
@@ -0,0 +1,115 @@
...
+		list( $email_username, ) = explode( '+', $email_username );
+
+		return [ $email_username, $email_hostname ];
+	}
+
+	public static function fetch_user_ids_for_emails( $emails ) {
```
rinatkhaziev, line 46: Thoughts on having this work in a strict match and wildcard modes?

It's not inconceivable for someone to try to use it to remove a specific email only to find out that other accounts were affected too.

vip-helpers/class-user-cleanup.php
mjangda: `empty` is probably better though in the off-chance that filter_var somehow returns `''`?

mjangda: But the "truthy" check might be better.

rinatkhaziev: I don't believe so. An empty string would result in `false` because it's not a valid email.

wp-cli/class-vip-user-cli-command.php
mjangda: This has been added.

vip-helpers/class-user-cleanup.php
mjangda: Going to pass on this. Since this is meant for offboarding, I think it's better to do a more thorough search. If someone wants to remove a specific user, they should use `wp user delete`.

mjangda:
```
$ wp user create user user@example.com
Success: Created user 2.

$ wp user create other other@example.com
Success: Created user 3.

$ wp user create user-fake user+fake@example.com
Success: Created user 4.

$ wp vip user cleanup --emails=test@example.com
This will remove all access for any users with email addresses that match the following:
- test@example.com
Are you sure? [y/n] y
Success: No users found for the specified email addresses.

$ wp vip user cleanup --emails=user@example.com
This will remove all access for any users with email addresses that match the following:
- user@example.com
Are you sure? [y/n] y
== Processing site http://vip-test-single.local (1) ==
- user (2)
- user-fake (4)

$ wp user get 2 --field=roles
<empty>

$ wp user get 1 --field=roles
administrator
```

There's a minor bug with Multisite where it processes the first site twice. Not sure why but will look.

mjangda:
```
$ wp site create --slug=testing
Success: Site 2 created: http://vip-test-multi.local/testing/

$ wp user create testing test@example.com --url=http://vip-test-multi.local/testing/
Success: Created user 41.

$ wp super-admin add testing
Success: Granted super-admin capabilities to 1 user.

$ wp super-admin list
wordpress
testing

$ wp vip user cleanup --emails=test@example.com
This will remove all access for any users with email addresses that match the following:
- test@example.com
Are you sure? [y/n] y
== Processing Super Admins ==
- testing (41)
== Processing site http://vip-test-multi.local (1) ==
- testing (41)
== Processing site http://vip-test-multi.local/testing (2) ==
- testing (41)
== Processing site http://vip-test-multi.local/testing (2) ==
- testing (41)

$ wp super-admin list
wordpress

$ wp user get testing --url=http://vip-test-multi.local/testing/ --field=roles
<empty>
```

wp-cli/class-vip-user-cli-command.php
```diff
@@ -0,0 +1,116 @@
...
+		}
+
+		$user_ids = User_Cleanup::fetch_user_ids_for_emails( $emails );
+
+		if ( empty( $user_ids ) ) {
+			return WP_CLI::success( 'No users found for the specified email addresses.' );
+		}
+
+		if ( is_multisite() ) {
+			$this->process_multisite( $user_ids );
```
mjangda, line 58:
```suggestion
			return $this->process_multisite( $user_ids );
```

mjangda:
> There's a minor bug with Multisite where it processes the first site twice. Not sure why but will look.

Was missing a return 🤦 Fixed in 0f584e3 (props @rinatkhaziev)

rinatkhaziev (approved):
Nice, thanks! Looks good!

PR #2348, opened 2021-08-11
Using id range instead of ids

pschoffer:
## Description

Currently, we store all the job_ids for the async search indexing queue as an argument for the cron instance. This is stored in autoloaded cron option. As this option is autoloaded on every request we would like to minimize the amount of data stored there.

Before the change, we would store an array of size up to 1000 per cron instance (up to 5). This change switches from using `[1,2,3,4,5]` to `1-5`.

### Plugin Updated: Search

This change optimizes the storage mechanism for cron processes in the async index queue.

## Checklist

## Steps to Test

One of the ways to get items to the queue is having a second index version. Updates to post then will be queued in `wp_vip_search_index_queue` for the inactive version.
1) Watch the table in DB - wp_vip_search_index_queue
2) Run sweeping job - `dev-env exec -- wp cron event run vip_search_queue_sweeper` (this will change status of chosen jobs)
3) Run the processing job - `dev-env exec -- wp cron event run vip_search_queue_processor` (this will reindex documents and remove rows from db)

Alternatively you can check that the changes got into the ES document after the processor.

### DIFF
--- a/search/includes/classes/class-queue.php
+++ b/search/includes/classes/class-queue.php
@@ -579,6 +579,9 @@ class Queue {
 		return $job;
 	}
 
+	/**
+	 * @deprecated this will no longer be used. The method is kept to ease deployment. See the usage for more information.
+	 */
 	public function get_jobs( $job_ids ) {
 		global $wpdb;
 
@@ -592,6 +595,26 @@ class Queue {
 		$escaped_ids = implode( ', ', $job_ids );
 
 		$jobs = $wpdb->get_results( "SELECT * FROM {$table_name} WHERE `job_id` IN ( {$escaped_ids} )" ); // Cannot prepare table name, ids already escaped. @codingStandardsIgnoreLine
+
+		return $jobs;
+	}
+
+	/**
+	 * Get the jobs using the lowest and highest job_id range
+	 */
+	public function get_jobs_by_range( $min_id, $max_id ) {
+		global $wpdb;
+
+		if ( ! $min_id || ! $max_id ) {
+			return array();
+		}
+
+		$table_name = $this->schema->get_table_name();
+
+		$min_job_id_value = intval( $min_id );
+		$max_job_id_value = intval( $max_id );
+
+		$jobs = $wpdb->get_results( "SELECT * FROM {$table_name} WHERE `job_id` >= $min_job_id_value AND `job_id` <= $max_job_id_value" ); // Cannot prepare table name, ids already escaped. @codingStandardsIgnoreLine
 
 		return $jobs;
 	}
@@ -612,11 +635,10 @@ class Queue {
 		$table_name = $this->schema->get_table_name();
 
 		// TODO transaction
-		// TODO only find objects that aren't already running
 
 		$jobs = $wpdb->get_results(
 			$wpdb->prepare(
-				"SELECT * FROM {$table_name} WHERE ( `start_time` <= NOW() OR `start_time` IS NULL ) AND `status` = 'queued' LIMIT %d", // Cannot prepare table name. @codingStandardsIgnoreLine
+				"SELECT * FROM {$table_name} WHERE ( `start_time` <= NOW() OR `start_time` IS NULL ) AND `status` = 'queued' ORDER BY `job_id` LIMIT %d", // Cannot prepare table name. @codingStandardsIgnoreLine
 				$count
 			)
 		);
--- a/search/includes/classes/class-queue.php
+++ b/search/includes/classes/class-queue.php
@@ -759,7 +759,7 @@ class Queue {
 	/**
 	 * We can't re-queue jobs that are already waiting in queue. We should remove such jobs instead.
 	 */
-	private function delete_jobs_on_the_already_queued_object( $deadlocked_jobs ) {
+	public function delete_jobs_on_the_already_queued_object( $deadlocked_jobs ) {
 		global $wpdb;
 
 		$table_name = $this->schema->get_table_name();
--- a/search/includes/classes/class-queue.php
+++ b/search/includes/classes/class-queue.php
@@ -614,7 +614,13 @@ class Queue {
 		$min_job_id_value = intval( $min_id );
 		$max_job_id_value = intval( $max_id );
 
-		$jobs = $wpdb->get_results( "SELECT * FROM {$table_name} WHERE `job_id` >= $min_job_id_value AND `job_id` <= $max_job_id_value" ); // Cannot prepare table name, ids already escaped. @codingStandardsIgnoreLine
+		$jobs = $wpdb->get_results(
+			$wpdb->prepare(
+				"SELECT * FROM {$table_name} WHERE `job_id` >= %d AND `job_id` <= %d",  // Cannot prepare table name. @codingStandardsIgnoreLine
+				$min_job_id_value,
+				$max_job_id_value
+			)
+		);
 
 		return $jobs;
 	}
--- a/search/includes/classes/queue/class-cron.php
+++ b/search/includes/classes/queue/class-cron.php
@@ -144,10 +144,18 @@ class Cron {
 	 *
 	 * This is the cron hook for indexing a batch of objects
 	 *
-	 * @param {array} $job_ids Array of job ids to process
+	 * @param {array} $options Containing max_id and min_id keys
 	 */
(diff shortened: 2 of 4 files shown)


search/includes/classes/class-queue.php
```diff
@@ -596,6 +599,26 @@ public function get_jobs( $job_ids ) {
...
+		if ( ! $min_id || ! $max_id ) {
+			return array();
+		}
+
+		$table_name = $this->schema->get_table_name();
+
+		$min_job_id_value = intval( $min_id );
+		$max_job_id_value = intval( $max_id );
+
+		$jobs = $wpdb->get_results( "SELECT * FROM {$table_name} WHERE `job_id` >= $min_job_id_value AND `job_id` <= $max_job_id_value" ); // Cannot prepare table name, ids already escaped. @codingStandardsIgnoreLine
```
rinatkhaziev, line 617: Looks like the comment on the #L641 is not accurate and the `$wpdb->prepare` is actually used? If that's the case, let's also use it here and remove the comment.

search/includes/classes/queue/class-cron.php
```diff
@@ -144,10 +144,18 @@ public function filter_cron_schedules( $schedule ) {
...
+	 * @param {array} $options Containing max_id and min_id keys
 	 */
-	public function process_jobs( $job_ids ) {
-		$jobs = $this->queue->get_jobs( $job_ids );
+	public function process_jobs( $options ) {
+
+		if ( ! array_key_exists( 'min_id', $options ) ) {
+			// This is a temporary fix to handle deployment correctly. To cover the case when
+			// the process job would be created with the full list of ids instead of min and max.
+			// It should be possible to remove within few minutes after deploy.
```
rinatkhaziev, line 154: Love the optimism here :)

search/includes/classes/class-queue.php
```diff
@@ -612,11 +635,10 @@ public function checkout_jobs( $count = 250 ) {
 		$table_name = $this->schema->get_table_name();
 
 		// TODO transaction
-		// TODO only find objects that aren't already running
 
 		$jobs = $wpdb->get_results(
 			$wpdb->prepare(
-				"SELECT * FROM {$table_name} WHERE ( `start_time` <= NOW() OR `start_time` IS NULL ) AND `status` = 'queued' LIMIT %d", // Cannot prepare table name. @codingStandardsIgnoreLine
+				"SELECT * FROM {$table_name} WHERE ( `start_time` <= NOW() OR `start_time` IS NULL ) AND `status` = 'queued' ORDER BY `job_id` LIMIT %d", // Cannot prepare table name. @codingStandardsIgnoreLine
```
rinatkhaziev, line 641: Mentioned above, it does seem the `prepare` is used and is the safer approach.

rinatkhaziev (changes requested):
Thank you! I left a couple of comments, otherwise looks good!

search/includes/classes/class-queue.php
pschoffer: I added the prepare statement. However the issue both here and #L641 has in common is that you can't prepare table name.

So you can not do `$wpdb->prepare("SELECT * FROM %s", "my-table")`. However if you inline the table name as here the coding standard complains. Hence the ignore statement and the comment

search/includes/classes/class-queue.php
pschoffer: Same as above, we still inline the table name, which can't be prepared, yet the coding standard complains if you have any inline variable in the query string.

search/includes/classes/class-queue.php
```diff
@@ -737,7 +765,7 @@ private function delete_jobs_on_the_same_object( $all_deadlocked_jobs ) {
 	/**
 	 * We can't re-queue jobs that are already waiting in queue. We should remove such jobs instead.
 	 */
-	private function delete_jobs_on_the_already_queued_object( $deadlocked_jobs ) {
+	public function delete_jobs_on_the_already_queued_object( $deadlocked_jobs ) {
```
pauarge, line 768: Why are we making this function public?

search/includes/classes/class-queue.php
```diff
@@ -596,6 +599,32 @@ public function get_jobs( $job_ids ) {
...
+	 * Get the jobs using the lowest and highest job_id range
+	 */
+	public function get_jobs_by_range( $min_id, $max_id ) {
+		global $wpdb;
+
+		if ( ! $min_id || ! $max_id ) {
+			return array();
+		}
+
+		$table_name = $this->schema->get_table_name();
```
pauarge, line 612: Given that we can't prepare the table name, would it make sense to further sanitize this variable?

search/includes/classes/class-queue.php
```diff
@@ -612,11 +641,10 @@ public function checkout_jobs( $count = 250 ) {
 		$table_name = $this->schema->get_table_name();
 
 		// TODO transaction
-		// TODO only find objects that aren't already running
 
 		$jobs = $wpdb->get_results(
 			$wpdb->prepare(
-				"SELECT * FROM {$table_name} WHERE ( `start_time` <= NOW() OR `start_time` IS NULL ) AND `status` = 'queued' LIMIT %d", // Cannot prepare table name. @codingStandardsIgnoreLine
+				"SELECT * FROM {$table_name} WHERE ( `start_time` <= NOW() OR `start_time` IS NULL ) AND `status` = 'queued' ORDER BY `job_id` LIMIT %d", // Cannot prepare table name. @codingStandardsIgnoreLine
```
pauarge, line 647: Do we need to have the tasks order by its id? I get that's the way it was, but unless we need to guarantee that property, we can save an expensive `ORDER BY`.

pauarge (changes requested):
Sick optimization! Left some nitpick issues in the comments

search/includes/classes/class-queue.php
pschoffer: Nah I added the Order by. For the lowest-highest range to work it needs to be ordered otherwise there could be conflicts between what different parallel processors will take. (one would have `1-10` from `[1,2,10]` and another one would have `5-20` from `[5,6,20]`, both of them would then try to process jobs `5,6,10`)

search/includes/classes/class-queue.php
pschoffer: We could yeah. But then again this is not a user-input and I think it is a pretty standard pattern we use all over (without sanitization). So I would lean towards not adding it. But I am happy to be convinced otherwise:)

search/includes/classes/class-queue.php
pschoffer: One of the tests (`test_free_deadlocked_jobs_handle_duplicates`) is mocking this function. It was throwing warnings before and kinda pass by mistake. So I made it public so it is easily mockable. I know not the best approach but as we dont really use inversion of controll it is hard to mock the dependencies that this function calls. It is way easier just to mock the function.

pauarge: If it's just for tests, we can use something similar to what I did here: https://github.com/Automattic/vip-go-mu-plugins/blob/14f6d02bb262fbbbe4952a12790d5caa2d525343/tests/vip-jetpack/connection-pilot/test-class-jetpack-connection-pilot.php#L8

search/includes/classes/class-queue.php
pauarge: It could be content that comes from a database or a third-party service (or any dynamic source, for that matter). In that case we should treat it as user input. 

In this case, it comes from a hardcoded string in a class, so I guess we're good.

search/includes/classes/class-queue.php
rinatkhaziev: ah I see, I misread the comment as "can't prepare table_name so won't be preparing at all"

search/includes/classes/class-queue.php
rinatkhaziev: @pauarge that's neat!

rinatkhaziev (approved):
Awesome 🎉  🚀

search/includes/classes/class-queue.php
pschoffer: Hmm, interesting trick:) I don't think it would work in this case though.

The test calls another public method, which in turn calls this previously private method. The way you suggested you operate on another ReflectionClass, so whatever config/mocking we would do there would not be used by the production code that calls this private methed..

PR #2349, opened 2021-08-12
Adding enhanced logging for ElasticPress and changing error name

pauarge:
## Description

This PR adds three additional fields to ElasticPress's error logs: `is_cli` (whether the error comes from a CLI or not), `reponse` (whole body of the response) and `request` (if the request is an error or not).

We are also removing the `vip_` prefix from the errors, as it was duplicated (the `log2logstash` function would add it as well).

### Adding enhanced logging for ElasticPress

Adds some additional logs for when an ElasticPress request fails.

## Checklist

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -861,7 +861,7 @@ class Search {
 
 			$this->logger->log(
 				'error',
-				'vip_search_http_error',
+				'search_http_error',
 				implode( ';', $error_messages ),
 				[
 					'is_cli' => $is_cli,
@@ -880,7 +880,7 @@ class Search {
 			$error_message = $response_error['reason'] ?? 'Unknown Elasticsearch query error';
 			$this->logger->log(
 				'error',
-				'vip_search_query_error',
+				'search_query_error',
 				$error_message,
 				[
 					'error_type' => $response_error['type'] ?? 'Unknown error type',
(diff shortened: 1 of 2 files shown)


search/includes/classes/class-search.php
```diff
@@ -870,13 +880,15 @@ public function ep_handle_failed_request( $response, $query, $statsd_prefix ) {
...
-				'vip_search_query_error',
+				'search_query_error',
 				$error_message,
 				[
 					'error_type' => $response_error['type'] ?? 'Unknown error type',
 					'root_cause' => $response_error['root_cause'] ?? null,
 					'query' => $query_for_logging,
 					'backtrace' => wp_debug_backtrace_summary(),
+					'is_cli' => $is_cli,
+					'request' => $encoded_request,
```
pschoffer, line 891: How do feel about adding the whole $response_body  here? I feel like we are getting these `Unknown Elasticsearch query error` quite a lot.

Or maybe add the body if there is no ` $response_error['reason']`? To avoid pollution?

pschoffer (approved):
This is great. I left another feature wish, but feel free to ignore :)

search/includes/classes/class-search.php
pauarge: I think we're fine by adding all the response all times. Actually, we're doing this PR especially for those unknown errors.

rinatkhaziev:
This is great, thanks @pauarge

pschoffer (approved):
Nice stuff

PR #2359, opened 2021-08-17
Search: Explicitly pass posts_per_page=1 instead of relying on the global default in health validate-counts command

rinatkhaziev:
## Description

When running a `wp vip-search health validate-counts` it may return 0 instead of actual count of objects in the db if query times out. The default value is `500` which doesn't quite work well for ranges that include the entire table.

Essentially this manifests as

```
❌  inconsistencies found when counting entity: post, type: attachment, index_version: 4 - (DB: 0, ES: 6804932, Diff: 6804932)
```

This PR addresses that by explicitly pass `posts_per_page=1` instead of relying on the global default.

We've fixed an issue that potentially may result in a wrong DB count when running `wp vip-search health validate-counts`

## Checklist

## Steps to Test

Unless you have a table with ten million posts, you'll have to trust me.

If you want to "trust but verify" ping me in Slack and I'll point you to the site where this can be tested.


pschoffer (approved):
LGTM tested locally on small site, but I trust you with the benefit on larger scale :)

PR #2374, opened 2021-08-23
Add more information to the stale queue alert

pschoffer:
## Changelog Description

Example for a plugin upgrade:

Add more information to the alert about stale async queue:

New message:
```
Average index queue wait time for application 200508 - http://vip-local.vipdev.lndo.site is currently 768 seconds. There are 2 items in the queue and the oldest item is 969 seconds old"
```

## Checklist

## Steps to Test

1) In automattic/vip-go-mu-plugins/search/includes/classes/class-search.php:1101 change the condition to:
```
if ( intval( $queue_stats->average_wait_time ) > self::STALE_QUEUE_WAIT_LIMIT || true ) {
```
1) In automattic/vip-go-mu-plugins/search/includes/classes/class-search.php:1110 print message:
```
var_dump(['message', $message]);
```
1) `dev-env exec -- wp eval "var_dump((\Automattic\VIP\Search\Search::instance())->maybe_alert_for_average_queue_time());"`


search/includes/classes/class-queue.php
```diff
@@ -1080,18 +1080,25 @@ public function get_average_queue_wait_time() {
...
+		$queue_stats = $wpdb->get_row(
+			// Cannot prepare table name. @codingStandardsIgnoreStart
 			$wpdb->prepare(
-				"SELECT FLOOR( AVG( TIMESTAMPDIFF( SECOND, queued_time, NOW() ) ) ) AS average_wait_time FROM $table_name WHERE 1" // Cannot prepare table name. @codingStandardsIgnoreLine
+				"SELECT
+					FLOOR( AVG( TIMESTAMPDIFF( SECOND, queued_time, NOW() ) ) ) AS average_wait_time,
+					FLOOR( MAX( TIMESTAMPDIFF( SECOND, queued_time, NOW() ) ) ) AS longest_wait_time,
+					COUNT( * ) AS queue_count
+				FROM $table_name
+				WHERE 1"
```
pauarge, line 1091: Do we need `WHERE 1` here? AFAIK this would translate to always true.

search/includes/classes/class-queue.php
```diff
@@ -1080,18 +1080,25 @@ public function get_average_queue_wait_time() {
...
 			)
+			// @codingStandardsIgnoreEnd
 		);
 
 		// Null value will usually mean empty table
-		if ( is_null( $average_wait_time ) || ! is_numeric( $average_wait_time ) ) {
+		if ( is_null( $queue_stats ) || ! is_object( $queue_stats ) ) {
 			return 0;
```
pauarge, line 1098: Maybe we could return `null` or `false`, now that the happy path returns an object.

search/includes/classes/class-queue.php
```diff
@@ -1080,18 +1080,25 @@ public function get_average_queue_wait_time() {
...
 		}
 
-		return intval( $average_wait_time );
+		return $queue_stats;
```
pauarge, line 1101: I see we are no longer checking for `is_numeric`. I suggest we'd do it here and also cast the result inside the function, in that way, no error handling has to be done from the caller.

pauarge (changes requested):
Some minor improvements on error handling.

search/includes/classes/class-queue.php
rinatkhaziev: We should make sure the return signature is consistent, with this PR `get_queue_stats` now can return either an integer 0 or an object of `average_wait_time, longest_wait_time, queue_count` which seems a bit explode-y.

In fact, I think the unhappy path will throw a warning now (same with null of course):

```
php > $b = 0;
php > intval( $b->my_unhappy_var );
PHP Warning:  Attempt to read property "my_unhappy_var" on int in php shell code on line 1
```

I think probably what makes sense from a defensive standpoint is to have those values set to 0 inside the `if` block.  I.e.
`$queue_stats = (object) [ 'average_wait_time' => 0, 'longest_wait_time' => 0, 'queue_count' => 0 ];`

pschoffer: Great point, fixed!

search/includes/classes/class-queue.php
pschoffer: Great point, fixed!

search/includes/classes/class-queue.php
rinatkhaziev: That's correct, but it's also a pattern that's used across WP, you can spot `WHERE 1` and `WHERE 1=1` in various parts of the core.

rinatkhaziev (approved):
LGTM

pauarge (approved):
LGTM

PR #2375, opened 2021-08-24
Search: add descriptions for the supported named arguments for the index command

rinatkhaziev:
## Description

It started as fixing the discrepancy between the ElasticPress command and the Enterprise Search command. We had `start-object-id` and `end-object-id` arguments which was deemed confusing when the PR was merged to the upstream, so it was swapped for `upper-limit-object-id` and `lower-limit-object-id` which is less confusing.  Unfortunately we didn't catch that during the merge.

While fixing it we realized that the command doesn't really explain what do all of those arguments mean so we decided to address that as well.

We've added detailed descriptions for each argument for the indexing command. 

Use `wp help vip-search index` to see them all.

## Checklist

## Steps to Test

1. Check out the PR
2. Run `wp help vip-search index`
3. Observe that arguments now have descriptions.

### DIFF
--- a/search/includes/classes/commands/class-corecommand.php
+++ b/search/includes/classes/commands/class-corecommand.php
@@ -173,6 +173,9 @@ class CoreCommand extends \ElasticPress\Command {
 	 * [--show-bulk-errors]
 	 * : displays the error message returned from Elasticsearch when a post fails to index (as opposed to just the title and ID of the post)
 	 *
+	 * [--skip-confirm]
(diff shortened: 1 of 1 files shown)


rinatkhaziev:
One caveat here, `--skip-confirm` skips our wrappers command checks but doesn't skip ElasticPress's confirmation prompts, something that we should address soon-ish.

pauarge (approved):
Nice catch!

rinatkhaziev:
r1777-stacks

PR #2384, opened 2021-08-26
Validate only the active indexables instead of trying post and user no matter what (and failing to validate term and comment)

rinatkhaziev:
## Description

Our `validate-counts` command tried to count posts and users, but the users are not typically active, resulting in `Cannot find indexable 'user', probably the feature is not enabled`. This is not particularly helpful and could be anxiety-inducing.

Now we switch iterating over only active indexables, which would include Term and Comment too.

We have updated `wp vip-search health validate-counts` command to only validate active Indexables.

## Checklist

## Steps to Test

1. Check out PR.
2. Make sure User feature is not enabled.
1. Run the command.
1. Behold, no more user-related message.


brandon-m-skinner (approved):
:shipit:

pauarge (approved):

PR #2405, opened 2021-09-06
Annotate the parsely option with meta_type => repeated_metas

jblz:
## Description

When running the plugin without UI, default to `meta_type`: `repeated_metas` (instead of `json_ld`).

This will result in printing the metadata in conventionally-named `<meta />` HTML tags instead of a structured schema inside a `<script type="json+ld" />` tag.

## Checklist


jblz:
@hbbtstar @pauarge  if y'all could pick this up, I'd really appreciate it.
I think all that remains is to add a test case or two in and around the suite here:
https://github.com/Automattic/vip-go-mu-plugins/blob/e2985f0837e05db1419b9ffed812f5041000894f/tests/parsely/test-mu-parsely-integration.php

pauarge (approved):

rinatkhaziev (approved):

PR #2411, opened 2021-09-08
Adding additional checks on correct Jetpack Connection

pauarge:
## Description

We were getting a lot of failed VaultPress and Akismet connection attempts due to us flagging some Jetpack connections are good when in reality, they were only partially connected. What that meant is that the `wpcomvip` was missing the user token in WPCOM's side.

This PR adds an additional check on the procedure to evaluate the existing connection using that logic.

### Adding additional checks on correct Jetpack Connection

Improving Jetpack Connection Pilot checks for an active connection.

## Checklist


rinatkhaziev (approved):
👍

PR #2417, opened 2021-09-09
Don't log the false returned from Queue Sweeper's schedule_batch_job

rinatkhaziev:
## Description

Don't log the false returned from Queue Sweeper's schedule_batch_job because it means that the queue is empty, instead only log a WP_Error which would mean a cron event schedule failed.

#2382 introduced some logging which was intended to log unsuccessful event scheduling attempts but due to a logic error it was also logging on `FALSE` which signifies an empty queue. An empty queue is a valid state and should not be logged.

We've removed logging from Queue Sweeper when the queue is empty.

## Checklist


WPprodigy (approved):

pauarge (approved):
LGTM

PR #2418, opened 2021-09-10
Remove Jetpack Search Menu if Enterprise Search is enabled

pauarge:
## Description

Jetpack 10 introduced a Jetpack Search marketing page in `wp-admin`. That page is not actionable if the client site has Enterprise Search enabled. 

Screen Shot 2021-09-10 at 10 12 37 AM

This PR removes that page if that's the case.

### Remove Jetpack Search Menu if Enterprise Search is enabled

We hide a Jetpack Search-related page if Enterprise Search is enabled.

## Checklist

## Steps to Test

1. Check out PR.
2. Make sure Jetpack and Enterprise Search are enabled.
3. You shouldn't be able to access the page above.


vip-jetpack/vip-jetpack.php
```diff
@@ -413,3 +413,18 @@ function add_jetpack_menu_placeholder(): void {
...
+	if ( defined( 'VIP_ENABLE_VIP_SEARCH' ) && true === VIP_ENABLE_VIP_SEARCH ) {
+		if ( class_exists( 'Jetpack_Search_Dashboard_Page' ) ) {
+			remove_submenu_page(
+				'jetpack',
+				'jetpack-search',
+			);
+		}
+	}
+}
+add_action( 'admin_menu', 'remove_jetpack_menu_search', 999 );
```
pschoffer, line 430: Awesome. Should we move here though ? - https://github.com/Automattic/vip-go-mu-plugins/blob/ab2e84eba1e548a518a53bf8c8a2ee6b394fd89c/search/search.php#L30

WPprodigy:
One potential use case to think through:

> I've just enabled VIP Search and am moving from JP Search. I now have VIP Search active, and want to deactivate JP Search.

If we automatically deactivate it, then all would be fine. But could be an annoyance if we make it hard to turn off.

pauarge:
@WPprodigy That is a fair point. However, I see in the snippet that Pavel shared just above that we disable Jetpack Search automatically if we have VIP Search enabled.

search/includes/classes/class-search.php
```diff
@@ -1283,6 +1283,20 @@ public function filter__jetpack_active_modules( $modules ) {
 		} ) ); // phpcs:ignore WordPress.WhiteSpace.DisallowInlineTabs.NonIndentTabsUsed
 	}
 
+	/**
+	 * Remove the marketing admin page for Jetpack Search if VIP Search is enabled
+	 */
+	function remove_jetpack_menu_search(): void {
+		if ( defined( 'VIP_ENABLE_VIP_SEARCH' ) && true === VIP_ENABLE_VIP_SEARCH ) {
```
rinatkhaziev, line 1290: I think this check here is redundant because the action is only getting added if the query integration is enabled, meaning Search is 100% enabled.

rinatkhaziev (approved):
Thank you, this looks good!

pschoffer (approved):

PR #2419, opened 2021-09-10
Prevent newrelic transaction name overrides during cron

WPprodigy:
## Description
Cron events in new relic were showing up with no transaction names, specifically when WooCommerce was active. This is because WC does the following when WP CLI is loaded, causing the filter we use to track rest requests to trigger and override the intended new relic settings that were setup for cron: https://github.com/woocommerce/woocommerce/blob/b19500728b4b292562afb65eb3a0c0f50d5859de/includes/cli/class-wc-cli-runner.php#L63

Also, since `rest_dispatch_request` is a filter, list it as such and just return the value rather than deciding it should be `null` (basically, ensure we don't override another hook that has already changed it.)

### New Relic: Fix transaction naming problems

Resolved issue with blank transaction names showing in New Relic, specifically when WooCommerce was active.

## Checklist

## Steps to Test

Kinda hard to test, since cron runs in it's own locations and not sandboxes where new relic doesn't log. But the logic is pretty straightforward :)


WPprodigy:
I also looked for a nice hook or filter we could use to grab the cron event name that is running so we could log that as a customer new relic paramater (similar to how we log what CLI command was run), but no viable hooks seem to currently exist. So perhaps in a future cron control update we could get that added in :)

vaurdan (approved):
Looks good!

rinatkhaziev:
r1826-stacks

PR #2420, opened 2021-09-10
Ignoring dev-env related file

pauarge:
## Description

If the `vip` CLI is used to spin up a local development environment, a file (`dev-env-plugin.php`) is created to make things work. We don't want that file to slip into production, hence we are ignoring it through `.gitignore`. 

### Ignoring dev-env related file

Internal code change to prevent conflicts with `vip dev-env`.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,7 @@
 jetpack-private.php
 
 # dev-env related files
-dev-env-plugin.php
+/dev-env-plugin.php
 
 # Local PHPDoc configuration file
 # overrides phpdoc.dist.xml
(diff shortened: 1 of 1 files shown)


WPprodigy:
Should we maybe do `./dev-env-plugin.php` to be specific about the location we want to ignore? Just in case (however unlikely) such a file is added somewhere much deeper like in a difference plugin sometime.

pauarge:
@WPprodigy I have applied your suggestion, but instead of `./dev-env-plugin.php` I've done `/dev-env-plugin.php` (it wouldn't work otherwise).

rinatkhaziev (approved):

rinatkhaziev:
No deploy needed.

PR #2424, opened 2021-09-15
Search: Health - validate counts: prevent potential fatal for term and comment indexables.

rinatkhaziev:
## Description

This is a hotfix for an issue introduced in #2384 which switched from hard coded user and post counts to all active indexables. 

Unfortunately, other indexables are not handled properly in `validate_indexable_count_for_site` method, resulting in 
`null` being passed to `render_results` which expects an array, thus causing an `Uncaught Type Error` fatal.

props @dhusakovic for the find

We've applied a hotfix to address a potential fatal error when running `wp vip-search health validate-counts` commands for Term and Comment indexables.

## Checklist

## Steps to Test

1. Check out PR.
1. Run the command on a site with Term or Comment Indexable enabled.
1. Verify there's no fatal anymore


brandon-m-skinner (approved):
That'll do it.

As an aside, we should only process indexables that we support for this. Maybe check against an allow list or something. Right now it just does:

```
~$ wp vip-search health validate-counts

Validating post count

✅ no inconsistencies found when counting entity: post, type: post, index_version: 1 - (DB: 1, ES: 1, Diff: 0)
✅ no inconsistencies found when counting entity: post, type: page, index_version: 1 - (DB: 1, ES: 1, Diff: 0)

Validating term count


~$
```

which also kinda looks something went wrong. Not blocking for this PR, but a fairly quick fix that will likely save some toil.

dhusakovic (approved):
Looks good! Solves the fatal error 👍

rinatkhaziev:
r1834-stacks

PR #2426, opened 2021-09-15
Removed legacy code which broke comment counts.

### LINKED ISSUE #1618: wpcom_vip_cache_full_comment_counts overrides Woocommerce comment count
## Expected/Desired Behavior

When viewing `/wp-admin/edit-comments.php` I would like the Woocommerce Order Notes excluded from the comment count. This is what happens by default in Woocommerce. https://github.com/woocommerce/woocommerce/blob/master/includes/class-wc-comments.php#L225

## Actual Behavior

`wpcom_vip_cache_full_comment_counts` overrides the results from the Woocommerce filter and returns the full comment count with Order Notes included.

1. Create a VIP site with Woocommerce
2. Place an order
3. Navigate to `/wp-admin/edit-comments.php` to see comment counts higher than the actual number of post comments or product reviews.

## Additional notes

When testing locally I updated `wpcom_vip_cache_full_comment_counts` to respect earlier priority filters by adding:
```
if ( ! empty( $counts ) ) {
    return $counts ;
}
```

Screenshot from production environment where comment count is over 24K and the comment table is displaying 2 comments.

Screen Shot 2020-06-10 at 3 12 22 PM

DustinHartzler:
## Description

Removed legacy code from WordPress.com days that cached the comment counts. The `wp_comment_count` function in core now caches the values and this code is no longer used.

### Fixed comment counts on dashboard

Removed legacy code that caused sites with WooCommerce to show the wrong count for total comments in the WordPress Dashboard.

## Checklist

- [ x ] This change works and has been tested locally (or has an appropriate fallback).
- [ x ] This change works and has been tested on a Go sandbox.
- [ x ] I've created a changelog description that aligns with the provided examples.

Closes #1618

### DIFF
--- a/performance/vip-tweaks.php
+++ b/performance/vip-tweaks.php
@@ -4,15 +4,6 @@
  * This is a list of performance tweaks that will become default for All VIP sites
  */
 function wpcom_vip_enable_performance_tweaks() {
-	/**
-	 * Improves performance of all the wp-admin pages that load comment counts in the menu.
-	 *
-	 * This caches them for 30 minutes. It does not impact the per page comment count, only
-	 * the total comment count that shows up in the admin menu.
-	 */
-	if ( function_exists( 'wpcom_vip_enable_cache_full_comment_counts' ) ) {
-		wpcom_vip_enable_cache_full_comment_counts();
-	}
 
 	// This disables the adjacent_post links in the header that are almost never beneficial and are very slow to compute.
 	remove_action( 'wp_head', 'adjacent_posts_rel_link_wp_head', 10, 0 );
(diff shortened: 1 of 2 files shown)


rinatkhaziev:
@DustinHartzler thanks for the PR, since this removes a function definition we need to check customer code first, if there's a code using that function. 
If there is, we'll have to adopt a more conservative approach and replace the function's body with a call to [_deprecated_function](https://developer.wordpress.org/reference/functions/_deprecated_function/)

Other than that, looks good!

I'll let you know once the check for that call runs through (will take some time)

rinatkhaziev:
it seems no customer code uses this function!

rinatkhaziev (approved):

DustinHartzler:
> since this removes a function definition we need to check customer code first, if there's a code using that function.

Good to know!

PR #2441, opened 2021-09-21
Cron: Action Scheduler Dynamic Queue

WPprodigy:
### What is Action Scheduler

[Action Scheduler](https://actionscheduler.org/) is an event-runner that can be heavily utilized by plugins like WooCommerce. "Actions" are scheduled and stored in the database, and then the "queue runner" picks up the actions in batches and processes through them. It is similar to how our cron works, except it more efficiently handles multiple actions within one request - making it ideal for running hundreds to thousands of "small" scheduled tasks (like order webhooks, emails, inventory syncs, etc).

The way the queue runner works by default in AS is through a recurring 60second cron job called `action_scheduler_run_queue`. It will claim and process say a batch of 25 actions repeatedly as long as it can given its time limit of 30 seconds. After which it will wait until the next iteration of the recurring cron job to go again. This alone has the main shortcoming of being pretty limited on concurrency.

There is also an "async runner" in the core plugin, which kicks off a queue once a minute off of a frontend request (`shutdown`). This is meant as a solution to normal WP cron being lacking, but it's not really necessary on our infrastructure - would be much better if we just let cron containers handle the background tasks.

Even with the async runner, it's still very easy for the queue to become backed up w/ high volume sites. Subscriptions are a common use case where thousands of tasks can become due all at once.

### Proposed Solution

Our cron runner is very well equipped to kick off multiple jobs at once, so let's utilize it :). This PR disables the async frontend queue in favor of scheduling one-time jobs when they are needed. If a high volume of actions hit or if the default recurring job starts to fall behind, this dynamic queue will jump in and dispatch additional cron jobs that are allowed to run concurrently with each other.

Rather than needing to tailor a bunch of filters per a site's needs, this scales directly off of `JOB_CONCURRENCY_LIMIT`. Allowing up to 33% of the concurrency limit. If that fails to be sufficient, then the site should have the concurrency limit increased at the application level at that point. But also notably, it is very lightweight in times where all the extra queues are not needed. So rather than having an unnecessary amount of recurring jobs that are always filling up the cron queue, it will only schedule jobs when it both needs them and when action scheduler has slots available and is ready to launch new queues.

Additionally, the default 30-second queue timeout limit is there because general WP hosts run cron on the frontend and thus it is limited by timeouts there. We can push this much farther on our cron containers, however we do not want to go too far as memory leaks become a concern if a single job tries process too many actions. So settling at a fairly-well tested 120seconds for now. This gives the default recurring queue and the additional queues a lot of efficiency, and lowers how often we need to worry about dispatching new queues.

### Cron: Introduce Dynamic Queue system for Action Scheduler

This new feature tailors Action Scheduler defaults and unleashes more processing power by syncing up with the Cron Control system. It will schedule additional AS queues to run concurrently when they are needed to process through a large list of "due" actions.

## Checklist

## Steps to Test

- Install this file as a client-mu-plugin and deploy to a test site, as you'll need it running live on real batch containers.
- Also add in this mini plugin, which will make it easy to set up a large backlog of actions: https://gist.github.com/WPprodigy/aebd49bc2f7f10bcb62dbe342948f924

1) Schedule 2000 actions that will run for 1 second each: `wp vip-action-scheduler-jammer --amount=2000 --duration=1`
2) Watch as over time the "In-progress" list of actions starts to increase, showing the concurrency: `/wp-admin/admin.php?page=wc-status&tab=action-scheduler&status=in-progress`. Can also `wp cron-control events list` and watch for the new events to be added (note that the dispatch is cached, so may need to wait for around a minute - else comment out the cache check & load a page on the sandbox).

(test site 3474 is available and setup for this 🙂 )


vip-woocommerce/action-scheduler.php
```diff
@@ -0,0 +1,166 @@
...
+	public function configure_action_scheduler_concurrency() {
+		// Increment by 1 to account for the default `action_scheduler_run_queue` recurring action.
+		return $this->get_max_allowed_queue_jobs() + 1;
+	}
+
+	private function get_max_allowed_queue_jobs() {
+		$total_cron_control_concurrency_limit = defined( '\Automattic\WP\Cron_Control\JOB_CONCURRENCY_LIMIT' ) ? \Automattic\WP\Cron_Control\JOB_CONCURRENCY_LIMIT : 10;
+
+		// Allow up to 33% of the cron-control queue to be used for action scheduler processing.
+		return (int) min( ceil( $total_cron_control_concurrency_limit / 3 ), self::MAX_ALLOWED_DYNAMIC_QUEUES );
```
pschoffer, line 73: I like this 33% percent approach. Smart.

vip-woocommerce/action-scheduler.php
```diff
@@ -0,0 +1,166 @@
...
+	 * Every few minutes, we determine if we should dispatch additional AS queues.
+	 */
+	public function maybe_dispatch_new_queues() {
+		$dispatch_interval = max( $this->get_queue_timeout_limit() - 10, 60 );
+		if ( ! wp_cache_add( 'dynamic-queue-scheduler-lock', 'locked', 'vip', $dispatch_interval ) ) {
+			// Only dispatch new queues around the time it takes to finish the previous round, or once a minute minimum.
+			return;
+		}
+
+		$pending_cron_jobs_count = $this->get_pending_queue_job_count();
```
pschoffer, line 91: So this `$pending_cron_jobs_count` is the number of queues as in Action Scheduler events that each will process jobs, correct? This naming is really hard to get right I think. It is easy to confuse it with the jobs that will be processed by the queue. 

What do you think about naming it `$current_queue_count` ?

pschoffer (changes requested):
Great stuff. I left few minor comments but it looks good to me otherwise

vip-woocommerce/action-scheduler.php
```diff
@@ -0,0 +1,166 @@
...
+	private function get_max_allowed_queue_jobs() {
+		$total_cron_control_concurrency_limit = defined( '\Automattic\WP\Cron_Control\JOB_CONCURRENCY_LIMIT' ) ? \Automattic\WP\Cron_Control\JOB_CONCURRENCY_LIMIT : 10;
+
+		// Allow up to 33% of the cron-control queue to be used for action scheduler processing.
+		return (int) min( ceil( $total_cron_control_concurrency_limit / 3 ), self::MAX_ALLOWED_DYNAMIC_QUEUES );
+	}
+
+	private function get_queue_timeout_limit() {
+		// Note: We set this to 120 seconds in a filter above.
+		return absint( apply_filters( 'action_scheduler_queue_runner_time_limit', 30 ) );
```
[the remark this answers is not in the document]
WPprodigy: Oh, it's not a filter I'm introducing. Is pre-existing: https://github.com/woocommerce/action-scheduler/blob/7f4aaf0e011bb0b41b266afd479f6935a495231b/classes/abstracts/ActionScheduler_Abstract_QueueRunner.php#L138

vip-woocommerce/action-scheduler.php
WPprodigy: Yeah, lots of shared wording.

Cron job = `vip_action_scheduler_run_queue` event
Each job starts up 1 AS queue.
Each AS queue processes actions.

____

I think the important factor to convey here though is that they are scheduled cron jobs that have yet to run. So it's not technically an AS queue yet, just pending cron jobs.

rinatkhaziev:
Would be nice to add some tests here

WPprodigy:
Yeah was thinking through how to best do that for this. I think will require some heavy mocking.

vip-woocommerce/action-scheduler.php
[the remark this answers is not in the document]
pschoffer: Oh got ya, so you want to have the same value in both places. Makes sense!

rinatkhaziev:
> Yeah was thinking through how to best do that for this. I think will require some heavy mocking.

Coincidentally, @pschoffer is our mocking champion.

rinatkhaziev:
This is getting a bit stale :)

vip-woocommerce/action-scheduler.php
WPprodigy: Added some extra documentation in the class docblock to help clarify this process.

WPprodigy:
1) Moved this functionality into a new `cron/` folder, as it's not actually WC specific. There have been other customers asking about this who are utilizing just Action Scheduler on their own for other tasks. The only dependencies are AS & Cron Control now, not WC.

2) In the future, I'm planning to move the other cron logic into this new `cron/` folder, and have `000-cron.php` just load things up. Perhaps even stick cron-control itself into the folder.

3) As for unit tests, I have a WIP version but it's very messy since this is tightly coupled w/ the Cron Control custom table (that doesn't exist in our tests). Rather than re-invent the wheel, going to wait until the new cron control changes are shipped that will allow more easier mocking / overriding of its Events Store method rather than forcing us to query the table directly in mu-plugins all the time.

001-cron.php
```diff
@@ -152,7 +156,7 @@ function wpcom_vip_log_cron_control_uncacheable_cron_option( $option_size, $buck
 	add_action( 'a8c_cron_control_event_threw_catchable_error', 'wpcom_vip_log_cron_control_event_for_caught_error', 10, 2 );
 	add_action( 'a8c_cron_control_freeing_event_locks_after_uncaught_error', 'wpcom_vip_log_cron_control_event_object' );
 	add_action( 'a8c_cron_control_uncacheable_cron_option', 'wpcom_vip_log_cron_control_uncacheable_cron_option', 10, 3 );
-
```
jesse-greathouse, line 155: love that. Thank you!

jesse-greathouse (approved):
This is a very sophisticated system and might be hard for me to understand, but I definitely appreciate all the comments and the straightforwardness of the code. Nicely done.

sjinks:
v20211214.0

PR #2449, opened 2021-09-28
Report correct name during settings check

pschoffer:
## Description

The reports we get for search settings inconsistencies would have incorrect index names. That happens if the non-active index has the issue. The reason behind this is that we would reset the current index to active BEFORE we would fetch the index name.

The fix here moves stuff around to make sure we get the name when the test with inconsistencies is still current.

Internal index settings reporting tweak.

## Checklist

## Steps to Test

1) Make sure you have 2 versions of index - `dev-env exec -- wp vip-search index-versions add post` and `dev-env exec -- wp vip-search index --setup --version=2 --indexables=post`
2) Make settings inconsistent. For example - change `filter__ep_default_index_number_of_shards` to return different value.
3) Run: `dev-env exec -- wp eval "var_dump((new Automattic\VIP\Search\Health(\Automattic\VIP\Search\Search::instance()))->get_index_settings_health_for_all_indexables());";`
4) Make sure the index names are correctly reported.


search/includes/classes/class-health.php
```diff
@@ -819,9 +815,18 @@ public function get_index_settings_diff_for_indexable( \ElasticPress\Indexable $
...
+		$result = [];
+		if ( ! empty($diff) ) {
+			$result = array(
+				'index_version' => $options['index_version'] ?? 1,
+				'index_name' => $indexable->get_index_name(),
+				'diff' => $diff,
+			);
+		}
+
 		$this->search->versioning->reset_current_version_number( $indexable );
```
pschoffer, line 827: Now this happens only after we have done `$indexable->get_index_name()` few lines above, so all is good with the world again.

rinatkhaziev (approved):
Amazing!

depeche-mode-enjoy-the-silence

search/includes/classes/class-health.php
jesse-greathouse: well done.

jesse-greathouse (approved):
LGTM

PR #2459, opened 2021-10-01
Hooking Parsely plugin tweaks to init

pauarge:
## Description

The code after the last conditional that removes some hooks doesn't work. That happens because `$parsely` is null. This is due to the fact that it has not yet been initialized (initialization of the wp-parsely plugin happens on the `plugins_loaded` hook, whereas this code happens on `muplugins_loaded`).

This PR moves this second part of the logic to the `init` hook, where wp-parsely is already initialized.

### Fix Parse.ly initialization

Some minor checks on the Parse.ly initialization were not being called.

## Checklist

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -88,6 +88,10 @@ function maybe_load_plugin() {
 add_action( 'muplugins_loaded', __NAMESPACE__ . '\maybe_load_plugin' );
 
 function maybe_disable_some_features() {
+	if ( ! apply_filters( 'wpvip_parsely_load_mu', get_option( '_wpvip_parsely_mu' ) === '1' ) ) {
+		return;
+	}
+
 	global $parsely;
 
 	if ( null != $parsely ) {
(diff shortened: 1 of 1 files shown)


pschoffer (approved):
LGTM

sjinks (approved):

wp-parsely.php
```diff
@@ -84,6 +82,19 @@ function maybe_load_plugin() {
 
 		require $entry_file;
 
+		return;
+	}
+}
+add_action( 'muplugins_loaded', __NAMESPACE__ . '\maybe_load_plugin' );
+
+function maybe_disable_some_features() {
+	if ( ! apply_filters( 'wpvip_parsely_load_mu', get_option( '_wpvip_parsely_mu' ) === '1' ) ) {
```
GaryJones, line 91: Seems like this line should be extracted into a function that returns a boolean, since its the same condition as in the function above.

PR #2467, opened 2021-10-04
Don't use global declaration on Parse.ly

pauarge:
## Description

Instead of using potentially problematic `global $parsely`, we're moving to the use of safer `$GLOBALS['parsely']`.

### Don't use global declaration on Parse.ly

We are implementing a safer way of loading the Parse.ly plugin.

## Checklist

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -91,16 +91,14 @@ function maybe_load_plugin() {
 add_action( 'muplugins_loaded', __NAMESPACE__ . '\maybe_load_plugin' );
 
 function maybe_disable_some_features() {
-	global $parsely;
-
-	if ( null != $parsely ) {
+	if ( $GLOBALS['parsely'] && is_a( $GLOBALS['parsely'], 'Parsely' ) ) {
 		// If the plugin was loaded solely by the option, hide the UI (for now)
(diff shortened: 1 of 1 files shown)


jesse-greathouse (approved):
Cool. I get it.

wp-parsely.php
```diff
@@ -91,16 +91,14 @@ function maybe_load_plugin() {
 add_action( 'muplugins_loaded', __NAMESPACE__ . '\maybe_load_plugin' );
 
 function maybe_disable_some_features() {
-	global $parsely;
-
-	if ( null != $parsely ) {
+	if ( $GLOBALS['parsely'] && is_a( $GLOBALS['parsely'], 'Parsely' ) ) {
```
rinatkhaziev, line 94: This may throw a notice, it would be safer to use an isset check
```suggestion
	if ( isset( $GLOBALS['parsely'] ) && is_a( $GLOBALS['parsely'], 'Parsely' ) ) {
```

rinatkhaziev (changes requested):
Please see the inline suggestion

wp-parsely.php
pauarge: I originally had this and I applied some IDE suggestion. Nice catch!

PR #2469, opened 2021-10-05
Fix code style issues and code smells in `cache`

sjinks:
## Description

This PR fixes code style compliance issues found by `phpcs` and some of the code smells found by `sonarlint`.

### Fix code style compliance issues in `cache` directory
  * Fix issues found by phpcs;
  * Add phpcs:ignore where appropriate to suppress false positives or known good code;
  * Fix code smells detected by `sonarlint`.

## Checklist

## Steps to Test

Make sure that the CI passes.

### DIFF
--- a/cache/class-vary-cache.php
+++ b/cache/class-vary-cache.php
@@ -4,16 +4,18 @@ namespace Automattic\VIP\Cache;
 
 use WP_Error;
 
+// phpcs:disable WordPress.PHP.DevelopmentFunctions.error_log_trigger_error
+
 class Vary_Cache {
 	const COOKIE_NOCACHE = 'vip-go-cb';
 	const COOKIE_SEGMENT = 'vip-go-seg';
-	const COOKIE_AUTH = 'vip-go-auth';
-	const HEADER_AUTH = 'HTTP_X_VIP_GO_AUTH';
+	const COOKIE_AUTH    = 'vip-go-auth';
+	const HEADER_AUTH    = 'HTTP_X_VIP_GO_AUTH';
 
 	// Allowed values in cookie are alphanumerics (A-Za-z0-9) and underscore (_) and hyphen (-).
 	const GROUP_SEPARATOR = '---__';
 	const VALUE_SEPARATOR = '_--_';
-	const VERSION_PREFIX = 'vc-v1__';
+	const VERSION_PREFIX  = 'vc-v1__';
 
 	/**
 	 * Flag to indicate if this an encrypted group request
@@ -107,7 +109,7 @@ class Vary_Cache {
 			return new WP_Error( 'did_send_headers', 'Failed to set nocache cookie; cannot be called after the `send_headers` hook has fired.' );
 		}
 
-		self::$is_user_in_nocache = true;
+		self::$is_user_in_nocache           = true;
 		self::$should_update_nocache_cookie = true;
 
 		return true;
@@ -128,7 +130,7 @@ class Vary_Cache {
 			return new WP_Error( 'did_send_headers', 'Failed to remove nocache cookie; cannot be called after the `send_headers` hook has fired.' );
 		}
 
-		self::$is_user_in_nocache = false;
+		self::$is_user_in_nocache           = false;
 		self::$should_update_nocache_cookie = true;
 
 		return true;
@@ -156,12 +158,12 @@ class Vary_Cache {
 
 		self::clear_groups();
 
-		self::$did_send_headers = false;
-		self::$encryption_enabled = false;
-		self::$is_user_in_nocache = false;
+		self::$did_send_headers             = false;
+		self::$encryption_enabled           = false;
+		self::$is_user_in_nocache           = false;
 		self::$should_update_nocache_cookie = false;
-		self::$should_update_group_cookie = false;
-		self::$cookie_expiry = MONTH_IN_SECONDS;
+		self::$should_update_group_cookie   = false;
+		self::$cookie_expiry                = MONTH_IN_SECONDS;
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


cache/class-vary-cache.php
```diff
@@ -414,24 +416,28 @@ private static function parse_nocache_cookie() {
 	 */
 	private static function parse_group_cookie() {
 		// If the cache layer supplies a decrypted segmentation header, use that instead of decrypting it again.
-		if ( self::is_encryption_enabled() && isset( $_SERVER[ self::HEADER_AUTH ] ) &&  ! empty( $_SERVER[ self::HEADER_AUTH ] ) ) {
+		if ( self::is_encryption_enabled() && ! empty( $_SERVER[ self::HEADER_AUTH ] ) ) {
```
sjinks, line 419: `isset()` check is superfluous because `! empty()` already includes it.

cache/class-vary-cache.php
```diff
@@ -414,24 +416,28 @@ private static function parse_nocache_cookie() {
...
+			// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
 			$cookie_value = $_SERVER[ self::HEADER_AUTH ];
-		} elseif ( self::is_encryption_enabled() && ! empty( $_COOKIE[ self::COOKIE_AUTH ] ) ) {
+		} elseif ( self::is_encryption_enabled() && ! empty( $_COOKIE[ self::COOKIE_AUTH ] ) && isset( $_SERVER['HTTP_COOKIE'] ) ) {
```
sjinks, line 422: We didn't check if `$_SERVER['HTTP_COOKIE']` is set; the lack of the check can cause warnings in WP CLI.

cache/class-vary-cache.php
```diff
@@ -414,24 +416,28 @@ private static function parse_nocache_cookie() {
...
-				if( self::COOKIE_AUTH === $k ) {
+				if ( self::COOKIE_AUTH === $k ) {
 					$auth_cookie = $v;
 					break;
 				}
 			}
 
-			$value = ltrim( $auth_cookie, VIP_GO_APP_ID . '.' ); // remove the site prefix
+			// Remove the site prefix
+			$value        = ltrim( $auth_cookie, VIP_GO_APP_ID . '.' );
```
sjinks, line 436: Had to move the comment to the top because of the weird complaint:

> Using the WPCS native whitelist comments is deprecated. Please use the PHPCS native "phpcs:ignore Standard.Category.SniffName.ErrorCode" annotations instead. Found: // Remove the site prefix\nPHPCS(WordPress.NamingConventions.PrefixAllGlobals.DeprecatedWhitelistCommentFound)

sjinks: As a side note: is this code correct? `ltrim()` will not remove the prefix; it will remove all leading characters specified by its second argument. Thus, if `VIP_GO_APP_ID` is `1234`, the code will remove `1234321.` from the prefix, which is *probably* not the intended behavior.

cache/class-vary-cache.php
```diff
@@ -457,7 +463,7 @@ private static function parse_group_cookie() {
 	 */
 	private static function stringify_groups() {
 		if ( empty( self::$groups ) ) {
-			return;
+			return null;
```
sjinks, line 466: Here and below: `Refactor this function to use "return" consistently`. Returning `null` is in line with what the doc block says: `@return null|string`

cache/class-vary-cache.php
```diff
@@ -486,7 +492,7 @@ private static function stringify_groups() {
 	 */
 	public static function set_cookie_expiry( int $expiry ) {
 		if ( $expiry < HOUR_IN_SECONDS ) {
-			trigger_error( sprintf( '%s: cookie expiry must be greater than or equal to 1 hour (%d)', __METHOD__, HOUR_IN_SECONDS ), E_USER_WARNING );
+			trigger_error( sprintf( '%s: cookie expiry must be greater than or equal to 1 hour (%d)', __METHOD__, intval( HOUR_IN_SECONDS ) ), E_USER_WARNING );
```
sjinks, line 495: Unfortunately, phpcs is not smart enough to figure out that `HOUR_IN_SECONDS` is integer. I am explicitly using `intval()` to avoid the suppression comment.

cache/class-vary-cache.php
rebeccahum: > Using the WPCS native whitelist comments is deprecated. Please use the PHPCS native "phpcs:ignore Standard.Category.SniffName.ErrorCode" annotations instead. Found: // Remove the site prefix\n PHPCS(WordPress.NamingConventions.PrefixAllGlobals.DeprecatedWhitelistCommentFound)

That weird complaint seems to be a false positive. However, that rule seems to be no longer there as of https://github.com/WordPress/WordPress-Coding-Standards/pull/1908/files, so in future versions of WPCS (3.0+), it won't be flagged.

rebeccahum: Re: the `ltrim()`, @dchymko can you speak to this? I noticed it was added in https://github.com/Automattic/vip-go-mu-plugins/pull/1285/files.

cache/class-vary-cache.php
rebeccahum: I would rather the suppression comment since it might be confusing — at first glance, one may wonder why `intval()` is being used on `HOUR_IN_SECONDS`.

sjinks: Fixed in 29f2d17d2

rebeccahum (approved):
Thanks!

cache/class-vary-cache.php
GaryJones: Might be better to see where this private function is called, and to return an empty string so at least the return type is consistent.

sjinks: Will do that in another PR, thanks

PR #2539, opened 2021-10-22
Add global wrapper function for Automattic\VIP\Environment::get_var

jesse-greathouse:
## Description

  * Added vip_get_env_var function so it can be available in the runtime.
  * Added unit tests to validate the functionality

### vip_get_env_var function

Added function: vip_get_env_var as a shortcut to requiring Automattic\VIP\Environment::get_var

```php
$my_env_var = vip_get_env_var( 'FOO',  'BAR' );
```
If the key, for the custom environment variable ("FOO") does not exist, the default value ( "BAR" ) will be returned.

## Checklist

## Steps to Test

1. Check out PR.
1. Create a user defined environment variable in VIP Cli, or just add a constant definition to wp-config.php
1. put the following code in the index.php file of your active template:
  ```php
    <?php
    // testing vip_get_env_var ( Automattic\VIP\Environment::get_var wrapper )
    $myVar = vip_get_env_var('MY_VAR');
    ?>

    <script>
      console.log('$myVar: <?php echo $myVar; ?> ');
    </script>
  ```
  On a page load, the javascript console should display the value


jesse-greathouse:
@rinatkhaziev I've added the helper function to lib/helpers/environment.php

The tests will probably not work until we merge in your new pre-mu-plugins file with a require statement on that new file.

lib/helpers/environment.php
```diff
@@ -0,0 +1,13 @@
+<?php
+/**
+ * Helper functions pretaining to the environment domain
```
rinatkhaziev, line 3: Typo here: 
```suggestion
 * Helper functions pertaining to the environment domain
```

lib/helpers/environment.php
```diff
@@ -0,0 +1,13 @@
...
+*/
+
```
rinatkhaziev, line 5: Let's add a `require_once` for the class here just for extra safety.

lib/helpers/environment.php
```diff
@@ -0,0 +1,13 @@
...
+use Automattic\VIP\Environment;
+
+/**
+ * A wrapper for Automattic\VIP\Environment::get_var
```
rinatkhaziev, line 9: PHPDoc is missing the signature for arguments, let's add that.

rinatkhaziev (changes requested):
Thanks! 

I left some notes inline. 

On top of that, I think the tests should go into tests/lib/class-environment.php rather than in tests/core.php

Additionally, we're expecting the notice - because it's the correct behavior, so instead of suppressing we should do something similar to what we do in https://github.com/Automattic/vip-go-mu-plugins/blob/master/tests/lib/test-vip-request-block.php

Also tests are failing with `Error: Call to undefined function vip_get_env_var()`

rinatkhaziev:
Please disregard my last note about the tests - I didn't see your previous message, it's ok to require the file in the test file, there are plenty of tests like that.

tests/test-core.php
```diff
@@ -1,10 +1,85 @@
...
+	 * tests the use-case where $key parameter does not have the prefix
+	 * @runInSeparateProcess
+	 * @preserveGlobalState disabled
+	 */
+	public function test_get_var_legacy_key() {
+		error_reporting( $this->error_reporting & ~E_USER_NOTICE );
+
+		$this->get_var_legacy_env();
+		$val = vip_get_env_var( 'MY_VAR', 'default_value' );
+		$this->assertEquals( 'MY_VAR', $val );
```
chriszarate, line 44: I would suggest making the value of the env var distinct from the name in test, for two reasons:

1. This is a bit confusing to read, initially I thought you were asserting that the value was equal to the name.
2. If by some future misfortune, the code accidentally returned the name instead of the value, this test would still pass.

lib/helpers/environment.php
jesse-greathouse: Thank you for spotting this typo. I have fixed it in the latest commit.

lib/helpers/environment.php
jesse-greathouse: Thank you. I have made this change in the latest commit.

lib/helpers/environment.php
jesse-greathouse: Good catch. I have made this update in the latest commit.

tests/test-core.php
jesse-greathouse: Thank you for the great suggestion. I have made this change in my latest commit.

sjinks (approved):

rinatkhaziev:
@jesse-greathouse thanks for making the changes! 

The final thing to address here is this: 
 
> We're expecting the notice - because it's the correct behavior, instead of suppressing we should do something similar to what we do in https://github.com/Automattic/vip-go-mu-plugins/blob/master/tests/lib/test-vip-request-block.php

More at https://phpunit.readthedocs.io/en/stable/writing-tests-for-phpunit.html#testing-php-errors-warnings-and-notices

jesse-greathouse:
> @jesse-greathouse thanks for making the changes!
> 
> The final thing to address here is this:
> 
> > We're expecting the notice - because it's the correct behavior, instead of suppressing we should do something similar to what we do in https://github.com/Automattic/vip-go-mu-plugins/blob/master/tests/lib/test-vip-request-block.php
> 
> More at https://phpunit.readthedocs.io/en/stable/writing-tests-for-phpunit.html#testing-php-errors-warnings-and-notices

Thank you for pointing this out. I have replaced the old code to take advantage of the methods in the ExpectPHPException class. Now the two test cases that will generate notices, have the expectNotice method. The rest of the tests did not require this and just had the line removed.

rinatkhaziev (approved):
Thank you for addressing the feedback!

jesse-greathouse:
r1984

PR #2544, opened 2021-10-25
Cache WP and WP Test Library

sjinks:
## Description

Initially, this PR was meant to bring up WordPress and WordPress Test Library caching, so that we don't get 429 errors from WP SVN.

However, it turned out that CircleCI's caching does not support our scenario (I failed to make it accept environment variables and/or pipeline parameters both in a cache name and a path to cache). The only scenario the caching would work was to cache all necessary WP/WPTL versions in a predefined location. However, that does not work well with our pipeline.

As a result, I ended up creating my own Docker image with all WP/WPTL baked in. As a bonus, the image provides several PHPUnit (7, 8, and 9) and PHP (7.3, 7.4, 8.0, and 8.1). WP, PHPUnit, and PHP versions are controlled with the environment variables. The image also contains Node.js 14, npm, MySQL client, Subversion, and a couple of other tools; this allows us to save some time because we don't need to install the system-wide dependencies anymore.

I have refactored the workflow to use this image; as a result, the complexity of the workflow has reduced a bit.

### CI Pipeline

Refactoring.

## Checklist

## Steps to Test

The CI should pass.


.circleci/config.yml
```diff
@@ -41,7 +41,7 @@ parameters:
     default: /tmp/wordpress-tests-lib
   wp_core_dir:
     type: string
-    default: /tmp/wordpress/
+    default: /tmp/wordpress
```
pauarge, line 44: Why are we changing this?

sjinks: Actually, for no particular reason… The scripts expect the unslashed path, and with the trailing slash it turns into smth like `/tmp/wordpres//something`. It works but does not look nice.

pauarge: If that's the case, it will indeed look better.

pauarge (approved):

PR #2562, opened 2021-10-29
Returning result of tracking functions

pauarge:
## Description

This fixes an issue in which some filters that require a return value would not get that value back, even if the called function was returning it.

### Returning result of tracking functions

Fix an issue when tracking would not work for filters that require a return value.

## Checklist


sjinks:
v20211102.0

PR #2563, opened 2021-10-29
Fix 2FA error during Lando: wp install

jesse-greathouse:
## Description

The dev-env start routine will do a WordPress install if the dev-env containers have not yet been created. When this happens, the Lando init routine will run the WP CLI to install the Wordpress Instance. This causes a benign but annoying PHP error notice because: in the 2FA plugin, `wp_parse_url` is called with a url parameter value of: "". When the plugin tries to use the parts['host'] from that parse, it doesn't exist in the dictionary and causes that annoying notice.

This PR addresses that notice with the following change:

  * Adds Automattic\VIP\Environment::is_dev_env_container method
  * Prevents loading the 2FA plugin in the context of WP CLI and vip dev-env
  * Unit test for Environment::is_dev_env_container

## Changelog Description

The command: `vip dev-env start` will no longer show a PHP Notice error when the WordPress environment container is first instantiated.

## Checklist

## Steps to Test

1. Check out PR.
1. Create a new dev-env container with the command: `vip dev-env create --slug=test-2fa-bug`
1. Run the container with `vip dev-env start --slug=test-2fa-bug`
1. Verify that there is no longer a PHP Notice error about the 2FA plugin.

### DIFF
--- a/lib/environment/class-environment.php
+++ b/lib/environment/class-environment.php
@@ -68,4 +68,14 @@ class Environment {
 
 		return false;
 	}
+
+	public static function is_dev_env_container( ) {
+		if ( isset( $_ENV['LANDO_APP_ROOT'] ) ) {
+			if ( preg_match( '/^[\w\-. \/\-_]+vip\/dev-environment\/[\w\-. \-_]+$/',  $_ENV['LANDO_APP_ROOT'] ) ) {
+				return true;
+			}
+		}
+
+		return false;
+	}
 }
--- a/lib/environment/class-environment.php
+++ b/lib/environment/class-environment.php
@@ -69,9 +69,9 @@ class Environment {
 		return false;
 	}
 
-	public static function is_dev_env_container( ) {
+	public static function is_dev_env_container() {
 		if ( isset( $_ENV['LANDO_APP_ROOT'] ) ) {
-			if ( preg_match( '/^[\w\-. \/\-_]+vip\/dev-environment\/[\w\-. \-_]+$/',  $_ENV['LANDO_APP_ROOT'] ) ) {
+			if ( preg_match( '/^[\w\-. \/\-_]+vip\/dev-environment\/[\w\-. \-_]+$/', $_ENV['LANDO_APP_ROOT'] ) ) {
 				return true;
 			}
 		}
--- a/tests/lib/environment/test-class-environment.php
+++ b/tests/lib/environment/test-class-environment.php
@@ -221,8 +221,8 @@ class Environment_Test extends TestCase {
 		return array(
 			array(
 				// path -- directory where a dev-env lando root is
-				'/Users/username/.local/share/vip/dev-environment/username'
-			)
+				'/Users/username/.local/share/vip/dev-environment/username',
+			),
 		);
 	}
 
@@ -233,6 +233,6 @@ class Environment_Test extends TestCase {
 	 */
 	public function test_is_dev_env_container( $path ) {
 		$_ENV['LANDO_APP_ROOT'] = $path;
-		$this->assertTrue( Environment::is_dev_env_container( ) );
+		$this->assertTrue( Environment::is_dev_env_container() );
 	}
 }
--- a/tests/lib/environment/test-class-environment.php
+++ b/tests/lib/environment/test-class-environment.php
@@ -216,4 +216,23 @@ class Environment_Test extends TestCase {
 
 		$this->assertEquals( $expected, $result );
 	}
+
+	public function is_dev_env_container_data() {
+		return array(
+			array(
+				// path -- directory where a dev-env lando root is
+				'/Users/username/.local/share/vip/dev-environment/username'
+			)
+		);
+	}
+
+	/**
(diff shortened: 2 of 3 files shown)


rinatkhaziev:
So I wonder whether the 2FA plugin is even helpful in the CLI context at all.

lib/environment/class-environment.php
```diff
@@ -68,4 +68,14 @@ public static function is_batch_container( $hostname, $env = array() ) {
 
 		return false;
 	}
+
+	public static function is_dev_env_container() {
+		if ( isset( $_ENV['LANDO_APP_ROOT'] ) ) {
+			if ( preg_match( '/^[\w\-. \/\-_]+vip\/dev-environment\/[\w\-. \-_]+$/', $_ENV['LANDO_APP_ROOT'] ) ) {
```
rinatkhaziev, line 74: I don't really think the regex is needed here - if the `$_ENV['LANDO_APP_ROOT']` we must be in Lando, right? :)

tests/lib/environment/test-class-environment.php
```diff
@@ -216,4 +216,23 @@ public function test_is_batch_container( $hostname, $env, $expected ) {
 
 		$this->assertEquals( $expected, $result );
 	}
+
+	public function is_dev_env_container_data() {
+		return array(
+			array(
+				// path -- directory where a dev-env lando root is
+				'/Users/username/.local/share/vip/dev-environment/username',
```
rinatkhaziev, line 224: Needs a Windows test case

two-factor.php
```diff
@@ -155,7 +158,10 @@ function() use ( $is_user_using_two_factor ) {
 	}
 }
 
-add_action( 'muplugins_loaded', 'wpcom_enable_two_factor_plugin' );
+// Do not enable 2FA plugin if WordPress is executed in the context of the CLI
```
rinatkhaziev, line 161: I think instead of wrapping it here maybe just move that inside the `wpcom_enable_two_factor_plugin`

rinatkhaziev (changes requested):
Thanks, left some feedback inline.

The biggest thing is that now I'm not sure whether we need to load 2FA in CLI at all.

tests/lib/environment/test-class-environment.php
jesse-greathouse: Good call. I will add that.

two-factor.php
jesse-greathouse: Ok I will try that and see how that works.

lib/environment/class-environment.php
jesse-greathouse: > if the `$_ENV['LANDO_APP_ROOT']` we must be in Lando, right? :)

This was my thought at first, but then I realized that I don't really know if mu-plugins is being used in Lando based projects which are not our dev-env. 

Perhaps a user is loading mu-plugins in their own iteration of a Lando env, and maybe this change would not be welcome in such a scenario. 

Trying to use a scalplel and not a hatchet, but maybe it's overkill? Im not in love with the computation efficiency of it. What do you think about that?

pschoffer (changes requested):
I think we might get away with not changing the 2fa plugin at all.

Tracking the stacktrace it is called when we are setting the 'fido' provider.

The relevant iteration over providers is here:
https://github.com/WordPress/two-factor/blob/cf55e92bef8f44ffeed2d4f144dc177a4cb742ca/class-two-factor-core.php#L150

Few lines, above there is a filter we could implement to remove fido provider from the providers. So in extreme case we could do:

```
add_filter('two_factor_providers', function() { return array(); });
```
This code then could live in here (so outside of production mu-plugins): 
https://github.com/Automattic/vip-container-images/blob/master/dev-tools/dev-env-plugin.php

rinatkhaziev:
Another alternative way to "fix" the issue is suppressing the notices via WP-CLI argument for the install

jesse-greathouse:
> I think we might get away with not changing the 2fa plugin at all.
> 
> Tracking the stacktrace it is called when we are setting the 'fido' provider.
> 
> The relevant iteration over providers is here: https://github.com/WordPress/two-factor/blob/cf55e92bef8f44ffeed2d4f144dc177a4cb742ca/class-two-factor-core.php#L150
> 
> Few lines, above there is a filter we could implement to remove fido provider from the providers. So in extreme case we could do:
> 
> ```
> add_filter('two_factor_providers', function() { return array(); });
> ```
> 
> This code then could live in here (so outside of production mu-plugins): https://github.com/Automattic/vip-container-images/blob/master/dev-tools/dev-env-plugin.php

I put this solution into the Automattic/vip-container-images repo, under this [PR here](https://github.com/Automattic/vip-container-images/pull/38/files).

It fixes the problem and I also love that it makes a very specific change to the list of providers, while retaining all the rest of the providers and functionality of the plugin.

This is a great suggestion by @pschoffer because it uses a scalpel to make a very surgical change to fix the problem. Thank you for this great suggestion, Pavel.

I'm still looking into why Lando's `wp install` is not handing off the desired hostname as a parameter, but I'm closing this PR in the interim, because I don't foresee it will be needed.

PR #2564, opened 2021-10-29
Create a new folder 000-pre-vip-config and move all the requires that happen before vip-config.php there for better visibility

rinatkhaziev:
## Description

We have a bunch of "secret sauce" that was loaded prior to vip-config.php but wasn't exposed anywhere, there's no particularly good reason to do so. Furthermore, it's a bit annoying because it's opaque and in case customers don't use our Local Dev Env they'll no doubt will be tripped up by that. 

This PR aims to address the issue by introducing a new folder, which is very explicitly named `000-pre-vip-config`. This needs to be accompanied with:

- [ ] Docs updates
- [ ] Dev Env bootstrap update
- [ ] Removal of the secret sauce from the secret place(s).

### WordPress dependency load order update

Previously we included a few classes, notably VIP_Request_Block and our custom error handler very early in the bootstrap, prior to loading the `vip-config.php`. Although those files are located in mu-plugins we didn't load them from there. 

We've made a change to include `000-pre-vip-config/requires.php` that will now contain those early dependencies so that it's more apparent.  

## Checklist

### DIFF
--- a/000-pre-vip-config/requires.php
+++ b/000-pre-vip-config/requires.php
@@ -1,0 +1,25 @@
+<?php
+/**
+ * This file contains everything that's loaded before vip-config.php
+ *
+ * You should make sure to include this file locally in case you're not using VIP Local Development Environment
+ * @see https://docs.wpvip.com/how-tos/local-development/use-the-vip-local-development-environment/
+ */
+
+$mu_plugins_base = dirname( __DIR__ );
+
+// Load custom error logging functions, if available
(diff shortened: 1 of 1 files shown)


jesse-greathouse (approved):
Great! thank you so much. lets merge this and I'll get the helper function in there.

chriszarate (approved):

rinatkhaziev:
Noting here, this is just the file, at this moment it's not included anywhere (but will be in the near future)

rinatkhaziev:
r1965-stacks

PR #2565, opened 2021-10-30
Don't log the default value of the VIP Env var if it wasn't set

rinatkhaziev:
## Description

We shouldn't leak the default value into logs, merely the fact of an expected variable absent should provide enough information for debug.

### VIP Env Variables

Don't log the default value when the specified key is not set.

## Checklist

## Steps to Test

1. Check out the PR 
2. `wp shell`
3. `Environment::get_var( 'notexisting', 'default_value' );`
4. Verify `default_value` is not being logged.


WPprodigy (approved):
✅ . I'm not completely sure I see the need to log at all, it's similar to a `get_option( 'key', 'default value' );`, in that having a default value is a feature not necessarily a bug if it's used.

I feel like if not having a value is an error to the codebase, then the application should be checking against a default null-like value on it's own. Otherwise, defaulting to something seems perfectly reasonable in many circumstances where you don't need an environmental override.

pauarge (approved):
Good catch.

rinatkhaziev:
@WPprodigy I see your point, the idea was mostly around catching potential regressions since the vars are getting propagated via other systems, it's not inconceivable that something may break there. We can always remove if it proves to be useless or too noisy.

rinatkhaziev:
r1966-stacks

PR #2573, opened 2021-11-04
Search: Do not perform current field count if no index exists

rebeccahum:
## Description
There seems to be a lot of index_not_found_exception errors in our logs stemming from the cron `maybe_alert_for_field_count` which calls `get_current_field_count()` — it attempts to do a GET request on an index that doesn't exist. We should return early if index does not exist.

### Search: Return early if no index on get_current_field_count()

No need to perform a remote request since it will throw an index_not_found_exception.

## Checklist

## Steps to Test
1) Check out PR
2) Spin up search site
3) If index exists, delete it `wp vip-search delete-index`
4) Use `Automattic\VIP\Search\Search::instance()->maybe_alert_for_field_count();` to see if `$current_field_count` returns NULL.

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -1165,7 +1165,7 @@ class Search {
 	 * @return null|int The current field count
 	 */
 	public function get_current_field_count( \ElasticPress\Indexable $indexable ) {
-		if ( ! $indexable ) {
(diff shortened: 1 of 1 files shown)


WPprodigy (approved):

rebeccahum:
r1972-stacks

PR #2581, opened 2021-11-04
jetpack-start: Update to use modern function

kraftbj:
This function has been deprecated since 9.5 and removed in 10.3. See https://github.com/Automattic/jetpack/pull/16434

## Changelog Description
Update `jetpack-start` to use modern function.

## Checklist

## Steps to Test
Unsure personally. Reported in vip-jetpack via p1636058453047900-slack-CDD9LQRSN

### DIFF
--- a/jetpack-start/wp-cli-provision.php
+++ b/jetpack-start/wp-cli-provision.php
@@ -1,5 +1,7 @@
 <?php
 
+use Automattic\Jetpack\Connection\Tokens;
+
 /**
  * These are extracted commands from Jetpack CLI.
  *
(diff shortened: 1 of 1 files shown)


jetpack-start/wp-cli-provision.php
```diff
@@ -230,7 +232,7 @@ public function partner_provision( $args, $named_args ) {
 
 		if ( isset( $body_json->access_token ) ) {
 			// authorize user and enable SSO
-			Jetpack::update_user_token( $user->ID, sprintf( '%s.%d', $body_json->access_token, $user->ID ), true );
+			( new Tokens )->update_user_token( $user->ID, sprintf( '%s.%d', $body_json->access_token, $user->ID ), true );
```
kraftbj, line 235:
```suggestion
			if ( method_exists( 'Jetpack', 'update_user_token' ) ) {
				Jetpack::update_user_token( $user->ID, sprintf( '%s.%d', $body_json->access_token, $user->ID ), true );
			} else {
				( new Tokens() )->update_user_token( $user->ID, sprintf( '%s.%d', $body_json->access_token, $user->ID ), true );
			}

```

WPprodigy (approved):
Confirmed broken in latest, but works w/ this patch.

Also noting that the connection pilot does not use this functionality, only the jetpack-start CLI which is very rarely used and only manually now. Ideally this whole module is deprecated and removed soon, as the connection pilot uses the much newer connection process.

PR #2583, opened 2021-11-05
Add Environment helpers requirement in 000-pre-vip-config

rinatkhaziev:
## Description

This PR adds a require to lib/helpers/environments, thus exposing `vip_get_env_var` early. 

### Load vip_get_env_var early

We've implemented early loading of `vip_get_env_var` so that it's available in vip-config.php

## Checklist

## Steps to Test

1. Checkout the PR
2. Add `require_once __DIR__ .  '/wp-content/mu-plugins/000-pre-vip-config/requires.php';` somewhere on your local install prior to loading `vip-config.php`
3. Verify that the helper function works as expected.


000-pre-vip-config/requires.php
```diff
@@ -21,5 +21,8 @@
 // Load Environment utility class and its helpers, if available
 if ( file_exists( $mu_plugins_base . '/lib/environment/class-environment.php' ) ) {
 	require_once $mu_plugins_base . '/lib/environment/class-environment.php';
```
WPprodigy, line 23: Could drop these then since it's required here: https://github.com/Automattic/vip-go-mu-plugins/blob/master/lib/helpers/environment.php#L6 ?

rinatkhaziev: I mean, we can drop it all later, just want to get it deployed first.

rinatkhaziev: Oh, I misread your comment a bit, I. think we just want to minimize the chances of breakage but also be explicit as much as possible.

WPprodigy (approved):

PR #2585, opened 2021-11-05
Tighten alloptions alerts

trepmal:
## Description
- Increases warning threshold. Review of current alerts show that current levels are generous
- Removes Slack alerts. These are redundant.
- Removes acking/silencing feature which only relevant to Slack alerts.
- Error-level thresholds get Opsgenie P2 (high), mid-level is P3 (moderate), and low-level is P5 (informational).
- Removes _disable by ID_, which is inefficient in this repository.
- Replaces unused email notification with flexible action.

### Plugin Updated: VIP AllOptions Safeguard

Tightens rules for VIP alerts regarding the size of autoloaded options.

## Checklist

## Steps to Test

Create some large options. I like to use a few wikipedia pages since they will not compress as well as other values. Example:

```
update_option( 'large_option_1', wp_remote_get( 'https://en.wikipedia.org/wiki/Language' ) );
```

Be ready to flush or bypass caches: `alloptions_size`, `alloptions_size_compressed` and `alloptions` (`throttle` group)

For local testing, you can add QM debugging above the `// Send notices to VIP staff` comment/conditionals. 

```
do_action( 'qm/debug', 'notifying... ' . ($really_blocked ? 'really blocked' : ' not blocked' ) );
```


alloptions-limit.php
```diff
@@ -39,38 +41,42 @@ function wpcom_vip_sanity_check_alloptions() {
...
-		wpcom_vip_sanity_check_alloptions_die( $alloptions );
+	if ( $maybe_blocked ) {
+		// It's likely at this point the site is already experiencing performance degradation.
+		// We're using gzdeflate here because pecl-memcache uses Zlib compression for large values.
+		// See https://github.com/websupport-sk/pecl-memcache/blob/e014963c1360d764e3678e91fb73d03fc64458f7/src/memcache_pool.c#L303-L354
+		$alloptions_size_compressed = wp_cache_get( 'alloptions_size_compressed' );
+		if ( ! $alloptions_size_compressed ) {
+			// phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.serialize_serialize
+			$alloptions_size_compressed = strlen( gzdeflate( serialize( wp_load_alloptions() ) ) );
+			wp_cache_add( 'alloptions_size_compressed', $alloptions_size_compressed, '', 60 );
```
WPprodigy, line 62: Could use `maybe_serialize()` here. While `wp_load_alloptions()` _should_ give an array, it is filterable and doesn't type check before the return.

Similarly,  `gzdeflate` will return false if it fails, so may need to handle that.

alloptions-limit.php
```diff
@@ -106,43 +112,30 @@ function wpcom_vip_sanity_check_alloptions_notify( $size, $blocked = false ) {
...
-
-		// If site is blocked, then the IRC alert is CRITICAL
-		$irc_alert_level      = 3;
-		$opsgenie_alert_level = 'P3';
+	} elseif ( $maybe_blocked ) {
+		$msg  = 'This will soon be BLOCKED from loading until if the options sizes increase.';
+		$msg .= PHP_EOL . PHP_EOL;
+		$msg .= sprintf( 'Blocking threshold: %s. Current size: %s', VIP_ALLOPTIONS_ERROR_THRESHOLD, $size_compressed );
 	} else {
 		$msg = 'Site will be blocked from loading if option sizes get too much bigger.';
```
WPprodigy, line 122: I like the extra context that $maybe_blocked is getting in the message, probably useful to give some here as well?

alloptions-limit.php
```diff
@@ -152,41 +145,37 @@ function wpcom_vip_sanity_check_alloptions_notify( $size, $blocked = false ) {
...
+		}
 
-		if ( $email_recipient ) {
-			$size = size_format( $size );
+		/**
+		 * Fires under alloptions warning conditions
+		 *
+		 * @param bool $really_blocked False if alloptions size is large. True if site loading is being blocked.
+		 */
+		do_action( 'vip_alloptions_notification', $really_blocked );
```
WPprodigy, line 178: Let's move this hook up higher before "Send notices to VIP staff". I don't think it needs to be under any conditionals aside from being under the rate-limiting ones.

Then can short circuit all the "Send notices to VIP staff" section conditionals and early abort if opsgenie won't be needed.

alloptions-limit.php
```diff
@@ -152,41 +145,37 @@ function wpcom_vip_sanity_check_alloptions_notify( $size, $blocked = false ) {
...
+					'description' => sprintf( 'The size of AllOptions has breached %s bytes', VIP_ALLOPTIONS_ERROR_THRESHOLD ),
+					'entity'      => (string) $site_id,
+					'priority'    => 'P3',
+					'source'      => 'sites/alloptions-size',
+				),
+				'alloptions-size-alert',
+				'10'
+			);
 
-		$email_recipient = defined( 'VIP_ALLOPTIONS_NOTIFY_EMAIL' ) ? VIP_ALLOPTIONS_NOTIFY_EMAIL : false;
```
WPprodigy, line 183: Let's just make sure nobody is using this constant, but +1 to removing and using the hook instead. If anybody is using it, we just need to inform them and help 'em migrate to the hook.

alloptions-limit.php
WPprodigy: Wait, i don't think `$msg` is used anyway. Perhaps remove the whole block?

trepmal: Ah right, that was removed along with the email function.

alloptions-limit.php
trepmal: at the moment, this is not user-definable

alloptions-limit.php
```diff
@@ -32,45 +34,50 @@ function wpcom_vip_sanity_check_alloptions() {
...
+		// It's likely at this point the site is already experiencing performance degradation.
+		// We're using gzdeflate here because pecl-memcache uses Zlib compression for large values.
+		// See https://github.com/websupport-sk/pecl-memcache/blob/e014963c1360d764e3678e91fb73d03fc64458f7/src/memcache_pool.c#L303-L354
+		$alloptions_size_compressed = wp_cache_get( 'alloptions_size_compressed' );
+		if ( ! $alloptions_size_compressed ) {
+			$alloptions_size_deflated = gzdeflate( maybe_serialize( wp_load_alloptions() ) );
+			if ( $alloptions_size_deflated ) {
+				$alloptions_size_compressed = strlen( $alloptions_size_deflated );
+				wp_cache_add( 'alloptions_size_compressed', $alloptions_size_compressed, '', 60 );
+			}
```
WPprodigy, line 63: In case of a fail during `gzdeflate()` (not sure when that would be), I think maybe better to still cache so it doesn't keep trying, and then just assume something is wrong (but not "time to block site" wrong):

```
$alloptions_size_deflated = gzdeflate( maybe_serialize( wp_load_alloptions() ) );
$alloptions_size_compressed = $alloptions_size_deflated !== false ? strlen( $alloptions_size_deflated ) : VIP_ALLOPTIONS_ERROR_THRESHOLD - 1;
wp_cache_add( 'alloptions_size_compressed', $alloptions_size_compressed, '', 60 );
```

alloptions-limit.php
```diff
@@ -106,87 +122,68 @@ function wpcom_vip_sanity_check_alloptions_notify( $size, $blocked = false ) {
...
-		$msg = 'Site will be blocked from loading if option sizes get too much bigger.';
-	}
-
-	$msg .= "\n\nDebug information can be found in the Fieldguide";
+	/**
+	 * Fires under alloptions warning conditions
+	 *
+	 * @param bool $really_blocked False if alloptions size is large. True if site loading is being blocked.
+	 */
+	do_action( 'vip_alloptions_notification', $really_blocked );
```
WPprodigy, line 130: If I saw `vip_alloptions_notification ` by itself I wouldn't know for sure what it does. Maybe `vip_alloptions_threshold_alert` could be good.

sjinks (approved):

sjinks:
r2016-stacks

rebeccahum:
v20211117.0

PR #2586, opened 2021-11-06
A8c files: Move out the Metadata Updater scheduling from web request path

rinatkhaziev:
## Description

There's no particular need to run this logic on the frontend requests, as it's mostly just wasting cycles.

## Checklist


a8c-files.php
```diff
@@ -86,13 +86,15 @@ public function __construct() {
...
-			// add new cron schedule for filesize update
-			add_filter( 'cron_schedules', array( $this, 'filter_cron_schedules' ), 10, 1 ); // phpcs:ignore WordPress.WP.CronInterval.CronSchedulesInterval
-
-			// Schedule meta update job
-			$this->schedule_update_job();
+		if ( defined( 'WP_CLI' ) && WP_CLI ) {
+			// Conditionally schedule the attachment filesize metadata update job
+			if ( defined( 'VIP_FILESYSTEM_SCHEDULE_FILESIZE_UPDATE' ) && true === VIP_FILESYSTEM_SCHEDULE_FILESIZE_UPDATE ) {
+				// add new cron schedule for filesize update
+				add_filter( 'cron_schedules', array( $this, 'filter_cron_schedules' ), 10, 1 ); // phpcs:ignore WordPress.WP.CronInterval.CronSchedulesInterval
```
WPprodigy, line 93: Hmm, I think we want this to occur even outside of Cron though. Events are "purged" when they are invalid at various places in WP, or during registration need to use a valid schedule, etc. Best to leave it globally set I think

WPprodigy: 0.o

```
// Not actually five minutes; we want it to run faster though to get through everything.
$schedule['vip_five_minutes'] = [ .... ]
```

rinatkhaziev: @WPprodigy the first comment was addressed, and the second one - yeah 0.o :)

mjangda: That's why you never put the interval in the `cron_schedule` name :)

a8c-files.php
```diff
@@ -86,13 +86,15 @@ public function __construct() {
...
 
-		// Conditionally schedule the attachment filesize metaata update job
+		// Conditionally schedule the attachment filesize metadata update job
 		if ( defined( 'VIP_FILESYSTEM_SCHEDULE_FILESIZE_UPDATE' ) && true === VIP_FILESYSTEM_SCHEDULE_FILESIZE_UPDATE ) {
 			// add new cron schedule for filesize update
 			add_filter( 'cron_schedules', array( $this, 'filter_cron_schedules' ), 10, 1 ); // phpcs:ignore WordPress.WP.CronInterval.CronSchedulesInterval
 
-			// Schedule meta update job
-			$this->schedule_update_job();
+			if ( defined( 'WP_CLI' ) && WP_CLI ) {
```
mjangda, line 94: If we ever move away from `WP_CLI`-based cron execution this will break.

Slightly better would be to use `is_admin()` (we could qualify to further to skip admin ajax as well if we want).

mjangda (changes requested):

a8c-files.php
rinatkhaziev: @mjangda if we ever decide to do that we'll have to audit everything regardless, I think, but fair point, I added a `wp_doing_cron` check for future proofing.

mjangda (approved):
🚢

rinatkhaziev:
r2006-stacks

sjinks:
v20211116.0

PR #2597, opened 2021-11-08
Search: Better request handling and caching

rinatkhaziev:
## Description

Currently, we don't really cache anything at all, this puts extra strain on the cluster.

This PR introduces low-level caching - we're caching the whole response object. The significant benefit of that is that any other logic would treat is a regular request. E.g. the query will appear in Search Dev Tools.

Currently, three following endpoints are cached `_doc`, `_mget`, `_search`.

Cache time is set at 5 minutes (for fallback purposes), but a request is deemed stale at 30s mark. This allows for content freshness + graceful fallback.

Introduces filters:

```apply_filters( 'vip_search_cache_es_response', bool $is_cacheable, string $url, array $args )```

```apply_filters( 'vip_search_stale_request_threshold', 30, $args );```

We have introduced a graceful fallback in case of a request to Enterprise Search backend fails. We would serve a previously-stored stale response (up to 5 minutes).

Additionally, only `query` requests are counted against the fail counter that blocks subsequent remote requests in case of a failure (via vip_safe_wp_remote_request).

## Checklist

## Steps to Test

1. Check out the PR
2. Set a breakpoint or  add an exit inside the https://github.com/Automattic/vip-go-mu-plugins/pull/2597/files#diff-dd3f02b793bcc2930a3ba1d70b0520c0865dc4a09d2dbd1f49a1c55af101c40dR853 block
3. Hit a `?s=test` or similar URL in the browser
4. Verify that. on the second run the cached code path is executing

### CODE REVIEW

File: search/includes/classes/class-search.php [hunk 1]
```diff
@@ -771,10 +780,37 @@ public function filter__ep_do_intercept_request( $request, $query, $args ) {
 			// Return a generic VIP Search WP_Error instead of the one from wp_remote_request
 			return new \WP_Error( 'vip-search-upstream-request-failed', 'There was an error connecting to the upstream search server' );
 		} else {
+			if ( $is_cacheable ) {
+				// phpcs:ignore WordPress.WP.AlternativeFunctions.rand_mt_rand
+				wp_cache_set( $cache_key, $response, self::SEARCH_CACHE_GROUP, mt_rand( 60, 70 ) );
```
rebeccahum, line 785: Are there certain responses that we don't want to cache?

rinatkhaziev: @rebeccahum I think we only want to cache the ones that are querying for or fetching documents. I don't think we want to cache results of indexing requests or DELETEs.

rebeccahum: Is there a chance that the response could return < 400 code or something and if so, do we want to return that?

rebeccahum: What if the response is a < 400 error code? Do we want to cache that?

rinatkhaziev: I think we may, but I don't have a strong opinion. We can probably cache it for a few seconds but I'm not sure what positive impact that would have.

pschoffer:
Have you thought about the invalidation of the cache records? I am thinking about the scenario when you add a say ost and right after you try to search for the thing, but fail due to the result being cached. A naive implementation could invalidate all the cache keys on any indexing request.

Secondly, what do you think of having some URL param to bypass caching?

rebeccahum:
@pschoffer I don't think cache invalidation is necessary if we have a short TTL (60 - 70s). However, I added that passing in `vip-debug=true` will bypass it on web requests.

rebeccahum:
@sjinks Since you've been working on concurrency stuff, any input for this one?

rinatkhaziev:
I think there are a couple of things: 

1) What are yall thoughts on actually caching for a longer period of time (say 5 minutes), but check the timestamp and discard the value and go through with the remote request, but fall back on the cached one if the request fails?

2) What about cache invalidation? I think @WPprodigy mentioned an interesting approach of using https://developer.wordpress.org/reference/functions/wp_cache_get_last_changed/

We could add a hook on relevant actions (e.g. meta updates and post/term updates) that would check whether the specific meta is an indexable one and whether the post and post_status is indexable, etc. Then based on that we'd bump last_changed for our own group, Say `vip_search:$indexable`, then we could use that value to construct the cache key.

rebeccahum:
For 1), on a really active site, do you think this would end up being evicted too quickly to reap the benefit of a longer cache time?

rinatkhaziev:
@rebeccahum I had to fix a regression introduced in 70a7199, but other than that I think it's more or less ready.

sjinks (approved):

rebeccahum (approved):

rinatkhaziev:
r2130-stacks

sjinks:
v20211228.1

PR #2599, opened 2021-11-09
[CANTINA-820] VIP Search: Implement Queue Priorities

sjinks:
## Description

This PR adds support for queue priorities to the VIP Search plugin.

Added support for queue priorities

## Checklist

## Steps to Test

The CI should pass — I added unit tests for priorities.


pschoffer:
It looks ok, but the issue is that that would revert what we have done in https://github.com/Automattic/vip-go-mu-plugins/pull/2348

Having the large list of ids in the option was an issue before. We might need to come up with a more complicated structure to pass data. Maybe something like:
```
[
  {
     priority: 1,
     min_id: 12,
     max_id: 23
  }, {
     priority: 2,
     min_id: 10,
     max_id: 17
  },
  ...

]
```

Great job on the processing logic that can handle both current and future ways of passing data. We will need to do that I think :+1:

search/includes/classes/class-queue.php
```diff
@@ -313,10 +315,9 @@ public function queue_object( $object_id, $object_type = 'post', $options = arra
 
 		$index_version = $this->get_index_version_number_from_options( $object_type, $options );
 
-		$table_name = $this->schema->get_table_name();
+		$priority = (int) ( $options['priority'] ?? 5 );
 
-		// Have to escape this separately so we can insert NULL if no start time specified
-		$start_time_escaped = $next_index_time ? $wpdb->prepare( '%s', array( $next_index_time ) ) : 'NULL';
```
sjinks, line 319: `wpdb` [supports](https://github.com/WordPress/WordPress/blob/5.5/wp-includes/wp-db.php#L2245) `NULL`

search/includes/classes/class-versioning.php
```diff
@@ -14,6 +14,8 @@ class Versioning {
 	const INDEX_VERSIONS_SELF_HEAL_LOCK_CACHE_GROUP          = 'vip_search';
 	const INDEX_VERSIONS_SELF_HEAL_LOCK_CACHE_TTL            = 10;
 	const INDEX_VERSIONS_SELF_HEAL_LOCK_CACHE_TTL_ON_FAILURE = 60 * 10; // 10 minutes
+	const VERSIONING_JOB_DEFAULT_PRIORITY                    = 15;
```
pschoffer, line 17: I think adding the `INACTIVE` somewhere in the name makes it clearer. What do you think?
```suggestion
	const INACTIVE_VERSION_JOB_DEFAULT_PRIORITY                    = 15;
```

search/includes/classes/class-versioning.php
```diff
@@ -741,7 +745,17 @@ public function replicate_queued_objects_to_other_versions( $queued_objects ) {
...
-					\Automattic\VIP\Search\Search::instance()->queue->queue_object( $object_id, $object_type, $options );
+					/**
+					 * Filter do determine the priority of the replication job
+					 *
+					 * @param int $priority         Priority
+					 * @param int $object_id        Object ID
+					 * @param string $object_type   Object type
+					 * @return int                  Priority
+					 */
+					$options['priority'] = apply_filters( 'ep_versioning_reindex_priority', self::VERSIONING_JOB_DEFAULT_PRIORITY, $object_id, $object_type );
```
pschoffer, line 756: I think we should name this filter with `vip_` prefix as this is not ElasticPress feature if I understand correctly.
```suggestion
					$options['priority'] = apply_filters( 'vip_inactive_version_reindex_priority', self::INACTIVE_VERSION_JOB_DEFAULT_PRIORITY, $object_id, $object_type );
```

search/includes/classes/class-health.php
```diff
@@ -807,7 +809,16 @@ public static function reconcile_diff( array $diff ) {
...
-					\Automattic\VIP\Search\Search::instance()->queue->queue_object( $obj_to_reconcile['id'], $obj_to_reconcile['type'] );
+					/**
+					 * Filter to determine the priority of the reindex job
+					 *
+					 * @param int $priority         Job priority
+					 * @param int $object_id        Object ID
+					 * @param string $object_type   Object type
+					 * @return int                  Job priority
+					 */
+					$priority = apply_filters( 'ep_healthcheck_reindex_priority', self::REINDEX_JOB_DEFAULT_PRIORITY, $obj_to_reconcile['id'], $obj_to_reconcile['type'] );
```
pschoffer, line 820:
```suggestion
					$priority = apply_filters( 'vip_healthcheck_reindex_priority', self::REINDEX_JOB_DEFAULT_PRIORITY, $obj_to_reconcile['id'], $obj_to_reconcile['type'] );
```

search/includes/classes/class-versioning.php
```diff
@@ -764,13 +778,26 @@ public function filter__pre_ep_index_sync_queue( $bail, $sync_manager, $indexabl
...
+			foreach ( $sync_manager->sync_queue as $object_id => $value ) {
+				/**
+				 * Filter do determine the priority of the index synchronization job
+				 *
+				 * @param int $priority         Priority
+				 * @param int $object_id        Object ID
+				 * @param string $object_type   Object type
+				 * @return int                  Priority
+				 */
+				$priority            = apply_filters( 'ep_index_sync_priority', self::INDEX_SYNC_JOB_DEFAULT_PRIORITY, $object_id, $indexable_slug );
```
pschoffer, line 797: I think we should use the INACTIVE_VERSION default priority. I don't think we need to differentiate in inactive versions too much. So using `INACTIVE_VERSION_JOB_DEFAULT_PRIORITY` is fine here I think.

Which would mean using `vip_inactive_version_reindex_priority` filter probably

search/includes/classes/queue/class-cron.php
```diff
@@ -147,10 +148,22 @@ public function filter_cron_schedules( $schedule ) {
 	 *
 	 * This is the cron hook for indexing a batch of objects
 	 *
-	 * @param array $options Containing max_id and min_id keys
+	 * @param array $options Containing either max_id and min_id or option keys
 	 */
 	public function process_jobs( $options ) {
-		$jobs = $this->queue->get_jobs_by_range( $options['min_id'], $options['max_id'] );
+		if ( ! empty( $options['option'] ) ) {
+			$job_ids = get_option( $options['option'] );
```
pschoffer, line 155: Using a separate option to store Ids to avoid it being in alloptions is smart I think :+1:

pschoffer (changes requested):
Awesome job. The option idea rocks!

I left few minor notes.

`+`

I am a bit "worried" about the unique constraint [here](https://github.com/Automattic/vip-go-mu-plugins/blob/a680fad82e5ca03c1e0703493e0898caab91de19/search/includes/classes/queue/class-schema.php#L133)

From me reading the code I think the duplicate entries are swallowed without error, but as priority is not part of the unique key it might have some unexpected scenarios. Say you enqueue object `20` with priority `999`, then you would try to enqueue the same object with priority `5`. That new record will then not be included.

I guess a good enough fix is to include priority in the unique constraint. Great but possibly overengineered with performance drawbacks is to make sure min(oldPrio, newPrio) stays on the record.

sjinks:
Re: unique constraint — the `priority` field existed in the table, I did not add it. So did the unique constraint. That is why I decided not to touch it because I was not sure it would be safe.

@rinatkhaziev, @rebeccahum, I need your wisdom here. Does it make sense to expand the unique key to include the priority field? Or maybe we should rewrite the enqueue method to do smth like `INSERT … ON DUPLICATE KEY UPDATE SET priority = MIN(priority, new.priority)`?

search/includes/classes/class-versioning.php
sjinks: Fixed in 09af12f04

search/includes/classes/class-health.php
sjinks: Fixed in 09af12f04

search/includes/classes/class-versioning.php
sjinks: Fixed in 09af12f04

search/includes/classes/class-versioning.php
sjinks: Fixed in 09af12f04

rinatkhaziev:
> That is why I decided not to touch it because I was not sure it would be safe.

It would be safe-ish as long as we wrote a proper migration routine with the `dbDelta` call we would've ran in advance

> INSERT … ON DUPLICATE KEY UPDATE SET priority = MIN(priority, new.priority)?

I would heavily favor this approach for its simplicity/elegance.

sjinks:
Implemented in 0cc81661b

search/includes/classes/class-queue.php
```diff
@@ -313,25 +316,22 @@ public function queue_object( $object_id, $object_type = 'post', $options = arra
...
-			$wpdb->prepare(
-				"INSERT INTO $table_name ( `object_id`, `object_type`, `start_time`, `status`, `index_version` ) VALUES ( %d, %s, {$start_time_escaped}, %s, %d )", // Cannot prepare table name. @codingStandardsIgnoreLine
-				$object_id,
-				$object_type,
-				'queued',
-				$index_version
+			// phpcs:ignore WordPress.DB
+			$wpdb->prepare( "INSERT INTO {$table_name} (object_id, object_type, status, index_version, start_time, priority)
+				VALUES (%d, %s, 'queued', %d, %s, %d) ON DUPLICATE KEY UPDATE priority = LEAST(priority, %d)",
+				[ $object_id, $object_type, $index_version, $next_index_time, $priority, $priority ]
```
sjinks, line 334: Using the placeholder in `LEAST` because otherwise, the monkeys will finally take over the zoo:
* In MariaDB until 10.3.2, you have to use `LEAST(priority, VALUES(priority))`;
* In MariaDB since 10.3.3, it will be `LEAST(priority, VALUE(priority))`;
* In MySQL since 8.0.20, it should be `LEAST(priority, new.priority)`

As always, this will break SUDDENLY.

pschoffer (approved):
Awesome thanks for all the fixes!

sjinks:
r2025-stacks

rebeccahum:
v20211118.0

PR #2601, opened 2021-11-10
Only require vip-jetpack.php when Jetpack file was truly required.

rinatkhaziev:
## Description

Currently, we're requiring `vip-jetpack.php` unconditionally in `vip_jetpack_load`, this PR changes that to only require it the file was successfully required and `VIP_JETPACK_LOADED_VERSION` is set.

The chances of this breaking are slim but not zero, so let's have an additional safe-guard.

### Plugin Updated: Jetpack loader

Do not load Jetpack customizations in case a main Jetpack file was not included properly.

## Checklist


WPprodigy (dismissed):
Seems sound, I checked the files that end up being loaded unconditionally atm and it doesn't look like there will be any surprises. Any constants/functions/classes are isolated to this portion.

Also, unlikely this happens like you said. If `VIP_JETPACK_LOADED_VERSION` is set then nothing down here would happen anyway.

pauarge (dismissed):

jetpack.php
```diff
@@ -139,7 +139,9 @@ function vip_jetpack_load() {
 		add_filter( 'instagram_cache_oembed_api_response_body', '__return_true' );
 	}
 
-	require_once __DIR__ . '/vip-jetpack/vip-jetpack.php';
+	if ( defined( 'VIP_JETPACK_LOADED_VERSION' ) ) {
```
rebeccahum, line 142: Should we also check that `VIP_JETPACK_LOADED_VERSION` is not a `none` value? https://github.com/Automattic/vip-go-mu-plugins/blob/b12bf1908960fdf9dcd0e52ccb39edb2d2677e99/jetpack.php#L97-L100

rinatkhaziev: @rebeccahum good catch, added that.

rebeccahum (approved):
Thanks! 🚢

sjinks:
r2012-stacks

sjinks:
v20211116.0

PR #2603, opened 2021-11-11
Fix 'Invalid value supplied for foreach' warning

sjinks:
## Description

This PR fixes the `Invalid argument supplied for foreach() in /wp-content/mu-plugins/search/includes/classes/class-health.php on line 777` warning, which happens when one of the keys contains a `null` value.

Fix PHP warning in `Health::diff_document_and_prepared_document()`

## Checklist

## Steps to Test

Added a unit test for this case — the CI should pass.

### DIFF
--- a/search/includes/classes/class-health.php
+++ b/search/includes/classes/class-health.php
@@ -774,7 +774,7 @@ class Health {
 		}
 
 		// Check that there is no missing key that would only be on $prepared_document
-		foreach ( $prepared_document as $key => $value ) {
+		foreach ( (array) $prepared_document as $key => $value ) {
 			if ( ! array_key_exists( $key, $checked_keys ) ) {
 				$diff[ $key ] = array(
 					'expected' => $value,
--- a/search/includes/classes/class-health.php
+++ b/search/includes/classes/class-health.php
@@ -760,7 +760,7 @@ class Health {
 			}
 
 			if ( is_array( $value ) ) {
-				$recursive_diff = self::diff_document_and_prepared_document( $value, $prepared_document[ $key ] );
+				$recursive_diff = self::diff_document_and_prepared_document( $value, $prepared_document[ $key ] ?? [] );
 
 				if ( ! empty( $recursive_diff ) ) {
 					$diff[ $key ] = $recursive_diff;
@@ -774,7 +774,7 @@ class Health {
 		}
 
 		// Check that there is no missing key that would only be on $prepared_document
-		foreach ( (array) $prepared_document as $key => $value ) {
+		foreach ( $prepared_document as $key => $value ) {
 			if ( ! array_key_exists( $key, $checked_keys ) ) {
 				$diff[ $key ] = array(
 					'expected' => $value,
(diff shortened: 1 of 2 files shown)


sjinks:
php74-core* tests fail with the recent 5.8.2 update :-(

pschoffer:
Could you describe the problem some more or is there an easy way to reproduce it? Maybe we could add that case to `test_simplified_diff_document_and_prepared_document`? I would personally prefer an earlier exit than casting but I don't think I understood the issue to be honest :)

pschoffer:
As we chatted with @sjinks .

I think the issue might be few steps above:
I guess the scenario is

```
$document = [
   'a' => [
   ]
]
$prepared_document = [
  'a' => NULL
]
or maybe 
$prepared_document = []
```

We might want to do something like:
```
if ( is_array( $value ) ) {
    $prepared_value = $prepared_document[ $key ] || array();
    $recursive_diff = self::diff_document_and_prepared_document( $value, $prepared_value );
```

sjinks:
I have extracted and run the code on the document that triggered the error in a PHP Sandbox, and it looks like we have more issues:

```
<br />
<b>Notice</b>:  Undefined index: _dt_aop_include_in_feed in <b>[...][...]</b> on line <b>984</b><br />
<br />
<b>Notice</b>:  Trying to access array offset on value of type null in <b>[...][...]</b> on line <b>984</b><br />
<br />
<b>Notice</b>:  Undefined index: _dt_article_type in <b>[...][...]</b> on line <b>984</b><br />
<br />
<b>Notice</b>:  Trying to access array offset on value of type null in <b>[...][...]</b> on line <b>984</b><br />
<br />
<b>Notice</b>:  Undefined index: dtseo_title in <b>[...][...]</b> on line <b>984</b><br />
<br />
<b>Notice</b>:  Trying to access array offset on value of type null in <b>[...][...]</b> on line <b>984</b><br />
<br />
<b>Notice</b>:  Undefined index: dtseo_description in <b>[...][...]</b> on line <b>984</b><br />
<br />
<b>Notice</b>:  Trying to access array offset on value of type null in <b>[...][...]</b> on line <b>984</b><br />
```

(line numbers do not match the original code :-( )

sjinks:
tl;dr: the issue happened when the document has some key, but the prepared document does not. Then `$prepared_document[ $key ]` generated the "Undefined index" notice and returned `null`. `diff_document_and_prepared_document()` was then called with `null` as its second argument, which generated the 'Invalid value supplied for foreach' warning. If the first argument happened to be an array, we would get the 'Trying to access array offset on value of type null' notice as well.

For example, this happened on this simplified version of two documents:

```php
$document = [
	'meta'                    => [
		'_dt_aop_include_in_feed' => [
			0 => [
				'value'    => '',
				'raw'      => '',
				'boolean'  => false,
				'date'     => '1971-01-01',
				'datetime' => '1971-01-01 00:00:01',
				'time'     => '00:00:01', 
			],
		],
	],
];

$prepared_document = [ 'meta' => [] ];
```

pschoffer (approved):
Awesome this is a great way to fix it I think!

sjinks:
r1996-stacks

PR #2604, opened 2021-11-11
Add type annotations so that static code analysis and code completion work properly

sjinks:
## Description

This PR adds type annotations to the files I have worked recently with so that static code analysis and code completion features work properly.

### Unit Tests

Add type annotations

## Checklist

## Steps to Test

N/A

### DIFF
--- a/search/includes/classes/class-health.php
+++ b/search/includes/classes/class-health.php
@@ -2,10 +2,6 @@
 
 namespace Automattic\VIP\Search;
 
-use ElasticPress\Indexable\Comment\Comment;
-use ElasticPress\Indexable\Post\Post;
-use ElasticPress\Indexable\Term\Term;
-use ElasticPress\Indexable\User\User;
 use \ElasticPress\Indexables as Indexables;
 
 use \WP_Query as WP_Query;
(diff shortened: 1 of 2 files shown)


search/includes/classes/class-health.php
```diff
@@ -131,7 +135,7 @@ public function validate_index_entity_count( array $query_args, \ElasticPress\In
 	 *      'post_status' => array( $post_statuses )
 	 * ];
 	 *
-	 * @param mixed $indexable Instance of an ElasticPress Indexable Object to search on
+	 * @param Comment|Post|Term|User $indexable Instance of an ElasticPress Indexable Object to search on
```
pschoffer, line 138: Should this be `Indexable` instead?

pschoffer (changes requested):

search/includes/classes/class-health.php
sjinks: I am afraid it shouldn't — `Indexable` does not have a `format_args()` method used by the code. That's why I had to enumerate all classes that have.

sjinks: https://github.com/Automattic/vip-go-mu-plugins/blob/2a822a6f5202daaf9085d7d45acdac0d69ae2fd3/search/includes/classes/class-health.php#L145

pschoffer: Hmm I see. Probably it should though? Maybe it should be an abstract class defining `format_args` as an abstract method. That it is how it is being used anyways I feel like, what do you think?

sjinks: I will address this in another PR: I saw several places where `Indexable` does not fit while working with the Search plugin; chances are that `format_args()` is not the only method.

sjinks: Actually, I am not sure this is even possible — those classes are defined in a third-party module, we cannot even attach an interface to them :-( The upstream will need to fix this.

pschoffer (approved):
LGTM

sjinks:
r1998-stacks

PR #2609, opened 2021-11-11
Removing custom Akismet filters

pauarge:
## Description

We introduced a couple of filters that checked if Akismet was correctly installed and active, and if that wasn't the case, it would hide some elements from `wp-admin`. This comes at the cost of an additional HTTP request, that might impact the server's performance. 

Given the fact that Akismet is always enabled (unless Jetpack is not) and VIP's Connection Pilot should connect it automatically, we no longer need to hide those messages. The only situation in which they would show up and they would redirect to a non-connected site is when CP fails, in which case is fine that the user sees that something is wrong.

### Removing custom Akismet filters

Getting rid of some custom code regarding Akismet messaging.

## Checklist

## Steps to Test

1. Check out PR on a Sandbox.
2. Disconnect Akismet manually through it's wp-admin page.
3. Expect to see normal messaging about Akismet not being active.
4. Let Connection Pilot reconnect Akismet.

### DIFF
--- a/akismet.php
+++ b/akismet.php
@@ -50,20 +50,6 @@ function wpcom_vip_akismet_spam_count_incr( $val ) {
 }
 add_filter( 'akismet_spam_count_incr', 'wpcom_vip_akismet_spam_count_incr' );
 
-function vip_remove_akismet_admin_menu() {
-	if ( is_akismet_key_invalid() ) {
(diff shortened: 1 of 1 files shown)


sjinks (approved):
:rocket:

sjinks:
v20211116.0

PR #2610, opened 2021-11-11
Exclude private taxonomies via filter

pschoffer:
## Description
We are excluding some taxonomies in our fork of [EP](https://github.com/Automattic/ElasticPress/blob/0f8e520f2cdda2bea6450cf04e2bffcc71446116/includes/classes/Indexable/Post/Post.php#L1014).

The change comes from this [requirement](https://github.com/Automattic/ElasticPress/pull/13).

In one of the previous EP releases new filter `ep_post_tax_excluded_wp_query_root_check` was introduced that can be leveraged to do the same thing without diverging from upstream.

This change starts to use the filter. In follow-up change, we can remove our custom change inside EP fork.
Follow-up change: https://github.com/Automattic/ElasticPress/pull/133

The first step towards moving excluded protected taxonomies from post queries. 

After this and follow-up change will be done, we will be able to enable the ability to change the list of excluded taxonomies instead of the current hard-coded solution.

## Checklist

## Steps to Test

This is a bit tough. This is what I have done:
1) In EP - https://github.com/Automattic/ElasticPress/blob/0f8e520f2cdda2bea6450cf04e2bffcc71446116/includes/classes/Indexable/Post/Post.php#L1014 comment out lines 1015 - 1017 with our custom code.
2) Inside the `if` adding the taxanomy to `$args['tax_query']` add a debug line: `var_dump(['including taxanomy', $tag_slug])`
3) Add following code somewhere (can be a `mu-plugin`):
```
		add_filter( 'posts_pre_query', function($posts, $query) {
			$query->query_vars['fields'] = '123';
		}, 10, 2);
		add_action('init', function() {
			register_taxonomy( 'fields', ['post'] );
		});
```
4) Search on site -> you should see `fields` being included in the query in post_filters and the debug line shown
5) Apply the change and search again. `fields` is no longer included.

image


search/includes/classes/class-search.php
```diff
@@ -63,6 +63,74 @@ class Search {
 		'advanced_seo_description',
 	);
 
+	public const POST_PROTECTED_TAXANOMIES = array(
```
sjinks, line 66: Probably a spelling error? TaxOnomies

sjinks: Question: are these really taxonomies? They look like query parameters to me.

pschoffer: Yeah probably not the best name. 

The issue is that those are kinda reserved keywords in a request towards elasticsearch. If you would create a custom taxanomy with the same search name and query for it, it would overwrite these keywords like `fields`.

Any suggestions on how to name it?

sjinks: Maybe a variation of RESERVED_NAMES? Summoning @rinatkhaziev, as his English-fu is much better than mine

pschoffer: How does `ES_QUERY_RESERVED_NAMES` sound?

search/includes/classes/class-search.php
```diff
@@ -2022,4 +2092,8 @@ public function maybe_change_index_version() {
 			$this->versioning->set_current_version_number( $indexable, $normalized_version );
 		}
 	}
+
+	public function exclude_es_query_reserved_names( $taxanomies ) {
+		return array_merge( $taxanomies, self::ES_QUERY_RESERVED_NAMES );
```
sjinks, line 2097: Could you please make it $tax**o**nomies? :-)

search/includes/classes/class-search.php
```diff
@@ -63,6 +63,74 @@ class Search {
 		'advanced_seo_description',
 	);
 
+	public const ES_QUERY_RESERVED_NAMES = array(
```
sjinks, line 66: Just as an idea — it does not necessarily go into this PR. We have this (or a similar to this) list in WordPress:

https://github.com/WordPress/WordPress/blob/5.8.2/wp-includes/class-wp.php#L17
https://github.com/WordPress/WordPress/blob/5.8.2/wp-includes/class-wp.php#L27

Does it make sense to generate our stop list from those WP lists? The advantage of this (as I see it) is that the code will be future-proof — if WP adds another reserved query variable, we won't have to update our code.

search/includes/classes/class-search.php
pschoffer: Such a weird wor**l**d :)

search/includes/classes/class-search.php
pschoffer: Oh that is a great idea!,,, I'll give it a try

sjinks (approved):

rebeccahum:
@pschoffer Do you mind making a PR in https://github.com/Automattic/ElasticPress/ to remove the VIP-specific changes?

search/includes/classes/class-search.php
```diff
@@ -2022,4 +2024,9 @@ public function maybe_change_index_version() {
 			$this->versioning->set_current_version_number( $indexable, $normalized_version );
 		}
 	}
+
+	public function exclude_es_query_reserved_names( $taxonomies ) {
+		global $wp;
+		return array_merge( $taxonomies, $wp->public_query_vars, $wp->private_query_vars );
```
rebeccahum, line 2030: With the latest change, I'm seeing certain names previously on the list left off now, e.g. `author__in`. Was that the intent, @pschoffer? Or were we supposed to include all parameters in https://developer.wordpress.org/reference/classes/wp_query/#parameters?

pschoffer: I am not sure to be honest... I couldn't find too much information in the original PR. I'll reach out to folks on slack to see if they remember more.

pschoffer:
@rebeccahum / @sjinks what do you think about this PR when we get the input from Nick? 
> oh! yes i think that's right - I think you were able to do things like $wp_query_vars[ $taxonomy_slug ] = ...  - but if your site had a taxonomy with a slug of one of the "built in" WP Query arguments (say, author...which is I think the one we had problems with), it would get interpreted as a taxonomy by the code that is now bypassed, breaking the "regular" WP Query behavior
so if you did WP_Query( array( 'author' => 'bob' ) ); , EP would turn that into a a taxonomy query for bob, rather than searching against the author field on the post document (edited) 

I think I would still lean towards using the core constants so that we don't need to maintain the list. Although that means we will miss some edge cases like `author__in`. But I am happy to revert to the list if we think we should cover all the edge cases too.

pschoffer:
I think that would increase the complexity of it. I will revert the use of core constants.

sjinks (approved):

PR #2619, opened 2021-11-15
Add :preserveSemverRanges preset to renovate.json

sjinks:
Pinned dependencies create complicated issues when it comes to security updates: because package versions are tightly locked, it is often impossible to update a vulnerable package without going to the upstream repo, fixing the issue there, waiting until the change gets published, etc — repeat for all affected packages (this is the way I went with, for example, `eslines`).

Pinning itself does not add much security — it may only protect against a bad release of a direct dependency, but not against a bad release of a transitive dependency (e.g., we may pin, say, webpack, but we can do nothing to 20+ of its dependencies).

We use lock files (`package-lock.json`) and `npm ci` during our CI process, which guarantees that we have a consistent set of dependencies both in our local and CI environments. Unless we run `npm up` explicitly, pinning provides no clear benefits for us.


rinatkhaziev (approved):
thanks, looks good.

PR #2620, opened 2021-11-15
Search: Bail and do not log when request is index_exists

rebeccahum:
## Description
We should not be logging any remote calls to `index_exists` as a failed request. 

`index_exists()` [does a remote request to determine if a 200 or a 404 is being returned](https://github.com/Automattic/ElasticPress/blob/fef833793a85e0229bf6fa3f66ba75333f00d8c8/includes/classes/Elasticsearch.php#L1001-L1023), but if there's no index, it gets logged as a failed request: https://github.com/Automattic/vip-go-mu-plugins/blob/3ee96e7db66a9345858f66eb0f28bd23c0172280/search/includes/classes/class-search.php#L729-L730

** DEPLOY NOTE: THIS MUST GO OUT AFTER (or in tandem with) https://github.com/Automattic/vip-go-mu-plugins/pull/2621 **

## Checklist

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -1990,7 +1990,7 @@ class Search {
 		}
 
 		$this->logger->log( 'warning', 'search_query_failure', $message, [
-			'request' => $request,
+			'request'    => $request,
 			'query_args' => $query_args,
 		] );
 	}
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -690,7 +690,16 @@ class Search {
 		return 500;
 	}
 
-	public function filter__ep_do_intercept_request( $request, $query, $args, $type ) {
+	/**
+	 * Filter to intercept EP remote requests.
+	 * 
+	 * @param  array  $request New remote request response
+	 * @param  array  $query   Remote request arguments
+	 * @param  array  $args    Request arguments
+	 * @param  string $type    Type of request
+	 * @return array  $request New request
+	 */
+	public function filter__ep_do_intercept_request( $request, $query, $args, $type = null ) {
 		// Add custom headers to identify authorized traffic
 		if ( ! isset( $args['headers'] ) || ! is_array( $args['headers'] ) ) {
 			$args['headers'] = [];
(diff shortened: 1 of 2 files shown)


search/includes/classes/class-search.php
```diff
@@ -690,7 +690,7 @@ public function filter__ep_bulk_items_per_page() {
 		return 500;
 	}
 
-	public function filter__ep_do_intercept_request( $request, $query, $args ) {
+	public function filter__ep_do_intercept_request( $request, $query, $args, $type ) {
```
rinatkhaziev, line 693: @rebeccahum Is there a scenario where changing this method's signature may break things?

search/includes/classes/class-search.php
```diff
@@ -1987,7 +1990,7 @@ public function log_ep_invalid_response( $request, $query, $query_args, $query_o
 		}
 
 		$this->logger->log( 'warning', 'search_query_failure', $message, [
-			'request'    => $request,
+			'request' => $request,
```
rinatkhaziev, line 1993: I think arrow alignment is the official way?

rebeccahum: Whoops didn't mean to do this.

search/includes/classes/class-search.php
rebeccahum: I don't believe so since `$type` is assigned a `null` value by default from `remote_request()`: https://github.com/Automattic/ElasticPress/blob/e623992b60a871c3d6396e35793f217ebb2a5b39/includes/classes/Elasticsearch.php#L1071.

rebeccahum: @rinatkhaziev I don't believe so since `$type` is passed in as a `null` value by default in `remote_request()` if not specified: https://github.com/Automattic/ElasticPress/blob/e623992b60a871c3d6396e35793f217ebb2a5b39/includes/classes/Elasticsearch.php#L1071.

rinatkhaziev: Just to be extra-safe and safe-guard from a potential revert upstream. 
```suggestion
	public function filter__ep_do_intercept_request( $request, $query, $args, $type = ''  ) {
```

rinatkhaziev:
Thanks! I left an inline suggestion. On a side note this method is missing the PHPDoc declaration, could you please add it while you're here?

rinatkhaziev (dismissed):
^^

rebeccahum:
@rinatkhaziev done and done!

pschoffer (approved):
This is great. Thank you!

rebeccahum:
v20211117.0

PR #2622, opened 2021-11-15
Update a8c-files.php

katag9k:
## Description
Changing comments in `init_photon` since eliminating VIP photon filters is yet to be determined.

### DIFF
--- a/a8c-files.php
+++ b/a8c-files.php
@@ -108,7 +108,7 @@ class A8C_Files {
 
 	public function init_photon() {
 		// Limit to certain contexts for the initial testing and roll-out.
-		// This will be phased out and become the default eventually.
+		// This maybe phased out and become the default eventually.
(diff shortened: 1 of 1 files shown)


a8c-files.php
```diff
@@ -108,7 +108,7 @@ private function init_vip_filesystem() {
 
 	public function init_photon() {
 		// Limit to certain contexts for the initial testing and roll-out.
-		// This will be phased out and become the default eventually.
+		// This maybe phased out and become the default eventually.
```
rinatkhaziev, line 111: Thought on making it even more explicit?
```suggestion
		// This check may be phased out and Photon may become the default eventually.
```

sjinks (approved):

sjinks:
r2010-stacks

sjinks:
v20211116.0

PR #2630, opened 2021-11-18
Remove XDebug and Speed Up CI

sjinks:
This PR tries to speed up our CI by eliminating the dependency on XDebug and mocking `header()` and related functions for tests that use it. In many cases, this also allowed us to get rid of the `@runInSeparateProcess` annotation, which also positively affected performance.

On averages, the builds are now 1 to 2 minutes faster (I guess that depends on CircleCI's workload, need to test this in a more quiet time). Well, it ain't much but it's honest work :grinning:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -101,6 +101,7 @@ job-references:
       MYSQL_USER: << pipeline.parameters.db_user >>
       MYSQL_PASSWORD: << pipeline.parameters.db_pass >>
       MYSQL_DB: << pipeline.parameters.db_name >>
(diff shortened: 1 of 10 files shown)


pschoffer (approved):
oh nice!

I wonder if this will enable us to not add those annotations to core tests - https://github.com/Automattic/vip-go-mu-plugins/blob/467244671acdd70470c3c397085a2fcd677cd535/bin/utils.sh#L134

sjinks:
@pschoffer not sure, TBH; I have not looked yet into the core tests. If they use things like `define()`, then we cannot avoid separate processes.

PR #2616, opened 2021-11-13
Drop eslines and eslint

sjinks:
## Description

As discussed in P2, this PR drops eslint and eslines, and moves the PHPCS checks into the pre-commit hook.

## Checklist

Please make sure the items below have been covered before requesting a review:

- [x] This change works and has been tested locally (or has an appropriate fallback).
- [ ] This change works and has been tested on a Go sandbox.
- [ ] This change has relevant unit tests (if applicable).
- [ ] This change has relevant documentation additions / updates (if applicable).
- [ ] I've created a changelog description that aligns with the provided examples.

## Steps to Test

The CI should pass.

PR #2636, opened 2021-11-19
Update search/elasticpress to 0f8e520f2cdda2bea6450cf04e2bffcc71446116

sjinks:
Undo the [erroneous change](https://github.com/Automattic/vip-go-mu-plugins/pull/2616/files#diff-9db63be1940a069c031ee68c9db04598ceffb8f0dedaa3ef5c681384992d0c51) from #2616.


rinatkhaziev (approved):
Thanks for catching this.

PR #2638, opened 2021-11-19
Jetpack plan: Update VIP object for jetpack_active_plan option

rebeccahum:
## Description
We're in the process of switching over Jetpack Professional (deprecated) to Complete. We need to update the default `jetpack_active_plan` option.

PR for the last time we updated: https://github.com/Automattic/vip-go-mu-plugins/pull/1431

This needs to go alongside with the below changes:
- https://github.com/Automattic/jetpack/pull/21833
- Update provisioning to use `jetpack-complete` plan

## Checklist


kraftbj (approved):

WPprodigy:
Are "active" ones going to be activated automatically for existing sites, or is this something that takes place during the initial provisioning only?

kraftbj:
`active` is different in this context than active modules. This means features that are active within that Plan, e.g. available. 

https://github.com/Automattic/jetpack/blob/master/projects/plugins/jetpack/class.jetpack-plan.php#L347 uses the `active` to field to determine if a feature is supported or not.

The terminology comes from the Store codebase and basically means "this feature is available as part of the current plan".

WPprodigy (approved):

rebeccahum:
@kraftbj A follow-up...this pre-option return should not negatively affect sites with a "Professional" plan in any way, correct?

kraftbj:
Correct. Since we are effectively telling Jetpack that these sites have a "vip" plan, Professional vs Complete is immaterial with Jetpack.

rinatkhaziev:
r2087-stacks

PR #2639, opened 2021-11-20
Cleanup JP connection pilot cron registration & run conditions

WPprodigy:
## Description

- Lower the max backoff for connection attempts to once a week.
- Cleanup cron registration:
  - Remove the internal job
  - Unregister the job when auto-connections are disabled
  - Limit conditions when it is registered
 - Expose CLI command regardless of `VIP_JETPACK_AUTO_MANAGE_CONNECTION` 

### Jetpack Connections: Cron job cleanup

Lower the max backoff time for connection attempts, and unregister the cron job when it's not needed.

## Checklist

## Steps to Test

Not much to test here, but could disable `VIP_JETPACK_AUTO_MANAGE_CONNECTION` and ensure the job is gone (keep batch containers in mind).

And then manually set a backoff factor to something more than 24 * 7 and set an old timestamp, then ensure the event still runs. Unit tests cover this already though :)

### DIFF
--- a/000-vip-init.php
+++ b/000-vip-init.php
@@ -139,6 +139,8 @@ $internal_cron_events = array(
 	),
 );
 
+define( 'CRON_CONTROL_ADDITIONAL_INTERNAL_EVENTS', $internal_cron_events );
+
 // Enable Jetpack private connection by default on non production sites
 if ( ! defined( 'VIP_JETPACK_IS_PRIVATE' ) && defined( 'VIP_GO_APP_ENVIRONMENT' ) && 'production' !== VIP_GO_APP_ENVIRONMENT ) {
 	define( 'VIP_JETPACK_IS_PRIVATE', true );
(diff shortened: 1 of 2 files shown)


000-vip-init.php
```diff
@@ -139,6 +139,8 @@
 	),
 );
 
+define( 'CRON_CONTROL_ADDITIONAL_INTERNAL_EVENTS', $internal_cron_events );
```
pauarge, line 142: Do we need to move this line up or is it fine to simply delete from line 152 to 158?

vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
```diff
@@ -21,18 +25,10 @@ class Connection_Pilot {
 	 */
 	const CRON_ACTION = 'wpcom_vip_run_jetpack_connection_pilot';
 
-	/**
-	 * The schedule the cron job runs on. Update in 000-vip-init.php as well.
-	 *
-	 * Schedule changes can take up to 24 hours to take effect.
-	 * See the a8c_cron_control_clean_legacy_data event for more details.
-	 */
-	const CRON_SCHEDULE = 'hourly';
```
pauarge, line 30: Why do we change this constant to a string literal?

vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
```diff
@@ -187,6 +177,13 @@ public function reconnect() {
 	private function should_back_off(): bool {
 		if ( ! empty( $this->last_heartbeat['backoff_factor'] ) && ! empty( $this->last_heartbeat['timestamp'] ) ) {
 			$backoff_factor = $this->last_heartbeat['backoff_factor'];
+
+			// Ensure we don't go past the max, and support future decreases should they occur.
+			if ( $backoff_factor > self::MAX_BACKOFF_FACTOR ) {
+				$backoff_factor = self::MAX_BACKOFF_FACTOR;
+				$this->update_heartbeat( $backoff_factor );
+			}
+
```
pauarge, line 186: 👍

pauarge (approved):
Looks good, I'm approving but left a couple of small comments.

000-vip-init.php
WPprodigy: Moved it up a few lines since the JP-related constants no longer affect it, just so it's closer to it's source. Left $internal_cron_events as an array though in case there is need in the future to add to it further based on conditions.

vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
WPprodigy: Only used in one-place, and I figured moving closer to it's source would make future devs more aware that changing the schedule requires more thought than just updating the const.

PR #2642, opened 2021-11-22
Make Alerts more testable

sjinks:
I have refactored Alerts tests so that they don't have to be run in a separate process.

The main changes are:
* the testable methods are now `protected`, not `private`. This allows us not to use the Reflection API and makes tests more readable;
* I have introduced two `protected static` helper methods: `get_service_address()` and `get_service_port`. They can be overridden by a descendant; their raison d'être is to allow us not to use PHP constants and, as a consequence, to be able to run all tests in a single process;
* another helper method is `clear_instance()`: because Singletons are hard to test :-( 

With these changes, it takes 2.68 seconds to complete the tests (vs 1.08 minutes).

### DIFF
--- a/lib/utils/class-alerts.php
+++ b/lib/utils/class-alerts.php
@@ -52,7 +52,7 @@ class Alerts {
 	 *
 	 * @return array|WP_Error Response details from wp_remote_post
 	 */
-	private function send( array $body ) {
+	protected function send( array $body ) {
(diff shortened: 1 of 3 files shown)


rebeccahum (approved):
This is fantastic, thank you!

sjinks:
r2031-stacks

sjinks:
There is no changelog here because there are no production code changes.

PR #2643, opened 2021-11-22
Refactor mail tests

sjinks:
Just like #2633 but for VIP Mail tests: refactor `@runInSeparateProcess` tests so that they don't need a separate process to run.


rebeccahum (approved):
Awesome!

PR #2644, opened 2021-11-22
Search: query ES for the last post ID in get_last_post_id()

rebeccahum:
## Description
`get_last_post_id()` only queries the DB for the last post ID, which means if we have an extra post in the ES index, our `validate-contents` will not include it in the range and therefore, it will be missed.

Fix bug where `get_last_post_id()` doesn't account for extra ID in ES index.

## Checklist

## Steps to Test
0. Check out PR
1. Create a post - `vip dev-env --slug test exec -- wp post create --post_title=test --post_status=publish`
2. Delete the created post directly in DB - `vip dev-env --slug test exec -- wp db cli + DELETE FROM wp_posts WHERE ID = 50;`
3. Flush the cache - `vip dev-env --slug test exec -- wp cache flush`
4. Run `dev-env exec --slug test -- wp vip-search health validate-contents`

You should see the last post in ES in the range. For this example, we are using `50`:

```
Validating posts 1 - 50...✘ (attempted to reconcile)
```


search/includes/classes/class-health.php
```diff
@@ -836,8 +836,24 @@ public static function get_last_post_id() {
...
-		$last = $wpdb->get_var( "SELECT MAX( `ID` ) FROM $wpdb->posts" );
+		$last_db_id = $wpdb->get_var( "SELECT MAX( `ID` ) FROM $wpdb->posts" );
 
+		$indexable = \ElasticPress\Indexables::factory()->get( 'post' );
+
+		$es_result = false;
+		if ( $indexable ) {
+			$query_args     = [
+				'posts_per_page' => 1,
+				'orderby'        => 'ID',
```
rinatkhaziev, line 847: I think it's better to be explicit here.

```suggestion
				'orderby'        => 'ID',
				'order'        => 'desc',
```

rinatkhaziev (changes requested):
Nice catch! It'd be good to add PHPDoc for the method that would state that this both checks DB and ES last post ID.

brandon-m-skinner:
I think the underlying issue around figuring out if an ES index has posts beyond the current last post ID should be resolved another way or the function should be renamed. The database is the source of truth in this case and having a function called `get_last_post_id` that may return a post ID outside of what the source of truth says is inconsistent with what would be expected. Can end up with some interesting bugs if someone uses it outside of this specific use case with the expectation that the returned ID is linked to an existing fetchable post.

rebeccahum:
OK, I've split it into `get_last_db_post_id()` and `get_last_es_post_id()` so the user will explicitly use `get_last_db_post_id()` if they are expecting the DB to be the source of truth.

rinatkhaziev (approved):

rebeccahum:
r2042-stacks

PR #2646, opened 2021-11-23
Refactor password tests

sjinks:
Refactor password tests to allow them to run in a single process.

This PR adds the necessary clean-up logic, and drops `@runInSeparateProcess` declarations for `test-password.php`.

Timings (for my local environment, it is *slower* than CircleCI):
* before: 21.34 seconds
* after: 3.95 seconds


pauarge (approved):
Looking great

PR #2651, opened 2021-11-23
Refactor tests for private sites and Lockout component

sjinks:
This PR refactors tests for Private Sites and Lockouts and Warnings by mocking all calls to `defined()` and `constant()`; this allows us to run all those tests in a single process.

For this to work, I had to patch `security/class-private-sites.php` and `security/class-lockout.php` to use `constant()` instead of accessing constants directly.

Mocking relies upon the [name resolution rules](http://php.net/manual/en/language.namespaces.rules.php) of PHP namespaces. However, because the PHP compiler tries to optimize away all calls to `defined()`, it is crucial that the mock is loaded before the tested class.

Timings (PS + Lockout)
* Before: 23.92 seconds + 31.89 seconds
* After: 2.68 seconds + 3.53 seconds

### DIFF
--- a/security/class-lockout.php
+++ b/security/class-lockout.php
@@ -43,9 +43,9 @@ class Lockout {
 		if ( defined( 'VIP_LOCKOUT_STATE' ) ) {
 			$user = wp_get_current_user();
 
-			switch ( VIP_LOCKOUT_STATE ) {
+			switch ( constant( 'VIP_LOCKOUT_STATE' ) ) {
 				case 'warning':
-					$show_notice = apply_filters( 'vip_lockout_show_notice', $user->has_cap( 'manage_options' ), VIP_LOCKOUT_STATE, $user );
+					$show_notice = apply_filters( 'vip_lockout_show_notice', $user->has_cap( 'manage_options' ), constant( 'VIP_LOCKOUT_STATE' ), $user );
 					if ( $show_notice ) {
 						$this->render_warning_notice();
 
@@ -55,7 +55,7 @@ class Lockout {
 					break;
 
 				case 'locked':
-					$show_notice = apply_filters( 'vip_lockout_show_notice', $user->has_cap( 'edit_posts' ), VIP_LOCKOUT_STATE, $user );
+					$show_notice = apply_filters( 'vip_lockout_show_notice', $user->has_cap( 'edit_posts' ), constant( 'VIP_LOCKOUT_STATE' ), $user );
 					if ( $show_notice ) {
 						$this->render_locked_notice();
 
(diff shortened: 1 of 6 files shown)


tests/security/test-class-private-sites.php
```diff
@@ -110,14 +96,8 @@ public function test__filter_jetpack_get_available_modules() {
 		$this->assertEquals( $expected, $filtered );
 	}
 
-	/**
-	 * @runInSeparateProcess
-	 * @preserveGlobalState disabled
-	 */
 	public function test__filter_blog_public_option_for_sync() {
-		define( 'VIP_JETPACK_IS_TRUE', true );
```
sjinks, line 118: This constant does not seem to be used anymore.

security/class-lockout.php
```diff
@@ -43,9 +43,9 @@ public function add_admin_notice() {
 		if ( defined( 'VIP_LOCKOUT_STATE' ) ) {
 			$user = wp_get_current_user();
 
-			switch ( VIP_LOCKOUT_STATE ) {
+			switch ( constant( 'VIP_LOCKOUT_STATE' ) ) {
```
jesse-greathouse, line 46: I don't understand why this requires the use of "constant()", usually I only see that where the name of the constant is dynamic. Can you explain for my curiosity?

jesse-greathouse (commented):
Very good tests however Im puzzled by the use of "constant()". I would prefer to see the conventional syntax but Im sure there is a reason Im just not seeing.

jesse-greathouse (approved):

security/class-lockout.php
sjinks: We can mock the `constant()` function, but cannot mock the constant itself.

If we use
```
define( 'CONSTANT', 'value' );
```

In a test, we have to annotate it with `@runInSeparateProcess`. PHPUnit will then run it in a separate process. But this incurs a quite significant performance penalty.

If we don't want a separate process, we need a way to "undefine" a constant somehow. But there is no way to do that on PHP (well, I can write a PHP extension for that, but that's a different story).

Because we can mock only functions, we have to use a function that will return the constant's value (`constant()`).

Now, since all our functions are namespace, PHP will look for a function in the current namespace first, and then in the global namespace.

Thus this:
```php
defined( 'VIP_JETPACK_IS_PRIVATE' ) && true === constant( 'VIP_JETPACK_IS_PRIVATE' )
```

is first tried like this:
```
\Automattic\VIP\Security\defined( 'VIP_JETPACK_IS_PRIVATE' ) && true === \Automattic\VIP\Security\constant( 'VIP_JETPACK_IS_PRIVATE' )
```

and only then as this:
```php
\defined( 'VIP_JETPACK_IS_PRIVATE' ) && true === \constant( 'VIP_JETPACK_IS_PRIVATE' )
```

Because we provide our own wrappers for [`defined()`](https://github.com/Automattic/vip-go-mu-plugins/pull/2651/files#diff-85fe9ea05584f9a0107947de85ab55ef1c121e50b020c71bc8c9de83a5367b80R38) and [`constant()`](https://github.com/Automattic/vip-go-mu-plugins/pull/2651/files#diff-85fe9ea05584f9a0107947de85ab55ef1c121e50b020c71bc8c9de83a5367b80R42), we don't actually have to define a "global" constant.

jesse-greathouse: Thanks for helping me understand this. You are a gem. 👍

sjinks:
r2036-stacks

PR #2652, opened 2021-11-23
Search: Cache index_exists requests

rebeccahum:
## Description
We should start caching index_exists requests since we use it as a check in many places (e.g. get_current_field_count).

Cache index_exists requests into an option.
## Checklist

## Steps to Test
1) Check out PR
2) Shell in
3) Do an index_exists request:
```
$indexable->index_exists();
```
4) Check it exists in the option: `wp option get es_index_exists_vip-200508-post-1` (you may need to change it accordingly to your index name since the format is `es_index_exists_<index-name>`
5) Delete the index: `wp vip-search delete-index`
6) Check the option does not exist `wp option get es_index_exists_vip-200508-post-1`

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -417,6 +417,8 @@ class Search {
 
 		// Network layer replacement to use VIP helpers (that handle slow/down upstream server)
 		add_filter( 'ep_intercept_remote_request', '__return_true', 9999 );
+		add_filter( 'ep_do_intercept_request', [ $this, 'get_cached_index_exists_request' ], 9998, 5 );
+		add_filter( 'ep_do_intercept_request', [ $this, 'invalidate_cached_index_exists_request' ], 9998, 5 );
 		add_filter( 'ep_do_intercept_request', [ $this, 'filter__ep_do_intercept_request' ], 9999, 5 );
 
 		// Disable query integration by default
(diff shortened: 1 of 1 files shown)


WPprodigy:
What are the reasons for caching via an option over memcache? The pro I can think of is of course longevity, but one negative that comes to mind is data syncs / SQL imports. The options could start to lie about the current state of things.

rebeccahum:
Using memcached feels a bit too volatile for this use case and I'm also under the impression that index re-building/deleting shouldn't be happening very often — thus, I don't expect the option to change much. When data syncs/SQL imports occur, bulk re-indexing should be kicked off afterwards. However, I can amend the PR to invalidate on bulk re-index if that makes sense?

sjinks (dismissed):

search/includes/classes/class-search.php
```diff
@@ -688,6 +690,77 @@ public function filter__ep_bulk_items_per_page() {
...
+		];
+		if ( ! in_array( $type, $index_actions, true ) ) {
+			return $request;
+		}
+
+		$option_name = $this->get_index_exists_option_name( $query['url'] );
+		$option      = get_option( $option_name );
+		if ( false !== $option ) {
+			delete_option( $option_name );
+		}
```
WPprodigy, line 746: Minor nitpick, but can just `delete_option()` option here directly without checking get_option first. WP won't delete if it can't SELECT first an ID to delete anyway.

WPprodigy (approved):

rebeccahum:
r2041-stacks

PR #2653, opened 2021-11-23
Run VIP Cache Manager tests in a single process

sjinks:
This PR removes the `@runTestsInSeparateProcesses` declaration from `VIP_Go_Cache_Manager_Test`, as the tests seem safe to run in the same process.

Timings:
* before: 25.2 seconds
* after: 3.88 seconds


rinatkhaziev (approved):
⏲️

PR #2656, opened 2021-11-23
Refactor Context tests

sjinks:
Run all Context tests in a single process.

Timings:
* before: 1.16 minutes
* after: 4.81 seconds


rebeccahum (approved):

sjinks:
r2039-stacks

PR #2657, opened 2021-11-23
Search: Add lock if there is already an ongoing healthcheck

rebeccahum:
## Description

If there is already an ongoing healthcheck job and it doesn't finish before the next one runs (i.e. it is scheduled within a minute of each other), we should not run it in parallel.

Add lock to healthcheck job to prevent parallel runs.

## Checklist

## Steps to Test
0) Check out PR
1) Shell in
2) `$es = new \Automattic\VIP\Search\Search();`
3) `$es->init();`
4) `$hj = new \Automattic\VIP\Search\HealthJob( $es );`
5) Enable healthchecks: `add_filter( 'enable_vip_search_healthchecks', '__return_true' );`
6) Set a lock manually: `$hj->set_healthcheck_lock();`
7) Check that `$hj->is_healthcheck_ongoing()` returns true
8) Attempt to run `$hj->check_health()` and check that it doesn't call `$hj->check_document_count_health()`.

### DIFF
--- a/search/includes/classes/class-healthjob.php
+++ b/search/includes/classes/class-healthjob.php
@@ -38,6 +38,16 @@ class HealthJob {
 	 */
 	const AUTOHEALED_ALERT_THRESHOLD = 0.1;
 
+	/**
+	 * @var string The lock name for healthchecks.
+	 */
+	const HEALTHCHECK_LOCK_NAME = 'vip_search_healthcheck_lock';
+
+	/**
+	 * @var int The timeout for the lock.
+	 */
+	const HEALTHCHECK_LOCK_TIMEOUT = 5 * MINUTE_IN_SECONDS;
+
 	public $health_check_disabled_sites = array();
 
 	/**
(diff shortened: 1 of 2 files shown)


search/includes/classes/class-healthjob.php
```diff
@@ -240,6 +252,8 @@ public function check_document_count_health() {
 
 			$this->process_document_count_health_results( $post_results );
 		}
+
+		$this->remove_healthcheck_lock();
```
sjinks, line 256: Does it make sense to use `try…finally` here? For example, if an unexpected exception happens?

sjinks (approved):

search/includes/classes/class-healthjob.php
rebeccahum: @sjinks I figured that IF something were to unexpectedly happen where the lock couldn't be removed, it would eventually expire after 5 minutes.

sjinks: Oh, right, `set_transient()`

rebeccahum:
r2038-stacks

PR #2659, opened 2021-11-24
Refactor ACL tests

sjinks:
Speed up ACL tests by running them in a single process.

* Time before: 13.32 seconds
* Time after: 2.58 seconds


rebeccahum (approved):

sjinks:
r2040-stacks

PR #2660, opened 2021-11-24
Refactor Vary Cache tests

sjinks:
Refactor Vary Cache tests to allow them to run in a single process.

Timings:
* before: 27.38 seconds
* after: 2.61 seconds

### DIFF
--- a/tests/mock-constants.php
+++ b/tests/mock-constants.php
@@ -79,3 +79,15 @@ namespace Automattic\VIP\Files\Acl {
 		return Constant_Mocker::constant( $constant );
 	}
 }
+
+namespace Automattic\VIP\Cache {
+	use Automattic\Test\Constant_Mocker;
+
+	function defined( $constant ) {
+		return Constant_Mocker::defined( $constant );
+	}
+
+	function constant( $constant ) {
+		return Constant_Mocker::constant( $constant );
+	}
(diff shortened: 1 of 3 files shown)


pauarge (approved):

tests/mock-constants.php
```diff
@@ -79,3 +79,15 @@ function constant( $constant ) {
 		return Constant_Mocker::constant( $constant );
 	}
 }
+
+namespace Automattic\VIP\Cache {
```
pschoffer, line 83: Is the reason we need this that the prod code is doing `defined` and `constant` directly? So this will kinda overwrite the php core functions with the namespaced versions?

pschoffer (approved):
Anyways, looks awesome. Thanks for that.

tests/mock-constants.php
sjinks: Yes :-) We inject our helpers into the respective namespace (this is why we have to duplicate the same function for every namespace), and then the production code, due to the namespace resolution rules, calls our implementation instead of the PHP's one.

rebeccahum:
r2043-stacks

sjinks:
Oops, I forgot to deploy, sorry :-( Thank you, @rebeccahum

PR #2681, opened 2021-12-06
Remove Bulk Edit Limitations

mjangda:
We no longer should be restricting these actions since they impact customer workflows.

### Removed Bulk Editing Restrictions

We've removed restrictions around bulk editing

## Checklist

## Steps to Test

1. Check out PR.
2. Make sure you have lots of posts (e.g. 500)
1. Go to `wp-admin` > `Posts` > `Edit`
1. Increase the post count to 100
1. Verify that the bulk edit banner doesn't show up and you can bulk edit all posts


WPprodigy:
In my WC support days, I remember a handful of times somebody set the "per page" limit to something really high in screen options and as a result ends up locked out of that particular edit posts page b/c the query would timeout/OOM/fail.

The only way to get yourself back into the page is to manually edit the particular usermeta I believe to drop back down the screen options value to something reasonable. It's not a great flow 😬 .

All that to say, are we sure we want to fully remove it?

mjangda:
I think we can counter those issues with a limit on the per_page value instead.

WPprodigy:
Looks like WP core already does both frontend and backend validation for the setting and doesn't let it go above 999 now: https://github.com/WordPress/WordPress/blob/ae9eae8f64aa84b965ca1d67e13164826a493b52/wp-admin/includes/misc.php#L658-L689

I did some testing on the simplest of scenarios, and loading 1000 is a bit slow but eventually works, but editing them is a very bad experience that eventually fails. So I do think we should probably still do something, while increasing `BULK_EDIT_LIMIT` by a bit (100 felt good, 150 was alright):

If we want to just limit the screen options, then something like this will give us both backend & frontend validation (so it's clear to the user):

```
add_action( 'wp_loaded', __NAMESPACE__ . '\limit_posts_per_page_for_registered_post_types' );
function limit_posts_per_page_for_registered_post_types() {
	$types = get_post_types( [ 'show_ui' => true ] );

	foreach ( $types as $type ) {
		add_action( "get_user_option_edit_{$type}_per_page", __NAMESPACE__ . '\limit_per_page' );
	}
}

function limit_per_page( $per_page ) {
	return $per_page > BULK_EDIT_LIMIT ? BULK_EDIT_LIMIT : $per_page;
}

add_action( 'all_admin_notices', __NAMESPACE__ . '\frontend_per_page_limit_validation' );
function frontend_per_page_limit_validation() {
	?>
	<script>
		var input = document.querySelector( 'input[type="number"].screen-per-page' );
		if ( null !== input ) {
			input.max = <?php echo BULK_EDIT_LIMIT ?>;
		}
	</script>
	<?php
}
```

The downside here though is that in a way it is more limiting. As right now you can view 999 (which works), just can't bulk edit them. W/ the above approach, you'll be limited to viewing the BULK_EDIT_LIMIT as well. Probably is okay though?

jesse-greathouse:
Mirroring my thoughts from the discussion in chat. I think removing this performance code is fine. I think if the core WP code is inadequate for hedging performance of bulk edits, then it would be better to contribute some help to core. I don't like the idea of mu-plugins trying to pick up the slack for performance gaps because this should not be a "VIP feature", this particular performance enhancements should be available in Wordpress out of the box.

I also think that doing more than 999+ bulk edits should probably be done by a specialized plugin because, on a lot of clients, the performance will never be adequate and the user experience will still be an unreasonable slog.

All that being said, I would like to see someone with more seniority than me offer the actual approval, but Im definitely putting my vote in for yes. 👍

GaryJones (commented):
Could the `BULK_EDIT_LIMIT` constant be removed as well? Or at least commented as not actively being used if there is a concern that a customer codebase is using it?

WPprodigy:
I checked for usage of the const, it's not technically used anywhere. So we can remove it if we do opt for removing everything completely.

I'm quite in favor of safeguarding here though. An editor that is trying to bulk change tags/categories for example may not understand that loading up 999 posts is a bad idea, and the bulk-update failing mid way is not ideal for anyone since you lose data integrity to some degree. Limiting to 100 or 150 seems smart, as it's going to ensure it works and prevents somebody from shooting themselves in the foot.

As for fixing in core, I think it'd be tough to "pull back" on the amount of posts somebody can list since it could break editing flows.

WPprodigy:
Closing in favor of https://github.com/Automattic/vip-go-mu-plugins/pull/2931

PR #2683, opened 2021-12-06
Adds "VIP_AKISMET_SKIP_LOAD" constant to allow disabling of Akismet

jesse-greathouse:
## Description
Akismet will check for the constant: "VIP_AKISMET_SKIP_LOAD"  when it loads.

If the value is set to `true` the plugin will be disabled.

## Changelog Description

The Akismet plugin, in Must-Use plugins, may now be disabled by adding the "VIP_AKISMET_SKIP_LOAD" constant, and setting the value to "true".

## Checklist

## Steps to Test

1. Check out PR.
1. Add a line with: `define( 'VIP_AKISMET_SKIP_LOAD', true );` to wp-config.php or wp-config-multisite.php 
1. Go to `wp-admin` > `Jetpack`
1. The Akismet Anti-Spam sub-menu should be missing

### DIFF
--- a/akismet.php
+++ b/akismet.php
@@ -16,6 +16,11 @@ if ( defined( 'VIP_JETPACK_SKIP_LOAD' ) && true === VIP_JETPACK_SKIP_LOAD ) {
 	return;
 }
 
+// Avoid loading Akismet if VIP_AKISMET_SKIP_LOAD is set to true
+if ( defined( 'VIP_AKISMET_SKIP_LOAD' ) && true === VIP_AKISMET_SKIP_LOAD ) {
+	return;
+}
+
 // Load the core Akismet plugin
 require_once __DIR__ . '/akismet/akismet.php';
 


rebeccahum (approved):

sjinks:
r2054-stacks

WPprodigy:
One other thing to be aware of is this flow: https://github.com/Automattic/vip-go-mu-plugins/blob/2f3c34c59df2148993326998e0de06d959bdd14a/vip-jetpack/connection-pilot/class-jetpack-connection-controls.php#L173

Used in two places:

- Cron: https://github.com/Automattic/vip-go-mu-plugins/blob/8606b21064137b21a3c48a82c6e686509219404d/vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php#L123
- CLI: https://github.com/Automattic/vip-go-mu-plugins/blob/2f3c34c59df2148993326998e0de06d959bdd14a/vip-jetpack/connection-pilot/class-jetpack-connection-cli.php#L51

I'm thinking could probably have the `connect_akismet()` method early return `null` if the constant is set. Then in those two places, check specifically for `=== false`. So the function signature is then: `connect_akismet(): ?bool` I suppose?

rinatkhaziev:
Nice catch @WPprodigy! @jesse-greathouse can we please address this in a follow-up PR?

jesse-greathouse:
> So to avoid false positives in alerting (that akismet connection failed when it wasn't needed), I'm thinking could probably have the `connect_akismet()` method early return `null` if the constant is set. Then in those two places, check specifically for `=== false`. So the function signature is then: `connect_akismet(): ?bool` I suppose?

Thank you for bringing this to my attention @WPprodigy I do agree with the analysis; and I am very grateful that you brought this to my attention. I do think, however, after looking at the code it would be better to go about it in a slightly different way.

Where my opinion differs, as I see it, is that we would be changing the `connect_akismet()` method signature; and changing the method to be more ambiguous in how it works.

```PHP
	/**
	 * Connect a site to Akismet.
	 *
	 * Uses Akismet's function to connect Akismet using the Jetpack. An active Jetpack connection on the site
	 * and the Akismet plugin are required.
	 *
	 * @return bool True if connection worked, false otherwise
	 */
	public static function connect_akismet(): bool {
```
I just think the idea of changing this method to a polymorphic result is going in the wrong direction. I don't think it's a terrible idea, but I think it's unnecessarily making this a bit more ambiguous than it needs to be to accomplish what we're trying to accomplish.

I think a better idea would be to simply wrap the whole call to Controls::connect_akismet in a condition predicated on whether the `VIP_AKISMET_SKIP_LOAD` exists and is true. Something more like this:

```PHP
                if ( ! defined( 'VIP_AKISMET_SKIP_LOAD' ) || true !== VIP_AKISMET_SKIP_LOAD ) {
		    $ak_connection = Controls::connect_akismet();
		    if ( ! $ak_connection ) {
			    WP_CLI::warning( '❌ Could not connect Akismet.' );
		    }
                }
```

```PHP
                if ( ! defined( 'VIP_AKISMET_SKIP_LOAD' ) || true !== VIP_AKISMET_SKIP_LOAD ) {
			$akismet_connection_attempt = Connection_Pilot\Controls::connect_akismet();
			if ( ! $akismet_connection_attempt ) {
				$this->send_alert( 'Alert: Could not connect Akismet automatically.' );
			}
                }
```

This would prevent changing the `connect_akismet()` method signature to being less ambiguous, and also adds one less logical operation. What do you think about this approach?

WPprodigy:
Seems good!

PR #2685, opened 2021-12-07
Speed up WP core tests

sjinks:
Speed up WordPress core tests by using a prebuilt image.

Before:
core-tests-before

After:
core-tests-now


pschoffer (approved):
Oh yeah! that's the stuff!

pschoffer (approved):

PR #2687, opened 2021-12-07
Search: Only ensure index existence for single-sites

rebeccahum:
## Description
We should not be creating new indexes via on `ensure_index_existence()` for multisites. This can result in unused indexes which may lead to resource consumption.

Only ensure index existence for single-sites.

## Checklist

## Steps to Test
1) Create a multisite ES env
2) Create a new subsite
3) Perform a search
4) Check `wp vip-search get-indexes` to ensure that an index has not been created


rebeccahum:
Updating to only use for single-sites because of https://github.com/Automattic/vip-go-mu-plugins/pull/2690.

pschoffer:
> we should not be creating new indexes

I think we might need to have some more complex logic in place. Maybe not create indexes on search? Otherwise you would:
1) create a new subsite  - which will not create an index
2) create a new post - which with this change will not create index, but push it to ES 
3) because there is no index in ES yet it will create a default one with incorrect mapping. The user will have to run `index --setup` to fix that.

But that is maybe ok. maybe instead of creating the index proactively, we should detect that the index is not set up with correct mappings (maybe that EP's index version feature?) and throw instead of attempting to communicate with the index. EP has already an admin notice that shows in that case I think.

rebeccahum:
@pschoffer Thanks for your feedback! I'm hoping you can clarify points 2 and 3:

> 2. create a new post - which with this change will not create index, but push it to ES
> 3. because there is no index in ES yet it will create a default one with incorrect mapping. The user will have to run index --setup to fix that.

What are your thoughts on disabling this functionality? If there is no index, to not push it to ES or create a new one.

rinatkhaziev:
I think it needs to act as a circuit-breaker - no index should mean no SyncManager activity and no health checks

rebeccahum:
@pschoffer I've opened #2695 to kill syncing if the index does not exist. Therefore, since step 2 will not be triggered, step 3 should not be either.

pschoffer:
Yeah that sounds good... although it might be tricky to differentiate between valid requests that we need to let go through even if the index doesn't exist. Like `create index` or query if `index exists`.

> What are your thoughts on disabling this functionality? If there is no index, do not push it to ES or create a new one.
Yeah, that would be good of course. The issue is that there are so many places to change in EP in order to achieve something like that (I assume). The only central place we have right now at least is that request handler where your change is now. But maybe we would be able to whitelist direct index operations there... and suppress everything else.

pschoffer:
Oh cool maybe that filter would sort out everything... I'll try to look at it some more. But as long as index creation gets through and nothing else that would create index implicitly then that's great. Then we can drop this whole ensure index thingie.

rebeccahum:
The ensure index functionality could prove to be useful for single sites, no?

rebeccahum:
@pschoffer I'll revise this PR to get rid of ensuring index existence if that PR is viable.

rebeccahum:
@pschoffer I've opened to not kill syncing attempts if the index does not exist.

PR #2689, opened 2021-12-07
Make use_ambiguous_login_error() more tolerant to broken user code

sjinks:
## Description

One of the client sites added an incorrect implementation of the `login_errors` filter, which under certain circumstances returned `null` instead of a string. This PR makes `use_ambiguous_login_error()` (which is our `login_errors` hook) more tolerant to such cases.

### Plugin Updated: VIP Security

`use_ambiguous_login_error()` is now more tolerant to broken user code.

## Checklist


security/login-error.php
```diff
@@ -10,11 +10,11 @@
  * 
  * @since 1.1
  */
-function use_ambiguous_login_error( string $error ): string {
+function use_ambiguous_login_error( $error ): string {
 	global $errors;
 
 	if ( ! is_wp_error( $errors ) ) {
-		return $error;
+		return (string) $error;
```
sjinks, line 17: If we are passed something other than a string, limit possible damage by explicitly casting the argument to a string.

rebeccahum: @sjinks Is this something we want to expose though since it gets directly outputted?

rebeccahum (approved):

sjinks:
r2059-stacks

jesse-greathouse:
Release v20211207.1

PR #2691, opened 2021-12-08
Increase the logstash limit to 100 in web and 500 in CLI contexts and set 'http_user_agent' based on _SERVER['HTTP_USER_AGENT']

rinatkhaziev:
## Description

Currently, we have a cut-off of 30 logstash entries per request, as a failsafe. The problem is we call trigger_error when we're over the threshold. This leads to an unhelpful situation where a useful log entry gets replaced by an unuseful generic error not providing any information beyond the fact that it happened.

This PR increases the limit to 100 for web requests and 500 for CLI requests - this may change pending feedback.
When a threshold is crossed only one E_USER_WARNING is triggered.

Additionally, this PR also surfaces user agent for request (if available) 

### Library updated: Logger

We've updated the internal logging library to surface the user agent information for custom log entries and to increase the maximum number of logs to be processed during a single request.

## Checklist

### DIFF
--- a/logstash/class-logger.php
+++ b/logstash/class-logger.php
@@ -300,8 +300,8 @@ class Logger {
 			'user_id'         => get_current_user_id(),                       // Optional.
 			'extra'           => [],                                          // Optional.
 			'timestamp'       => gmdate( 'Y-m-d H:i:s' ),                     // Required.
-			'index'           => 'log2logstash',                              // Required,
-			'http_user_agent' => self::get_user_agent(),
+			'index'           => 'log2logstash',                              // Required.
+			'http_user_agent' => self::get_user_agent(),                      // Optional.
 		];
 
 		if ( ! isset( $params['file'] ) && ! isset( $params['line'] ) ) {
--- a/logstash/class-logger.php
+++ b/logstash/class-logger.php
@@ -33,7 +33,7 @@ class Logger {
 	 *
 	 * @var int
 	 */
-	protected const MAX_ENTRIES = 30;
+	protected const MAX_ENTRIES = 100;
 
 	/**
 	 * Maximum entries to send in one API request.
@@ -302,7 +302,8 @@ class Logger {
 
 			'extra'     => [],                                        // Optional.
 			'timestamp' => gmdate( 'Y-m-d H:i:s' ),                   // Required.
-			'index'     => 'log2logstash',                            // Required
+			'index'     => 'log2logstash',                            // Required,
+			'user_ua'   => self::get_user_agent(),
 		];
 
 		if ( ! isset( $params['file'] ) && ! isset( $params['line'] ) ) {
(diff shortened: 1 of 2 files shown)


logstash/class-logger.php
```diff
@@ -535,4 +540,28 @@ public static function wp_debug_log( array $entry ) : void {
...
+	 */
+	public static function get_user_agent() {
+		if ( defined( 'WP_CLI' ) && WP_CLI ) {
+			return 'cli';
+		} else {
+			// Match handling in other places
+			// phpcs:ignore WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___SERVER__HTTP_USER_AGENT__
+			return substr( sanitize_text_field( $_SERVER['HTTP_USER_AGENT'] ?? '' ), 0, 255 );
+		}
+	}
```
WPprodigy, line 566: Small nitpick, but removes the `else` :)

```suggestion
	public static function get_user_agent() {
		if ( defined( 'WP_CLI' ) && WP_CLI ) {
			return 'cli';
		}
		
		// Match handling in other places
		// phpcs:ignore WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___SERVER__HTTP_USER_AGENT__
		return substr( sanitize_text_field( $_SERVER['HTTP_USER_AGENT'] ?? '' ), 0, 255 );
	}
```

WPprodigy (commented):
So the 101st entry will fail and log the warning. But the 100 prev ones will still be logged right?

rinatkhaziev:
> So the 101st entry will fail and log the warning. But the 100 prev ones will still be logged right?

Yep, that logic doesn't change, what's changing is the limit and not logging the useless message over and over, instead only log once.

logstash/class-logger.php
```diff
@@ -290,19 +291,17 @@ public function debug( string $feature, string $message, array $context = [] ) :
...
+			'blog_id'         => get_current_blog_id(),                       // Required.
+			'http_host'       => strtolower( $_SERVER['HTTP_HOST'] ?? '' ),   // phpcs:ignore -- Optional.
+			'severity'        => '',                                          // Optional.
+			'feature'         => '',                                          // Required.
+			'message'         => '',                                          // Required.
+			'user_id'         => get_current_user_id(),                       // Optional.
+			'extra'           => [],                                          // Optional.
+			'timestamp'       => gmdate( 'Y-m-d H:i:s' ),                     // Required.
+			'index'           => 'log2logstash',                              // Required,
+			'http_user_agent' => self::get_user_agent(),
```
WPprodigy, line 304: Just to keep the pattern of the rest (assuming it is optional):

```suggestion
			'timestamp'       => gmdate( 'Y-m-d H:i:s' ),                     // Required.
			'index'           => 'log2logstash',                              // Required.
			'http_user_agent' => self::get_user_agent(),                      // Optional.
```

WPprodigy (approved):
Small commenting change, otherwise looks good

WPprodigy (approved):

rinatkhaziev:
r2063-stacks

PR #2692, opened 2021-12-08
Add caching to the months dropdown filter for posts

WPprodigy:
## Description
Follow-up to https://github.com/Automattic/vip-go-mu-plugins/pull/1883.

tl;dr, the query to generate the year/month filter dropdown on posts pages can be a bit expensive on large sites and is never cached. There is a new filter in WP core that will allow us to intercept and cache the results: https://core.trac.wordpress.org/ticket/51660

### Performance: Cache the year/month dropdown on admin post pages

To prevent an expensive query occurring on every admin posts-edit page load, this caches the available months/years for the dropdown filter.

## Checklist

## Steps to Test

1. Go to a listing page e.g.: `/wp-admin/edit.php` or a custom post type listing page (`/wp-admin/edit.php?post_type=myCustomPostType`) or media library (`/wp-admin/upload.php`)
2. When the page is first loaded, the query should take place and the cache is populated.
3. Reload the listing page and note the query does not need to re-occur.
4. Bust the cache by added a post that belongs to a new month/year: `wp post generate --count=1 --post_type=post --post_date=2021-08-25`. Reload and check that the new month/year appears in the dropdown filter.

### DIFF
--- a/performance/vip-tweaks.php
+++ b/performance/vip-tweaks.php
@@ -3,9 +3,10 @@
 /**
  * This is a list of performance tweaks that will become default for All VIP sites
  */
+add_action( 'after_setup_theme', 'wpcom_vip_enable_performance_tweaks' );
 function wpcom_vip_enable_performance_tweaks() {
 
-	// This disables the adjacent_post links in the header that are almost never beneficial and are very slow to compute.
+	// Disables the adjacent_post links in the header that are almost never beneficial and are very slow to compute.
 	remove_action( 'wp_head', 'adjacent_posts_rel_link_wp_head', 10, 0 );
 
 	if ( function_exists( 'wpcom_vip_enable_old_slug_redirect_caching' ) ) {
@@ -15,8 +16,20 @@ function wpcom_vip_enable_performance_tweaks() {
 	if ( function_exists( 'wpcom_vip_enable_maybe_skip_old_slug_redirect' ) ) {
 		wpcom_vip_enable_maybe_skip_old_slug_redirect();
 	}
+
+	if ( is_admin() ) {
+		// Prevent expensive media library queries, and cache the available months for filtering.
+		add_filter( 'media_library_show_video_playlist', '__return_true' );
+		add_filter( 'media_library_show_audio_playlist', '__return_true' );
+		add_filter( 'media_library_months_with_files', 'wpcom_vip_media_library_months_with_files' );
+
+		// Cache the available months for filtering on posts/CPTs.
+		add_filter( 'pre_months_dropdown_query', 'wpcom_vip_available_post_listing_months', 10, 2 );
+	}
+
+	// Busts the month's filtering caches when needed for both attachments and posts/CPTs.
+	add_action( 'save_post', 'wpcom_vip_maybe_bust_available_months_cache', 10, 2 );
 }
-add_action( 'after_setup_theme', 'wpcom_vip_enable_performance_tweaks' );
 
 /**
  * Use this function to disable the loading of performance tweaks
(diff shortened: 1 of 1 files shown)


WPprodigy:
And to summarize the previous PR's feedback / what I plan to look into:

1) Switch to wp_cache_* API instead of transients.
2) Look into new cache invalidation strategy (only invalidate when really needed)
3) Perhaps un-duplicate the logic that exists in `wpcom_vip_media_library_months_with_files()` already.

WPprodigy:
Made some adjustments:

- Both media library and posts dropdown are handled the same way.
- They cache the results for 24hours, and only invalidate when a new post w/ a "fresh" month is added.
- Things got messy quickly when dealing w/ filtered post statuses, so opted for just avoiding the cache entirely if somebody selects the "trash" post status filter for example.

performance/vip-tweaks.php
```diff
@@ -25,50 +39,118 @@ function wpcom_vip_disable_performance_tweaks() {
...
+	foreach ( $existing_months as $date ) {
+		if ( $post_year === $date->year && $post_month === $date->month ) {
+			$found = true;
 		}
+	}
 
-		return $months;
+	if ( ! $found ) {
+		// The month/year doesn't exist yet, so just bust the cache and let it re-generate.
+		wp_cache_delete( $cache_key, 'vip' );
```
WPprodigy, line 154: Could technically just add the new month to the cache directly, but I figured busting is okay here since we are already leaning on the 24hour TTL for "cleanup". It can't be that often that an entirely new month is added under normal editorial conditions, and even if somehow lots of new months are added (like an import or something) then the cache will just be busted the first time and not each import. So editorial flow wouldn't be interrupted or notice anything.

performance/vip-tweaks.php
```diff
@@ -25,50 +39,118 @@ function wpcom_vip_disable_performance_tweaks() {
...
+		return false;
+	}
+
+	$post_year  = get_the_time( 'Y', $post_id );
+	$post_month = get_the_time( 'n', $post_id );
+
+	$found = false;
+	foreach ( $existing_months as $date ) {
+		if ( $post_year === $date->year && $post_month === $date->month ) {
+			$found = true;
```
jesse-greathouse, line 148: I think there should be a `break;` here. I see no reason for the the loop to continue after a match is found. Do you?

jesse-greathouse (changes requested):
Looks good. I did find a small suggestion that I think may possibly help.

performance/vip-tweaks.php
WPprodigy: Nice, good find.

performance/vip-tweaks.php
```diff
@@ -15,8 +16,21 @@ function wpcom_vip_enable_performance_tweaks() {
 	if ( function_exists( 'wpcom_vip_enable_maybe_skip_old_slug_redirect' ) ) {
 		wpcom_vip_enable_maybe_skip_old_slug_redirect();
 	}
+
+	if ( is_admin() ) {
+		// Prevent expensive media library queries, and cache the available months for filtering.
+		add_filter( 'media_library_show_video_playlist', '__return_true' );
```
rebeccahum, line 22: Do we still need to force return true if the default value is already true? https://github.com/WordPress/WordPress/blob/39bff93b6b274a873158a362725b1338ab1400e5/wp-includes/media.php#L4323

performance/vip-tweaks.php
```diff
@@ -15,8 +16,21 @@ function wpcom_vip_enable_performance_tweaks() {
...
+		add_filter( 'media_library_show_audio_playlist', '__return_true' );
```
rebeccahum, line 23: Same as above ^^.

performance/vip-tweaks.php
```diff
@@ -25,50 +39,118 @@ function wpcom_vip_disable_performance_tweaks() {
...
-					FROM $wpdb->posts
-					WHERE post_type = %s
-					ORDER BY post_date DESC
-				", 'attachment' ) );
-			set_transient( 'wpcom_media_months_array', $months );
+/**
+ * Helper method to bust the "available months" cache when needed.
+ *
+ * Note that we only bust the cache when a new month is added,
+ * For deletions, we just let the 24-hour TTL handle that.
```
rebeccahum, line 127: Why did you choose this approach over having no cache expiry and invalidating cache on post deletion?

rebeccahum, line 127: Why did you choose this approach over having no cache expiry and invaliding on post deletion?

rebeccahum (changes requested):

performance/vip-tweaks.php
WPprodigy: Nice find, looks like we can remove these now since WP 4.8 made it the default. These filters predate this PR, I just moved them up a few lines.

performance/vip-tweaks.php
WPprodigy: Most post deletions are quite unlikely to result in any change to the "available months". It would have to be the last post within that month to result in a change. Worst case is that the "last post" is deleted and somebody filters by a month within the same day and finds no posts there, but truly unlikely. Vs the high likelihood of deletes being fairly often and resulting in cache busts.

rebeccahum (approved):

rebeccahum:
v20211215.0

PR #2690, opened 2021-12-08
Search: Remove hook to automatically create index for new blogs on multisites

rebeccahum:
## Description
We want to allow the constant `EP_IS_NETWORK` to be enabled (for cross-blog searching) without automatically creating indexes everytime a subsite is created on a multisite.

### Plugin Updated: Enterprise Search

Remove hook to automatically create index for new blogs.

## Checklist

Please make sure the items below have been covered before requesting a review:

- [x] This change works and has been tested locally (or has an appropriate fallback).
- [ ] This change works and has been tested on a Go sandbox.
- [ ] This change has relevant unit tests (if applicable).
- [ ] This change has relevant documentation additions / updates (if applicable).
- [x] I've created a changelog description that aligns with the provided examples.

## Steps to Test
0) Check out PR
1) Create multisite with ES and `EP_IS_NETWORK` defined
2) Create new subsite
3) Run `wp vip-search get-indexes`
4) Ensure index has not been created for it.

PR #2696, opened 2021-12-08
Search: Add hard limit on command for setting up multiple indexes at …

rebeccahum:
## Description
Since multisites will no longer automatically create indexes on subsite creation per #2690, we should add a hard limit on the CLI command to prevent sites with over 100 blogs automatically creating indexes via the `--network-wide` flag. This will ensure that indexes are purposefully being created after a certain amount.

Added a `--blog-ids` parameter to pass in as an alternative as well.

Ensure multisites with over 100 subsites cannot automatically create indexes by CLI.

## Checklist

## Steps to Test
To test hard limit:
1) Create multisite with over 100 blogs
2) Attempt to set up indexes all at once via `wp vip-search index --setup --network-wide`
3) Expect it to halt with error

To test new blog-ids flag:
1) Attempt to index 2-3 blogs via `wp vip-search index --setup --blog-ids=1,2,3`


WPprodigy:
Hmm, I do wonder if this is just causing tedious work down the line though if there ever is a large MS that does need each site to have an index.

rebeccahum:
What about adding a flag to override it then? Similar to what we have for the `--force_parallel_execution`. This is just an initial stop-gap to ensure that `--network-wide` is being run purposely for sites with blogs #s > 100 . @rinatkhaziev what say you?

rinatkhaziev:
@WPprodigy @rebeccahum what do you all think about `--site-ids=`? like `--site-ids=3,1,33,7`

I don't think our goal is to make it cumbersome but rather it's "Check yourself before you wreck yourself"

rebeccahum:
I think that's a great idea, having that alternative flag to mass-index but not mass-index all once a threshold has been reached. Will work on updating the command to reflect that.

rebeccahum:
Updated the command per @rinatkhaziev's idea with `--blog-ids` flag. Will need to update docs accordingly after this is merged.

sjinks (approved):

rebeccahum:
r2079-stacks

sjinks:
v20211214.0

PR #2697, opened 2021-12-08
Wrapped Connection_Pilot\Controls::connect_akismet() in conditional

jesse-greathouse:
## Description
Added `VIP_AKISMET_SKIP_LOAD` condition to Connection_Pilot\Controls::connect_akismet() calls

## Changelog Description

Setting the "VIP_AKISMET_SKIP_LOAD" constant to true will prevent calls made to connect Akismet during the Jetpack Connect procedures in Cron and CLI.

## Checklist

## Steps to Test
1. Check out PR.
1. Set `VIP_AKISMET_SKIP_LOAD` constant to true in vip-config.php: `define( 'VIP_AKISMET_SKIP_LOAD', true );`
1. Run the Jetpack connect cli command: `wp jetpack-start connect`
1. Find the Jetpack connect cron entry ID: `wp cron-control events list`
1. Run the Jetpack connect cron event:  `wp cron-control events run [event id]`


WPprodigy (dismissed):
Thanks!

vip-jetpack/connection-pilot/class-jetpack-connection-cli.php
```diff
@@ -48,9 +48,11 @@ private function connect_site() {
 			return false;
 		}
 
-		$ak_connection = Controls::connect_akismet();
-		if ( ! $ak_connection ) {
-			WP_CLI::warning( '❌ Could not connect Akismet.' );
+		if ( ! defined( 'VIP_AKISMET_SKIP_LOAD' ) || true !== VIP_AKISMET_SKIP_LOAD ) {
+			$ak_connection = Controls::connect_akismet();
```
sathyapulse, line 52: The variable `$ak_connection` might need to be defined with default `false`. It's used [here](https://github.com/Automattic/vip-go-mu-plugins/pull/2697/files#diff-3ba288381cb921454bae71e6735b3867e0eb585f41e9efe9aed15dd3015569abR63).

vip-jetpack/connection-pilot/class-jetpack-connection-cli.php
```diff
@@ -48,9 +48,11 @@ private function connect_site() {
...
+			if ( ! $ak_connection ) {
+				WP_CLI::warning( '❌ Could not connect Akismet.' );
+			}
 		}
 
 		$vp_connection = Controls::connect_vaultpress();
```
sathyapulse, line 58: We also need to check the existence of VaultPress using the constant `VIP_VAULTPRESS_SKIP_LOAD`.

vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
```diff
@@ -120,9 +120,11 @@ public function run_connection_pilot() {
 			$this->update_heartbeat();
 
 			// Attempting Akismet connection given that Jetpack is connected
-			$akismet_connection_attempt = Connection_Pilot\Controls::connect_akismet();
-			if ( ! $akismet_connection_attempt ) {
-				$this->send_alert( 'Alert: Could not connect Akismet automatically.' );
+			if ( ! defined( 'VIP_AKISMET_SKIP_LOAD' ) || true !== VIP_AKISMET_SKIP_LOAD ) {
```
pauarge, line 123: I would mirror the way the conditional is written with VaultPress (variable plus a &&). I think its easier to read and we maintain consistency.

pauarge (changes requested):
Good catch, left a small cosmetic change

vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
jesse-greathouse: Thank you for this! I have done as you requested in the most recent commit.

vip-jetpack/connection-pilot/class-jetpack-connection-cli.php
jesse-greathouse: Thank you very much for catching this. I have refactored this in the latest commit so that it will not try to evaluate against an undefined variable.

vip-jetpack/connection-pilot/class-jetpack-connection-cli.php
jesse-greathouse: Thank you, again for this wonderful catch. I have made the necessary adjustments for vaultpress skip in my latest commit.

WPprodigy (approved):
Nice, this does feel cleaner.

One thought I had is that the method returns are a bit inconsistent. `connect_jetpack()` returns `false` if it skipped the attempt, while vp/ak ones return `true` if they were skipped (making the docblocks not totally true).

It's definitely fine tbh, since it's needed for the control flow, but just mentioning in case you have any ideas for a cleaner flow there.

jesse-greathouse:
> Nice, this does feel cleaner.
> 
> One thought I had is that the method returns are a bit inconsistent. `connect_jetpack()` returns `false` if it skipped the attempt, while vp/ak ones return `true` if they were skipped (making the docblocks not totally true).
> 
> It's definitely fine tbh, since it's needed for the control flow and usage is quite limited, but just mentioning in case you have any ideas for a cleaner flow there.

Very good point. Im glad you noticed that too, because thats something Im dissatisfied with also. I'll let that ping pong around my head a bit and see if I can come up with something.

sathyapulse (approved):
Thank you for working on the changes. It looks good to me.

pauarge (approved):
Great work Jesse!

rebeccahum:
r2075-stacks

sjinks:
v20211214.0

PR #2698, opened 2021-12-08
Reduce the maximum amount of jobs to 3 to reduce indexing pressure

rinatkhaziev:
## Description

Reduce the maximum amount of jobs to 3 to reduce indexing pressure.

Don't log the fact the processor job couldn't be scheduled as it's correct behavior in case the concurrency limit is hit. 

We've reduced the maximum number of concurrent jobs to prevent indexing operations potentially affecting search performance. 

## Checklist

### DIFF
--- a/search/includes/classes/queue/class-cron.php
+++ b/search/includes/classes/queue/class-cron.php
@@ -46,7 +46,7 @@ class Cron {
 	/**
 	 * The maximum number if processor jobs allowed at one time
 	 */
-	const MAX_PROCESSOR_JOB_COUNT = 5;
+	const MAX_PROCESSOR_JOB_COUNT = 3;
 
 	/**
 	 * Instance of Automattic\VIP\Search\Queue that created this Cron instance
@@ -194,6 +194,7 @@ class Cron {
 		// WP_Query args for looking up posts that match the term taxonomy id and indexable
 		// post types/statuses
 		$args = array(
+			// phpcs:ignored WordPress.WP.PostsPerPage.posts_per_page_posts_per_page
 			'posts_per_page'         => self::TERM_UPDATE_BATCH_SIZE,
 			'post_type'              => $indexable_post_types,
 			'post_status'            => $indexable_post_statuses,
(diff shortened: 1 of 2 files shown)


tests/search/includes/classes/queue/test-class-cron.php
```diff
@@ -252,11 +252,11 @@ public function configure_concurrency_data() {
 			],
 			[
 				20,
-				[ 'vip_search_queue_processor' => 5 ],
+				[ 'vip_search_queue_processor' => 3 ],
 			],
 			[ // max of 5 takes over
```
rebeccahum, line 257: Minor nitpick, can we update the comment to reflect the change of 3?

search/includes/classes/queue/class-cron.php
```diff
@@ -46,7 +46,7 @@ class Cron {
 	/**
 	 * The maximum number if processor jobs allowed at one time
 	 */
-	const MAX_PROCESSOR_JOB_COUNT = 5;
+	const MAX_PROCESSOR_JOB_COUNT = 3;
```
rebeccahum, line 49: Just wondering, but will this have much of an impact of it evaluates to 2.5 usually? https://github.com/Automattic/vip-go-mu-plugins/blob/90ae5a0b46c0231fce4384075f314a5cb20f7613/search/includes/classes/queue/class-cron.php#L86-L89

rinatkhaziev: I spot-checked some heavy hitters, they have much bigger concurrency, which kind of adds up to the problem.

tests/search/includes/classes/queue/test-class-cron.php
rinatkhaziev: thanks for the catch, addressed!

rebeccahum (changes requested):
Thanks for the changes. Just one more thing, can you update the test here to reflect the new value:

https://github.com/Automattic/vip-go-mu-plugins/blob/90ae5a0b46c0231fce4384075f314a5cb20f7613/tests/search/includes/classes/queue/test-class-cron.php#L225-L226

rinatkhaziev:
Sure thing, funny nothing failed though.

rebeccahum (dismissed):
thxu

rebeccahum (approved):
gg

PR #2703, opened 2021-12-10
Implement concurrency limit for ES

sjinks:
## Description

This PR implements a concurrency limit for ES.

Implemented concurrency limit to avoid overloading of the ES server.

## Checklist

## Steps to Test

The CI should pass.


search/includes/classes/class-search.php
```diff
@@ -772,6 +772,10 @@ public function filter__ep_bulk_items_per_page() {
 	 * @return array  $request  New request
 	 */
 	public function get_cached_index_exists_request( $request, $query, $args, $failures = 0, $type = null ) {
+		if ( is_wp_error( $request ) && 503 === $request->get_error_code() ) {
+			return $request;
+		}
```
sjinks, line 777: If we hit the concurrency limit, we should not make a request to ES.

bin/test.sh
```diff
@@ -57,6 +57,8 @@ done
 : "${PHPUNIT_VERSION:=""}"
 : "${DOCKER_OPTIONS:=""}"
 
+PHP_OPTIONS="-d apc.enable_cli=1 ${PHP_OPTIONS}"
```
sjinks, line 60: APCu is not enabled in the CLI mode by default. We cannot move this to `phpunit.xml.dist` because `ap.enable_cli` is `PHP_INI_SYSTEM`.

tests/search/includes/classes/class-concurrency-limiter-helper.php
```diff
@@ -0,0 +1,13 @@
...
+namespace Automattic\VIP\Search;
+
+require_once __DIR__ . '/../../../../search/includes/classes/class-concurrency-limiter.php';
+
+// phpcs:disable Generic.CodeAnalysis.UselessOverridingMethod.Found
+
+class Concurrency_Limiter_Helper extends Concurrency_Limiter {
+	public function get_key(): int /* NOSONAR */ {
+		return parent::get_key();
+	}
```
sjinks, line 12: My personal preference is not to use Reflection.

search/includes/classes/class-concurrency-limiter.php
```diff
@@ -0,0 +1,103 @@
...
+use WP_Error;
+
+class Concurrency_Limiter {
+	const KEY_NAME = 'vip_es_request_count';
+
+	/** @var int */
+	private $cache_ttl = 60;
+
+	/** @var int */
+	private $max_concurrent_requests = 100;
```
sjinks, line 14: These values are to be discussed.

sjinks:
To reviewers: this PR is mostly a PoC: it will be refined and improved by subsequent pull requests. For example, we will not be limited with APCu, and there is #2706 that will address this.

This PR is mostly about the approach to concurrency limiting.

bin/test.sh
pschoffer: What does this do?

search/includes/classes/class-concurrency-limiter.php
pschoffer: This seems kinda low... I wonder how to come up with optimal value.. Do we know how many concurent request clients do currently?

tests/search/includes/classes/class-concurrency-limiter-helper.php
pschoffer: What does that imply for us non-php native folks? :)

pschoffer (changes requested):
More filling my knowledge gaps rather than change requests.. It feels like you will be able to give me better answers specifically to our problem than duckduckgo (I am starting a new trend here:)) ever will... Looks good overall.

tests/search/includes/classes/class-concurrency-limiter-helper.php
sjinks: Nothing :-) I will kill this class in #2706 anyway, so please don't pay attention :-)

https://www.cprime.com/resources/blog/unit-testing-smells/

search/includes/classes/class-concurrency-limiter.php
sjinks: Unfortunately, I have no idea. I am open to changes here :-)

And, for corner cases, we provide a filter to adjust this value.

bin/test.sh
sjinks: It prepends `-d apc.enable_cli=1` to the `PHP_OPTIONS` variable. It is then [passed to Docker](https://github.com/Automattic/vip-go-mu-plugins/blob/57636750224f1c3732076f07b29601dd4226f41d/bin/test.sh#L116). The entrypoint script [passes these options](https://github.com/Automattic/vip-container-images/blob/master/wp-test-runner/runner.sh#L13) to PHP.

We will end up with

```bash
php -d apc.enable_cli=1 other parameters go here
```

The `-d` option [defines a php.ini value](https://www.php.net/manual/en/features.commandline.options.php).

In this case, we [enable APCu in the CLI](https://www.php.net/manual/en/apcu.configuration.php#ini.apcu.enable-cli).

search/includes/classes/class-concurrency-limiter.php
```diff
@@ -0,0 +1,103 @@
...
+			return $value;
+		}
+
+		return PHP_INT_MAX;
+	}
+
+	protected function dec_key(): void {
+		$success = null;
+		apcu_dec( self::KEY_NAME, 1, $success, $this->cache_ttl );
+	}
```
sjinks, line 102: The APCu stuff will be replaced with configurable backends in #2706.

search/includes/classes/class-concurrency-limiter.php
sjinks: PS: this is a per-pod limit.

WPprodigy: Since it's per-pod, and PHP is basically always synchronous, I'm not sure how much benefit there will be here since a pod already has a concurrency limit built-in with the "php workers" limit. This is a pretty standard number across the board, except in rare cases of manual vertical scaling rather than horizontal.

rinatkhaziev: >  Do we know how many concurent request clients do currently?

I don't think we have a good understanding of this currently. I think the first step would be to gauge that value by disabling the limiting functionality and instead just log the fact. We can keep it running for a bit to gather a better understanding.

rebeccahum:
Should we have some sort of logging when concurrency has been reached?

search/includes/classes/class-search.php
```diff
@@ -840,6 +852,10 @@ private function get_index_exists_option_name( $url ) {
 	 * @return array  $request  New request
 	 */
 	public function filter__ep_do_intercept_request( $request, $query, $args, $failures = 0, $type = null ) {
+		if ( is_wp_error( $request ) && 503 === $request->get_error_code() ) {
```
rebeccahum, line 855: What are your thoughts on putting this logic in a later filter to run after the other ones hooked onto `ep_do_intercept_request` instead of modifying `filter__ep_do_intercept_request` and `get_cached_index_exists_request`?

search/includes/classes/class-search.php
```diff
@@ -772,6 +780,10 @@ public function filter__ep_bulk_items_per_page() {
 	 * @return array  $request  New request
 	 */
 	public function get_cached_index_exists_request( $request, $query, $args, $failures = 0, $type = null ) {
+		if ( is_wp_error( $request ) && 503 === $request->get_error_code() ) {
```
rebeccahum, line 783: Same as above.

search/includes/classes/class-search.php
sjinks: The idea is that when we reach the concurrency limit, we should not make a request to ES.

Whether we have reached the limit is detected by another `ep_do_intercept_request` filter with a higher priority (0). Now, if we need to stop the request, there is no other way for us (or maybe I miss one?) to stop *this* filter, because it unconditionally calls `vip_safe_wp_remote_request()` (this is what we don't need it to do).

search/includes/classes/class-concurrency-limiter.php
sjinks: I have created another backend (in #2706) that will use Memcached, and then this limit will be per-client.

For the per-pod limit, I would like to confirm if we are not using php-fpm's dynamic process scheduler (it spawns new PHP workers on demand and kills them when they are not in use).

search/includes/classes/class-search.php
rebeccahum: I'm not a fan of putting this logic into each and every hook onto `ep_do_intercept_request` (especially if we do decide to add more on later). But I understand your point. 

Would it be possible to have a filter that runs the check and then, short-circuit it before even adding the other filters?

search/includes/classes/class-search.php
rebeccahum: What if the request is cached?

search/includes/classes/class-search.php
rebeccahum: I'm not a fan of replicating this logic into each and every hook onto ep_do_intercept_request. But I understand your point. I'm going to combine these filters into one.

rebeccahum: @sjinks I've opened https://github.com/Automattic/vip-go-mu-plugins/pull/2712.

search/includes/classes/class-concurrency-limiter.php
sjinks: I have updated #2706 to use Object Cache (= Memcached). This makes the limit global (per-client instead of per-pod).

Also, that PR changes the behavior to log warnings for excessive requests instead of dropping them.

search/includes/classes/concurrency-limiter/class-apcu-backend.php
```diff
@@ -0,0 +1,69 @@
+<?php
```
sjinks, line 1: Please disregard this file

search/includes/classes/concurrency-limiter/class-semaphore-backend.php
```diff
@@ -0,0 +1,65 @@
+<?php
```
sjinks, line 1: Please disregard this file

search/includes/classes/class-concurrency-limiter.php
```diff
@@ -0,0 +1,121 @@
...
+
+	public function init(): bool {
+		$backend_class = Object_Cache_Backend::class;
+		$backend_class = apply_filters( 'vip_es_concurrency_limit_backend', $backend_class );
+
+		if ( ! empty( $backend_class ) && is_subclass_of( $backend_class, BackendInterface::class, true ) ) {
+			/** @psalm-var class-string<BackendInterface> $backend_class */
+			if ( $backend_class::is_supported() ) {
+				$this->max_concurrent_requests = (int) apply_filters( 'vip_es_max_concurrent_requests', $this->max_concurrent_requests );
```
rinatkhaziev, line 40:
```suggestion
				$this->max_concurrent_requests = (int) apply_filters( 'vip_search_max_concurrent_requests', $this->max_concurrent_requests );
```

search/includes/classes/class-concurrency-limiter.php
```diff
@@ -0,0 +1,121 @@
...
+				$this->cache_ttl               = (int) apply_filters( 'vip_es_cache_ttl', $this->cache_ttl );
+
+				/** @var BackendInterface */
+				$this->backend = new $backend_class();
+				$this->backend->initialize( $this->max_concurrent_requests, $this->cache_ttl );
+
+				add_filter( 'ep_do_intercept_request', [ $this, 'ep_do_intercept_request' ], 0 );
+				add_action( 'ep_remote_request', [ $this, 'ep_remote_request' ] );
+				// We will remove this one once we have enough stats
+				apply_filters( 'vip_es_should_fail_excessive_request', [ $this, 'vip_es_should_fail_excessive_request' ] );
```
rinatkhaziev, line 50:
```suggestion
				apply_filters( 'vip_es_should_fail_excessive_request', [ $this, 'vip_search_should_fail_excessive_request' ] );
```

search/includes/classes/class-concurrency-limiter.php
```diff
@@ -0,0 +1,121 @@
...
+	 * @return mixed 
+	 */
+	public function ep_do_intercept_request( $response ) {
+		// This filter can be called inside a loop; we need to make sure not to increment the counter more than once
+		if ( ! $this->doing_request ) {
+			$this->doing_request = true;
+			$this->should_fail   = ! $this->backend->inc_value();
+		}
+
+		$fail = apply_filters( 'vip_es_should_fail_excessive_request', $this->should_fail );
```
rinatkhaziev, line 86:
```suggestion
		$fail = apply_filters( 'vip_search_should_fail_excessive_request', $this->should_fail );
```

search/includes/classes/class-concurrency-limiter.php
```diff
@@ -0,0 +1,121 @@
...
+	 * Currently, we allow all requests to ES while we are gathering stats.
+	 * 
+	 * @param bool $should_fail 
+	 * @return bool 
+	 */
+	public function vip_es_should_fail_excessive_request( bool $should_fail ): bool {
+		if ( $should_fail ) {
+			log2logstash( [
+				'severity' => 'warning',
+				'feature'  => 'es_concurrency_limiter',
```
rinatkhaziev, line 112:
```suggestion
				'feature'  => 'search_concurrency_limiter',
```

tests/search/includes/classes/test-concurrency-limiter.php
```diff
@@ -0,0 +1,101 @@
...
+require_once __DIR__ . '/../../../../search/elasticpress/includes/utils.php';
+
+class Test_Concurrency_Limiter extends WP_UnitTestCase {
+	public function setUp(): void {
+		parent::setUp();
+
+		remove_all_filters( 'ep_do_intercept_request' );
+		remove_all_actions( 'ep_remote_request' );
+
+		remove_all_actions( 'vip_es_should_fail_excessive_request' );
```
rinatkhaziev, line 25:
```suggestion
		remove_all_actions( 'vip_search_should_fail_excessive_request' );
```

tests/search/includes/classes/test-concurrency-limiter.php
```diff
@@ -0,0 +1,101 @@
...
+	 * @dataProvider data_concurrency_limiting
+	 * @param string $backend
+	 * @psalm-param class-string<\Automattic\VIP\Search\ConcurrencyLimiter\BackendInterface> $backend
+	 */
+	public function test_concurrency_limiting( $backend ): void {
+		if ( ! $backend::is_supported() ) {
+			self::markTestSkipped( sprintf( 'Backend "%s" is not supported', $backend ) );
+		}
+
+		add_filter( 'vip_es_concurrency_limit_backend', function() use ( $backend ) {
```
rinatkhaziev, line 38:
```suggestion
		add_filter( 'vip_search_concurrency_limit_backend', function() use ( $backend ) {
```

tests/search/includes/classes/test-concurrency-limiter.php
```diff
@@ -0,0 +1,101 @@
...
+			return $backend;
+		} );
+
+		add_filter( 'ep_intercept_remote_request', '__return_true' );
+		add_filter( 'vip_es_max_concurrent_requests', function() {
```
rinatkhaziev, line 43:
```suggestion
		add_filter( 'vip_search_max_concurrent_requests', function() {
```

search/includes/classes/concurrency-limiter/class-object-cache-backend.php
```diff
@@ -0,0 +1,56 @@
+<?php
+
+namespace Automattic\VIP\Search\ConcurrencyLimiter;
+
+require_once __DIR__ . '/backendinterface.php';
+
+class Object_Cache_Backend implements BackendInterface {
+	const KEY_NAME   = 'vip_es_request_count';
```
rinatkhaziev, line 8:
```suggestion
	const KEY_NAME   = 'vip_search_concurrent_requests_count';
```

search/includes/classes/concurrency-limiter/class-object-cache-backend.php
```diff
@@ -0,0 +1,56 @@
...
+	const GROUP_NAME = 'vip_es';
```
rinatkhaziev, line 9:
```suggestion
	const GROUP_NAME = 'vip_search';
```

rinatkhaziev (changes requested):
Thanks for this! I left a bunch of naming suggestions (mostly `s/es/vip_search/`)

search/includes/classes/class-concurrency-limiter.php
```diff
@@ -0,0 +1,121 @@
...
+	 */
+	public function ep_do_intercept_request( $response ) {
+		// This filter can be called inside a loop; we need to make sure not to increment the counter more than once
+		if ( ! $this->doing_request ) {
+			$this->doing_request = true;
+			$this->should_fail   = ! $this->backend->inc_value();
+		}
+
+		$fail = apply_filters( 'vip_es_should_fail_excessive_request', $this->should_fail );
+		return $fail ? new WP_Error( 503, 'Concurrency limit exceeded' ) : $response;
```
rinatkhaziev, line 87: Nitpick but I think 429 would be more appropriate rather than 503?

search/includes/classes/class-concurrency-limiter.php
sjinks: Fixed in f84e495ca

search/includes/classes/class-concurrency-limiter.php
sjinks: Fixed in Fixed in f84e495ca

search/includes/classes/class-concurrency-limiter.php
sjinks: Fixed in f84e495ca

search/includes/classes/class-concurrency-limiter.php
sjinks: Fixed in f84e495ca

tests/search/includes/classes/test-concurrency-limiter.php
sjinks: Fixed in f84e495ca

tests/search/includes/classes/test-concurrency-limiter.php
sjinks: Fixed in f84e495ca

tests/search/includes/classes/test-concurrency-limiter.php
sjinks: Fixed in f84e495ca

search/includes/classes/concurrency-limiter/class-object-cache-backend.php
sjinks: Fixed in f84e495ca

search/includes/classes/concurrency-limiter/class-object-cache-backend.php
sjinks: Fixed in f84e495ca

search/includes/classes/class-concurrency-limiter.php
sjinks: OK, set to 429. If Systems implement limiting with nginx, it will return 503. I have updated `class-search.php` to check for both 429 and 503.

rebeccahum (approved):

rebeccahum:
r2124-stacks

PR #2706, opened 2021-12-13
Add concurrency limit backends

sjinks:
## Description

Add various backends for concurrency limiting.

## Plugin Updated: VIP Search

Add three backends for concurrency limiting (APCu, Object Cache, Semaphore).

## Checklist

## Steps to Test

The CI should pass.


search/includes/classes/class-concurrency-limiter.php
```diff
@@ -16,27 +22,52 @@ class Concurrency_Limiter {
...
 
-	public function __destruct() {
-		if ( $this->doing_request ) {
-			$this->dec_key();
+	public function init(): bool {
+		if ( defined( 'WP_TESTS_DOMAIN' ) ) {
+			$backend_class = Object_Cache_Backend::class;
+		} else {
+			$backend_class = 'Windows' === PHP_OS_FAMILY ? APCu_Backend::class : Semaphore_Backend::class;
+		}
```
sjinks, line 40: For tests, we use the most harmless class.

For Windows, we have to use APCu because PHP does not support semaphores there.

For other platforms, we use semaphores because they are the most reliable. However, once a semaphore I created, it is impossible to change its limit (and consequently, the number of concurrent connections allowed). This should not be much of an issue because all changes to the code trigger redeployment, and this will recreate a semaphore with the correct limits.

search/includes/classes/concurrency-limiter/class-semaphore-backend.php
```diff
@@ -0,0 +1,65 @@
...
+			while ( $this->increments > 0 ) {
+				sem_release( $this->sem );
+				--$this->increments;
+			}
+		}
+
+		--self::$instances;
+		if ( defined( 'WP_TESTS_DOMAIN' ) && $this->sem && 0 === self::$instances ) {
+			sem_remove( $this->sem );
+		}
```
sjinks, line 33: Semaphores are not deleted automatically; they persist even if the process that created them, died. For tests, we remove what we create so that we don't create too many semaphores (in case the tests are run on the live system, not in a container).

sjinks:
While we are not going to use APCu/Semaphore backends (at least yet), I decided not to remove them now.

PR #2711, opened 2021-12-14
Use a sub-directory for CF7 uploads to avoid CF7 cleanup deleting unrelated files

### LINKED ISSUE #2710: Contact Form 7 tmp workaround causes unexpected tmp files to be deleted
## Expected/Desired Behavior

The Contact Form 7 tmp folder workaround for attachments should be self-contained so that the clean-up routine in this plugin does not attempt to delete unrelated files.

## Actual Behavior

We have a workaround to avoid contact form 7 trying to use a path within the file system wrapper for uploaded attachment files. It is currently using `/tmp`

https://github.com/Automattic/vip-go-mu-plugins/blob/4156bc395bace0a6ccbf832b320ecfc927bf9de6/plugin-fixes.php#L18

Contact Form 7 contains a clean-up routine which periodically attempts to delete everything in the specified directory, which can have unintended effects e.g. deleting compiled blade template files used by sage theme, and returning warnings where it has tried and failed to delete other files within '/tmp' (e.g. newrelic log)

1. Install and activate a recent version of Contact Form 7 on a vip go site and sandbox it. 
2. Add some files to `/tmp` both as a privileged user and unprivileged user
3. The cleanup is initiated on `template_redirect` so navigate through some pages.
4. Check logs for warnings for failed deletes to files in the `/tmp` dir
5. Check for removed files added as an unprivileged user.

1. Note that as Contact Form 7 uses an ajax form submit, when sandboxed this will try to submit to the original domain rather than the sandbox domain. This makes debugging this tricky - it might be necessary to add "production" logging to the non-sandboxed version.
2. Recent versions of Contact Form 7 (5.0.5 and up) will not send attachments without patching the plugin to avoid the check performed by the function `wpcf7_is_file_path_in_content_dir`
3. This works as a solution to the initial issue described here, (but not the issue noted in 2. above):-

```
define( 'WPCF7_UPLOADS_TMP_DIR', get_temp_dir().'/cf7/' );
```

I will create a PR for the above.

rickhurst:
Fixes #2710 

## Description

This changes our Contact Form 7 workaround to use a subdirectory rather than storing files directly in `/tmp`, which CF7 then tries to clean-up, including any unrelated files.

### Compatibility: Contact Form 7
Changed path of `WPCF7_UPLOADS_TMP_DIR` to use a subdirectory of `/tmp`

## Checklist

## Steps to Test
1. Install and activate a recent version of Contact Form 7 on a VIP Go site and sandbox it.
2. Check for the presence of `/tmp/cf7`
3. On the sandbox, add some files directly to `/tmp` both as a privileged user and unprivileged user
4. The cleanup is initiated on `template_redirect` so navigate through some pages.
5. Check logs for warnings for failed deletes to files in the `/tmp` dir
6. Check for removed files added as an unprivileged user in `/tmp`.

Note there are some workarounds needed to actually test attachments with CF7 versions 5.0.5 and above - see related issue

### DIFF
--- a/plugin-fixes.php
+++ b/plugin-fixes.php
@@ -15,7 +15,7 @@ License: GPL version 2 or later - http://www.gnu.org/licenses/old-licenses/gpl-2
  * so let's just tell CF7 to put them in `/tmp/cf7/`.
  */
 if ( ! defined( 'WPCF7_UPLOADS_TMP_DIR' ) ) {
-	define( 'WPCF7_UPLOADS_TMP_DIR', get_temp_dir() . '/cf7/' );
+	define( 'WPCF7_UPLOADS_TMP_DIR', get_temp_dir() . 'cf7/' );
 }
 
 /**
--- a/plugin-fixes.php
+++ b/plugin-fixes.php
@@ -12,10 +12,10 @@ License: GPL version 2 or later - http://www.gnu.org/licenses/old-licenses/gpl-2
  *
  * The plugin attempts to write to a `wp-content` path which will fail.
  * These files are transient and only meant to be included as attachment,
- * so let's just tell CF7 to put them in `/tmp/`.
+ * so let's just tell CF7 to put them in `/tmp/cf7/`.
  */
 if ( ! defined( 'WPCF7_UPLOADS_TMP_DIR' ) ) {
-	define( 'WPCF7_UPLOADS_TMP_DIR', get_temp_dir() );
+	define( 'WPCF7_UPLOADS_TMP_DIR', get_temp_dir() . '/cf7/' );
 }
 
 /**


rinatkhaziev:
@rickhurst changes look good, can we adjust the Changelog description to read something like

`### Compatibility: Contact Form 7` and then the rest of it?

rinatkhaziev (dismissed):

rickhurst:
@rinatkhaziev Changelog updated!
@GaryJones Thanks for spotting the typo in the pull request title!

WPprodigy:
`get_temp_dir()` always returns w/ a trailing slash: https://developer.wordpress.org/reference/functions/get_temp_dir/

So this has double slashes atm: 

```
wp> get_temp_dir() . '/cf7/'
wp> /tmp//cf7/
```

WPprodigy (approved):

rebeccahum:
v20211215.0

PR #2715, opened 2021-12-15
Unblocks jobs that are missing scheduled_time

pschoffer:
## Description
Some data for search plugins are syncked to the ES using async queue.

However that queue can sometimes fail due to timeout or other causes.

For that reason, we have a [deadlock cleanup](https://github.com/Automattic/vip-go-mu-plugins/blob/2a30dcef9e8759989b1ba488fa7bfab692942228/search/includes/classes/class-queue.php#L714) mechanism that basically marks "jobs" to be ready for pickup again.

We had a scenario where the scheduled time for a job was null (which it should not be if it was already scheduled). I was not able to find the root cause behind this invalid state. However, this change makes deadlock jobs also pick up these loose jobs so that they can be rerun.

### Plugin Updated: Search

Enhancement for the asynchronous synchronization mechanism. This change address an edge case in which stuck jobs would not be retried by requeueing them again.

## Checklist

## Steps to Test

Monitor your queue `select status, scheduled_time from wp_vip_search_index_queue;` 
wp eval 'do_action("vip_search_queue_sweeper");'


sjinks (approved):

sjinks:
r2102-stacks

rebeccahum:
v20211215.0

PR #2718, opened 2021-12-15
Search: For index_exists cached option, only cache if response code is valid

rebeccahum:
## Description
Since only need to cache if the response is a 200 or 404. All other ones (i.e. 410) should be ignored and not cached. If a bad response has been stored, invalidate it.

### Plugin Updated: Search

Only cache index_exists requests that are valid. Invalidate if response code is not in safelist.

## Checklist

Please make sure the items below have been covered before requesting a review:

- [x] This change works and has been tested locally (or has an appropriate fallback).
- [ ] This change works and has been tested on a Go sandbox.
- [ ] This change has relevant unit tests (if applicable).
- [ ] This change has relevant documentation additions / updates (if applicable).
- [x] I've created a changelog description that aligns with the provided examples.

## Steps to Test
To test the actual caching of the index_exists option:

1) Make an index_exists request:
```
 $indexable = \ElasticPress\Indexables::factory()->get( 'post' );
$indexable->index_exists();
```
2) Get the option `wp option get es_index_exists_vip-200508-post-1`

PR #2719, opened 2021-12-15
Search: Add logging for more details on 401 response codes

rebeccahum:
## Description
We were seeing invalid responses (i.e. 401) cached in the index_exists option. Let's add logging to figure out the root cause. Related to #2718.

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -890,6 +890,27 @@ class Search {
 			update_option( $index_exists_option_name, $response );
 		}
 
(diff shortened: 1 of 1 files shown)


sjinks (approved):

rebeccahum:
@WPprodigy Before I deploy, just wanting to check if you have any other parameters you want fed into logging.

WPprodigy:
Looks good. Maybe we want to watch for more than just 401s since we don't really know what is happening out there, but limiting for now is okay so we can track down this particular thing.

rebeccahum:
r2110-stacks

PR #2720, opened 2021-12-15
Search: Alert if we don't have a queue count value cached OR if the queue count isn't decreasing

rebeccahum:
## Description
To reduce the noise, let's add caching for the queue count and alert when the queue count isn't decreasing. 

## Checklist

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -1157,15 +1157,21 @@ class Search {
 		$queue_stats = $this->queue->get_queue_stats();
 
 		if ( $queue_stats->average_wait_time > self::STALE_QUEUE_WAIT_LIMIT ) {
-			$message = sprintf(
-				'Average index queue wait time for application %d - %s is currently %d seconds. There are %d items in the queue and the oldest item is %d seconds old',
-				FILES_CLIENT_SITE_ID,
-				home_url(),
-				$queue_stats->average_wait_time,
-				$queue_stats->queue_count,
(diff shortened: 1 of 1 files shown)


sjinks (approved):

rebeccahum:
@sjinks What are your thoughts on not alerting it there is no cached value and only caching it the first time. Then on subsequent times, only alerting if the comparison of the cached value is not less than the actual value?

WPprodigy:
That could work, but I'd increase the cache timeout to longer than 15m then.

The only other thing to consider is an unhealthy site that has lots of memcache ejections, is possible our cache key goes away each time before it's up again to run in 5min. Seems unlikely, but worth considering.

sjinks:
> What are your thoughts on not alerting if there is no cached value and thus, only caching it the first time

:+1:

rebeccahum:
> The only other thing to consider is an unhealthy site that has lots of memcache ejections, is possible our cache key goes away each time before it's up again to run in 5min. Seems unlikely, but worth considering.

That did occur to me but I'm not sure what kind of failsafe we would have in place for it that wouldn't rely on memcache. I've updated the caching TTL to 45 minutes instead and I'll leave it to alert for the time being on no cached count value (just in case we do run into an unhealthy site that has constant memcache ejections — I don't think we want to be ina situation where we are expecting that alert and it doesn't alert).

WPprodigy (approved):

rebeccahum:
r2115-stacks

PR #2724, opened 2021-12-16
Changelog creator action

pschoffer:
## Description
Adds an action. This action is however only triggerable after merge...

## Checklist

### DIFF
--- a/bin/mark-production-changelog.js
+++ b/bin/mark-production-changelog.js
@@ -21,7 +21,7 @@ const fetchPage = async page => {
 		page,
 	}
 
-	const argString = Object.keys(queryArgs).map(argument => `${argument}=${queryArgs[argument]}`).join('&');
+	const argString = new URLSearchParams(queryArgs).toString();
 
 	const result = await fetch(`${CHANGELOG_ENDPOINT}?${argString}`, { headers });
 	if (!result.ok) {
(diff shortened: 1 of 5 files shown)


bin/mark-production-changelog.js
```diff
@@ -0,0 +1,87 @@
...
+	};
+	const queryArgs = {
+		'release-channel': RELEASE_CHANNEL_STAGING,
+		tags: 1784989, // mu-plugins
+		per_page: 100,
+		status: 'draft,publish',
+		page,
+	}
+
+	const argString = Object.keys(queryArgs).map(argument => `${argument}=${queryArgs[argument]}`).join('&');
```
sjinks, line 24:
```suggestion
	const argString = new URLSearchParams(queryArgs).toString();
```

sjinks (changes requested):

bin/mark-production-changelog.js
sjinks: (just in case we might have something that needs to be urlencoded first)

pschoffer: Cool.. yeah I was thinking:  I am rediscovering a wheel here right?

sjinks (dismissed):

sjinks (approved):

PR #2725, opened 2021-12-16
Search: Do not alert as an error if content validation is already ongoing

rebeccahum:
## Description
It's not necessary to alert that a validate-contents error occurred if it's just a validation that's already ongoing.


sjinks (approved):

rebeccahum:
r2113-stacks

PR #2726, opened 2021-12-16
Search: remove autohealing alert

rebeccahum:
## Description
The autohealing alert is not necessary — I think we only need to alert if an error occurred. Otherwise, it's safe to assume that it went fine and dandy.


sjinks (approved):

rebeccahum:
r2118-stacks

PR #2732, opened 2021-12-19
Replace $_ENV with getenv()

sjinks:
## Description

There is a setting in php.ini, [variables_order](https://www.php.net/manual/en/ini.core.php#ini.variables-order), controlling which superglobals PHP parses and sets. Although the default value is `EGPCS` (meaning that PHP sets the `$_ENV` superglobal), a notice in the `php.ini` [discourages](https://github.com/php/php-src/blob/701e5fe4aecfa4ba75d71c4ebe37e0e0cc8ed131/php.ini-production#L636-L647) this:

> ; This directive determines which super global arrays are registered when PHP
> ; starts up. G,P,C,E & S are abbreviations for the following respective super
> ; globals: GET, POST, COOKIE, ENV and SERVER. There is a performance penalty
> ; paid for the registration of these arrays and because ENV is not as commonly
> ; used as the others, ENV is not recommended on productions servers. You
> ; can still get access to the environment variables through getenv() should you
> ; need to.

This PR updates our code to use `getnev()` instead of `$_ENV`, should we want someday to follow the best practices :-) 

### Plugin Updated: VIP Init

Do not rely upon the `$_ENV` superglobal: depending on the PHP configuration parameters, it may or may not be populated.
## Checklist

## Steps to Test

The CI should pass.


GaryJones:
> should we want someday to follow the best practices 

Will you be following up to check for `$_ENV` usage in customer application code, and working towards getting that changed so we can turn off `ENV` support in the PHP config? Seems like this could be rolled out with the PHP 8 deployment?

pschoffer (approved):
Thank you!

rebeccahum:
r2129-stacks

sjinks:
v20211228.1

PR #2737, opened 2021-12-20
Filters plugin URLs to use sandboxed hostnames.

emrikol:
## Description

Adds a `wpvip_filter_sandbox_plugins_url` filter to the `plugins_url` hook to properly return sandboxed hostnames.

Currently `plugins_url( 'test' );` will return a URL like `https://example.com/wp-content/plugins/test` which does not allow for enqueuing assets that are sandboxed.

After the filter, it should properly return URLs like `https://example-com.sandbox-ingress-hostname.example.com/wp-content/plugins/test`

### Plugin Updated: vip-helpers

* Fixes a bug where incorrect URLs are returned by `plugins_url()` when sites are sandboxed by VIP support.

## Checklist


vip-helpers/sandbox.php
```diff
@@ -88,3 +88,21 @@ function wpcom_do_sandbox_bar() {
...
+
+/**
+ * Filters plugin URLs to use sandboxed hostnames.
+ *
+ * @param string $url The complete URL to the plugins directory including scheme and path.
+ *
+ * @return string Filtered URL using sandboxed hostname.
+ */
+function wpvip_filter_sandbox_plugins_url( $url ) {
+	if ( defined( 'WPCOM_SANDBOXED' ) && WPCOM_SANDBOXED ) {
```
rebeccahum, line 100: We don't need this check since the file returns if the constant isn't defined. Other than that, tested and LGTM.

rebeccahum (changes requested):

rebeccahum (changes requested):
Actually, I tested this further and in multi sites where the site URL contains another site URL, we can have search-replace overlapping.

e.g. subsite 1 is example.com and subsite 2 is example.com/test

Say we want to look at the plugin_url for subsite 2, it gets replaced with subsite 1 first.

vip-helpers/sandbox.php
```diff
@@ -88,3 +88,21 @@ function wpcom_do_sandbox_bar() {
...
+		global $sandbox_vhosts;
+
+		$url = str_replace( array_values( $sandbox_vhosts ), array_keys( $sandbox_vhosts ), $url );
```
rinatkhaziev, line 103:
```suggestion
		$url = str_replace( $_SERVER['HTTP_HOST'], array_flip( $sandbox_vhosts )[ $_SERVER['HTTP_HOST'] ], $url );
```

vip-helpers/sandbox.php
rinatkhaziev: I think this should need to be swapped with the check for the old platform?

rinatkhaziev (dismissed):
Thanks! Left a bit of inline feedback

sjinks:
@rebeccahum re: overlapping - could you please check the update implementation? It makes sure that we only replace the matching part in the URL domain.

rebeccahum (approved):

rebeccahum:
Tested & works after the most recent change.

rebeccahum:
r2248-stacks

PR #2739, opened 2021-12-21
Search: Fix is_cli logic in filter__ep_do_intercept_request

rebeccahum:
## Description
Follow up to #2719, let's define `$is_cli` before doing the check to assign the request URI for the 401 logging.


pschoffer (approved):

pschoffer (changes requested):
This is a quite small change but what do you think about adding changelog entry if it is touching production code?

rebeccahum:
I don't think it's necessary since it's just for our own logging purposes of 401s — it doesn't touch any client functionality otherwise.

rinatkhaziev (approved):

rebeccahum:
r2128-stacks

sjinks:
v20211228.1

PR #2742, opened 2021-12-21
Fix tests for WP trunk

sjinks:
Tests for WP trunk [fail](https://app.circleci.com/pipelines/github/Automattic/vip-go-mu-plugins/9248/workflows/fbf2e035-0f44-4be1-8eee-7b693fc44074/jobs/88669) because of the modifications to how the blocks are registered (arre yaar, I wish there was a way to tell WP not to load all that block stuff if I don't need it).

Here I am trying to fix that.

### DIFF
--- a/tests/search/includes/classes/test-class-versioning.php
+++ b/tests/search/includes/classes/test-class-versioning.php
@@ -19,6 +19,8 @@ class Versioning_Test extends WP_UnitTestCase {
 	public static function setUpBeforeClass(): void {
 		parent::setUpBeforeClass();
 
+		remove_all_actions( 'init' );
+
 		if ( ! defined( 'VIP_ELASTICSEARCH_ENDPOINTS' ) ) {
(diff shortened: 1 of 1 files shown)


rinatkhaziev:
@sjinks can we target the init hook of the specifically braking part instead of blanket remove_all_actions?

sjinks:
https://github.com/WordPress/WordPress/tree/master/wp-includes/blocks - every single file there registers its own `init` hook :-( 

Actually, it does not harm that we kill all hooks — `init` runs before the tests; we have to run `init` in our tests because WP kills all installed hooks when a test ends; EP installs its hooks during `init` or `plugins_loaded`. Basically, we [run `plugins_loaded` and `init`](https://github.com/Automattic/vip-go-mu-plugins/blob/ac8451a18c591dcc59051a98a42762c9dc0a6e36/tests/search/includes/classes/test-class-versioning.php#L38-L39) only to let ES re-install its hooks.

rinatkhaziev (approved):

PR #2743, opened 2021-12-22
Search: Basic request cache invalidation via last_changed approach

rinatkhaziev:
## Description

This adds basic invalidation via adding `last_changed` as a suffix for the cache key.

We have implemented cache invalidation after index and bulk index to make sure that any cached requests will not use stale data.

## Checklist

## Steps to Test

1. Check out PR.
2. Hit a search `?s=test`
3. Open Debug Bar or Query Monitor and verify there's a remote request to ES
4. Hit the search again
5. Verify there's no request
6. Perform an action that results in an indexing (e.g. save a post)
7. Hit the search again and verify the request is made again.

### DIFF
--- a/search/includes/classes/class-syncmanager-helper.php
+++ b/search/includes/classes/class-syncmanager-helper.php
@@ -52,6 +52,9 @@ final class SyncManager_Helper {
 		add_action( 'saved_term', [ $this, 'saved_term' ] );
 
 		add_filter( 'ep_skip_action_edited_term', [ $this, 'ep_skip_action_edited_term' ], 10, 4 );
+
+		// Invalidate the request cache on these hooks
+		add_action( 'save_post', [ $this, 'bump_last_changed' ] );
 	}
 
 	public function cleanup(): void {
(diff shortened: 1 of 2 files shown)


search/includes/classes/class-syncmanager-helper.php
```diff
@@ -64,6 +69,43 @@ public function cleanup(): void {
...
+			$this->bump_last_changed();
+		}
+	}
+
+	/**
+	 * Bump last updated what we can use to invalidate any cached requests
+	 *
+	 * @return bool 
+	 */
+	public function bump_last_changed() {
```
WPprodigy, line 105: Let's make this private until/if it's needed externally.

WPprodigy (approved):
Looks good, did a bunch of testing.

One note for our future persons - we should probably not increment the rate limit when it's a cache reply. The whole `filter__ep_do_intercept_request` could use some love/refactoring as it's complexity probably belongs in it's own class at this point.

WPprodigy (approved):

rinatkhaziev:
r2132-stacks

sjinks:
v20211228.1

PR #2749, opened 2021-12-24
Support Composer 2.2

sjinks:
This PR adds support for Composer 2.2.

Composer 2.2 introduces [more secure plugin execution](https://blog.packagist.com/composer-2-2/#more-secure-plugin-execution) and this PR adds support for it.

The idea was taken from Parsely/wp-parsely#559 :-)


GaryJones (approved):
✅

PR #2750, opened 2021-12-24
Fix flaky test

sjinks:
Ref: https://app.circleci.com/pipelines/github/Automattic/vip-go-mu-plugins/9317/workflows/0fb186e8-f2c8-496f-9e21-4f3b2cb19fd4/jobs/89967

> 1) Automattic\VIP\Search\Queue_Test::test_checkout_jobs
> Job 3 has the wrong scheduled_time
> Failed asserting that two strings are equal.
> --- Expected
> +++ Actual
> @@ @@
> -'2021-12-24 02:59:01'
> +'2021-12-24 02:59:02'

This happens when the `$expected_scheduled_time` value is captured on the last microseconds of the current second; `checkout_jobs()` happens then the next second and the test fails. This PR tries to fix it.


pschoffer (approved):
Smart!

PR #2755, opened 2021-12-27
Fix X-Forwarded-For in VIP_Request_Block

sjinks:
## Description

The original implementation of `VIP_Request_Block::ip()` incorrectly handles the `X-Forwarded-For` header, which sometimes leads to up to 2,625 additional IPs being blocked. This PR attempts to fix this issue.

For example, `VIP_Request_Block::ip( '1.1.1.1' )` could result in `11.1.1.1`, `21.1.1.1`,…, `251.1.1.199` being blocked.

As a side effect, this PR makes the `VIP_Request_Block` class more testable.

### Plugin Updated: VIP Request Block

The plugin now correctly handles partial IP matches in the X-Forwarded-For header.

## Checklist

## Steps to Test

I have added a test for my bug fix. The CI should pass.

### DIFF
--- a/lib/class-vip-request-block.php
+++ b/lib/class-vip-request-block.php
@@ -23,20 +23,23 @@ class VIP_Request_Block {
 	 * 🛑 ALWAYS: use `whois {IP}` to look up the IP before making the changes.
 	 *
 	 * @param string $value target IP address to be blocked.
+	 * @return bool|void
 	 */
 	public static function ip( string $value ) {
 		// Don't try to block if the passed value is not a valid IP.
 		if ( ! filter_var( $value, FILTER_VALIDATE_IP ) ) {
-			// phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
-			error_log( 'VIP Request Block: The value passed is not a correct IP address: ' . $value );
-			return;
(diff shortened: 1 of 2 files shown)


pschoffer (approved):
Thank you for the tests part of it as well!

sjinks:
r2136-stacks

lib/class-vip-request-block.php
```diff
@@ -41,32 +45,36 @@ public static function ip( string $value ) {
...
+		return false;
 	}
 
 	/**
 	 * Block by exact match for an arbitrary header.
 	 *
 	 * @param string $header HTTP header.
 	 * @param string $value header value.
-	 * @return void
+	 * @return void|bool
```
GaryJones, line 77: Seems like these functions can't return void anymore?

sjinks: @GaryJones they can, please check https://github.com/Automattic/vip-go-mu-plugins/pull/2755/files/49a367ea0b1e502ee6c3de2394bad6ce1b25856a#diff-2aaa18fe2bb071a15fb049eb6bc4dcf92755b5f1baa6171d13e26330f629b38fR108

These functions return `true` (or sometimes `false`) only when we run tests; in the production environment, `block_and_log()` will call `exit()`.

sjinks:
v20211228.1

PR #2756, opened 2021-12-27
VIP Request Block: handle different IPv6 notations

sjinks:
## Description

This PR tries to address the issue of multiple representations of an IPv6 address.

For example,
* 2001:4860:4860::88ff
* 2001:4860:4860:0:0:0:0:88ff
* 2001:4860:4860:0000:0000:0000:0000:88ff
* 2001:4860:4860::88FF

are representations of the same IP address.

We try to address this by using the `inet_pton()` function and comparing the internal representations of IP addresses, falling back to the case-insensitive comparison of IPs if PHP was compiled without the IPv6 support.

### Plugin Updated: VIP Request Block

Improved comparison of IPv6 addresses to handle some corner cases

## Checklist

## Steps to Test

I have added test cases to cover the new functionality; the CI should pass.


lib/class-vip-request-block.php
```diff
@@ -26,6 +26,8 @@ class VIP_Request_Block {
 	 * @return bool|void
 	 */
 	public static function ip( string $value ) {
+		$value = strtolower( $value );
```
sjinks, line 29: If PHP is compiled without IPv6, we fall back to case-insensitive string comparisons.

lib/class-vip-request-block.php
```diff
@@ -36,21 +38,31 @@ public static function ip( string $value ) {
 			return false;
 		}
 
+		// In case this is an IPv6 address and PHP is compiled without the IPV6 support, make sure `false`'s won't match.
+		if ( false === $ip ) {
```
rinatkhaziev, line 42: Should this be more explicit to include a check whether PHP was compiled without IPV6 support? As I'm reading the docs it says "false if a syntactically invalid ip is given",  so we kind of side-stepping the intent of that `false` .

rinatkhaziev (changes requested):
Thanks for this, left a minor note inline.

lib/class-vip-request-block.php
sjinks: @rinatkhaziev we validate the IP here: https://github.com/Automattic/vip-go-mu-plugins/pull/2756/files#diff-2aaa18fe2bb071a15fb049eb6bc4dcf92755b5f1baa6171d13e26330f629b38fL30

If it is not valid, we log a warning and exit.

rinatkhaziev (approved):

sjinks:
r2148-stacks

sjinks:
v20211229.0

PR #2748, opened 2021-12-24
Use Context for request type checks

sjinks:
## Description

Use `Context::is_XXX()` instead of "manual" request type checks

### Plugin Updated: VIP Init

`Context` class is now available in a single-site mode.

## Checklist

Please make sure the items below have been covered before requesting a review:

- [x] This change works and has been tested locally (or has an appropriate fallback).
- [ ] This change works and has been tested on a Go sandbox.
- [ ] This change has relevant unit tests (if applicable).
- [ ] This change has relevant documentation additions / updates (if applicable).
- [x] I've created a changelog description that aligns with the provided examples.

## Steps to Test

The CI should pass.

PR #2758, opened 2021-12-27
Revert "Revert "Use Context for request type checks""

sjinks:
Yo dawg, I heard you like reverts so I created a revert for your revert so you can revert while you revert.

Reverts Automattic/vip-go-mu-plugins#2757
See #2748 

### Plugin Updated: VIP Init

`Context` class is now available in a single-site mode.


sjinks:
Comparing to #2748, all the magic now happens in 8ad5e007a43a68a169ef170a3bf4ed2a7bbd15ae: we now load the `Context` class from `000-vip-init.php` if it has not been loaded from `wp-config.php` via `000-pre-vip-config/requires.php`.

This approach does not break the dev environment (it does not load `000-pre-vip-config/requires.php` from `wp-config.php`).

jesse-greathouse (approved):
This is really cool for syntax declutter. I like it.

PR #2759, opened 2021-12-28
Retrieve all existing tags before push

sjinks:
See https://github.com/Automattic/vip-go-mu-plugins/runs/4649774955?check_suite_focus=true


pschoffer (approved):
Thank you!

PR #2760, opened 2021-12-28
More robust tag generation

sjinks:
Use `git tag -l` + `awk` to generate the tag for the next release. Avoids loops and does not have a limit of 8 tags per day,


pschoffer (approved):
Perfect

PR #2761, opened 2021-12-28
Fix concurrency limit value increment

sjinks:
Ensure the key exists before trying to increment it.

Improvements to the concurrency limiter:
  * make sure the key exists before trying to increment it;
  * make sure that the value will not become negative after a decrement;
  * fix possible race conditions.


search/includes/classes/concurrency-limiter/class-object-cache-backend.php
```diff
@@ -24,12 +22,13 @@ public function __destruct() {
...
 	public function initialize( int $limit, int $ttl ): void {
 		$this->limit = $limit;
+		$this->ttl   = $ttl;
 
 		$found = false;
 		$value = wp_cache_get( self::KEY_NAME, self::GROUP_NAME, false, $found );
 		if ( ! $found || ! is_int( $value ) ) {
 			// phpcs:ignore WordPressVIPMinimum.Performance.LowExpiryCacheTime.CacheTimeUndetermined
-			wp_cache_set( self::KEY_NAME, 0, self::GROUP_NAME, $ttl );
+			wp_cache_set( self::KEY_NAME, 0, self::GROUP_NAME, $this->ttl );
```
rinatkhaziev, line 31: Does it make sense to use `wp_cache_add` here too to avoid potential race conditions?

rinatkhaziev (approved):
Thanks!

rinatkhaziev:
r2143-stacks

PR #2767, opened 2021-12-29
Better check for if current query is cacheable

rinatkhaziev:
## Description

Instead of a `foreach` check using Jetpack's helper function  `wp_in` this check can be simplified by using a regex.

## Checklist

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -980,7 +980,7 @@ class Search {
 			return $is_cacheable;
 		}
 
-		$is_cacheable = (bool) preg_match( '#(/_(search|mget|doc))', $url );
+		$is_cacheable = (bool) preg_match( '#(/_(search|mget|doc))#', $url );
 
 
 		/**
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -980,7 +980,7 @@ class Search {
 			return $is_cacheable;
 		}
 
-		$is_cacheable = (bool) preg_match( '#(/_(search|mget|doc))#', $url );
+		$is_cacheable = (bool) preg_match( '#/_(search|mget|doc)#', $url );
 
 
 		/**
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -980,12 +980,8 @@ class Search {
 			return $is_cacheable;
 		}
 
-		foreach ( [ '_search', '_mget', '_doc' ] as $needle ) {
-			if ( wp_in( $needle, $url ) ) {
-				$is_cacheable = true;
-				break;
-			}
-		}
+		$is_cacheable = (bool) preg_match( '#(/_(search|mget|doc))', $url );
+
 
 		/**
 		 * Filter if request is cacheable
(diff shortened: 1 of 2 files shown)


rinatkhaziev:
props @sjinks for the catch

tests/search/includes/classes/test-class-search.php
```diff
@@ -142,6 +142,78 @@ public function vip_search_filter_ep_index_name_with_versions_data() {
...
+				false,
+			),
+			// Random other url
+			array(
+				// The $query object
+				array(
+					'url' => 'https://foo.com/index/type/_anything',
+				),
+				// The expected result
+				2,
```
WPprodigy, line 199: Why is it 2 instead of a boolean?

search/includes/classes/class-search.php
```diff
@@ -980,12 +980,8 @@ public function is_url_query_cacheable( string $url, $args ): bool {
 			return $is_cacheable;
 		}
 
-		foreach ( [ '_search', '_mget', '_doc' ] as $needle ) {
-			if ( wp_in( $needle, $url ) ) {
-				$is_cacheable = true;
-				break;
-			}
-		}
+		$is_cacheable = (bool) preg_match( '#/_(search|mget|doc)#', $url );
```
rebeccahum, line 983: While this is a cleaner solution, regex typically is slower. 

Using `hrtime()` for benchmarking:
With preg_match - 40167
Using old solution - 7417

Do you think the above is negligible enough to not affect the performance since it runs on every intercepted request?

tests/search/includes/classes/test-class-search.php
rebeccahum: Seeing tests fail from this:
> Failed asserting that false matches expected 2.

search/includes/classes/class-search.php
rinatkhaziev: @rebeccahum  That's a good point, I obviously didn't measure. over how many iterations did you do the benchmarking?

rebeccahum: @rinatkhaziev Just 5-6 times and those were the averages.

rinatkhaziev: @rebeccahum https://replit.com/@rinatkhaziev/SpecializedAnguishedBruteforceprogramming#main.php

The old one is indeed more performant but the fluctuation on small repeats (where $i < 10) is wildly different, I got anywhere from 2 to 10 times slower. However, on 100000 iterations the difference is like 50%-200%, given that this is in milliseconds land  I think it should be fine?

rinatkhaziev: Scratch that, silly me forgot to reset the timer, the worst I could get is this:

Over 1 million iterations, the first one is the old approach.

```
string(10) "Elapsed ms"
float(277.970179)

string(10) "Elapsed ms"
float(363.893794)
```

So I think it's definitely fine :)

rebeccahum (approved):

rinatkhaziev:
r2156-stacks

sjinks:
v20220111.1

PR #2776, opened 2022-01-05
Search: Build new index in the background if shard requirements are not met

rebeccahum:
## Description
We should automatically build new indexes for sites that do not meet the shard requirements. 

When the health job runs and it detects that shard requirements are not met, it will schedule a cron job that utilizes the handy `--using-versions` command option of indexing to create a new index, index it and hot-swap it. There is alerting in place to let us know if something fails or if it cannot be automatically built and hot-swapped.

**Right now, this is limited to NON-PROD sites.**

New index versions will be automatically created in the background if shard requirements are not met.

## Checklist

## Steps to Test
You can shell in and use the below steps to test on a myriad of scenarios:
```
$es->init();
$hj->maybe_schedule_build_new_index( $indexable );
```


search/includes/classes/class-settingshealthjob.php
```diff
@@ -200,4 +220,84 @@ public function send_alert( $channel, $message, $level, $type = '' ) {
...
+			// No lock, check if event has been scheduled already.
+			if ( ! wp_next_scheduled( self::CRON_EVENT_BUILD_NAME ) ) {
+				$scheduled_event = wp_schedule_single_event( time() + 30, self::CRON_EVENT_BUILD_NAME, [ $indexable ] );
+
+				if ( is_wp_error( $scheduled_event ) ) {
+					// Send alert that failed to schedule event.
+					$error_message = $scheduled_event->get_error_message() ?? 'Event could not be scheduled';
+					$message       = sprintf( 'Failed to schedule event for building new %s index version on %s to meet shard requirements: %s', $indexable->slug, home_url(), $error_message );
+					$this->send_alert( '#vip-go-es-alerts', $message, 2 );
+				}
```
WPprodigy, line 248: Gotta pass the 4th flag to `wp_schedule_single_event` to get a WP Error back on failure: https://developer.wordpress.org/reference/functions/wp_schedule_single_event/

Honestly though this is very very unlikely, I'd personally just drop the extra error handling / alerting here.

search/includes/classes/class-settingshealthjob.php
```diff
@@ -200,4 +220,84 @@ public function send_alert( $channel, $message, $level, $type = '' ) {
...
+		update_option( self::BUILD_LOCK_NAME, time() ); // Set lock for starting rebuild.
+
+		// Do the indexing.
+		$assoc_args = [
+			'using-versions' => true,
+			'skip-confirm'   => true,
+			'indexables'     => $indexable->slug,
+		];
+		$cmd        = 'vip-search index ' . WP_CLI\Utils\assoc_args_to_str( $assoc_args );
+		$result     = WP_CLI::runcommand(
```
WPprodigy, line 284: Is it possible to call the underlining functions rather than running the CLI in a child process?

rebeccahum: @WPprodigy That was my initial approach but then, I realized that I wanted to capture any errors returned. To accomplish that  I think the underlining function would require re-factoring to capture any errors to be used outside a non-CLI context.

search/includes/classes/class-settingshealthjob.php
rebeccahum: Good catch and suggestion noted!

search/includes/classes/class-settingshealthjob.php
```diff
@@ -149,24 +162,31 @@ public function heal_index_settings( $unhealthy_indexables ) {
...
+					continue;
 				}
+				// Check if index needs to be re-built in the background or just do auto-healing.
+				if ( true === array_key_exists( 'index.number_of_shards', $result['diff'] ) ) {
+					$this->maybe_build_new_index( $indexable );
+				} else {
+					$diff = $this->health::limit_index_settings_to_keys( $result['diff'], $this->health::INDEX_SETTINGS_HEALTH_AUTO_HEAL_KEYS );
+					if ( empty( $diff ) ) {
+						continue;
+					}
```
WPprodigy, line 175: This will stop the auto-healing I think if shards count is a mismatch, is that intended?

So if a site ends up "stuck" in the position where it has the wrong number of shards but can't auto-create a new one (already has 2+ indexes) - it will also stop auto-healing the one it is using.

search/includes/classes/class-settingshealthjob.php
WPprodigy: Gotcha. Should be fine, just typically worth avoiding the extra complexity if possible.

search/includes/classes/class-settingshealthjob.php
rebeccahum: @WPprodigy It was intentional because it doesn't make sense to update the index settings if the index is being re-built (whether automatically or manually after alerting). 

Good point though — might be worth it to do it anyways to prevent a scenario where a site with 2+ indexes does not get the autohealing of the other index settings.

sjinks (approved):

rebeccahum:
r2158-stacks

sjinks:
v20220111.1

PR #2781, opened 2022-01-07
Search: Add namespaced shims for str_contains, str_starts_with, str_ends_with 

rinatkhaziev:
## Description

There were a couple of places where we used Jetpack helpers in our code, since there's no guarantee Jetpack is loaded we should get rid of those calls.

Opted to go with namespaced shims for PHP8 functions (up for discussion though)

## Checklist


rebeccahum:
Are these helper functions worth adding into a general non-search utils file? I'm seeing them used on other parts of `vip-go-mu-plugins` like https://github.com/Automattic/vip-go-mu-plugins/blob/41c4b4ac10c08dbfd0a9d5ad3faaa64ea1614ba3/a8c-files.php#L514.

search/includes/functions/helpers.php
```diff
@@ -0,0 +1,18 @@
...
+
+// source: Laravel Framework
+// https://github.com/laravel/framework/blob/8.x/src/Illuminate/Support/Str.php
+
+function str_starts_with( $haystack, $needle ) {
+	return (string) $needle !== '' && strncmp( $haystack, $needle, strlen( $needle ) ) === 0;
+}
+
+function str_ends_with( $haystack, $needle ) {
+	return $needle !== '' && substr( $haystack, -strlen( $needle ) ) === (string) $needle;
```
sjinks, line 13: Should we cast `$needle` to `string` here like we do this in `str_starts_with()`?

Or, what do you think if we add types to function arguments, like this:

```php
function str_starts_with( string $haystack, string $needle ): bool
```

Then we won't need explicit type casts…

GaryJones:
I'd rather see each of these new functions wrapped in `! function_exists()` conditions, and then use the functions without namespace prefixes - see https://github.com/WordPress/WordPress/blob/3ec2d2fcfae0cad2144d21b5ca7d8243b06bf245/wp-includes/compat.php#L438.

Since they are in WP 5.9 core as well, then once all customers are on 5.9, this compatibility shim could be removed, without having to also find each instance of usage.

PR #2782, opened 2022-01-07
Fix Search Dev Tools fatal error on when response body is empty

rinatkhaziev:
## Description

In a rare case where the request is 200 but the body of it is empty Search Dev Tools triggered a fatal due to sanitize_query_response expecting an object. but `json_decode('')` results in `NULL`. 

We catch that via null coalescing operator

We fixed an edge case that may have caused a fatal error in Search Dev Tools when the Enterprise Search request was technically valid but its body was empty.

## Checklist


rebeccahum (approved):

rinatkhaziev:
r2197-stacks

PR #2788, opened 2022-01-10
Fix E2E workflow

sjinks:
E2E workflow appears to be broken:

broken-e2e

Trying to fix it.

Reason: https://github.com/Marak/colors.js/blob/master/lib/index.js#L16-L23

It will take some time until `vip` fixes its dependencies; this PR tries to fix the tests now.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -43,7 +43,7 @@ jobs:
       - name: Install VIP CLI
         run: |
           npm install --prefix=$HOME/.local -g @automattic/vip
-          cd $HOME/.local/lib/node_modules/@automattic/vip && npm i colors@1.4.1
+          cd $HOME/.local/lib/node_modules/@automattic/vip && npm i colors@1.4.1 --force
 
       - name: Determine WP version
         id: wpver
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -28,25 +28,45 @@ jobs:
         with:
           submodules: true
           path: vip-go-mu-plugins
+
       - name: Setup Node
         uses: actions/setup-node@v2
         with:
           node-version: '16'
+          cache: npm
+
       - name: Install dependencies
-        run: cd ./vip-go-mu-plugins && npm install
-      - name: Install VIP Cli
+        run: npm ci
+        working-directory: vip-go-mu-plugins
+
+      - name: Install VIP CLI
         run: npm install --prefix=$HOME/.local -g @automattic/vip
+
+      - name: Determine WP version
+        id: wpver
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e.yml
```diff
@@ -28,30 +28,53 @@ jobs:
...
-        run: cd ./vip-go-mu-plugins && npm install
-      - name: Install VIP Cli
-        run: npm install --prefix=$HOME/.local -g @automattic/vip
+        run: npm ci
+        working-directory: vip-go-mu-plugins
+
+      - name: Install VIP CLI
+        run: |
+          npm install --prefix=$HOME/.local -g @automattic/vip
+          cd $HOME/.local/lib/node_modules/@automattic/vip && npm i colors@1.4.1 --force
```
sjinks, line 46: colors@1.4.2 is broken; install the last working version (`--force` is required because of the unsatisfied peer dependencies).

.github/workflows/e2e.yml
```diff
@@ -28,30 +28,53 @@ jobs:
...
         with:
           node-version: '16'
+          cache: npm
+          cache-dependency-path: vip-go-mu-plugins/package-lock.json
+
       - name: Install dependencies
-        run: cd ./vip-go-mu-plugins && npm install
-      - name: Install VIP Cli
-        run: npm install --prefix=$HOME/.local -g @automattic/vip
+        run: npm ci
```
sjinks, line 40: Use `npm ci` instead of `npm install` to ensure that we won't install any new package versions; only what is set in `package-lock.json` will be installed.

.github/workflows/e2e.yml
```diff
@@ -28,30 +28,53 @@ jobs:
...
+          cd $HOME/.local/lib/node_modules/@automattic/vip && npm i colors@1.4.1 --force
+
+      - name: Determine WP version
+        id: wpver
+        run: |
+          if [ -z "${{ github.event.inputs.wpversion }}" ]; then
+            echo ::set-output name=wpversion::latest
+          else
+            echo ::set-output name=wpversion::${{ github.event.inputs.wpversion }}
+          fi
```
sjinks, line 55: For pull requests, `github.event.inputs.wpversion` is empty; this makes `setup-e2e-env` complain.

.github/workflows/e2e.yml
pschoffer: This is a temporary fix untill `vip` is fixed right?

Should we leave the note here about that?

sjinks: Once I merge this PR, I will create another one reverting this change.

pschoffer: Sounds good:)

pschoffer (approved):

PR #2795, opened 2022-01-11
Search: in build_new_job(), pass in string slug instead

rebeccahum:
## Description

In `build_new_job()`, pass in the slug rather than the entire object since we don't need it and I want to prevent any potential bugs that occur during serialization/deserialization.


WPprodigy (approved):

rebeccahum:
r2179-stacks

PR #2805, opened 2022-01-13
Query Monitor 3.7.1 -> 3.8.2

johnbillion:
## Description

This updates the Query Monitor plugin to version 3.8.2 which includes native support for the VIP-specific format for logged queries in HyperDB, thus removing the need for the custom `WPCOM_VIP_QM_Collector_DB_Queries` class.

The reason QM now supports the VIP-specific format for logged queries is that the presence of the custom query collector on VIP was preventing me from making some planned wider changes that would require me to take it into consideration for VIP compatibility, thus increasing time and effort.

The benefit this has is that QM can now display a more useful call stack for each query instead of `hyperdb->query` being shown for every query, and it can also show information about duplicate queries, both of which are not currently available on VIP.

For code review purposes I recommend viewing the changes without whitespace changes included because quite a few whitespace adjustments were made in 3.8.0.

## Changelog Description

Full changelog here: https://github.com/johnbillion/query-monitor/releases (see 3.8.0 to 3.8.2).

This is in addition to the VIP logged query format support which I didn't mention in the changelog because it's not technically supported until it's rolled out on VIP :-)

## Checklist

## Steps to Test

This needs to be tested on a VIP environment that uses HyperDB in the same manner that VIP production environments do.

1. View a page while logged in as a user who can view Query Monitor
2. Verify that `hyperdb->query` is no longer shown in the Caller column of the Queries panel in Query Monitor and that a useful calling function is shown instead
3. Verify that the query SQL and time is correctly shown for each query
4. Verify the Queries -> Query by Caller panel shows information about each caller
5. Verify that no PHP errors are triggered by Query Monitor itself


rinatkhaziev:
Thanks, @johnbillion!

WPprodigy:
Oh, was literally just looking at improving the query collector last week! Thanks a bunch

WPprodigy:
Gonna pop this into a temp branch so it can be rebased and tested against production stacks. Will have a follow-up PR for merging into master.

PR #2812, opened 2022-01-14
Introduce php-compat helper that currently contains polyfills for PHP 8 new string functions

rinatkhaziev:
## Description

This is an alternative approach to #2781

Instead of introducing the namespaced versions, we'll define polyfills. Those are copied from core's compat.php

This file is not included in the bootstrap on purpose to avoid any potential deploy-related issues.

## Checklist


WPprodigy:
I like this better for sure. Just noting that we'll want to make sure the file is loaded after core defines the functions, but early enough so that users of them find them defined in time.

rinatkhaziev:
@WPprodigy I was thinking of pre-vip-config, but there's a major setback we need to address first - namely we need to reach out to customers and let them know that this file will be a hard dependency, so folks need to address their non-VIP staging environments, CI pipelines, etc. It'll take time, so probably during the initial rollout we can load it on top of 000-vip-init.php

GaryJones:
> Just noting that we'll want to make sure the file is loaded after core defines the functions

If core is defining the functions, why are we also trying to define them as well?

sjinks:
@GaryJones the core defines these functions only in 5.9; we still need to support earlier versions of WP.

lib/helpers/php-compat.php
```diff
@@ -0,0 +1,68 @@
...
+	 *
+	 * @param string $haystack The string to search in.
+	 * @param string $needle   The substring to search for in the `$haystack`.
+	 * @return bool True if `$haystack` starts with `$needle`, otherwise false.
+	 */
+	function str_starts_with( $haystack, $needle ) {
+		if ( '' === $needle ) {
+			return true;
+		}
+		return 0 === strpos( $haystack, $needle );
```
sjinks, line 44: What I dislike in this implementation is that `strpos()` will have to scan `strlen($haystack) - strlen($needle)` bytes if the string does not start with `$needle`. The longer the string is, the worse performance will be.

What do you think about

```php
return 0 === strncmp( $haystack, $needle, strlen( $needle ) );
```
?

sjinks: In this case, we can even drop the `if ( '' === $needle )` condition, because `strncmp( $haystack, '', 0 )` will always return 0.

lib/helpers/php-compat.php
```diff
@@ -0,0 +1,68 @@
...
+	 * @param string $haystack The string to search in.
+	 * @param string $needle   The substring to search for in the `$haystack`.
+	 * @return bool True if `$haystack` ends with `$needle`, otherwise false.
+	 */
+	function str_ends_with( $haystack, $needle ) {
+		if ( '' === $haystack && '' !== $needle ) {
+			return false;
+		}
+		$len = strlen( $needle );
+		return 0 === substr_compare( $haystack, $needle, -$len, $len );
```
sjinks, line 66: We [don't need](https://www.php.net/substr_compare) to specify the fourth parameter; that's the matter of taste, but we then can simplify the code to

```php
return '' === $needle || ( '' !== $haystack && 0 === substr_compare( $haystack, $needle, -strlen( $needle ) ) );
```
(and even drop the outer parentheses).

GaryJones:
Oh, of course, I knew this previously 😖

lib/helpers/php-compat.php
rinatkhaziev: Thanks, I adjusted the code.

sjinks (approved):

lib/helpers/php-compat.php
rinatkhaziev: Your suggestion is actually harder to read (to me at least), so let's just keep it as-is. At this point, it's also a bit of a moot discussion because the absolute majority of the sites are running 5.9 and the core would load its version first.

jesse-greathouse:
r2241-stacks

jesse-greathouse:
Release v20220127.0

PR #2813, opened 2022-01-15
Make Context available for non-MU sites

sjinks:
## Description

#2758 had to be reverted because of a sudden spike in errors; supposedly, that was because of the asynchronous nature of deployments: if the "main" code gets updated before `requires.php` does, this will cause fatal errors.

We now want to try another approach: first, deploy the code that makes the Context class globally available. After that, try to deploy the rest of the original PR.

### Plugin Updated: VIP Init

`Context` class is now available in a single-site mode.

## Checklist

## Steps to Test

Unfortunately, tests do not catch the bugs here :-(

### DIFF
--- a/000-pre-vip-config/requires.php
+++ b/000-pre-vip-config/requires.php
@@ -8,21 +8,16 @@
 
 $mu_plugins_base = dirname( __DIR__ );
 
-// Load custom error logging functions, if available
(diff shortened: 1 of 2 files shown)


rinatkhaziev:
@sjinks looks good, remember though we need to give a heads-up before deploying the other one.

rinatkhaziev (approved):

PR #2815, opened 2022-01-16
Harden GitHub workflows with StepSecurity

sjinks:
Trying the [suggestions by StepSecurity](https://app.stepsecurity.io/) to secure our GHA workflows.


varunsh-coder:
Happy to see you using `harden-runner`! I am the developer for it. Please let me know if you have feedback about ease of use, or anything else. 

I observed that you tried to restrict the endpoints, but the workflow failed. Sorry about that. 

The reason was that frsnacprodeus2file1.blob.core.windows.net is the GitHub caching storage account for your workflow, which did not show up as external call in the first run. So, it was not added to the allowed endpoints list. I will investigate why it did not show up in the first run. It is showing up in the more recent runs: https://app.stepsecurity.io/github/Automattic/vip-go-mu-plugins/actions/runs/1703876152

sjinks:
@varunsh-coder thank you! There was another IP missing from the logs:

Log: https://github.com/Automattic/vip-go-mu-plugins/runs/4831388885?check_suite_focus=true
Message: `Error: request to *** failed, reason: connect ECONNREFUSED 54.185.253.63:443`

sjinks:
> It is showing up in the more recent runs: https://app.stepsecurity.io/github/Automattic/vip-go-mu-plugins/actions/runs/1703876152

I am confused by these lines:

```
- uses: step-security/harden-runner@v1
  with:
    allowed-endpoints: 
...
      undefined:3306
      undefined:53
      undefined:65535
...
```

As far as I understand the logic, the system will try to resolve these names (`undefined`) when adding them to `iptables`, but that will fail with NXDOMAIN.

varunsh-coder:
> @varunsh-coder thank you! There was another IP missing from the logs:
> 
> Log: https://github.com/Automattic/vip-go-mu-plugins/runs/4831388885?check_suite_focus=true Message: `Error: request to *** failed, reason: connect ECONNREFUSED 54.185.253.63:443`

That IP address is StepSecurity's sinkhole IP Address (code [here](https://github.com/step-security/agent/blob/af1ba5a6df552deacb93c3fc53a3b67ee2c101a4/dnsproxy.go#L47)). When any of the job steps try to resolve a domain name, `harden-runner` agent checks if the domain is in the `allowed-endpoints` list. If not, the DNS proxy returns a sinkhole IP address (code [here](https://github.com/step-security/agent/blob/af1ba5a6df552deacb93c3fc53a3b67ee2c101a4/dnsproxy.go#L175)). This is to let the step continue so it makes the network call to that IP address, which is then blocked at the firewall. This helps to know what process made the call. A call to the sinkhole IP address (54.185.253.63) will only happen when `allowed-endpoints` are specified, and a call is made to a domain not in the allowed list. 

In this case, the call was to the cache storage domain, which was not in the `allowed-endpoints` list. `harden-runner` then creates a annotation to show the domain resolution that was blocked:
https://github.com/Automattic/vip-go-mu-plugins/actions/runs/1703869264
> DNS resolution for domain frsnacprodeus2file1.blob.core.windows.net. was blocked

varunsh-coder:
> > It is showing up in the more recent runs: https://app.stepsecurity.io/github/Automattic/vip-go-mu-plugins/actions/runs/1703876152
> 
> I am confused by these lines:
> 
> ```
> - uses: step-security/harden-runner@v1
>   with:
>     allowed-endpoints: 
> ...
>       undefined:3306
>       undefined:53
>       undefined:65535
> ...
> ```
> 
> As far as I understand the logic, the system will try to resolve these names (`undefined`) when adding them to `iptables`, but that will fail with NXDOMAIN.

That is a bug. I will work to fix it. 2 of those calls were to the local network (`:::65535` and `:::3306` from step `Setup test environment`), so they should not have been shown in the report. The third call to the IPv6 address (`2001:4860:4860::8888:53` from step `Run e2e tests`) needs more investigation. I will investigate that and get back soon...

sjinks:
@varunsh-coder I believe `[2001:4860:4860::8888]:53` is `dns.google`; the query came from Chromium (we use headless Chromium in E2E tests). It *probably* by default uses Google's DNS.

varunsh-coder:
> @varunsh-coder I believe `[2001:4860:4860::8888]:53` is `dns.google`; the query came from Chromium (we use headless Chromium in E2E tests). It _probably_ by default uses Google's DNS.

Yes, found it here: https://developers.google.com/speed/public-dns/docs/using. So, looks like Chromium uses the DNS IPv6 address directly. 

> The Google Public DNS IPv6 addresses are as follows:
   2001:4860:4860::8888
   2001:4860:4860::8844

sjinks:
So if I add `dns.google:53` to `allowed-endpoints`, this should cover both IPv4 and IPv6 addresses, correct?

varunsh-coder:
> So if I add `dns.google:53` to `allowed-endpoints`, this should cover both IPv4 and IPv6 addresses, correct?

I don't think so. If Chromium was using `dns.google` as a domain name, the recommended `allowed-endpoints` policy would have shown `dns.google` as a domain name. Looks like Chromium calls the IPv6 DNS address `2001:4860:4860::8888` directly. It does not resolve the IP address from `dns.google`. This makes sense because that IP address is fixed. 

So, you will need to add `2001:4860:4860::8888:53` to the `allowed-endpoint` list. `allowed-endpoint` list does not support IP addresses as of now. 

I could detect use of Google DNS IP addresses and show them in the recommended policy as `dns.google`. Then handle the firewall rules when that domain is added to `allowed-endpoints` list. That way developers would not be surprised to see the odd IP addresses. Would this work for you? Thanks for your patience!

sjinks:
Yes, sure, let's do this.

sjinks:
@varunsh-coder the most recent run: https://github.com/Automattic/vip-go-mu-plugins/actions/runs/1706745975

varunsh-coder:
> @varunsh-coder the most recent run: https://github.com/Automattic/vip-go-mu-plugins/actions/runs/1706745975

I found the issue. `e2e-test-site.vipdev.lndo.site` resolves to `127.0.0.1` while in audit mode. Since it resolves to `127.0.0.1`, the domain is not being shown in the recommended policy. 
https://github.com/Automattic/vip-go-mu-plugins/runs/4833376619?check_suite_focus=true#step:26:744

Please note, the `dns.google` is not resolved yet. 

Can you please run this in `audit` mode for the time being. I will fix these issues, test this out on a fork, and get back once I fix these issues?

sjinks:
Running.

https://app.stepsecurity.io/github/Automattic/vip-go-mu-plugins/actions/runs/1706848146

sjinks:
@varunsh-coder it worked in the block mode, but with two alerts:

```
Error: DNS resolution for domain planet.wordpress.org. was blocked
Error: DNS resolution for domain api.snapcraft.io. was blocked
```

https://github.com/Automattic/vip-go-mu-plugins/runs/4837955540?check_suite_focus=true

sjinks:
It did not mention those endpoints in the audit mode: https://app.stepsecurity.io/github/Automattic/vip-go-mu-plugins/actions/runs/1706848146

varunsh-coder:
> @varunsh-coder it worked in the block mode, but with two alerts:
> 
> ```
> Error: DNS resolution for domain planet.wordpress.org. was blocked
> Error: DNS resolution for domain api.snapcraft.io. was blocked
> ```
> 
> https://github.com/Automattic/vip-go-mu-plugins/runs/4837955540?check_suite_focus=true

`api.snapcraft.io` is not used by your workflow. It is a called made by Ubuntu VM. I will report it to GitHub. I had reported similar outbound calls that are unnecessary. You can ignore this. It is fine to block it. It will not affect the workflow. 

`planet.wordpress.org` - I see this in [some](https://app.stepsecurity.io/github/Automattic/vip-go-mu-plugins/actions/runs/1706802127) workflows, but not all. I will investigate. Can you please add to `allowed-endpoints`? sorry about all the bugs - it is a new system...

sjinks:
It worked, thanks!

.github/workflows/e2e.yml
```diff
@@ -19,20 +19,58 @@ on:
...
         required: false
         default: ''
+
+permissions: read-all
+
 jobs:
   Run_e2e_tests:
+    name: Run E2E Tests
+    permissions:
+      contents: read
```
GaryJones, line 29: Since you currently have `permissions: read-all`, then these two lines seem redundant.

varunsh-coder: Hi @GaryJones, `read-all` also has additional `read` permissions like `security-events: read`, `pull-requests: read` etc. This is why the StepSecurity automation adds `contents: read` explicitly here. I am getting similar feedback though on GitHub Action starter workflows as well (see this [issue](https://github.com/actions/starter-workflows/pull/1341)). One proposal is to have a config setting/ checkbox `Only add job level permissions if they are higher than read-all` in the automation, which would be more aligned with what you are looking for. Or would you prefer that the automation not add the workflow level `permissions: read-all` and just add the job level permissions, which are specific?

sjinks: > Or would you prefer that the automation not add the workflow level permissions: read-all and just add the job level permissions, which are specific?

@varunsh-coder OSSF [does not recommend](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) that. The reason is that when you add a new job to the workflow, it is easy to forget to add the `permissions:` option to it.

The best way is probably to have the most restrictive `permissions` for the workflow and adjust the per-job permissions as necessary.

PS: that OSSF document is how I have got to know about StepSecurity :-)

.github/workflows/e2e.yml
```diff
@@ -19,20 +19,58 @@ on:
         description: 'Alternative Password'
         required: false
         default: ''
+
+permissions: read-all
```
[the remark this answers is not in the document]
sjinks: Good point, will do that now.

.github/workflows/e2e.yml
GaryJones: I found that the default for those additional permissions is (from another repo):

```
Actions: write
Checks: write
Contents: write
Deployments: write
Discussions: write
Issues: write
Metadata: read
Packages: write
Pages: write
PullRequests: write
RepositoryProjects: write
SecurityEvents: write
Statuses: write
```

So, in that sense, adding `permission: read-all` makes sense to restrict the `write` permission.

But as noted in [my other comment](https://github.com/Automattic/vip-go-mu-plugins/pull/2815#discussion_r786843895), by using the repo settings, this can be dropped even further down to just:

```
Contents: read
Metadata: read
```

and then only job-level permissions need to be added where absolutely necessary.

For this particular issue, if `permissions:read-all` already includes `content: read`, then it still seems redundant to include `content:read` at the job-level.

.github/workflows/e2e.yml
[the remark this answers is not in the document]
sjinks: I have set the workflow permissions to `contents: read`. This seems to be more restrictive than `permissions: read-all`, because setting only one permission [behaves as if all the other permissions were set to `none`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions).

.github/workflows/e2e.yml
```diff
@@ -19,20 +19,58 @@ on:
...
   Run_e2e_tests:
+    name: Run E2E Tests
+    permissions:
+      contents: read
     runs-on: ubuntu-latest
     steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@14dc64f30986eaa2ad2dddcec073f5aab18e5a24 # v1
+        with:
+          egress-policy: block
```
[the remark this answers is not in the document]
sjinks: Yeah, I left it here to be able to switch between `block` and `audit` easily — this was very useful for debugging.

.github/workflows/e2e.yml
sjinks: @GaryJones [`contents: read` is more restrictive than `permissions: read-all`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions)

> Any permission that is absent from the list will be set to none.

That is,

```yaml
permissions:
  contents: read
```

is equivalent to

```yaml
permissions:
  contents: read
  actions: none
  checks: none
  deployments: none
  issues: none
  packages: none
  pull-requests: none
  repository-projects: none
  security-events: none
  statuses: none
```

sjinks: While `permissions: read-all` is equivalent to setting all those permissions above to `read`.

I am trying to set `contents: read` at the global level and see how it works — this seems to be a more restrictive setting.

.github/workflows/release-changelog.yml
```diff
@@ -19,33 +35,43 @@ jobs:
...
+      - uses: softprops/action-gh-release@6232f0b438cb856c39d14f8743e3a7c99fc879af # v0.1.14
         with:
           generate_release_notes: true
-          tag_name: ${{needs.get-release-tag.outputs.id}}
+          tag_name: ${{ steps.id-generator.outputs.id }}
+
   release-changelog:
+    name: Publish Changelog
+    permissions:
+      contents: read
```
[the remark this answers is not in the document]
sjinks: Removed in favor of `contents: read` in the workflow confioguration

.github/workflows/e2e.yml
GaryJones: > If you specify the access for any of these scopes, all of those that are not specified are set to none.

Ah - I think that's a key bit of information I was missing. Thank you for that!

sjinks: @varunsh-coder does it make sense to replace the top-level

```yaml
permissions: read-all
```

to

```yaml
permissions:
  contents: read
```
and then adjust the per-job permissions? This seems to be a more restrictive configuration than the original one and it ensures that all additional read or write permissions need to be requested explicitly.

sjinks:
OK, the global `permissions: { contents: read-all }` worked fine; this allowed us to remove the `permissions` option from the per-job configurations (these restrictive permissions are [inherited from the workflow level](https://github.com/Automattic/vip-go-mu-plugins/runs/4856079202?check_suite_focus=true#step:1:13)).

.github/workflows/e2e.yml
varunsh-coder: Thanks @sjinks for the feedback. Let me discuss this and I will get back.

rinatkhaziev (approved):
This is great!

PR #2818, opened 2022-01-17
Search: Add --mode to validate-contents command

rebeccahum:
## Description
To assist with debugging, it'd be nice to have a `--mode` flag to the `wp vip-search health validate-contents` command to limit to a subset of inconsistencies.

This PR also updates the reporting term to use "mismatch" instead of "inconsistency" for posts that exist but have differing content.

Added --mode command options to validate-contents command.

## Checklist

## Steps to Test
1) Create a few posts
2) Delete one from the DB:
```
wp db cli
> DELETE FROM wp_posts WHERE id = <postid>
```
3) Modify the post_content for another in the DB:
```
wp db cli
> UPDATE wp_posts SET post_content = 'blahblahblah' WHERE id = <another-postid>
```
4) Run `wp vip-search health validate-contents --do_not_heal` and expect to see one marked as `extra_in_index` and another marked as `inconsistent`
5) Apply PR

For the `missing` input:
6) Run `wp vip-search health validate-contents --mode=missing --do_not_heal` and expect to see only the one marked as `extra_in_index` reported back

For the `--mismatch` flag:
7) Run `wp vip-search health validate-contents --mode=mismatch --do_not_heal` and expect to see only the one marked as `mismatch` reported back


rinatkhaziev:
I know it's not strictly the subject of the pr, but that's quite a few arguments now, what do you think about providing some examples, [as per WP-CLI docs](https://make.wordpress.org/cli/handbook/guides/commands-cookbook/#annotating-with-phpdoc)? On that note, I think the whole docblock is not conforming to what it should be.

search/includes/classes/class-health.php
```diff
@@ -602,9 +573,12 @@ public function remove_validate_content_lock() {
 		delete_transient( self::CONTENT_VALIDATION_LOCK_NAME );
 	}
 
-	public function validate_index_posts_content_batch( $indexable, $start_post_id, $next_batch_post_id, $inspect ) {
+	public function validate_index_posts_content_batch( $indexable, $start_post_id, $next_batch_post_id, $options ) {
 		global $wpdb;
 
+		$inspect = isset( $options['inspect'] );
+		$mode    = $options['mode'] ?? null;
```
rinatkhaziev, line 580: What do you think about making the validation a bit more explicit, like

```
$supported_modes = [
	'complete',
	'missing',
	'mismatch', 
];

$mode    = isset( $options['mode'] ) && in_array( $options['mode'], $supported_modes, true ) ? $options['mode'] : 'complete';
```

This would also allow simplifying the `$mode` checks further down the method.

rebeccahum: Is it necessary if the CLI already throws an error when you attempt to pass in values that is not `mismatch` or `missing`? Also, unsure if we need a `complete` mode since it's there to limit a subset of index inconsistencies.

rinatkhaziev: Good points

rinatkhaziev (approved):

rebeccahum:
r2230-stacks

PR #2823, opened 2022-01-19
Fixing Parse.ly 3.0 release

pauarge:
## Description

Fixing some errors we saw on Parse.ly 3.0 due to type errors. This PR makes some additional checks on parameters to ensure types are always consistent.

This PR changes files on the wp-parsely source code. We will back port them to the main plugin.

### Fixing Parse.ly 3.0 release

Fixes potential issues on wp-parsely 3.0.

## Checklist

## Steps to Test

1. Check out PR.
2. Activate Parse.ly on your site and check it renders correctly.

### DIFF
--- a/wp-parsely-3.0/src/class-parsely.php
+++ b/wp-parsely-3.0/src/class-parsely.php
@@ -186,15 +186,7 @@ class Parsely {
 
 		// Assign default values for LD+JSON
 		// TODO: Mapping of an install's post types to Parse.ly post types (namely page/post).
-		if ( is_int( $post ) ) {
-			$parsed_post = WP_Post::get_instance( $post );
-			if ( $parsed_post === false ) {
-				return;
-			}
-			$parsely_page = $this->construct_parsely_metadata( $parsely_options, $parsed_post );
-		} else {
-			$parsely_page = $this->construct_parsely_metadata( $parsely_options, $post );
-		}
+		$parsely_page = $this->construct_parsely_metadata( $parsely_options, get_post( $post ) );
 
 		// Something went wrong - abort.
 		if ( empty( $parsely_page ) || ! isset( $parsely_page['headline'] ) ) {
--- a/wp-parsely-3.0/src/class-parsely.php
+++ b/wp-parsely-3.0/src/class-parsely.php
@@ -730,7 +730,7 @@ class Parsely {
 				$term_name = $this->get_bottom_level_term( $post_obj->ID, $parsely_options['custom_taxonomy_section'] );
 			}
 
-			if ( $term_name !== false ) {
+			if ( is_string( $term_name ) ) {
 				$category = $term_name;
 			}
 		}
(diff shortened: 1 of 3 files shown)


wp-parsely-3.0/src/class-parsely.php
```diff
@@ -186,7 +186,7 @@ public function insert_page_header_metadata(): void {
 
 		// Assign default values for LD+JSON
 		// TODO: Mapping of an install's post types to Parse.ly post types (namely page/post).
-		$parsely_page = $this->construct_parsely_metadata( $parsely_options, $post );
+		$parsely_page = $this->construct_parsely_metadata( $parsely_options, get_post( $post ) );
```
jblz, line 189: Another possibility that might be more robust, revert this change:

```suggestion
		$parsely_page = $this->construct_parsely_metadata( $parsely_options, $post );
```
...& up on L#180, something like:

```
$post = get_post( $GLOBALS['post'] );
if ( ! $post instanceof WP_Post ) {
    return;
}
```

sjinks (dismissed):

acicovic (approved):

wp-parsely-3.0/src/class-parsely.php
jblz: Yeah, definitely don't override the global :)

jblz (changes requested):
It's deprecated, but we should probably hit `insert_parsely_page`, while we're at it:

https://github.com/Parsely/wp-parsely/blob/4121bb9ebc35a1523e12648e20f09e7008fe875e/src/class-parsely.php#L236-L253

sjinks (dismissed):

jblz (approved):

wp-parsely-3.0/src/class-parsely.php
```diff
@@ -763,12 +774,17 @@ private function get_top_level_term( int $term_id, string $taxonomy_name ) {
...
 	 */
-	private function get_bottom_level_term( int $post_id, string $taxonomy_name ): string {
-		$terms    = get_the_terms( $post_id, $taxonomy_name );
-		$term_ids = is_array( $terms ) ? wp_list_pluck( $terms, 'term_id' ) : array();
-		$parents  = is_array( $terms ) ? array_filter( wp_list_pluck( $terms, 'parent' ) ) : array();
+	private function get_bottom_level_term( int $post_id, string $taxonomy_name ) {
+		$terms = get_the_terms( $post_id, $taxonomy_name );
+
+		if ( ! is_array( $terms ) ) {
+			return false;
```
GaryJones, line 783: Could we at least keep the type consistency, and return an empty string? What advantage is there to returning boolean `false`?

pauarge: Actually we want to do this to keep consistency with `get_top_level_term` and the check on line 736.

GaryJones: The check on line 736 can check for string length instead, or a `=== ''`.

sjinks (approved):

sjinks:
r2214-stacks

sjinks:
v20220119.2

PR #2827, opened 2022-01-19
Fix wp-parsely Type Errors for the 3.0 tree

pauarge:
## Description

Fixing some errors we saw on Parse.ly 3.0 due to type errors. This PR makes some additional checks on parameters to ensure types are always consistent.

This PR changes files on the wp-parsely source code. We will back port them to the main plugin.

This pulls in previous fixes in #2823 & #2824 (which got reverted in #2826 & #2825, respectively) & also adds  a couple of additional fixes to guard against type errors.

## Changelog Description

Fix type errors on Parse.ly 3.0
Fix two issues when getting metadata in certain circumstances.

## Checklist

Please make sure the items below have been covered before requesting a review:

- [x] This change works and has been tested locally (or has an appropriate fallback).
- [x] This change works and has been tested on a Go sandbox.
- [ ] This change has relevant unit tests (if applicable).
- [ ] This change has relevant documentation additions / updates (if applicable).
- [x] I've created a changelog description that aligns with the provided examples.

### Confirm the taxonomy metadata issue

* Run the main branch.
* Add the following to an mu-plugin:

```

add_filter( 'get_term', function ( $term, $taxonomy ) {
        if ( 'category' === $taxonomy && 'uncategorized' === $term->slug ) {
                $term->name = null;
        }
        return $term;
}, 10, 2 );
```

* Browse to `/category/uncategorized/`
* Confirm fatal error

### Confirm the taxonomy metadata fix

* Switch to this branch (leaving the `get_term` filter in place)
* Browse to `/category/uncategorized/` while logged out of the site.
* Confirm no fatal error is shown

### Confirm the non-array metadata issue

* Run the main branch
* Add the following to an mu-plugin: `add_filter( 'wp_parsely_metadata', '__return_false' );`
* Browse to any front end post or page while logged out of the site.
* Confirm fatal error

### Confirm the non-array metadata fix

* Switch to this branch (leaving the `wp_parsely_metadata` filter in place)
* Browse to any front end post or page while logged out of the site.
* Confirm no fatal error is shown & metadata are correctly not present in the page source

PR #2828, opened 2022-01-19
Revert "Fix wp-parsely Type Errors for the 3.0 tree"

jblz:
Reverts Automattic/vip-go-mu-plugins#2827


jblz:
not currently needed

PR #2833, opened 2022-01-24
Split CI jobs between GitHub and CircleCI

sjinks:
Recently, we often hit the limit of concurrent jobs in CircleCI. This PR moves some of those jobs into GitHub so that CircleCI runs only the required checks and GitHub runs the rest.

This allows us to reduce the load on CircleCI and spend less time waiting until CircleCI has a free slot for the queued checks.


jesse-greathouse (approved):
Looks alright.

PR #2847, opened 2022-01-27
Add E2E tests for Search Dev Tools

sjinks:


__tests__/e2e/bin/setup-env.sh
```diff
@@ -17,11 +17,17 @@ done
...
 # Install specified version of WordPress
-docker exec e2etestsite_php_1 wp core update --allow-root --version=$version --force
\ No newline at end of file
+docker exec e2etestsite_php_1 wp core update --allow-root --version="${version}" --force
+
+# Enable Enterprise Search
+docker exec e2etestsite_wordpress_1 sh -c 'echo "define( \"VIP_ENABLE_VIP_SEARCH\", true );" | tee -a /app/config/wp-config.php'
+docker exec e2etestsite_wordpress_1 sh -c 'echo "define( \"VIP_ENABLE_VIP_SEARCH_QUERY_INTEGRATION\", true );" | tee -a /app/config/wp-config.php'
+docker exec e2etestsite_php_1 wp --allow-root vip-search index --setup --skip-confirm
+docker exec e2etestsite_php_1 wp --allow-root vip-search index
```
rinatkhaziev, line 33:
```suggestion
```
This line is not needed.

rinatkhaziev (commented):
I left a small note about a redundant line, let's see what @bsessions85 says about the rest.

bsessions85:
Hi @sjinks! I love that you're writing e2e tests! The general guidance with e2e tests using this type of Page Object Model is to keep all of the expectations and assertions in the spec file and to limit the page file to just functions and utilities that will help you perform the tasks needed to test it. I would move all the `expect` calls in to the spec file and then change the methods in the page file from things like `ensureDevToolsOpen()` to `isDevToolsOpen` and return it as a boolean to the spec file where you can make sure it is in the state you're expecting. 

I did notice that when I ran them locally the test is failing. I checked the CI run and it is marked as "flaky", meaning it failed on the first run and passed on the second run. You'll want to watch that 1 sec timeout in `runQuery`

Reach out  if I can help in any way!

bsessions85:
Also, adding in some JSdocs would be nice on the page and helper files

sjinks:
@bsessions85 I have updated the code. Regarding POM — I saw that in the official Playwright docs [they use assertions](https://playwright.dev/docs/test-pom), and I thought that was OK :-)

bsessions85:
@sjinks  Yeah, I have seen them do other things that go against general guidance. Always pushing boundaries! I find it easier to debug when the assertion is right there at the beginning. Thanks for making the updates. Great work on this!

bsessions85 (approved):
Looks great, thanks!

PR #2848, opened 2022-01-27
Do not contact wordpress.org for translations

sjinks:
This is one of the reasons for transient failures ([example](https://github.com/Automattic/vip-go-mu-plugins/runs/4970502794?check_suite_focus=true)).

We should not make external requests in our tests unless absolutely necessary.


rinatkhaziev (approved):

WPprodigy (approved):
Seems logical. IDK what the response structure is supposed to be looking at that filter's function, but I'm assuming you've done the research on that :)

sjinks:
@WPprodigy [yes](https://github.com/WordPress/WordPress/blob/5.9/wp-admin/includes/translation-install.php#L135) :-)

sjinks:
WP bug report: https://core.trac.wordpress.org/ticket/54959

PR #2854, opened 2022-01-28
Performance: Disable CAP's join filter

rebeccahum:
## Description
More details at https://lobby.vip.wordpress.com/2017/11/07/co-authors-plus-global-filter/.

### Plugin Updated: VIP Performance

Enable global CAP taxonomy filter for better performance.

## Checklist


sjinks (approved):

performance/plugins.php
```diff
@@ -9,3 +9,13 @@
 add_action( 'amp_extract_image_dimensions_callbacks_registered', function() {
 	remove_filter( 'amp_extract_image_dimensions', array( 'AMP_Image_Dimension_Extractor', 'extract_from_attachment_metadata' ) );
 } );
+
+// CAP: By default, co-authors will query in a sort of compatibility mode.
+// Making the SQL queries work for posts with the old author style (as in the built-in post author
+// or by using the co-authors taxonomy.
```
GaryJones, line 15: The English doesn't make sense here. 

It's also missing a `)`.

I'd also rather see it as a function DocBlock for `vip_coauthors_plus_should_query_post_author` below rather than just a random multiline-but-styled-as-a-single-line comment.

rebeccahum: It was a copy and paste from the description in the WPCOM snippet. I've re-written it now, @GaryJones.

performance/plugins.php
```diff
@@ -3,3 +3,20 @@
...
+/**
+ * By default, the Co-Authors Plus plugin will query in a compatibility mode where it uses an expensive JOIN for post_author in it.
+ * This filter disables it and uses a simpler, taxonomy-only query.
+ * Run `wp co-authors-plus create-terms-for-posts` to ensure all authors have an author taxonomy attached to them.
+ * 
+ * @see https://lobby.vip.wordpress.com/2017/11/07/co-authors-plus-global-filter/
+ * 
+ * @return void
+ */
+function vip_coauthors_plus_should_query_post_author() {
```
GaryJones, line 16:
```suggestion
function vip_coauthors_plus_should_query_post_author(): void {
```

performance/plugins.php
```diff
@@ -3,3 +3,20 @@
 // Plugin-specific tweaks go in here
 
 namespace Automattic\VIP\Performance;
+
+/**
+ * By default, the Co-Authors Plus plugin will query in a compatibility mode where it uses an expensive JOIN for post_author in it.
+ * This filter disables it and uses a simpler, taxonomy-only query.
+ * Run `wp co-authors-plus create-terms-for-posts` to ensure all authors have an author taxonomy attached to them.
```
GaryJones, line 10: Separates a clear Summary (first line) from the Description (rest of the notes).

```suggestion
 * Disable an expensive term-counting query for Co-Authors Plus.
 *
 * By default, the Co-Authors Plus plugin will query in a compatibility mode where it uses an expensive JOIN for post_author in it.
 * This filter disables it and uses a simpler, taxonomy-only query.
 * Run `wp co-authors-plus create-terms-for-posts` to ensure all authors have an author taxonomy attached to them.
```

performance/plugins.php
```diff
@@ -3,3 +3,20 @@
...
+ * 
+ * @see https://lobby.vip.wordpress.com/2017/11/07/co-authors-plus-global-filter/
+ * 
+ * @return void
+ */
+function vip_coauthors_plus_should_query_post_author() {
+	if ( class_exists( 'CoAuthors_Plus' ) ) {
+		add_filter( 'coauthors_plus_should_query_post_author', '__return_false' );
```
GaryJones, line 18: Would this filter ever run if the `CoAuthors_Plus` class _didn't_ exist?

rebeccahum: No, why should it?

WPprodigy: Ah I think he means it could drop the `class_exists( 'CoAuthors_Plus' )` check, and the whole wrapper even I suppose?

Guess the overhead for non-CAP sites would be the same, either `vip_coauthors_plus_should_query_post_author` is the extra filter being registered or `coauthors_plus_should_query_post_author` is.

rebeccahum: Ah, got it. I'm not great at reading between the lines 🙂

jesse-greathouse (approved):
LGTM

rebeccahum:
Eh, it's in the readme. I'm not sure if it's worth breaking for users who haven't run the command, especially on smaller-scale sites where the query isn't a huge deal.

PR #2872, opened 2022-02-02
Search: Re-factor build_new_index() to not use WP_CLI::runcommand()

rebeccahum:
## Description
Due to limitations with PHP-fpm requiring `WP_CLI::runcommand()` to set the 'launch' flag to `false`, I've decided to go about another way since it wouldn't capture any STDERR. This is continuation of the work from #2776.

## Checklist

## Steps to Test
To test with the post Indexable:
1) `$es = new \Automattic\VIP\Search\Search();`
2) `$es->init();`
3) `$hj = new \Automattic\VIP\Search\SettingsHealthJob( $es );`
4) `$hj->build_new_index( 'post' );`

### DIFF
--- a/search/includes/classes/class-settingshealthjob.php
+++ b/search/includes/classes/class-settingshealthjob.php
@@ -239,8 +239,7 @@ class SettingsHealthJob {
 		}
 
 		// Do not schedule new index build if index version limit is reached (2 versions per Indexable).
-		$search           = \Automattic\VIP\Search\Search::instance();
-		$current_versions = $search->versioning->get_versions( $indexable );
+		$current_versions = $this->search->versioning->get_versions( $indexable );
 		if ( count( $current_versions ) > 1 ) {
 			$message = sprintf(
 				'Cannot automatically build new %s index on %s to meet shard requirements. Please ensure there is less than 2 index versions.',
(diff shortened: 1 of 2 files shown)


search/includes/classes/class-settingshealthjob.php
```diff
@@ -263,27 +262,69 @@ public function maybe_schedule_build_new_index( $indexable ) {
 	public function build_new_index( $indexable_slug ) {
 		update_option( self::BUILD_LOCK_NAME, time() ); // Set lock for starting rebuild.
```
rinatkhaziev, line 263: Do we need that option to be autoloaded?

rebeccahum: no

search/includes/classes/class-settingshealthjob.php
```diff
@@ -250,7 +249,7 @@ public function maybe_schedule_build_new_index( $indexable ) {
 			$this->send_alert( '#vip-go-es-alerts', $message, 2 );
 			
 			return;
-		} elseif ( ! wp_next_scheduled( self::CRON_EVENT_BUILD_NAME, [ $indexable ] ) ) {
+		} elseif ( ! wp_next_scheduled( self::CRON_EVENT_BUILD_NAME, [ $indexable->slug ] ) ) {
```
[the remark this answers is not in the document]
rebeccahum: Yes, it is intentional. It was an oversight on my part earlier in #2795, since I changed it to pass in the slug to the cron job so we didn't have to worry about any deserialization/serialization bugs.

rinatkhaziev (commented):
Thanks, looks good but I left an inline comment about the argument for the `wp_next_scheduled` check.

Also, what do you think about adding the blog id everywhere in the error messages, will it be helpful for us or the URL is sufficient?

rebeccahum:
The URL is sufficient IMHO, because if something went wrong, I'm probably going to want to sandbox and do a `wp site list` anyways before debugging.

rebeccahum:
r2261-stacks

PR #2873, opened 2022-02-03
Search: Remove logging for 401

rebeccahum:
## Description
Let's remove the logging we added for 401s in #2719 since looking at the logs, it hasn't occurred in the last 2.5 months.


sjinks (approved):

rebeccahum:
r2258-stacks

PR #2886, opened 2022-02-08
CANTINA-855 - Account for all indexable post types and statuses when determining the number of shards for index

rebeccahum:
## Description
In `filter__ep_default_index_number_of_shards()`, we assume the only indexed post types are `post` with a post status of `publish`, which is incorrect. The default is still set at 1, but we will use the hooks `ep_post_mapping` and `ep_user_mapping` to determine if it needs an increase.

Account for all indexable post types and statuses when determining the number of shards to index

## Checklist

## Steps to Test
1) Create a large site with only one 'post' and over a million of a public CPT
2) Add a new index version and check that the shards are set at 4


rinatkhaziev (approved):
Awesome! :shipit:

rebeccahum:
r2302-stacks

sjinks:
v20220222.0

PR #2887, opened 2022-02-08
Feature: Add enabling or disabling by ID

rebeccahum:
## Description
It would be nice to be able to selectively enable or disable a feature by an array of ids. This PR also gets rid of the `site_id()` method since we can use the Constant_Mocker class now.

## Checklist

## Steps to Test
1) Sandbox into a site
2) Edit in `$feature_ids` with `[ 'feature' => [ <your-site-id> => true ] ]`
3) shell in and do `\Automattic\VIP\Feature::is_enabled_by_ids( 'feature' )` and expect it to return true
4) Repeat step 2 but replace `<your-site-id>` with random numbers
5) shell in and do `\Automattic\VIP\Feature::is_enabled_by_ids( 'feature' )` and expect it to return false


rinatkhaziev (changes requested):
Thanks!

I've got two things:
1) I'm not sure I understand the intent of getting rid of the setter and using the constant instead, 
2) The enabled/disabled approach looks clever but confusing. If we stuff IDs in a single array it will be hard to discern what should be disabled and what should be enabled. Thus, increasing the chances of someone accidentally enabling something that shouldn't be enabled and vice versa.

rebeccahum:
@rinatkhaziev 1) My understanding was that the setter was more in place for testing purposes — I don't really see the purpose in encapsulating the constant that's not tied to the class?
2) OK, fair enough. I'm thinking we can do something like this then:
```
$feature_ids = [ 'feature-flag' => [ 123 => true, 345 => true, 567 => false ] ];
```
So in this instance, anything with a value of true would be enabled and anything with a value of false would be disabled. We can also use pass it in the rollout as a bool whether we want to use the $feature_ids as well.

rinatkhaziev (approved):

rebeccahum:
r2274-stacks

jesse-greathouse:
v20220210.0

PR #2890, opened 2022-02-10
Search: Make log message more filterable for successful deletes

rebeccahum:
## Description
It's hard to filter through the Kibana logs for `search_versioning` since the messages can differ.


rebeccahum:
r2281-stacks

PR #2895, opened 2022-02-11
Do not use Jetpack Photon when Jetpack disabled

emrikol:
When a site is disabling Jetpack according to [our documentation](https://docs.wpvip.com/technical-references/restricting-site-access/controlling-content-distribution-via-jetpack/#h-preventing-jetpack-from-loading), it will cause `A8C_Files::init_jetpack_photon_filters()` to continually throw the warning `Cannot initialize Photon filters as the Jetpack_Photon class is not loaded. Please verify that Jetpack is loaded and active to restore this functionality.`

This will skip attempting to use Jetpack Photon without having to have the site add an additional constant  `define( 'WPCOM_VIP_USE_JETPACK_PHOTON', false )`

## Description

When Jetpack is completely disabled on a site via the `VIP_JETPACK_SKIP_LOAD` we should also skip attempting to load Jetpack Photon.

### Plugin Updated: Automattic File Hosting Service

Fixed a bug where disabling Jetpack does not disable attempted Jetpack Photon loading.

## Checklist


rebeccahum:
I was able to easily reproduce by the following steps:

1) Disable JP via `define( 'VIP_JETPACK_SKIP_LOAD', true );`
2) Access a url with the params `?jetpack-photon=yes`

Marking this PR as ready for review.

rebeccahum (approved):
Thanks!

rebeccahum:
r2290-stacks

PR #2897, opened 2022-02-11
Jetpack: Return on no object for $jetpack_search in debug bar

rebeccahum:
## Description
To prevent the below fatal from occurring:

```
PHP Fatal error:  Uncaught Error: Call to a member function get_last_query_info() on null
```


rinatkhaziev (approved):

rebeccahum:
r2268-stacks

PR #2909, opened 2022-02-14
Search: Ensure option in cache is deleted after we invalidate it

rebeccahum:
## Description
We're seeing that sometimes when we delete an option, the option still remains floating in the cache. This adds a check to ensure that the option cache actually gets deleted after we delete the option.

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -822,6 +822,11 @@ class Search {
 				} else {
 					// Invalidate index_exists caching on certain actions.
 					delete_option( $index_exists_option_name );
+
+					// Ensure the cache for the option was actually deleted.
+					if ( false !== wp_cache_get( $index_exists_option_name, 'options' ) ) {
+						wp_cache_delete( $index_exists_option_name, 'options' );
+					}
 				}
 			}
 		}


rebeccahum:
r2289-stacks

WPprodigy:
The option is going into alloptions cache group, not the individual `options` cache (or at least it shouldn't be unless wp_installing() or something): https://github.com/Automattic/vip-go-mu-plugins/blob/c6bd8a186483357fb72d4f7722ca11f6378a5c75/search/includes/classes/class-search.php#L943

So I don't think this PR will have much effect, it's purging a cache that doesn't exist. It's more-so a problem that even though request A deletes the option (and the entry from the alloptions cache), a concurrent request B overwrites the alloptions cache a millisecond afterwards and the value ends up stuck there until alloptions is purged sometime.

If it's a big problem, could pull it out of the `alloptions` cache (autoload=false). The downside is that it'll take an additional query/cache get on pages where the option is retrieved, so generally speaking its ideal to leave things in alloptions that are fetched on every page load.

rebeccahum:
> The option is going into alloptions cache group, not the individual options cache

Oddly enough, we were seeing that it was in the individual `options` cache and once we deleted it, it resolved the issue at hand. How it got there? I'm not sure.

```
wp cache delete es_index_exists_vip-XXX-post-1 options
Success: Object deleted.
```

PR #2911, opened 2022-02-15
Use add_option over update_option for better concurrency handling

rebeccahum:
## Description
Part of the issue described in #2909, we were finding that the option was stored in the `options` group rather than `alloptions`. Let's use `add_option()` instead since it checks if the `$autoload` parameter is set before deciding which option group it's set at: https://github.com/WordPress/WordPress/blob/a4026420cc1845aec29661239e625c5d3cfee452/wp-includes/option.php#L645-L651

Compared to `update_option()` where it only checks the current state of alloptions: https://github.com/WordPress/WordPress/blob/a4026420cc1845aec29661239e625c5d3cfee452/wp-includes/option.php#L508-L514


sjinks (approved):

sjinks:
r2295-stacks

sjinks:
v20220216.0

PR #2914, opened 2022-02-16
Search: json_decode the body of the query before logging a failed request

rinatkhaziev:
## Description

Some logged errors result in error_type and root_cause values to be `null` when in reality we have the data.

Logging improvements.

## Checklist

## Steps to Test

Tricky to test manually, need to force a bad query so that request is successful but the resulting response contains the error from ES on a non-sandbox.


search/includes/classes/class-search.php
```diff
@@ -1062,12 +1062,19 @@ public function ep_handle_failed_request( $request, $response, $query, $statsd_p
...
+	 * Or format parts of the query object to be more readable.
 	 */
 	public function sanitize_ep_query_for_logging( $query ) {
 		if ( isset( $query['args']['headers']['Authorization'] ) ) {
 			$query['args']['headers']['Authorization'] = '<redacted>';
 		}
 
+		// Try to parse the body if possible to make it better readable in the log entry.
+		if ( isset( $query['args']['body'] ) ) {
+			$decoded_body          = json_decode( $query['args']['body'] );
```
jesse-greathouse, line 1074: nothing wrong with this, but my preference is to force the [associative flag](https://www.php.net/manual/en/function.json-decode.php) here, to be absolutely sure that the product is always an associative array and not a generic object.

rinatkhaziev: @jesse-greathouse thanks for the comment. Generally speaking, I would agree with you but in this case, it doesn't really matter much as it's used right on the next line and it'll get json_encode'd again when logging and would become an object.

sjinks (approved):

sjinks:
r2312-stacks

sjinks:
v20220222.0

PR #2917, opened 2022-02-17
Don't typecast $user object for wp_login callback

sjinks:
Moving Automattic/vip-support#62 into mu-plugins repo.

### DIFF
--- a/vip-support/class-vip-support-user.php
+++ b/vip-support/class-vip-support-user.php
@@ -614,7 +614,13 @@ class User {
 	 *
 	 * @return void
 	 */
-	public function action_wp_login( $_user_login, WP_User $user ) {
+	public function action_wp_login( $_user_login, $user ) {
+		if ( ! ( $user instanceof WP_User ) ) {
+			// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped -- gettype() is safe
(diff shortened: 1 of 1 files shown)


GaryJones:
From the original issue:

> Some manually called instances may incorrectly pass in a non-WP_User object which results in fatals

Is that a good thing?
The rest of the method only seems to use `$user->ID`, so it at least needs to be an object, but the type hint was protecting against `WP_Post` being passed in, for instance.

I'm not sure how the new approach in this PR really helps engineers - it just hides the issue which will give harder-to-debug runtime logical issues.

sjinks:
@GaryJones I added an `instanceof` check here. In debug mode, it will spit E_USER_NOTICE (or what `_doing_it_wrong()` does); in production mode, it will just return early.

We already had a similar issue (#2689) with another plugin in another place. It looks like it is safer to do the type checking inside a hook rather than crash the site because of a misbehaving plugin.

GaryJones:
> It looks like it is safer to do the type checking inside a hook rather than crash the site because of a misbehaving plugin.

On the assumption that customers are checking in non-production first, I'd rather it threw a fatal there (non-prod) from the type declaration than just a `E_USER_NOTICE` (not as obvious as a fatal), so this can be addressed well before the code makes it to production.

rinatkhaziev (approved):

sjinks:
r2304-stacks

sjinks:
v20220222.0

PR #2918, opened 2022-02-17
Disable admin email check for VIP Support users

sjinks:
Cloned from Automattic/vip-support#88

Differences from the original PR:
  - use `__return_zero` instead of a custom function returning zero in `disable_admin_email_check()`


sjinks:
cc @WPprodigy

rebeccahum (approved):

sjinks:
r2303-stacks

sjinks:
v20220222.0

PR #2924, opened 2022-02-18
Set retry window to 10 and threshold for failing at 5 for Search requests

rinatkhaziev:
## Description

We previously had it set that after 3 failed requests any subsequent requests would be blocked. This is not great because with a higher rate of requests we essentially block requests even though the blip could be momentary. This PR reduces the retry window to 10s (minimum allowed value) and cautiously sets the threshold to 5. It can probably be bumped up, but I'd like to err on the side of caution.

We have tweaked thresholds that control early server request termination in case of repeated request errors.

## Checklist


rebeccahum:
r2311-stacks

sjinks:
v20220222.0

PR #2931, opened 2022-02-22
Increase bulk edit limit to 100 posts

WPprodigy:
## Description

Alternative approach to https://github.com/Automattic/vip-go-mu-plugins/pull/2681. After various testing, 100 seems to be a safe number where bulk edits will consistently work. A bit higher than that and you risk the edit failing mid-way through, much higher and it's guaranteed to fail on even the simplest of sites.

I did explore just limiting the screen options value ([more info](https://github.com/Automattic/vip-go-mu-plugins/pull/2681#issuecomment-990149848)), and while it seems like slightly better UX, this ends up being a bit more limiting as it also restricts querying/loading up to 999 posts which works fine (is just the edit/updating that fails).

### WP Admin: Bulk editing limit increased from 40 to 100.

Can now bulk edit up to 100 posts by default through the bulk-editing UI in the WordPress admin dashboard.

## Checklist

## Steps to Test

- Add 200 posts & pages to a test site.
- Edit screen options to view 101 on each post type admin list, ensure it restricts bulk editing.
- Change to 100 and it should allow bulk editing again.


sjinks (approved):

rebeccahum:
r2323-stacks

PR #2935, opened 2022-02-22
Search: Add FILES_CLIENT_SITE_ID to alerting

rebeccahum:
## Description
Sometimes data syncs can lead to weirdness with homeurls, so we should also log the application ID to avoid confusion in our alerting.


sjinks (approved):

rebeccahum:
r2317-stacks

PR #2942, opened 2022-02-24
Ignore Parse.ly WINDOWS file on deploys

pauarge:
## Description

Adding the WINDOWS file to deploy ignore so it doesn't get included.


GaryJones (approved):
✅

PR #2943, opened 2022-02-24
Search: Fix updated variable in case it was affecting the cron job

rebeccahum:
## Description
Updated variable to `$delete_version` but it still checks if `$versioning` is a WP_Error. This could potentially cause the cron job to silently fail if there is a notice being thrown.


rinatkhaziev (approved):

rebeccahum:
20220301.0

PR #2959, opened 2022-03-03
Search: Add temporary command clean-ep-feature-settings before we backfill ep_feature_settings

rebeccahum:
## Description
Just in-case we have rogue ep_feature_settings option on multisites with the `EP_IS_NETWORK` constant is defined, let's add a command to delete it.

## Checklist

Please make sure the items below have been covered before requesting a review:

- [x] This change works and has been tested locally (or has an appropriate fallback).
- [ ] This change works and has been tested on a Go sandbox.
- [ ] This change has relevant unit tests (if applicable).
- [ ] This change has relevant documentation additions / updates (if applicable).
- [ ] I've created a changelog description that aligns with the provided examples.

## Steps to Test
1) `wp option update ep_feature_settings test`
2) `wp vip-search clean-ep-feature-settings`

PR #2958, opened 2022-03-02
Search: Update variable name in vip_maybe_backfill_ep_option()

rebeccahum:
## Description
Bugfix: Change the variable name to match the check

PR #2957, opened 2022-03-02
Revert PR #2956

rebeccahum:
This PR reverts https://github.com/Automattic/vip-go-mu-plugins/pull/2956

PR #2956, opened 2022-03-02
Search: Bring in EP to cfa08368e168f47ed537c833c288d17b069be889

rebeccahum:
## Description
This pulls in EP to include the following PRs:
- https://github.com/Automattic/ElasticPress/pull/151
- https://github.com/Automattic/ElasticPress/pull/149

PR #2960, opened 2022-03-03
Search: Bring in EP to cfa08368e168f47ed537c833c288d17b069be889

rebeccahum:
## Description
Previously done in #2956 and reverted in #2957. Now that the bugfix #2958 is in, we'll be OK. But first, we must run #2959 on all multisites.


jesse-greathouse (approved):
👍

rebeccahum:
r2346-stacks

PR #2961, opened 2022-03-03
Remove wp-parsely 3.0

pauarge:
## Description

We no longer have users of Parse.ly 3.0 on the platform, so it is time for it to go away.

Tests have also been updated to test against 3.1 instead of 3.0.

### Plugin Removed: Parse.ly 3.0

We have removed the 3.0 version of Parse.ly from the platform. The recommended version is the latest, 3.1.

## Checklist


GaryJones:
> We no longer have users of Parse.ly 3.0 on the platform

How was this determined?

pauarge:
> How was this determined?

We have used our internal tools to check for plugins and their versions. All the sites that have wp-parsely version 3.0.x, they have installed manually. We have also went through the client code repositories searching for the filter that defines a version, and we haven't found any sites using the filter with value `3.0`.

rebeccahum (approved):

rebeccahum:
r2345-stacks

PR #2972, opened 2022-03-07
SDI: Index wporg plugin slugs

WPprodigy:
## Description

This PR adds the WordPress.org plugin slug to the plugin data for the Sites Details Index.

The process for finding the proper `slug` of a WordPress.org plugin is pretty complicated, as you can't guarantee that the folder/file path is an accurate representation. The only way to know for sure is to use the same process that WP core uses when checking for plugin updates. So that's what we are doing here.

### Unit Testing Changes

Cleaned up the unit test a bit as well, removing a bit of duplicate logic. Testing the output of the `vip_site_details_index_data` filter is sufficient, don't really need to do the exact same checks for one of the functions that are hooked onto the filter. And then rather than testing if the key exists, is of the right type, and has the right value - can just use `assertSame()` which has type safety and will of course throw an error if the index doesn't exist in the array.

For the new logic being tested, I'm mocking the `update_plugins` site transient to both prevent external calls & ensure we are testing all possible scenarios.

### Site Details Index: Index the plugin slug for WordPress.org plugins.

Indexes the proper `slug` for WordPress.org-listed plugins.

## Checklist

## Steps to Test

Is a pretty complicated setup for a full end to end test for this, but the unit tests here are pretty comprehensive for the feature being added.

### DIFF
--- a/config/class-site-details-index.php
+++ b/config/class-site-details-index.php
@@ -89,36 +89,65 @@ class Site_Details_Index {
 	 * Does not contain anything from vip-go-mu-plugins.
 	 */
 	public function get_plugin_info() {
-		// Needed or get_plugins can't be found in some instances
+		// Ensure get_plugins() & wp_update_plugins() are available.
 		require_once ABSPATH . 'wp-admin/includes/plugin.php';
+		require_once ABSPATH . 'wp-admin/includes/update.php';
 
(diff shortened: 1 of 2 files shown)


rebeccahum (approved):
LGTM.

PR #2974, opened 2022-03-08
Prevent local dev-env fatals during SDI syncs

WPprodigy:
## Description

Calling `$this->alerts->send_to_chat()` will result in a fatal error on local dev-envs because `$this->alerts` could have be initiated as a WP_Error(), [seen here](https://github.com/Automattic/vip-go-mu-plugins/blob/3498a07756cc0296458a6b4e33da14b5c76e5479/lib/utils/class-alerts.php#L277-L303). Notably this makes testing the SDI syncs locally pretty tough.

The alerts class does provide a [helper method](https://github.com/Automattic/vip-go-mu-plugins/blob/3498a07756cc0296458a6b4e33da14b5c76e5479/lib/utils/class-alerts.php#L352-L363) that protects against this, but you can't test/mock statics so some of the existing tests end up failing.

Ultimately I think maybe the Alerts class could use some refactoring so that it always works peacefully even if not hooked up all the way (but stays mockable). This PR felt like the best answer for the short term as it doesn't make any wide-arcing decisions and instead just works to prevent some fatals.

## Changelog Description
n/a

## Checklist

## Steps to Test

Run `do_action( 'vip_config_sync_cron' );` in a local dev-env.

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -795,7 +795,7 @@ class Search {
 
 	/**
 	 * Generate option name for cached index_exists request.
-	 * 
+	 *
 	 * @return string $option_name Name of generated option.
 	 */
 	private function get_index_exists_option_name( $url ) {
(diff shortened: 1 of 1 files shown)


rebeccahum (approved):

rebeccahum:
r2356-stacks

PR #2975, opened 2022-03-08
ci: update allowed endpoints in e2e.yml workflow

varunsh-coder:
## Description
This PR updates the allowed-endpoints set in step-security/harden-runner for the e2e.yml workflow. 
https://github.com/Automattic/vip-go-mu-plugins/blob/master/.github/workflows/e2e.yml#L31

I am the maintainer for [step-security/harden-runner](https://github.com/step-security/harden-runner) and monitor if any new endpoints get called. In this case, new outbound calls were being made to `update.traefik.io` and `update.containous.cloud` (redirect from `update.traefik.io`). So, adding these two endpoints. I looked up the reason, and it is just traefik image trying to see if new version is available - [code here](https://github.com/traefik/traefik/blob/cb4fb973b281ac5b4be743bfafde710145bc3f36/pkg/version/version.go#L72). 

I have run this change on a fork, and the new endpoints are no longer being blocked. 
https://app.stepsecurity.io/github/varunsh-coder/vip-go-mu-plugins-1/actions/runs/1949213318

Note: the e2e.yml workflow is failing. Not sure why. But it is not related to these endpoints. I had run `harden-runner` in audit mode and the test cases were still failing. 

Please let me know if you have any questions about this. Thanks!


sjinks (approved):

sjinks:
Thank you!

PR #2978, opened 2022-03-08
Search: Make error message more explicit in build_new_index()

rebeccahum:
## Description
Let's make the alert more explicit when re-building fails + some whitespacing changes.


jesse-greathouse (approved):
^ ditto

rebeccahum:
r2361-stacks

PR #2954, opened 2022-03-01
Search: Auto re-building indexes account for pods rolling over

rebeccahum:
## Description
There is an auto-rebuilding of indexes job to meet shard requirements. However, this didn't account for interruptions in the indexing process. This PR now takes it into account and depending on the state of the index rebuilding, it can either
1) Start the process of rebuilding an index afresh
2) Continue indexing from where it left off at
3) Hot-swap the indexes after building has completed

Currently, it's only enabled for non-production.

## Checklist

Please make sure the items below have been covered before requesting a review:

- [x] This change works and has been tested locally (or has an appropriate fallback).
- [ ] This change works and has been tested on a Go sandbox.
- [ ] This change has relevant unit tests (if applicable).
- [ ] This change has relevant documentation additions / updates (if applicable).
- [ ] I've created a changelog description that aligns with the provided examples.

### Scenario: `in-progress`
1) Call the indexing cron command directly:
```
$es = new \Automattic\VIP\Search\Search();
$es->init();
$indexable = \ElasticPress\Indexables::factory()->get( 'post' );
$hj = new \Automattic\VIP\Search\SettingsHealthJob( $es );
$hj->build_new_index( 'post' );
```
2) Attempt to kick off the cron job concurrently:
```
$es = new \Automattic\VIP\Search\Search();
$es->init();
$indexable = \ElasticPress\Indexables::factory()->get( 'post' );
$hj = new \Automattic\VIP\Search\SettingsHealthJob( $es );
$hj->maybe_process_build();
```
3) Check to see that one has not been scheduled `wp cron event list`

### Scenario: `resume`
1) Call the indexing cron command directly:
```
$es = new \Automattic\VIP\Search\Search();
$es->init();
$indexable = \ElasticPress\Indexables::factory()->get( 'post' );
$hj = new \Automattic\VIP\Search\SettingsHealthJob( $es );
$hj->build_new_index( 'post' );
```
2) Kill it off half-way
3) Check to ensure that the last processed ID was stored in options: `wp option get vip_search_cron_last_processed_id`
4) Delete the `vip_search_cron_new_version_building_now` transient: `wp delete transient vip_search_cron_new_version_building_now`
5) Kick off a cron job:
```
$es = new \Automattic\VIP\Search\Search();
$es->init();
$indexable = \ElasticPress\Indexables::factory()->get( 'post' );
$hj = new \Automattic\VIP\Search\SettingsHealthJob( $es );
$hj->maybe_process_build();
```
[...]

PR #2979, opened 2022-03-08
Search: Update add_action to account for build_new_index taking two args

rebeccahum:
## Description
Re-factored in #2954, I didn't update the `add_action()` call to account for the fact that `build_new_index()` can take two arguments now.


jesse-greathouse (approved):

rebeccahum:
r2363-stacks

PR #2981, opened 2022-03-09
e2e- Fix e2e failures related to themes update

bsessions85:
## Description

The default theme for WP changed and the selectors for some things we're testing changed as well.

This PR updates those selectors to get the tests passing again.

## Checklist

## Steps to Test

Make sure e2e tests pass in CI, with the exception of the Search test which is failing for unrelated issues

### DIFF
--- a/__tests__/e2e/lib/pages/published-page-page.ts
+++ b/__tests__/e2e/lib/pages/published-page-page.ts
@@ -4,7 +4,7 @@
 import { Page } from 'playwright';
 
 const selectors = {
-    entryTitle: '.entry-title',
+    entryTitle: '.wp-block-post-title',
     pageImage: '.entry-content img',
     pageText: ( text: string ) => `text=${ text }`,
 };
--- a/__tests__/e2e/lib/pages/published-post-page.ts
+++ b/__tests__/e2e/lib/pages/published-post-page.ts
@@ -4,7 +4,7 @@
 import { Page } from 'playwright';
 
 const selectors = {
-    entryTitle: '.entry-title',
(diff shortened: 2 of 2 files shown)


jesse-greathouse (approved):
If I understand this correctly, the values in the selectors need to change because the default template for WordPress changed.

I suppose the next logical question would be:
Would it be better for us to test against an environment with a contrived theme that will not change? I sense that the control in this experiment is dynamic.

bsessions85:
> If I understand this correctly, the values in the selectors need to change because the default theme for WordPress changed.
> 
> I suppose the next logical question would be: Would it be better for us to test against an environment with a contrived theme that will not change? I sense that the control in this experiment is dynamic.

Yeah, that is a good point. I'll look in to that now.

GaryJones:
If something else in WP Core changed (like a selector in the media uploader), then you'd have to make changes anyway.

In this case, it looks like only two strings are needing to be changed, and that's because the default theme is now one that supports full-site editing; the now-replaced `entry-title` has been around for many many years - see [this](https://themes.trac.wordpress.org/browser/twentyten/1.0.3/single.php?rev=12716) from 12 years ago, and there are probably earlier examples too.

Creating and maintaining a contrived theme that still takes advantage of new WP core features that validate how customers and the code in this repo would be expected to engage with it, doesn't seem like the best use of time.

jesse-greathouse:
> Creating and maintaining a contrived theme that still takes advantage of new WP core features that validate how customers and the code in this repo would be expected to engage with it, doesn't seem like the best use of time.

Yea that makes sense. Thanks for answering.

PR #2989, opened 2022-03-14
Production changelog

pschoffer:
# Description

On Release workflow also create a summary changelog entry with a list of all the changelog entries composing it.


GaryJones:
> On Release workflow also create a summary changelog entry with a list of all the changelog entries composing it.

That's going to be very useful (more useful than individual changelog entries I would say, since a release bundles several changes together).

jesse-greathouse (approved):
Very good, @pschoffer

PR #2991, opened 2022-03-15
Add support for WP 6.0's attachment filesize changes

WPprodigy:
## Description

Fix the currently broken nightly tests :), and adds support for how WP 6.0 will be handling this feature.

With https://core.trac.wordpress.org/ticket/49412 landing in core (WP 6.0), the attachment `filesize` will be saved in attachment postmeta. We happen to [already do this](https://github.com/Automattic/vip-go-mu-plugins/blob/30acf90f5a2cda3466f11fdcdc36b6b8482eb5c4/files/class-vip-filesystem.php#L337-L348), and notably our filter will essentially just early return once core starts adding the meta beforehand. So the switchover on that side of things will be pretty smooth.

Core also stores the `filesize` for each additional attachment/thumbnail sizes though, which we do not currently do. This PR adds in early support for that because once core has this, it becomes an expectation that it exists. Note though that because we don't truly create addition files/thumbnails, there isn't an actual different filesize to be listed. So I opted for just storing the original file's filesize in the place. The other option would be listing `0` which is core's new default for when it can't retrieve a filesize, but I'm leaning towards showing the original filesize being a bit better.

Example:

```
wp> wp_get_attachment_metadata(7);

=> array(6) {
  ["width"]=> int(1353)
  ["height"]=> int(1235)
  ["file"]=> string(24) "2022/03/example.png"
  ["filesize"]=> int(2015937)
  ["sizes"]=>
  array(4) {
    ["medium"]=>
    array(5) {
      ["file"]=> string(24) "example-300x274.png"
      ["width"]=> int(300)
      ["height"]=> int(274)
      ["mime-type"]=> string(9) "image/png"
      ["filesize"]=> int(2015937)
    }
    ["large"]=>
    array(5) {
      ["file"]=> string(25) "example-1024x935.png"
      ["width"]=> int(1024)
      ["height"]=> int(935)
      ["mime-type"]=> string(9) "image/png"
      ["filesize"]=> int(2015937)
    }
    [...]
  }
}
``` 

### Deploy Notes

Let me deploy this please, as I'd like to give it some more testing on live VIP envs while just in staging stacks - both pre-6.0 and nightly builds.

### Attachment metadata: Store the filesize within each "additional size" attachment entry.

Stores the attachment's `filesize` within each "virtual sizes" in the attachment metadata. Keeps us in line with how WP core will be handling this same behavior in WP 6.0+: https://core.trac.wordpress.org/ticket/49412

## Checklist

## Steps to Test

- Upload an image.
- Run `wp_get_attachment_metadata( $attachment_id );`

Note that the behavior is different on local dev-envs vs live VIP envs. Live VIP envs should now always have the `filesize` data in the meta and within each additional size. Locally, you'll see this take place only on WP 6.0+ b/c our related code doesn't run when the VIP filesystem isn't initiated.

### DIFF
--- a/a8c-files.php
+++ b/a8c-files.php
@@ -175,7 +175,7 @@ class A8C_Files {
 		if ( defined( 'VIP_JETPACK_SKIP_LOAD' ) && true === VIP_JETPACK_SKIP_LOAD ) {
 			return false;
 		}
-		
+
 		if ( defined( 'WPCOM_VIP_USE_JETPACK_PHOTON' ) && true === WPCOM_VIP_USE_JETPACK_PHOTON ) {
 			return true;
 		}
--- a/files/class-image.php
+++ b/files/class-image.php
@@ -32,6 +32,9 @@ class Image {
 	/** @var bool $is_resized Whether the attachment has been resized yet, or not. */
 	private $is_resized = false;
 
+	/** @var int $filesize Image original filesize */
+	private $filesize;
+
 	/**
 	 * Constructs the image object.
 	 *
@@ -50,6 +53,7 @@ class Image {
 		$this->width           = $data['width'];
 		$this->height          = $data['height'];
 		$this->mime_type       = $mime_type;
+		$this->filesize        = isset( $data['filesize'] ) ? $data['filesize'] : 0;
 	}
 
 	/**
@@ -82,7 +86,7 @@ class Image {
 	 *
 	 * @param array $size_data Array of width, height, and crop properties of a size.
 	 *
-	 * @return array|\WP_Error An array containing file, width, height, and mime-type keys and it's values. WP_Error on failure.
+	 * @return array|\WP_Error An array containing file, width, height, filesize, and mime-type keys and it's values. WP_Error on failure.
 	 */
 	public function get_size( $size_data ) {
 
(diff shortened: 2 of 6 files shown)


WPprodigy:
Ah, so in the tests we don't truly load up all the stream wrapper VIP filesystem stuffs, including the `filter_wp_generate_attachment_metadata` filter that adds in the `filesize` metadata pre-WP6.0

So...either our tests need to diverge for 6.0+, or we could move this filter up a level and just always have it work regardless of if `VIP_FILESYSTEM_USE_STREAM_WRAPPER` is enabled.  I think the latter probably is best, as the only effect afaik will just be that it starts working on dev-envs essentially.

a8c-files.php
```diff
@@ -175,7 +175,7 @@ private function use_jetpack_photon() {
 		if ( defined( 'VIP_JETPACK_SKIP_LOAD' ) && true === VIP_JETPACK_SKIP_LOAD ) {
 			return false;
 		}
-
```
jesse-greathouse, line 178: 🥇

files/class-image.php
```diff
@@ -157,6 +162,18 @@ public function get_mime_type() {
 		return $this->mime_type;
 	}
 
+	/**
```
jesse-greathouse, line 165: would be good to note the file size measuring unit  whether it be bytes, kb, or mb just so its clear what this integer represents in case someone ends up in this code later.

jesse-greathouse (dismissed):
Good coding, I'd just like to see the file size unit mentioned in the comment.

rebeccahum (approved):

PR #2995, opened 2022-03-15
Search: Add X-Opaque-Id header for ES Task API

rebeccahum:
## Description
Let's send a `X-Opaque-Id` header for identifying sites. Using `vip-<site id>` as an identifier so we don't run into the partial match issue of looking for site 60 and receiving 260.

Add `X-Opaque-Id` header to requests for ES Task API to access.

## Checklist

## Steps to Test

1. Check out PR.
2. Sandbox a site with VIP Search enabled
3. `var_dump()` to see if the headers get sent


WPprodigy (dismissed):

search/includes/classes/class-search.php
```diff
@@ -856,6 +856,7 @@ public function filter__ep_do_intercept_request( $request, $query, $args, $failu
 				'X-Client-Site-ID' => FILES_CLIENT_SITE_ID,
 				'X-Client-Env'     => VIP_GO_ENV,
 				'Accept-Encoding'  => 'gzip, deflate',
+				'X-Opaque-Id'      => 'vip-' . FILES_CLIENT_SITE_ID, // for IDing tasks: https://www.elastic.co/guide/en/elasticsearch/reference/master/tasks.html#_identifying_running_tasks
```
rinatkhaziev, line 859: A few things:
1. `vip-` prefix is probably redundant.
2. I think we may want this to be unique-ish and also more telling. Something like:  
```suggestion
				'X-Opaque-Id'      => sprintf( '%d:%s:%d', FILES_CLIENT_SITE_ID, $type, hrtime( true ) ); , 
```

Here we'd identify the request by the site id, then the request type, and its timestamp with nanoseconds

rinatkhaziev: I removed the comment from the suggestion cause we can use this in many other ways to identify requests.

rinatkhaziev (changes requested):
Thanks, left some feedback inline.

search/includes/classes/class-search.php
rebeccahum: @rinatkhaziev Is the timestamp really necessary when that information will already be in the logs? I was hoping for something not _too_ unique so that it can be used to be more "filterable". How about something like this?

```
'X-Opaque-Id'      => sprintf( '%d:%s', FILES_CLIENT_SITE_ID, $type ); , 
```

WPprodigy: Can `$type` end up being `null` here like in the function signature? Maybe want a specific whitelist for it?

```
$logged_type = in_array( $type, $allowlist, true ) ? $type : 'unknown'
```

jesse-greathouse (approved):

search/includes/classes/class-search.php
rinatkhaziev: > Don’t generate a unique X-Opaque-Id header for every request. Too many unique X-Opaque-Id values can prevent Elasticsearch from deduplicating warnings in the deprecation logs.

That's an interesting bit I didn't know about. Sure, let's try it this way.

rinatkhaziev (approved):

search/includes/classes/class-search.php
rebeccahum: @WPprodigy I think just a check that it isn't a `null` value should suffice.

jesse-greathouse (approved):
LGTM

rebeccahum:
r2394-stacks

PR #2810, opened 2022-01-14
Search: Only check index settings for active version

rebeccahum:
## Description
It is not necessary to check the index settings for inconsistencies on an inactive version.

## Checklist

Please make sure the items below have been covered before requesting a review:

- [x] This change works and has been tested locally (or has an appropriate fallback).
- [ ] This change works and has been tested on a Go sandbox.
- [ ] This change has relevant unit tests (if applicable).
- [ ] This change has relevant documentation additions / updates (if applicable).
- [ ] I've created a changelog description that aligns with the provided examples.

## Steps to Test
1) Add some "bad" settings to current index "1", e.g.
```
$bad_settings = [ 'index.number_of_replicas' => 8, 'index.max_result_window' => 50 ];
$indexable = \ElasticPress\Indexables::factory()->get( 'post' );
$indexable-> update_index_settings( $bad_settings );
```
2) Create a new post index version "2" `wp vip-search index-versions add post`
3) Swap to the new index "2" `wp vip-search index-versions activate post next`
4) Ensure that the old index "1" isn't returned from `get_index_settings_health_for_all_indexables()`

PR #2999, opened 2022-03-17
Search: Do healthcheck on inactive indexes and build new index only for active indexes

rebeccahum:
## Description
This PR does two things:
- reverts #2810 which made health check only on active indexes
- builds new index to meet shard requirements only for active indexes

1) Healthcheck will fire on all indexes regardless of active state. 
2) To meet shard requirements, new indexes will only be automatically built for active indexes.

## Checklist

## Steps to Test
Scenario 1: Check if healthcheck fires for all indexes regardless of state
1) Add some "bad" settings to current index "1", e.g.
```
$bad_settings = [ 'index.number_of_replicas' => 8, 'index.max_result_window' => 50 ];
$indexable->update_index_settings( $bad_settings );
```
2) Create a new post index version "2" `wp vip-search index-versions add post`
3) Swap to the new index "2" `wp vip-search index-versions activate post next`
4) Ensure that the old index "1" isn't returned from `get_index_settings_health_for_all_indexables()`

Scenario 2: Build new index requirements only for active index
1) Fake index shard requirement mismatch:

```
add_filter( 'ep_post_mapping', function( $mapping ) {
    $mapping['settings']['index.number_of_shards'] = 3;
return $mapping;
}, 20 );
```

2) Shell in and kick off a health check:
```
$es->init();
$hj->check_settings_health();
```

(Can also throw a dump before `maybe_process_build()` to ensure that `$result['index_version']` matches the active index version)


search/includes/classes/class-health.php
```diff
@@ -954,17 +954,19 @@ public function get_index_settings_health_for_all_indexables() {
...
 		$diff = [];
 
-		if ( ! empty( $version_result ) ) {
-			$diff[] = $version_result;
+		foreach ( $versions as $version ) {
+			$version_result = $this->get_index_settings_diff_for_indexable( $indexable, array(
+				'index_version' => $version['number'],
+			) );
+
+			if ( ! empty( $version_result ) ) {
```
jesse-greathouse, line 967: I would prefer to use something more explicit than `empty` and I admit it is entirely personal preference.

Consider the merit of this code:

```php
if ( is_array( $version_result ) && ( 0 < count( $version_result ) ) {
```

I think it's less terse, but it's more direct, and it doesn't carry any of the pitfalls that can happen if and when `$version_result` ends up as the wrong type for some reason. I like the preciseness of the statement that this code could only execute if its an array and if it has a length of greater than zero.

I looked into the function `get_index_settings_diff_for_indexable` and it looks like this can only return an empty array in the form of: `[]` or a dictionary type array with the specific object graph that this function is meant to produce.

This is a subtle nuance of PHP being kind of gross and wanting to call a potential list the same thing as a potential dictionary.

jesse-greathouse (approved):
I left a comment, but you can take it or leave it.

search/includes/classes/class-health.php
rebeccahum: Thanks for the feedback. It does look like it can also be a WP_Error, so we definitely don't want that to bubble up. I've addressed your feedback now 🙂

PR #3003, opened 2022-03-18
Search: Add alerting to detect when no origin DC detected

rebeccahum:
## Description
Add alerting to see how often no origin DC is detected during mapping process.


rinatkhaziev:
@rebeccahum thanks, I think we also need an alert in the above block (when `settings` is not an array. Neither of these two conditions should be possible, yet it happened somehow.

rebeccahum:
@rinatkhaziev Added.

rinatkhaziev (approved):

rebeccahum:
r2398-stacks

PR #3012, opened 2022-03-25
VIP Feed Cache: Fix PHP 8 error non-static method VIP_Go_Feed_Cache::create() cannot be called statically

rebeccahum:
## Description
Calling `fetch_feed()` causes a PHP 8 fatal `non-static method VIP_Go_Feed_Cache::create() cannot be called statically`. Similar to https://core.trac.wordpress.org/changeset/49565, we can use register the `VIP_Go_Feed_Cache_Transient` class.

Note: After all sites have migrated to k8s, we can remove the entire VIP Go Feed Cache plugin since `mtime()` won't differ across pods.

### Plugin Updated: VIP Go Feed Cache

Fixes PHP 8 fatal "non-static method VIP_Go_Feed_Cache::create() cannot be called statically".

## Checklist

## Steps to Test
1) `$feed = fetch_feed('https://www.nasa.gov/rss/dyn/breaking_news.rss');`
2) Dump out `$feed` and take note of the build # (ctrl + F "build")
3) Apply patch
4) Repeat step 2 and ensure it matches up
5) Check Debug Monitor and notice that the error is gone


vip-feed-cache/class-vip-go-feed-transient.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-class VIP_Go_Feed_Cache_Transient extends WP_Feed_Cache_Transient {
+class VIP_Go_Feed_Cache_Transient extends WP_Feed_Cache_Transient implements SimplePie_Cache_Base {
```
rebeccahum, line 3: I added the interface b/c of https://github.com/WordPress/WordPress/blob/270f2011f8ec7265c3f4ddce39c77ef5b496ed1c/wp-includes/SimplePie/Cache/Base.php#L47-L49

rinatkhaziev (dismissed):
THANKS!! 🚀 🎱

rinatkhaziev (approved):

rebeccahum:
r2404-stacks

vip-feed-cache/class-vip-go-feed-cache.php
```diff
@@ -11,7 +11,7 @@ class VIP_Go_Feed_Cache extends SimplePie_Cache {
 	 * @param string $extension 'spi' or 'spc'.
 	 * @return VIP_Go_Feed_Cache_Transient Feed cache handler object that uses transients and normalizes SimplePie Build number.
 	 */
-	public function create( $location, $filename, $extension ) {
+	public static function get_handler( $location, $filename, $extension ) {
```
WPprodigy, line 14: Technically a backwards compatibility break if somebody were using this class/method directly.

Generally would be best to leave the existing method and show a deprecation notice/warning so we're more prepared if somebody is extending/using it, rather than causing fatals upon deploy.

rebeccahum: @WPprodigy When I grepped for it, I didn't find any references, so I figured we should be a-ok?

WPprodigy: Could be tricky to cache if some custom code was using the abstraction layers and stuff though, or the `wp_feed_options` hook for example 😬 

Low probability, but definitely gotta be careful when removing public callables.

PR #3010, opened 2022-03-23
wp-admin: add notice for Jetpack Search migrations

rebeccahum:
## Description
This PR does two things:
- Allows the anchor tag to be used in Admin_Notice
- For sites that still have the `JETPACK_SEARCH_VIP_INDEX` constant defined, adds a notice regarding the EOL.

### Plugin Updated: Jetpack Search

Add wp-admin notice to notify user of upcoming end-of-life for Jetpack Search custom indexes

## Checklist

Please make sure the items below have been covered before requesting a review:

- [x] This change works and has been tested locally (or has an appropriate fallback).
- [ ] This change works and has been tested on a Go sandbox.
- [ ] This change has relevant unit tests (if applicable).
- [ ] This change has relevant documentation additions / updates (if applicable).
- [x] I've created a changelog description that aligns with the provided examples.

## Steps to Test
1) Check out PR
2) Add `define( 'JETPACK_SEARCH_VIP_INDEX', true );` in vip-config
3) Go to wp-admin and expect to see notice panel

PR #3023, opened 2022-03-28
wp-admin: Do not display Search Migrations admin notice for extensions

rebeccahum:
## Description
Exclude certain IDs from displaying the message in #3010.


rinatkhaziev (approved):

rebeccahum:
r2405-stacks

PR #3027, opened 2022-03-29
Search: Load vip-search adapter no matter what

rebeccahum:
## Description
This PR removes the other adapters since we only need the vip-search one for Enterprise Search for our mu-plugins. Therefore, if one attempts to load via `es_wp_query_load_adapter( 'jetpack-search' );` when Enterprise Search is enabled and they are testing with the `vip-search-enabled=1` query vars, there won't be the below fatal anymore:

> PHP Fatal error:  Cannot redeclare vip_es_field_map() (previously declared in /var/www/wp-content/mu-plugins/search/es-wp-query/adapters/vip-search.php:109) in /var/www/wp-content/mu-plugins/search/es-wp-query/adapters/jetpack-search.php on line 107

I realize we already surface the error at: https://github.com/Automattic/vip-go-mu-plugins/blob/5c591826c8a6f10590865b56853a7c3d44adeed1/search/includes/classes/class-search.php#L725-L728 
...but that's only if someone is using their own version of `es-wp-query` and it doesn't apply if someone has already removed their version of the `es-wp-query` plugin but still has the `es_wp_query_load_adapter( 'jetpack-search' );` bit remaining, this would be an extra step in removing any Search migrations friction.

Remove adapters that are not vip-search and only load vip-search adapters.

## Checklist

## Steps to Test
1) Have some code that calls `es_wp_query_load_adapter( 'jetpack-search' );`
2) Do a search `?s=test&vip-search-enabled=1` and expect to see fatal
3) Apply PR
4) Repeat step 2 and no longer see fatal


katag9k:
Tested on site ID 2022 where `es_wp_query_load_adapter( 'jetpack-search' );` was loaded manually. 👍🏻

WPprodigy (approved):

rebeccahum:
r2407-stacks

PR #3031, opened 2022-03-30
Search: Do not alert if index is in midst of being re-built

rebeccahum:
## Description
We don't need multiple alerts firing on index inconsistencies if the active index is already being re-built.

## Checklist

## Steps to Test
1. Fake index shard requirements not being met:
```
add_filter( 'ep_post_mapping', function( $mapping ) {
    $mapping['settings']['index.number_of_shards'] = 3;
return $mapping;
}, 20 );
```
2. Fake index in the process of being built:
```
update_option( 'vip_search_cron_new_version_building_lock', true );
```
3. Throw a die in the condition
4. Shell in and kick off a health check: 
```
$es->init();
$hj->check_settings_health();
```


search/includes/classes/class-settingshealthjob.php
```diff
@@ -149,6 +149,14 @@ public function process_indexables_settings_health_results( $results ) {
 					continue;
 				}
 
+				$indexable = $this->indexables->get( $indexable_slug );
+				if ( true === array_key_exists( 'index.number_of_shards', $result['diff'] ) && 1 === count( $result['diff'] )
```
jesse-greathouse, line 153: I believe this reduces to:

```php
if ( isset( $result['diff']['index.number_of_shards'] ) && 1 === count( $result['diff'] )
```

This is because I presume that `$result['diff']['index.number_of_shards']` cannot be a `NULL` strongly typed value. Am I wrong?

[citation here](https://www.geeksforgeeks.org/difference-between-isset-and-array_key_exists-function-in-php/#:~:text=%3F%3E,-Output%3A%20array%20key&text=Difference%20between%20isset()%20and%20array_key_exists()%20Function%3A%20The%20main,exists%20and%20is%20not%20null.)

Forgive me also that I did not see a check for the 'diff' dictionary on $result, but that may be safe to ignore? or maybe I just don't see it.

jesse-greathouse (changes requested):
Just had one and a half questions.

search/includes/classes/class-settingshealthjob.php
rebeccahum: Ah, good call. Yes, `number_of_shards` should never be a `NULL` value. @jesse-greathouse I fixed it per your recommendation, thank you!

jesse-greathouse: [Pulled this little gem out of the recesses of my memory](https://ilia.ws/archives/247-Performance-Analysis-of-isset-vs-array_key_exists.html) just to put a spotlight on why the syntax difference is relevant. However, I admit that some time has passed and it may now be different.

rebeccahum: I fixed the other part in the code where we are using `array_key_exists()` FWIW.

jesse-greathouse (approved):
very nice, @rebeccahum 🔆

rebeccahum:
r2422-stacks

jesse-greathouse:
Release v20220405.0

PR #3039, opened 2022-04-05
Healthcheck that fully loads customer code and plugins before determining whether a site is healthy or not.

rinatkhaziev:
## Description

This should help us to determine whether a site is healthy or not (we can use it as a readiness probe and to potentially power other places).

The main idea here is to perform the health check at a point in the request lifecycle where all the code is loaded (as opposed to `cache-healthcheck?` which gets executed early on.

### Plugin Introduced: Healthcheck

We've introduced a new Healthcheck which runs at a point where all of the site's codebase should be loaded. Any errors at this point could be used as a strong signal that the application is unhealthy.

## Checklist

## Steps to Test

- Check out PR.
- Hit `/vip-healthcheck` in the browser or via curl:

```
 curl --head http://mu.vipdev.lndo.site/vip-healthcheck
HTTP/1.1 200 OK
Cache-Control: no-cache, must-revalidate, max-age=0
Content-Type: text/plain;charset=UTF-8
Date: Wed, 27 Apr 2022 22:35:29 GMT
Expires: Wed, 11 Jan 1984 05:00:00 GMT
```

- It should display return an empty document with `200` status code.
- Now add an error via `$hc->add_error( 'test error'  )` and hit the URL again, it should display the error text and return a 500 error.


healthcheck.php
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Utility Healthcheck class that runs further than cache-healthcheck and allow to verify the app code is working.
+ */
+class Healthcheck {
+	/**
+	 * By default the site is considered healthy unless filter value returns falsy value
+	 *
+	 * @var boolean
+	 */
+	protected bool $healthy = true;
```
jesse-greathouse, line 14: personally I like one newline space after a property declaration but I noticed the linter didn't agree so :shrug:

healthcheck.php
```diff
@@ -0,0 +1,36 @@
...
+	public function __construct() {
+		$this->check();
+	}
```
jesse-greathouse, line 17: I think it would look nicer with a newline after this function.

healthcheck.php
```diff
@@ -0,0 +1,36 @@
...
+
+	public function is_healthy() {
+		return $this->healthy;
+	}
+}
+
+/**
+ * It's safe to assume that by the time of `template_redirect` the codebase is loaded fully.
+ */
+if ( isset( $_SERVER['REQUEST_URI'] ) && '/vip-healthcheck' === $_SERVER['REQUEST_URI'] ) {
```
jesse-greathouse, line 30: Can the `'/vip-healthcheck'` string be defined as a constant of class? This way if the endpoint ever changes, we can just change the constant instead of touching the logic.

I think the code would look nicer if it was:

```php
if ( isset( $_SERVER['REQUEST_URI'] ) && VIP_HEALTHCHECK_ENDPOINT === $_SERVER['REQUEST_URI'] ) {
```

I admit this is a pedantic preference, I just think it looks better.

jesse-greathouse (dismissed):
Had a few cosmetic suggestions. Feel free to take it or leave it.

healthcheck.php
```diff
@@ -0,0 +1,36 @@
...
+	 * By default the site is considered healthy unless filter value returns falsy value
+	 *
+	 * @var boolean
+	 */
+	protected bool $healthy = true;
+	public function __construct() {
+		$this->check();
+	}
+	public function check() {
+		$this->healthy = (bool) apply_filters( 'vip_site_healthcheck', $this->healthy );
```
bnu0, line 19: i would propose that rather than a bool, this returns a `string[]`, where empty array means no errors (aka healthy) and non-empty indicates any number of reasons for the unhealthiness - then we can echo these reasons before exiting so it is trivial to look at a pod and ask it why it thinks it is not healthy, and not have to guess.

jesse-greathouse:
Needs 1 unit test I think.

need to be able to assert that this gives the correct response when the URI is the endpoint, and the correct response when the customer code doesn't load, so that we can feel confident there is no false positive.

healthcheck.php
GaryJones: Agree - the conversion to bool can be done in `is_healthy()`, but a new function could return the array of errors (`get_health_errors()` or whatever).

healthcheck.php
```diff
@@ -0,0 +1,36 @@
...
+}
+
+/**
+ * It's safe to assume that by the time of `template_redirect` the codebase is loaded fully.
+ */
+if ( isset( $_SERVER['REQUEST_URI'] ) && '/vip-healthcheck' === $_SERVER['REQUEST_URI'] ) {
+	add_action( 'template_redirect', function() {
+		nocache_headers();
+		$is_healthy = ( new Healthcheck() )->is_healthy();
+		exit( $is_healthy ? 'Ok' : 'Not ok' );
```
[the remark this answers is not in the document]
chriszarate: Would it be worth exiting with `503` or some other unique non-200 code, to make filtering easier?

healthcheck.php
rinatkhaziev: Yeah, I personally like the inline hardcoded strings in cases like these (where they are used only once), which makes searching easier. IF there were two or more instances of using it, I'd agree with you.

healthcheck.php
rinatkhaziev: @bnu0 Addressed.

healthcheck.php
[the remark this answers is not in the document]
rinatkhaziev: @chriszarate ummm I'd defer to @bnu but I think 500 makes sense here since we're trying to determine whether there's a runtime error.

chriszarate:
Just to record it on the PR, my main question is whether we should add a new endpoint or adapt the existing `/cache-healthcheck` endpoint for this purpose.

rinatkhaziev:
@chriszarate I think cache-healthcheck is for a different use-case where we check whether there are any PHP worker issues. It's meant to be super-quick so it bails as we start loading the mu-plugins.

healthcheck.php
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * Utility Healthcheck class that runs further than cache-healthcheck,
+ * and allows to verify the app code is working as expected.
+ */
+class Healthcheck {
+	/**
+	 * By default, an app is considered healthy unless filter value returns a non-empty array
+	 *
+	 * @var array
```
GaryJones, line 12:
```suggestion
	 * @var array<string>
```

(Could also be `string[]` like elsewhere in this file, but I prefer the former, as the former feels more like `array()`, and there's nothing like `string[]` syntax in PHP where the type comes first. The former also allows consistency when doing something like `array<int, string>` and other array shapes.)

healthcheck.php
```diff
@@ -0,0 +1,74 @@
...
+	/**
+	 * Apply the filters to and set the `$errors` property accordingly.
+	 *
+	 * @return void
+	 */
+	public function check() {
+		/**
+		 * Filters the errors that would indicate the app is unhealthy.
+		 *
+		 * @param string[]     $errors either empty array (no errors) or array of strings: [ 'an error has occurred', 'another error' ]
```
GaryJones, line 29:
```suggestion
		 * @param string[] $errors Either empty array (no errors) or array of strings: [ 'an error has occurred', 'another error' ].
```

healthcheck.php
```diff
@@ -0,0 +1,74 @@
...
+		 */
+		$this->errors = apply_filters( 'vip_site_healthcheck_errors_array', $this->errors );
+		if ( ! is_array( $this->errors ) ) {
+			$this->errors = [ 'Unexpected value in Healthcheck->errors' ];
+		}
+	}
+
+	/**
+	 * Check if the app is healthy.
+	 * @return boolean true
```
GaryJones, line 39:
```suggestion
	 * @return bool True if there are no errors, false if there are.
```

healthcheck.php
```diff
@@ -0,0 +1,74 @@
...
+	 */
+	public function is_healthy() {
+		return is_array( $this->errors ) && empty( $this->errors );
```
GaryJones, line 42:
```suggestion
		return count( $this->errors ) === 0;
```

The array check seems redundant, as it is instantiated as an array, and there's a guard clause when assigning to it that if it isn't an array, then it's made an array.

See https://beberlei.de/2021/02/19/when_to_use_empty_in_php_i_say_never.html for why it's clearer code to stay away from `empty()`.

healthcheck.php
```diff
@@ -0,0 +1,74 @@
...
+	 * 200 or 500 with the reasons listed in the response body.
+	 *
+	 * @return void
+	 */
+	public function render() {
+		$is_healthy = $this->is_healthy();
+
+		nocache_headers();
+		header( 'Content-Type: text/plain' );
+		http_response_code( $is_healthy ? 200 : 500 );
```
GaryJones, line 57:
```suggestion
		nocache_headers();
		header( 'Content-Type: text/plain' );
		$response_code = $this->is_healthy() ? 200 : 500;
		http_response_code( $response_code );
```

Avoids having a ternary condition inside a function call.

healthcheck.php
jesse-greathouse: fair enough.

jesse-greathouse (approved):

healthcheck.php
```diff
@@ -0,0 +1,74 @@
...
+	 *
+	 * @return void
+	 */
+	public function check() {
+		/**
+		 * Filters the errors that would indicate the app is unhealthy.
+		 *
+		 * @param string[]     $errors either empty array (no errors) or array of strings: [ 'an error has occurred', 'another error' ]
+		 */
+		$this->errors = apply_filters( 'vip_site_healthcheck_errors_array', $this->errors );
```
WPprodigy, line 31: What sorts of things do we expect customer's to add in checks for? Asking because maybe we should ship this endpoint initially without a filter like this. If/once the use case comes up where there is some sort of dependency to people want to check, we'd maybe end up going a different route with how we want to handle it.

bnu0 (approved):
LGTM 🌵🐫

healthcheck.php
rinatkhaziev: > What sorts of things do we expect customer's to add in checks for?

Not sure just yet, I think that hugely depends on doc coverage and our advice in implementing it, the goal here was to implement something more advanced from the get-go. There are plenty of ways someone can break their app with PHP alone which would otherwise be considered "healthy" by our simple "if it loaded the codebase it's ok" check. So I thought providing some flexibility by allowing to check something critical for the app but not resulting in a fatal would be nice.

rinatkhaziev: @simonwheatley @mjangda thoughts on this?

WPprodigy: My thought process is that it's always easier to add a filter later, but much harder to remove one. And particularly in this case, I'm not able to think of any really good use cases - but a number of bad ones 🙃 .

A "readiness" check to me just means that the application built (no fatals), and required services are online/connected (DB/memcache). Since this endpoint returns after all the `init` stuff takes place, I think we cover both pretty well.

If a customer wants a particular "ensure this is present before a deploy" check, then at any point during the `init` process they could throw a PHP fatal.

healthcheck.php
```diff
@@ -0,0 +1,95 @@
...
+	/**
+	 * By default, an app is considered healthy unless filter value returns a non-empty array
+	 *
+	 * @var string[] - Either empty array (no errors) or array of strings: [ 'an error has occurred', 'another error' ].
+	 */
+	protected array $errors = [];
+
+	public function __construct() {
+		$this->check();
+	}
```
GaryJones, line 18: Having side effects like this, within the constructor, makes the code significantly harder to add tests for. Constructors are there to set up the class, not run a WP filter function.

Wherever the class is instantiated, should call `->check()`, directly itself.

WPprodigy (approved):

rinatkhaziev:
r2485-stacks

PR #3045, opened 2022-04-11
GH-2980: make sure `wpcom_vip_override_role_caps()` persists capabilities

### LINKED ISSUE #2980: Function wpcom_vip_override_role_caps doesn't work as expected
## Expected/Desired Behavior

Given a role, the function ``wpcom_vip_override_role_caps`` is expected to override the role's current capabilities by the one(s) sent as a parameter. 

## Actual Behavior

The role capabilities remain unchanged. 

1. Via WP Shell: execute the function. sending the role to be modified and the caps to be applied;
In the code below, I'm using as an example the role `contributor` and the capability ``read``.

```
wp> wpcom_vip_override_role_caps( 'contributor', array( 'read' => false ) );
```

2. Via WP CLI: Check whether the capabilities for the role 'contributor' have changed.

```
~# wp cap list "contributor"
	my-custom-cap
	edit_posts
	read
	level_1
	level_0
```

The capabilities remain the same while it should be empty. 

## Additional notes

**Code:**  https://github.com/Automattic/vip-go-mu-plugins/blob/master/vip-helpers/vip-roles.php#L82-L92
**VIP how to customize user roles public docs:** https://docs.wpvip.com/how-tos/customize-user-roles/

sjinks:
## Description

This PR fixes #2980 by adding the code to persist the overridden capabilities to the database.

Bug fix: `wpcom_vip_override_role_caps()` now persists the capabilities to the database.

## Checklist

## Steps to Test

The CI should pass -- I have added a test case for this issue.

### DIFF
--- a/vip-helpers/vip-roles.php
+++ b/vip-helpers/vip-roles.php
@@ -88,6 +88,11 @@ function wpcom_vip_override_role_caps( $role, $caps ) {
 
 	$role_obj->capabilities = (array) $caps;
 
+	$roles = wp_roles();
+
+	$roles->roles[ $role ]['capabilities'] = $caps;
+	update_option( $roles->role_key, $roles->roles );
+
 	_wpcom_vip_maybe_refresh_current_user_caps( $role );
 }
 
--- a/tests/vip-helpers/test-vip-roles.php
+++ b/tests/vip-helpers/test-vip-roles.php
@@ -1,0 +1,18 @@
+<?php
+
+namespace Automattic\VIP\Helpers;
+
+use WP_UnitTestCase;
+
+class Test_VIP_Roles extends WP_UnitTestCase {
+	public function test_override_role_caps(): void {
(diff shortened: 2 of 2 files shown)


vip-helpers/vip-roles.php
```diff
@@ -88,6 +88,11 @@ function wpcom_vip_override_role_caps( $role, $caps ) {
 
 	$role_obj->capabilities = (array) $caps;
 
+	$roles = wp_roles();
+
+	$roles->roles[ $role ]['capabilities'] = $caps;
+	update_option( $roles->role_key, $roles->roles );
```
sjinks, line 94: There was an option to use `remove_cap()` and `add_cap()` in a loop instead; however, I did not like that because it would trigger `update_option()` for every operation, and that does not look database- and cache-friendly.
_

sjinks:
r2433-stacks

sjinks:
v20220413.0

PR #3064, opened 2022-04-19
WP-CLI: Add Support for the db query <querystring> Command

jblz:
## Description

The `wp db` commands currently fail because we neither set the appropriate constants, nor provide any other way for the commands to know how to connect to the database.

See the `DB_Command` code for more info:
https://github.com/wp-cli/db-command/blob/de3eddeaa224e617a05fb3a8a97d77fefc456c4f/src/DB_Command.php#L1595-L1597
https://github.com/wp-cli/db-command/blob/de3eddeaa224e617a05fb3a8a97d77fefc456c4f/src/DB_Command.php#L453

This change:

* Creates a new file `lib/helpers/wp-cli-db.php` (& backing functionality in directory:  `lib/helpers/wp-cli-db`) which registers a function to the `before_run_command` [WP-CLI hook](https://make.wordpress.org/cli/handbook/references/internal-api/wp-cli-add-hook/#notes) that:
  * Returns early if the `db` command is not seen
  * Pulls db settings out of the global var
  * Determines which server to use
  * Defines the appropriate constants
  * Validates the db subcommands and throws an error for everything except `db query <querystring>`

Note: This PR does not include sourcing this new functionality. See the Test Plan for more info.

Also Note: We still have to add support to the API to permit the `db` command. This only fixes the backend issue.

### Read-only By Default

Unless a PHP constant is set correctly when the configuration is loaded, a read-only replica will be the only server this can use.

## Changelog Description

WP-CLI: Add initial support for the db command.

## Checklist

### Automated Tests

* Review the added tests for completeness and accuracy
* Tests should pass locally (e.g. `./bin/test.sh --filter WP_Cli_Db_Test`) and in CI

#### Disabled by Default

* Spin up a sandbox for a test site
* Pull the code from this branch to the sandbox
* Try out various `wp db` commands
  * They should all fail since you did not enable the feature

#### Enabled Read-Only

* Add the following code to your `vip-config/vip-config/php`:
  ```
  if ( ! defined( 'WPVIP_ENABLE_WP_DB' ) ) {
  	define( 'WPVIP_ENABLE_WP_DB', 1 );
  }
  
  if ( defined( 'ABSPATH' ) && is_readable( ABSPATH . 'wp-content/mu-plugins/lib/helpers/wp-cli-db.php' ) ) {
  	require_once ABSPATH . 'wp-content/mu-plugins/lib/helpers/wp-cli-db.php';
  }
  ```
* Try out various `wp db` commands
  * You should be able to run read queries / ops (e.g. `wp db query "SHOW TABLES"`, etc.)
  * Write queries (like `INSERT`, etc.) should not work
  * `wp db query` (without a query string) should fail
  * Subcommands other than `wp db query <querystring>` should fail

#### Enabled Read and Write

* Add the following code to your `vip-config/vip-config.php`:
  ```
  if ( ! defined( 'WPVIP_ENABLE_WP_DB_WRITES' ) ) {
  	define( 'WPVIP_ENABLE_WP_DB_WRITES', 1 );
  }
  ```
* Try out various `wp db` commands
  * You should be able to run read and write queries with `wp db query <querystring>`
  * `wp db query` (without a query string) should fail
  * Subcommands other than `wp db query <querystring>` should fail


lib/helpers/wp-cli-db.php
```diff
@@ -0,0 +1,95 @@
...
+
+	$write_specific_subcommands = [
+		'clean',
+		'create',
+		'drop',
+		'import',
+		'optimize',
+		'repair',
+		'reset',
+	];
```
jblz, line 39: We may want to disallow some of these entirely :thinking:

lib/helpers/wp-cli-db.php
```diff
@@ -0,0 +1,97 @@
...
+
+	$subcommand = $command[1];
+	if ( ! $allow_writes && in_array( $subcommand, $write_specific_subcommands ) ) {
+		echo "Error: The 'wp db $subcommand' subcommand is not currently allowed for this site.\n.";
```
jblz, line 43:
```suggestion
		// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
		echo "Error: The 'wp db $subcommand' subcommand is not currently allowed for this site.\n.";
```

lib/helpers/wp-cli-db.php
```diff
@@ -0,0 +1,98 @@
...
+	global $db_servers;
+
+	if ( ! ( isset( $command[0] ) && 'db' === $command[0] ) ) {
+		// Don't do anything for any command other than `db`
+		return;
+	}
+
+	if ( ! is_array( $db_servers ) ) {
+		echo "Error: No database servers are configured for this environment.\n";
+		exit( 10 );
```
rinatkhaziev, line 26: Why the `10` specifically?

jblz: No specific reason other than to have a non-zero exit code (to signify an error). We should decide which code(s) we want to support for each error if we want to differentiate (or just use `1` if not).

lib/helpers/wp-cli-db/class-wp-cli-db.php
```diff
@@ -0,0 +1,117 @@
...
+	
+		$hostname = gethostname();
+		$env      = getenv();
+	
+		if ( ! (
+			Environment::is_sandbox_container( $hostname, $env ) ||
+			Environment::is_batch_container( $hostname, $env )
+		) ) {
+			return;
+		}
```
jblz, line 32: Are these checks useful? Should we also include a check for the local dev env?

WPprodigy (commented):
This is pretty awesome, thanks for working on this.

lib/helpers/wp-cli-db/class-db-server.php
```diff
@@ -0,0 +1,52 @@
...
+		}
+		if ( ! defined( 'DB_USER' ) ) {
+			define( 'DB_USER', $this->user );
+		}
+		if ( ! defined( 'DB_PASSWORD' ) ) {
+			define( 'DB_PASSWORD', $this->password );
+		}
+		if ( ! defined( 'DB_NAME' ) ) {
+			define( 'DB_NAME', $this->name );
+		}
```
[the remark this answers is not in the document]
jblz: Thanks much for taking a look, @WPprodigy!

> Do we have to use the "main" constants

It might be possible to override the arguments that get passed to the WP-CLI invocation, but it's not immediately apparent how to do it:

https://github.com/wp-cli/db-command/blob/de3eddeaa224e617a05fb3a8a97d77fefc456c4f/src/DB_Command.php#L1595-L1597
https://github.com/wp-cli/db-command/blob/de3eddeaa224e617a05fb3a8a97d77fefc456c4f/src/DB_Command.php#L453

> This is something that might interfere with our other WP bootup logic in the future.

These constants are only going to be set in the context of the `wp db` scope / lifetime and WordPress never loads in that context, so it oughtn't affect other things.

> We're initializing multiple DB_Servers.

This code intentionally chooses a single server from the list (if any suitable ones are present in the global) to fulfill the command.

I feel like I'm missing something from your point `B`, though. Can you expand, please?

jblz:
We chatted about this in our team call today and there's definitely more we want to do with this. For example:

* Append comments to SQL statements we can use to include attribution data in queries (e.g. in the [binlog](https://dev.mysql.com/doc/refman/5.7/en/binary-log.html))
* Prevent shell REPL access (e.g. `wp db cli`, or `wp db query` without including a statement) -- we will use an emulated REPL in the vip CLI instead (like we do with `wp shell`).

That said, I think we can push those changes to a follow up so far as they affect mu-plugins since the code is not loaded without the theme code change from the test plan and, even then, it will not be accessible without the Parker piece (& its gated to VIP personnel).

/cc @nickdaugherty

lib/helpers/wp-cli-db/class-db-server.php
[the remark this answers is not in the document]
WPprodigy: Okay I see now. there are multiple initializations of `DB_Server` happening here: https://github.com/Automattic/vip-go-mu-plugins/pull/3064/files#diff-a182b0d077b2b93f40cb08b85e87943fea9086cac0b60c145c93215bfa636148R55-R63

But eventually is filtered down to just one, and only that one has `define_variables()` called on it.

jblz:
Only the query subcommand with a querystring are allowed for now.
Inclusion of an attribution comment (for binlog inspection) has been implemented in the API.

@Automattic/vip-platform-cantina Can y'all please take another look?

jblz:
Noting that this requires https://github.com/Automattic/cron-control-runner/pull/20 on the site for this to work as well as API changes.

lib/helpers/wp-cli-db/class-wp-cli-db.php
jblz: Pulled it out.  The presence of the `WP_CLI` class should suffice.

sjinks (approved):

sjinks:
r2451-stacks

PR #3078, opened 2022-04-20
Show plugin update notices

WPprodigy:
## Description
Updated approach for https://github.com/Automattic/vip-go-mu-plugins/pull/1836/.

Trying to get the WP Core methods to play nice for us whilst `activate_plugins` & `install_plugins` caps are gone is very tricky, felt like walking into a minefield. Every time you get past one block, there is another. As an example, [the core update notice](https://github.com/WordPress/WordPress/blob/9fb5112732c62a33d3e56d15fd402f88a97943fa/wp-admin/includes/update.php#L488-L500) shows a "click here for more details" link - even if the user cannot `update_plugins`. But this link opens up an iframe modal for `/wp-admin/install-plugins.php` that will die because [of permissions issues](https://github.com/WordPress/WordPress/blob/8a90b8691ff7a7cf511ec624c912aacb6dd2b30a/wp-admin/plugin-install.php#L18-L20).

So I opted for just rolling our own system - but still using the same information flow from core so that custom 3rd party plugins can still interact accordingly (seen in the below SS w/ WC Subs). And planning ahead a bit, this new module will be particularly useful when we start to expose more custom functionality that interacts with Codebase Manager later down the road.

versionupdates

### Dev Review Notes

This module will later hold functionality for displaying custom update/delete actions for plugins, eventually themes, and likely language files later as well. So that's the purpose behind the folder structure and architecture.

`PluginsManager` sets up the stuff that needs to be done one time, while individual `Plugin` objects will delegate the specific internal stuffs that each plugin needs to do.

### Plugin Added: Codebase Manager

Shows available plugin update notices in the admin.

## Checklist

## Steps to Test

Go to `/wp-admin/plugins.php` (and `/wp-admin/network/plugins.php` on a MS) as an admin and ensure all looks correct for outdated plugins.


sjinks (approved):

codebase-manager.php
```diff
@@ -0,0 +1,11 @@
+<?php
+/**
+ * Plugin Name: VIP Codebase Manager
+ * Description: Tools for managing the codebase (plugins, themes, etc) within the VIP ecosystem.
+ * Version: 1.0.0
+ * Author: Automattic
+ */
+
+if ( is_admin() && file_exists( __DIR__ . '/codebase-manager/codebase-manager.php' ) ) {
```
GaryJones, line 9: Under what uncontrolled condition might the file not exist? Is this file_exists() check redundant?

mjangda:
@WPprodigy I wonder if this would be a good candidate for an e2e test?

bsessions85:
@mjangda Yeah, I think it would make sense. We would need to always install an outdated version of plugins though, so would need to plan for that.

mjangda:
I think an old version of Hello Dolly is probably the safest bet.

PR #3101, opened 2022-04-26
copy() returns true for non-existing files

rbcorrales:
## Description

The return value of the PHP [copy()](https://www.php.net/manual/en/function.copy.php) function is always true, even when the source file doesn’t exist on the VIP file system. It also creates a zero-byte file at the destination.

This PR addresses the issue by adding a check to the `stream_open()` method of the stream wrapper to throw an error when the mode is `r` (read), and there's an error reading the file (this includes attempts to open a non-existing file).

### Handle file read errors in the stream wrapper

Fixed an issue where `copy()` calls returned `true` for non-existing files.

## Checklist

## Steps to Test

1. Check out PR.
2. Open a `wp` shell
3. Try to copy a non-existing file via the `copy()` function:
```
wp> $from = wp_upload_dir()['basedir'] . '/non-existing-asset.jpg';
string(47) "vip://wp-content/uploads/non-existing-asset.jpg"

wp> $to = wp_upload_dir()['path'] . '/destination.jpg';
string(48) "vip://wp-content/uploads/2022/04/destination.jpg"

wp> copy( $from, $to );
bool(false)
```
4. If the PR is implemented, `copy()` should return `false` instead of `true` and no file should be created in the FS.


sjinks (approved):

rebeccahum:
20220427.0

PR #3103, opened 2022-04-27
Search: Do not alert on backfill of ep_feature_settings option

rebeccahum:
## Description
Just wanting to remove the alerting since so far, it's been a false positive due to race conditions. So far, the option has been added each time.


sjinks (approved):

rebeccahum:
r2464-stacks

PR #3136, opened 2022-05-09
Search: Use dashes for query parameter in forcing index version

rebeccahum:
## Description
Modifies #2340 to use the query parameter with dashes instead of underscores for consistency since we already have the `vip-search-enabled` query parameter.

Modifies query parameter `vip_search_<indexable>_version` to `vip-search-<indexable>-version`.

## Checklist

## Steps to Test
1) Add a new post index version `wp vip-search index-versions add post` (do not index or activate it). For example, we create a new post index version of `4`
2) Pick in a search term that returns results for the current index version. For example, let's use "hello"
3) Append it to the URL `&vip-search-post-version=4`: http://example.com/?s=helllo&vip-search-post-version=4
4) Expect to see no results because the new index version has not been indexed. Remove query parameter `&vip-search-post-version=4` and see results on the current active index version


sjinks (approved):

sjinks:
r2487-stacks

PR #3149, opened 2022-05-13
Search: Add ability to deactivate in index versioning

rebeccahum:
## Description
Since we support the ability to activate in index-versioning, we should also support being able to deactivate them. We do not allow deleting index versions if they are active, so in the use case where one may want to migrate away from Enterprise Search, we should be allowing them the ability to deactivate and then, delete all their indexes if they wanted to.

This PR adds that ability and changes the below functionality:
- `get_active_version_number()` will not return default of 1 if there is no active version at all, as that is not expected. Instead, it will return `false`.
- It is now possible to have no active versions, if the user decides to deactivate them all. If there is no active versions, the index name will be `vip-indexable-no-active-index`, which will throw a `_doing_it_wrong` error when the user attempts to do an ES request.

Also fixes a bug where the error message doesn't display the indexable type when you attempt to activate an already-active index.

It is now possible to deactivate indexes with versioning.

## Checklist

## Steps to Test
1) Create a new index `wp vip-search index-versions add post` (which will add an index 2)
2) Activate the new index `wp vip-search index-versions activate post 2` (this will in turn, deactivate index 1, the old index)
3) Deactivate the new index `wp vip-search index-versions deactivate post 2`
4) Check the current index versions `wp vip-search index-versions list post` and see that none are marked active now


rinatkhaziev:
> so in the use case where one may want to migrate away from Enterprise Search, we should be allowing them the ability to deactivate and then, delete all their indexes if they wanted to.

What are the chances of customers migrating away from search actually bothering to deactivate as opposed to removing the constant from their codebase?

PR #3166, opened 2022-05-17
Add additional data to MemcachePool errors.

emrikol:
## Description

While we have a maximum size of (compressed) data per individual cache key, right now we have no way of determining what keys might be causing these issues at the application layer.

Currently, we will see PHP notices such as this:

> `Notice: MemcachePool::set(): Server example.local. (tcp 12345, udp 0) failed with: SERVER_ERROR object too large for cache (3)`

I'm hoping we can use our custom error handler to extract additional information from the backtrace and add it to the message. I'm also moving this to `E_WARNING` because, in my opinion, this is more in line with the severity of an error like this. Having objects uncached can cause site stability issues.

The new messages would look more like this:

> `Warning: MemcachePool::set(): Server example.local. (tcp 12345, udp 0) failed with: SERVER_ERROR object too large for cache (3) (Key: test123, Group: default, Data Size: 19999999)`

## Checklist

## Steps to Test

I've been testing this in a sandbox using this snippet in a client-mu-plugin:

```php
if ( isset( $_GET['mc-test'] ) && 'thisisabadidea' === $_GET['mc-test'] ) {
	ini_set('display_errors', 1);
	ini_set('display_startup_errors', 1);
	error_reporting(E_ALL);
	$string = substr(str_shuffle(str_repeat($x='12345', ceil(20000000/strlen($x)) )),1,20000000);
	wp_cache_set( 'test123', $string );
	wp_die( 'Ran a large MC test with a size of ' . number_format( strlen( $string ) ) );
}
```

### DIFF
--- a/lib/wpcom-error-handler/wpcom-error-handler.php
+++ b/lib/wpcom-error-handler/wpcom-error-handler.php
@@ -104,6 +104,24 @@ function wpcom_custom_error_handler( $whether_i_may_die, $type, $message, $file,
 		return true;
 	}
 
+	// Capture MemcachePool errors, throw a warning, and provide more debugging data.
+	if (
+		E_NOTICE === $type &&
+		( '/var/www/wp-content/object-cache-stable.php' === $file || '/var/www/wp-content/object-cache-next.php' === $file )
+	) {
+		$type = E_WARNING;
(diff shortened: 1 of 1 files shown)


rinatkhaziev (approved):
A bit scary, but definitely cool :)

rinatkhaziev:
Let's do some manual testing once this gets merged into develop

PR #3169, opened 2022-05-19
Removed the contingency for unprefixed environment variables

jesse-greathouse:
## Description
Removes the use of unprefixed environment variables in the VIP mu-plugins

### Removed the retrieval of unprefixed VIP environment variables

VIP environment variables, enabled through the dashboard or CLI, will now always be prefixed with a VIP string. This will not affect the use of environment variables currently used, but aims to prevent collisions between VIP environment variables and constants that may be defined elsewhere.

## Checklist

## Steps to Test
1. Check out PR.
1. Define a constant in wp-config like `define('MY_CONSTANT', 'foo');`
1. Define an environment variable with the same name, through the dashboard or VIP CLI: `MY_CONSTANT = 'bar'`
1. echo the constant in the code like: `echo MY_CONSTANT`.
2. verify the environment variable did not change the constant to 'bar'

### DIFF
--- a/lib/environment/class-environment.php
+++ b/lib/environment/class-environment.php
@@ -20,14 +20,8 @@ class Environment {
 		$key       = strtoupper( $key );
 		$env_const = sprintf( '%s_%s', self::VAR_PREFIX, $key );
 
-		// First check to see if the VIP_ENV_VAR constant exists and return it
-		// If constant does not exist fallback to const without the prefix
-		// If neither constant exist, return default
 		if ( defined( $env_const ) ) {
 			return constant( $env_const );
-		} elseif ( defined( $key ) ) {
-			// interim migration step
-			return constant( $key );
 		}
 
 		// The call was not able to retrieve an env variable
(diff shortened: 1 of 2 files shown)


sjinks (approved):

chriszarate (dismissed):
This looks good, but please hold on deploying. There is an accompanying GOOP change and we also need to triple-check that other customers are not relying on the legacy behavior anymore. Thanks!

chriszarate:
@jesse-greathouse This is ok to be deployed. (It was actually always ok to deploy. It doesn't have any effect or dependency on whether the unprefixed constant is defined, which is done in GOOP.)

sjinks (changes requested):
Could you please take a look at the failing tests?

rinatkhaziev:
@jesse-greathouse this is getting a bit stale. Can we please get it over the finish line?

jesse-greathouse:
Opened a new PR #3498

PR #3171, opened 2022-05-20
Fix flaky tests

sjinks:
It looks like `lastpostmodified_Test` is sensitive to the order in which our tests run and may fail if other tests update a post and run before it.

This PR tries to fix this behavior.

### DIFF
--- a/tests/test-performance-lastpostmodified.php
+++ b/tests/test-performance-lastpostmodified.php
@@ -30,12 +30,6 @@ class lastpostmodified_Test extends WP_UnitTestCase {
 		$this->assertEquals( 1, did_action( 'wpcom_vip_bump_lastpostmodified' ) );
 	}
 
-	public function test__transition_post_status__save_on_update() {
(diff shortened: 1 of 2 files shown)


tests/test-performance-lastpostmodified.php
```diff
@@ -20,12 +25,6 @@ public function test__transition_post_status__save_on_publish() {
 		$this->assertEquals( 1, did_action( 'wpcom_vip_bump_lastpostmodified' ) );
 	}
 
-	public function test__transition_post_status__save_on_update() {
-		\wp_transition_post_status( 'publish', 'publish', $this->post );
-
-		$this->assertEquals( 1, did_action( 'wpcom_vip_bump_lastpostmodified' ) );
-	}
```
sjinks, line 27: Exactly the same implementation as in `test__transition_post_status__save_on_publish()`

pschoffer (approved):
Looks great

PR #3059, opened 2022-04-15
Fix to enable MU plugins when running core tests

sjinks:

PR #3176, opened 2022-05-22
Check if headers has already been sent before sending new ones

sjinks:
This PR fixes several tests by adding a check whether the headers have already been sent before invoking a function that sends additional headers.

This primarily affects WP Core Tests and the cases when it is impossible to mock `header()` and related functions.

Split from #3059


pschoffer (approved):
This change makes sense to me even if it would not be done for tests sake.

sjinks:
r2504-stacks

PR #3177, opened 2022-05-22
Fix PHP warnings

sjinks:
Fix warnings that manifest themselves under some rare circumstances.

Related: #3059


pschoffer (approved):

sjinks:
r2505-stacks

PR #3178, opened 2022-05-22
Skip problematic plugins on WP installation

sjinks:
This PR disables the loading of several problematic plugins when `WP_INSTALLING` is `true`. The affected plugins access the database before WP has a chance to create tables, which leads to unpleasant error messages.

Related: #3059

### DIFF
--- a/jetpack.php
+++ b/jetpack.php
@@ -175,4 +175,6 @@ function vip_jetpack_load() {
 	}
 }
 
-vip_jetpack_load();
+if ( ! defined( 'WP_INSTALLING' ) ) {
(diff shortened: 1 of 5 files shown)


pschoffer:
I am not sure on that one? What is the scenario when WP is still installing? Is that only in core tests? It feels safer to have error than missing plugins in prod.

sjinks:
It happens only when you run `wp core install`. The plugins won't be missing, they just won't load during that step.

pschoffer (approved):

sjinks:
r2506-stacks

PR #2913, opened 2022-02-16
Search: add a few headers to allow to better identify the requests

rinatkhaziev:
## Description

To allow us to better identify the requests across various logs it makes sense to pass through a few client-related requests (particularly, User-agent and IP addresses should allow us to easily identify abusive activity.

### Plugin Updated: Enterprise Search

Logging improvements that would allow us to better identify potential abusive activity.

## Checklist

Please make sure the items below have been covered before requesting a review:

- [x] This change works and has been tested locally (or has an appropriate fallback).
- [x] This change works and has been tested on a Go sandbox.
- [ ] This change has relevant unit tests (if applicable).
- [ ] This change has relevant documentation additions / updates (if applicable).
- [x] I've created a changelog description that aligns with the provided examples.

PR #3190, opened 2022-05-27
Fix regression that was introduced by varying the  by adding unique headers, use body of the query to calculate the hash

rinatkhaziev:
## Description

In #2913 we added a few headers to be able to identify requests along their lifecycle better, this, however, had an unfortunate side-effect of varying the resulting `$cache_key`.
This PR addresses that by only using the query body to determine the cache key hash.

Fix a regression that resulted in request cache using unique keys thus not using cached responses.

## Checklist

## Steps to Test

On a sandbox: 

1. Checkout the PR
2. Search for something and note the HTTP request made 
3. Refresh the page, verify the request is not made anymore but search results are present

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -906,21 +906,13 @@ class Search {
 		if ( $is_cacheable ) {
 			$cache_key = 'es_query_cache:' . md5( $query['url'] . $args['body'] ) . ':' . wp_cache_get_last_changed( self::SEARCH_CACHE_GROUP );
 			/**
-			 * Serve cached response right away, if available and not stale and the query is cacheable
+			 * Serve cached response right away, if available
 			 */
-			$cached_response = $is_cacheable ? wp_cache_get( $cache_key, self::SEARCH_CACHE_GROUP ) : false;
+			$cached_response = wp_cache_get( $cache_key, self::SEARCH_CACHE_GROUP );
 
 			if ( $cached_response ) {
 				return $cached_response;
 			}
-		}
-
-		/**
-		 * Serve cached response right away, if available and not stale and the query is cacheable
-		 */
-		$cached_response = $is_cacheable ? wp_cache_get( $cache_key, self::SEARCH_CACHE_GROUP ) : false;
-		if ( $cached_response ) {
-			return $cached_response;
 		}
 
 		$start_time = microtime( true );
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -902,7 +902,9 @@ class Search {
 		$statsd_prefix          = $this->get_statsd_prefix( $query['url'], $statsd_mode );
 
 		// Cache handling
-		$is_cacheable = $this->is_url_query_cacheable( $query['url'], $args );
+		$is_cacheable    = $this->is_url_query_cacheable( $query['url'], $args );
+		$cached_response = false;
+
 		if ( $is_cacheable ) {
 			$cache_key = 'es_query_cache:' . md5( $query['url'] . $args['body'] ) . ':' . wp_cache_get_last_changed( self::SEARCH_CACHE_GROUP );
 			/**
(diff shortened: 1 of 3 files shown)


search/includes/classes/class-search.php
```diff
@@ -903,7 +903,17 @@ public function filter__ep_do_intercept_request( $request, $query, $args, $failu
 
 		// Cache handling
 		$is_cacheable = $this->is_url_query_cacheable( $query['url'], $args );
-		$cache_key    = 'es_query_cache:' . md5( $query['url'] . wp_json_encode( $args ) ) . ':' . wp_cache_get_last_changed( self::SEARCH_CACHE_GROUP );
+		if ( $is_cacheable ) {
+			$cache_key = 'es_query_cache:' . md5( $query['url'] . $args['body'] ) . ':' . wp_cache_get_last_changed( self::SEARCH_CACHE_GROUP );
+			/**
+			 * Serve cached response right away, if available and not stale and the query is cacheable
+			 */
+			$cached_response = $is_cacheable ? wp_cache_get( $cache_key, self::SEARCH_CACHE_GROUP ) : false;
```
rebeccahum, line 911: `$is_cacheable` will never be false b/c of the check in L906. I think we can just go straight to the source via `wp_cache_get()`.

rebeccahum (changes requested):

search/includes/classes/class-search.php
```diff
@@ -1017,7 +1021,7 @@ public function is_url_query_cacheable( string $url, $args ): bool {
 			return $is_cacheable;
 		}
 
-		$is_cacheable = (bool) preg_match( '#/_(search|mget|doc)#', $url );
+		$is_cacheable = (bool) preg_match( '#/_(search|mget|doc)#', $url ) && in_array( $args['method'], [ 'DELETE', 'PUT' ], true );
```
rebeccahum, line 1024: Aren't search requests a `POST` request?

rebeccahum (changes requested):

search/includes/classes/class-search.php
rebeccahum: Or was it intended for an `! in_array()` based on the `Skip DELETE and PUT from cacheable methods` comment?

rinatkhaziev: typo!

rinatkhaziev:
whew, what a struggle :)

search/includes/classes/class-search.php
```diff
@@ -902,15 +902,21 @@ public function filter__ep_do_intercept_request( $request, $query, $args, $failu
...
-		/**
-		 * Serve cached response right away, if available and not stale and the query is cacheable
-		 */
-		$cached_response = $is_cacheable ? wp_cache_get( $cache_key, self::SEARCH_CACHE_GROUP ) : false;
-		if ( $cached_response ) {
-			return $cached_response;
+		if ( $is_cacheable ) {
+			// Some requests may not have body
+			$body      = $args['body'] ?? '';
+			$cache_key = 'es_query_cache:' . md5( $query['url'] . $body ) . ':' . wp_cache_get_last_changed( self::SEARCH_CACHE_GROUP );
```
rebeccahum, line 911: Since `$cache_key` is only defined in this block here, will it be a `null` value when it comes to line 1000 at `wp_cache_set( $cache_key, $response, self::SEARCH_CACHE_GROUP, 5 * MINUTE_IN_SECONDS );`?

rebeccahum (changes requested):

search/includes/classes/class-search.php
rinatkhaziev: It's wrapped in the similar `$is_cacheable` check so I think it should be fine?

rebeccahum (approved):

rinatkhaziev:
r2517-stacks

sjinks:
v20220531.0

Issue #3179, opened 2022-05-23
VIP Admin Notices do not persist dismissed notice


dougaxe1:
## Expected/Desired Behavior

Upon dismissing a VIP Admin Notice (`Automattic\VIP\Admin_Notice\Admin_Notice`), the dismissal should be persisted to user meta.

## Actual Behavior

The dismissal is not persisted in user meta due to a timing bug in `admin-notice/js/script.js`. On window load `registerDismissHooks()` is called which attaches click events to each notice based on the selector `.vip-notice .notice-dismiss`. The `.notice-dismiss` button is added later during the WordPress core `makeNoticesDismissible()`.

## Steps to Reproduce the Problem

1. Create a VIP Admin Notice with matching conditions
2. Dismiss the notice via the UI (observe lack of network request)
3. Refresh the page
4. See same notice, again ☹️.

## (Optional) Additional notes

`makeNoticesDismissible()` is triggered by many events, including `$document.trigger( 'wp-updates-notice-added' );`. One option may be to use the same event to trigger running `registerDismissHooks()`. Another option would be to use a more general "live" selector capturing bubbling click events within the notice.

sjinks:
Fixed in [d19bf3d](https://github.com/Automattic/vip-go-mu-plugins/commit/d19bf3d593604449dd759ce93e590e20b6c13177)

PR #3192, opened 2022-05-27
GH-3179: more robust notice dismissal

sjinks:
## Description

See: #3179

This PR attempts to fix race conditions happening when `common.js` gets loaded after `admin-notice/js/script.js`

### Plugin Updated: Admin Notice

More robust notice dismissal

## Checklist

## Steps to Test

See #3179

### DIFF
--- a/admin-notice/class-admin-notice-controller.php
+++ b/admin-notice/class-admin-notice-controller.php
@@ -31,7 +31,7 @@ class Admin_Notice_Controller {
 	}
 
 	public function enqueue_scripts() {
-		wp_enqueue_script( 'vip-admin-notice-script', plugins_url( '/js/script.js', __FILE__ ), [], '1.0', true );
+		wp_enqueue_script( 'vip-admin-notice-script', plugins_url( '/js/script.js', __FILE__ ), [ 'common' ], '1.1', true );
 		wp_localize_script( 'vip-admin-notice-script', 'dismissal_data', [
 			'nonce'          => wp_create_nonce( self::DISMISS_NONCE_ACTION ),
 			'data_attribute' => Admin_Notice::DISMISS_DATA_ATTRIBUTE,
--- a/admin-notice/js/script.js
+++ b/admin-notice/js/script.js
@@ -1,14 +1,7 @@
-function tryGetNoticeContainer(currentElement) {
-	if (currentElement === null || currentElement.hasAttribute(dismissal_data.data_attribute)) {
-		return currentElement;
-	}
-	return tryGetNoticeContainer(currentElement.parentElement);
-}
-
 function persistDismissedNotice(containerElement) {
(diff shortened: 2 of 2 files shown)


admin-notice/class-admin-notice-controller.php
```diff
@@ -31,7 +31,7 @@ public function display_notices() {
 	}
 
 	public function enqueue_scripts() {
-		wp_enqueue_script( 'vip-admin-notice-script', plugins_url( '/js/script.js', __FILE__ ), [], '1.0', true );
+		wp_enqueue_script( 'vip-admin-notice-script', plugins_url( '/js/script.js', __FILE__ ), [ 'common' ], '1.1', true );
```
sjinks, line 34: Dismiss buttons get added by `common.js`; therefore, it makes sense to list it as a dependency. This also ensures that `common.js` will be loaded before `script.js`.

admin-notice/js/script.js
```diff
@@ -21,19 +14,18 @@ function persistDismissedNotice(containerElement) {
 	}
 }
 
-function onDismissed(event) {
-	const noticeContainer = tryGetNoticeContainer(event && event.target);
-	if (noticeContainer) {
-		persistDismissedNotice(noticeContainer);
+function vipNoticeClickHandler( ev ) {
+	const button = ev.target.closest( '.notice-dismiss' );
```
sjinks, line 18: [closest()](https://caniuse.com/element-closest) is supported by all browsers supporting [fetch()](https://caniuse.com/fetch).

admin-notice/js/script.js
```diff
@@ -21,19 +14,18 @@ function persistDismissedNotice(containerElement) {
...
+	if ( button ) {
+		const noticeContainer = button.closest( `[data="${dismissal_data.data_attribute}"]` );
```
sjinks, line 20: [Template literals](https://caniuse.com/template-literals) are supported even better than `fetch()`.

admin-notice/js/script.js
```diff
@@ -21,19 +14,18 @@ function persistDismissedNotice(containerElement) {
...
 	}
 }
 
 function registerDismissHooks() {
-	const notices = document.querySelectorAll( '.vip-notice .notice-dismiss' );
-
-	for ( const notice of notices ) {
-		notice.addEventListener( "click", onDismissed );
-	}
+	document.querySelectorAll( '.vip-notice' ).forEach( notice => notice.addEventListener( 'click', vipNoticeClickHandler ) );
```
sjinks, line 28: Because events bubble, we can set up an event listener on a parent element and then check which child was clicked.

This also prevents a race condition when we set up event listeners but the dismiss button is not there yet.

pschoffer (dismissed):
Oh that `closest` is pretty handy! Didn't know about that before.

pschoffer (approved):

sjinks:
r2519-stacks

sjinks:
v20220531.0

PR #3195, opened 2022-05-30
Ensure we are using the latest WP branch release

sjinks:
Now that Automattic/vip-container-images#237 has been deployed, we can update our workflows to use `.x` WP revisions to make sure that we use the latest WordPress release available for the given branch.

No more manual workflow updates when a new WP patch is released.


pschoffer (approved):
:heart:

PR #3197, opened 2022-05-31
Search: Fix bug where health validate-counts on multisite returns incorrect results

rebeccahum:
## Description
We currently cache if an index exists via an option: https://github.com/Automattic/vip-go-mu-plugins/blob/6292ffbc0a6202c3070dd629bd50ac3babd610ab/search/includes/classes/class-search.php#L999-L1002

In multisite, we should be switching into the subsite context BEFORE checking if the index exists. If not, this will result in the option being stored in site 1, which will also return it after as not existing subsequently after any changes are made when `wp vip-search health validate-counts --network-wide` is run. This is because the invalidation logic below will not fire:
https://github.com/Automattic/vip-go-mu-plugins/blob/6292ffbc0a6202c3070dd629bd50ac3babd610ab/search/includes/classes/class-search.php#L846

Fix bug where health validate-counts command was not returning correct results when using network-wide flag.

## Checklist

## Steps to Test
1) Create multisite with main site and a subsite
2) Index only the main site: `wp vip-search index --setup`
3) Run `wp vip-search health validate-counts --network-wide` to store the option incorrectly on the main site
4) Index now the subsite: `wp vip-search index --setup --url=subsite.example.com`
5) Run `wp vip-search health validate-counts --network-wide` again to see the incorrect results being returned for the subsite
6) Validate with `wp vip-search health validate-counts --url=subsite.example.com` that the index does infact, exist
7) Apply the PR
8) Run `wp vip-search health validate-counts --network-wide` and see the expected results

### DIFF
--- a/search/includes/classes/commands/class-healthcommand.php
+++ b/search/includes/classes/commands/class-healthcommand.php
@@ -148,6 +148,8 @@ class HealthCommand extends \WPCOM_VIP_CLI_Command {
 					continue;
 				}
 
+				switch_to_blog( $site['blog_id'] );
+
 				if ( ! $indexable->index_exists( $site['blog_id'] ) ) {
 					$blog_id = $site['blog_id'];
 					WP_CLI::line( "Skipping validation of '$indexable_slug' index for site $blog_id as it doesn't exist.\n\n" );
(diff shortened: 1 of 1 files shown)


pschoffer (approved):

rebeccahum:
r2532-stacks

sjinks:
v20220531.0

PR #3235, opened 2022-06-07
Show plugin vulnerabilities in wp-admin

WPprodigy:
## Description

Followup to https://github.com/Automattic/vip-go-mu-plugins/pull/3078 - alongside update notices now any known vulnerabilities are also shown:

vulns

### Plugin Updated: Codebase Manager

Nows shows known plugin vulnerabilities in the plugins list in WP Admin.

## Checklist

## Steps to Test

Live:
1. Apply the two file changes to a Go sandbox (`297` is a good option).
1. Go to the plugins page and ensure all works correctly.

Dev:
1. Apply the PR.
1. Set up api-gw & ccm locally, and update relevant services api constants.
1. Or ping me :)

### DIFF
--- a/codebase-manager/plugins/plugin.php
+++ b/codebase-manager/plugins/plugin.php
@@ -29,7 +29,7 @@ class Plugin {
 		}
 
 		printf( '<tr class="plugin-update-tr%s">', $this->is_active() ? ' active' : '' );
-		printf( '<td colspan="%s" class="plugin-update colspanchange%s">', esc_attr( $number_of_columns ), $this->has_vulnerabilities() ? ' hide-box-shadow' : '' );
+		printf( '<td colspan="%d" class="plugin-update colspanchange%s">', esc_attr( $number_of_columns ), $this->has_vulnerabilities() ? ' hide-box-shadow' : '' );
 		print( '<div class="update-message notice inline notice-warning notice-alt"><p>' );
 
 		// In the future, could do the fancy iframe that core does here. It also allows 3rd party plugins to hook in with `plugins_api` filter to better expose their changelogs.
@@ -53,7 +53,7 @@ class Plugin {
 		}
 
 		printf( '<tr class="plugin-vuln-tr%s">', $this->is_active() ? ' active' : '' );
-		printf( '<td colspan="%s" class="plugin-vuln colspanchange">', esc_attr( $number_of_columns ) );
+		printf( '<td colspan="%d" class="plugin-vuln colspanchange">', esc_attr( $number_of_columns ) );
 		print( '<div class="vuln-message error inline notice-error notice-alt">' );
 
 		/* translators: 1: The number of vulnerabilities associated with the plugin. */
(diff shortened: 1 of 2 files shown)


codebase-manager/plugins/plugin.php
```diff
@@ -25,7 +29,7 @@ public function display_version_update_information( int $number_of_columns ): vo
 		}
 
 		printf( '<tr class="plugin-update-tr%s">', $this->is_active() ? ' active' : '' );
-		printf( '<td colspan="%s" class="plugin-update colspanchange">', esc_attr( $number_of_columns ) );
+		printf( '<td colspan="%s" class="plugin-update colspanchange%s">', esc_attr( $number_of_columns ), $this->has_vulnerabilities() ? ' hide-box-shadow' : '' );
```
sjinks: Why not just use `%d` or `%u` for `colspan`?

codebase-manager/plugins/plugin.php
```diff
@@ -43,6 +47,46 @@ public function display_version_update_information( int $number_of_columns ): vo
...
+
+		echo '<ul>';
+		foreach ( $this->vulnerabilities as $vuln ) {
+			// "None" could be confusing, let's convert it to "Low".
+			$severity = 'NONE' === $vuln->severity ? 'Low' : ucwords( strtolower( $vuln->severity ) );
+
+			$severity_info = sprintf( '%s severity', esc_html( $severity ) );
+			if ( null !== $vuln->severityScore ) {
+				$severity_info = sprintf( '%s severity: %s/10', esc_html( $severity ), esc_html( $vuln->severityScore ) );
+			}
```
sjinks: If we use `_n()` for `$message`, it probably makes sense to use translation functions here as well.

codebase-manager/plugins/plugin.php
```diff
@@ -43,6 +47,46 @@ public function display_version_update_information( int $number_of_columns ): vo
...
+			'There are %s known vulnerabilities for this plugin on the currently installed version:',
+			count( $this->vulnerabilities )
+		), number_format_i18n( count( $this->vulnerabilities ) ) );
+
+		echo '<p>' . esc_html( $message ) . '</p>';
+
+		echo '<ul>';
+		foreach ( $this->vulnerabilities as $vuln ) {
+			// "None" could be confusing, let's convert it to "Low".
+			$severity = 'NONE' === $vuln->severity ? 'Low' : ucwords( strtolower( $vuln->severity ) );
```
sjinks, line 71: In which cases the severity could be `NONE`? Does it make sense to use `Unknown` instead of `Low` for such a case?

WPprodigy: So "none" is a valid CVSS v2 score: https://nvd.nist.gov/vuln-metrics/cvss. [This recent vuln](https://wpscan.com/vulnerability/2c3f9038-632d-40ef-a099-6ea202efb550) got the 0.0/none rating.

"Unknown" dates back to vulns before WPscan was tracking it mainly.

codebase-manager/plugins/plugin.php
WPprodigy: No particular reason, maybe copy/pasta being the main one 😆. Ultimately it does end up being a string, but I can swap it.

codebase-manager/plugins/plugin.php
sjinks: That recent vulnerability became of [Medium severity](https://nvd.nist.gov/vuln/detail/CVE-2021-25118).

My concern is that if we set the severity to Low, the customer may think it is unimportant and ignore/postpone the fix. But the vulnerability could get reevaluated (when researchers have more information) and get higher severity.

WPprodigy: When WPScan updates, we'll have the updated information to show in the UIs as well. I don't think "None" is always a mistake necessarily, it could be something that just isn't exploitable by itself?

In either case, all we can do is show the data we know about from the API.

PR #3236, opened 2022-06-08
Changelog creation

pschoffer:
This PR introduces a GitHub Action that on merge to staging creates a draft changelog that contains all the changelog entries from PRs since the last changelog.

It's not bullet proof, but how it works it lists or the closed prs ordering by updated the latest.. then walks through them and ignores all that have no changelog or `[Status] No files to Deploy`. It does that as long till it hits the first one with `[Status] Deployed to <branch on which it is running>`.

Then combines the changelog entries into one WP post.


ci/changelog-summary.php
```diff
@@ -0,0 +1,240 @@
...
+        exit( 0 );
+    }
+
+    $title = 'Staging deploy - ' . date("Y/m/d");
+    $content = join( "\n<hr />\n", $changelog_entries );
+
+    create_changelog_post( $title, $content, $changelog_tags, $changelog_channels );
+}
+
+create_changelog_summary();
```
pschoffer:
```suggestion
create_changelog_summary();

```

pschoffer:
It now also marks all the PRs with GitHub label `Deployed to <branch>` after its done.

.github/workflows/changelog-summary.yml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # tag=v3.0.2
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '7.4'
+
+      - name: Install
+        run: composer install -n
```
sjinks:
```suggestion
      - name: Install
        uses: ramsey/composer-install@2.1.0
```

sjinks (approved):

.github/workflows/changelog-summary.yml
```diff
@@ -0,0 +1,42 @@
...
+
+on:
+  push:
+    branches:
+      - staging
+      - update/ci
+
+permissions:
+  contents: read
+  pull-requests: write
```
sjinks, line 11: BTW, do we really need this permission?

sjinks: I mean, maybe `read` is enough?

pschoffer: We are changing labels on PR. I think that's why we need write.

sjinks: Makes sense

PR #3238, opened 2022-06-08
GH-2267: disable IRC for dev env

### LINKED ISSUE #2267: Local error message about missing IRC host configuration
## Expected/Desired Behavior

Fill with either one:

I think I would expect not to have IRC mentions in my logs if I haven't done anything to show that IRC is a desired method for communicating.

## Actual Behavior

How does the application currently work.

I see a regular log message like this: `[04-Jul-2021 06:23:06 UTC] Missing IRC host configuration in ALERT_SERVICE_ADDRESS constant`. It comes from [these lines](https://github.com/Automattic/vip-go-mu-plugins/blob/master/vip-helpers/vip-utils.php#L1531-L1535).

1. I'm not sure how to reproduce, but here's how it happens for me. It doesn't happen on every request, but it does run more often than once a day. I notice it most often when I shut my dev computer for the day and then come back to it the next day, and the logs have a lot of this entry in them.

sjinks:
## Description

This PR disables IRC for the dev environment -- we don't provide any IRC server anyway.

Fixes #2267

### Plugin Updated: VIP Init

Disable IRC Alerts for the dev environment

## Checklist

### DIFF
--- a/lib/utils/class-alerts.php
+++ b/lib/utils/class-alerts.php
@@ -54,6 +54,9 @@ class Alerts {
 	 */
 	protected function send( array $body ) {
 		$fallback_error = new WP_Error( 'alerts-send-failed', 'There was an error connecting to the alerts service' );
+		if ( empty( $this->service_url ) ) {
+			return $fallback_error;
+		}
 
 		$response = vip_safe_wp_remote_request( $this->service_url, $fallback_error, 3, 1, 10, [
 			'method' => 'POST',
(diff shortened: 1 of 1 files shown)


pschoffer:
I am not a big fan of changing the production code in order to make tests or in this case local dev-env work. Is there a way we can disable it differently by adding some filter or something to the dev-env plugin?
https://github.com/Automattic/vip-container-images/blob/master/dev-tools/dev-env-plugin.php

pschoffer (changes requested):
The above comment

sjinks:
I have updated the code to silence the error for the dev env if no IRC server has been configured.

This allows the users to use their own URC server should they want to, and won't pollute the error log if they have chosen not to use any IRC server (the default).

pschoffer (approved):

PR #3266, opened 2022-06-15
Speed up "Lint PHP files" workflow

sjinks:
If we recursively check out all submodules, it takes ages for `php-lint` to scan all files.

This PR:
1. Turns off module checkout; as a result, the linter has much fewer files to check (it does not make sense to check 3rd party modules anyway)
2. Turns off linting for Jetpack and Parse.ly

These changes reduce the run time from 06:50 down to 01:15. Not bad.


pschoffer (approved):
Not bad indeed!

PR #3274, opened 2022-06-20
Add coverage reporting

sjinks:
This PR adds a generation of a code coverage report to some tests. These reports are then uploaded to Codecov (we can explore them there and see how far we are from the ideal coverage). Codecov comments on the PR and says how that PR contributes to the global coverage and whether all changes in the PR are covered with tests.

This will help us identify places we should cover with tests.


jesse-greathouse (approved):
👍

PR #3275, opened 2022-06-20
Update docs in README.md to use develop branch

naxoc:
## Description

It looks like this repo is using the `main` branch instead of `master` now, so this is just a PR to update the docs in the README.

## Steps to Test

This is docs only, but please do make sure I got things right :)


README.md
```diff
@@ -121,4 +121,4 @@ bin/create-release.sh
 
 This is a repo primarily meant for local non-development use.
 
-Every commit merged into `master` is automatically pushed to the public copy at [Automattic/vip-go-mu-plugins-built](https://github.com/Automattic/vip-go-mu-plugins-built/). This is handled via CI by the [`deploy.sh` script](https://github.com/Automattic/vip-go-mu-plugins/blob/master/ci/deploy.sh) script, which builds pushes a copy of this repo and expanded submodules.
+Every commit merged into `main` is automatically pushed to the public copy at [Automattic/vip-go-mu-plugins-built](https://github.com/Automattic/vip-go-mu-plugins-built/). This is handled via CI by the [`deploy` action](https://github.com/Automattic/vip-go-mu-plugins/blob/develop/.github/workflows/deploy.yml), which builds pushes a copy of this repo and expanded submodules.
```
sjinks: I am afraid we don't have the `main` branch.

Looking [at the workflow](https://github.com/Automattic/vip-go-mu-plugins/blob/d21bfcf04f77506497126a9bd7ead85412276a14/.github/workflows/deploy.yml#L5-L7), I see that the workflow is triggered by a push into the `develop` branch.

```suggestion
Every commit merged into `develop` is automatically pushed to the public copy at [Automattic/vip-go-mu-plugins-built](https://github.com/Automattic/vip-go-mu-plugins-built/). This is handled via CI by the [`deploy` action](https://github.com/Automattic/vip-go-mu-plugins/blob/develop/.github/workflows/deploy.yml), which pushes a copy of this repo and expanded submodules.
```

sjinks (changes requested):

README.md
naxoc: Woops, yes. That is indeed on push to `develop`. Out of curiosity, what is the `main` branch used for in this repo? https://github.com/Automattic/vip-go-mu-plugins/tree/main

sjinks: Good question, I was not aware of it 😅

@pschoffer, could you please help?

sjinks (approved):

PR #3297, opened 2022-06-26
[CANTINA-21] Do not hardcode Jetpack IPs

### LINKED ISSUE #1563: Update vip_is_jetpack_request() IP Addresses
These are outdated: https://github.com/Automattic/vip-go-mu-plugins/blob/be41b3572428f257db4d5803dac70824626d8293/vip-helpers/vip-utils.php#L1379-L1405

Current list is here: https://jetpack.com/support/hosting-faq/#jetpack-whitelist

Could also potentially request and cache these IPs from here: https://jetpack.com/ips-v4.json, falling back to a hardcoded list if the request fails maybe?

sjinks:
## Description

This PR fixes SonarCloud's complaints about the hardcoded IP addresses we use in `vip_is_jetpack_request()`. Instead of hardcoding those IPs, we retrieve them from the [official endpoint](https://jetpack.com/support/how-to-add-jetpack-ips-allowlist/). We cache the list using Transients API and periodically update it using WP Cron.

Fixes: #1563

Do not hardcode Jetpack IPs; instead, retrieve them from the official API endpoint.

## Checklist

## Steps to Test

The CI should pass.


vip-helpers/vip-utils.php
```diff
@@ -1494,24 +1500,10 @@ function vip_is_jetpack_request() {
...
-	// From https://jetpack.com/support/hosting-faq/#jetpack-whitelist
-	// Or https://jetpack.com/ips-v4.json
-	$jetpack_ips = array(
-		'122.248.245.244/32',
-		'54.217.201.243/32',
-		'54.232.116.4/32',
-		'192.0.80.0/20',
-		'192.0.96.0/20',
-		'192.0.112.0/20',
-		'195.234.108.0/22',
```
sjinks, line 1506: These IPs come from the JSON.

vip-helpers/vip-utils.php
```diff
@@ -1494,24 +1500,10 @@ function vip_is_jetpack_request() {
...
-		'192.0.96.202/32',
-		'192.0.98.138/32',
-		'192.0.102.71/32',
-		'192.0.102.95/32',
```
sjinks, line 1510: These ones belong to the 192.0.96.0/20 range.

I assume it is safe to remove them?

pschoffer:
Looks smart to me, but I don't really understand that JP flow. @rinatkhaziev , @rebeccahum , @WPprodigy do you have an input on this one?

rinatkhaziev (changes requested):
I don't think that using object cache is a good strategy here - since it may fall off at any given moment. Probably it's better to store as an autoloaded option. The IPs don't change often.

WPprodigy:
I agree regarding saving as an option, but also could strip out the cron job even as it seems a little overkill IMO.

```
$option = get_option( 'option_name', [ 'data' => [], expiration => 0 ]  );

if ( empty( $option['data'] ) || $option['expiration'] < time() ) {
  $IPs = wp_remote_request();
  update_option( 'option_name',  [ 'data' => $IPs, expiration => time() + DAY_IN_SECONDS ] );
}

return $option['data'];
```

rinatkhaziev:
@WPprodigy what happens when your approach fails due to a network connectivity issue?

WPprodigy:
We don't need to update the option unless the remote request worked. Should always fall back to just returning from the data from the option.

sjinks:
@WPprodigy if we do requests in the foreground, won't it negatively affect load times?

rinatkhaziev:
> We don't need to update the option unless the remote request worked. Should always fall back to just returning from the data from the option.

Ok but imagine a scenario where option is missing and the request has failed, how would that potentially affect a request?

WPprodigy:
So before we ever have a successful request? Could only ever happen early on in an app's life, and I mean, that's doomed either way if we can never get the list in the first place. So if the request has only ever failed then the option's IP list remains empty and `vip_is_jetpack_request()` will just have to return `false` to be safe. Worst case scenario is that the brand new site won't initially be able to connect to JP during maintenance mode. Doesn't seem super problematic IMO.

> if we do requests in the foreground, won't it negatively affect load times?

Could swap it to doing the lookup/option update on `shutdown` if it's expired. So the current request will use the existing list and on shutdown it'll check if it needs to update. Ultimately not that different from cron - I'm just pretty opposed to an entirely new cron job for such a tiny task.

Rough mockup / example:

```
function get_ips() {
	$option = get_option( 'option_name', [ 'data' => [], 'expiration' => 0 ] );

	if ( empty( $option['data'] ) || $option['expiration'] < time() ) {
		add_action( 'shutdown', 'update_ips' );
	}

	return $option['data'] ?? [];
}

function update_ips() {
	if ( ! wp_cache_add( 'some_lock_key', true, 'vip', MINUTE_IN_SECONDS ) ) {
		return;
	}

	$IPs = wp_remote_request();
	// do stuff with the request, etc.

	if ( $request_was_successfull ) {
		update_option( 'option_name',  [ 'data' => $IPs, expiration => time() + DAY_IN_SECONDS ] );
	}
}
```

vip-helpers/class-jetpack-ip-manager.php
```diff
@@ -0,0 +1,65 @@
+<?php
+
+namespace Automattic\VIP\Utils;
+
+class JetPack_IP_Manager {
```
nickdaugherty: Jetpack has a lowercase P ... `_lowercase_p_dangit()`! 😄

sjinks:
I went with the mixed approach way.

WPprodigy (dismissed):
Thanks for the changes, looks good!

rinatkhaziev (changes requested):
Thoughts on actually not doing it on the shutdown but rather just checking at the call site (vip_is_jetpack_request)? Since this is only needed for machines to talk to each other tacking 50ms+ on random requests seems to be a bit much?

WPprodigy:
Pretty sure it is only triggered at the call site still.

sjinks:
Done but as far as I can tell, the only change is that we refresh the IPs right when `vip_is_jetpack_request()` is called vs we deferred the update until `shutdown`.

WPprodigy:
I kinda liked the deferred version, but tbh probably doesn't make any notable difference to justify the complexity increase.

I would however still consider having the `wp_cache_add()` lock, to prevent a possible stampede of requests all trying to refresh the option at the same time.

rinatkhaziev (approved):

PR #3298, opened 2022-06-26
Refactor CI

sjinks:
* Create a composite action to run tests (shared by CI workflows)
* Create a workflow that reports coverage to CodeCov on pushes top the default branch


pschoffer (approved):
Najs

PR #3299, opened 2022-06-27
Lockout plugin: Fix capability checks on wp-admin notices

vaurdan:
## Description

When the lockout plugin is enabled, in certain scenarios the lockout notice does not display because it fails the capabilities check. Since it also removes all the capabilities - that can also be done by other plugins, such as the 2FA plugin - and we're only showing the notice to users with `manage_options` and `edit_posts`, the message is never displayed (unless the user is an Automattician).

By accessing directly the `allcaps` array, we can check the original user's capabilities, and make sure that it had the required capabilities (`manage_options` and `edit_posts`) to display the notices. Using `has_cap` in this scenario can be unreliable as the capabilities might have been filtered somewhere else.

### Lockout: Check for user capabilities using `allcaps` instead of `has_cap`.

Changes the logic that decides if lockout notices should be displayed depending on the user capabilities.

## Checklist

## Steps to Test

1. Check out PR.
1. Set `VIP_LOCKOUT_STATE` environment variable to `warning` and `VIP_LOCKOUT_MESSAGE` to some string. (or, if using a sandboxed site, you can use the CLI subcommand)

### DIFF
--- a/security/class-lockout.php
+++ b/security/class-lockout.php
@@ -45,7 +45,8 @@ class Lockout {
 
 			switch ( constant( 'VIP_LOCKOUT_STATE' ) ) {
 				case 'warning':
-					$show_notice = apply_filters( 'vip_lockout_show_notice', $user->has_cap( 'manage_options' ), constant( 'VIP_LOCKOUT_STATE' ), $user );
+					$has_caps    = isset( $user->allcaps['manage_options'] ) && true === $user->allcaps['manage_options'];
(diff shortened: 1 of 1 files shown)


nickdaugherty (commented):
Makes sense to me and the code looks good ... would love to hear opinions from @Automattic/vip-platform-cantina

rebeccahum (dismissed):
Thanks for fixing this bug!

rebeccahum:
Sorry, fail rebasing :(

rebeccahum (approved):

PR #3319, opened 2022-06-29
Prevent stale admin editor changes when Taxonomy cache clear is deferred.

jesse-greathouse:
## Description
Adds is_admin and admin_limit logic to prevent stale taxonomy data in the topic editor when taxonomy cache is normally deferred

### Updated taxonomy editor cache clearing behavior.
Typically the platform will defer large taxonomy cache invalidations so that the cache can be cleared asynchronously. Now, when editing taxonomies, the behavior will tolerate significantly larger taxonomy caches, so that the taxonomy editor might not be out of sync.

## Checklist

## Steps to Test
Example:

1. Check out PR.
1. Populate the site with a large amount of articles related to a specific topic (Over 2000 less than 10000) 
1. Go to `wp-admin` > `Articles` > `Topics`
1. Alter the name of the topic and save.
1. Verify the topic name is changed in the editor.

### DIFF
--- a/vip-helpers/vip-clean-term-cache.php
+++ b/vip-helpers/vip-clean-term-cache.php
@@ -58,7 +58,7 @@ class VIP_Suspend_Cache_Invalidation {
 
 		// Allows admin users to see saved changes in the edit terms page, when editing a topic
 		// Short circuits scheduling and runs the method synchronously
-		if ( ( $num_objects < $this->admin_limit ) && is_admin() ) {
+		if ( ( $num_objects < $this->admin_limit ) && is_admin() && ! wp_doing_ajax() ) {
 			$this->cron_action( $tt_id, $taxonomy );
 			return;
 		}
--- a/vip-helpers/vip-clean-term-cache.php
+++ b/vip-helpers/vip-clean-term-cache.php
@@ -6,6 +6,8 @@ class VIP_Suspend_Cache_Invalidation {
 
 	private $limit = 2000; //That's what used to work on WPCOM
 
+	private $admin_limit = 10000; //Limit for users who are in the admin UI
+
 	public function __construct() {
 		// Disable {$taxonomy}_relationships cache invalidation
 		// when saving categories, as some have a very large number of posts,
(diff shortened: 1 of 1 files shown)


vip-helpers/vip-clean-term-cache.php
```diff
@@ -52,8 +54,17 @@ public function edited_term_taxonomy_action( $tt_id, $taxonomy ) {
 		// phpcs:ignore WordPress.DB.DirectDatabaseQuery
 		$objects = $wpdb->get_var( $wpdb->prepare( "SELECT COUNT(*) FROM $wpdb->term_relationships WHERE term_taxonomy_id = %d", $tt_id ) );
 
+		$num_objects = intval( $objects );
+
+		// Allows admin users to see saved changes in the edit terms page, when editing a topic
+		// Short circuits scheduling and runs the method synchronously
+		if ( ( $num_objects < $this->admin_limit ) && is_admin() ) {
```
rinatkhaziev: I don't think `is_admin` is what we want here because it fires on AJAX requests too? cc @WPprodigy

rebeccahum: Maybe `wp_doing_ajax()`? https://developer.wordpress.org/reference/functions/wp_doing_ajax/

jesse-greathouse: I expanded the condition to include a negative check for wp_doing_ajax. What do you think about this?

```PHP
    if ( ( $num_objects < $this->admin_limit ) && is_admin() && ! wp_doing_ajax() ) {
```

vip-helpers/vip-clean-term-cache.php
```diff
@@ -52,8 +54,17 @@ public function edited_term_taxonomy_action( $tt_id, $taxonomy ) {
 		// phpcs:ignore WordPress.DB.DirectDatabaseQuery
 		$objects = $wpdb->get_var( $wpdb->prepare( "SELECT COUNT(*) FROM $wpdb->term_relationships WHERE term_taxonomy_id = %d", $tt_id ) );
 
+		$num_objects = intval( $objects );
+
+		// Allows admin users to see saved changes in the edit terms page, when editing a topic
+		// Short circuits scheduling and runs the method synchronously
+		if ( ( $num_objects < $this->admin_limit ) && is_admin() && ! wp_doing_ajax() ) {
+			$this->cron_action( $tt_id, $taxonomy );
```
WPprodigy, line 62: I don't think we need to run the cron action here - would just an early return like in the conditional below be sufficient?

If I understand this all correctly, this module is meant to to override some defaults if the term count is too high and instead schedule a cache purge later on. If under the "limit" it doesn't suspend cache invalidation. Here we just simply need to not suspend if it's in the admin and under the "admin limit" ?

rebeccahum:
@jesse-greathouse Any plans on picking this back up?

PR #3351, opened 2022-07-07
CI: Fix tagging release not running

rebeccahum:
Changes were made in https://github.com/Automattic/vip-go-mu-plugins/pull/3316 towards renaming the workflow, but since it didn't get synced to `production`, the workflow isn't running on `push` to `production` as expected.

Edit: I cherry-picked this change directly onto `staging` branch.

### DIFF
--- a/.github/workflows/tag-release.yml
+++ b/.github/workflows/tag-release.yml
@@ -1,7 +1,10 @@
-name: release-changelog
+name: Tag Release
 
 on:
-  workflow_dispatch
+  push:
+    branches:
+      - production
+      - update/ci
 
 permissions:
   contents: read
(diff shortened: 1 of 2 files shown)


.github/workflows/tag-release.yml
```diff
@@ -1,7 +1,10 @@
-name: release-changelog
+name: Tag Release
 
 on:
-  workflow_dispatch
+  push:
+    branches:
+      - production
+      - update/ci
```
rinatkhaziev, line 7: I think this should be removed and was used by @pschoffer for testing purposes?

rinatkhaziev (approved):

PR #3362, opened 2022-07-11
Staging release 20220711.0

rebeccahum:
Includes just the one PR: https://github.com/Automattic/vip-go-mu-plugins/pull/3354

PR #3363, opened 2022-07-11
Production release v20220711.1

rebeccahum:
Syncing with https://github.com/Automattic/vip-go-mu-plugins/pull/3362


jesse-greathouse (approved):
🦖

PR #3369, opened 2022-07-12
CI: For staging changelog summary, exclude the deployed to prod PRs

rebeccahum:
If something that was marked as deployed to prod somehow made it into the staging commit history, we don't need to include it in the changelog.


jesse-greathouse (approved):
LGTM

PR #3392, opened 2022-07-15
Search dev tools usability improvements

rinatkhaziev:
## Description

This PR contains a few UI updates:

1. A copy button is implemented (allows to quickly copy the response into clipboard)
2. Buttons were made larger
3. Query response time is color-coded now depending on how slow or fast any given query is.
4. We've disabled syntax highlighting in the query response as a workaround for syntax highlighting performance being severely affected by wp-emoji's MutationObserver.

Highlighting large JSON blobs results in a lot of HTML nodes which, in conjunction, with wp-emoji MutationObserver makes it's extremely slow, to the point of being unusable. 

The proper fix is to provide API in core wp-emoji to allow disabling the MutationObserver

1. A copy button is implemented (allows to quickly copy the response into clipboard)
2. Buttons were made larger
3. Query response time is color-coded depending on how slow or fast any given query is.
4. We've disabled syntax highlighting in the query response as a workaround for syntax highlighting performance being severely affected by wp-emoji's MutationObserver.

## Checklist


search/search-dev-tools/src/style/prism.scss
```diff
@@ -14,7 +14,7 @@
 		white-space: pre;
 		word-spacing: normal;
 		word-break: normal;
-		font-size: 0.9em;
+		font-size: 12px;
```
sjinks: Maybe we should try to find the proper `em` value? `em`s scale better if the user wants to zoom in or zoom out

search/search-dev-tools/src/components/query/index.js
```diff
@@ -72,9 +72,9 @@ const Query = ( { args, request, url, query_args, backtrace = [] } ) => {
 	}, [ state.query, state.editing ] );
 
 	// Re-highlight the query on result change (after the response is received).
-	useEffect( () => {
-		highlightElement( queryResultRef.current );
-	}, [ queryResultRef, state.result ] );
+	// useEffect( () => {
+	// 	highlightElement( queryResultRef.current );
+	// }, [ queryResultRef, state.result ] );
```
sjinks: I would suggest that we remove the commented-out code - we can always restore it from Git.

sjinks: Also, `highlightElement` on line 5 also needs to be removed.

search/search-dev-tools/src/style/prism.scss
```diff
@@ -28,7 +28,7 @@
 	}
 
 	pre > code[class*='language-'] {
-		font-size: 1em;
+		// font-size: 1em;
```
sjinks: I would suggest to remove the entire CSS rule :-)

sjinks (changes requested):

search/search-dev-tools/src/style/prism.scss
rinatkhaziev: thanks, sloppy Friday afternoon coding right here :)

sjinks (dismissed):

sjinks (approved):

PR #3393, opened 2022-07-18
Automate rebuild of search-dev-tools

sjinks:
This PR automates the rebuild of search-dev-tools when one of its essential dependencies gets updated. This is especially useful when dealing with Dependabot PRs: you don't need to check out every PR, install dependencies, rebuild the bundle and commit the changes anymore.

**How it works:**
The action checks out the code, installs the dependencies, and runs the build. If the build has succeeded, it runs `git diff` against the `search/search-dev-tools` directory. If there are any changes, it commits and pushes them.

The commit is authored by `"WordPress VIP Bot" <github-actions@github.com>` but we can change that if necessary. However, we use the bot's repo access token so that the commit triggers another round of CI.

**How to test:**

1. Check out this branch.
2. Modify a JS file in `search/search-dev-tools/src` (this should be a change to the code, not formatting).
3. Commit the changes.
4. If everything is fine, you should see a new commit with the rebuilt files soon.

gha-log

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,10 +123,15 @@ jobs:
         run: |
           if [ "$(git diff --ignore-space-at-eol search/search-dev-tools/ | wc -l)" -gt "0" ]; then
             echo "::warning::There are changes in the built files."
-            git config user.name "WordPress VIP Bot"
-            git config user.email github-actions@github.com
-            git config push.default "current"
-            git add search/search-dev-tools/
-            git commit -m "Regenerate the bundle"
-            git push
+            if [ "${{ github.event.head_commit.committer.email }}" != "github-actions@github.com" ]; then
+              git config user.name "WordPress VIP Bot"
+              git config user.email github-actions@github.com
+              git config push.default "current"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -121,3 +118,20 @@ jobs:
       - name: Build files
         run: npm run build
         working-directory: search/search-dev-tools
+
+      - name: Check the diff
+        run: |
+          if [ "$(git diff --ignore-space-at-eol search/search-dev-tools/ | wc -l)" -gt "0" ]; then
+            echo "::warning::There are changes in the built files."
+            if [ "${{ github.event.head_commit.committer.email }}" != "github-actions@github.com" ]; then
```
sjinks, line 126: This is to avoid infinite loops if something goes terribly wrong. As an alternative, we can check that `${{ github.actor }}` is not the name of the bot. However, there could be conflicts if we add some other automation later, as the name of the bot cannot be changed but the email address of the committer can.

rebeccahum (approved):
Nice!

PR #3397, opened 2022-07-18
Add slack hook to the CI pipeline

pschoffer:
Adds a slack ping at the end of the changelog draft creation. That way we are notified about the release and can easily review/publish the changelog entry.


.github/workflows/changelog-summary.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - production
       - staging
+      - update/ci
```
rebeccahum: I think we removed this branch, did this accidentally sneak in? 🙂

rebeccahum (changes requested):

.github/workflows/changelog-summary.yml
pschoffer: Damn, I thought I would get away with keeping my test branch in there ... I need a higher Sneak level clearly. I can remove it again ... till the next time:)

pschoffer: Done

sjinks (approved):

PR #3409, opened 2022-07-19
[CANTINA-889] Add Prometheus plugin

sjinks:
## Description

This PR introduces Prometheus plugin and several basic collectors. Later, I am going to add more collectors.

### Plugin Added: Prometheus

Added Prometheus plugin to collect performance metrics.

## Checklist

## Steps to Test

1. Check out the code
2. Create a dev env with mu-plugins pointing to the checked-out code.
3. Go to `http://vip-local.vipdev.lndo.site:8000/metrics` (or whatever the URL is) and refresh the page a couple of times.

metrics


prometheus/inc/class-plugin.php
```diff
@@ -0,0 +1,94 @@
...
+		add_action( 'init', [ $this, 'init' ] );
+	}
+
+	public function plugins_loaded(): void {
+		$collectors = apply_filters( 'vip_prometheus_collectors', [] );
+		if ( ! is_array( $collectors ) ) {
+			$collectors = [];
+		}
+
+		$collectors = array_filter( $collectors, fn ( $collector) => is_object( $collector ) && $collector instanceof CollectorInterface );
```
rinatkhaziev: Any particular reason to use array_filter + array_map vs a basic foreach?

prometheus/inc/class-plugin.php
```diff
@@ -0,0 +1,94 @@
...
+
+	public function init(): void {
+		add_rewrite_endpoint( 'metrics', EP_ROOT );
+		add_filter( 'wp_headers', [ $this, 'wp_headers' ], 10, 2 );
+		add_action( 'template_redirect', [ $this, 'template_redirect' ] );
+
+		$version = (int) get_option( 'vip_prometheus_version', 0 );
+		if ( $version < 1 ) {
+			// phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.flush_rewrite_rules_flush_rewrite_rules -- there are no activation hooks for mu-plugins
+			flush_rewrite_rules();
```
rinatkhaziev: This is a bit dangerous because flushing rewrite rules randomly in an inopportune moment can cause issues.

prometheus/inc/class-plugin.php
```diff
@@ -0,0 +1,94 @@
...
+		if ( ! is_array( $collectors ) ) {
+			$collectors = [];
+		}
+
+		$collectors = array_filter( $collectors, fn ( $collector) => is_object( $collector ) && $collector instanceof CollectorInterface );
+		array_map( fn ( CollectorInterface $collector ) => $collector->initialize( $this->registry ), $collectors );
+	}
+
+	public function init(): void {
+		add_rewrite_endpoint( 'metrics', EP_ROOT );
```
rinatkhaziev: Since it's for internal use I don't think we should expose it as a rewrite rule.

prometheus/inc/class-plugin.php
sjinks: There are no activation hooks for mu-plugins. And we do need to flush the rules once after we add a new endpoint. Ideally, this happens only once.

I will try to use `wp_cache_add()` as a mutex though

prometheus/inc/class-plugin.php
sjinks: Well, by default Prometheus scrapes the `site.com/metrics` endpoint. And even if we rename it, we still have to make it visible to Prometheus.

Or do you suggest checking `$_SERVER['REQUEST_URI']` upon plugin instantiation?

I tried to do that the WordPress way :-)

prometheus/inc/class-plugin.php
sjinks: No :-) Two lines instead of 6-7.

sjinks: Rewrote the code.

prometheus/inc/class-plugin.php
sjinks: Implemented `/metrics` as a hidden endpoint.

prometheus/inc/class-plugin.php
sjinks: Implemented `/metrics` as a hidden endpoint.

rinatkhaziev (approved):

PR #3418, opened 2022-07-21
Use filter instead of a show_admin_bar call to force Admin Bar in debug mode

rinatkhaziev:
## Description

Some customers disable the admin bar via a filter, and we always want to show the admin bar in debug mode.

This PR takes it a bit further by trying to add a filter at the max priority on a `9999` priority. It may seem extreme, but there's no guarantee that a hook with the same priority would be added later during the execution of the request, but this approach should cover the absolute majority of cases.

## Checklist

## Steps to Test

1. Pass `?a8c-debug=true` to a request, not logged in.
2. Verify admin bar shows up as expected.

### DIFF
--- a/000-debug/debug-mode.php
+++ b/000-debug/debug-mode.php
@@ -116,9 +116,9 @@ function enable_debug_tools() {
 		return $user_caps;
 	}, 10, 3 );
 
-	add_action('init', function () {
+	add_action( 'init', function () {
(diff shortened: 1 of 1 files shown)


WPprodigy:
Think we should limit it to vip_support accounts / proxied users?

rinatkhaziev:
Already is, if you look at surrounding lines.

sjinks (approved):

PR #3420, opened 2022-07-21
Fix rate-limiting logic incorrectly increasing the counter even for queries that are not supposed to be offloaded to ES

rinatkhaziev:
## Description

Current logic is faulty - it increments the value on every filter call, which may lead to an artificially increased counter number, which would trigger rate-limiting logic and send some queries to DB. This may lead to random spikes in SQL.

The new approach only increments the counter AFTER the request is made, whether it was successful or not. 

We've reworked query rate-limiting logic to avoid potential over-counting of number of queries made to Enterprise Search backend.

## Checklist

## Steps to Test

1. Before applying this PR, run a search query on the site, make note of total number of remote requests in QM or Search Dev Tools
2. `wp cache get query_count vip_search` - make note that this number is bigger
3. Apply the PR
4. Flush memcached
5. Do steps 1, 2 and verify the counter is `1`

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -2219,7 +2219,7 @@ class Search {
 				self::$query_count_ttl
 			);
 
-			$this->logger->log( 'warning', 'search_query_rate_limiting', $message );
+			$this->logger->log( 'warning', 'search_query_rate_limiting', $message, [ 'count' => self::get_query_count() ] );
 		}
 	}
 
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -521,8 +521,10 @@ class Search {
 		// Disable query integration by default
 		add_filter( 'ep_skip_query_integration', array( __CLASS__, 'ep_skip_query_integration' ), 5, 2 );
 		add_filter( 'ep_skip_user_query_integration', array( __CLASS__, 'ep_skip_query_integration' ), 5 );
+
 		// Rate limit query integration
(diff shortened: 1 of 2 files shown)


search/includes/classes/class-search.php
```diff
@@ -1258,32 +1263,49 @@ public static function ep_skip_query_integration( $skip, $query = null ) {
...
+	 * This doesn't care whether a request was successful or not.
+	 *
+	 * @param array $query formatted ElasticPress Query object
+	 * @param string $type Current query type
+	 *
+	 * @return void
+	 */
+	public function increment_rate_limit_counter_after_request( $query, $type ) {
+		if ( $type === 'query' ) {
+			self::query_count_incr();
```
nickdaugherty: Rather than counting it here, can't we just count it inside `filter__ep_do_intercept_request()`, after the cache/not cached check, and after we make the network request to ES? We don't have to use the EP hook for this, we know when the network request completed, which is right here https://github.com/Automattic/vip-go-mu-plugins/blob/11c4384b5b38c791dab46a4426af1949dbf5d716/search/includes/classes/class-search.php#L929-L936

rinatkhaziev: I feel dumb now :)

nickdaugherty (dismissed):
💥 Thank you for fixing this!

search/includes/classes/class-search.php
nickdaugherty: I feel dumb multiple times a day 😄

WPprodigy (dismissed):
Nice work tracking this down.

rinatkhaziev:
Had to fix a bunch of tests, please re-review

pschoffer (approved):

PR #3431, opened 2022-07-26
AI-16: Add ability to shutdown site when overdue

vaurdan:
## Description

Adds the ability to effectively shutdown a WordPress application, when the ~`WPCOM_VIP_OVERDUE_LOCKOUT`~ `VIP_OVERDUE_LOCKOUT` constant is set and truthy. 

## Checklist

## Steps to Test

1. Check out PR.
2. Edit `config/wp-config.php` and add the following: 
```php
define( 'VIP_OVERDUE_LOCKOUT', true );
```
3. If testing on a VIP environment, you should see the `site-shutdown.html` page on all requests. 
4. If you are testing on a local environment, nothing should happen. Fake a VIP environment by adding 
```php
define('VIP_GO_ENV', true);
```
to your `wp-config.php` file


errors/site-shutdown.html
```diff
@@ -0,0 +1,160 @@
...
+			</div>
+
+			<div class="col-right">
+				<h1>Site Unavailable</h1>
```
vaurdan, line 145: The copy here is temporary and will likely need a revision.

lib/sunrise/sunrise.php
```diff
@@ -67,12 +67,18 @@ function handle_not_found_error( $error_type ) {
...
 		if ( $is_maintenance_mode ) {
 			// 503 prevents page from being cached.
 			// We handle healthchecks earlier and don't have to worry about them.
 			$status_code = 503;
 			header( 'X-VIP-Go-Maintenance: true' );
 			$error_doc = sprintf( '%s/mu-plugins/errors/site-maintenance.html', WP_CONTENT_DIR );
+		} elseif ( $is_overdue_locked && Context::is_vip_env() ) {
+			// Similar with maintenance mode, but with a different message.
+			$status_code = 503;
+			header( 'X-VIP-Go-Maintenance: true' ); // TODO: different header?
```
vaurdan: Should we use a different header here? I don't think there is any disadvantage on leveraging the same header, but let me know your thoughts.

000-vip-init.php
```diff
@@ -64,6 +68,29 @@
 	}
 }
 
+// Sites can be disabled if there is an overdue payment.
+// This constant is defined by VIP Go in config/wp-config.php.
+if ( Context::is_vip_env() && WPCOM_VIP_OVERDUE_LOCKOUT ) {
+	// WP CLI is allowed, but disable cron
+	if ( Context::is_wp_cli() ) {
+		add_filter( 'pre_option_a8c_cron_control_disable_run', function() {
```
rinatkhaziev: This can be simplified with `__return_true` rather than  using a closure

000-vip-init.php
```diff
@@ -64,6 +68,29 @@
 	}
 }
 
+// Sites can be disabled if there is an overdue payment.
+// This constant is defined by VIP Go in config/wp-config.php.
+if ( Context::is_vip_env() && WPCOM_VIP_OVERDUE_LOCKOUT ) {
```
rinatkhaziev:
```suggestion
if ( Context::is_vip_env() && Context::is_overdue_locked() ) {
```

rinatkhaziev (dismissed):
Left a couple of inline suggestions; aside from that, what do you think of dropping the `WPCOM_` part on the constants, the old ones are like that because of the legacy but probably we shouldn't keep on doing it.

lib/sunrise/sunrise.php
nickdaugherty: I would say yes - the site isn't in maintenance mode so this header would be confusing / misleading.

lib/sunrise/sunrise.php
```diff
@@ -67,12 +67,18 @@ function handle_not_found_error( $error_type ) {
...
+		$is_overdue_locked   = Context::is_overdue_locked();
 		if ( $is_maintenance_mode ) {
 			// 503 prevents page from being cached.
 			// We handle healthchecks earlier and don't have to worry about them.
 			$status_code = 503;
 			header( 'X-VIP-Go-Maintenance: true' );
 			$error_doc = sprintf( '%s/mu-plugins/errors/site-maintenance.html', WP_CONTENT_DIR );
+		} elseif ( $is_overdue_locked && Context::is_vip_env() ) {
+			// Similar with maintenance mode, but with a different message.
+			$status_code = 503;
```
nickdaugherty: 503 doesn't seem like the right status code, and will probably end up setting off various alarms and polluting metrics.

There's an "unofficial" 402 code for Payment Required that sounds right for this https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/402

000-vip-init.php
```diff
@@ -12,6 +12,8 @@
 use Automattic\VIP\Utils\Context;
 use Automattic\VIP\Utils\WPComVIP_Restrictions;
 
+// @codeCoverageIgnoreStart
```
mjangda, line 15: Can we add a comment noting why we're doing this?

000-vip-init.php
```diff
@@ -64,6 +70,27 @@
...
+	// WP CLI is allowed, but disable cron
+	if ( Context::is_wp_cli() ) {
+		add_filter( 'pre_option_a8c_cron_control_disable_run', '__return_true', 9999 );
+	} else {
+		// Don't try to short-circuit Jetpack requests, otherwise it will break the connection.
+		require_once __DIR__ . '/vip-helpers/vip-utils.php';
+		if ( ! vip_is_jetpack_request() ) {
+			http_response_code( 503 );
+
+			header( 'X-VIP-Go-Maintenance: true' );
```
```suggestion
			header( 'X-VIP-402: true' );
```

errors/site-shutdown.html
```diff
@@ -0,0 +1,160 @@
...
+				<p>Payment for this WordPress VIP account is overdue, and it has been disabled.</p>
+				<p>Contact <a href="mailto:accounts@wpvip.com">accounts@wpvip.com</a> to pay your bill.</p>
+				<footer>
+					<span class="error">Site Unavailable</span>
```
mjangda:
```suggestion
					<span class="error">402</span>
```

errors/site-shutdown.html
```diff
@@ -0,0 +1,160 @@
...
+			</div>
+
+			<div class="col-right">
+				<h1>Site Unavailable</h1>
+				<p>Payment for this WordPress VIP account is overdue, and it has been disabled.</p>
+				<p>Contact <a href="mailto:accounts@wpvip.com">accounts@wpvip.com</a> to pay your bill.</p>
```
mjangda:
```suggestion
				<p>This site is currently not available. Please try again later.</p>
```

errors/site-shutdown.html
```diff
@@ -0,0 +1,160 @@
...
+				<footer>
+					<span class="error">Site Unavailable</span>
+				</footer>
+			</div>
+		</div>
+
+		<script type="text/javascript" src="https://stats.wp.com/wpcom.js?22"></script>
+		<script type="text/javascript">
+			_error = 'site-shutdown';
```
mjangda:
```suggestion
			_error = 'vip-site-402';
```

lib/sunrise/sunrise.php
mjangda: Will 402s be cached in Varnish? Will need to look into how we handle that (e.g. if cached, we probably want to flush the cache after the lockout is removed)

lib/sunrise/sunrise.php
mjangda:
```suggestion
			header( 'X-VIP-402: true' );
```

errors/site-shutdown.html
```diff
@@ -0,0 +1,160 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<meta charset="utf-8" />
+		<meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1" />
+		<meta name="viewport" content="width=device-width, initial-scale=1.0" />
+
+		<title>Maintenance</title>
```
mjangda:
```suggestion
		<title>Site Unavailable</title>
```

mjangda (dismissed):
Added some suggestions for the language / wording.

errors/site-shutdown.html
nickdaugherty: Good question about 402 cacheability. I don't see any special rules for 4xx, aside from 404 which is cached for 1 minute, so I think our normal cache logic will be in play.

lib/sunrise/sunrise.php
nickdaugherty: Replied in another spot, but I couldn't find any exceptions in varnish.vcl that would affect 402, so AFAIK it will receive the normal cache rules, which means we will need to do a cache flush on lockout...and on lockout for that matter.

vaurdan: That's a very good point, and thanks for checking @nickdaugherty! I'm going to add the extra step for the cache flush on the respective GOOP endpoints.

000-vip-init.php
vaurdan: @sjinks I believe you added this line. I would say it is related with the ~SonarCloud~CodeCov report, but I wanted to double check and ask for your input before adding a comment here.

sjinks: @mjangda please disregard this for now - I will write a P2 post explaining all these things.

In brief, PHPUnit is unable to generate a coverage report for the files loaded from the bootstrap script, even if those files are used in tests (I suspect that the code needs to be instrumented when it is compiled, and compilation of the bootstrap script and the files it requires happens before the code coverage driver is initialized). This is why these files are reported as uncovered.

I want to try another approach: initialize the autoloader and load the test library in `bootstrap.php` and use PHPUnit's Test Listener feature to load all mu-plugins before the very first test (the driver should be initialized by that time). However, I am not sure this will play nice with the way the WordPress Test Library initializes all hooks and filters.

vaurdan:
Is there anything else that needs to be addressed before we can merge this PR to `develop`?

tests/lib/utils/context.php
```diff
@@ -47,6 +47,24 @@ public function test__is_maintenance_mode__yep() {
...
 	}
 
+	public function test__is_overdue_locked__nope() {
+		// Note: `VIP_OVERDUE_LOCKOUT` not defined
+
+		self::assertFalse( defined( 'VIP_OVERDUE_LOCKOUT' ) );
+
+		$actual_result = Context::is_overdue_locked();
+
+		$this->assertFalse( $actual_result, '`VIP_OVERDUE_LOCKOUT` constant should not be defined for this test' );
```
nickdaugherty:
```suggestion
		$this->assertFalse( $actual_result, 'The `VIP_OVERDUE_LOCKOUT` constant is not defined, but site is reported as locked' );
```

The text of the error message doesn't reflect what's being asserted.

tests/lib/utils/context.php
```diff
@@ -47,6 +47,24 @@ public function test__is_maintenance_mode__yep() {
...
+	}
+
+	public function test__is_overdue_locked__yep() {
+		Constant_Mocker::define( 'VIP_OVERDUE_LOCKOUT', true );
+
+		$actual_result = Context::is_overdue_locked();
+
+		$this->assertTrue( $actual_result );
```
nickdaugherty: If we want to keep with the theme of having a custom error message for the assertion, we can do this:

```suggestion
		$this->assertTrue( $actual_result, 'The `VIP_OVERDUE_LOCKOUT` constant is set to `true`, but site is not reported as locked' );
```

lib/utils/class-context.php
```diff
@@ -22,6 +22,11 @@ public static function is_maintenance_mode() {
 			&& true === constant( 'WPCOM_VIP_SITE_MAINTENANCE_MODE' );
 	}
 
+	public static function is_overdue_locked() {
+		return defined( 'VIP_OVERDUE_LOCKOUT' )
+			&& true === constant( 'VIP_OVERDUE_LOCKOUT' );
```
nickdaugherty, line 27: For simplicity, we can skip the `constant()` function here, as it's only really needed when the constant name is not known beforehand (a variable).
```suggestion
			&& true === VIP_OVERDUE_LOCKOUT;
```

lib/sunrise/sunrise.php
```diff
@@ -67,12 +67,18 @@ function handle_not_found_error( $error_type ) {
...
 			// 503 prevents page from being cached.
 			// We handle healthchecks earlier and don't have to worry about them.
 			$status_code = 503;
 			header( 'X-VIP-Go-Maintenance: true' );
 			$error_doc = sprintf( '%s/mu-plugins/errors/site-maintenance.html', WP_CONTENT_DIR );
+		} elseif ( $is_overdue_locked && Context::is_vip_env() ) {
+			// Similar with maintenance mode, but with a different message.
+			$status_code = 402;
+			header( 'X-VIP-402: true' );
+			$error_doc = sprintf( '%s/mu-plugins/errors/site-shutdown.html', WP_CONTENT_DIR );
```
nickdaugherty: Looking at this closer, I wonder if we actually need this code block at all.

The regular locked / not locked check happens separately in 000-vip-init.php ... do we need to handle it again here too? The cross section of sites that are 1) locked out, and 2) have site not found errors is probably pretty tiny, and worst case, the end user is still seeing an error page and not real content.

Just trying to reduce the amount of code and duplication we have, if it makes sense.

000-vip-init.php
```diff
@@ -64,6 +70,27 @@
 	}
 }
 
+// Sites can be disabled if there is an overdue payment.
+// This constant is defined by VIP Go in config/wp-config.php.
+if ( Context::is_vip_env() && Context::is_overdue_locked() ) {
```
nickdaugherty: Do we need any exceptions for logged-in VIP members? Could be a future enhancement I suppose. Just thinking of the case where someone on our end needs to do _something_ on a site while it's in lockout.

Folks that have the ability to sandbox sites can hack around it, just wondering if people think it's worthwhile to account for now.

nickdaugherty:
@vaurdan I left a few more non-blocking thoughts, I think this is really close.

lib/utils/class-context.php
rinatkhaziev: @nickdaugherty, we prefer `constant()` these days because it's better testable.

lib/sunrise/sunrise.php
vaurdan: That's a very good point. I think that I will remove this from the sunrise.php. As you said, worse case scenario, an error will still be displayed instead of any content.

000-vip-init.php
vaurdan: Good point! I have created AI-22 to track this improvement.

lib/sunrise/sunrise.php
```diff
@@ -67,6 +67,7 @@ function handle_not_found_error( $error_type ) {
 	$is_web_request = Context::is_web_request();
 	if ( $is_web_request ) {
 		$is_maintenance_mode = Context::is_maintenance_mode();
+		$is_overdue_locked   = Context::is_overdue_locked();
```
nickdaugherty:
```suggestion
```

@vaurdan we have a stray line, now that we've removed the other code.

nickdaugherty (dismissed):
This looks great!

lib/sunrise/sunrise.php
vaurdan: Good catch! :D

aswasif007 (dismissed):

aswasif007:
@vaurdan Let's merge it and try locking out a test site.

vaurdan:
Just waiting for the final approval by @Automattic/vip-platform-cantina :)

nickdaugherty:
Tested and works great! @Automattic/vip-platform-cantina any help you can offer to get this across the finish line is greatly appreciated 🙇

WPprodigy (dismissed):
Looks good! Sorry for the delayed review here.

Couple (non-blocking) thoughts:

1) Lets double check that it is cacheable, and ensure the cache is purged when a site is unlocked.

2) Mozilla says "This response code is reserved for future use": https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses. Seems adoption for it was too low. But also, super unlikely there is any movement on this, as these things seem to rarely change 😆 

3) Disabling cron means we won't be getting SDS syncs. Having that data fall out of sync isn't super ideal, but atm there's not a means to only run one type of job. (could be a fun future project to have a backup means of syncing to SDS for when cron is turned off / broke)

nickdaugherty:
Thanks for looking @WPprodigy !

> Lets double check that it is cacheable, and ensure the cache is purged when a site is unlocked.

I've looked at Varnish VCL and 4xx is cacheable AFAICT ... the only specific handling we have is for 404, which is cached for 60s, otherwise it should behave the same.

> But also, super unlikely there is any movement on this, as these things seem to rarely change

It's easy enough to change in the future if needed, but this code makes the most sense IMO since it's 1) in the 4xx error range, and 2) related to payments (a subtle jab 😄) 

> Disabling cron means we won't be getting SDS syncs.

Very good point - I think we're safe to leave cron running...it's not really the target of these lockouts. @vaurdan what do you think about leaving cron alone?

vaurdan:
> Disabling cron means we won't be getting SDS syncs. Having that data fall out of sync isn't super ideal, but atm there's not a means to only run one type of job. (could be a fun future project to have a backup means of syncing to SDS for when cron is turned off / broke)

Great point! I just committed a change to ignore the lockout when running from WP-CLI, cron included.

WPprodigy (approved):

PR #3439, opened 2022-07-28
Changelog creation

pschoffer:
Mostly I added bunch of debug lines. 

But I think I found the source of the problem. We might have picked wrong merge PR.

### DIFF
--- a/ci/changelog-summary.php
+++ b/ci/changelog-summary.php
@@ -78,24 +78,27 @@ function debug( $arg ) {
  * @return int $merged_pr The PR object
  */
 function fetch_pr_merged_to_branch() {
-    $merged_pr = false;
     $missing_label = BRANCH === 'production' ? LABEL_DEPLOYED_PROD : LABEL_DEPLOYED_STAGING;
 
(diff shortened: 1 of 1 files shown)


ci/changelog-summary.php
```diff
@@ -78,24 +78,27 @@ function debug( $arg ) {
...
+            echo "PR #{$pr['number']} is not merged, skipping.\n";
             continue;
         }
 
         $labels = array_map( fn ($label) => $label['name'], $pr['labels'] );
         if ( ! in_array( $missing_label, $labels ) ) {
-            // Look for PR with missing label.
-            $merged_pr = $pr;
+            echo "PR #{$pr['number']} - {$pr['title']} {$pr['html_url']} is it!\n";
+            return $pr;
```
pschoffer, line 95: I think this will be the fix we need. Picks the most recent merge not the last one on the first page.

sjinks (approved):

ci/changelog-summary.php
rebeccahum: Thank you!

PR #3440, opened 2022-07-28
Search Dev Tools UI enhancements

rinatkhaziev:
## Description

This PR addresses some CSS issues resulting in theme's CSS overtaking the Search Dev Tools buttons styles.

Also adds concurrent requests panel. As a side-effect of that, we had to remove Semaphore_Backend for Concurrency_Limiter

We've updated Search Dev Tools to include CSS fixes to prevent the "COPY" button from potentially being invisible to the user.

## Checklist

- [] This change has relevant documentation additions / updates (if applicable).


jesse-greathouse (dismissed):
👍  Very good!

rinatkhaziev:
@sjinks curious to hear your thoughts around the interface/implementation changes for ConcurrencyLimiter, particularly about returning `$this->increment` vs fetching value from remote in object cache backend.

search/includes/classes/concurrency-limiter/class-semaphore-backend.php
```diff
@@ -62,4 +62,8 @@ public function dec_value(): void {
 			}
 		}
 	}
+
+	public function get_value(): int {
+		return (int) $this->increments;
```
sjinks: The Bad Thing™ is that we cannot retrieve the value of the semaphore.

`increments` is how many times we acquired the semaphore in the current session (so that we release it as many times as we have acquired it). Ideally, `increments` shouldn't be more than one, but it is still possible to have a larger value if we call ES in, say, `ep_remote_request` hook (we do that in tests).

sjinks:
@rinatkhaziev the reason I did not add the `value()` method or did not return the old value from `inc_value()` is that not all backends support this operation. This is possible with APCu or Object Cache but not possible with Semaphore.

If we drop the Semaphore backend, we can do that.

rinatkhaziev:
@sjinks 
> This is possible with APCu or Object Cache but not possible with Semaphore.

I must've confused the purpose of the `$increments` property. I'm not against removing it IMO the getter method would be more valuable than the backend we don't really use.

sjinks (approved):

PR #3445, opened 2022-07-29
Fix ESLint warnings

sjinks:
```
vip-go-mu-plugins/search/search-dev-tools/src/components/query/index.js
  67:28  warning  Identifier name 'e' is too short (< 2)                  id-length
  89:32  warning  Unexpected parentheses around single function argument  arrow-parens
  90:7   warning  Identifier name 'c' is too short (< 2)                  id-length
```


rinatkhaziev (approved):

rinatkhaziev:
thanks for that, been meaning to address.

PR #3453, opened 2022-08-02
[CANTINA-889] Change the way the collectors are loaded

sjinks:
## Description

This PR changes the way how Prometheus collectors are loaded.

Now there are three different points when collectors can be loaded:
1. `vipgo_mu_plugins_loaded`: happens after all our MU plugins have been loaded and before `z-client-mu-plugins.php` loads MU plugins from the `client-mu-plugins` directory. This is when we should load *our* collectors;
2. `mu_plugins_loaded`: gives a chance for client MU plugins to set up their collectors;
3. `plugins_loaded`: gives a chance to client plugins to set up their collectors.

For now, (2) and (3) exist in case we will ever want to allow for user-provided collectors. In other words, these extension points exist, but we do not advertise them nor provide any user-visible feedback if they want to use them.

## Checklist

## Steps to Test

The CI should pass.

### DIFF
--- a/prometheus.php
+++ b/prometheus.php
@@ -15,12 +15,15 @@ if ( defined( 'ABSPATH' ) ) {
 		require_once __DIR__ . '/prometheus-collectors/class-apcu-collector.php';
 		require_once __DIR__ . '/prometheus-collectors/class-opcache-collector.php';
 
-		add_filter( 'vip_prometheus_collectors', function ( array $collectors ): array {
-			$collectors[] = new Cache_Collector();
-			$collectors[] = new APCu_Collector();
-			$collectors[] = new OpCache_Collector();
+		add_filter( 'vip_prometheus_collectors', function ( array $collectors, string $hook ): array {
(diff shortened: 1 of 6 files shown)


rebeccahum:
> For now, (2) and (3) exist in case we will ever want to allow for user-provided collectors. In other words, these extension points exist, but we do not advertise them nor provide any user-visible feedback if they want to use them.

Do we really want to enable these yet if we don't plan to advertise the existence? I guess, it feels easier to add these points later on because if we remove them later on and we don't want things break as we didn't expect many users to use them.

sjinks:
I remember when we used logstash to see (and count) something when we applied patches to the core (I think the last time it was something encoding-related).

We can use these hooks instead to collect metrics; they can also be useful when we apply our security patches to zero day exploits (the graphs can show if the vulnerability is already exploited in the wild).

rebeccahum (dismissed):

rebeccahum (approved):

PR #3457, opened 2022-08-02
[CANTINA-889] Add login stats collector

sjinks:
## Description

This PR adds a Prometheus collector for login stats.

It introduces the following counters:
* `security_login_limit_exceeded_total`: number of rate-limited login requests;
* `security_password_reset_limit_exceeded_total`: number of rate-limited password reset requests;
* `security_successful_logins_total`: number of successful logins;
* `security_failed_logins_total`: number of failed logins.

All counters have the `site_id` label with the value equal to the return value of `get_current_blog_id()`.

### DIFF
--- a/prometheus-collectors/class-login-stats-collector.php
+++ b/prometheus-collectors/class-login-stats-collector.php
@@ -50,19 +50,19 @@ class Login_Stats_Collector implements CollectorInterface {
 	}
 
 	public function login_limit_exceeded(): void {
-		$this->login_limit_exceeded_counter->inc( [ 'site_id' => (string) get_current_blog_id() ] );
+		$this->login_limit_exceeded_counter->inc( [ (string) get_current_blog_id() ] );
 	}
 
 	public function password_reset_limit_exceeded(): void {
-		$this->password_reset_limit_exceeded_counter->inc( [ 'site_id' => (string) get_current_blog_id() ] );
+		$this->password_reset_limit_exceeded_counter->inc( [ (string) get_current_blog_id() ] );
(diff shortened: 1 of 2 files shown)


prometheus-collectors/class-login-stats-collector.php
```diff
@@ -0,0 +1,71 @@
...
+use Prometheus\Counter;
+use Prometheus\RegistryInterface;
+
+/**
+ * @codeCoverageIgnore
+ */
+class Login_Stats_Collector implements CollectorInterface {
+	private Counter $login_limit_exceeded_counter;
+	private Counter $password_reset_limit_exceeded_counter;
+	private Counter $successful_login_counter;
```
rebeccahum, line 14: Just wondering, but why do we need a counter for successful logins? The other counters are more for failed/rate-limited requests.

sjinks: I was thinking on using the rate() function (or maybe something similar) in Prometheus to see the spikes in the failed to successful login ratio.

The counter is cheap, we can always remove it if we find that we don't need it.

rebeccahum (dismissed):

prometheus-collectors/class-login-stats-collector.php
rebeccahum: OK. What would the ratio of failed to successful would be helpful in determining?

rinatkhaziev: For one that would show us abnormal patterns?

PR #3469, opened 2022-08-04
Rename vipgo_mu_plugins_loaded to vip_mu_plugins_loaded

sjinks:


pschoffer (commented):
Do we know if no client code is relying on the old `vipgo_mu_plugins_loaded`?

sjinks:
Yes - that hook was introduced yesterday

sjinks:
Rinat's suggestion was to rename it to match our existing naming scheme

pschoffer (approved):

PR #3481, opened 2022-08-05
Lockout Plugin: Add `VIP_ACCOUNT_STATUS` constant to display notice/lock site.

vaurdan:
## Description
This PR changes the functionality of the Lockout plugin a bit, by using the `VIP_ACCOUNT_STATUS` to decide whether if the WordPress site should display a notice, or if it should be locked. The messaging is also "hard coded", and changes depending on the value of `VIP_ACCOUNT_STATUS`.

This constant can have four values:
 * `normal` - should be present on all sites without any overdue.
 * `warned` - slightly overdue, show a warning
 * `locked` - next overdue stage, where all the access is locked
 * `shutdown` - the application is shutdown (handled differently, see #3431)

I tried to maintain retro-compatibility with the current lockout architecture, so if `VIP_ACCOUNT_STATUS` is `normal`, it will fallback to the original logic: if `VIP_LOCKOUT_STATE` and `VIP_LOCKOUT_MESSAGE` are set, that message will be shown with the correct state. However, if at any time `VIP_ACCOUNT_STATUS` is changed to a state other than `normal`, it will bypass and override any of the values set in `VIP_LOCKOUT_STATE` and `VIP_LOCKOUT_MESSAGE`.

### Lockout: Adds VIP_ACCOUNT_STATUS constant to control the notice/locking functionalities

Adds a new `VIP_ACCOUNT_STATUS` that can be used to set the site's account status, and warn/lock the application, if required.

## Checklist

## Steps to Test

1. Check out PR.
1. Validate the original logic: Set `VIP_LOCKOUT_STATE` environment variable to `warning` and `VIP_LOCKOUT_MESSAGE` to some string. 
2. Now, add `define('VIP_ACCOUNT_STATUS', 'locked');`. The admin dashboard should be locked, and a red notice should be visible in all pages. This should override what was set in step 2.

### DIFF
--- a/security/class-lockout.php
+++ b/security/class-lockout.php
@@ -43,7 +43,7 @@ class Lockout {
 
 	private function get_lockout_state() {
 		// VIP_ACCOUNT_STATUS has precedence over VIP_LOCKOUT_STATE
-		if ( defined( 'VIP_ACCOUNT_STATUS' ) && VIP_ACCOUNT_STATUS !== self::ACCOUNT_STATUS_NORMAL ) {
+		if ( defined( 'VIP_ACCOUNT_STATUS' ) && constant( 'VIP_ACCOUNT_STATUS' ) !== self::ACCOUNT_STATUS_NORMAL ) {
 			return constant( 'VIP_ACCOUNT_STATUS' );
 		}
 
@@ -52,7 +52,7 @@ class Lockout {
 
 	private function get_lockout_message() {
 		// If the account is locked, use the proper lockout message
-		if ( defined( 'VIP_ACCOUNT_STATUS' ) && VIP_ACCOUNT_STATUS !== self::ACCOUNT_STATUS_NORMAL ) {
+		if ( defined( 'VIP_ACCOUNT_STATUS' ) && constant( 'VIP_ACCOUNT_STATUS' ) !== self::ACCOUNT_STATUS_NORMAL ) {
 			switch ( $this->get_lockout_state() ) {
 				case self::ACCOUNT_STATUS_WARNING:
 					return 'Payment for this WordPress VIP account is overdue and it will be disabled.<br />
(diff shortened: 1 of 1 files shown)


security/class-lockout.php
```diff
@@ -36,17 +41,44 @@ public function __construct() {
...
+		if ( defined( 'VIP_ACCOUNT_STATUS' ) && VIP_ACCOUNT_STATUS !== self::ACCOUNT_STATUS_NORMAL ) {
+			return constant( 'VIP_ACCOUNT_STATUS' );
+		}
+
+		return defined( 'VIP_LOCKOUT_STATE' ) ? constant( 'VIP_LOCKOUT_STATE' ) : false;
+	}
+
+	private function get_lockout_message() {
+		// If the account is locked, use the proper lockout message
+		if ( defined( 'VIP_ACCOUNT_STATUS' ) && VIP_ACCOUNT_STATUS !== self::ACCOUNT_STATUS_NORMAL ) {
```
sjinks:
```suggestion
		if ( defined( 'VIP_ACCOUNT_STATUS' ) && constant( 'VIP_ACCOUNT_STATUS' ) !== self::ACCOUNT_STATUS_NORMAL ) {
```

for consistency

security/class-lockout.php
```diff
@@ -36,17 +41,44 @@ public function __construct() {
 		}
 	}
 
+	private function get_lockout_state() {
+		// VIP_ACCOUNT_STATUS has precedence over VIP_LOCKOUT_STATE
+		if ( defined( 'VIP_ACCOUNT_STATUS' ) && VIP_ACCOUNT_STATUS !== self::ACCOUNT_STATUS_NORMAL ) {
```
sjinks:
```suggestion
		if ( defined( 'VIP_ACCOUNT_STATUS' ) && constant( 'VIP_ACCOUNT_STATUS' ) !== self::ACCOUNT_STATUS_NORMAL ) {
```

sjinks (changes requested):
LGTM except for two minor things.

sjinks (approved):

PR #3483, opened 2022-08-05
Search: Pull in ElasticPress to d655394c176b718b3284cc552671daf0776c4b1e

rebeccahum:
## Description
Includes the below PRs:
- https://github.com/Automattic/ElasticPress/pull/168
- https://github.com/Automattic/ElasticPress/pull/169

Two ElasticPress bug fixes:
- format_hits_as_terms() will return with a WP_Term object instead of stdClass
- Protected Content feature will take into account posts that are autosaved only


jesse-greathouse (approved):
👍

PR #3490, opened 2022-08-08
Debug bar: Change to earlier hook to utilize remote request logging

### LINKED ISSUE #544: Load Debug Bar earlier
Currently we hook Debug Bar into `init`, which is fairly late and misses some stuff. This was primarily done so we could limit it by user - but maybe we should look at proxy / sandbox status instead.

rebeccahum:
## Description
Fixes https://github.com/Automattic/vip-go-mu-plugins/issues/544.

### Plugin Updated: Debug Bar

Change to load at earlier hook to utilize remote request logging

## Checklist

## Steps to Test
1) Add in theme:
```
$url = 'http://www.wordpress.com';
$args = array(
    'method' => 'GET'
);
$response = wp_remote_request( $url, $args );
```
2) Go to Debug Bar and notice "WP_http" panel does not exist
3) Apply PR
4) Repeat step 2 and notice the panel now

### DIFF
--- a/debug-bar.php
+++ b/debug-bar.php
@@ -23,7 +23,7 @@ add_filter( 'debug_bar_enable', function( $enable ) {
 }, PHP_INT_MAX );
 
 // We only need to load the files if it's enabled
-add_action( 'init', function() {
+add_action( 'set_current_user', function() {
 	$enable = apply_filters( 'debug_bar_enable', false );
 
 	if ( ! $enable ) {


sjinks (dismissed):

sjinks:
Cool, I was looking into this issue as well :-)

WPprodigy (commented):
One possible problem would be existing code using the `debug_bar_enable` filter on an earlier priority `init` hook or so.

This change would make is so they now filter too late and lose the debug bar. As such, at a minimum will require some communications. Maybe could look at usage across the platform.

sjinks:
Actually yes :-(

This change could affect these sites:
* https://github.com/wpcomvip/nesn/blob/581e3462c912c0d80f05559fb8497dfd827e3089/plugins/nesn-development-assistant/includes/NESN/Development/DebugBar.php (assuming `init` means that the code is run on `init`);
* https://github.com/wpcomvip/bring-a-trailer/blob/14065fd55e88b1e7087fdb9b6233a7126e20e99f/plugins/bringatrailer/components/integrations/vip-go/debug-bar/debug-bar.php (`add_global_filters` seems to be run on `init`);
* https://github.com/wpcomvip/bethurbantrial/blob/0736cf6fb5f3d52574c09564aa0ff34d8106f611/plugins/debug-bar/debug-bar.php (probably)

rebeccahum:
Based on the previous list curated by @sjinks, it doesn't actually affect any customers. They are loading it earlier at `plugins_loaded`.

sjinks (approved):

PR #3499, opened 2022-08-09
Update jetpack_is_mobile() to account for X-Mobile-Class header

### LINKED ISSUE #363: Filter jetpack_is_mobile() to use the value of the X-Mobile-Class header
JP mobile detection should be based on the X-Mobile-Class header (set by Varnish) on VIP Go to ensure consistency.

This is because the JP mobile rules are implemented in Varnish VCL so we can correctly bucket requests for caching purposes. Since there are now 2 sources of the 'mobile class' for devices, we need to choose one as The Source of Truth, and that must be Varnish/VCL because only it runs at the edge.

rebeccahum:
## Description
Fixes #363.

### Plugin Updated: Jetpack: VIP Specific Changes

Changed the behavior of jetpack_is_mobile() to account for X-Mobile-Class header.

## Checklist

## Steps to Test
1) Fake coming from a desktop:
```
	$_SERVER['HTTP_X_MOBILE_CLASS'] = 'desktop';
}, 998);
```

2) Test out `jetpack_is_mobile()`:
```
	var_dump( jetpack_is_mobile() );
}, 999);
```
Expect it to return `false`.
4) Fake coming from a tablet:
```
	$_SERVER['HTTP_X_MOBILE_CLASS'] = 'tablet';
}, 998);
```
5) Repeat step 2) but expect it to return `true`
6) Fake coming from a smartphone:
```
	$_SERVER['HTTP_X_MOBILE_CLASS'] = 'smart';
}, 998);
```
7) Repeat step 2) but expect it to return `true`
8) Use `jetpack_is_mobile()` to check if request is coming from a non-smartphone:
```
	var_dump( jetpack_is_mobile( 'dumb' ) );
}, 999);
```
Expect it to return `false` b/c step 6)

### DIFF
--- a/vip-jetpack/vip-jetpack.php
+++ b/vip-jetpack/vip-jetpack.php
@@ -749,4 +749,8 @@ function vip_jetpack_is_mobile( $matches, $kind, $return_matched_agent ) {
 
 	return $matches;
 }
-add_filter( 'pre_jetpack_is_mobile', 'vip_jetpack_is_mobile', PHP_INT_MAX, 3 );
+$is_rolled_out = ( defined( 'VIP_GO_APP_ENVIRONMENT' ) && 'production' !== VIP_GO_APP_ENVIRONMENT ) || \Automattic\VIP\Feature::is_enabled_by_percentage( 'jetpack-is-mobile' );
+
+if ( $is_rolled_out ) {
+	add_filter( 'pre_jetpack_is_mobile', 'vip_jetpack_is_mobile', PHP_INT_MAX, 3 );
+}
--- a/vip-jetpack/vip-jetpack.php
+++ b/vip-jetpack/vip-jetpack.php
@@ -716,3 +716,37 @@ function vip_remove_jetpack_search_menu_page() {
 	);
 }
 add_action( 'admin_menu', 'vip_remove_jetpack_search_menu_page', PHP_INT_MAX );
+
+/**
+ * Account for X-Mobile-Class header in jetpack_is_mobile()
+ *
+ * @param bool|string $matches Boolean if current UA matches $kind or not. If
+ * $return_matched_agent is true, should return the UA string
+ * @param string      $kind Category of mobile device being checked. Can be 'any', 'smart' or 'dumb'.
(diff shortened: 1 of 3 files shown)


vip-jetpack/vip-jetpack.php
```diff
@@ -716,3 +716,37 @@ function vip_remove_jetpack_search_menu_page() {
...
+	$x_mobile_class = sanitize_text_field( $_SERVER['HTTP_X_MOBILE_CLASS'] ); // "desktop", "smart", "dumb", "tablet"
+
+	if ( 'desktop' === $x_mobile_class ) {
+		return false;
+	}
+
+	if ( 'smart' === $kind || 'dumb' === $kind ) {
+		$matches = $kind === $x_mobile_class;
+	} elseif ( 'any' === $kind ) {
+		$matches = true;
```
sjinks, line 747: If `$return_matched_agent` is true, shouldn't we return the UA string? At least the doc block says so...

rebeccahum: Good catch, updated now!

sjinks (dismissed):

vip-jetpack/vip-jetpack.php
```diff
@@ -716,3 +716,41 @@ function vip_remove_jetpack_search_menu_page() {
...
+	}
+
+	return $matches;
+}
+$is_rolled_out = ( defined( 'VIP_GO_APP_ENVIRONMENT' ) && 'production' !== VIP_GO_APP_ENVIRONMENT ) || \Automattic\VIP\Feature::is_enabled_by_percentage( 'jetpack-is-mobile' );
```
rinatkhaziev: How about for Constant_Mocker purposes? 
```suggestion
$is_rolled_out = ( defined( 'VIP_GO_APP_ENVIRONMENT' ) && 'production' !== constant( 'VIP_GO_APP_ENVIRONMENT' ) ) || \Automattic\VIP\Feature::is_enabled_by_percentage( 'jetpack-is-mobile' );
```

rinatkhaziev (approved):

PR #3503, opened 2022-08-10
Add Jetpack loading fallback

pschoffer:
## Description
After we have moved versioned Jetpack folders to https://github.com/Automattic/vip-go-mu-plugins-ext we want to still be able to use Jetpack locally, mostly for development purposes.

The change adds a fallback to the "unversioned" `jetpack` folder in mu-plugins.

### Plugin Updated: Jetpack

We upgraded jetpack loader logic which chooses what version of JP should be loaded. This change provides one extra fallback in case the default version of JP is not included in the mu-plugins.

## Checklist

## Steps to Test

1) Check the JP page in wp-admin loads and look for version
2) Remove that version folder from mu-plugins
3) JP page still loads but with another version


jetpack.php
```diff
@@ -31,6 +31,8 @@
 	}
 }
 
+
+
```
pschoffer:
```suggestion
```

rebeccahum (approved):

rinatkhaziev:
PR title and descriptions need a bit of a cleanup @pschoffer

PR #3511, opened 2022-08-11
Bump Jetpack submodule to 11.2

rinatkhaziev:
## Description

This concludes the process of gradually bringing up the Jetpack submodule to the latest.

See #3492 #3287 

## Checklist

## Steps to Test

Nothing to test


rinatkhaziev:
This will come handy with #3503

jesse-greathouse (approved):
👍

renatonascalves:
Just a quick notice that I started to see errors with a conflict of versions after this update:

```
Fatal error: Cannot declare class Automattic\Jetpack\Autoloader\jpf11009ded9fc4592b6a05b61ce272b3c_jetpackⓥ11_2\Autoloader, because the name is already in use in /wordpress/wp-content/mu-plugins/jetpack/vendor/jetpack-autoloader/class-autoloader.php on line 15
```

It looks like both the `jetpack` and `jetpack-11.2` plugins are using the same namespace. This is mainly affecting unit tests, the site itself is not being affected.

rinatkhaziev:
@renatonascalves, thanks for reporting. Do you mind sharing more about the context?

We don't load anything from `jetpack` by default. It's merely for legacy purposes because some folks have direct includes. I'm assuming something like that is happening in your tests?

renatonascalves:
Thanks for reaching out. I’m not intentionally loading jetpack directly though. 🤔  I’ll do a bit of digging to see what could be loading it.

This is however only happening while running unit tests (PHPUnit).

PR #3491, opened 2022-08-08
Move the object-cache files from stacks to mu-plugins

rebeccahum:
## Description
All `object-cache-*.php` logic should live in MU-plugins for transparency. Fixes [#696](https://github.com/Automattic/vip-go-mu-plugins/issues/696) and https://github.com/Automattic/vip-go-mu-plugins/issues/2077.

We'll need to update the current `object-cache.php` in stacks with the below contents:

```
<?php

/*
Plugin Name: Memcached
Description: Memcached backend for the WP Object Cache.
Version: 4.0.0
Plugin URI: http://wordpress.org/extend/plugins/memcached/
Author: Automattic

Install this file to wp-content/object-cache.php
*/

if ( file_exists( ABSPATH . '/wp-content/mu-plugins/drop-ins/object-cache/object-cache-stable.php' && file_exists( ABSPATH . '/wp-content/mu-plugins/drop-ins/object-cache/object-cache-next.php' ) ) ) {
	define( 'VIP_OBJECT_CACHE_DROPIN_STABLE', ABSPATH . '/wp-content/mu-plugins/drop-ins/object-cache/object-cache-stable.php' );
	define( 'VIP_OBJECT_CACHE_DROPIN_NEXT', ABSPATH . '/wp-content/mu-plugins/drop-ins/object-cache/object-cache-next.php' );
} else {
	// TODO: Remove after mu-plugins deploy
	define( 'VIP_OBJECT_CACHE_DROPIN_STABLE', __DIR__ . '/object-cache-stable.php' );
	define( 'VIP_OBJECT_CACHE_DROPIN_NEXT', __DIR__ . '/object-cache-next.php' );
}

// Load the next version on specified environment types
if ( ! defined( 'VIP_USE_NEXT_OBJECT_CACHE_DROPIN' ) ) {
	if ( in_array( VIP_GO_APP_ENVIRONMENT, [ 'develop', 'preprod', 'staging' ], true ) ) {
		define( 'VIP_USE_NEXT_OBJECT_CACHE_DROPIN', true );
	}
}

// If site is testing next version of object cache dropin, load it
if ( defined( 'VIP_USE_NEXT_OBJECT_CACHE_DROPIN' ) && true === VIP_USE_NEXT_OBJECT_CACHE_DROPIN && is_readable( VIP_OBJECT_CACHE_DROPIN_NEXT ) ) {
	require_once( VIP_OBJECT_CACHE_DROPIN_NEXT );
} else {
	// Else, load stable cache dropin (default)
[...]

PR #3528, opened 2022-08-13
Revert "Move the object-cache files from stacks to mu-plugins"

yvettep321:
Reverts Automattic/vip-go-mu-plugins#3491


rinatkhaziev:
@yvettep321, thanks for your contribution. Do you mind sharing what the issue was?

PR #3542, opened 2022-08-16
Add wp-parsely loader information to SDS

jesse-greathouse:
## Description

This adds wp-parsely plugin-related information to the site details index, with the helper of a new `Parsely_Loader_Info` class that is introduced to help keep track of how (and if) the plugin is being loaded.

Additionally:

- Avoid fatal errors when activating a self-installed version of the wp-parsely plugin whilst the mu-plugins version was already active.
- Cleanup loader logic, so that we can more effectively take into account the `wpvip_parsely_load_mu` filter & `_wpvip_parsely_mu` option.
- Move `maybe_load_plugin()` logic to the later `plugins_loaded` hook priority 1, rather than `muplugins_loaded`. This allows us to safely ensure a self-installed version of the plugin is not being loaded.

## Changelog Description
Information about how a site uses the parse.ly plugin has been added to the Site Details Index. The bundled version of wp-parsely plugin now loads on the `plugins_loaded` hook rather than on `muplugins_loaded` when enabled.

## Checklist

## Steps to test

Try out the various states, ensure all is loading correctly & reporting the correct loader information.

- Parsely not enabled.
- Parsely enabled by the option.
- Parsely enabled by the filter.
- Parsely enabled via a self-installed version.
- Parsely enabled by the option but disabled by the filter.


sjinks (dismissed):

config/class-site-details-index.php
```diff
@@ -228,6 +252,100 @@ public function get_jetpack_info() {
...
+	}
+
+	/**
+	 * The way that the service is being used
+	 * @param array $parsely The Parsely options.
+	 * @return string service type.
+	 */
+	public function get_parsely_service_type( $parsely ) {
+		if ( is_array( $parsely ) && isset( $parsely['apikey'] ) && '' !== $parsely['apikey'] ) {
+			$result = $this->call_parsely_api( $parsely );
```
rebeccahum: Leaving for historical purposes — there were concerns about this API call flagged on Slack by @WPprodigy

rebeccahum (dismissed):

acicovic:
I've opened a Slack thread about this PR.

config/class-site-details-index.php
```diff
@@ -228,6 +248,74 @@ public function get_jetpack_info() {
...
+		return $parsely['plugin_version'];
+	}
+
+	/**
+	 * The way that the plugin is used in WordPress
+	 * @return string integration type.
+	 */
+	public function get_parsely_integration_type() {
+		// detect a hidden installation
+		if ( false !== get_option( '_wpvip_parsely_mu', false ) && ! has_filter( 'wpvip_parsely_load_mu' ) ) {
```
mjangda: Rather than using `has_filter` to detect things, I wonder if it would be better to set have the code in `wp-parsely.php` track / share necessary details instead which would we infer the data more directly from.

e.g. could have a static class called `Parsely_Loader_Info` with methods `(get|set)_integration_type`. `wp-parsely.php` would call `Parsely_Loader_Info::set_integration_type( ... )` and here we'd fetch it via `Parsely_Loader_Info::get_integration_type()`.

jesse-greathouse: > Rather than using `has_filter` to detect things, I wonder if it would be better to set have the code in `wp-parsely.php` track / share necessary details instead which would we infer the data more directly from.
> 
> e.g. could have a static class called `Parsely_Loader_Info` with methods `(get|set)_integration_type`. `wp-parsely.php` would call `Parsely_Loader_Info::set_integration_type( ... )` and here we'd fetch it via `Parsely_Loader_Info::get_integration_type()`.

Im not opposed to this change but I'm not clear on what problem it is attempting to solve by this. I'm anxious to hear more of your wisdom on this.

config/class-site-details-index.php
```diff
@@ -228,6 +248,74 @@ public function get_jetpack_info() {
...
```
acicovic: I think that `false !==` is not needed here?

config/class-site-details-index.php
```diff
@@ -228,6 +248,74 @@ public function get_jetpack_info() {
...
+		}
+
+		// mu-plugins
+		if ( has_filter( 'wpvip_parsely_load_mu' ) ) {
+			return self::PARSELY_INTEGRATION_TYPE_MUPLUGINS;
+		}
+
+		// is enabled as a standard plugin
+		$plugin = 'wp-parsely/wp-parsely.php';
+		if ( in_array( $plugin, (array) get_option( 'active_plugins', array() ), true ) || is_plugin_active_for_network( $plugin ) ) {
```
acicovic: Have we considered using [is_plugin_active()](https://developer.wordpress.org/reference/functions/is_plugin_active/) here, or it isn't possible for some reason?

config/class-site-details-index.php
jesse-greathouse: I have moved most of the logic into the new `Parsely_Loader_info` class in `wp-parsely.php`. It has all of the functionality that you've asked for and will keep these values in static properties so that they only have to be resolved once per property.

config/class-site-details-index.php
jesse-greathouse: Thank you for the feedback. I have made this change in the most recent commit.

config/class-site-details-index.php
jesse-greathouse: Thank you for the feedback. I'm not sure why I didn't do that initially. I have made the change in the most recent commit.

config/class-site-details-index.php
```diff
@@ -228,6 +231,20 @@ public function get_jetpack_info() {
...
+	/**
+	 * Gather all the information about Parsely
+	 * @return array Parsely plugin info.
+	 */
+	public function get_parsely_info() {
+		$parsely_info                     = [];
+		$parsely_info['active']           = Pinfo::get_active();
+		$parsely_info['integration_type'] = Pinfo::get_integration_type();
+		$parsely_info['service_type']     = Pinfo::get_service_type();
+		$parsely_info['verison']          = Pinfo::get_version();
```
acicovic:
```suggestion
		$parsely_info['version']          = Pinfo::get_version();
```

config/class-site-details-index.php
```diff
@@ -228,6 +231,20 @@ public function get_jetpack_info() {
 		return $jetpack_info;
 	}
 
+	/**
+	 * Gather all the information about Parsely
```
acicovic:
```suggestion
	 * Gather all the information about Parse.ly.
```

config/class-site-details-index.php
```diff
@@ -228,6 +231,20 @@ public function get_jetpack_info() {
...
+	 * @return array Parsely plugin info.
```
acicovic:
```suggestion
	 * @return array Parse.ly plugin info.
```

wp-parsely.php
```diff
@@ -21,6 +21,189 @@
 	'3.1',
 ];
 
+/**
+ * Class to hold Parse.ly loading info
```
acicovic:
```suggestion
 * Class to hold Parse.ly loading info.
```

wp-parsely.php
```diff
@@ -21,6 +21,189 @@
...
+ * Will prevent this information from being queried multiple times
+ * by keeping the status in a static variable
```
acicovic:
```suggestion
 * by keeping the status in a static variable.
```

wp-parsely.php
```diff
@@ -21,6 +21,189 @@
...
+	 */
+	private static $service_type;
+
+	/**
+	 * @var string
+	 */
+	private static $version;
+
+	/**
+	 * fetches the active status.
```
acicovic:
```suggestion
	 * Fetches the active status.
```

wp-parsely.php
```diff
@@ -21,6 +21,189 @@
...
+	public static function get_active() {
+		if ( null === self::$active ) {
+			self::set_active();
+		}
+
+		return self::$active;
+	}
+
+	/**
+	 * populates the private $active property if it hasn't been set.
```
acicovic:
```suggestion
	 * Populates the private $active property if it hasn't been set.
```

wp-parsely.php
```diff
@@ -21,6 +21,189 @@
...
+			self::INTEGRATION_TYPE_SELF_MANAGED === $integration_type
+		) {
+			self::$active = true;
+		}
+
+		self::$active = false;
+	}
+
+	/**
+	 * fetches the integration type.
```
acicovic:
```suggestion
	 * Fetches the integration type.
```

wp-parsely.php
```diff
@@ -21,6 +21,189 @@
...
+	public static function get_integration_type() {
+		if ( null === self::$integration_type ) {
+			self::set_integration_type();
+		}
+
+		return self::$integration_type;
+	}
+
+	/**
+	 * populates the private $integration_type property if it hasn't been set.
```
acicovic:
```suggestion
	 * Populates the private $integration_type property if it hasn't been set.
```

wp-parsely.php
```diff
@@ -21,6 +21,189 @@
...
+		if ( get_option( '_wpvip_parsely_mu', false ) && ! has_filter( 'wpvip_parsely_load_mu' ) ) {
+			self::$integration_type = self::INTEGRATION_TYPE_MUPLUGINS_SILENT;
+		}
+
+		// mu-plugins
+		if ( has_filter( 'wpvip_parsely_load_mu' ) ) {
+			self::$integration_type = self::INTEGRATION_TYPE_MUPLUGINS;
+		}
+
+		// is enabled as a standard plugin
```
acicovic:
```suggestion
		// Is enabled as a standard plugin.
```

wp-parsely.php
```diff
@@ -21,6 +21,189 @@
...
+		$plugin = 'wp-parsely/wp-parsely.php';
+		if ( is_plugin_active( $plugin ) ) {
+			self::$integration_type = self::INTEGRATION_TYPE_SELF_MANAGED;
+		};
+
+		self::$integration_type = self::INTEGRATION_TYPE_NONE;
+	}
+
+	/**
+	 * fetches Parse.ly options dictionary.
```
acicovic:
```suggestion
	 * Fetches Parse.ly options dictionary.
```

wp-parsely.php
```diff
@@ -21,6 +21,189 @@
...
+	public static function get_parsely() {
+		if ( null === self::$parsely ) {
+			self::set_parsely();
+		}
+
+		return self::$parsely;
+	}
+
+	/**
+	 * populates the private $parsely property if it hasn't been set yet.
```
acicovic:
```suggestion
	 * Populates the private $parsely property if it hasn't been set yet.
```

wp-parsely.php
```diff
@@ -21,6 +21,189 @@
...
+	 */
+	private static function set_parsely() {
+		self::$parsely = get_option( 'parsely', [] );
+	}
+
+	/**
+	 * fetches the service type.
```
acicovic:
```suggestion
	 * Fetches the service type.
```

wp-parsely.php
```diff
@@ -21,6 +21,189 @@
...
+	public static function get_service_type() {
+		if ( null === self::$service_type ) {
+			self::set_service_type();
+		}
+
+		return self::$service_type;
+	}
+
+	/**
+	 * populates the private $service_type property if it hasn't been set.
```
acicovic:
```suggestion
	 * Populates the private $service_type property if it hasn't been set.
```

wp-parsely.php
```diff
@@ -21,6 +21,189 @@
...
+		) {
+			self::$service_type = self::SERVICE_TYPE_PAID;
+		}
+
+		self::$service_type = self::SERVICE_TYPE_NONE;
+
+	}
+
+	/**
+	 * fetches the version.
```
acicovic:
```suggestion
	 * Fetches the version.
```

wp-parsely.php
```diff
@@ -21,6 +21,189 @@
...
+	public static function get_version() {
+		if ( null === self::$version ) {
+			self::set_version();
+		}
+
+		return self::$version;
+	}
+
+	/**
+	 * populates the private $version property if it hasn't been set.
```
acicovic:
```suggestion
	 * Populates the private $version property if it hasn't been set.
```

acicovic (commented):
Left some spelling suggestions.

wp-parsely.php
```diff
@@ -21,6 +21,189 @@
...
+		}
+
+		return self::$integration_type;
+	}
+
+	/**
+	 * populates the private $integration_type property if it hasn't been set.
+	 */
+	private static function set_integration_type() {
+		// detect a hidden installation
```
acicovic:
```suggestion
		// Detect a hidden installation.
```

wp-parsely.php
```diff
@@ -21,6 +21,189 @@
...
+		if ( get_option( '_wpvip_parsely_mu', false ) && ! has_filter( 'wpvip_parsely_load_mu' ) ) {
+			self::$integration_type = self::INTEGRATION_TYPE_MUPLUGINS_SILENT;
+		}
+
+		// mu-plugins
```
acicovic:
```suggestion
		// Is enabled by mu-plugins.
```

wp-parsely.php
```diff
@@ -21,6 +21,189 @@
...
+		if ( has_filter( 'wpvip_parsely_load_mu' ) ) {
```
acicovic: What happens here if the `wpvip_parsely_load_mu` filter is set to `false`, and would there be any unwanted side-effects to the `set_active()` function? Is there a fix that must be applied here or in the `set_active()` function?

wp-parsely.php
```diff
@@ -21,6 +21,189 @@
...
+
+	/**
+	 * @var string
+	 */
+	private static $integration_type;
+
+	/**
+	 * @var array The Parse.ly WordPress options dictionary.
+	 */
+	private static $parsely;
```
acicovic: For clarity, I would suggest to rename this variable and related functions to `*parsely_option` or similar.

wp-parsely.php
```diff
@@ -21,6 +21,189 @@
...
+		$integration_type = self::get_integration_type();
+
+		if ( self::INTEGRATION_TYPE_MUPLUGINS === $integration_type ||
+			self::INTEGRATION_TYPE_SELF_MANAGED === $integration_type
+		) {
+			self::$service_type = self::SERVICE_TYPE_PAID;
+		}
+
+		self::$service_type = self::SERVICE_TYPE_NONE;
+
```
acicovic: This empty line (178) can be removed.

wp-parsely.php
```diff
@@ -21,6 +21,189 @@
...
+	private static function set_version() {
+		$parsely = self::get_parsely();
+
+		if ( array_key_exists( 'plugin_version', $parsely ) ) {
+			self::$version = $parsely['plugin_version'];
+		}
+
+		self::$version = self::VERSION_NONE;
+	}
+
```
acicovic: This empty line (204) can be removed.

wp-parsely.php
```diff
@@ -21,6 +21,189 @@
 	'3.1',
 ];
 
+/**
+ * Class to hold Parse.ly loading info
+ * Will prevent this information from being queried multiple times
+ * by keeping the status in a static variable
+ */
+final class Parsely_Loader_Info {
+
```
acicovic: This empty line (30) can be removed.

acicovic:
For the functions within `Parsely_Loader_Info` that return something, you might want to consider adding `@return` tags in the function comments.

wp-parsely.php
jesse-greathouse: Is there a way to tell what values are present in an applied filter?

wp-parsely.php
jesse-greathouse: You are right. This should be renamed. I have refactored this property, as well as its getter/setter methods, in the latest commit.

wp-parsely.php
jesse-greathouse: Thank you. I have corrected this in the latest commit.

wp-parsely.php
jesse-greathouse: Very good. I have changed this in the latest commit.

wp-parsely.php
jesse-greathouse: Great! I have removed this line in the latest commit.

jesse-greathouse:
> For the functions within `Parsely_Loader_Info` that return something, you might want to consider adding `@return` tags in the function comments.

Thank you. This has been added in the latest commit.

wp-parsely.php
acicovic: I would think there is, but a quick search doesn't reveal anything. I can look into that more tomorrow if needed.

acicovic: Maybe what we could do (and maybe what @mjangda was suggesting?) is to change the `Parsely_Loader_Info` state from within the `maybe_load_plugin()` and `maybe_disable_some_features()` functions where the actual operations take place? @jesse-greathouse, can you give it a shot and see if this approach can work?

If the answer is yes, I can help you out more if needed.

jesse-greathouse: @acicovic,

The latest commit is trying to use this suggestion and I think it works well. The commit notes are as follows:

```
Changed Parsely_Loader_Info setter methods to public
Gave Parsely_Loader_Info setter methods a parameter to pass the value
Integreated Parsely_Loader_Info with maybe_load_plugin and maybe_disable_some_features methods in wp-parsely.php
Added appropriate docblock metadata for the new parameters in setters
Removed unnecessary code in Parsely_Loader_Info methods
```

If you do not mind, please look over the way I used the setter methods in `maybe_load_plugin` and `maybe_disable_some_features` to be sure that it is correctly setting the appropriate values. If I messed it up, it should be easy to fix.

wp-parsely.php
```diff
@@ -47,12 +226,44 @@ function maybe_load_plugin() {
...
+		if ( $plugin_is_already_loaded || $plugin_loading_queued ) {
+			Parsely_Loader_Info::set_active( true );
+			Parsely_Loader_Info::set_integration_type( Parsely_Loader_Info::INTEGRATION_TYPE_SELF_MANAGED );
+
+			//Get the version of the self-managed wp-parsely plugin.
+			$parsely_options = Parsely_Loader_Info::get_parsely_options();
+			if ( array_key_exists( 'plugin_version', $parsely_options ) ) {
+				Parsely_Loader_Info::set_version( $parsely_options['plugin_version'] );
+			}
+		}
```
acicovic: Question: I'm wondering whether we want to run this during plugin activation? Opinions?

acicovic:
I'm ready to approve this PR when [this](https://github.com/Automattic/vip-go-mu-plugins/pull/3542#discussion_r957059149) gets sorted out,but I'd need feedback from other reviewers as well. Thanks!

wp-parsely.php
jesse-greathouse: I don't have strong opinions on this. I've removed it in the latest commit.

wp-parsely.php
```diff
@@ -47,12 +226,44 @@ function maybe_load_plugin() {
...
 	 * add_filter( 'wpvip_parsely_load_mu', '__return_false' );
 	 */
-	if ( ! apply_filters( 'wpvip_parsely_load_mu', get_option( '_wpvip_parsely_mu' ) === '1' ) ) {
-		return;
-	}
+	$do_not_load_plugin = ! apply_filters( 'wpvip_parsely_load_mu', get_option( '_wpvip_parsely_mu' ) === '1' );
+
+	// Check if wp-parsely has already been loaded, for example when the user
+	// has a self-managed wp-parsely installation.
+	$plugin_is_already_loaded = in_array( PARSELY_PLUGIN_SIGNATURE, get_option( 'active_plugins' ) );
```
chriszarate: guard against missing option
```suggestion
	$plugin_is_already_loaded = in_array( PARSELY_PLUGIN_SIGNATURE, get_option( 'active_plugins', [] ), true );
```

wp-parsely.php
```diff
@@ -47,12 +226,44 @@ function maybe_load_plugin() {
 	 *
 	 * add_filter( 'wpvip_parsely_load_mu', '__return_false' );
 	 */
-	if ( ! apply_filters( 'wpvip_parsely_load_mu', get_option( '_wpvip_parsely_mu' ) === '1' ) ) {
-		return;
-	}
+	$do_not_load_plugin = ! apply_filters( 'wpvip_parsely_load_mu', get_option( '_wpvip_parsely_mu' ) === '1' );
```
chriszarate: having this spelled out and separate is useful

```suggestion
	$load_plugin_requested_via_option = get_option( '_wpvip_parsely_mu' ) === '1';
	$do_not_load_plugin = ! apply_filters( 'wpvip_parsely_load_mu', $load_plugin_requested_via_option );
```

wp-parsely.php
```diff
@@ -47,12 +226,44 @@ function maybe_load_plugin() {
...
+
+		// Attempt to detect force-disables invoked by the wpvip_parsely_load_mu filter.
+		// Set Loader Info data as an inactive mu-plugins integration.
+		$force_disabled = $do_not_load_plugin && has_filter( 'wpvip_parsely_load_mu' )
+			&& false === $plugin_is_already_loaded && false === $plugin_loading_queued;
+		if ( $force_disabled ) {
+			Parsely_Loader_Info::set_active( false );
+			Parsely_Loader_Info::set_integration_type( Parsely_Loader_Info::INTEGRATION_TYPE_MUPLUGINS );
+
+			if ( get_option( '_wpvip_parsely_mu' ) === '1' ) {
```
chriszarate:
```suggestion
			if ( $load_plugin_requested_via_option ) {
```

wp-parsely.php
```diff
@@ -47,12 +226,44 @@ function maybe_load_plugin() {
...
+	// Check if wp-parsely has already been loaded, for example when the user
+	// has a self-managed wp-parsely installation.
+	$plugin_is_already_loaded = in_array( PARSELY_PLUGIN_SIGNATURE, get_option( 'active_plugins' ) );
+
+	// Check if wp-parsely is in a state of attempting to be activated
+	// for a self-managed wp-parsely installation.
+	$plugin_loading_queued = is_queued_for_activation();
+
+	// If the plugin must not be loaded, set Loader Info data and exit.
+	if ( $do_not_load_plugin || $plugin_is_already_loaded || $plugin_loading_queued ) {
```
chriszarate: This is ... incredibly complicated with not enough comments. It also feels like this is in the wrong order. It would benefit from very descriptive variable names and early returns, like:

```
if ( $self_managed_plugin_has_loaded || $self_managed_plugin_is_queued_for_activation ) {
  Parsely_Loader_Info::set_active( true );
  Parsely_Loader_Info::set_integration_type( Parsely_Loader_Info::INTEGRATION_TYPE_SELF_MANAGED );
  return;
}

if ( $load_plugin_requested_via_option && $force_disabled ) {
  Parsely_Loader_Info::set_integration_type( Parsely_Loader_Info::INTEGRATION_TYPE_MU_PLUGINS );
  return;
}

// etc
```

wp-parsely.php
jesse-greathouse: Thank you. This has been corrected in the most recent commit.

wp-parsely.php
jesse-greathouse: Awesome suggestion. Thanks 👍

wp-parsely.php
jesse-greathouse: Very good. I have added this.

wp-parsely.php
jesse-greathouse: Thank you for the excellent suggestions for this very convoluted logic. This section has been one of the most difficult things about working on this piece of code. Just understanding what's happening is a nightmare. Thanks to your suggestion I have made this section a lot less convoluted and more readable, and I've added a lot of relevant comments to help the next person who might have to read this. I feel that the latest commit makes this section a lot better. Thank you very much for your time and great thoughts on this.

wp-parsely.php
```diff
@@ -47,12 +226,43 @@ function maybe_load_plugin() {
...
+	// The activation of this plugin has been queued in this request.
+	$self_managed_plugin_is_queued_for_activation = is_queued_for_activation();
+
+	// Check if wp-parsely has already been loaded, for example when the user
+	// has a self-managed wp-parsely installation.
+	$self_managed_plugin_has_loaded = in_array( PARSELY_PLUGIN_SIGNATURE, get_option( 'active_plugins', [] ) );
+
+	if ( $self_managed_plugin_has_loaded || $self_managed_plugin_is_queued_for_activation ) {
+		$parsely_options = Parsely_Loader_Info::get_parsely_options();
+		Parsely_Loader_Info::set_active( true );
```
acicovic: My only remaining question here is whether we should execute this for `$self_managed_plugin_is_queued_for_activation`, instead of just returning. @chriszarate, @mjangda, opinions?

jesse-greathouse: > My only remaining question here is whether we should execute this for `$self_managed_plugin_is_queued_for_activation`, instead of just returning. @chriszarate, @mjangda, opinions?

Adding @rinatkhaziev and @WPprodigy. They may have some thoughts on the question by @acicovic

acicovic: Thanks!

WPprodigy: All the `is_queued_for_activation()` related code paths (and tests) feels like it's adding more complication than is worth IMO. Can we just strip that part out? I don't _think_ we really need to know right that second on the activation requests themselves. The next PHP request will know about it's activated/deactivation, and we're never going to be activating & syncing sds data on the same request.

wp-parsely.php
```diff
@@ -47,12 +226,43 @@ function maybe_load_plugin() {
...
 	 * add_filter( 'wpvip_parsely_load_mu', '__return_false' );
 	 */
-	if ( ! apply_filters( 'wpvip_parsely_load_mu', get_option( '_wpvip_parsely_mu' ) === '1' ) ) {
+
+	// The activation of this plugin has been queued in this request.
+	$self_managed_plugin_is_queued_for_activation = is_queued_for_activation();
+
+	// Check if wp-parsely has already been loaded, for example when the user
+	// has a self-managed wp-parsely installation.
+	$self_managed_plugin_has_loaded = in_array( PARSELY_PLUGIN_SIGNATURE, get_option( 'active_plugins', [] ) );
```
WPprodigy: This isn't a safe enough protection to use to abort early if the plugin is already loaded. The plugin signature does not have to be `wp-parsely/wp-parsely.php`, it could be anything like `wp-parsely-2/wp-parsely.php` for example. Or it could be manually `require`'d in client-mu-plugins, etc.

We need to return to the safety of `class_exists( 'Parsely' ) || class_exists( 'Parsely\Parsely' )` type checks.

wp-parsely.php
jesse-greathouse: This part was not added to record the info for SDS. `is_queued_for_activation()` was added because when a user activates a self managed Parse.ly plugin installation, and the code has `add_filter( 'wpvip_parsely_load_mu', '__return_true' );` somewhere in their code, there will be a class loading collision between the mu-plugins integrated wp-parsely, and their self-managed installation.
Screen Shot 2022-08-26 at 9 21 12 AM

wp-parsely.php
jesse-greathouse: In my testing, `class_exists( 'Parsely' ) || class_exists( 'Parsely\Parsely' )` never actually worked when the user was using a self-managed plugin. I believe it is because this code is executed before the plugin was loaded.

wp-parsely.php
jesse-greathouse: > ...and we're never going to be activating & syncing sds data on the same request.

@acicovic, I think this is the answer we've been seeking on your question. Based on @WPprodigy's input, there will not be a need to populate this data for use in the SDS request, because there will not be a sync to SDS on this request. I will make the necessary change that will short circuit this logic when activation is queued on this request.

wp-parsely.php
jesse-greathouse: Please see the changes in my latest commit which address all of the concerns that you brought up. I have tested the scenarios and it seems to be all just fine. 

The `class_exists( 'Parsely' ) || class_exists( 'Parsely\Parsely' )` logic does work as expected when a user is doing a direct require in `client-mu-plugins/plugin-loader.php` or putting their wp-parsely directly in `client-mu-plugins`.

If this class is not found, it reverse to the methodology of checking the `active_plugins` list, but this time it uses a regex pattern to identify anything that looks like it's a wp-parsely plugin. e.g. `wp-parsely2/wp-parsely.php`, `wp-parsely-1.10/wp-parsely.php`, etc...

I believe that if there is any more of a wild Parse.ly plugin out there, then this would be such a small statistical anomaly that it's not really all that relevant for this scope.

Here is the commit log:

* Added back in `class_exists( 'Parsely' ) || class_exists( 'Parsely\Parsely' )` logic to account for when the plugin is loaded through `client-mu-plugins`.
* Added function: `has_plugin_signature` which uses a regex pattern to determine if a plugin is loaded that looks like wp-parsely.
* Changed constant `PARSELY_PLUGIN_SIGNATURE` to `PARSELY_PLUGIN_SIGNATURE_REGEX` to represent the pattern which a wp-parsely plugin could look like.
* Updated tests and added tests for `has_plugin_signature`
* Thoroughly tested all scenarios so that I know it works.

wp-parsely.php
jesse-greathouse: I have addressed this in my latest commit. [The SDS info will not be captured when the request has an activation queued](https://github.com/Automattic/vip-go-mu-plugins/pull/3542/files#diff-41f0c604e0848ecb9ac1b6e07d5d2fab90b2ef3f715792c15243c5eed180894cR249).

acicovic: Great, thank you both!

rinatkhaziev:
@jesse-greathouse Parse.ly-related tests are failing.

jesse-greathouse:
> @jesse-greathouse Parse.ly-related tests are failing.

Tnx, @rinatkhaziev. I fixed the tests in the latest commit.

wp-parsely.php
```diff
@@ -21,6 +21,67 @@
...
+		self::$integration_type = $integration_type;
+	}
+
+	public static function get_service_type(): string {
+		return isset( self::$service_type ) ? self::$service_type : self::SERVICE_TYPE_UNKNOWN;
+	}
+
+	public static function set_service_type( string $service_type ): void {
+		self::$service_type = $service_type;
+	}
```
WPprodigy: Should we strip these out until/if we need them? Not a big fan of shipping unused "maybe we use this in the future" type code. Always easier to add than to remove later :)

WPprodigy (dismissed):
I updated the PR description to account for what all is changed here.

Also did a bunch of testing and all looks good as far as I can tell! Used this in my testing as I went through the various states:

```
	$parsely_info                     = [];
	$parsely_info['active']           = ParselyInfo::is_active();
	$parsely_info['integration_type'] = ParselyInfo::get_integration_type();
	$parsely_info['service_type']     = ParselyInfo::get_service_type();
	$parsely_info['version']          = ParselyInfo::get_version();
	error_log( print_r( $parsely_info, true ) );
} );

// add_filter( 'wpvip_parsely_load_mu', '__return_true' );
// add_filter( 'pre_option__wpvip_parsely_mu', fn() => '1' );
```

And then I confirmed `Automattic\VIP\Config\Site_Details_Index::instance()->get_site_details();` output looks good.

We do need to be vigilant on deploy due to the `plugins_loaded` hook change, but I'm quite sure it should go over smoothly.

wp-parsely.php
acicovic: IMO we can strip these out.

jesse-greathouse: I have removed this in the latest commit.

jesse-greathouse: I will mirror that change in the SDS code.

rinatkhaziev:
> I'm quite sure it should go over smoothly.

Famous last words :)

We should check/test against customers who load the standalone versions prior to the release.

WPprodigy:
Can definitely check/test those to be safe. More details on how it could break:

If a self-installed version is loaded via UI/wpcom_vip_load_plugin(), all is fine. The only time it could go poorly is if the plugin file is manually require'd on a hook later than `plugins_loaded` priority 1 - but doing so would have caused the previous (current) code to fatal out anyway though (unless using the filter to disable).

For things to break that weren't currently broken, you'd have to use the `muplugins_loaded` hook to see if we loaded parsely, and then use the result of that check to manually require the plugin on `plugins_loaded` after priority 1. Lots of random hoops to jump through in order to make this change not backwards-compat. All in all, extremely unlikely bordering impossible.

The other scenario that would change is unhooking the `maybe_load_plugin` hook - but we've documented using the `wpvip_parsely_load_mu ` filter instead of that route.

WPprodigy (approved):

PR #3543, opened 2022-08-16
Adjust VIP Jetpack loader to safe-guard against an edge-case where Jetpack is present in active_plugins

rinatkhaziev:
## Description

VIP loads Jetpack via a custom bootstrap function vip_jetpack_load(). Jetpack's [Autoloader](https://github.com/Automattic/jetpack-autoloader/blob/trunk/src/class-plugin-locator.php#L54) tries to determine plugin paths based on `active_plugins`. 

In case `jetpack/jetpack.php` is present in the option, Jetpack's autoloader would try to load both. In the best-case scenario, nothing happens except loading a bunch of Jetpack classes for a different version, but in the worst-case scenario (where the current global deployed version matches the one in `jetpack`, it results in a fatal.

Screen Shot 2022-08-16 at 1 45 04 PM

This PR addresses that by explicitly removing entries for Jetpack via the `option_active_plugins` filter. 

### Plugin Updated: VIP Jetpack Loader

We implemented a safeguard against potentially loading Jetpack classes from different paths.

## Checklist

## Steps to Test

The easiest way to test is to step through XDebug.

Before applying the PR: 
1. In `wp shell`: `update_option( 'active_plugins', [ 'jetpack/jetpack.php' ] );`
2. Verify the site is now fataling
3. Apply the PR
4. Verify site is not fataling anymore
5. For a good measure go to `jetpack/vendor/jetpack-autoloader/class-autoloader` and set a breakpoint at `class Autoloader`
6. Run a request with Xdebug and verify the breakpoint doesn't trigger.

### DIFF
--- a/jetpack.php
+++ b/jetpack.php
@@ -164,7 +164,7 @@ function vip_jetpack_load() {
 					return $option;
 				}
 
-				foreach( $option as $i => $plugin ) {
+				foreach ( $option as $i => $plugin ) {
 					if ( wp_endswith( $plugin, '/jetpack.php' ) ) {
 						unset( $option[ $i ] );
 						break;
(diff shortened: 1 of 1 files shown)


rinatkhaziev:
props @renatonascalves for the bug report.

sjinks (approved):

PR #3569, opened 2022-08-25
Hotfix - Bring back unprefixed environment variables

jesse-greathouse:
Legacy systems may require un-prefixed environment variables.


chriszarate (changes requested):
I'd rather we not "fix" it this way

renatonascalves:
I'm not sure why this was removed. ¯\_(ツ)_/¯ But I'd guess others, like me, relied on this functionality.

PR #3574, opened 2022-08-29
Fix fatal error when Jetpack network active

anubisthejackle:
## Description
This corrects a fatal error where Jetpack attempts to autoload the same class when it has been network activated previously.

The existing work around makes sure that if Jetpack is in the `active_plugins` array that it is ignored, however network activated plugins are stored in a different option, which isn't impacted by this filter.

## Changelog Description
Updates the Jetpack loader to account for network activated installations of Jetpack.

## Checklist

## Steps to Test
1. Create a new local Multisite
1. Install Jetpack
1. Network activate Jetpack
1. Without disabling Jetpack, remove Jetpack from the plugins
1. Install the vip-go-mu-plugins-built to your local site
1. Attempt to load your site.

### DIFF
--- a/jetpack.php
+++ b/jetpack.php
@@ -173,20 +173,22 @@ function vip_jetpack_load() {
 				return $option;
 			} );
 
-		    	// The same edge case as above, but for when Jetpack is network activated.
-			add_filter( 'site_option_active_sitewide_plugins', function( $option ) {
-				if ( ! is_array( $option ) ) {
-					return $option;
-				}
-
-				foreach ( $option as $plugin => $i ) {
-					if ( wp_endswith( $plugin, '/jetpack.php' ) ) {
-						unset( $option[ $plugin ] );
-						break;
+			if ( is_multisite() ) {
+			    	// The same edge case as above, but for when Jetpack is network activated.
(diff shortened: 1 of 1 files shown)


jetpack.php
```diff
@@ -173,6 +173,21 @@ function vip_jetpack_load() {
 				return $option;
 			} );
 
+		    	// The same edge case as above, but for when Jetpack is network activated.
+			add_filter( 'site_option_active_sitewide_plugins', function( $option ) {
```
rebeccahum: Can you add a `is_multisite()` check to this?

rebeccahum (changes requested):
This seems unlikely to ever happen on production since AFAIK for multisite, Jetpack is activated on a site-by-site basis on VIP. However, it's not a bad idea to account for edge cases like this locally too.

jetpack.php
anubisthejackle: Done and done.

anubisthejackle:
> ... it's not a bad idea to account for edge cases like this locally too.

That's exactly where I noticed the issue. Setting up a client site locally, and the DB backup contained Jetpack network activated. It took some debugging to figure out exactly what was happening.

rebeccahum (approved):

PR #3580, opened 2022-08-30
CANTINA-902: Define DB_NAME, DB_HOST, DB_USER, DB_PASSWORD if not defined

rebeccahum:
## Description
To prevent PHP 8 fatal errors, we should provide the DB_* constants to connect to the primary database to support 3rd party integrations utilizing them. Since we use HyperDB, these constants aren't defined by default.

### Plugin Updated: VIP Init (non-production environments only)

Define DB_NAME, DB_HOST, DB_USER, DB_PASSWORD if not defined.

## Checklist

## Steps to Test
1) Sandbox and shell in
2) Check that the constants are defined now


000-vip-init.php
```diff
@@ -265,4 +265,28 @@
 	add_filter( 'wp_sitemaps_enabled', '__return_false' );
 }
 
+// We use HyperDB, so DB_* constants are not automatically defined
+if ( ! defined( 'DB_NAME' ) && ! defined( 'DB_HOST' ) && ! defined( 'DB_PASSWORD' ) && ! defined( 'DB_NAME' ) ) {
```
sjinks:
```suggestion
if ( ! defined( 'DB_NAME' ) && ! defined( 'DB_HOST' ) && ! defined( 'DB_PASSWORD' ) && ! defined( 'DB_USER' ) ) {
```

000-vip-init.php
```diff
@@ -265,4 +265,28 @@
...
+	global $db_servers;
+	if ( ! is_array( $db_servers ) ) {
+		return;
+	}
+
+	$writable_servers = array_filter( $db_servers, function( $db_server ) {
+		return isset( $db_server[5] ) && $db_server[5] > 0;
```
sjinks:
```suggestion
		return ! isset( $db_server[5] ) || $db_server[5] > 0;
```

Per the [documentation](https://github.com/Automattic/HyperDB/blob/trunk/db-config.php#L81-L82), this parameter is optional and is considered to be `1` if not defined.

000-vip-init.php
```diff
@@ -265,4 +265,28 @@
...
+	} );
+
+	usort( $writable_servers, function( $a, $b ) {
+		return $a[5] > $b[5];
```
sjinks:
```suggestion
		return ( $a[5] ?? 1 ) - ( $b[5] ?? 1 );
```

`??` is because the fifth index, in theory, may not be defined (see above). `-` instead of `>` is because the callback is [expected to return](https://www.php.net/usort) a tri-state integer, not a boolean, value.

000-vip-init.php
```diff
@@ -265,4 +265,28 @@
...
+	});
+
+	$db = $writable_servers[0];
```
sjinks:
```suggestion
	$db = $writable_servers[0] ?? null;
```

or

```php
	$db = $writable_servers[0] ?? $db_servers[0];
```

in the case when there are no writable servers (won't happen on VIP, but can happen in a dev environment).

sjinks (changes requested):

rebeccahum:
Need to look to see why e2e is failing now...

rebeccahum:
E2E fixed via https://github.com/Automattic/vip-go-mu-plugins/pull/3587

000-vip-init.php
```diff
@@ -288,5 +288,29 @@
 	add_filter( 'wp_sitemaps_enabled', '__return_false' );
 }
 
+// We use HyperDB, so DB_* constants are not automatically defined
+if ( ! defined( 'DB_NAME' ) && ! defined( 'DB_HOST' ) && ! defined( 'DB_PASSWORD' ) && ! defined( 'DB_USER' ) ) {
+	global $db_servers;
```
mjangda: Can we move this logic to a function and add some tests for it?

We could also simplify it by fetching the db details from `$wpdb->hyper_servers` (https://github.com/Automattic/HyperDB/blob/trunk/db.php#L266-L268), although, it would be nicer/safer if hyperdb had a `get` method for that.

mjangda (changes requested):
Added a suggestion.

We'll want to roll this out carefully with advance notice to customers to avoid unexpected surprises.

000-vip-init.php
```diff
@@ -288,5 +288,29 @@
 	add_filter( 'wp_sitemaps_enabled', '__return_false' );
 }
 
+// We use HyperDB, so DB_* constants are not automatically defined
```
```suggestion
// We use HyperDB, so DB_* constants are not automatically defined.
// We define them here in case any customer code is referencing them to avoid fatals errors in PHP 8+.
// As a best practice, these constants should not be used directly (`_doing_it_wrong()`).
```

000-vip-init.php
```diff
@@ -288,5 +288,46 @@
...
+ * As a best practice, these constants should not be used directly (`_doing_it_wrong()`).
+ *
+ * @return void
+ */
+function vip_define_db_constants() {
+	if ( defined( 'DB_NAME' ) || defined( 'DB_HOST' ) || defined( 'DB_PASSWORD' ) || defined( 'DB_USER' ) ) {
+		return;
+	}
+
+	if ( ! method_exists( 'hyperdb', 'get_hyper_servers' ) ) {
```
sjinks:
```suggestion
    global $wpdb;
	if ( ! method_exists( $wpdb, 'get_hyper_servers' ) ) {
```

because we call `get_hyper_servers()` on `$wpdb`.

000-vip-init.php
```diff
@@ -288,5 +288,46 @@
...
+	}
+
+	global $wpdb;
+	$db_servers = $wpdb->get_hyper_servers( 'write' );
+
+	if ( ! is_array( $db_servers ) ) {
+		return;
+	}
+
+	if ( isset( $db_servers[1] ) ) {
```
sjinks: Maybe we should add a comment explaining why we are taking `$db_servers[1]` instead of `$db_servers[0]`...

sjinks (changes requested):

000-vip-init.php
rebeccahum: Moved to a function now...couldn't quite figure out how to mock hyperdb with PHPUnit atm since it uses the native $wpdb.

sjinks: What do you think if:

```diff
-function vip_define_db_constants() {
+function vip_define_db_constants( $db ) {
...
-add_action( 'vip_loaded', 'vip_define_db_constants' );
+add_action( 'vip_loaded', function () {
+    global $wpdb;
+    vip_define_db_constants( $wpdb );
+} );
```

?

Then we can create a helper class with `get_hyper_servers()` and test the function.

sjinks: But we won't be unable to mock `defined()` in `if ( defined( 'DB_NAME' ) || defined( 'DB_HOST' ) || defined( 'DB_PASSWORD' ) || defined( 'DB_USER' ) ) {` because the function is not in a namespace. We will probably have to move it.

sjinks: If you are OK with that, I can implement this.

sjinks (dismissed):

000-vip-init.php
```diff
@@ -288,5 +288,14 @@
 	add_filter( 'wp_sitemaps_enabled', '__return_false' );
 }
 
+require_once __DIR__ . '/001-core/constants.php'; // Define the DB constants
+// Roll out to non-prods only for now
+if ( defined( 'VIP_GO_APP_ENVIRONMENT' ) && 'production' !== constant( 'VIP_GO_APP_ENVIRONMENT' ) ) {
```
sjinks, line 297: 👍🏼

rebeccahum:
@mjangda Do you think this is something that will require a Lobby post or if the VIP Cloud Changelog is enough?

001-core/constants.php
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * We use HyperDB, so DB_* constants are not automatically defined.
+ * We define them here in case any customer code is referencing them to avoid fatals errors in PHP 8+.
+ * As a best practice, these constants should not be used directly (`_doing_it_wrong()`).
+ *
+ * @param Object $wpdb
+ * @return void
+ */
+function vip_define_db_constants( $wpdb ) {
```
GaryJones: The function doesn't need the `vip_` prefix, since it is namespace sufficiently to avoid a clash.

Also, can we at least add the return type `: void`?

001-core/constants.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+namespace Automattic\VIP\Core\Constants;
+
+/**
+ * We use HyperDB, so DB_* constants are not automatically defined.
+ * We define them here in case any customer code is referencing them to avoid fatals errors in PHP 8+.
+ * As a best practice, these constants should not be used directly (`_doing_it_wrong()`).
```
GaryJones: Give a more accurate Summary (first line of a DocBlock), and let the rest be the Description.

```suggestion
 * Define DB_HOST, DB_USER, DB_PASSWORD, and DB_NAME constants.
 *
 * WPVIP uses use HyperDB, so DB_* constants are not automatically defined.
 * As a best practice, these constants should not be used directly, but we define them here in
 * case any customer code does to avoid fatal errors in PHP 8+.
```

001-core/constants.php
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @param Object $wpdb
+ * @return void
+ */
+function vip_define_db_constants( $wpdb ) {
+	if ( defined( 'DB_NAME' ) || defined( 'DB_HOST' ) || defined( 'DB_PASSWORD' ) || defined( 'DB_USER' ) ) {
+		return;
```
GaryJones, line 16: Could there ever be a situation where only some of these are defined?

I wonder if this function could be tweaked so that the early guard clause checks for at least one of them being undefined, and then adds a `defined()` wrapper around each of the `define()`'s at the end. That way, we fill the gaps, rather than do an all-or-nothing approach.

```suggestion
	if ( defined( 'DB_NAME' ) && defined( 'DB_HOST' ) && defined( 'DB_PASSWORD' ) && defined( 'DB_USER' ) ) {
	    // All DB_* constants are defined, so nothing else to do here.
		return;
```

then:

```
if ( ! defined( `DB_HOST` ) {
	define( 'DB_HOST', $db[0]['host'] );
}
```

etc.

Of course, the downside is that then the collection of constants may be inconsistent with each other in terms of a workable config, so maybe instead of just a return if only some of the constants are defined, there's an opportunity to capture / log that somewhere?

sjinks: > Could there ever be a situation where only some of these are defined?

This should not happen. "Normal" WordPress installations have all these constants defined. VIP installations have no constants defined because they use `$db_servers` and HyperDB.

In theory, a user on a "normal" installation can define only some of these constants; but if this is the case, they probably know what they are doing and why.

001-core/constants.php
sjinks: Done.

sjinks (dismissed):

001-core/constants.php
rebeccahum: >  That way, we fill the gaps, rather than do an all-or-nothing approach.

I prefer the all-or-nothing approach in this aspect because if only some are defined, then something is definitely off and the data being supplied from us may not be in tangent with what is already defined.

sjinks:
Failing nightly tests are fixed in #3621

sjinks (dismissed):

000-vip-init.php
```diff
@@ -287,5 +289,13 @@
 	add_filter( 'wp_sitemaps_enabled', '__return_false' );
 }
 
+require_once __DIR__ . '/001-core/constants.php'; // Define the DB constants
+
+// Roll out to non-prods only for now
+if ( defined( 'VIP_GO_APP_ENVIRONMENT' ) && 'production' !== constant( 'VIP_GO_APP_ENVIRONMENT' ) ) {
+	global $wpdb;
+	define_db_constants( $wpdb );
```
mjangda: Any concerns with setting `global` in the file execution like this? Would it be cleaner / safer to do something like this instead:

```suggestion
	define_db_constants( $GLOBALS['wpdb'] );
```

001-core/constants.php
```diff
@@ -0,0 +1,43 @@
...
+function define_db_constants( $wpdb ): void {
+	if ( defined( 'DB_NAME' ) || defined( 'DB_HOST' ) || defined( 'DB_PASSWORD' ) || defined( 'DB_USER' ) ) {
+		return;
+	}
+
+	if ( ! method_exists( $wpdb, 'get_hyper_servers' ) ) {
+		return;
+	}
+
+	$db_servers = $wpdb->get_hyper_servers( 'write' );
```
mjangda: We'll want the hyperdb update rolled out everywhere otherwise this obviously won't work.

001-core/constants.php
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Define DB_HOST, DB_USER, DB_PASSWORD, and DB_NAME constants.
+ *
+ * WPVIP uses use HyperDB, so DB_* constants are not automatically defined.
+ * As a best practice, these constants should not be used directly, but we define them here in
+ * case any customer code does to avoid fatal errors in PHP 8+.
+ *
+ * @param object $wpdb
+ */
+function define_db_constants( $wpdb ): void {
```
mjangda: To avoid confusion with the global `wpdb`, we should probably call this `$hyperdb`

001-core/constants.php
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	if ( isset( $db_servers[1] ) ) {
+		// The index is based on the "write priority", which defaults to 1.
+		// See https://github.com/Automattic/HyperDB/blob/06dc2449535c35aaaf2f7e1e8a28fb40e59ff23e/db.php#L257-L267
+		$db = $db_servers[1];
+	} else {
+		$keys = array_keys( $db_servers );
+		$db   = $db_servers[ $keys[0] ] ?? null;
+	}
```
mjangda: Minor but I think the better approach here would be to sort by priority and pull the one with the highest value / index:  

https://github.com/Automattic/HyperDB/blob/06dc2449535c35aaaf2f7e1e8a28fb40e59ff23e/db.php#L558-L559

Not sure if that impacts performance though.

mjangda (dismissed):
Looking good. Made some minor notes but nothing blocking.

000-vip-init.php
rinatkhaziev: ++ to what Mo said.

001-core/constants.php
rinatkhaziev: I don't think a sort on an array of < 10 elements would be that big of a deal

001-core/constants.php
rinatkhaziev: I think that's rolled out already

000-vip-init.php
```diff
@@ -287,5 +289,13 @@
 	add_filter( 'wp_sitemaps_enabled', '__return_false' );
 }
 
+require_once __DIR__ . '/001-core/constants.php'; // Define the DB constants
```
rinatkhaziev: Need a file_exists check to prevent intermittent fatals. Alternatively can split in two deploys.

rinatkhaziev (dismissed):
Left some feedback.

000-vip-init.php
```diff
@@ -287,5 +289,14 @@
 	add_filter( 'wp_sitemaps_enabled', '__return_false' );
 }
 
+if ( file_exists( __DIR__ . '/001-core/constants.php' ) ) {
+	require_once __DIR__ . '/001-core/constants.php'; // Define the DB constants
+}
+
+// Roll out to non-prods only for now
+if ( defined( 'VIP_GO_APP_ENVIRONMENT' ) && 'production' !== constant( 'VIP_GO_APP_ENVIRONMENT' ) ) {
+	define_db_constants( $GLOBALS['wpdb'] );
```
rinatkhaziev: this would need a `function_exists` check given the `file_exists` check above.

sjinks (approved):

PR #3621, opened 2022-09-06
Fix failing tests

sjinks:
Ref: https://github.com/Automattic/vip-go-mu-plugins/runs/8217157462?check_suite_focus=true


tests/test-performance-lastpostmodified.php
```diff
@@ -103,6 +103,6 @@ public function test__override_lastpostmodified__is_set_post() {
 	public function test__override_lastpostmodified__is_not_set_post() {
 		$actual = get_lastpostmodified( 'gmt', 'post' );
 
-		$this->assertEquals( $this->post_modified_gmt, $actual );
```
sjinks, line 106: This property did not exist :-) interestingly, the test failed only with WP nightly.

rebeccahum (approved):

PR #3678, opened 2022-09-22
add vip admin notice filter

davidsword:
## Description

Make it easier to conditionally not show specific VIP admin notices.

Useful for anyone wanting to dynamically suppress our messages for certain caps, or any other reason. Ex hiding php notice from non super admins:

```php
add_filter('vip_admin_notice_hide_php8-migrations-3', function($show) {
	return ! is_super_admin();
});
```

Might be better to have just `vip_hide_admin_notice` to blanket hide all VIP admin_notices as an option instead of requiring the optional ID for each new message.

## Checklist

### DIFF
--- a/admin-notice/class-admin-notice-controller.php
+++ b/admin-notice/class-admin-notice-controller.php
@@ -19,8 +19,8 @@ class Admin_Notice_Controller {
 		$dismissed_notices = get_user_meta( get_current_user_id(), self::DISMISS_USER_META );
 
 		$filtered_notices = array_filter( $this->all_notices, function ( $notice ) use ( $dismissed_notices ) {
-			$is_dismissed = in_array( $notice->dismiss_identifier, $dismissed_notices );
-			$is_hidden_with_filter = apply_filters('vip_admin_notice_hide_'.$notice->dismiss_identifier, false);
+			$is_dismissed          = in_array( $notice->dismiss_identifier, $dismissed_notices );
+			$is_hidden_with_filter = apply_filters( 'vip_admin_notice_hide_' . $notice->dismiss_identifier, false );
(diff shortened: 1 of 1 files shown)


admin-notice/class-admin-notice-controller.php
```diff
@@ -19,7 +19,9 @@ public function display_notices() {
 		$dismissed_notices = get_user_meta( get_current_user_id(), self::DISMISS_USER_META );
 
 		$filtered_notices = array_filter( $this->all_notices, function ( $notice ) use ( $dismissed_notices ) {
-			if ( $notice->dismiss_identifier && in_array( $notice->dismiss_identifier, $dismissed_notices ) ) {
+			$is_dismissed = in_array( $notice->dismiss_identifier, $dismissed_notices );
+			$is_hidden_with_filter = apply_filters('vip_admin_notice_hide_'.$notice->dismiss_identifier, false);
```
sjinks:
```suggestion
			$is_dismissed          = in_array( $notice->dismiss_identifier, $dismissed_notices );
			$is_hidden_with_filter = apply_filters( 'vip_admin_notice_hide_' . $notice->dismiss_identifier, false );
```

to keep phpcs happy

sjinks (changes requested):

rebeccahum:
Given that the notices usually only display important information, I'm not sure we really want to grant the capability to dynamically disable them. 🙂 Especially if the notices have a [Capability_Condition ](https://github.com/Automattic/vip-go-mu-plugins/blob/develop/admin-notice/conditions/class-capability-condition.php) tied to them, then we definitely wouldn't want that overwritten.

mjangda:
Going to close this as #wontfix per @rebeccahum's note. We do need to be mindful of the capability we use (will likely default to is_super_admin() moving forward) and other more impactful ways to surface this information.

PR #3684, opened 2022-09-23
Expand function parameters according to WP Core

vitaliy-kulish-dt:
Added default WordPress parameters to `wpcom_vip_count_user_posts` function to allow using with custom post types

### DIFF
--- a/vip-helpers/vip-caching.php
+++ b/vip-helpers/vip-caching.php
@@ -361,17 +361,17 @@ function wpcom_vip_old_slug_redirect() {
  * if it's called several times per request. This allows bypassing the db queries in favor of
  * the cache
  */
-function wpcom_vip_count_user_posts( $user_id ) {
+function wpcom_vip_count_user_posts( $user_id, $post_type = 'post', $public_only = false ) {
 	if ( ! is_numeric( $user_id ) ) {
 		return 0;
 	}
 
-	$cache_key   = 'vip_' . (int) $user_id;
+	$cache_key   = 'vip_' .  md5(wp_json_encode($post_type)) . '_' . (int) $user_id;
 	$cache_group = 'user_posts_count';
-	
+
 	$count = wp_cache_get( $cache_key, $cache_group );
 	if ( false === $count ) {
-		$count = count_user_posts( $user_id );
+		$count = count_user_posts( $user_id, $post_type, $public_only );
 
 		wp_cache_set( $cache_key, $count, $cache_group, 5 * MINUTE_IN_SECONDS );
 	}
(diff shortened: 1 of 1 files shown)


vip-helpers/vip-caching.php
```diff
@@ -361,17 +361,17 @@ function wpcom_vip_old_slug_redirect() {
...
  * the cache
  */
-function wpcom_vip_count_user_posts( $user_id ) {
+function wpcom_vip_count_user_posts( $user_id, $post_type = 'post', $public_only = false ) {
 	if ( ! is_numeric( $user_id ) ) {
 		return 0;
 	}
 
-	$cache_key   = 'vip_' . (int) $user_id;
+	$cache_key   = 'vip_' .  md5(wp_json_encode($post_type)) . '_' . (int) $user_id;
```
rebeccahum: Minor styling suggestion:
```suggestion
	$cache_key   = 'vip_' .  md5( wp_json_encode( $post_type ) ) . '_' . (int) $user_id;
```

rebeccahum (changes requested):
Thanks for this!

rebeccahum (dismissed):

vip-helpers/vip-caching.php
```diff
@@ -361,17 +361,17 @@ function wpcom_vip_old_slug_redirect() {
...
  * the cache
  */
-function wpcom_vip_count_user_posts( $user_id ) {
+function wpcom_vip_count_user_posts( $user_id, $post_type = 'post', $public_only = false ) {
 	if ( ! is_numeric( $user_id ) ) {
 		return 0;
 	}
 
-	$cache_key   = 'vip_' . (int) $user_id;
+	$cache_key   = 'vip_' .  md5( wp_json_encode( $post_type ) ) . '_' . (int) $user_id;
```
rinatkhaziev, line 374: @vitaliy-kulish-dt What's the reason for `wp_json_encode` here? The argument is a string and probably can be used as-is?

rinatkhaziev (changes requested):
@vitaliy-kulish-dt thanks for the contribution, I left a question inline.

vip-helpers/vip-caching.php
vitaliy-kulish-dt: @rinatkhaziev param $post_type can be a single post type or an array of post types. To avoid an extra check, I decided to process both options immediately.

rinatkhaziev: Thanks for clarifying! Do you mind adjusting the PHPDoc for the function to list the parameters and their respective possible types?

vitaliy-kulish-dt: @rinatkhaziev yep, sure

rebeccahum (approved):

PR #3695, opened 2022-09-27
VIP Support: Fix PHP warning of undefined $user

rebeccahum:
## Description
Saw in the logs this error:

```
PHP message: PHP Warning:  Undefined variable $user in /var/www/wp-content/mu-plugins/vip-support/class-vip-support-user.php on line 475
```

### Plugin Updated: VIP Support
Fix warning "Undefined variable $user"

### DIFF
--- a/vip-support/class-vip-support-user.php
+++ b/vip-support/class-vip-support-user.php
@@ -460,16 +460,19 @@ class User {
 			return $url;
 		}
 
-		// Get the user's email address.
 		if ( is_numeric( $id_or_email ) ) {
+			// Get the user's email address.
 			$user = get_user_by( 'id', $id_or_email );
 			if ( false !== $user ) {
 				$user_email = $user->user_email;
 			}
 		} elseif ( is_string( $id_or_email ) ) {
+			// Since we already have user email, get WP_User object.
 			$user_email = $id_or_email;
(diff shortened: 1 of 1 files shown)


vip-support/class-vip-support-user.php
```diff
@@ -472,7 +472,7 @@ public function filter_vip_support_email_gravatars( $url, $id_or_email, $args )
 			$user_email = $id_or_email->user_email;
 		}
 
-		if ( isset( $user_email ) && self::is_a8c_email( $user_email ) && ( ! $user || $this->has_vip_support_meta( $user->ID ) ) ) {
+		if ( isset( $user_email ) && self::is_a8c_email( $user_email ) && ( isset( $user->ID ) && $this->has_vip_support_meta( $user->ID ) ) ) {
```
rinatkhaziev, line 478: Was there a bug in the original condition? I'm also not clear on the intent of this check.

rebeccahum: Yep, looks like there was. It checks if the user has an a8c email and also checks if it has the meta attached to it. In an odd case where `$user` isn't defined, we should check if it's set.

rinatkhaziev (approved):

PR #3708, opened 2022-09-27
Search: Cache existence of Custom Search Results and return early in filter__ep_enable_do_weighting for existence of CSR

rebeccahum:
## Description
This PR does two things:
- Caches the existence of Custom Search Results (CSR) in an option
- Bails out of filter__ep_enable_do_weighting under two conditions:
1) if CSR is active and exists since CSR won't work as expected with weighting disabled
2) if CSR is active and the pointer preview display is shown for accuracy since weighting will be enabled

Cache existence of Custom Search Results and do not use filter__ep_enable_do_weighting if Custom Search Results are used

## Checklist

## Steps to Test
1) On a fresh install with no weighting, add a custom search result.
2) Search for the custom search result and notice it doesn't return as expected.
3) Apply PR and repeat step 2.
4) Check if the option is set `vip_custom_results_existence` to `"1"`


rinatkhaziev:
@rebeccahum since we attach on publish and trash hooks we should test that behavior.

rebeccahum:
@rinatkhaziev Added a test case to account for those hooks.

rinatkhaziev (approved):
Thanks for this!

search/includes/classes/class-search.php
```diff
@@ -2401,6 +2417,67 @@ public function filter__ep_enable_do_weighting( $should_do_weighting, $weight_co
...
+	public function is_custom_results_exist() {
+		$args           = [
+			'post_type'           => 'ep-pointer',
+			'post_status'         => 'publish',
+			'fields'              => 'ids',
+			'posts_per_page'      => 1,
+			'no_found_rows'       => true,
+			'ignore_sticky_posts' => true,
+		];
+		$custom_results = new \WP_Query( $args );
```
lukaspawlik, line 2476: @rinatkhaziev @rebeccahum a question - given the issue (https://github.com/Automattic/vip-go-mu-plugins/pull/3771) we observed today, would it make sense to add, above this line, something like 

```php
add_filter( 'ep_skip_query_integration', '__return_true', PHP_INT_MAX, 0 );
```
to ensure this query doesn't trigger somehow another Elasticpress call like we had in mentioned bug? I am not saying #3771 is not a proper solution, but maybe ensuring this `WP_Query` never goes with Elasticpress (what would cause this infinite loop).
After this line `remove_filter` call would be needed
```php
remove_filter( 'ep_skip_query_integration', '__return_true', PHP_INT_MAX );
```

(side note: probably it would make sense to use dedicated class method instead of `__return_true` in case any other piece of code would be doing the same using `__return_true` callback)

PR #3711, opened 2022-09-29
Search: Fix commands get-indexes and get-mapping

rebeccahum:
## Description
This PR does two things:
1) Removes system operation indexes from being returned in `wp vip-search get-indexes`
2) For multisites, it fixes an error that is returned in `wp vip-search get-mapping` when one indexable is enabled on a subsite, but not another

The error returned from 2) is because features on VIP are activated by a site-basis, rather than network-wide. So, the assumption that each site will have the same indexables does not apply at https://github.com/10up/ElasticPress/blob/043ef0dc6ef1865644da5a33bd68d5838f7ee4e3/includes/classes/Command.php#L443-L465

Search: Fix commands get-indexes and get-mapping

## Checklist

## Steps to Test
For 1):
1) Do `wp vip-search get-indexes` and see `.geoip_databases` returned from it
2) Apply PR
3) Repeat step 1) and see no `.geoip_databases`

For 2):
1) Create a multisite with two subsites
2) For subsite 1 enable the `terms` feature via `wp vip-search activate-feature terms`
3) Do a network-wide re-index `wp vip-search index --setup --network-wide`
4) `wp vip-search get-mapping` and see an error returned, as it is attempting to query for the terms index on subsite 2
5) Apply PR and repeat step 4) and expect to see the mapping settings returned properly


search/includes/classes/commands/class-corecommand.php
```diff
@@ -90,6 +91,35 @@ private function set_version( $indexable, $version ) {
...
+		}
+
+		$indexes = array_column( $body, 'index' );
+		// Remove system operation indexes from list.
+		$indexes = array_filter( $indexes, function( $index ) {
+			if ( function_exists( 'str_starts_with' ) ) { // Only available for PHP 8+
+				return ! str_starts_with( $index, '.' );
+			}
+			return substr( $index, 0, 1 ) !== '.';
+		} );
```
sjinks: What do you think if we simplify this piece of code a bit?

```suggestion
		$indexes = array_filter( $indexes, fn( $index ) => ! $index || $index[0] !== '.' );
```

If `$index` cannot be empty, `fn( $index ) => $index[0] !== '.'`

rebeccahum: Sounds good, thank you!

sjinks (approved):

PR #3717, opened 2022-09-30
Update jetpack loader

pschoffer:
## Description
Adds extra validation before we set `VIP_JETPACK_LOADED_VERSION`. That way if Jetpack fails to load we don't end up in broken state but "only" with site without Jetpack.

`VIP_JETPACK_LOADED_VERSION` is then later used to load internal plugins that rely on Jetpack which would cause a fatal error if Jetpack wasn't loaded.

### Plugin Updated: Jetpack loader

We made Jetpack loader more resilient as it now validates existence of `Jetpack` class, before proceeding to load more related logic.

## Checklist

## Steps to Test

1) Add a return statement at the top of jetpack.php you are loading (similar to as if the version requirements weren't met)
2) Navigate to wp-admin
3) Page still works (although no JP is loaded)

### DIFF
--- a/jetpack.php
+++ b/jetpack.php
@@ -191,7 +191,10 @@ function vip_jetpack_load() {
 			}
 
 			require_once $path;
-			define( 'VIP_JETPACK_LOADED_VERSION', $version );
+			if ( class_exists( 'Jetpack' ) ) {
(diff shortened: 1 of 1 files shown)


jesse-greathouse (approved):
👍

PR #3718, opened 2022-09-30
Lockout: Show notices to VIP Support users and updated copy

vaurdan:
## Description

This PR  updates the copy of the wp-admin notices, and changes the display logic of these notices to also show them to VIP Support users.

The validation was made with `is_automattician()` to be consistent with the class implementation. Another alternative would be to check if the user role is VIP Support.

VIP Support users will retain all the capabilities, while regular users will be reduced to Subscriber capabilities.

### Lockout Update

Updated the copy of the admin notices and made them visible to VIP Support users.

## Checklist

## Steps to Test

1. Check out PR.
1. Set `VIP_ACCOUNT_STATUS` environment variable to `warning`. 
2. Now, change it to `locked`. The admin dashboard should be locked, and a red notice should be visible in all pages.

### DIFF
--- a/security/class-lockout.php
+++ b/security/class-lockout.php
@@ -55,12 +55,12 @@ class Lockout {
 		if ( defined( 'VIP_ACCOUNT_STATUS' ) && constant( 'VIP_ACCOUNT_STATUS' ) !== self::ACCOUNT_STATUS_NORMAL ) {
 			switch ( $this->get_lockout_state() ) {
 				case self::ACCOUNT_STATUS_WARNING:
-					return 'Payment for this WordPress VIP account is overdue and access will be disabled.<br />
+					return 'Payment for this WordPress VIP account is overdue and access will be suspended soon.<br />
 Please contact accounts@wpvip.com to settle your bill.';
 				case self::ACCOUNT_STATUS_LOCK:
 				case self::ACCOUNT_STATUS_SHUTDOWN:
-					return 'Payment for this WordPress VIP account is overdue and access has been disabled.<br />
-Please contact accounts@wpvip.com to settle your bill.';
+					return 'Payment for this WordPress VIP account is overdue and access has been suspended.<br />
+Please contact accounts@wpvip.com to settle your bill and restore access.';
 			}
 		}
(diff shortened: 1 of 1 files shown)


security/class-lockout.php
```diff
@@ -55,12 +55,12 @@ private function get_lockout_message() {
 		if ( defined( 'VIP_ACCOUNT_STATUS' ) && constant( 'VIP_ACCOUNT_STATUS' ) !== self::ACCOUNT_STATUS_NORMAL ) {
 			switch ( $this->get_lockout_state() ) {
 				case self::ACCOUNT_STATUS_WARNING:
-					return 'Payment for this WordPress VIP account is overdue and access will be disabled.<br />
+					return 'Payment for this WordPress VIP account is overdue and access will be suspended soon.<br />
 Please contact accounts@wpvip.com to settle your bill.';
```
vaurdan: Should we also make the email here a link?

security/class-lockout.php
```diff
@@ -55,12 +55,12 @@ private function get_lockout_message() {
...
 				case self::ACCOUNT_STATUS_LOCK:
 				case self::ACCOUNT_STATUS_SHUTDOWN:
-					return 'Payment for this WordPress VIP account is overdue and access has been disabled.<br />
-Please contact accounts@wpvip.com to settle your bill.';
+					return 'Payment for this WordPress VIP account is overdue and access has been suspended.<br />
+Please contact accounts@wpvip.com to settle your bill and restore access.';
```
nickdaugherty: Should we make the email address a clickable `mailto` link? I did so in the Dashboard.

nickdaugherty (dismissed):
Looks good to me, just left 1 note.

security/class-lockout.php
nickdaugherty: Oh i didn't see this until just now, I asked the same thing 😂 ... yes I think we should.

vaurdan: And, done :)

nickdaugherty (approved):
LGTM! 🚢

rebeccahum (approved):

PR #3719, opened 2022-09-30
Update two-factor submodule

rebeccahum:
## Description
Bump to latest version 7.2

### Plugin Updated: Two-Factor

We upgraded Two-Factor to 7.2


jesse-greathouse (approved):
👍

PR #3742, opened 2022-10-06
Sync Site Details when the siteurl/home changes

andrea-sdl:
## Description
This PR improves the SDS syncing by adding a sync when the `siteurl/home changes`
This change should support both single and multisites.

### Improved Site Details sync

Small improvements in the site details sync to trigger a sync when the home/siteurl changes.

## Checklist

## Steps to Test

*Single site*

1. Update the home/siteurl of a site `wp option update siteurl 'http://example.com' && wp option update home 'http://example.com'` 
1. Check that the SDS endpoint has been call and that the data has changed.

*Network site*

1. Update the home/siteurl of a site `wp option update siteurl 'http://example.com' --url=<your-old-url> && wp option update home 'http://example.com'  --url=<your-old-url>` 
1. Check that the SDS endpoint has been call and that the data has changed.


config/class-sync.php
```diff
@@ -63,7 +77,43 @@ public function filter_cron_schedules( $schedule ) {
...
+				// Using the cron protects us from data consistency issues
+				if ( ! wp_next_scheduled( self::CRON_EVENT_NAME, [ 'is_faster_cron' => true ] ) ) {
+					wp_schedule_single_event( time() + 1, self::CRON_EVENT_NAME, [ 'is_faster_cron' => true ] );
+				}
+				restore_current_blog();
+			}
+		}
+
+		if ( false !== array_search( $original_blog_id, $this->blogs_to_sync ) ) {
+			if ( function_exists( 'fastcgi_finish_request' ) ) {
```
rinatkhaziev: This block should be moved to the top of the method because everything below is expensive too (especially switch_to_blog/restore_current_blog)

config/class-sync.php
```diff
@@ -12,6 +12,11 @@ class Sync {
 
 	const JETPACK_PRIVACY_SETTINGS_SYNC_STATUS_OPTION_NAME = 'vip_config_jetpack_privacy_settings_synced_value';
 
+	/**
+	 * @var array List of blogIDs that need a sync
+	 */
+	private $blogs_to_sync = [];
```
rinatkhaziev, line 22: If we use an associative array `[ $blog_id => $blog_id,  $blog_2_id => $blog_2_id ]`  we can reduce check's complexity to O(1): 
instead of `array_search` or `is_array` this becomes a simple `isset( $this->blog_ids_to_sync[ get_current_blog_id() ] )`

rinatkhaziev (changes requested):
Thanks, looking mostly good, but left some inline feedback

config/class-sync.php
WPprodigy: This list should never be big, in fact I'm not sure that it will ever be more than 1 - at least through core functionality. It's either the blog you are on, or another blog being updated in network admin. Pretty sure the only way it grows is through custom code that loops through blogs and does an update.

Even in that case though, with 10k entries we're still talking milliseconds. I'd argue if it's worth the complexity & memory difference. But also, neither is large enough to matter either 😆

config/class-sync.php
andrea-sdl: @rinatkhaziev moved up in https://github.com/Automattic/vip-go-mu-plugins/pull/3742/commits/9af15e80e1088b8aaf620612ab6a437937ce82ea 
I also like the idea that we run this _before_ the `switch_to_blog`.

config/class-sync.php
andrea-sdl: @rinatkhaziev I don't have strong opinions on either implementation. Is this a blocker for you?

rinatkhaziev: yikes, missed I this comment, sorry! @andrea-sdl

rinatkhaziev (dismissed):

rinatkhaziev (dismissed):

config/class-sync.php
```diff
@@ -22,17 +32,27 @@ public static function instance() {
...
 
 	public function maybe_setup_cron() {
 		add_filter( 'cron_schedules', [ $this, 'filter_cron_schedules' ] ); // phpcs:ignore WordPress.WP.CronInterval.ChangeDetected
 		add_action( self::CRON_EVENT_NAME, [ $this, 'do_cron' ] );
 
 		// Only register cron event from admin or CLI, to keep it out of frontend request path
-		$is_cli = defined( 'WP_CLI' ) && WP_CLI;
-
-		if ( ! is_admin() && ! $is_cli ) {
+		if ( ! is_admin() && ! Context::is_wp_cli() ) {
```
andrea-sdl, line 57: @WPprodigy do you see any problems moving the `$is_cli` to the `Context::is_wp_cli()`?

config/class-sync.php
```diff
@@ -63,7 +83,65 @@ public function filter_cron_schedules( $schedule ) {
...
-	public function do_cron() {
+	public function get_blogs_to_sync() {
+		return $this->blogs_to_sync;
+	}
+
+	public function queue_sync_for_blog() {
+		// Queue the sync only from admin or CLI, to keep it out of frontend request path
+		if ( ! is_admin() && ! Context::is_wp_cli() ) {
+			return;
+		}
```
andrea-sdl, line 96: @WPprodigy @rinatkhaziev one thing I've also added is to not queue the sync if we're not on admin/WP CLI, like we already do for the `maybe_setup_cron`. 

I didn't include this into the `init_listener` primarily to make it easier to test (I was having troubles defining the test to check that the behavior would actually change if we're not on WP CLI)

config/class-sync.php
```diff
@@ -63,7 +83,65 @@ public function filter_cron_schedules( $schedule ) {
...
+		$needs_sync = false;
+
+		// Check if the current request changed important data on this blog.
+		$blog_had_changes = false !== array_search( $this->original_blog_id, $this->blogs_to_sync );
+		if ( $blog_had_changes ) {
+			$needs_sync = true;
+		}
+
+		if ( $needs_sync && ! $this->is_sync_ratelimited() ) {
```
andrea-sdl, line 117: rate-limiting the sync ( @WPprodigy suggestion )

config/class-sync.php
```diff
@@ -63,7 +83,65 @@ public function filter_cron_schedules( $schedule ) {
...
+	public function queue_sync_for_blog() {
+		// Queue the sync only from admin or CLI, to keep it out of frontend request path
+		if ( ! is_admin() && ! Context::is_wp_cli() ) {
+			return;
+		}
+		$blog_id = get_current_blog_id();
+
+		// To avoid performance issues, don't add if the count would surpass BLOGS_TO_SYNC_LIMIT.
+		// Can rely on the default cron schedules to sync the rest.
+		if ( count( $this->blogs_to_sync ) >= self::BLOGS_TO_SYNC_LIMIT ) {
```
andrea-sdl, line 100: We will also limit the number of blogs we sync via a cron trigger, to limit inadvertently triggering a mass update on SDS that might cause problems.

We will rely on our cronjobs for when there is a mass update.

config/class-sync.php
```diff
@@ -22,17 +32,27 @@ public static function instance() {
 	}
 
 	public function init() {
+		// Saving the initial blog_id in the init to be extra sure nobody changed it later on via a switch_to_blog().
+		$this->original_blog_id = get_current_blog_id();
```
andrea-sdl: @WPprodigy I'm not sure if placing this here is early enough or if we should add it even before the init, to make sure we are getting the "first blog_id". Ideas?

WPprodigy: This should be good, happens on the `init` WP hook right now which is quite early. If we want to be really safe, could make sure it can't be re-set (in case some code triggers the hook again). So something like this perhaps would be good:

```
if ( is_null( $this->original_blog_id ) ) {
	$this->original_blog_id = get_current_blog_id();
}
```

config/class-sync.php
WPprodigy: Nope, I like the change :)

config/class-sync.php
WPprodigy: I like the restriction as it brings about some extra safety.

There's a (small) chance the option is updated via REST/XMLRPC endpoints, which we wouldn't catch here. Maybe that's an okay trade-off though. (Like you said earlier, we can always rely on the normal cron schedule to catch)

Also, `is_admin()` does unfortunately include frontend (possibly unauthenticated) AJAX requests. We do need to support authenticated admin ajax requests though since that is a likely route used for updating options in the admin.

config/class-sync.php
```diff
@@ -63,7 +83,65 @@ public function filter_cron_schedules( $schedule ) {
...
+		if ( count( $this->blogs_to_sync ) == 0 ) {
+			return;
+		}
+		$needs_sync = false;
+
+		// Check if the current request changed important data on this blog.
+		$blog_had_changes = false !== array_search( $this->original_blog_id, $this->blogs_to_sync );
+		if ( $blog_had_changes ) {
+			$needs_sync = true;
+		}
```
WPprodigy:
```suggestion
		if ( 0 === count( $this->blogs_to_sync ) ) {
			return;
		}

		$needs_sync = false !== array_search( $this->original_blog_id, $this->blogs_to_sync );
```

config/class-sync.php
andrea-sdl: Change added in https://github.com/Automattic/vip-go-mu-plugins/pull/3742/commits/d1755d951d6421890ff5fd06801b4ad93a8ac4a6

config/class-sync.php
```diff
@@ -63,7 +85,61 @@ public function filter_cron_schedules( $schedule ) {
...
+				// Avoid syncing in this request as switch_to_blog() is not reliable for gaining the full state (plugins, etc).
+				if ( ! wp_next_scheduled( self::CRON_EVENT_NAME, [ 'is_faster_cron' => true ] ) ) {
+					wp_schedule_single_event( time() + 1, self::CRON_EVENT_NAME, [ 'is_faster_cron' => true ] );
+				}
+
+				restore_current_blog();
+			}
+		}
+		//clearing the sync queue after the run so that we don't sync the same blog twice
+		$this->blogs_to_sync = array();
```
andrea-sdl: @WPprodigy I've also added this to clear the sync queue after we're done with it. If there is a slight possibility that the function might get triggered more than once then I want to be sure it's not sending double the notifications it needs.

config/class-sync.php
WPprodigy: Nope, I like the change!

WPprodigy (dismissed):
Did some testing, looking good!

PR #3772, opened 2022-10-12
Use a different approach to run WP Core Tests

sjinks:
WP 6 changed the way how the tests should be run. This PR modified the workflow to run the tests properly.


sjinks:
The "Run WP Core Tests" jobs should be ignored - they are to be replaced with the "CI (WP Core) / Run WP Core Tests" job.

Once this PR is merged, I will update branch protection rules.

jesse-greathouse (approved):
LGTM 👍

PR #3776, opened 2022-10-12
CANTINA-911: add support for touch()

sjinks:
## Description

This PR adds support for the `touch()` function to our VIP Filesystem Wrapper.

### Plugin Updated: VIP File Service

Added support for the `touch()` function to the `vip://` filesystem wrapper,

## Checklist

### DIFF
--- a/files/class-api-client.php
+++ b/files/class-api-client.php
@@ -4,8 +4,8 @@ namespace Automattic\VIP\Files;
 
 use WP_Error;
 
-require __DIR__ . '/class-curl-streamer.php';
-require __DIR__ . '/class-api-cache.php';
+require_once __DIR__ . '/class-curl-streamer.php';
+require_once __DIR__ . '/class-api-cache.php';
 
 function new_api_client() {
 	return new API_Client(
--- a/files/class-vip-filesystem-stream-wrapper.php
+++ b/files/class-vip-filesystem-stream-wrapper.php
@@ -105,7 +105,7 @@ class VIP_Filesystem_Stream_Wrapper {
 	/**
 	 * Flush empty file flag
 	 *
-	 * Flag to determine if an empty file should be flushed to the 
+	 * Flag to determine if an empty file should be flushed to the
 	 * Filesystem.
 	 *
 	 * @since   1.0.0
@@ -694,9 +694,25 @@ class VIP_Filesystem_Stream_Wrapper {
 	 * @return  bool
 	 */
 	public function stream_metadata( $path, $option, $value ) {
-		$this->debug( sprintf( 'stream_metadata =>  %s + %s + %s', $path, $option, $value ) );
+		$this->debug( sprintf( 'stream_metadata =>  %s + %s + %s', $path, $option, json_encode( $value ) ) );   // phpcs:ignore WordPress.WP.AlternativeFunctions.json_encode_json_encode
 
-		return false;
+		switch ( $option ) {
+			case STREAM_META_TOUCH:
+				if ( false === file_exists( $path ) ) {
+					$file = fopen( $path, 'w' );    // phpcs:ignore WordPress.WP.AlternativeFunctions.file_system_read_fopen
+					if ( is_resource( $file ) ) {
+						fclose( $file );
+						return true;
+					}
+
+					return false;
+				}
+
+				return true;
+
+			default:
+				return false;
+		}
 	}
 
 	/**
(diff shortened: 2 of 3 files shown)


files/class-api-client.php
```diff
@@ -4,8 +4,8 @@
 
 use WP_Error;
 
-require __DIR__ . '/class-curl-streamer.php';
-require __DIR__ . '/class-api-cache.php';
+require_once __DIR__ . '/class-curl-streamer.php';
+require_once __DIR__ . '/class-api-cache.php';
```
sjinks, line 8: Unrelated, but SonarCloud insisted on that change :-(

files/class-vip-filesystem-stream-wrapper.php
```diff
@@ -694,9 +694,23 @@ public function mkdir( $path, $mode, $options ) {
 	 * @return  bool
 	 */
 	public function stream_metadata( $path, $option, $value ) {
-		$this->debug( sprintf( 'stream_metadata =>  %s + %s + %s', $path, $option, $value ) );
+		$this->debug( sprintf( 'stream_metadata =>  %s + %s + %s', $path, $option, json_encode( $value ) ) );   // phpcs:ignore WordPress.WP.AlternativeFunctions.json_encode_json_encode
```
sjinks, line 698: `$value` can be an array; `json_encode()` is used to avoid "array to string conversion" warnings.

files/class-vip-filesystem-stream-wrapper.php
```diff
@@ -694,9 +694,25 @@ public function mkdir( $path, $mode, $options ) {
...
+			case STREAM_META_TOUCH:
+				if ( false === file_exists( $path ) ) {
+					$file = fopen( $path, 'w' );    // phpcs:ignore WordPress.WP.AlternativeFunctions.file_system_read_fopen
+					if ( is_resource( $file ) ) {
+						fclose( $file );
+						return true;
+					}
+
+					return false;
+				}
```
sjinks, line 710: This seems to be the simplest way to implement `touch()`: `file_exists()` will skip downloading the file if it exists; `fopen()` will create an empty file, and then `fclose()` will make sure that the empty file is uploaded to the server.

If we skip `file_exists()`, I suspect the system will download the file if it exists, which might not be desirable for large files.

files/class-vip-filesystem-stream-wrapper.php
```diff
@@ -114,6 +114,8 @@ class VIP_Filesystem_Stream_Wrapper {
 	 */
 	private $should_flush_empty;
 
+	public static ?API_Client $default_client = null;
```
sjinks, line 117: This is to make this class more testable. If we call the FS-related function in the `vip://` inside the class, it will be difficult to override the client without this change.

files/class-vip-filesystem-stream-wrapper.php
```diff
@@ -694,9 +695,25 @@ public function mkdir( $path, $mode, $options ) {
...
-		$this->debug( sprintf( 'stream_metadata =>  %s + %s + %s', $path, $option, $value ) );
+		$this->debug( sprintf( 'stream_metadata =>  %s + %s + %s', $path, $option, json_encode( $value ) ) );   // phpcs:ignore WordPress.WP.AlternativeFunctions.json_encode_json_encode
 
-		return false;
+		switch ( $option ) {
+			case STREAM_META_TOUCH:
+				if ( false === file_exists( $path ) ) {
+					$file = fopen( $path, 'w' );    // phpcs:ignore WordPress.WP.AlternativeFunctions.file_system_read_fopen
+					if ( is_resource( $file ) ) {
+						fclose( $file );
```
rebeccahum, line 705: What if `fclose()` returns `false`?

rebeccahum (changes requested):

files/class-vip-filesystem-stream-wrapper.php
sjinks: Then I will be very sad :-(

But jokes aside, the file is empty - there are no writes, I don't think it is possible for the filesystem to return an `ENOSPC` error.

Anyway, `fopen()` and `fclose()` will be handled by the methods of our FS wrapper; in case of an error, they will show error messages; I don't think we need to duplicate them here.

The `is_resource()` check is only to avoid calling `fclose()` on an invalid file descriptor - to avoid further errors.

rebeccahum (approved):

PR #3782, opened 2022-10-17
Stale GH workflow: Bump time for stale and close

rebeccahum:
## Description
Seeing too many issues that are still applicable being closed and marked stale when we haven't been able to get to them yet. We should bump the time to account for that.


sjinks (approved):
But do we really need this workflow at all?

rebeccahum:
In theory, I think the workflow is a nice reminder to keep us on top of issues...if we are finding that we are re-marking and re-opening things, we might have to remove it in the future.

PR #3799, opened 2022-10-20
refactor: Move all E2E stuff to __tests__/e2e

sjinks:
This PR moves all the stuff related to our E2E tests into `__tests__/e2e` directory.

This will help us keep our root `package.json` clean and reduce execution time for workflows that need to set up Node.js dependencies (for example, they won't have to download 100+ MB of Chromium which is a dependency of Playwright).

The next step will be to move ES E2E stuff to a separate directory.

We also update VIP CLI to 2.20.0; this also tests the `vip dev-env`-related things.

As a bonus, it is now possible to run E2E tests locally by running `npm test` from `__tests__/e2e` directory (you may need to do `npm ci` first to install the dependencies).


jesse-greathouse (approved):
👍

PR #3818, opened 2022-10-27
Suppress codebase manager

pschoffer:
## Description
There is a bug in class-wp-plugins-list-table.php which makes the search parameter unescaped. CodebaseManager makes this bug more prevalent as it loads the problematic page on every admin_init (as opposed to only the plugin list). 

1) CodebaseManager during [init](https://github.com/Automattic/vip-go-mu-plugins/blob/2e00dd5111f6fa26fef54f1a776a34854b930849/codebase-manager/plugins/plugins-manager.php#L18) constructs WP_Plugins_List_Table
2) The [constructor](https://github.com/WordPress/wordpress-develop/blob/b71a7bfcfff6532d6f83b38dbe52bcb24208d37f/src/wp-admin/includes/class-wp-plugins-list-table.php#L57) of WP_Plugins_List_Table does the unencoding as a side effect

This workaround will avoid loading Codebase Manager when search parameter is present avoiding the problem.

### Plugin Updated: Codebase Manager

Workaround related to search unescaping in wp-admin.

## Checklist

## Steps to Test
1) Go to posts list
1) Search for `test&me` 
1) Search term changes to `test`

### DIFF
--- a/codebase-manager/codebase-manager.php
+++ b/codebase-manager/codebase-manager.php
@@ -11,6 +11,7 @@ add_action( 'admin_init', function() {
 		return;
 	}
 
+	// phpcs:ignore WordPress.Security.NonceVerification.Recommended
 	if ( ! isset( $_REQUEST['s'] ) ) {
 		// Avoid loading stuff due to bug in plugin manager list for search requests.
 		$plugins_manager = new PluginsManager();
--- a/codebase-manager/codebase-manager.php
+++ b/codebase-manager/codebase-manager.php
@@ -11,5 +11,9 @@ add_action( 'admin_init', function() {
 		return;
 	}
 
-	( new PluginsManager() )->init();
+	if ( ! isset( $_REQUEST['s'] ) ) {
+		// Avoid loading stuff due to bug in plugin manager list for search requests.
+		$plugins_manager = new PluginsManager();
+		$plugins_manager->init();
+	}
 } );
--- a/codebase-manager/codebase-manager.php
+++ b/codebase-manager/codebase-manager.php
@@ -11,10 +11,6 @@ add_action( 'admin_init', function() {
 		return;
 	}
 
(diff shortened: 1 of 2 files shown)


sjinks (dismissed):

WPprodigy:
Eh, idk. This will result in plugin update notices not showing after searching by plugin name in the plugins ui.

rebeccahum:
What about hooking on later to `current_screen` and using `get_current_screen()` and only loading it for `plugins`?

GaryJones:
> There is a bug in class-wp-plugins-list-table.php which makes the search parameter unescaped

Can you be more specific on this?

rebeccahum:
@GaryJones See BB8-9820 for more details.

GaryJones:
Looks like the [class is only instantiated to get the column count of a table](https://github.com/Automattic/vip-go-mu-plugins/blob/2e00dd5111f6fa26fef54f1a776a34854b930849/codebase-manager/plugins/plugins-manager.php#L18-L19).

And that column count is only used in a [callback](https://github.com/Automattic/vip-go-mu-plugins/blob/2e00dd5111f6fa26fef54f1a776a34854b930849/codebase-manager/plugins/plugins-manager.php#L32-L37) to [the `after_plugin_row` hook](https://github.com/Automattic/vip-go-mu-plugins/blob/2e00dd5111f6fa26fef54f1a776a34854b930849/codebase-manager/plugins/plugins-manager.php#L21), so moving the first two lines into the callback may be sufficient such that the search-handling code that uses `$_REQUEST` has finished.

sjinks:
@GaryJones @WPprodigy could you please take a look?

WPprodigy (approved):

GaryJones (commented):
I've not checked the code still works as intended, but certainly the new changes match up to what I was intending to convey should/could work.

PR #3832, opened 2022-11-01
Query Monitor: Add cron panel

rebeccahum:
## Description
This adds the cron panel to query monitor.
Screenshot 2022-11-01 at 10 43 02 AM

We added a panel to display cron events.

## Checklist

## Steps to Test
1) Go to query monitor > Cron and verify it matches with what is in Debug Bar

### DIFF
--- a/qm-plugins/qm-cron/class-qm-cron-collector.php
+++ b/qm-plugins/qm-cron/class-qm-cron-collector.php
@@ -1,0 +1,168 @@
+<?php
+
+class QM_Cron_Collector extends \QM_Collector {
+
+	/**
+	 * @var string
+	 */
+	public $id = 'qm-cron';
+
+	/**
+	 * Lists all crons that are defined in WP Core.
+	 *
+	 * @var array
+	 *
+	 * @internal To find all, search WP trunk for `wp_schedule_(single_)?event`.
+	 */
+	private $core_cron_hooks = [
+		'do_pings',
+		'importer_scheduled_cleanup',     // WP 3.1+.
+		'publish_future_post',
+		'update_network_counts',          // WP 3.1+.
+		'upgrader_scheduled_cleanup',     // WP 3.3+.
+		'wp_maybe_auto_update',           // WP 3.7+.
+		'wp_scheduled_auto_draft_delete', // WP 3.4+.
+		'wp_scheduled_delete',            // WP 2.9+.
+		'wp_split_shared_term_batch',     // WP 4.3+.
+		'wp_update_plugins',
+		'wp_update_themes',
+		'wp_version_check',
+	];
+
+	/**
+	 * @return string
+	 */
+	public function name() {
+		return esc_html__( 'Cron', 'query-monitor' );
+	}
+
+	/**
+	 * @return void
+	 */
+	public function process() {
+		$this->process_crons();
+		$this->process_schedules();
+	}
+
+	/**
+	 * Processes the cron jobs.
+	 *
+	 * @return void
+	 */
+	private function process_crons() {
+		$this->data['doing_cron'] = get_transient( 'doing_cron' ) ? true : false;
+
+		$crons = _get_cron_array();
+		if ( is_array( $crons ) && ! empty( $crons ) ) {
+			$this->data['crons'] = $crons;
+			$this->sort_count_crons( $crons );
+			$this->process_next_event_time( $crons );
+		}
+	}
+
+	/**
+	 * Sort and count crons.
+	 *
+	 * This function sorts the cron jobs into core crons, and custom crons. It also tallies
+	 * a total count for the crons as this number is otherwise tough to get.
+	 *
+	 * @param array $crons Cron events
+	 */
+	private function sort_count_crons( $crons ) {
+		$total_crons      = 0;
+		$total_core_crons = 0;
+		$total_user_crons = 0;
+		$core_crons       = [];
+		$user_crons       = [];
+		foreach ( $crons as $time => $time_cron_array ) {
+			foreach ( $time_cron_array as $hook => $data ) {
+				$total_crons += count( $data );
+
+				if ( in_array( $hook, $this->core_cron_hooks, true ) ) {
+					$core_crons[ $time ][ $hook ] = $data;
+					$total_core_crons            += count( $data );
+				} else {
+					$user_crons[ $time ][ $hook ] = $data;
+					$total_user_crons            += count( $data );
+				}
+			}
+		}
+
+		$this->data['total_crons']      = $total_crons;
+		$this->data['total_core_crons'] = $total_core_crons;
+		$this->data['core_crons']       = $core_crons;
+		$this->data['user_crons']       = $user_crons;
+		$this->data['total_user_crons'] = $total_user_crons;
+	}
+
+	/**
+	 * Process next cron event time
+	 *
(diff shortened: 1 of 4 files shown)


WPprodigy:
At some point I'd like to make this panel play nicer with Cron Control. But merging as-is will be fine so that we can deprecate debug bar faster.

qm-plugins/qm-cron/class-qm-cron-collector.php
```diff
@@ -0,0 +1,168 @@
...
+	 *
+	 * @return int Return 1 if $schedule_a argument 'interval' greater then $schedule_b argument 'interval',
+	 *             0 if both intervals equivalent and -1 otherwise.
+	 */
+	private function schedules_sorting( $schedule_a, $schedule_b ) {
+		if ( (int) $schedule_a['interval'] === (int) $schedule_b['interval'] ) {
+			return 0;
+		} else {
+			return ( ( (int) $schedule_a['interval'] > (int) $schedule_b['interval'] ) ? 1 : -1 );
+		}
```
sjinks: What do you think about this:

```suggestion
		return (int) $schedule_a['interval'] <=> (int) $schedule_b['interval'];
```

Ref: [spaceship operator](https://www.php.net/manual/en/migration70.new-features.php#migration70.new-features.spaceship-op)

qm-plugins/qm-cron/class-qm-cron-collector.php
```diff
@@ -0,0 +1,168 @@
+<?php
+
+class QM_Cron_Collector extends \QM_Collector {
```
sjinks: Here and below: does it make sense to use a fully qualified class name? As far as I can see, we don't have a namespace here.

```suggestion
class QM_Cron_Collector extends QM_Collector {
```

qm-plugins/qm-cron/class-qm-cron-output-html.php
```diff
@@ -0,0 +1,333 @@
...
+						echo '<br>';
+						echo '<i>' . esc_html( $this->display_past_time( human_time_diff( $data['next_event_time']['unix'] ), $data['next_event_time']['unix'] ) ) . '</i>';
+						?>
+					</td>
+					<td><?php echo esc_html( gmdate( 'H:i:s' ) ); ?></td>
+				</tr>
+			</tbody>
+		</table>
+		<br>
+		<h3><b>Schedules</b></h3>
```
sjinks: Isn't `<h3>` already bold (I didn't check, just asking :-) )?

rebeccahum: It makes it extra bold then D:

sjinks: See also: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/b

query-monitor.php
```diff
@@ -160,3 +160,4 @@ function change_dispatchers_shutdown_priority( array $dispatchers ) {
 require_once __DIR__ . '/qm-plugins/qm-alloptions/qm-alloptions.php';
 require_once __DIR__ . '/qm-plugins/qm-object-cache/qm-object-cache.php';
 require_once __DIR__ . '/qm-plugins/qm-apcu-cache/qm-apcu-cache.php';
+require_once __DIR__ . '/qm-plugins/qm-cron/qm-cron.php';
```
sjinks: Do we need to wrap this with `file_exists()` check to make sure that the very first deploy won't cause a spike of fatal errors?

qm-plugins/qm-cron/class-qm-cron-output-html.php
```diff
@@ -0,0 +1,333 @@
...
+	/**
+	 * Transform a time in seconds to hours rounded to 2 decimals.
+	 *
+	 * @param int $time Unix timestamp.
+	 *
+	 * @return float
+	 */
+	private function get_hours( $time ) {
+		return round( ( (int) $time / 3600 ), 2 );
+	}
```
sjinks, line 324: I think I have already seen exactly the same functions in `class-qm-cron-collector.php`.

Do you think it makes sense to create a helper class and move these methods into it to avoid code duplication?

qm-plugins/qm-cron/class-qm-cron-output-html.php
```diff
@@ -0,0 +1,333 @@
...
+	/**
+	* Displays the cron arguments in a readable format.
+	*
+	* @param mixed $args Cron argument(s).
+	*
+	* @return void
+	*/
+	private function display_event_cron_arguments( $args ) {
+		// Arguments defaults to an empty array if no arguments are given.
+		if ( is_array( $args ) && array() === $args ) {
```
sjinks:
```suggestion
		if ( is_array( $args ) && empty( $args ) ) {
```

?

(so that our code looks the same [everywhere](https://github.com/Automattic/vip-go-mu-plugins/pull/3832/files#diff-7f540689ad3ddbab9b7839e5f2bdc5e2ac83869f7c7aecbb64fc94e17b6120eaR56))

qm-plugins/qm-cron/class-qm-cron-output-html.php
rebeccahum: Ah, whoops. I actually intended to move it and accidentally left the one in the other file. I've removed the duplicated one.

query-monitor.php
rebeccahum: Wouldn't hurt!

qm-plugins/qm-cron/class-qm-cron-output-html.php
```diff
@@ -0,0 +1,333 @@
...
+	*
+	* @param int $time        Timestamp.
+	* @param int $event_count Number of events running at this time.
+	*/
+	private function display_event_time( $time, $event_count ) {
+		$row_span = ( $event_count > 1 ) ? ' rowspan="' . esc_attr( $event_count ) . '"' : '';
+
+		// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+		echo '<td' . $row_span . '>
+		   ', esc_html( gmdate( 'Y-m-d H:i:s', $time ) ), '<br />
```
sjinks, line 215: We use self-closing `br` tags here; does it make sense to use them everywhere for consistency? For example, we use `<br>`, not `<br />` in `output()`.

qm-plugins/qm-cron/class-qm-cron-output-html.php
```diff
@@ -0,0 +1,333 @@
...
+
+			foreach ( $time_cron_array as $hook => $data ) {
+				$row_attributes = $this->get_event_row_attributes( $time, $hook );
+				$arg_set_count  = count( $data );
+				$show_hook      = true;
+
+				foreach ( $data as $hash => $info ) {
+					echo '<tr', $row_attributes, '>'; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+
+					if ( true === $show_time ) {
```
sjinks:
```suggestion
					if ( $show_time ) {
```

`$show_time` and `$show_hook` can only be `true` or `false`; we can probably remove the comparison.

sjinks (dismissed):
LGTM

sjinks (dismissed):

sjinks (approved):

PR #3758, opened 2022-10-10
(WIP) Block automatic loading of wp-parsely and Jetpack for FedRAMP sites.

jblz:
## Description
WIP

## Checklist

Please make sure the items below have been covered before requesting a review:

- [ ] This change works and has been tested locally (or has an appropriate fallback).
- [ ] This change works and has been tested on a Go sandbox.
- [ ] This change has relevant unit tests (if applicable).
- [ ] This change has relevant documentation additions / updates (if applicable).
- [ ] I've created a changelog description that aligns with the provided examples.

PR #3662, opened 2022-09-15
parsely.yml: Add wp-parsely 3.5 tests

acicovic:
## Description
This PR adds wp-parsely 3.5 tests in `.github/workflows/parsely.yml`.

### Added wp-parsely 3.5 tests in our parsely.yml GitHub Workflow

We added wp-parsely 3.5 tests in our parsely.yml GitHub Workflow.

## Checklist

Please make sure the items below have been covered before requesting a review:

- [ ] This change works and has been tested locally (or has an appropriate fallback).
- [ ] This change works and has been tested on a Go sandbox.
- [ ] This change has relevant unit tests (if applicable).
- [ ] This change has relevant documentation additions / updates (if applicable).
- [x] I've created a changelog description that aligns with the provided examples.

## Steps to Test
The newly added tests should be executed by GitHub.

PR #3845, opened 2022-11-03
Fix wp-parsely integration tests

jblz:
## Description

While working on limiting the way the plugin defaults to load on certain sites (#3758), I noticed our Integration Tests were not correct in a number of ways. The most glaring was that the "integration type" was showing as `INTEGRATION_TYPE_SELF_MANAGED` for the cases where it's loaded via filter and option.

It's not possible to accurately test for the integration type and other "initialization" related parameters in the current setup as the plugin gets automatically loaded prior to the test suite running.

This change works around these issues by:
* unhooking the early hooks that load the plugin in the bootstrap file
* running all tests in isolated processes that do not share state
* manually calling the aforementioned loading functions in ways that are expected by the functionality under test

Additionally, this change:
* Allows for specifying the targeted plugin version on the command line
* Allows for specifying the targeted loading mode on the command line
* Improves validation of the loaded plugin version
* Fixes tests for plugin version 3.5

### Plugin Wrapper Testing Improved: wp-vip

We increased coverage of our integration with various versions of the wp-parsely plugin.

## Checklist

## Steps to Test

1. Check out PR.
1. Check all changes to the test suite and bootstrap file for correctness
1. Test all targeted plugin versions and integration modes:
    ````
    for ver in 3.3 3.5; do for mode in disabled option_enabled filter_enabled filter_and_option_enabled; do WPVIP_PARSELY_INTEGRATION_PLUGIN_VERSION="$ver" WPVIP_PARSELY_INTEGRATION_TEST_MODE="$mode" ./bin/test.sh --filter MU_Parsely_Integration_Test; done; done
    ````

### DIFF
--- a/bin/test.sh
+++ b/bin/test.sh
@@ -128,6 +128,8 @@ docker run \
     -e MYSQL_HOST \
     -e DISABLE_XDEBUG=1 \
     -e PRETEST_SCRIPT=/home/circleci/project/bin/pretest.sh \
+    -e WPVIP_PARSELY_INTEGRATION_TEST_MODE="${WPVIP_PARSELY_INTEGRATION_TEST_MODE}" \
+    -e WPVIP_PARSELY_INTEGRATION_PLUGIN_VERSION="${WPVIP_PARSELY_INTEGRATION_PLUGIN_VERSION}" \
     ${DOCKER_OPTIONS} \
     -v "$(pwd):/home/circleci/project" \
     ghcr.io/automattic/vip-container-images/wp-test-runner:latest \
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -153,6 +153,16 @@ switch ( getenv( 'WPVIP_PARSELY_INTEGRATION_TEST_MODE' ) ) {
 		break;
 }
 
+tests_add_filter( 'muplugins_loaded', function () {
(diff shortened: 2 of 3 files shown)


jblz:
@acicovic Can you please take a look at this when you get a sec?

We should probably also go ahead and add 3.5 to the test matrix:
https://github.com/Automattic/vip-go-mu-plugins/blob/f021daaa96f3293ebcb74f53d0bb110a00375872/.github/workflows/parsely.yml#L23-L31

...but these are pretty heavy, so it would probably be prudent to pull out 3.1 at the same time.  I'll leave that decision to y'all.

jblz:
> We should probably also go ahead and add 3.5 to the test matrix

Thinking a bit more about this, I think we can probably handle this at the test level now that we're running in isolated processes and speed things up considerably :thinking:

acicovic:
Hi @jblz, thanks for the awesome work.

To my knowledge wp-parsely 3.1 is still actively being used, but versions 3.2 and 3.3 should be removed soon, so IMO you can remove them from the matrix. Also feel free to add 3.5. I had done this [here](https://github.com/Automattic/vip-go-mu-plugins/pull/3662) but tests were failing and I never got back to it. So if you do this, we can close that PR as well. 😃

acicovic (approved):
Looks good to me 👍

rebeccahum (approved):

PR #3853, opened 2022-11-08
fix(files): `touch()` should fail if `fclose()` fails

sjinks:
## Description

The `touch()` implementation introduced in #3776 assumed that `close()` could not fail. As always, assumptions and the truth dine at totally separate tables.

### Plugin Updated: VIP Filesystem Services

Fix `touch()` implementation to return `false` if `fclose()` fails.

## Checklist

## Steps to Test

`touch("vip://wp-content/uploads/test.php");` in the WP Shell should return `false` if `stream_flush` fails.

### DIFF
--- a/files/class-vip-filesystem-stream-wrapper.php
+++ b/files/class-vip-filesystem-stream-wrapper.php
@@ -702,8 +702,7 @@ class VIP_Filesystem_Stream_Wrapper {
 				if ( false === file_exists( $path ) ) {
 					$file = fopen( $path, 'w' );    // phpcs:ignore WordPress.WP.AlternativeFunctions.file_system_read_fopen
 					if ( is_resource( $file ) ) {
-						fclose( $file );
-						return true;
+						return fclose( $file );
 					}
 
 					return false;
--- a/files/class-vip-filesystem-stream-wrapper.php
+++ b/files/class-vip-filesystem-stream-wrapper.php
@@ -239,12 +239,14 @@ class VIP_Filesystem_Stream_Wrapper {
 	public function stream_close() {
 		$this->debug( sprintf( 'stream_close => %s + %s', $this->path, $this->uri ) );
 
+		$result = true;
+
 		// Don't attempt to flush new file when in read mode
 		if ( $this->should_flush_empty && 'r' !== $this->mode ) {
-			$this->stream_flush();
+			$result = $this->stream_flush();
 		}
 
-		return $this->close_handler( $this->file );
+		return $this->close_handler( $this->file ) && $result;
 	}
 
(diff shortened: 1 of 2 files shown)


files/class-vip-filesystem-stream-wrapper.php
```diff
@@ -239,12 +239,14 @@ public function stream_open( $path, $mode, $options, &$opened_path ) {
...
+		$result = true;
+
 		// Don't attempt to flush new file when in read mode
 		if ( $this->should_flush_empty && 'r' !== $this->mode ) {
-			$this->stream_flush();
+			$result = $this->stream_flush();
 		}
 
-		return $this->close_handler( $this->file );
+		return $this->close_handler( $this->file ) && $result;
```
sjinks, line 249: The order matters - if we check `$result` first and it is `false`, PHP won't run `close_handler()`.

files/class-vip-filesystem-stream-wrapper.php
```diff
@@ -314,16 +316,14 @@ public function stream_flush() {
 
 		try {
 			// Upload to file service
-			$result = $this->client
-				->upload_file( $this->uri, $this->path );
+			$result                   = $this->client->upload_file( $this->uri, $this->path );
+			$this->should_flush_empty = false;
```
sjinks, line 320: This avoids the "double flush": if we run `flush()` before we close the file, we don't want to upload the file twice (see [this line](https://github.com/Automattic/vip-go-mu-plugins/pull/3853/files#diff-9ff26734fc2ae14c41038dfa03c7c03f156e31d68a8ffac8d379f808ad9f20c5L243)).

files/class-vip-filesystem-stream-wrapper.php
```diff
@@ -702,8 +702,8 @@ public function stream_metadata( $path, $option, $value ) {
 				if ( false === file_exists( $path ) ) {
 					$file = fopen( $path, 'w' );    // phpcs:ignore WordPress.WP.AlternativeFunctions.file_system_read_fopen
 					if ( is_resource( $file ) ) {
-						fclose( $file );
-						return true;
+						$result = fflush( $file );
+						return fclose( $file ) && $result;
```
sjinks, line 706: Now the most interesting part: PHP [does not report errors from `fclose()`](https://heap.space/xref/PHP-8.0/ext/standard/file.c?r=7e947908#910). Even if `stream_close()` fails, `fclose()` will return `true`.

This is why we have to call `flush()` to make sure we catch the error if the file upload fails, and then return the proper result.

rebeccahum (approved):

PR #3865, opened 2022-11-08
fix(files): Clear stats cache when upload of an empty file fails

sjinks:
## Description

If we create an empty file and then close it, and for some reason the upload fails, subsequent calls to the `stats` family pretend that the file exists:

```
wp> touch("vip://wp-content/uploads/test.php");
[08-Nov-2022 22:01:01 UTC] Warning: stream_flush failed for wp-content/uploads/test.php with error: Failed to upload file `/tmp/phpdimqHB` to `wp-content/uploads/test.php` (response code: 404) #vip-go-streams in /chroot/var/www/wp-content/mu-plugins/files/class-vip-filesystem-stream-wrapper.php on line 323
bool(false)
wp> is_file("vip://wp-content/uploads/test.php");
bool(true)
```

This happens because `stream_open()` calls `cache_file_stats()` on a temporary file, but `stream_close()` does not clear it if the upload fails (and file is not created).

This PR fixes that.

### Plugin Updated: VIP Filesystem Services

Fix the case when the upload of an empty file fails, but the cached stats are not cleared.

## Checklist

## Steps to Test

Follow the steps in the description; it is very hard to write meaningful tests because `call_api()` method is private.

### DIFF
--- a/files/class-vip-filesystem-stream-wrapper.php
+++ b/files/class-vip-filesystem-stream-wrapper.php
@@ -316,17 +316,22 @@ class VIP_Filesystem_Stream_Wrapper {
 
 		try {
 			// Upload to file service
-			$result                   = $this->client->upload_file( $this->uri, $this->path );
-			$this->should_flush_empty = false;
+			$result = $this->client->upload_file( $this->uri, $this->path );
 			if ( is_wp_error( $result ) ) {
 				trigger_error(
 					sprintf( 'stream_flush failed for %s with error: %s #vip-go-streams', esc_html( $this->path ), esc_html( $result->get_error_message() ) ),
 					E_USER_WARNING
 				);
 
+				if ( $this->should_flush_empty ) {
+					$this->should_flush_empty = false;
+					$this->client->cache_file_stats( $this->path, [] );
+				}
+
 				return false;
 			}
 
+			$this->should_flush_empty = false;
 			return fflush( $this->file );
 		} catch ( \Exception $e ) {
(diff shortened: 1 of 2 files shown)


files/class-vip-filesystem-stream-wrapper.php
```diff
@@ -316,17 +316,22 @@ public function stream_flush() {
...
+			$result = $this->client->upload_file( $this->uri, $this->path );
 			if ( is_wp_error( $result ) ) {
 				trigger_error(
 					sprintf( 'stream_flush failed for %s with error: %s #vip-go-streams', esc_html( $this->path ), esc_html( $result->get_error_message() ) ),
 					E_USER_WARNING
 				);
 
+				if ( $this->should_flush_empty ) {
+					$this->should_flush_empty = false;
+					$this->client->cache_file_stats( $this->path, [] );
```
sjinks, line 336: There is no `clear_file_stats()` exposed by the API client, but passing an empty array effectively discards the cache.

rebeccahum: Can we add a note in the code for why we are passing an empty array?

rebeccahum (changes requested):

files/class-vip-filesystem-stream-wrapper.php
sjinks: Done in aa75268

rebeccahum (approved):

PR #3872, opened 2022-11-10
ci: Update E2E workflow

sjinks:
* Add Foresight domains to the list of the allowed endpoints ([ref](https://github.com/Automattic/vip-go-mu-plugins/actions/runs/3433244834))
* Reduce the number of workers to 1 for CI to reduce CPU/memory usage
* Bump @automattic/vip to 2.22.0

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -108,7 +108,7 @@ jobs:
         working-directory: __tests__/e2e
 
       - name: Install VIP CLI
-        run: npm install -g @automattic/vip@2.20.0
+        run: npm install -g @automattic/vip@2.22.0
 
       - name: Determine WP version
         run: |
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -75,8 +75,11 @@ jobs:
             packages.microsoft.com:443
             download.cypress.io:443
             cdn.cypress.io:443
-            foresight.service.thundra.io
-            foresight-cli-test-artifact-prod.s3-accelerate.amazonaws.com
+            foresight.service.thundra.io:443
+            foresight-cli-test-artifact-prod.s3-accelerate.amazonaws.com:443
+            upload.service.runforesight.com:443
+            api.service.runforesight.com:443
+            api.wpvip.com:443
 
       - name: Check out repository code
         uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


iamchughmayank:
@sjinks are we sure that the flakiness is being caused due to chromium crashing? reducing the worker to 1 may increase test execution time by a minute or so.

sjinks:
@iamchughmayank comparing with [this](https://github.com/Automattic/vip-go-mu-plugins/actions/runs/3433957257/jobs/5724817555), the execution time increased only by 6 seconds (4:53 vs. 4:47 for the Run e2e tests step). This difference makes me think that if parallelism really helped, we would see much bigger difference.

One instance of Chrome should use fewer resources, and ideally, this would result in more stable performance. If it does not work out, we can always revert this change :-)

iamchughmayank:
> @iamchughmayank comparing with [this](https://github.com/Automattic/vip-go-mu-plugins/actions/runs/3433957257/jobs/5724817555), the execution time increased only by 6 seconds (4:53 vs. 4:47 for the Run e2e tests step). This difference makes me think that if parallelism really helped, we would see much bigger difference.
> 
> One instance of Chrome should use fewer resources, and ideally, this would result in more stable performance. If it does not work out, we can always revert this change :-)

I think the compared it with [this one](https://github.com/Automattic/vip-go-mu-plugins/actions/runs/3433856817/jobs/5724612188) where it took around a minute less. But, by looking at more runs it appears that the tests do tend to take more than 4.5 minutes with or without the additional worker.

I agree that it is worth trying 👍

.github/workflows/e2e.yml
```diff
@@ -75,8 +75,11 @@ jobs:
             packages.microsoft.com:443
             download.cypress.io:443
             cdn.cypress.io:443
-            foresight.service.thundra.io
-            foresight-cli-test-artifact-prod.s3-accelerate.amazonaws.com
+            foresight.service.thundra.io:443
+            foresight-cli-test-artifact-prod.s3-accelerate.amazonaws.com:443
```
iamchughmayank, line 79: Thanks for pointing this out. I will keep this in mind in future to specify the port number :)

iamchughmayank (dismissed):
LGTM :)

jesse-greathouse (approved):
LGTM

iamchughmayank (approved):
🚢

PR #3880, opened 2022-11-11
refine alloptions alerting thresholds

trepmal:
## Description

- do no monitor on convenience domains
- no action if uncompressed size is less than 4MB
- if greater than 4M: calculate the compressed size
- if *compressed* size is over 850K: log warning and send P2 alert
- if *compressed* size is over 1M: throw 503, log warning, and send P1 alert

## Changelog Description

Refined alloptions alerting thresholds

## Checklist

## Steps to Test

Create large options for testing

```
# 3 copies of this option will put a basic site at alerting level without throwing 503
update_option( 'large_option_3', wp_remote_get( 'https://en.wikipedia.org/wiki/Language' ) );
```

### DIFF
--- a/alloptions-limit.php
+++ b/alloptions-limit.php
@@ -26,6 +26,14 @@ function run_alloptions_safeguard() {
 		return;
 	}
 
+	$home_url_parsed = wp_parse_url( home_url() );
+	if (
+		wp_endswith( $home_url_parsed['host'], '.go-vip.co' ) ||
+		wp_endswith( $home_url_parsed['host'], '.go-vip.net' )
+	) {
+		return;
+	}
+
 	// To avoid performing a potentially expensive calculation of the compressed size we use 4MB uncompressed (which is likely less than 1MB compressed)
 	$alloptions_size_worry_level = MB_IN_BYTES * 4;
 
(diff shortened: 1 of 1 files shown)


alloptions-limit.php
```diff
@@ -25,12 +26,16 @@ function run_alloptions_safeguard() {
 		return;
 	}
 
-	// Uncompressed size thresholds.
-	// Warn should *always* be =< die
-	$alloptions_size_warn = MB_IN_BYTES * 2.5;
+	$home_url_parsed = wp_parse_url( home_url() );
```
rebeccahum, line 29: Can we also check that it's a production environment before alerting? We can use the `VIP_GO_APP_ENVIRONMENT` constant

trepmal: https://github.com/Automattic/vip-go-mu-plugins/blob/16236fd34061ebbc6501e2c7fb59d40e33c37ef6/alloptions-limit.php#L134


it probably makes sense to shuffle that around though

WPprodigy:
Hey @trepmal, given https://github.com/Automattic/vip-go-mu-plugins/pull/4725 & https://github.com/Automattic/vip-go-mu-plugins/pull/4725/files I think we can actually start to remove the alerting here.

A site shouldn't completely fall over anymore if > 1mb alloptions. What do you think? Maybe we could leave the severe warning/alert in place but drop all the rest? If after a few rounds of checking on those alerts and turns out the sites were doing fine, we can drop it all and no longer alert?

trepmal:
Given the current state, the only alert I'd keep is the "critical" level, as long as we're throwing a 503.
Alerts at the warning level can be kept a bit longer if desired just to be safe. Anecdotally, they're rarely helpful due to the rough math we use. This PR tries to make adjustments there, but it may no longer be necessary with the other work that's been done.
P5/info alerts can certainly be removed (as would have been done by this PR).

WPprodigy:
> as long as we're throwing a 503.

I think we stop doing that directly here. If the site is struggling enough, it'll start 503ing anyway. But with the new mitigations, I really don't think an extra large alloptions will bring down a site.

rinatkhaziev:
Closing for the time being

PR #3881, opened 2022-11-11
Estimate the thumbnail image sizes

WPprodigy:
## Description
In WP 6.0, the `filesize` for attachment thumbnails was added to the attachment metadata. We added support for this here: https://github.com/Automattic/vip-go-mu-plugins/pull/2991

Since we don't actually generate attachments, the filesize was just set to that of the original image. This has some unfortunate implications wherein a sized-down image will be seen as too large for certain tasks (like social media usage, [seen here in yoast seo](https://github.com/Yoast/wordpress-seo/blob/265ff5a8c065b85aea28b0b8305d29a948fb4b25/inc/class-wpseo-image-utils.php#L146-L172)).

As a workaround to this, it was proposed that we can instead just estimate what the resized image will be. So this is what I attempted in this PR.

## Problems

It's not very consistent >.<.

### JPG Image

```
// Original size: 2192x2921
curl -sIXGET https://{redacted}/wp-content/uploads/2022/11/sample.jpg | grep -w content-length
content-length: 1033414

// Resized to 225x300
curl -sIXGET https://{redacted}/wp-content/uploads/2022/11/sample.jpg\?resize\=225,300 | grep -w content-length
content-length: 14703

// Resized to 768x1024
curl -sIXGET https://{redacted}/wp-content/uploads/2022/11/sample.jpg\?resize\=768,1024 | grep -w content-length
content-length: 114954
```

The estimates by the current logic in this PR (pretty close):

```
wp_get_attachment_metadata(8061);

'file' => string(23) "sample.jpg?resize=225,300"
'filesize' => int(10894)

'file' => string(23) "sample.jpg?resize=768,1024"
'filesize' => int(126930)
```

### PNG Image

```
// Original size: 792x725
curl -sIXGET https://{redacted}/wp-content/uploads/2022/11/test.png | grep -w content-length
content-length: 71061

// Resized to 300x275
curl -sIXGET https://{redacted}/wp-content/uploads/2022/11/test.png\?resize\=300,275 | grep -w content-length
content-length: 57071

// Resized to 768x703
curl -sIXGET https://{redacted}/wp-content/uploads/2022/11/test.png\?resize\=768,703 | grep -w content-length
content-length: 388501
```

The estimates by the current logic in this PR (off by half):

```
wp_get_attachment_metadata(8060);

'file' => string(23) "test.png?resize=300,275"
'filesize' => int(102099)

'file' => string(23) "test.png?resize=768,703"
'filesize' => int(668167)
```

So as you can see, we could divide our current estimate by 2 and be pretty close for the PNG. But JPG would then be further off. This doesn't really account for webp conversions either (which I think is fine since those change based on browser support, though maybe we need to account for it if a webp image is uploaded as the original).

So what should we do? Alter the formula based on image type? Am I missing something?


GaryJones:
Presuming it's a one-off task on upload, could you actually do the resize, capture the bytesize, and then dump the thumbnail image?

WPprodigy:
> Presuming it's a one-off task on upload, could you actually do the resize, capture the bytesize, and then dump the thumbnail image?

We generate attachment metadata "on the fly". One big benefit to this is that thumbnails don't need to be "regenerated" when new sizes are added by plugins/themes.

Actually resizing and capturing the bitesize for every attachment thumbnail size on every image could take a very large amount of time on many of our sites 🙃

PR #3888, opened 2022-11-11
ci: Fix Jetpack bootstrapping issue

sjinks:
Because Jetpacks now live elsewhere, several workflows need to be updated to check them out from the other repository.

This fixes the "Jetpack could not be loaded and initialized due to a bootstrapping issue" warning.

Ref: https://github.com/Automattic/vip-go-mu-plugins/actions/runs/3447629250/jobs/5755529162#step:12:956

### DIFF
--- a/.github/actions/prepare-source/action.yml
+++ b/.github/actions/prepare-source/action.yml
@@ -1,0 +1,19 @@
+name: Prpeare source code for testing
+description: Checks out and installs all dependencies
+runs:
+  using: composite
+  steps:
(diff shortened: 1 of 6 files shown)


.github/actions/prepare-source/action.yml
```diff
@@ -0,0 +1,19 @@
+name: Prpeare source code for testing
```
rinatkhaziev:
```suggestion
name: Prepare source code for testing
```

rinatkhaziev (approved):

PR #3899, opened 2022-11-14
Search: Auto re-build post index when mapping is bad

rebeccahum:
## Description
This PR only executes the jobs for non-prods atm. When `wp vip-search index` is run on a new site without the `--setup` flag, we get the error message `field [post_date_gmt] is of type [org.elasticsearch.index.mapper.TextFieldMapper$TextFieldType@f663aff], but only numeric types are supported.` and thus, ES isn't run on it.

Re-build index if incorrect post mapping found (non-prods only)

## Checklist

## Steps to Test
1) Create a new second site
2) Run `wp vip-search index`
3) Go do a search `?s=test` and check Search Dev Tools for the error being returned `field [post_date_gmt] is of type [org.elasticsearch.index.mapper.TextFieldMapper$TextFieldType@f663aff], but only numeric types are supported.`
4) Shell in:
```
$es->init();
$health = new \Automattic\VIP\Search\Health( $es );
$mapping = $health->validate_post_index_mapping( $indexable );
```
Ensure `$mapping` returns `false` to ensure that it is picking up the incorrect mapping
5) Run `wp vip-search index --setup`
6) Repeat step 4 and ensure it returns `true` instead

### DIFF
--- a/.github/workflows/build-docker-image.yml
+++ b/.github/workflows/build-docker-image.yml
@@ -6,14 +6,19 @@ on:
       - staging
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
   build:
     name: Build Docker image
     runs-on: ubuntu-latest
+    permissions: write-all
     steps:
       - name: Request to build Docker image
         uses: actions/github-script@v6
         with:
+          github-token: ${{ secrets.WPCOM_VIP_BOT_TOKEN }}
           script: |
             github.rest.repos.createDispatchEvent({
               owner: 'automattic',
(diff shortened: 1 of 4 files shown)


search/includes/classes/class-health.php
```diff
@@ -2,8 +2,6 @@
 
 namespace Automattic\VIP\Search;
 
-use \ElasticPress\Indexables as Indexables;
```
rebeccahum, line 5: This doesn't appear to be used anywhere.

.github/workflows/build-docker-image.yml
```diff
@@ -13,6 +13,7 @@ jobs:
   build:
     name: Build Docker image
     runs-on: ubuntu-latest
+    permissions: write-all
```
sjinks: ?

rebeccahum: No idea...snuck in somehow >.<

sjinks (approved):

PR #3908, opened 2022-11-16
Introduce Environment::has_var()

gmazzap:
# Description

## The issue

`Environment::get_var()`, as described in [docs](https://docs.wpvip.com/how-tos/manage-environment-variables/#wordpress), allows us to retrieve the value of a custom environment variable set via VIP CLI.

We have found this particularly useful for feature flags. For example, a code like this:

```php
if (Environment::get_var('FEATURE_X_ENABLED')) {
  // Use feature X
}
```

The problem is that `Environment::get_var()` [triggers a notice](https://github.com/Automattic/vip-go-mu-plugins/blob/69e63d26f61fb4b5d152128dd796aacdce06acda/lib/environment/class-environment.php#L29-L36) whenever the target variable is not set.

This is annoying because, for feature flags, undefined variables are _expected_.

## The solution

This PR introduces the method `Environment::has_var()`, alongside its wrapper `vip_has_env_var()`, to check if a custom variable is defined before attempt accessing it.

Or snippet above now becomes:

```php
if (Environment::has_var('FEATURE_X_ENABLED') && Environment::get_var('FEATURE_X_ENABLED')) {
  // Use feature X
}
```

or, if we set it only if true, just:

```php
if (Environment::has_var('FEATURE_X_ENABLED')) {
  // Use feature X
}
```

as a result, the behavior is the same but now we don't have notice triggered.

## Notes

1. unchecked accesses to undefined variables _still_ trigger the notice after this PR
2. the wrapper `vip_has_env_var()` behaves differently from the wrapped method: it returns a boolean, whereas the method returns a string (the name of the constant). That is to avoid call `strtoupper` and `sprintf` in both methods, improving performance
3. The changes are 100% backward compatible

# Changelog Description

#### Introduce `Environment::has_var()` method and `vip_has_env_var()` wrapper function

They allow custom environment variables existence checks without triggering notices.

# Checklist

# Steps to Test

1. Check out PR.
4. Write a code that make use of the new class method
5. Verify that you can check existence of set environment variables without trigger notices

### DIFF
--- a/lib/environment/class-environment.php
+++ b/lib/environment/class-environment.php
@@ -9,19 +9,19 @@ class Environment {
 	 */
 	const VAR_PREFIX = 'VIP_ENV_VAR';
 
-    /**
-     * Checks existence of custom environment variable values based on the supplied key and return
-     * its name if it is set, null otherwise.
-     *
-     * @param string $key The name of the environment variable.
-     * @return string|null
-     */
-    public static function has_var( string $key ): ?string {
-        $key       = strtoupper( $key );
-        $env_const = sprintf( '%s_%s', self::VAR_PREFIX, $key );
+	/**
+	 * Checks existence of custom environment variable values based on the supplied key and return
+	 * its name if it is set, null otherwise.
+	 *
+	 * @param string $key The name of the environment variable.
+	 * @return string|null
+	 */
+	public static function has_var( string $key ): ?string {
+		$key       = strtoupper( $key );
+		$env_const = sprintf( '%s_%s', self::VAR_PREFIX, $key );
 
-        return defined( $env_const ) ? $env_const : null;
-    }
+		return defined( $env_const ) ? $env_const : null;
+	}
 
 	/**
 	 * Attempts to return custom environment variable values based on the supplied key
@@ -31,9 +31,9 @@ class Environment {
 	 * @return string
 	 */
 	public static function get_var( string $key, $default_value = '' ) {
-		$env_const = self::has_var( $key );
+		$env_const = static::has_var( $key );
 
-		if ( $env_const !== null ) {
+		if ( null !== $env_const ) {
 			return constant( $env_const );
 		}
 
--- a/lib/environment/class-environment.php
+++ b/lib/environment/class-environment.php
@@ -9,47 +9,38 @@ class Environment {
 	 */
 	const VAR_PREFIX = 'VIP_ENV_VAR';
 
-	/**
-	 * Checks existence of custom environment variable values based on the supplied key and return
-	 * its name if it is set, null otherwise.
-	 *
-	 * @param string $key The name of the environment variable.
-	 * @return string|null
-	 */
-	public static function has_var( string $key ): ?string {
-		$key       = strtoupper( $key );
-		$env_const = sprintf( '%s_%s', self::VAR_PREFIX, $key );
-
-		return defined( $env_const ) ? $env_const : null;
-	}
+    /**
+     * Checks existence of custom environment variable values based on the supplied key.
+     *
+     * @param string $key The name of the environment variable.
+     * @return bool
+     */
+    public static function has_var( string $key ): bool {
+        return defined( self::calculate_env_const_name( $key ) );
+    }
 
 	/**
-	 * Attempts to return custom environment variable values based on the supplied key
+	 * Attempts to return custom environment variable values based on the supplied key.
 	 *
 	 * @param string $key The name of the environment variable.
-	 * @param string $default_value The value to return if no environment variable matches the key
-	 * @return string
+	 * @param mixed $default_value The value to return if no environment variable matches the key
+	 * @return mixed
 	 */
 	public static function get_var( string $key, $default_value = '' ) {
-		$env_const = static::has_var( $key );
+		$env_const = self::calculate_env_const_name( $key );
 
-		if ( null !== $env_const ) {
-			return constant( $env_const );
-		}
-
-		// The call was not able to retrieve an env variable
-		// log this as an E_USER_NOTICE for debugging
-		trigger_error( // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_trigger_error
-			sprintf(
-				'The ENV variable: "%s" did not exist. %s::get_var returned the default value',
-				$key, // phpcs:ignore WordPress.Security.EscapeOutput
-				__CLASS__
-			),
-			\E_USER_NOTICE
-		);
-
-		return $default_value;
+		return defined( $env_const ) ? constant( $env_const ) : $default_value;
 	}
+
+	/**
+	 * Calculates environment variable constant name based on the supplied key.
+	 *
+	 * @param string $key
+	 * @return string
+	 */
+    private static function calculate_env_const_name( string $key ): string {
+        return sprintf( '%s_%s', self::VAR_PREFIX, strtoupper( $key ) );
+    }
 
 	public static function is_sandbox_container( $hostname, $env = array() ) {
 		if ( false !== strpos( $hostname, '_web_dev_' ) || false !== strpos( $hostname, '-sbx-u' ) ) {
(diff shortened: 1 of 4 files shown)


jesse-greathouse (changes requested):
I love this change overall. Thank you for submitting this contribution.

It needs unit tests and some code quality changes to be merged. [I have pushed a PR to your fork which has the changes that I'm requesting](https://github.com/gmazzap/vip-go-mu-plugins/pull/2). If you you can merge the changes I suggested then I can approve this.

❤️  Thanks

gmazzap:
Thanks @jesse-greathouse !

Merged your commits! I'll never get used to core CS...

jesse-greathouse:
@gmazzap would removing the notice also be an adequate solution for this problem? We are contemplating whether removing the notice would be a way to solve this without additions to the code. It is still yet to be decided but I wanted to find out if that would work for your use-case?

jesse-greathouse (dismissed):
This code is a fine solution to the problem but we may decide to just remove the notice. Holding off on merging for just a few days while I analyze the outcome of that path.

gmazzap:
> would removing the notice also be an adequate solution for this problem? We are contemplating whether removing the notice would be a way to solve this without additions to the code. It is still yet to be decided but I wanted to find out if that would work for your use-case?

Yes, it would @jesse-greathouse .

I did not ask for that because that would be a breaking change (kind of), and I thought it was easier to get merged a 100% backward-compatible code.

lib/environment/class-environment.php
```diff
@@ -9,6 +9,20 @@ class Environment {
...
 	const VAR_PREFIX = 'VIP_ENV_VAR';
 
+	/**
+	 * Checks existence of custom environment variable values based on the supplied key and return
+	 * its name if it is set, null otherwise.
+	 *
+	 * @param string $key The name of the environment variable.
+	 * @return string|null
+	 */
+	public static function has_var( string $key ): ?string {
```
rebeccahum: `has_var()` should only check for the existence of the constant and return a `boolean` value rather than the constant if it's defined or `null` for undefined.

rebeccahum (changes requested):
Thanks for your contribution! Just a few notes:
- We can remove the notice itself since it's not proving to be that helpful
- Let's change the behavior of `has_var()` to return a boolean

gmazzap:
Thanks for the feedback @rebeccahum 

> We can remove the notice itself since it's not proving to be that helpful

Very fine by me, thanks!

> Let's change the behavior of has_var() to return a boolean

If we do that, we must re-calculate the constant's name in `get_var`.

In current code, `get_var()` looks [like this](https://github.com/gmazzap/vip-go-mu-plugins/blob/feature/environment-has-var/lib/environment/class-environment.php#L33-L38):

```php
public static function get_var( string $key, $default_value = '' ) {
	$env_const = static::has_var( $key );

	if ( null !== $env_const ) {
		return constant( $env_const );
	}

	// ...omissis...
}
```

As you can see, the constant's name is returned by `has_var()`.

But if `has_var()` returns a boolean, the code above would become:

```php
public static function get_var( string $key, $default_value = '' ) {
	if ( static::has_var( $key ) ) {
		$key = strtoupper( $key );
		$env_const = sprintf( '%s_%s', self::VAR_PREFIX, $key );

		return constant( $env_const );
	}

	// ...omissis...
}
```

This means we calculate the constant's name _again_ after `has_var` already did it.

That has 2 downsides:

1. we have duplicated logic
2. we have worse performance because we call `strtoupper` and `sprint` twice each instead of once each.

The first problem could be solved by creating a separate (possibly private) method that calculates the name and is used in both `has_var` and `get_var`.

The second problem could be solved by duplicating the code of `has_var()` in `get_var()`.

So the whole code, assuming the notice is removed, becomes:

```php
public static function has_var( string $key ) {
    return defined( static::calculate_env_const_name( $key ) );
}

public static function get_var( string $key, $default ) {
    $env_const = static::calculate_env_const_name( $key );

    return defined( $env_const ) ? constant( $env_const ) : $default;
}

private static function calculate_env_const_name( string $key ) {
    return sprintf( '%s_%s', self::VAR_PREFIX, strtoupper( $key ) );
}
```

That said, in all honesty, if we remove the notice from `get_var()` I'm not sure `has_var()` is really needed at all...

WPprodigy:
Your example code at the end looks good me! Since our minimum php version is 7.4, let's go ahead and add a typed return as well:

```
public static function has_var( string $key ): bool {
    return defined( static::calculate_env_const_name( $key ) );
}
```

_____

> That said, in all honesty, if we remove the notice from get_var() I'm not sure has_var() is really needed at all...

I can see some situations where it could help with code readability. Situations where the logic flow just depends on whether or not the env variable is there versus caring what it is at that moment. Without has_var, you'd have to do some form of `false !== Environment::get_var( 'key', false )`, but also ensuring not to use a "default" that is possibly stored in the env variable. So a bit messier code readability-wise than having a `has_var()` option :)

gmazzap:
Thank you @WPprodigy 

With my last commit:

- `Environment::get_var()` won't trigger a notice anymore
- `Environment::has_var()` returns a boolean
- `Environment::calculate_env_const_name()` is added
- Tests are adjusted for the changed return value of `Environment::has_var()`
- Type declarations are added wherever possible

---


> I can see some situations where it could help with code readability. 

I think the mistake was to use `''` as the default value for `get_env()`. If that had used `null`, considering env vars (which are set via VIP CLI) can not be null

```php
if ( null !== Environment::get_var( 'X' ) ) {
```

would have been identical to the not-that-much-different:

```php
if ( Environment::has_var( 'X' ) ) {
```

But anyway, `Environment::has_var` is a literal oneliner, I guess it won't harm.

lib/environment/class-environment.php
```diff
@@ -9,35 +9,39 @@ class Environment {
...
-				'The ENV variable: "%s" did not exist. %s::get_var returned the default value',
-				$key, // phpcs:ignore WordPress.Security.EscapeOutput
-				__CLASS__
-			),
-			\E_USER_NOTICE
-		);
+		$env_const = self::calculate_env_const_name( $key );
 
-		return $default_value;
+		return defined( $env_const ) ? constant( $env_const ) : $default_value;
```
rebeccahum, line 32: What do you think about simplifying it more to just:

```suggestion
		return self::has_var( $key ) ? constant( self::calculate_env_const_name( $key ) ) : $default_value;
```

rebeccahum (changes requested):
Thank you for addressing the changes. Just one more note:

lib/environment/class-environment.php
rebeccahum: Ah, nevermind. Then `calculate_env_const_name()` would be called redundantly if the var exists.

rebeccahum:
Currently seeing two failures with the tests:

```
1) Automattic\VIP\Helpers\Environment_Test::test_get_var_empty_key
Failed asserting that exception of type "PHPUnit\Framework\Error\Notice" is thrown.

2) Automattic\VIP\Helpers\Environment_Test::test_get_default_var
Failed asserting that exception of type "PHPUnit\Framework\Error\Notice" is thrown.
```

We can remove https://github.com/Automattic/vip-go-mu-plugins/blob/5e2df3194a1278884dc5c5b4c0955c478101beff/tests/lib/helpers/test-environment.php#L39 and https://github.com/Automattic/vip-go-mu-plugins/blob/5e2df3194a1278884dc5c5b4c0955c478101beff/tests/lib/helpers/test-environment.php#L29

gmazzap:
Thank you @rebeccahum 

Done.

lib/environment/class-environment.php
gmazzap: Yep, that's what I meant in my previous comment when I said _"the second problem could be solved by duplicating the code of has_var() in get_var()."_ :)

jesse-greathouse:
I think we may need some technical guidance on why some of the checks are not being run on this PR. I can only guess its because it's merging from a repo of a different origin but I can't say that for sure. Maybe @sjinks would know 🤷

jesse-greathouse (approved):
This looks great. 👍 Thanks for all the help from @gmazzap and input from @rebeccahum and @WPprodigy

sjinks:
@jesse-greathouse yes, CI does not run for forked repos.

I'll check what can be done there.

PR #3918, opened 2022-11-16
Prometheus: update the endpoint and query_var to be more unique, remove proxy restriction

rinatkhaziev:
## Description

It was decided that `metrics` is way too generic;  to avoid any sort of collision with userland code this PR makes the slug/query variable more unique - `vip-prom-metrics`.

## Checklist

## Steps to Test

1. Merge the PR
2. Hit the https://vip-local.vipdev.lndo.site/vip-prom-metrics
3. Verify it's there.


sjinks (approved):

jesse-greathouse (commented):
It seems like this string: `vip-prom-metrics` is being used 5 different times in this class. I think it may be better to define this string in a constant of the class so that it can be referred to and changed without making multiple edits.

PR #3924, opened 2022-11-18
Query Monitor: Add panel for VIP-specifics

rebeccahum:
## Description
Adds a VIP panel in Query Monitor.
screenshot_2022-11-18_at_12 13 38_pm

Adds a "VIP" panel with display of stacks information
## Checklist

## Steps to Test
1) Create a `.version` in `vip-go-mu-plugins` if using VIP dev-env with content of `{ "tag": "staging", "stack_version": "20221115190626-cb35ffb7e-1234" }`
2) Log in and go to Query Monitor and expect to see screenshot in Description

### DIFF
--- a/qm-plugins/qm-vip/class-qm-vip-collector.php
+++ b/qm-plugins/qm-vip/class-qm-vip-collector.php
@@ -35,8 +35,8 @@ class QM_VIP_Collector extends QM_Collector {
 
 		$info          = json_decode( $version );
 		$branch        = $info->tag ?? '';
-		$stack_version = $info->stack_version ?? [];
-		$version_info  = explode( '-', $stack_version );
+		$stack_version = $info->stack_version ?? null;
+		$version_info  = $stack_version ? explode( '-', $stack_version ) : [];
 		$date          = isset( $version_info[0] ) ? gmdate( 'F j, Y', strtotime( $version_info[0] ) ) : null;
 		$commit        = $version_info[1] ?? null;
 
--- a/qm-plugins/qm-vip/class-qm-vip-collector.php
+++ b/qm-plugins/qm-vip/class-qm-vip-collector.php
@@ -22,24 +22,38 @@ class QM_VIP_Collector extends QM_Collector {
 	}
 
 	private function process_version_file() {
-		$version_file = WP_CONTENT_DIR . '/mu-plugins/.version';
+		$version_file = WPMU_PLUGIN_DIR . '/.version';
 		if ( ! file_exists( $version_file ) ) {
-			return null;
+			$this->set_default_version_info();
+			return;
 		}
 		$version = file_get_contents( $version_file ); // phpcs:ignore WordPressVIPMinimum.Performance.FetchingRemoteData.FileGetContentsUnknown
 		if ( ! $version ) {
-			return null;
+			$this->set_default_version_info();
+			return;
 		}
 
 		$info          = json_decode( $version );
-		$tag           = $info->tag ?? null;
-		$stack_version = $info->stack_version ?? null;
+		$branch        = $info->tag ?? '';
+		$stack_version = $info->stack_version ?? [];
 		$version_info  = explode( '-', $stack_version );
 		$date          = isset( $version_info[0] ) ? gmdate( 'F j, Y', strtotime( $version_info[0] ) ) : null;
 		$commit        = $version_info[1] ?? null;
 
-		$this->data['mu-plugins-stack']  = $tag;
-		$this->data['version']['date']   = $date;
-		$this->data['version']['commit'] = $commit;
+		$this->data['mu-plugins'] = [
+			'branch' => $branch,
+			'commit' => $commit,
+			'date'   => $date,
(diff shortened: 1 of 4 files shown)


qm-plugins/qm-vip/class-qm-vip-collector.php
```diff
@@ -0,0 +1,45 @@
...
+
+	/**
+	 * @return void
+	 */
+	public function process() {
+		$this->process_version_file();
+	}
+
+	private function process_version_file() {
+		$version_file = WP_CONTENT_DIR . '/mu-plugins/.version';
```
rinatkhaziev: We can probably use `WPMU_PLUGIN_DIR` here, to simplify things.

qm-plugins/qm-vip/qm-vip.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+/**
+ * Plugin Name: Query Monitor VIP
+ * Description: Additional collector for Query Monitor for VIP-specific information.
+ * Version: 1.0
+ * Author: Automattic, Rebecca Hum
+ */
+
+if ( ! defined( 'VIP_GO_APP_ENVIRONMENT' ) || 'local' === constant( 'VIP_GO_APP_ENVIRONMENT' ) ) {
```
rinatkhaziev: In what case do we set `'local' === constant( 'VIP_GO_APP_ENVIRONMENT' )`?

rinatkhaziev (changes requested):
Few things I spotted:

1. I think we may be better off if we modify the data structure to be of the following shape:

```
$this->data[ 'mu-plugins' ] = [
	'branch' => 'production|staging',
	'date'   => 'Y-m-d',
	'commit' => 'gitsha'
];

2. Ideally we would return some defaults if the file is not present rather than bailing altogether.

File: qm-plugins/qm-vip/class-qm-vip-output-html.php [hunk 3]
```diff
@@ -0,0 +1,31 @@
...
+
+		return $menu;
+	}
+
+	public function output() {
+		$data       = $this->collector->get_data();
+		$commit_url = 'https://github.com/automattic/vip-go-mu-plugins/commit/' . $data['version']['commit'];
+		?>
+		<div class="qm qm-non-tabular" id="<?php echo esc_attr( $this->collector->id ); ?>">
+			<h3><strong>MU-Plugins Stack: </strong><?php echo esc_html( $data['mu-plugins-stack'] ); ?></h3>
```
rinatkhaziev: We use words "branch" both in announcement and docs (I think, @yolih correct me if I'm wrong), so we should match that here to avoid any confusion. 

Customers don't necessarily know what a "stack" is.

qm-plugins/qm-vip/qm-vip.php
rebeccahum: dev-env

qm-plugins/qm-vip/class-qm-vip-output-html.php
yolih: Correct! "Branch" is preferred to "stack".

rebeccahum:
Screenshot 2022-12-01 at 3 30 40 PM

qm-plugins/qm-vip/qm-vip.php
rinatkhaziev: I think it would be valuable to have it on because: 
a) We can bake in the version file in our mu-plugins image for dev env - the plan is to update it weekly with the release, so it may be valuable there
b) the file missing may indicate that this is a local/unbuilt copy of dev env

qm-plugins/qm-vip/class-qm-vip-collector.php
```diff
@@ -0,0 +1,59 @@
...
+			$this->set_default_version_info();
+			return;
+		}
+		$version = file_get_contents( $version_file ); // phpcs:ignore WordPressVIPMinimum.Performance.FetchingRemoteData.FileGetContentsUnknown
+		if ( ! $version ) {
+			$this->set_default_version_info();
+			return;
+		}
+
+		$info          = json_decode( $version );
```
rinatkhaziev, line 36: There's a bug in this block:

```php
php > $a = explode( '-', null );
php > var_dump( isset( $a[0] ) );
bool(true)
php > echo strtotime( $a[0] );
php > var_dump( strtotime( $a[0] ));
bool(false)
php > var_dump( gmdate( 'F j, Y', strtotime( $a[0] ) ) );
string(15) "January 1, 1970"
```

qm-plugins/qm-vip/class-qm-vip-collector.php
```diff
@@ -0,0 +1,59 @@
...
+		$branch        = $info->tag ?? null;
```
rinatkhaziev: On a separate note, here and forth we're expecting strings, so assigning `null` as a fallback value is not entirely semantically correct.

rinatkhaziev (approved):
Let's get this merged 🚀

PR #3934, opened 2022-11-23
Search: Add 'inactive' as an alias

rebeccahum:
## Description
Currently, we have `next`, `previous`, `active` as aliases in versioning. It would be good to have `inactive` as one as well, especially so we could do something like `wp vip-search index-versions delete post inactive`.

Add `inactive` as alias in versioning

## Checklist

## Steps to Test
1) Spin up new environment and create an index `wp vip-search index --setup`
2) Add another version `wp vip-search index-versions add post`
3) Verify there are two versions `wp vip-search index-versions list post`, with one being active and one inactive:
```
+--------+--------+--------------+----------------+----------------+
| number | active | created_time | activated_time | document_count |
+--------+--------+--------------+----------------+----------------+
| 1      | 1      |              |                | 2              |
| 2      |        | 1669227879   |                | 0              |
+--------+--------+--------------+----------------+----------------+
```

### Testing the `get_inactive_version_number()` directly from shell:
1) `$v = new \Automattic\VIP\Search\Versioning();`
2) `$indexable = \ElasticPress\Indexables::factory()->get( 'post' );`
3) `$v->get_inactive_version_number( $indexable );`
Expect to see `2` returned

### Testing it as an alias in CLI context
1) `wp vip-search index-versions delete post inactive` and confirm
2) Expect to see `Success: Successfully deleted index version 2 for type post`

### DIFF
--- a/search/includes/classes/class-versioning.php
+++ b/search/includes/classes/class-versioning.php
@@ -185,7 +185,8 @@ class Versioning {
 	}
 
 	/**
-	 * Grab just the version number for the inactive version
+	 * Grab just the version number for the inactive version. If there is more than one inactive version, grab
+	 * the first one.
 	 *
 	 * @param \ElasticPress\Indexable $indexable The Indexable to get the inactive version number for
 	 * @return int|WP_Error The current inactive version number
@@ -195,10 +196,6 @@ class Versioning {
 
 		if ( empty( $inactive_versions ) ) {
 			return new WP_Error( 'no-inactive-versions-found', 'No inactive versions.' );
-		}
-
-		if ( count( $inactive_versions ) > 1 ) {
-			return new WP_Error( 'many-inactive-versions-found', 'More than one inactive version found.' );
 		}
 
 		return array_key_first( $inactive_versions ); // We only need the first key since there's only one.
--- a/search/includes/classes/class-versioning.php
+++ b/search/includes/classes/class-versioning.php
@@ -187,7 +187,7 @@ class Versioning {
 	/**
 	 * Grab just the version number for the inactive version
 	 *
-	 * @param \ElasticPress\Indexable $indexable The Indexable to get the active version number for
+	 * @param \ElasticPress\Indexable $indexable The Indexable to get the inactive version number for
 	 * @return int|WP_Error The current inactive version number
 	 */
 	public function get_inactive_version_number( Indexable $indexable ) {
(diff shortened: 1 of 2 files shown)


search/includes/classes/class-versioning.php
```diff
@@ -184,6 +184,26 @@ public function get_active_version_number( Indexable $indexable ) {
 		return $active_version['number'];
 	}
 
+	/**
+	 * Grab just the version number for the inactive version
+	 *
+	 * @param \ElasticPress\Indexable $indexable The Indexable to get the active version number for
```
rinatkhaziev:
```suggestion
	 * @param \ElasticPress\Indexable $indexable The Indexable to get the inactive version number for
```

search/includes/classes/class-versioning.php
```diff
@@ -184,6 +184,26 @@ public function get_active_version_number( Indexable $indexable ) {
...
+	 * @return int|WP_Error The current inactive version number
+	 */
+	public function get_inactive_version_number( Indexable $indexable ) {
+		$inactive_versions = $this->get_inactive_versions( $indexable );
+
+		if ( empty( $inactive_versions ) ) {
+			return new WP_Error( 'no-inactive-versions-found', 'No inactive versions.' );
+		}
+
+		if ( count( $inactive_versions ) > 1 ) {
```
rinatkhaziev: I'm wondering if this is a correct approach to handling the many inactive versions. Like, we're asking for any inactive version, so maybe it's forth returning just the first one, rather than bail?

rebeccahum: @rinatkhaziev That was my first approach initially...but I changed it to throw an error because if the user is requesting for an inactive version, I figured it'd be better to throw an error to ensure that that the user instead targets a specific inactive version. I'm happy go either or.

rinatkhaziev: @rebeccahum I think having a single inactive version returned has more usefulness - i.e. if I as a user want to clean up some inactives I'd rather have a first inactive deleted that have an error thrown at me without any clue.

An alternative scenario I think might be to report all the inactive versions in the error:

`Found following inactive versions: 2,4,5. Please specify which one you would like to delete`

rebeccahum: @rinatkhaziev Let's do the first scenario you've proposed. Feedback should be addressed now.

rinatkhaziev (approved):

PR #3945, opened 2022-11-30
Search: check response size before trying to cache the ES response to avoid potentially hammering Memcached

rinatkhaziev:
## Description

Previous approach didn't take the size of response into the account when making a decision whether to cache or not to cache. This PR adds a safeguard to not try to set large values.

The 6MB mark is rather arbitrary, with compression it'll likely fit in Memcached but since our goal is to reduce the chances of `wp_cache_set` barrages this felt like a good compromise.

We modified the request cache behavior to not try to cache any responses over 6 megabytes in size to prevent potential negative impact on object cache.

## Checklist

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -1028,7 +1028,7 @@ class Search {
 			add_site_option( $index_exists_option_name, $response );
 		}
 
-		if ( $is_cacheable ) {
+		if ( $is_cacheable && isset( $response_body_json ) && strlen( $response_body_json ) < 6 * MB_IN_BYTES ) {
 			// phpcs:ignore WordPressVIPMinimum.Performance.LowExpiryCacheTime.CacheTimeUndetermined
 			wp_cache_set( $cache_key, $response, self::SEARCH_CACHE_GROUP, 5 * MINUTE_IN_SECONDS );
 		}


search/includes/classes/class-search.php
```diff
@@ -1028,7 +1028,7 @@ public function filter__ep_do_intercept_request( $request, $query, $args, $failu
 			add_site_option( $index_exists_option_name, $response );
 		}
 
-		if ( $is_cacheable ) {
+		if ( $is_cacheable && isset( $response_body_json ) && strlen( $response_body_json ) < 6 * MB_IN_BYTES ) {
```
rebeccahum, line 1031: Do we have any data on what size it b0rks Memcached?

rinatkhaziev: yes we do. anything over 1MB compressed. The compression rate is usually 8-12 or so. We do a more complex check in alloptions-limit (see below) but I figured this will be cheap and easy. 

https://github.com/Automattic/vip-go-mu-plugins/blob/fdd99cf535e55eeda46d7a11a4cbb1603d8bc666/alloptions-limit.php#L60-L64

Do you have any particular concern about this approach?

rebeccahum: Sorry, I wasn't talking about the general overall limit, but if we have had an average of certain cached responses that did break Memcached, that's all. 🙂  Is it too expensive to calculate from the `gzdeflate()` length? Feels like that would be more accurate.

rinatkhaziev: > we have had an average of certain cached responses that did break Memcached

There were a few instances where this functionality certainly didn't help :) I don't think I can share those details here though.

>  Is it too expensive to calculate from the gzdeflate() length

I haven't measured in this particular context TBH but knowing that some sites perform dozens of queries per request I'd say that potential overhead may be significant in terms of cost vs benefit.

rebeccahum (approved):

PR #3946, opened 2022-11-30
CANTINA-914: Account for multisite in Feature rollout

rebeccahum:
## Description
When we use the feature roll-out, it should be taking into account the subsites in multisites since we don't want all of the subsites to be enabled.

### Plugin Updated: Feature

Feature percentage roll out will take into account a multisite's subsites to ensure not all subsites are rolled out together at once.

## Checklist

### DIFF
--- a/lib/feature/class-feature.php
+++ b/lib/feature/class-feature.php
@@ -77,7 +77,7 @@ class Feature {
 		// Is the bucket enabled?
 		$threshold = $percentage * 100; // $percentage is decimal
 
-		if ( is_multisite() && ( $bucket < $threshold ) ) {
+		if ( is_multisite() && $bucket < $threshold ) {
 			// For multisites, we don't want to roll out for all subsites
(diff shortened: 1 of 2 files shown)


rebeccahum:
Tests failing are unrelated I believe. Looks like it is happening on latest + multisite + JP. Will take a closer look later.

lib/feature/class-feature.php
```diff
@@ -79,6 +77,11 @@ public static function is_enabled_by_percentage( $feature ) {
 		// Is the bucket enabled?
 		$threshold = $percentage * 100; // $percentage is decimal
 
+		if ( is_multisite() && ( $bucket < $threshold ) ) {
```
rinatkhaziev:
```suggestion
		if ( is_multisite() && $bucket < $threshold ) {
```

rinatkhaziev (approved):

PR #3957, opened 2022-12-06
Query Monitor (VIP): Open commit in new window

rebeccahum:
## Description
Maybe me, but it's an annoyance if I open a link in a working window and it doesn't pop up in a new one.

### Plugin Updated: Query Monitor (VIP)

Open commit in new window.

## Checklist

## Steps to Test
1) Create `.version` file with contents: `{ "tag": "staging", "stack_version": "20221115190626-cb35ffb7e-1234" }`
2) Go to Query Monitor > VIP and click on date displayed
3) Expect it to take you to a new window

### DIFF
--- a/qm-plugins/qm-vip/class-qm-vip-output-html.php
+++ b/qm-plugins/qm-vip/class-qm-vip-output-html.php
@@ -26,7 +26,7 @@ class QM_VIP_Output extends QM_Output_Html {
 			<?php
 			if ( isset( $data['mu-plugins']['commit'] ) && isset( $data['mu-plugins']['date'] ) ) {
 				echo '<p><a href="https://github.com/automattic/vip-go-mu-plugins/commit/' . rawurlencode( $data['mu-plugins']['commit'] ) .
-				'" alt="GitHub URL of the commit that the stack was deployed from."><i><strong>Last modified: </strong>' . esc_html( $data['mu-plugins']['date'] ) .
+				'" alt="GitHub URL of the commit that the stack was deployed from." target="_blank"><i><strong>Last modified: </strong>' . esc_html( $data['mu-plugins']['date'] ) .
 				'</i></a></p>';
 			}
 			?>
--- a/qm-plugins/qm-vip/class-qm-vip-output-html.php
+++ b/qm-plugins/qm-vip/class-qm-vip-output-html.php
@@ -26,7 +26,7 @@ class QM_VIP_Output extends QM_Output_Html {
 			<?php
 			if ( isset( $data['mu-plugins']['commit'] ) && isset( $data['mu-plugins']['date'] ) ) {
 				echo '<p><a href="https://github.com/automattic/vip-go-mu-plugins/commit/' . rawurlencode( $data['mu-plugins']['commit'] ) .
-				'"><i><strong><span class="screen-reader-text">Open in new tab </span>Last modified: </strong>' .
+				' target="_blank"><i><strong><span class="screen-reader-text">Open in new tab </span>Last modified: </strong>' .
 				esc_html( $data['mu-plugins']['date'] ) . '</i></a></p>';
 			}
 			?>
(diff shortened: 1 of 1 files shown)


qm-plugins/qm-vip/class-qm-vip-output-html.php
```diff
@@ -26,7 +26,7 @@ public function output() {
 			<?php
 			if ( isset( $data['mu-plugins']['commit'] ) && isset( $data['mu-plugins']['date'] ) ) {
 				echo '<p><a href="https://github.com/automattic/vip-go-mu-plugins/commit/' . rawurlencode( $data['mu-plugins']['commit'] ) .
-				'" alt="GitHub URL of the commit that the stack was deployed from."><i><strong>Last modified: </strong>' . esc_html( $data['mu-plugins']['date'] ) .
+				'" alt="GitHub URL of the commit that the stack was deployed from." target="_blank"><i><strong>Last modified: </strong>' . esc_html( $data['mu-plugins']['date'] ) .
```
GaryJones: `alt` attribute isn't valid on `a` elements - I think it meant `title`.

And the `title` attribute isn't accessibility friendly, and shouldn't be used here.

And if the link is being forced into a new tab, then it needs a screen-reader-text span to indicate to screen reader users that it opens in a new tab.

WPprodigy:
Gotta `cmd-click` links you aren't sure about ;p

qm-plugins/qm-vip/class-qm-vip-output-html.php
```diff
@@ -26,8 +26,8 @@ public function output() {
 			<?php
 			if ( isset( $data['mu-plugins']['commit'] ) && isset( $data['mu-plugins']['date'] ) ) {
 				echo '<p><a href="https://github.com/automattic/vip-go-mu-plugins/commit/' . rawurlencode( $data['mu-plugins']['commit'] ) .
-				'" alt="GitHub URL of the commit that the stack was deployed from."><i><strong>Last modified: </strong>' . esc_html( $data['mu-plugins']['date'] ) .
-				'</i></a></p>';
+				'" aria-label="GitHub URL of the commit that the stack was deployed from."><i><strong><span class="screen-reader-text">Open in new tab </span>Last modified: </strong>' .
```
GaryJones: Screen readers use `aria-label` to replace the whole of the anchor text - so screen reader users wouldn't hear the "Last modified: _date_" or the new "Open in a new tab" bit.
```suggestion
				'"><i><strong><span class="screen-reader-text">Open in new tab </span>Last modified: </strong>' .
```

GaryJones (changes requested):

qm-plugins/qm-vip/class-qm-vip-output-html.php
rebeccahum: @GaryJones So if I'm understanding you correctly, remove `aria-label` then?

GaryJones: Yes - it's rarely needed on links - the most common use for links is when it is more deeply describing one of a collection of links with repeated texr e.g. Read more, to make them sufficiently unique.

In this case, it was being used to describe where the link went, even though that's not something we would see for any other link. As it would have meant the loss of the Last modified: _date_ information to describe a GitHub commit URL, it is best to remove the attribute.

qm-plugins/qm-vip/class-qm-vip-output-html.php
```diff
@@ -26,8 +26,8 @@ public function output() {
 			<?php
 			if ( isset( $data['mu-plugins']['commit'] ) && isset( $data['mu-plugins']['date'] ) ) {
 				echo '<p><a href="https://github.com/automattic/vip-go-mu-plugins/commit/' . rawurlencode( $data['mu-plugins']['commit'] ) .
-				'" alt="GitHub URL of the commit that the stack was deployed from."><i><strong>Last modified: </strong>' . esc_html( $data['mu-plugins']['date'] ) .
-				'</i></a></p>';
+				'"><i><strong><span class="screen-reader-text">Open in new tab </span>Last modified: </strong>' .
```
GaryJones: I think your `target="_blank"` got lost...

GaryJones (approved):
✅

PR #3958, opened 2022-12-06
Mitigate corrupted get_sites cache entry in multisite context

rinatkhaziev:
## Description

WordPress doesn't handle get_sites population gracefully, and in case the query fails the corrupted value may get stored, essentially breaking sites in a hard-to-recover fashion. 

Luckily it's simple to detect - `get_sites()` return value should never be empty and if it is the things went south. 

The recovery is pretty simple too, we bump the sites_last_changed - this would invalidate the cache key, and the next request would try to run the query again. 

### Sunrise.php update

We've updated sunrise.php to recover from certain cache pollution edge case. 

## Checklist

## Steps to Test

1. The cache key is a bit salty, so the quickest way to test would be to hack `wp-includes/class-wp-site-query.php#361` to read as follows:

```php
		wp_cache_set( $cache_key, array(
			'site_ids' => [],
			'found_sites' => 0,
		), 'sites' );
		$cache_value = wp_cache_get( $cache_key, 'sites' );
```
2. Hit an existing subsite, verify you see the error
3. Comment out the hack
4. Hit the subsite again, verify that it recovered.

### DIFF
--- a/lib/sunrise/sunrise.php
+++ b/lib/sunrise/sunrise.php
@@ -66,6 +66,13 @@ function handle_not_found_error( $error_type ) {
 		exit;
 	}
 
+	// Empty get_sites() return would signify potential cache pollution
+	// We side-step by bumping the last changed which would result in the cache key change
+	// So the next request can recover
+	if ( 'site' == $error_type && ! get_sites() ) {
+		wp_cache_set_sites_last_changed();
+	}
+
 	$is_web_request = Context::is_web_request();
 	if ( $is_web_request ) {
 		$mu_plugin_dir       = defined( 'WPMU_PLUGIN_DIR' ) ? constant( 'WPMU_PLUGIN_DIR' ) : constant( 'WP_CONTENT_DIR' ) . '/mu-plugins';


lib/sunrise/sunrise.php
```diff
@@ -66,6 +66,13 @@ function handle_not_found_error( $error_type ) {
 		exit;
 	}
 
+	// Empty get_sites() return would signify potential cache pollution
+	// We side-step by bumping the last changed which would result in the cache key change
+	// So the next request can recover
+	if ( 'site' == $error_type && ! get_sites() ) {
+		wp_cache_set_sites_last_changed();
```
WPprodigy, line 73: `get_sites()` could be an unbounded query on a large MS. Since really we just want to make sure we aren't involved with cache pollution, maybe lets just check the cache itself: https://github.com/WordPress/WordPress/blob/97ba08426dd1ca896596646e708b2d17387ed287/wp-includes/class-wp-site-query.php#L357-L361

Could check the key that is used when `$args` is empty to get the same result you're currently doing, just without the extra queries running at the end. Perhaps better would be to check the query args/key that WP core uses when looking for a matching subsite.

rinatkhaziev:
So this likely won't work because `get_sites()` call would result in a different cache key because the call with corrupted cache entry would include domain and path.

PR #3963, opened 2022-12-08
Search: Allow no active versions and change initial state to no versions

rebeccahum:
## Description
This change does the below:
- Allows for a state of having no active versions in the index version management (before, we would default to 1 no matter what)
- Adds a `deactivate` WP-CLI command: `wp vip-search index-versions deactivate post 1`
- Defaults to having no index versions on initial state

If there's no active versions but search is enabled, search will fall back to DB.

Add deactivate CLI command, allow a state of having no active indexes and set initial state to no versions

## Checklist

## Steps to Test
Note: when testing on dev-env, you'll need to run `wp vip-search index --setup` since the current version uses `wp elasticsearch index --setup` which doesn't take into account the changes in this PR. see https://github.com/Automattic/vip-container-images/pull/390
1) Run `wp vip-search index-versions deactivate post <index-id>` on an active version and use `wp vip-search index-versions list post` to verify it's no longer marked as active
2) Run a search query via `?s=<searchTerm>` and observe that in Search Dev Tools, there's no query run anymore
3) Run `wp vip-search index-versions delete post <index-id` until there's no more indexes left in `wp vip-search index-versions list post`
4) Run `wp vip-search index --using-versions --setup --indexables=post` and expect an error message returned
5) Run`wp vip-search index --setup` and verify that a new version has been created per `wp vip-search index-versions list post`
6) Repeat step 5 but with the expectation that the is only one index and it's active and different from the one in step 6)
7) Run`wp vip-search index --setup --indexables=post --version=<index id from step 6>` and verify that no new indexes have been added and only the index from step 6 was indexed


search/includes/classes/class-versioning.php
```diff
@@ -221,15 +226,14 @@ public function get_versions( Indexable $indexable, bool $provide_default = true
 		$versions = [];
 
 		if ( $indexable->global ) {
-			$versions = get_site_option( self::INDEX_VERSIONS_OPTION_GLOBAL, array() );
+			$versions = get_site_option( self::INDEX_VERSIONS_OPTION_GLOBAL );
```
rinatkhaziev, line 231: What's the reason behind changing the default value to empty string here?

rebeccahum: Isn't it being changed to `false`? And in regards for the _why_ it is being changed to `false` is because we have a way of knowing that the index versioning hasn't been touched yet — when the option does not exist. Then, we can use that later on to determine whether to return `1` as a default version versus just not having any indexes (albeit, if they get in a state where none exist, as they have been deleted).

rinatkhaziev: You're right, it's `false`, I thought it's `''` (an empty string). My question was more around the fact we're expecting `$versions` to be an array.

According to my testing, on a fresh dev-env instance this option is present:

```
wp> get_option( 'vip_search_index_versions' );
=> array(1) {
  ["post"]=>
  array(1) {
    [1]=>
    array(4) {
      ["number"]=>
      int(1)
      ["active"]=>
      bool(true)
      ["created_time"]=>
      NULL
      ["activated_time"]=>
      NULL
    }
  }
}
```

rinatkhaziev: furhtermore, IF I delete it it get's recreated on the next request:

```
wp> delete_option( 'vip_search_index_versions' );
=> bool(true)
wp> get_option( 'vip_search_index_versions' );
=> bool(false)
wp> exit
 ~  vip dev-env exec --slug search -- wp shell
Running validation steps...
✓ Check for docker installation
✓ Check access to docker for current user
✓ Check DNS resolution
wp> get_option( 'vip_search_index_versions' );
=> array(1) {
  ["post"]=>
  array(1) {
    [1]=>
    array(4) {
      ["number"]=>
      int(1)
      ["active"]=>
      bool(true)
      ["created_time"]=>
      NULL
      ["activated_time"]=>
      NULL
    }
  }
}
```

rebeccahum: Hmmm on a fresh dev-env instance where the versioning hasn't been touched at all, I'm getting:
```
wp> get_option( 'vip_search_index_versions' );
=> bool(false)
```

rinatkhaziev: Same happens not-locally, and for multiple versions it even detects it correctly.:

```
wp> get_option( 'vip_search_index_versions' );
array(1) {
  'post' =>
  array(2) {
    [1] =>
    array(4) {
      'number' =>
      int(1)
      'active' =>
      bool(true)
      'created_time' =>
      NULL
      'activated_time' =>
      NULL
    }
    [2] =>
    array(4) {
      'number' =>
      int(2)
      'active' =>
      bool(false)
      'created_time' =>
      int(1644466599)
      'activated_time' =>
      NULL
    }
  }
}
wp> update_option( 'vip_search_index_versions_backup', get_option( 'vip_search_index_versions' ) );
bool(true)
wp> get_option( 'vip_search_index_versions' );
bool(false)
root@cantina-trunk-staging-sbx-u139-54bb787cd9-hppgz:~# wp shell

wp> get_option( 'vip_search_index_versions' );
bool(false)
root@cantina-trunk-staging-sbx-u139-54bb787cd9-hppgz:~# wp shell
wp> get_option( 'vip_search_index_versions' );
array(1) {
  'post' =>
  array(2) {
    [1] =>
    array(4) {
      'number' =>
      int(1)
      'active' =>
      bool(true)
      'created_time' =>
      NULL
      'activated_time' =>
      NULL
    }
    [2] =>
    array(4) {
      'number' =>
      int(2)
      'active' =>
      bool(false)
      'created_time' =>
      NULL
      'activated_time' =>
      NULL
    }
  }
}
```

rebeccahum:
@rinatkhaziev This is ready for review now.

rinatkhaziev:
I've been testing it a bit and it's a bit weird in terms that you can deactivate the version and the queries are not running but `wp vip-search index` would still run. I'm not sure whether that's a correct behavior.

rebeccahum:
OK, this is what I have for possible combinations when using `index` for `maybe_setup_index_version()`:

`wp vip-search index --setup`
* No indexes exist at all => Create a new one and activate it
* No active indexes => Find the first one and activate it
* An active index exists => Use the active one

`wp vip-search index --using-versions --indexables=post` or 
* No indexes exist => throw error
* No active indexes exist => throw error
* More than 1 index exists => throw error
* An active index exists but less than 2 indexes => Continue with logic in creating a new one and activating it

`wp vip-search index --version=<int> --indexables=post`
State of index existence doesn't matter => use target version passed in

search/includes/classes/commands/class-corecommand.php
```diff
@@ -121,44 +121,78 @@ private function list_indexes() {
 	}
 
 	protected function maybe_setup_index_version( $assoc_args ) {
```
rinatkhaziev, line 127: This method looks a bit intense now, can we make it less of a 
image

search/includes/classes/commands/class-versioncommand.php
```diff
@@ -332,6 +332,106 @@ protected function activate_helper( Indexable $indexable, $version_number_to_act
...
+	protected function deactivate_helper( Indexable $indexable, $version_number_to_deactivate ) {
+		$search = \Automattic\VIP\Search\Search::instance();
+
+		$new_version_number = $search->versioning->normalize_version_number( $indexable, $version_number_to_deactivate );
+
+		if ( is_wp_error( $new_version_number ) ) {
+			return WP_CLI::error( sprintf( 'Index version %s is not valid: %s', $version_number_to_deactivate, $new_version_number->get_error_message() ) );
+		}
+
+		$inactive_versions = $search->versioning->get_inactive_versions( $indexable );
```
rinatkhaziev: Is this really needed? `deactivate_version` seems to be handling already inactive version?

rinatkhaziev: In fact, do we need this method at all?

rinatkhaziev (changes requested):
Left some feedback.

rinatkhaziev (dismissed):
Manually tested in all possible scenarios I could come up with, seems to work great, but would still be nice to figure out the rollout strategy for this, as the change is pretty big.

PR #3967, opened 2022-12-10
ci: Fix failing core tests

sjinks:
Ref: https://github.com/Automattic/vip-go-mu-plugins/actions/runs/3660081303/jobs/6186815697
Ref: https://core.trac.wordpress.org/ticket/57306


GaryJones (approved):
Only a single instance of `test_multiple_location_headers` in that file, so no unexpected side effects from the swap ✅

rinatkhaziev (approved):

PR #3980, opened 2022-12-15
Default setting Debug Bar to false and add deprecation + admin notice

rebeccahum:
## Description
As part of https://lobby.vip.wordpress.com/2022/12/14/deprecation-notice-debug-bar-january-31-2023/, @WPprodigy suggested that we disable it right now and add in deprecation notices.
Screenshot 2022-12-15 at 12 38 35 PM

### Plugin Updated: Debug Bar

We added a deprecation notice to debug bar if it is set to be enabled.

## Checklist

## Steps to Test
1) In `client-mu-plugins`, add the below:
```
remove_all_filters( 'debug_bar_enable' );
add_filter( 'debug_bar_enable', function( $enable ) {
    if ( ! is_admin() ) {
        $enable = false;
    }
    return $enable = true;
} );
```
2) Log in as admin and expect to see admin notice. Check logs and also see deprecation notice.
3) Log out and log in as user, and expect to not see admin notice

### DIFF
--- a/debug-bar.php
+++ b/debug-bar.php
@@ -9,19 +9,6 @@ Author URI: https://wordpress.org/
 Text Domain: debug-bar
 */
 
-// If the user is an Automattician (typically a vip_support user), then force-enable Debug Bar.
-add_filter( 'debug_bar_enable', function( $enable ) {
-	if ( is_automattician() ) {
-		return true;
-	}
-
-	if ( defined( 'WP_ENVIRONMENT_TYPE' ) && 'local' === WP_ENVIRONMENT_TYPE ) {
-		return true;
-	}
-
-	return $enable;
-}, PHP_INT_MAX );
-
 // We only need to load the files if it's enabled
 add_action( 'set_current_user', function() {
 	$enable = apply_filters( 'debug_bar_enable', false );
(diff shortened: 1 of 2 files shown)


rebeccahum:
@WPprodigy What say you? Do you think throwing a deprecation notice is too log spammy?

debug-bar.php
```diff
@@ -9,18 +9,8 @@
...
-	}
-
-	if ( defined( 'WP_ENVIRONMENT_TYPE' ) && 'local' === WP_ENVIRONMENT_TYPE ) {
-		return true;
-	}
-
-	return $enable;
-}, PHP_INT_MAX );
+// Disabling as first step to deprecation, see https://lobby.vip.wordpress.com/2022/12/14/deprecation-notice-debug-bar-january-31-2023/
+add_filter( 'debug_bar_enable', '__return_false', PHP_INT_MAX );
```
WPprodigy: I would leave out this part, just delete our usage of the filter is enough.

debug-bar.php
```diff
@@ -30,6 +20,8 @@
 		return;
 	}
 
+	_deprecated_function( 'debug_bar_enable', '1.1.3', 'wpcom_vip_qm_enable' );
```
WPprodigy: Unfortunately won't end up in logs. Need to `trigger_error( $message, E_USER_WARNING )` if we want to be able to look for the notice / expose to customers.

debug-bar.php
rebeccahum: Haha, you beat me to it, I actually deleted it the second pass 😄

WPprodigy (approved):

PR #3989, opened 2022-12-20
Prom: speed things up by moving out the rendering logic into a non-WP endpoint

rinatkhaziev:
## Description

This PR moves the rendering logic entirely out of WP context. This speeds up the rendering significantly + we're entirely eliminating the issues with sunrise.php in multisite contexts and eliminating any sort of randomness due to the slow customer code. 

The only drawback is right now Post Stats Collector is not working (it requires some core functions to be defined and will need a bit of a refactor)

## Checklist

## Steps to Test

I use a bash function to test:

```bash
function perf {
  curl -o /dev/null -s -w "%{time_connect} + %{time_starttransfer} = %{time_total}\n" "$1"
}
```

Measure the performance locally:

```
perf http://instance.vipdev.lndo.site/.vip-prom-metrics

0.005555 + 0.204087 = 0.204367
```

Apply the PR, and re-run the measurement, you should see an improvement.

```
0.006917 + 0.035189 = 0.035304
```

On VIP, port-forward and do something like 

```
perf http://localhost:8090/.vip-prom-metrics
```


prometheus/endpoint.php
```diff
@@ -0,0 +1,46 @@
...
+use Automattic\VIP\Utils\Context;
+use Prometheus\CollectorRegistry;
+use Prometheus\RegistryInterface;
+use Prometheus\RenderTextFormat;
+use Prometheus\Storage\Adapter;
+use Prometheus\Storage\APCng;
+use Prometheus\Storage\InMemory;
+
+require_once __DIR__ . '/vendor/autoload.php';
+require_once dirname( __DIR__ ) . '/lib/utils/class-context.php';
```
rebeccahum, line 13: Is this redundant? `dirname( __DIR__)`

rebeccahum (changes requested):

rebeccahum (approved):

PR #3998, opened 2022-12-26
Check plugin updates after every 6 hours

gagan0123:
## Description

The Codebase Manager plugin keeps making POST requests to the WordPress' `update-check` (`ttps://api.wordpress.org/plugins/update-check/1.1/`) endpoint on every request.

In the code it was apparent that the developer intended to make such requests after every 6 hours, but instead of `<` they used `>` thereby ending up making those POST requests on every request instead.

## Changelog Description

This PR will ensure that sites using vip-go-mu-plugins will make requests to check for plugin updates after every 6 hours rather than on every request that they are doing currently.

## Checklist

## Steps to Test

Before this PR:
1. Enable Query Monitor for your user
2. Open any `wp-admin` page (Dashboard page)
3. Check the `HTTP API Calls` section in the Query Monitor dashboard.
4. You will always see a POST request to `https://api.wordpress.org/plugins/update-check/1.1/` 
image

After this PR:
1. Repeat the same steps as mentioned above, but you will only see the request after every 6 hours.

### DIFF
--- a/codebase-manager/plugins/plugins-manager.php
+++ b/codebase-manager/plugins/plugins-manager.php
@@ -67,7 +67,7 @@ class PluginsManager {
 
 		// Maybe refresh.
 		$last_refreshed = isset( $update_cache->last_checked ) ? ( time() - $update_cache->last_checked ) : null;
-		if ( ! $last_refreshed || 6 * HOUR_IN_SECONDS < $last_refreshed ) {
+		if ( ! $last_refreshed || $last_refreshed > ( 6 * HOUR_IN_SECONDS ) ) {
 			wp_update_plugins();
 			$update_cache = get_site_transient( 'update_plugins' );
 		}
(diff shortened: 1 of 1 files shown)


GaryJones:
Looks like a good catch cc @WPprodigy. Thanks, @gagan0123!

I think switching it to `$last_refreshed > (6 * HOUR_IN_SECONDS)` would be easier for more folks to comprehend.

gagan0123:
Updated patch as suggested @GaryJones

gagan0123:
Thanks @GaryJones 
Updated the patch as suggested

rebeccahum (commented):
This change makes sense to me...will leave it to @WPprodigy for what he thinks 🙂 
Side note: I am not seeing an API call on **every** request though as reported since when `wp_update_plugins()` is called, it checks the time gate (12 hrs) before making the remote request: https://github.com/WordPress/WordPress/blob/3ba44120d0ffa7ac330a1b7e3f8b363ce78f8060/wp-includes/update.php#L361

WPprodigy:
Oops! Nice catch. I'll test and then merge in this week.

Also of note, this logic only runs on `admin_init` & not during ajax, so frontend performance shouldn't be affected. Definitely should get patched up though.

WPprodigy (approved):
Works good. thanks again!

PR #4001, opened 2022-12-30
VIP Request Block: Add function to allow for blocking by partial User Agent string.

mslinnea:
## Description
I would like to block requests by a partial match on the user agent string.  The use case is a User Agent header like 'WordPress/6.1.1; https://www.example.com' which is the default format of a User Agent string to WordPress.  If I block by exact match, as soon as WordPress is upgraded, this exact match will no longer work.  

## Changelog Description

Adds VIP_Request_Block::ua_partial_match

## Checklist

## Steps to Test

Example:

1. Check out PR.
1. Follow documentation in https://docs.wpvip.com/how-tos/block-requests/ to test, except use `VIP_Request_Block::ua_partial_match` with a substring instead of `VIP_Request_Block::ua`

### DIFF
--- a/lib/class-vip-request-block.php
+++ b/lib/class-vip-request-block.php
@@ -88,8 +88,8 @@ class VIP_Request_Block {
 	 * @return void|bool
 	 */
 	public static function ua_partial_match( string $user_agent_substring ) {
-		// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotValidated, WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___SERVER__HTTP_USER_AGENT__
-		if ( isset( $_SERVER['HTTP_USER_AGENT'] ) && str_contains( $_SERVER['HTTP_USER_AGENT'], $user_agent_substring ) ) {
+		// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotValidated, WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___SERVER__HTTP_USER_AGENT__, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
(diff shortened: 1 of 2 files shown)


lib/class-vip-request-block.php
```diff
@@ -81,6 +81,21 @@ public static function ua( string $user_agent ) {
...
 
+	/**
+	 * Block by partial match of the user agent header
+	 *
+	 * @param string $user_agent_substring target user agent to be blocked.
+	 * @return void|bool
+	 */
+	public static function ua_partial_match( string $user_agent_substring ) {
+		// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotValidated, WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___SERVER__HTTP_USER_AGENT__
+		if ( isset( $_SERVER['HTTP_USER_AGENT'] ) && str_contains( $_SERVER['HTTP_USER_AGENT'], $user_agent_substring ) ) {
```
sjinks: `str_contains()` appeared in PHP 8; we still need to be compatible with PHP 7.4. Could you please update the code to use `strpos()` or whatever is appropriate?

Thank you!

sjinks (changes requested):

rebeccahum:
@sjinks I committed the changes on behalf.

sjinks:
@rebeccahum thank you, I will take a closer look tomorrow - I see SonarCloud complains etc.

GaryJones:
It's suggesting that [this line](https://github.com/Automattic/vip-go-mu-plugins/pull/4001/files?w=1#diff-2aaa18fe2bb071a15fb049eb6bc4dcf92755b5f1baa6171d13e26330f629b38fR93) should be `static::block_and_log()` and not `self::block_and_log()` - i.e. make use of late static binding (PHP 5.3) to reference any child class that may have extended this class which may overwrite `block_and_log()`. 

Same fix will be needed in at least one other method too.

sjinks (approved):

PR #4021, opened 2023-01-04
CI: Only run search-e2e workflow when search path is modified and on PRs only

rebeccahum:
## Description
It's adding too much waiting time during deploys


rinatkhaziev (approved):
This is fantastic!!!

PR #3979, opened 2022-12-14
Deprecate debug bar

rebeccahum:
## Description
https://lobby.vip.wordpress.com/2022/12/14/deprecation-notice-debug-bar-january-31-2023/

### Plugin Removed: Debug Bar

We removed Debug Bar. Please see https://lobby.vip.wordpress.com/2022/12/14/deprecation-notice-debug-bar-january-31-2023/ for more details.

## Checklist

Please make sure the items below have been covered before requesting a review:

- [x] This change works and has been tested locally (or has an appropriate fallback).
- [ ] This change works and has been tested on a Go sandbox.
- [ ] This change has relevant unit tests (if applicable).
- [ ] This change uses a rollout method to ease with deployment (if applicable - especially for large scale actions that require writes).
- [ ] This change has relevant documentation additions / updates (if applicable).
- [x] I've created a changelog description that aligns with the provided examples.

## Steps to Test
1) `add_filter( 'debug_bar_enable', '__return_true' );`
2) Expect to see no moar debug bar

PR #4035, opened 2023-01-09
Query Monitor: Add tab for DB Connections

rebeccahum:
## Description
It's being removed in https://github.com/Automattic/vip-go-mu-plugins/pull/3979, so let's add it into QM.

Screenshot 2023-01-09 at 2 19 07 PM

Added panel for DB Connections and only load qm-plugins if user has QM capability

## Checklist

## Steps to Test
1) Enable Debug Bar
2) Go to Query Monitor > DB Connections and see if it matches the Debug Bar > DB Connections

### DIFF
--- a/qm-plugins/qm-db-connections/class-qm-db-connections-collector.php
+++ b/qm-plugins/qm-db-connections/class-qm-db-connections-collector.php
@@ -18,28 +18,23 @@ class QM_DB_Connections_Collector extends QM_Collector {
 	 * @return void
 	 */
 	public function process() {
-		foreach ( $GLOBALS as $global ) {
-			if ( ! is_object( $global ) || ! get_class( $global ) || ! is_a( $global, 'wpdb' ) ) {
-				continue;
-			}
+		if ( ! isset( $GLOBALS['wpdb'] ) || ! property_exists( $GLOBALS['wpdb'], 'db_connections' ) ) {
+			return;
+		}
 
-			if ( ! property_exists( $global, 'db_connections' ) ) {
-				break;
-			}
+		$db_connections = $GLOBALS['wpdb']->db_connections;
+		if ( ! is_array( $db_connections ) || empty( $db_connections ) ) {
+			return;
+		}
 
-			if ( is_array( $global->db_connections ) && ! empty( $global->db_connections ) ) {
-				$elapsed = 0;
+		$elapsed = 0;
+		foreach ( $db_connections as $conn ) {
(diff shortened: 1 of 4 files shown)


qm-plugins/qm-db-connections/class-qm-db-connections-collector.php
```diff
@@ -0,0 +1,45 @@
...
+	 */
+	public function name() {
+		return esc_html__( 'DB Connections', 'query-monitor' );
+	}
+
+	/**
+	 * @return void
+	 */
+	public function process() {
+		foreach ( $GLOBALS as $global ) {
```
rinatkhaziev: I know it's a copy-pasta but this to me reads weird? It seems our intent is to check the global $wpdb instance, why don't we do just that rather than iterating over all globals.

Seems like a single  `isset( $GLOBALS['wpdb'] ) && property_exists( $GLOBALS['wpdb'] ,'db_connections' )` would've been just fine?

qm-plugins/qm-db-connections/qm-db-connections.php
```diff
@@ -0,0 +1,34 @@
...
+
+add_action( 'plugins_loaded', 'register_qm_db_connections' );
+function register_qm_db_connections() {
+	if ( ! class_exists( 'QM_Collectors' ) ) {
+		return;
+	}
+
+	require_once __DIR__ . '/class-qm-db-connections-collector.php';
+
+	QM_Collectors::add( new QM_DB_Connections_Collector() );
```
rinatkhaziev, line 22: Pardon the ignorance, Is this idiomatic for a QM_Collector to be added even though we may not know that a user is logged in? In other words, do we need to check here if a user is logged in as to not add any extra overhead for no benefit?

rebeccahum: You know what. I think we could actually move the entire block https://github.com/Automattic/vip-go-mu-plugins/blob/90bd992af476e75c21cc075c012c031f79aeafd7/query-monitor.php#L157-L168 into `wpcom_vip_qm_require()` so that the capability filter can catch it at https://github.com/Automattic/vip-go-mu-plugins/blob/90bd992af476e75c21cc075c012c031f79aeafd7/query-monitor.php#L84-L86

rinatkhaziev (approved):
Code looks good, but since we've changed the bootstrap let's indicate that in PR title and description.

PR #4039, opened 2023-01-10
Unregister JP-related cron events when disabled

WPprodigy:
## Description
Remove JP/VP cron events when they are disabled. Runs on just `cli_init` as we don't need these hooks except during cron execution.

### Cleanup Jetpack-related cron events when JP is disabled

When Jetpack is disabled, make sure all the cron events are also disabled/removed.

## Checklist

## Steps to Test

1. Run `wp cron-control events list`, and note the events.
2. With this code and JP disabled, the events will unregister themselves as they run.

### DIFF
--- a/cron/cron.php
+++ b/cron/cron.php
@@ -11,3 +11,31 @@ if ( file_exists( __DIR__ . '/action-scheduler-dynamic-queue.php' ) ) {
 		( new Action_Scheduler_Dynamic_Queue() )->init();
 	}, 9 );
 }
+
+// Unregister Jetpack-related cron events when disabled.
+add_action( 'cli_init', function() {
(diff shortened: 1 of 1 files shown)


cron/cron.php
```diff
@@ -11,3 +11,31 @@
...
+		] );
+	}
+
+	if ( $vaultpress_is_disabled ) {
+		$events_to_unregister = array_merge( $events_to_unregister, [ 'vp_scan_site', 'vp_scan_next_batch' ] );
+	}
+
+	// When/if the cron event hook runs, unschedule the event so it runs no more.
+	foreach ( $events_to_unregister as $event_hook ) {
+		add_action( $event_hook, function() use ( $event_hook ) {
```
rinatkhaziev: Minor but we can do:		`add_action( $event_hook, fn() =>  wp_unschedule_hook( $event_hook ) );`

PR #4042, opened 2023-01-11
Remove versioned WP-Parse.ly (part of ext now) and add unversioned as a submodule

rinatkhaziev:
## Description

Now that we moved Parsely to the [ext](https://github.com/Automattic/vip-go-mu-plugins-ext) repo we should clean it up from this one.

This PR removes all versioned copies and adds a submodule as a fallback.

This may break tests and also we need to modify the loader to include the fallback to minimize potential breakage on customer CI side (integration tests and workflows).

## Checklist


rebeccahum (commented):
Do we need any kind of version gating like we do in JP? https://github.com/Automattic/vip-go-mu-plugins/blob/ba26d4502b8f658526af3f7c6634bcb86bd74dea/jetpack.php#L20-L34

rinatkhaziev:
@rebeccahum I'm not sure, I don't remember an instance of WP-Parsely bumping the WP version requirements but I think @acicovic would be better equipped to answer that.

acicovic:
Since wp-parsely 3.0 we [require WP version 5.0](https://github.com/Parsely/wp-parsely/blob/3.0.0/wp-parsely.php) and this has not changed. The wp-parsely versions we have here are from 3.1 to 3.5.

rebeccahum (dismissed):

rebeccahum (dismissed):
@rinatkhaziev This doesn't work because it's expecting a versioned folder: https://github.com/Automattic/vip-go-mu-plugins/blob/da2ceb9444f0a1963a525da260b133d6d140623f/wp-parsely.php#L195

We'll have to modify the logic to account for it being `wp-parsely` or rename the submodule.

rebeccahum:
Testing steps:
1) Add `add_filter( 'wpvip_parsely_load_mu', '__return_true' );` to `plugin-loader.php`
2) Check out PR and do not include `vip-go-mu-plugins-ext` in mu-plugins
3) Go to wp-admin and expect to see notice `The Parse.ly plugin is not active. You need to [provide your Parse.ly Dash Site ID](http://es-site.vipdev.lndo.site/wp-admin/options-general.php?page=parsely) before things get cooking.` that Parsely plugin has been loaded

rinatkhaziev:
@acicovic can you take a look at the modified loader and think of any scenario this may break?

wp-parsely.php
```diff
@@ -191,30 +191,35 @@ function maybe_load_plugin() {
...
+		$integration_type = Parsely_Loader_Info::INTEGRATION_TYPE_MUPLUGINS_SILENT;
+	}
 
-		// Require VIP's customizations over wp-parsely.
-		$vip_parsely_plugin = __DIR__ . '/vip-parsely/vip-parsely.php';
-		if ( is_readable( $vip_parsely_plugin ) ) {
-			require_once $vip_parsely_plugin;
-		}
+	// Require the actual wp-parsely plugin.
+	require_once $entry_file;
```
rinatkhaziev, line 217: Does this need a file_exists check just to be extra safe?

rebeccahum: @rinatkhaziev IMHO, that seems to be excessive given the `is_readable()` check up above.

acicovic:
What happens if we're pinned to a specific wp-parsely version, using the `wpvip_parsely_version` filter?

```php
add_filter(
    'wpvip_parsely_version',
    function() {
        return '3.1';
    } 
);
```
Users use this to pin to old releases or test new ones.
Tip: If you go to Settings->Parse.ly, in the WP dashboard, the active plugin version is displayed on the top left of the page.

wp-parsely.php
rinatkhaziev: but we're only calling `is_readable`for the list of versioned files, not the fallback `$entry_file`

rebeccahum: Ah, good catch.

wp-parsely.php
```diff
@@ -211,7 +211,9 @@ function maybe_load_plugin() {
 	}
 
 	// Require the actual wp-parsely plugin.
-	require_once $entry_file;
+	if ( is_readable( $entry_file ) ) {
+		require_once $entry_file;
+	}
```
rinatkhaziev, line 216: Maybe flip around? If not readable then return early, otherwise load and let the active versions and whatnot

rebeccahum:
@acicovic It will still fall back to the default version if it cannot find that set version. But the pinning won't be successful (given the fact the version won't exist)

rinatkhaziev:
LGTM but since I created the PR I can't approve it :) @Automattic/vip-platform-cantina

acicovic (approved):

acicovic:
I approved but apparently this won't work as we need someone from the Cantina team :)

rebeccahum (approved):

PR #4044, opened 2023-01-12
Disable Contact Form 7 upload cleanup if it ignores WPCF7_UPLOADS_TMP…

rickhurst:
## Description
TL;DR this stops Contact Form 7 attempting to open a directory within wp-content, which raises excessive `opendir() Failed to open directory` warnings.

This PR addresses a situation where recent versions of the Contact Form 7 (CF7) enforces a policy of `WPCF7_UPLOADS_TMP_DIR` being located somewhere within `wp-content`, which isn't compatible with VIP platform. CF7 has a method `wpcf7_upload_tmp_dir()` which uses the value of `WPCF7_UPLOADS_TMP_DIR` but expects it to be within the `wp-content` directory, and ultimately returns a path within `wp-content/uploads/` on VIP Platform. 

The current version of the plugin (5.7.2 at time of writing) does not currently work for attachments, without patching - this PR _does not_ address that.

The reason for this PR is that a method run on shutdown of every request attempts to run a method `wpcf7_cleanup_upload_files()`, which tries to run `opendir()` in a sub-folder of `wp-content`. This generates excessive `opendir() Failed to open directory` warnings.

### Plugin fixes updated to include a workaround for Contact Form 7 upload file cleanup

An action was added to conditionally disable the `wpcf7_cleanup_upload_files` shutdown action, if the tmp directory path returned by `wpcf7_upload_tmp_dir()` does not match the value in `WPCF7_UPLOADS_TMP_DIR`.

## Checklist

## Steps to Test

1. Install a recent Contact Form 7 (e.g. 5.7.2) on a VIP test site
1. Visit the test site and request some pages
1. Check PHP logs for `opendir() Failed to open directory` warnings.
1. Checkout this PR and switch the site to use the appropriate version of mu-plugins
1. Visit the test site and request some pages
1. Check that there are no longer`opendir() Failed to open directory` warnings being triggered.


plugin-fixes.php
```diff
@@ -18,6 +18,29 @@
...
+function vip_disable_wpcf7_cleanup_upload_files() {
+	// Return early if the relevant functions do not exist
+	if ( ! function_exists( 'wpcf7_cleanup_upload_files' ) || ! function_exists( 'wpcf7_upload_tmp_dir' ) ) {
+		return;
+	}
+
+	// Check if the action is queued and if the tmp directories match
+	// (they might match if the plugin is an old version, or has been patched)
+	$priority = has_action( 'shutdown', 'wpcf7_cleanup_upload_files' );
+	if ( false !== $priority && WPCF7_UPLOADS_TMP_DIR !== wpcf7_upload_tmp_dir() ) {
```
WPprodigy, line 38: For the sake of extra safety, let's make sure the `WPCF7_UPLOADS_TMP_DIR` constant is defined to avoid potential future PHP errors.

WPprodigy:
Where do CF7-uploaded files currently go? Do we need to make sure they are cleaned up still somehow?

plugin-fixes.php
rinatkhaziev: +1

rinatkhaziev (dismissed):
Please see Caleb's comment.

plugin-fixes.php
rickhurst: I will add this, though it is defined a few lines above :)

https://github.com/Automattic/vip-go-mu-plugins/blob/90bd992af476e75c21cc075c012c031f79aeafd7/plugin-fixes.php#L17-L19

rickhurst:
> Where do CF7-uploaded files currently go? Do we need to make sure they are cleaned up still somehow? 

If the Plugin has been patched (our only current workaround) to allow uploads somewhere other than within `wp-content`, they go into `/tmp/CF7/`:

https://github.com/Automattic/vip-go-mu-plugins/blob/90bd992af476e75c21cc075c012c031f79aeafd7/plugin-fixes.php#L17-L19

The clean-up will also run in this directory with my suggested action, as long as `WPCF7_UPLOADS_TMP_DIR` matches `wpcf7_upload_tmp_dir()`, which will only happen for older versions of the plugin, or a patched plugin.

If the plugin isn't old or patched, uploads do not work, so there is nothing to clean up.

Note that I have reached out to the plugin author directly and [here](https://wordpress.org/support/topic/feature-request-optional-exception-to-is_file_path_in_content_dir/) to see if we can create a workaround where we do not need to patch the plugin for attachments to work.

rinatkhaziev (approved):

PR #4048, opened 2023-01-12
Debug Bar: Only trigger deprecation once every 10 minutes and on admin

rebeccahum:
## Description
For less noisy logs

## Checklist

## Steps to Test
1) Force enable debug bar to be true `add_filter( 'debug_bar_enable', '__return_true' );`
2) Go to wp-admin and see warning in logs
3) Go to another wp-admin page and do not see warning appear again in log
4) `wp cache delete debug_bar_deprecation_warning vip`
5) Repeat step 2

### DIFF
--- a/debug-bar.php
+++ b/debug-bar.php
@@ -17,7 +17,14 @@ add_action( 'set_current_user', function() {
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


WPprodigy (approved):
Usually would suggest wp_cache_add() instead of the get+set, but in this case probably doesn't really matter.

PR #4050, opened 2023-01-13
Search: Change frequency of cron jobs that run maybe_alert_for_field_count and maybe_alert_for_average_queue_time

rebeccahum:
## Description
This PR changes the frequency for two cron jobs currently hooked onto `vip_config_sync_cron`, which runs every 5 minutes which is too frequent for these jobs.
- maybe_alert_for_field_count will run once a week
- maybe_alert_for_average_queue_time will run hourly

They are both hooked onto existing cron jobs since Caleb has been doing work around removing cron jobs and I don't want to add to that pile.

Change frequency of cron jobs that run maybe_alert_for_field_count and maybe_alert_for_average_queue_time

## Checklist

### DIFF
--- a/search/includes/classes/class-queuewaittimejob.php
+++ b/search/includes/classes/class-queuewaittimejob.php
@@ -2,14 +2,12 @@
 
 namespace Automattic\VIP\Search;
 
+require_once __DIR__ . '/class-settingshealthjob.php';
+
 class QueueWaitTimeJob {
-	const CRON_EVENT_NAME = 'vip_config_sync_cron';
-
-	const FIELD_COUNT_GAUGE_DISABLED_SITES = array();
-
 	public function init() {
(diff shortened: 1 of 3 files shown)


search/includes/classes/class-queuewaittimejob.php
```diff
@@ -2,14 +2,12 @@
 
 namespace Automattic\VIP\Search;
 
-class QueueWaitTimeJob {
-	const CRON_EVENT_NAME = 'vip_config_sync_cron';
-
-	const FIELD_COUNT_GAUGE_DISABLED_SITES = array();
```
rebeccahum, line 8: Removed `FIELD_COUNT_GAUGE_DISABLED_SITES` since it doesn't seem to be used anywhere.

WPprodigy (approved):

PR #4058, opened 2023-01-17
ci: Update E2E workflow

sjinks:
Update E2E workflow:
* install latest version of VIP CLI (so that we don't have to modify the workflow when new VIP CLI gets released);
* remove Cypress-related endpoints from the list of allowed outgoing connections;
* update the list of the allowed endpoints according to the [recommendations](https://app.stepsecurity.io/github/Automattic/vip-go-mu-plugins/actions/runs/3945923788).

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -43,14 +43,15 @@ jobs:
             1.gravatar.com:80
             2.gravatar.com:80
             api.github.com:443
+            api.service.runforesight.com:443
             api.wordpress.org:443
             artifactcache.actions.githubusercontent.com:443
             auth.docker.io:443
-            codeload.github.com:443
             dl-cdn.alpinelinux.org:443
-            dl-cdn.alpinelinux.org:80
             downloads.wordpress.org:443
-            frsnacprodeus2file1.blob.core.windows.net:443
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e.yml
```diff
@@ -62,27 +64,10 @@ jobs:
...
-            e2e-test-site.vipdev.lndo.site:80
-            playwright-akamai.azureedge.net:443
-            playwright-verizon.azureedge.net:443
-            playwright.azureedge.net:443
-            ppa.launchpad.net:80
-            ppa.launchpad.net:443
-            azure.archive.ubuntu.com:80
-            azure.archive.ubuntu.com:443
-            packages.microsoft.com:443
-            download.cypress.io:443
```
rebeccahum, line 79: We don't need the Cypress anymore for https://github.com/Automattic/vip-go-mu-plugins/blob/develop/.github/workflows/search-e2e.yml?

sjinks: Why, that's a different workflow.

I moved all Playwright-related stuff into a dedicated directory in #3799 so that we don't have to install Cypress and company when running Playwright tests.

And now that this workflow does not try to install Cypress, we don't need those endpoints anymore.

rebeccahum (approved):

PR #4062, opened 2023-01-17
maybe_load_plugin: Remove nullable return type for $filtered_load_status

rebeccahum:
## Description
Since we're checking if `$filtered_load_status` is a boolean @ https://github.com/Automattic/vip-go-mu-plugins/blob/731bd7db3abbd1237eebb7f3e781371f1d6f4352/wp-parsely.php#L158-L159

TODO: Update https://docs.wpvip.com/technical-references/plugins/parse-ly/#h-enable-only-on-selected-network-sites after

### Plugin Updated: VIP Parse.ly Integration

maybe_load_plugin: Remove nullable return for $filtered_load_status

## Checklist

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -152,8 +152,8 @@ function maybe_load_plugin() {
 		return;
 	}
 
-	$option_load_status   = get_option( '_wpvip_parsely_mu', null );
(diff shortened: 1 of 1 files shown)


rebeccahum:
Wanted to note, closing because the `null` value is needed to know if a filter was applied or not.

PR #4079, opened 2023-01-25
Wp-admin: Add notice for Search customers without the constant

rebeccahum:
## Description
Since March 1 will be here before we know it, it would be good to display a notice.
Screenshot 2023-01-25 at 3 11 28 PM

### Updated: Wp-admin

Add reminder to test out Enterprise Search upgrade 

## Checklist

## Steps to Test
1) Define search constant:
```
define( 'VIP_ENABLE_VIP_SEARCH', true ); // Enables Enterprise Search
```
2) Hit up wp-admin with super admin
3) Expect to see notice
4) Add testing constant:
```
define( 'VIP_SEARCH_USE_NEXT_EP', true );
```
5) Repeat step 2 
6) Expect to see notice gone


GaryJones:
Since admin notices can come from different sources (plugins), then perhaps make it clearer that this is related to WPVIP / the platform, so folks don't just dismiss it as unimportant?

Also, who is this targeted at? Seems like everyone? Which means a lot of authors, editors and other roles that can't take any action, or won't know what Enterprise Search is, will also be seeing it (as noise).

rebeccahum:
@GaryJones This only displays it for Enterprise Search customers and users with a super admin. However, I've amended the messaging to indicate that this is on WordPress VIP.

rinatkhaziev:
> Since admin notices can come from different sources (plugins), then perhaps make it clearer that this is related to WPVIP / the platform, so folks don't just dismiss it as unimportant?

One would hope that that someone who has explicitly Enterprise Search constant set in the code is well-aware what Enterprise Search is enabled for their site?

GaryJones:
If they are a developer, sure, but not all site/super-admins are developers.

admin-notice/admin-notice.php
```diff
@@ -47,3 +47,23 @@ function( $admin_notice_controller ) {
 		);
 	}
 );
+
+// Enterprise Search upgrade
+add_action(
+	'vip_admin_notice_init',
+	function( $admin_notice_controller ) {
+		global $wp_version;
```
GaryJones: Why is this line needed?

rebeccahum: Oops. Bad copy pasta.

rinatkhaziev (approved):

PR #4082, opened 2023-01-30
Query Monitor: Revise Object Cache to play with get_stats()

rebeccahum:
## Description
This only works once https://github.com/Automattic/wp-memcached/pull/127 has been deployed. Otherwise, it will fall back onto the old method if `get_stats()` doesn't exist. 
Screenshot 2023-01-30 at 10 19 08 AM

This PR plays nice with dark mode and adds a "Operations" sub-panel which allows for filtering and sorting of Memcache operations. 
Screenshot 2023-01-31 at 11 08 51 AM

The main panel will display a summary of the statistics.

Re-factored Object Cache panel to play nice with dark mode and sort/filter operations

## Checklist

## Steps to Test
1) Add https://github.com/Automattic/wp-memcached/pull/127 as drop-in
2) Go to Query Monitor > Object Cache and Query Monitor > Object Cache > Operations to see it matches screenshot. 
3) In Operations sub-panel, play around with sorting and filtering.
4) Revert step 1) to drop-in and expect it to still work as it did before


qm-plugins/qm-object-cache/css/style.css
```diff
@@ -1,64 +1,69 @@
...
-	clear: none !important;
-	min-width: 150px !important;
-	border: 1px solid #eee !important;
-	padding: 5px 10px 15px !important;
-	text-align: center !important;
-	color: #000 !important;
-	font-family: georgia,times,serif !important;
-	font-size: 28px !important;
-	margin: 15px 10px 15px 0 !important;
+  float: left  !important;
```
GaryJones: The WordPress Coding Style is to use tab indentation for CSS files. See https://core.trac.wordpress.org/browser/trunk/.editorconfig#L14

(I'm surprised we don't have a `.editorconfig` file in the root of this repo to help catch this.)

qm-plugins/qm-object-cache/html/class-qm-output-object-cache-ops.php
```diff
@@ -0,0 +1,156 @@
...
+		$this->before_tabular_output();
+
+		echo '<thead>';
+		echo '<tr>';
+		$this->output_filterable_table_col( 'Operation', array_keys( $ops ) );
+		$this->output_sortable_table_col( 'Key' );
+		$this->output_sortable_table_col( 'Size' );
+		$this->output_sortable_table_col( 'Time' );
+		$this->output_filterable_table_col( 'Group', array_keys( $groups ) );
+		$this->output_sortable_table_col( 'Result' );
```
GaryJones: Should these column header strings be made translatable the same as some of the menu/title strings?

qm-plugins/qm-object-cache/html/class-qm-output-object-cache-ops.php
```diff
@@ -0,0 +1,156 @@
...
+				$result = 'Not in Memcache';
+				break;
+			case 'memcache':
+				$result = 'Found in Memcache';
+				break;
+			case '[mc already]':
+				$result = 'Already in Memcache';
+				break;
+			case '[lc already]':
+				$result = 'Local cache already';
```
GaryJones: Make strings translatable?

Also, since this is custom code, and not part of QM directly, then all the translatable strings should be custom e.g. `vip-qm-object-cache` or similar.

qm-plugins/qm-object-cache/html/class-qm-output-object-cache-ops.php
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	/**
+	 * Outputs the table column header
+	 *
+	 * @param $title string
+	 */
+	public function output_sortable_table_col( $title ) {
+		echo '<th scope="col" class="qm-sortable-column" role="columnheader">';
+		echo $this->build_sorter( esc_html__( $title, 'query-monitor' ) ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped, WordPress.WP.I18n.NonSingularStringLiteralText
```
GaryJones: `esc_html__()` is trying to do a translatable on a PHP variable; it doesn't work like that. The translation should be at the first point the string is defined, and only the escaping happens here.

That means you can likely get rid of one of those PHPCS ignore violations codes as well.

qm-plugins/qm-object-cache/html/class-qm-output-object-cache-ops.php
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	/**
+	 * Outputs the table column header
+	 *
+	 * @param $title string
+	 */
+	public function output_filterable_table_col( $title, $values, $args = [] ) {
+		echo '<th scope="col" class="qm-filterable-column">';
+		echo $this->build_filter( sanitize_title( strtolower( $title ) ), $values, esc_html__( $title, 'query-monitor' ), $args ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped, WordPress.WP.I18n.NonSingularStringLiteralText
```
GaryJones: As above - trying to translate a non-literal string.

qm-plugins/qm-object-cache/html/class-qm-output-object-cache.php
```diff
@@ -0,0 +1,117 @@
...
+		echo '</tr>';
+	}
+
+	public function output_before_section( $heading ) {
+		echo '<section>';
+		if ( $heading ) {
+			echo '<h3>' . esc_html( $heading ) . '</h3>';
+		}
+		echo '<table>';
+		echo '<thead class="qm-screen-reader-text">';
```
GaryJones, line 115: I came here to make a comment about the table not appearing to have table headings which screen reader software would need, and you've already handled it 👏️

qm-plugins/qm-object-cache/html/class-qm-output-object-cache.php
```diff
@@ -0,0 +1,117 @@
...
+				$this->output_table_row( 'Size', size_format( $totals['size'], 2 ) );
+			}
+			$this->output_after_section();
+		}
+
+		$stats = $data['stats'] ?? false;
+		if ( $stats ) {
+			$this->output_before_section( false );
+			foreach ( $stats as $stat => $value ) {
+				$this->output_table_row( $stat, $value );
```
GaryJones: Would be nice to have these `$value` numbers formatted to include the locale's thousand separators, etc.

rebeccahum: I believe it is later down on L49 via `number_format_i18n()`?

GaryJones: That's for the Query Time value in the first table. This line (L61) is for the numerical outputs in the second table (get, update, etc.), which isn't formatted.

The screenshot included in the PR description even shows `1888` unformatted.

qm-plugins/qm-object-cache/html/class-qm-output-object-cache-ops.php
```diff
@@ -139,7 +139,7 @@ public function process_result( $result ) {
 	 */
 	public function output_sortable_table_col( $title ) {
```
GaryJones: Should be able to add some strong types to these method parameters.

qm-plugins/qm-object-cache/css/style.css
```diff
@@ -1,69 +1,69 @@
 #query-monitor-main .qm-hide-operation, #query-monitor-main .qm-hide-group {
-    display: none !important;
-  }
+				display: none !important;
+		}
```
GaryJones: Looks like each one space was replaced with one tab, which now means this is indented 4 tabs in (should be 1) and two tabs (should be 0).

The other lines are affected too.

qm-plugins/qm-object-cache/html/class-qm-output-object-cache.php
rebeccahum: @GaryJones If you're referring to the `1888` value for `get_local`, those are the counts and thus, no formatting is needed.

GaryJones: If a count is 4324987419041510 then it's going to be hard to read. Hence, formatting the number to use thousand-separators will make it easier.

qm-plugins/qm-object-cache/html/class-qm-output-object-cache-ops.php
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	/**
+	 * Returns in human readable result.
+	 *
+	 * @param $result  string Memcached operation result
+	 * @return $result string Human readable memcached operation result
+	 */
+	public function process_result( string $result ) {
+		switch ( trim( $result ) ) {
```
rinatkhaziev, line 135: I'm not sure how accurate this is going to be, cc @WPprodigy

rinatkhaziev (dismissed):

qm-plugins/qm-object-cache/html/class-qm-output-object-cache-ops.php
WPprodigy: The description / rephrasing here is accurate. There's a couple types of operations where we don't leave these comments consistently, but that can be solved later on.

qm-plugins/qm-object-cache/html/class-qm-output-object-cache-ops.php
```diff
@@ -0,0 +1,212 @@
...
+	 *
+	 * @param array $array Array to be outputted in table cell
+	 */
+	public function maybe_output_toggle_table_cell_for_array( array $array ) {
+		if ( empty( $array ) ) {
+			return;
+		}
+
+		if ( count( $array ) === 1 ) {
+			echo '<td class="qm-nowrap qm-ltr">' . esc_html( $array[0] ) . '</td>';
```
WPprodigy: Maybe a bit more DRY to call `$this->output_table_cell()` inside this conditional, but not a big deal.

qm-plugins/qm-object-cache/html/class-qm-output-object-cache-ops.php
```diff
@@ -0,0 +1,212 @@
...
+			return;
+		}
+
+		echo '<td class="qm-nowrap qm-ltr qm-has-toggle">';
+		echo static::build_toggler(); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+		echo '<ol><li>' . esc_html( $array[0] ) . '...</li>';
```
WPprodigy:
```suggestion
		echo '<ol><li>' . esc_html( $array[0] ) . ' [+' . ( count( $array ) - 1 ) . ']</li>';
```

qm-plugins/qm-object-cache/html/class-qm-output-object-cache-slow-ops.php
```diff
@@ -0,0 +1,189 @@
...
+		$this->output_sortable_table_col( __( 'Backtrace', 'qm-object-cache' ) );
+		echo '</tr>';
+		echo '</thead>';
+
+		echo '<tbody>';
+		foreach ( $ops as $op_name => $data ) {
+			foreach ( $data as $op ) {
+				echo '<tr data-qm-operation="' . esc_attr( $op_name ) . '" data-qm-group="' . esc_attr( $op['group'] ) . '">';
+				echo '<td class="qm-nowrap qm-ltr">' . esc_html( $op_name ) . '</td>';
+				echo '<td class="qm-nowrap qm-ltr">' . esc_html( $op['key'] ) . '</td>';
```
WPprodigy: Slow-ops will be able to have an array of keys here as well, so need to support the same type of drop-down thingy as the ops panel. Currently `get_multi()` is excluded from slow_ops, but the new drop-in won't exclude get_multiple().

ATM, this will end up being an array to string conversion / just displaying `Array`.

WPprodigy (changes requested):

WPprodigy (approved):

PR #4109, opened 2023-02-08
Prometheus: re-add the dedicated endpoint with better locking mechanism, introduce Error stats collector

rinatkhaziev:
## Description

This is a significant re-work of #3989, it contains the following changes:

1. Collector interface is amended to include a new `process_metrics` method. All collectors had that implemented.
- The purpose of the new method is to do any potentially heavy computations off the web request path and store it into temporary storage, so that `collect_metrics` can grab that data without having to waste time. See Post Stats Collector for example implementation.
- A new cron event is added `vip_prom_process_metrics`. It's get scheduled at around ~10 min interval (as a single event). This kicks off the aforementioned `process_metrics` for each collector.
- On the web request's shutdown on authenticated requests (might change in the future depending on initial rollout results) every 90 seconds we'll collect the metrics and put it into APCu for the scraper to collect.
2. Error Stats collector is introduced, to report the number of runtime error per network site and error type
- To achieve that our current error handler adds a new action `php_error_handler`. `do_action` gets called if it's defined, Error Stats Collector hooks onto that to increment the relevant counter.

## Checklist

## Steps to Test

1. Check out the PR
4. For ease of testing, comment out the current_user_can. 
5. Add an error_log inside if `( wp_cache_add( 'vip_prometheus_last_collection_run', time(), 'vip_prometheus', 5 * MINUTE_IN_SECONDS ) ) {`
6. Comment out the auth check
7. Kick off something like `ab -c 10 -n 300 http://site.vipdev.lndo.site`
8. Make sure there's just one message logged (i.e. the locking works)
9. On the v2.27.0 of VIP-CLI, visit `http://site.vipdev.lndo.site/.vip-prom-metrics` and verify it's rendered correctly.
10. Add some test code like `trigger_error( 'My test warning', E_USER_WARNING );`, verify that `error_count` is present in the prev step.
11. Disable the Search collector (by nerfing the method) and verify Search is working in all contexts and there are no fatals.


prometheus/inc/class-collectorinterface.php
```diff
@@ -13,4 +13,9 @@ public function initialize( RegistryInterface $registry ): void;
 	 * This can be useful, for, for example, gauges which measure something external wrt the application (e.g., APC cache stats)
 	 */
 	public function collect_metrics(): void;
+
+	/**
+	 * Process metrics off the web request path, this is useful for metrics that are expensive to calculate on every request.
+	 */
+	public function process_metrics(): void;
```
rinatkhaziev: On a second thought this may be safer to exclude from the initial rollout just to avoid the deploy thing.

prometheus/endpoint.php
```diff
@@ -0,0 +1,43 @@
...
+use Prometheus\CollectorRegistry;
+use Prometheus\RegistryInterface;
+use Prometheus\RenderTextFormat;
+use Prometheus\Storage\Adapter;
+use Prometheus\Storage\APCng;
+use Prometheus\Storage\InMemory;
+
+require_once __DIR__ . '/vendor/autoload.php';
+require_once dirname( __DIR__ ) . '/lib/utils/class-context.php';
+// @codeCoverageIgnoreStart -- APCu may or may not be available during tests
```
sjinks: I guess this comment needs to be amended :-)

sjinks (dismissed):

lib/wpcom-error-handler/wpcom-error-handler.php
```diff
@@ -208,6 +208,20 @@ function wpcom_custom_error_handler( $whether_i_may_die, $type, $message, $file,
 		error_log( sprintf( '%s: %s in %s on line %d [%s] [%s]', $string, $message, $file, $line, $source, $backtrace ) );
 	}
 
+	if ( function_exists( 'do_action' ) ) {
+		/**
+		 * Fires when a PHP error occurs.
+		 *
+		 * @since 2.1.0
```
sjinks: What is 2.1.0? :-) I mean, how do we version mu-plugins?

prometheus.php
```diff
@@ -1,37 +1,47 @@
...
+		$files[] = '/prometheus-collectors/class-post-stats-collector.php';
+	}
+
 	foreach ( $files as $file ) {
 		if ( ! file_exists( __DIR__ . $file ) ) {
-			return; // Bail early if one of the files doesn't exist.
+			continue; // Skip to the next one
 		} else {
 			require_once __DIR__ . $file;
 		}
```
sjinks: Maybe 

```php
if ( file_exists( __DIR__ . $file ) ) {
    require_once __DIR__ . $file;
}
```
?

prometheus.php
```diff
@@ -1,37 +1,47 @@
...
-			$collectors[] = new Login_Stats_Collector();
-			$collectors[] = new Post_Stats_Collector();
+			$collectors['cache']   = new Cache_Collector();
+			$collectors['apcu']    = new APCu_Collector();
+			$collectors['opcache'] = new OpCache_Collector();
+			$collectors['login']   = new Login_Stats_Collector();
+			if ( class_exists( Post_Stats_Collector::class ) ) {
+				$collectors['post'] = new Post_Stats_Collector();
+			}
+			$collectors['error'] = new Error_Stats_Collector();
```
sjinks: If we do not bail when one of the file does not exist, we have to guard every `new` with `if ( class_exists() )`.

lib/wpcom-error-handler/wpcom-error-handler.php
rinatkhaziev: a copy-pasta 🍝

prometheus.php
rinatkhaziev: yeah I was conflicted about this, maybe something like

```php
$c = [
 $slug => $class
];

foreach( $c as $slug => $class ) {
  if ( class_exists() ) {
     $collectors[ $slug ] => new $class();
  }
}

File: prometheus.php (hunk 4 above)
Comment (rinatkhaziev) [reply prometheus.php]: yes good point.

File: prometheus.php (hunk 5 above)
Comment (rinatkhaziev) [reply prometheus.php]: Could do more OOP and add a `addCollector` or something

**Review by sjinks — DISMISSED** [review]

**Comment by rinatkhaziev** [comment]
This should probably go into March 14 staging release so I slapped `Do not deploy` in the mean time.

File: search/includes/classes/class-search.php [hunk 6]
```diff
@@ -1164,9 +1162,10 @@ public function ep_handle_failed_request( $request, $response, $query, $type, $c
 			$response_failure_code = $response->get_error_code();
 
 			foreach ( $error_messages as $error_message ) {
-				$reason = $this->is_curl_timeout( $error_message ) ? Prometheus_Collector::QUERY_FAILED_TIMEOUT : Prometheus_Collector::QUERY_FAILED_ERROR;
-
-				Prometheus_Collector::increment_failed_query_counter( $query_method, $query['url'], $reason );
+				if ( class_exists( Prometheus_Collector::class ) ) {
```
sjinks: I would move this check out of `foreach` loop: if the class does not exist, we don't need the loop.

rinatkhaziev: thanks, addressed

prometheus.php
```diff
@@ -1,37 +1,49 @@
...
-		'/prometheus/index.php',
 		'/prometheus-collectors/class-cache-collector.php',
 		'/prometheus-collectors/class-apcu-collector.php',
 		'/prometheus-collectors/class-opcache-collector.php',
 		'/prometheus-collectors/class-login-stats-collector.php',
-		'/prometheus-collectors/class-post-stats-collector.php',
+		'/prometheus-collectors/class-error-stats-collector.php',
 	];
 
+	$should_enable_post_collector = Feature::is_enabled( 'prom-post-collection' ) || ( defined( 'VIP_GO_ENV' ) && 'production' !== VIP_GO_ENV );
```
rebeccahum: Can we use the new `is_enabled_by_env()`? https://github.com/Automattic/vip-go-mu-plugins/blob/b7b33680a72eb2bcf30cda6eba76628be40ef803/lib/feature/class-feature.php#L85

Then we could just call `is_enabled()`: https://github.com/Automattic/vip-go-mu-plugins/blob/b7b33680a72eb2bcf30cda6eba76628be40ef803/lib/feature/class-feature.php#L57-L59

rinatkhaziev: oh yes we can, this PR was started so long ago. 

I also think there's a few IDs we'd have to force-disable out of abundance of caution.

rinatkhaziev: addressed

prometheus-collectors/class-cache-collector.php
```diff
@@ -59,13 +66,13 @@ public function shutdown(): void {
 		global $wp_object_cache;
 
 		if ( $this->cache_hits_counter ) {
-			$this->cache_hits_counter->incBy( $wp_object_cache->cache_hits, [ (string) get_current_blog_id() ] );
-			$this->cache_misses_counter->incBy( $wp_object_cache->cache_misses, [ (string) get_current_blog_id() ] );
+			$this->cache_hits_counter->incBy( $wp_object_cache->cache_hits, [ $this->blog_id ] );
+			$this->cache_misses_counter->incBy( $wp_object_cache->cache_misses, [ $this->blog_id ] );
```
WPprodigy, line 70: I'm not so sure if this will go well if there are `switch_to_blog()` calls happening. Kinda need to make sure we're the same blog_id as when we started by the time we get to `shutdown` I think?

prometheus/inc/class-plugin.php
```diff
@@ -157,14 +92,71 @@ private static function create_registry(): RegistryInterface {
...
+		// This is expensive, potentially, so be mindful about when this method is called
+		// Currently it only runs on web requests with a 90 interval, see the constructor
+		if ( function_exists( 'fastcgi_finish_request' ) ) {
+			fastcgi_finish_request();
+		}
+
+		// Don't run on unauthorized requests.
+		if ( ! function_exists( 'wp_get_current_user' ) || ! current_user_can( 'edit_posts' ) ) {
+			return;
+		}
```
WPprodigy: Not sure I understand the `! function_exists( 'wp_get_current_user' )`. And then for the latter, it's quite possible a request from a user with `edit_posts` permissions doesn't occur for a pretty long period of time, is that intended/okay to go so long without collecting?

prometheus/inc/class-plugin.php
```diff
@@ -157,14 +92,71 @@ private static function create_registry(): RegistryInterface {
...
+
+		if ( wp_cache_add( 'vip_prometheus_last_collection_run', time(), 'vip_prometheus', 5 * MINUTE_IN_SECONDS ) ) {
+			foreach ( $this->collectors as $collector ) {
+				$collector->collect_metrics();
+			}
+
+			// Roughly every hour
+			if ( ! wp_next_scheduled( 'vip_prometheus_process_metrics' ) ) {
+				wp_schedule_single_event( time() + HOUR_IN_SECONDS, 'vip_prometheus_process_metrics' );
+			}
```
WPprodigy: I don't love adding in a new cron job for this 😬  - it particularly really starts to inflate cron jobs for large multisites. Would really like to start pairing up our internal jobs together on shared hooks.

prometheus/inc/class-plugin.php
rinatkhaziev: I don't remember in which context exactly but I got `wp_get_current_user` fatal so I added this check. 

`collect_metrics` is something that grabs the result of computationally-intensive of `process_metrics` (which happens off request path`. It doesn't really matter when it's exactly called, see the Posts Stats collector for the example implementation.

rinatkhaziev: The intent of this whole method is purely to limit the situations where metrics can be collected and when the single job event is scheduled.

prometheus/inc/class-plugin.php
rinatkhaziev: @WPprodigy  I don't love this either :)

Open to suggestions but I think here's what helps: the event is only scheduled when there's logged-in activity, so even on the large multisite the scenario where ALL bajillion sites are scheduling at the same time is of relatively low probability.

I'm open to suggestions though, alternative design would be to run on on a root site and do `switch_to_blog` but that may lead to some edge-cases or a long-running job that may be terminated by pod rolls.

prometheus-collectors/class-cache-collector.php
rinatkhaziev: This particular instance should probably be addressed with `Plugin::get_instance()->get_site_label()` but switch_to_blog would probably mess things up anyway.

rinatkhaziev: What would be the for sure way to tell `a switch_to_blog` happened?

rebeccahum: Mebbe:
```
if ( is_multisite() && $GLOBALS['switched'] ) {
restore_current_blog();
}
```

prometheus/inc/class-plugin.php
rinatkhaziev: Addressed

prometheus-collectors/class-cache-collector.php
rinatkhaziev: Addressed this by introducing Plugin::$site_label property which can only be set once.

WPprodigy (approved):

PR #4110, opened 2023-02-09
Skip loading Jetpack for FedRAMP sites by default

jblz:
## Description

In the `000-vip-init.php` file, when the `VIP_IS_FEDRAMP` constant is true, set the `VIP_JETPACK_SKIP_LOAD` constant to `true` as well.

## Checklist

## Steps to Test

1. Check out PR.
2. Test that Jetpack is enabled for non-FedRAMP sites
    * Browse to `http://$SITE_URL/index.php?rest_route=/jetpack/v4`
        * You should see a list of Jetpack REST API endpoints
    * Browse to `wp-admin` while logged in
        * You should see Jetpack settings in the menu & the options page should function as expected
3. Test that Jetpack is *NOT* enabled for FedRAMP sites
    * Add the following to your `vip-config/vip-config.php` file:
    `define( 'VIP_IS_FEDRAMP', true );`
    * Browse to `http://$SITE_URL/index.php?rest_route=/jetpack/v4`
        * You should see a `404` w/ `rest_no_route` code.
    * Browse to `wp-admin` while logged in
        * You should *NOT* see Jetpack settings in the menu & manually browsing to `http://$SITE_URL/wp-admin/admin.php?page=jetpack` should `403` with a "Sorry, you are not allowed to access this page." message.


rinatkhaziev (changes requested):
@jblz I think maybe rather than check for the constant like that we should make this a [Context](https://github.com/Automattic/vip-go-mu-plugins/blob/develop/lib/utils/class-context.php) helper? Like `Context::is_fedramp` or something.

The rationale is likely we'll re-use this check in many other parts of codebase (At least Parse.ly comes to mind)

sjinks:
@rinatkhaziev Done, please take a look

rinatkhaziev (approved):

jblz:
Thanks for the refactor @sjinks !

000-vip-init.php
```diff
@@ -128,6 +128,11 @@
 // Define constants for custom VIP Go paths
 define( 'WPCOM_VIP_CLIENT_MU_PLUGIN_DIR', WP_CONTENT_DIR . '/client-mu-plugins' );
 
+// FedRAMP sites do not load Jetpack by default
+if ( method_exists( Context::class, 'is_fedramp' ) && Context::is_fedramp() && ! defined( 'VIP_JETPACK_SKIP_LOAD' ) ) {
```
jblz, line 132: The class is probably as likely to be missing as the function, so I'd probably use `is_callable` instead of `method_exists`.

rinatkhaziev: No, class is definitely there. But lib/ will be copied over later than 000-vip-init

PR #4111, opened 2023-02-09
Revert "Reverting wp-cli-db"

rebeccahum:
Reverts Automattic/vip-go-mu-plugins#3683

Previously, this broke the `import` command on dev-env:
```
Uncaught Exception: The db command is not currently supported in this environment. in /wp/wp-content/mu-plugins/lib/helpers/wp-cli-db/class-config.php:38
```

But it looks like that's not the case anymore:
```
╰─$ vip dev-env --slug es-site import sql ~/Downloads/test.sql
...
Success: Database imported.
```

### Plugin added: WP-DB-CLI
Previously reverted, this reverts that change and adds back `/lib/helpers/wp-cli-db.php`

## Checklist

## Steps to Test

1. Check out PR.
2. On Local dev-env, attempt to import `vip dev-env import sql ~/Downloads/test.sql` and expect no exception

### DIFF
--- a/000-pre-vip-config/requires.php
+++ b/000-pre-vip-config/requires.php
@@ -17,7 +17,7 @@ $files = [
 ];
 
 $cli_files = [
-	// '/lib/helpers/wp-cli-db.php', - Reverting as it breaks dev-env import
+	'/lib/helpers/wp-cli-db.php',
 ];
 
 foreach ( $files as $file ) {
--- a/lib/helpers/wp-cli-db/class-config.php
+++ b/lib/helpers/wp-cli-db/class-config.php
@@ -9,9 +9,11 @@ class Config {
 	private bool $enabled      = false;
 	private bool $allow_writes = false;
 	private bool $is_sandbox   = false;
+	private bool $is_local     = false;
 
 	public function __construct() {
 		$this->is_sandbox   = class_exists( Environment::class ) && Environment::is_sandbox_container( gethostname(), [] );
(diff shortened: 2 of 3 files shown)


rinatkhaziev (changes requested):
We first should check there are no more envs being spun up on the VIP-CLI version that had the error. If there are we can't really merge this.

sjinks:
I offer that we add a check whether the code runs in a dev env (if I am not mistaken, we had a constant for that). If it does, we should not disable `wp db` commands, and import will work for those old clients.

rebeccahum:
The problem with that approach is that `000-pre-vip-config` is loaded before the constants are set, so we'd need to move the constants to load beforehand.  Uncertain if there would be a case where this would cause fatals where it was somehow being defined in vip-config without a proper check.

Edit: Maybe for dev-env, we can just move https://github.com/Automattic/vip-container-images/blob/a81001d59d292db332942cc15b0d7a7c3696c613/dev-tools/wp-config-defaults.php#L106-L110 upwards before pre-vip-config?

rinatkhaziev:
@rebeccahum I'd argue those 4 constants are internal so whoever is defining them in vip-config.php is "holding it wrong'

rinatkhaziev:
Well apparently there's are quite a few instances of defining  it, so maybe it's me who's holding it wrong 😅

rebeccahum:
OK, tested on VIP-CLI 2.11 and it works now.

Edit: The only downside where I can see it not working now is if they are running on an outdated version of dev-tools where the constants aren't available yet.

rinatkhaziev (approved):

PR #3645, opened 2022-09-13
Allow for running `wp db` CLI commands in a sandbox

sjinks:
## Description

This PR allows us to use all sorts of `wp db` commands in a sandboxed environment.

Omitting changelog description, as this feature is for internal use only.

## Checklist

Please make sure the items below have been covered before requesting a review:

- [x] This change works and has been tested locally (or has an appropriate fallback).
- [x] This change works and has been tested on a Go sandbox.
- [ ] This change has relevant unit tests (if applicable).
- [ ] This change has relevant documentation additions / updates (if applicable).
- [x] I've created a changelog description that aligns with the provided examples.

## Steps to Test

1. Check out this branch in a sandbox
2. Run `wp db cli` and make sure it works.

PR #4127, opened 2023-02-15
Take #3: WP_CLI_DB - Allow certain `wp db` subcommands

rebeccahum:
## Description
Previously, this was reverted in https://github.com/Automattic/vip-go-mu-plugins/pull/4125 due to database backup copies breaking due to the error `Uncaught Exception: The db command is not currently supported in this environment.`. This PR:
- adds the addition for a batch container property in `$is_batch`
- catches the Exception thrown from `get_database_server()` in `before_run_command()`
- changes `validate_subcommand()` to return a WP_Error instead of an exception
- adds `validate_query()` to check the query can be run
- adjusts + adds tests for all of the above 
- adds a few extra subcommands in addition to `query`: `columns`, `size`, `prefix`
- adds a `--read-write` flag for the `query` subcommand to enable writes on-demand basis

TODO: 
- Unblock subcommands at Parker level 

Previous reverts:
- https://github.com/Automattic/vip-go-mu-plugins/pull/3683
- https://github.com/Automattic/vip-go-mu-plugins/pull/4125

Attempted redos:
- https://github.com/Automattic/vip-go-mu-plugins/pull/4111
- https://github.com/Automattic/vip-go-mu-plugins/pull/3645

OG PR: https://github.com/Automattic/vip-go-mu-plugins/pull/3064

### Plugin Updated: WP_CLI_DB

Adds back WP_CLI_DB and accounts for both batch containers & dev-env environments

## Checklist

### DIFF
--- a/lib/helpers/wp-cli-db/class-wp-cli-db.php
+++ b/lib/helpers/wp-cli-db/class-wp-cli-db.php
@@ -72,9 +72,14 @@ class Wp_Cli_Db {
 	public function validate_query( string $query ): ?WP_Error {
 		$query = strtolower( $query );
 
-		if ( false !== strpos( $query, 'drop' ) ) {
-			return new \WP_Error( 'db-cli-disallowed-query', 'This query is disallowed.' );
+		$disallowed_syntax = [ 'drop', 'create' ];
+
+		foreach ( $disallowed_syntax as $syntax ) {
+			if ( false !== strpos( $query, $syntax ) ) {
+				return new \WP_Error( 'db-cli-disallowed-query', 'This query is disallowed.' );
+			}
 		}
+
 		return null;
 	}
 
--- a/lib/helpers/wp-cli-db/class-wp-cli-db.php
+++ b/lib/helpers/wp-cli-db/class-wp-cli-db.php
@@ -94,7 +94,7 @@ class Wp_Cli_Db {
 			return;
 		}
 
-		if ( $this->config->is_local() && $this->config->is_batch() ) {
+		if ( $this->config->is_local() ) {
 			return;
 		}
 
--- a/lib/helpers/wp-cli-db/class-wp-cli-db.php
+++ b/lib/helpers/wp-cli-db/class-wp-cli-db.php
@@ -22,11 +22,18 @@ class Wp_Cli_Db {
 			return;
 		}
 
-		if ( ! class_exists( 'WP_CLI' ) ) {
+		if ( ! class_exists( 'WP_CLI' ) || ! class_exists( 'DB_Command' ) ) {
 			return;
 		}
 
 		WP_CLI::add_hook( 'before_run_command', [ $this, 'before_run_command' ] );
+
+		$db_command_file = __DIR__ . '/class-vip-db-command.php';
+		if ( file_exists( $db_command_file ) ) {
+			require_once $db_command_file;
+		}
+
+		WP_CLI::add_command( 'db', 'VIP_DB_Command' );
 	}
 
 	/**
@@ -87,8 +94,9 @@ class Wp_Cli_Db {
 	 * Added to the WP_CLI `before_run_command` hook.
 	 *
 	 * @param array $command
+	 * @param array $assoc_args
 	 */
-	public function before_run_command( array $command ): void {
+	public function before_run_command( array $command, array $assoc_args = [] ): void {
 		if ( ! ( isset( $command[0] ) && 'db' === $command[0] ) ) {
 			// Don't do anything for any command other than `db`
 			return;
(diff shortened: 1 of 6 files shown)


lib/helpers/wp-cli-db/class-wp-cli-db.php
```diff
@@ -30,40 +31,74 @@ public function early_init() {
...
+	 * @param array $command
+	 * @return WP_Error|null
 	 */
-	public function validate_subcommand( array $command ): void {
+	public function validate_subcommand( array $command ): ?WP_Error {
 		$subcommand = $command[1] ?? '';
 
 		if ( 'query' !== $subcommand ) {
-			throw new Exception( 'ERROR: Only the `wp db query` subcommand is permitted for this site.' );
+			return new \WP_Error( 'db-cli-disallowed-subcmd', 'Only the `wp db query` subcommand is permitted for this site.' );
```
rinatkhaziev: I don't think this is correct, there are some other commands that are used  in other parts of the stack like `wp db prefix`, and others are pretty innocuous.  We should come up with an explicit list.

mjangda:
@rebeccahum @rinatkhaziev `wp db` is also used in k8s-site-jobs are part of exports (see BB8-10113). We should skip any restrictions from those pods as well.

rebeccahum:
@mjangda Batch containers are skipped from these restrictions.

lib/helpers/wp-cli-db/class-wp-cli-db.php
rebeccahum: @rinatkhaziev What's the best way to go about compiling that list? 🙂

lib/helpers/wp-cli-db/class-wp-cli-db.php
```diff
@@ -30,40 +31,83 @@ public function early_init() {
...
+		}
+
+		if ( 'query' === $subcommand && 2 === count( $command ) ) {
+			// Doing `wp db query` without a DB query is the equivalent of doing `wp db cli`
+			return new \WP_Error( 'db-cli-missing-query', 'Please provide the database query as a part of the command.' );
 		}
 
-		if ( 2 === count( $command ) ) {
-			throw new Exception( 'ERROR: Please provide the database query as a part of the command.' );
+		$query      = $command[2];
```
rinatkhaziev: This seems to be missing check that it's indeed a `query` subcommand?

lib/helpers/wp-cli-db/class-wp-cli-db.php
```diff
@@ -30,40 +31,83 @@ public function early_init() {
...
+		}
+	}
+
+	/**
+	 * Ensure the query is allowed.
+	 *
+	 * @param string $query
+	 * @return WP_Error|void
+	 */
+	public function validate_query( string $query ) {
```
rinatkhaziev, line 79: I'd modify the signture to return bool|WP_Error and add `return true` at the end. Which woud signify that the query is valid.

rinatkhaziev (changes requested):
Thanks, left some inline feedback.

Additionally, have this been tested in all the contexts? Locally, on a sandbox and on batch containers.

lib/helpers/wp-cli-db/class-wp-cli-db.php
rebeccahum: Hmmm, no I think I did account for that 🙂

lib/helpers/wp-cli-db/class-wp-cli-db.php
rebeccahum: I wonder if that would be confusing because someone may expect it to return false on invalidation 🤔

rebeccahum:
@rinatkhaziev Yep, it was deployed to 2160 and tested in different contexts.

mjangda:
> Batch containers are skipped from these restrictions.

I believe site jobs run on `job` containers so this wouldn't help.

rebeccahum:
@mjangda The jobs containers seem to have the `IS_VIP_BATCH_CONTAINER` env set, so this would include jobs containers for them. 🙂

mjangda:
> The jobs containers seem to have the IS_VIP_BATCH_CONTAINER env set, so this would include jobs containers for them. 🙂

And that's why you never listen to @mjangda ...

Thanks for double-checking!

lib/helpers/wp-cli-db/class-wp-cli-db.php
```diff
@@ -30,40 +30,85 @@ public function early_init() {
...
+			// This will throw an error if db commands are not enabled for this env
+			WP_CLI::error( $e->getMessage() );
+		}
 
-		if ( ! $this->config->is_sandbox() ) {
-			// This will throw an exception if the db subcommand is not valid:
+		try {
 			$this->validate_subcommand( $command );
+		} catch ( Exception $e ) {
+			WP_CLI::Error( $e->getMessage() );
```
rinatkhaziev:
```suggestion
			WP_CLI::error( $e->getMessage() );
```

rinatkhaziev (approved):

PR #4128, opened 2023-02-15
Add more files to ignore list 

rinatkhaziev:
## Description

This should recursively ignore many dotfiles, tests, and other stuff that shouldn't be going into the built version.

## Checklist

## Steps to Test

Needs a bit of testing :)


rinatkhaziev:
I diffed the prod and this branch images, the results are in pb 2f24f - I think everything looks good any nothing is missing but would appreciate extra 👀

rebeccahum (approved):
This looks good to me. Do we need to do an audit on https://github.com/Automattic/vip-go-mu-plugins-ext/blob/trunk/.dockerignore?

rinatkhaziev:
@rebeccahum I don't think so, this should catch it

PR #4129, opened 2023-02-16
Jetpack: Removed unneeded versions

luiztiago:
## Description

Removing Jetpack 9.4 and 9.8 fallback versions, because we don't have support with Wordpress 5.6 and 5.7

### With your local Wordpress version:

1. Check out PR.
2. On Local dev-env, run wp shell using `vip dev-env exec -- wp shell`
3. Type `VIP_JETPACK_DEFAULT_VERSION`. Then, you should receive the mapped Jetpack version for your Wordpress version. If you are using `6.1.1`, the shell should print `11.8`.

### With other Wordpress versions:

1. Set `$wp_version = '...'` with some Wordpress version.
2. type `vip_default_jetpack_version()` to reload the version.
3. type `VIP_JETPACK_DEFAULT_VERSION`
4. You should see the related Jetpack version

To help the debug, the versions map is listed below:

```php
$versions_map = [
	// WordPress version => Jetpack version
	'5.8.6' => '10.9',
	'5.9.0' => '11.4',
	'6.0.0' => '11.8',
];
```

### DIFF
--- a/jetpack.php
+++ b/jetpack.php
@@ -21,13 +21,13 @@ function vip_default_jetpack_version() {
 	global $wp_version;
 
 	if ( version_compare( $wp_version, '5.9', '<' ) ) {
-		// Wordpress 5.8.x and older. Not including 5.9.
+		// WordPress 5.8.x and older. Not including 5.9.
 		return '10.9';
 	} elseif ( version_compare( $wp_version, '6.0', '<' ) ) {
-		// Wordpress 5.9.x and newer. Not including 6.0.
+		// WordPress 5.9.x and newer. Not including 6.0.
 		return '11.4';
 	} else {
-		// Wordpress 6.0 and newer
+		// WordPress 6.0 and newer
 		return '11.8';
 	}
 }
--- a/jetpack.php
+++ b/jetpack.php
@@ -17,16 +17,24 @@
 
 // Choose an appropriate default Jetpack version, ensuring that older WordPress versions
 // are not using a too modern Jetpack version that is not compatible with it
-if ( ! defined( 'VIP_JETPACK_DEFAULT_VERSION' ) ) {
-	if ( version_compare( $wp_version, '5.8', '<' ) ) {
-		define( 'VIP_JETPACK_DEFAULT_VERSION', '10.4' );
-	} elseif ( version_compare( $wp_version, '5.9', '<' ) ) {
-		define( 'VIP_JETPACK_DEFAULT_VERSION', '10.9' );
+function vip_default_jetpack_version() {
+	global $wp_version;
+
+	if ( version_compare( $wp_version, '5.9', '<' ) ) {
+		// Wordpress 5.8.x and older. Not including 5.9.
+		return '10.9';
 	} elseif ( version_compare( $wp_version, '6.0', '<' ) ) {
-		define( 'VIP_JETPACK_DEFAULT_VERSION', '11.4' );
+		// Wordpress 5.9.x and newer. Not including 6.0.
+		return '11.4';
 	} else {
-		define( 'VIP_JETPACK_DEFAULT_VERSION', '11.8' );
+		// Wordpress 6.0 and newer
+		return '11.8';
 	}
+}
+
+// Set the default Jetpack version if it's not already defined
+if ( ! defined( 'VIP_JETPACK_DEFAULT_VERSION' ) ) {
+	define( 'VIP_JETPACK_DEFAULT_VERSION', vip_default_jetpack_version() );
 }
 
 // Bump up the batch size to reduce the number of queries run to build a Jetpack sitemap.
--- a/jetpack.php
+++ b/jetpack.php
@@ -18,11 +18,7 @@
 // Choose an appropriate default Jetpack version, ensuring that older WordPress versions
 // are not using a too modern Jetpack version that is not compatible with it
 if ( ! defined( 'VIP_JETPACK_DEFAULT_VERSION' ) ) {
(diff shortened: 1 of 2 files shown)


rebeccahum (dismissed):
🧨

rinatkhaziev (changes requested):
As discussed in Slack this contains one extra condition, since we only have 5.8 and above the first condition should be

```if ( version_compare( $wp_version, '5.9', '<' ) ) {```
for all sites less than 5.9 set default as 10.9

But maybe we should use this PR to rework the whole block to be easier testable and less confusing by abstracting this as a function which we can write tests for.

**Comment by luiztiago** [comment]
> As discussed in Slack this contains one extra condition, since we only have 5.8 and above the first condition should be
> 
> `if ( version_compare( $wp_version, '5.9', '<' ) ) {` for all sites less than 5.9 set default as 10.9
> 
> But maybe we should use this PR to rework the whole block to be easier testable and less confusing by abstracting this as a function which we can write tests for.

Totally agree. The logic is not easy to understand at first time, with some code pranks. Now we have a lower quantity of Jetpack versions and I think the key/value usage would not help to get the code (because we'll keep with pranks such as `{ "5.9": "10.9" }`, without including 5.9.x versions).

I researched a bit and I think the comments on this case should help us to don't add code complexity and have good readability.

```js
if ( ! defined( 'VIP_JETPACK_DEFAULT_VERSION' ) ) {
	if ( version_compare( $wp_version, '5.9', '<' ) ) {
		// Wordpress 5.8.x and older. Not including 5.9.
		define( 'VIP_JETPACK_DEFAULT_VERSION', '10.9' );
	} elseif ( version_compare( $wp_version, '6.0', '<' ) ) {
		// Wordpress 5.9.x and newer. Not including 6.0.
		define( 'VIP_JETPACK_DEFAULT_VERSION', '11.4' );
	} else {
		// Wordpress 6.0 and newer
		define( 'VIP_JETPACK_DEFAULT_VERSION', '11.8' );
	}
}
```

Other than that, tests are always good to be added. 🧶

tests/test-jetpack.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+// phpcs:disable PEAR.NamingConventions.ValidClassName.Invalid
+
+class VIP_Go__Core__Default_VIP_Jetpack_Version extends WP_UnitTestCase {
+	public function test__vip_default_jetpack_version() {
+		$versions_map = [
+			// WordPress version => Jetpack version
+			'5.8.6' => '10.9',
+			'5.9.0' => '11.4',
```
rinatkhaziev: WordPress doesn't exactly follow semver, so the versions would be like 5.9, 5.9.1 ,etc
```suggestion
			'5.9' => '11.4',
```

tests/test-jetpack.php
```diff
@@ -0,0 +1,24 @@
...
+			// WordPress version => Jetpack version
+			'5.8.6' => '10.9',
+			'5.9'   => '11.4',
+			'5.9.5' => '11.4',
+			'6.0'   => '11.8',
+			'6.1'   => '11.8',
+			'6.1.1' => '11.8',
+		];
+
+		foreach ( $versions_map as $wordpress_version => $jetpack_version ) {
```
rinatkhaziev, line 20: @sjinks will overwriting the global potentially mess up other tests?

jetpack.php
```diff
@@ -17,22 +17,26 @@
...
-		define( 'VIP_JETPACK_DEFAULT_VERSION', '9.8' );
-	} elseif ( version_compare( $wp_version, '5.8', '<' ) ) {
-		define( 'VIP_JETPACK_DEFAULT_VERSION', '10.4' );
-	} elseif ( version_compare( $wp_version, '5.9', '<' ) ) {
-		define( 'VIP_JETPACK_DEFAULT_VERSION', '10.9' );
+function vip_default_jetpack_version() {
+	global $wp_version;
+
+	if ( version_compare( $wp_version, '5.9', '<' ) ) {
+		// Wordpress 5.8.x and older. Not including 5.9.
```
rinatkhaziev:
```suggestion
		// WordPress 5.8.x and older. Not including 5.9.
```

jetpack.php
```diff
@@ -17,22 +17,26 @@
...
+		return '10.9';
 	} elseif ( version_compare( $wp_version, '6.0', '<' ) ) {
-		define( 'VIP_JETPACK_DEFAULT_VERSION', '11.4' );
+		// Wordpress 5.9.x and newer. Not including 6.0.
```
rinatkhaziev:
```suggestion
		// WordPress 5.9.x and newer. Not including 6.0.
```

jetpack.php
```diff
@@ -17,22 +17,26 @@
...
+		return '11.4';
 	} else {
-		define( 'VIP_JETPACK_DEFAULT_VERSION', '11.8' );
+		// Wordpress 6.0 and newer
```
rinatkhaziev:
```suggestion
		// WordPress 6.0 and newer
```

tests/test-jetpack.php
rebeccahum: Might be best practice anyways to reset it back after the test is done for clean up?

rinatkhaziev: @luiztiago can you address this please? and then this is good to go

rebeccahum: @rinatkhaziev @luiztiago Addressed ahead of tmrw's release.

luiztiago: Thanks @rebeccahum 🚀

rinatkhaziev (approved):

PR #4176, opened 2023-03-03
Query Monitor: Enable for non-production environments

rebeccahum:
## Description
Enables Query Monitor for non-production environments by default. You still need the view_query_monitor capability to see QM (which is by default, granted for users that have the manage_options cap). 

Enable QM for non-production environments by default. Only applies for users with `manage_options` cap and above. Filter `wpcom_vip_qm_enable` to opt out.

## Checklist

## Steps to Test
1) Create administrator user
2) See no QM
3) Apply PR
4) See QM

### DIFF
--- a/query-monitor.php
+++ b/query-monitor.php
@@ -66,6 +66,10 @@ function wpcom_vip_qm_enable( $enable ) {
 		return true;
 	}
 
+	if ( defined( 'VIP_GO_APP_ENVIRONMENT' ) && 'production' !== constant( 'VIP_GO_APP_ENVIRONMENT' ) ) {
+		return true;
+	}
+
 	return $enable;
 }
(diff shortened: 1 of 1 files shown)


rinatkhaziev:
@rebeccahum Will it be visible for other roles?

rebeccahum:
@rinatkhaziev Is what you're asking is if the intention is to have it visible for roles that are non-administrators? 
Because to clarify, that will **not** happen with this PR, as you still need the `view_query_monitor` capability to see QM (which is by default, granted for users that have the `manage_options` cap). So, all in all, this only enables it for roles with the `manage_options` caps on non-production VIP environments.

rinatkhaziev (approved):

PR #4183, opened 2023-03-07
VaultPress: Stop loading VaultPress plugin by default for non-production VIP environments

saroshaga:
## Description

VaultPress is being deprecated and as a first step, we will stop loading it for non-production environments by default from the 14th of March 2023. 

Adding the following line to a site's `vip-config.php` file should override this for now:

```
define( 'VIP_VAULTPRESS_SKIP_LOAD', false );
```

Find more information here: https://lobby.vip.wordpress.com/2023/02/28/upcoming-vaultpress-deprecation/

### Plugin Updated: VaultPress

Stop loading VaultPress plugin by default for non-production VIP environments


WPprodigy:
Made a slight change to instead set the `VIP_VAULTPRESS_SKIP_LOAD` constant, which will then return early in the next condition block all the same.

Reason being there are a few other places in the codebase that currently look for this constant in order to avoid interacting with VP if needed (during connections for example).

WPprodigy (approved):

PR #4185, opened 2023-03-07
Index theme information

WPprodigy:
## Description

Starts indexing relevant theme information, similar to what we gather for plugins.

## Checklist

## Steps to Test

A few ways to go about this. Easiest might be to just var_dump this result: https://github.com/Automattic/vip-go-mu-plugins/blob/4192b4b2e88c174bf58ff7d7de21c0e87389891a/config/class-site-details-index.php#L288

And then run the `vip_config_sync_cron` cron event via `wp cron-control events run {$ID}`


config/class-site-details-index.php
```diff
@@ -191,6 +192,64 @@ private function get_plugin_marketplace( $plugin_url ) {
 		return null;
```
rinatkhaziev, line 192: I know it was already like that but should this be `null`?

WPprodigy: Yeah, null is expected for any of the update-related fields that we aren’t able to find.

rinatkhaziev (approved):

PR #4187, opened 2023-03-07
Add new cron control folders

WPprodigy:
## Description

- Add `cron/cron-control` plugin as a subtree located inside the `cron` folder. This will receive the new updates, and will be the main plugin moving forward.
- Add `cron/cron-control-legacy` as a submodule in the same place, to be used for legacy envs until they are all extinct. This will track the `legacy` branch in the remote repo - which will mostly never change barring any urgent fixes.
- Keeping the existing `cron-control` submodule just for now to help avoid any temp issues during non-atomic-rollout. After this lands in the staging branch, we'll start loading from the new locations. And then look to remove the old submodule completely.

Also did a few updates to the phpcs & dockerignore files.

## Code Review / Deploy Steps

Can just ignore the subtree in `cron/cron-control` since the whole plugin doesn't need a re-review of course :). And then like mentioned earlier, the plan is to:

- Deploy this to staging alongside any other changes per usual.
- Then the next week we'll merge a new PR that starts loading from the new locations.
- Lastly, we can remove the old submodule completely.

Non-atomic deploy fun 🙃 . Step 2's code will just be like:

```
if ( ( defined( 'IS_VIP_ON_KUBERNETES' ) && true === IS_VIP_ON_KUBERNETES ) || true !== WPCOM_IS_VIP_ENV ) {
	require_once __DIR__ . '/cron/cron-control/cron-control.php';
} else {
	// Load an older version of the plugin on environments still on the legacy architecture.
	require_once __DIR__ . '/cron/cron-control-legacy/cron-control.php';
}
```

## Checklist

## Steps to Test

Make sure cron control is still loading in the appropriate way :)


rebeccahum (approved):
Since we aren't modifying any loading yet

PR #4191, opened 2023-03-08
Feature: is_enabled() should also check if enabled by ID and adds is_enabled_by_env() & get_features() and `wp vip-feature` CLI commands

rebeccahum:
## Description
This PR includes the below:
1. When we use `is_enabled()` to check if a feature is enabled, we should also take into account if it's enabled on an ID-basis rather than just percentage
2. Adds `is_enabled_by_env()` to control enabling by environment.
3. Adds `get_features()` to see a list of existing features in place
4. Adds `wp vip feature` CLI commands.

### Plugin Updated: Feature

- Adds two new functions `is_enabled_by_env()` and `get_features()`
- `is_enabled()` will take into account if the feature is also enabled on an ID-basis or env-basis
- Adds `wp vip-feature` CLI commands to check the status of a feature

## Checklist

## Steps to test
1) Add to `testing` feature to `$feature_envs` for local only:
```
	public static $feature_envs = [
		'testing' => [
			'local' => true,
		],
	];
```
2) Do `\Automattic\VIP\Feature::is_enabled_by_env( 'testing' );` and expect it to return `true`
3) Do `\Automattic\VIP\Feature::get_features();` and expect it to return `[ 'testing' ]`
4) Do `wp vip-feature list` and expect to see `testing` returned:
```
+------------------------+----------+
| Feature                | Status   |
+------------------------+----------+
| testing                | enabled  |
+------------------------+----------+
```

### DIFF
--- a/lib/feature/class-feature-cli.php
+++ b/lib/feature/class-feature-cli.php
@@ -6,7 +6,7 @@ use WP_CLI;
 use \Automattic\VIP\Feature as Feature;
 
 // phpcs:ignore WordPressVIPMinimum.Classes.RestrictedExtendClasses.wp_cli
-class FeatureCLI extends \WP_CLI_Command {
+class Feature_CLI_Command extends \WP_CLI_Command {
 	/**
 	 * List all available VIP features.
 	 *
--- a/lib/feature/class-feature-cli.php
+++ b/lib/feature/class-feature-cli.php
@@ -63,4 +63,4 @@ class FeatureCLI extends \WP_CLI_Command {
 	}
 }
 
-WP_CLI::add_command( 'vip feature', __NAMESPACE__ . '\FeatureCLI' );
+WP_CLI::add_command( 'vip feature', __NAMESPACE__ . '\Feature_CLI_Command' );
--- a/lib/feature/class-feature-cli.php
+++ b/lib/feature/class-feature-cli.php
@@ -1,0 +1,65 @@
+<?php
+
+namespace Automattic\VIP;
+
+use WP_CLI_Command;
+use WP_CLI;
+use \Automattic\VIP\Feature as Feature;
+
+class FeatureCLI extends WP_CLI_Command {
+	/**
+	 * List all available VIP features.
+	 *
+	 * @subcommand list
+	 *
+	 * ## EXAMPLES
+	 *
+	 *     wp vip-feature list [--format=<table|json|csv|yaml>]
+	 */
+	public function list_features( $args, $assoc_args ) {
+		$format = $assoc_args['format'] ?? 'table';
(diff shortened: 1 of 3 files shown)


lib/feature/class-feature.php
```diff
@@ -30,8 +30,70 @@ class Feature {
...
-	public static function is_enabled( $feature ) {
-		return static::is_enabled_by_percentage( $feature );
+	/**
+	 * Holds feature slug and then, key of environments with bool value to enable E.g.
+	 * // Enable feature for non-production sites
+	 * // 'feature-flag' => [ 'non-production' => true ],
+	 *
+	 * @var array
+	 */
+	public static $feature_envs = [];
```
rinatkhaziev, line 48: I think this needs future clarification? Looking at the code below it allows to do both `non-production` and `specific_environment_slug`

lib/feature/class-feature.php
```diff
@@ -30,8 +30,70 @@ class Feature {
...
+	}
+
+	/**
+	 * Returns all features that exist.
+	 *
+	 * @return array $features Array of features.
+	 */
+	public static function get_features() {
+		$features = array_merge(
+			array_keys( static::$feature_percentages ),
```
rinatkhaziev: Should this be more defensive on the off-chance someone messes up the property definition in the future?

lib/feature/class-feature-cli.php
```diff
@@ -0,0 +1,66 @@
+<?php
+
+namespace Automattic\VIP;
+
+use WP_CLI;
+use \Automattic\VIP\Feature as Feature;
+
+// phpcs:ignore WordPressVIPMinimum.Classes.RestrictedExtendClasses.wp_cli
+class FeatureCLI extends \WP_CLI_Command {
```
rinatkhaziev:
```suggestion
class Feature_CLI_Command extends \WP_CLI_Command {
```

lib/feature/class-feature-cli.php
```diff
@@ -0,0 +1,66 @@
...
+		$features = Feature::get_features();
+		if ( ! in_array( $feature, $features, true ) ) {
+			WP_CLI::error( "Invalid feature slug '$feature'" );
+		}
+
+		WP_CLI::line( sprintf( '%1$s is %2$s.', $feature, Feature::is_enabled( $feature ) ? 'enabled' : 'disabled' ) );
+	}
+}
+
+WP_CLI::add_command( 'vip feature', __NAMESPACE__ . '\FeatureCLI' );
```
rinatkhaziev:
```suggestion
WP_CLI::add_command( 'vip feature', __NAMESPACE__ . '\Feature_CLI_Command' );
```

rinatkhaziev (approved):

PR #4200, opened 2023-03-09
tests: Fix "Undefined constant 'Automattic\VIP\Files\WP_CLI'" error

### LINKED ISSUE #4192: Error: Undefined constant "Automattic\VIP\Files\WP_CLI"
```
Automattic\VIP\Files\API_Client_Test::test__get_unique_filename with data set "WP_Error" (WP_Error Object (...), WP_Error Object (...))
Error: Undefined constant "Automattic\VIP\Files\WP_CLI"

/home/runner/work/vip-go-mu-plugins/vip-go-mu-plugins/files/class-api-client.php:43
/home/runner/work/vip-go-mu-plugins/vip-go-mu-plugins/tests/files/test-api-client.php:44
/home/runner/work/vip-go-mu-plugins/vip-go-mu-plugins/tests/files/test-api-client.php:23
```

Ref: https://github.com/Automattic/vip-go-mu-plugins/actions/runs/4296707896/jobs/7488766006#step:5:341

sjinks:
This PR‌ fixes another hard-to-reproduce heisenbug which depends on the specific order of tests.

Closes: #4192

### DIFF
--- a/files/class-api-client.php
+++ b/files/class-api-client.php
@@ -37,10 +37,10 @@ class API_Client {
 		$this->files_token   = $files_token;
 
 		// Add some context to the UA to simplify debugging issues
-		if ( defined( 'DOING_CRON' ) && DOING_CRON ) {
+		if ( defined( 'DOING_CRON' ) && constant( 'DOING_CRON' ) ) {
 			// current_filter may not be totally accurate but still better than nothing
 			$current_context = sprintf( 'Cron (%s)', current_filter() );
-		} elseif ( defined( 'WP_CLI' ) && WP_CLI ) {
+		} elseif ( defined( 'WP_CLI' ) && constant( 'WP_CLI' ) ) {
 			$current_context = 'WP_CLI';
 		} else {
 			$current_context = add_query_arg( [] );
--- a/tests/cache/test-vary-cache.php
+++ b/tests/cache/test-vary-cache.php
@@ -30,6 +30,7 @@ class Vary_Cache_Test extends WP_UnitTestCase {
 	}
 
 	public function tearDown(): void {
+		Constant_Mocker::clear();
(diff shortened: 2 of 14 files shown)


tests/cache/test-vary-cache.php
```diff
@@ -30,6 +30,7 @@ public function setUp(): void {
 	}
 
 	public function tearDown(): void {
+		Constant_Mocker::clear();
```
sjinks, line 33: We have to clear mocked constants in `tearDown()`; otherwise, it is possible that another test will break because another test has mocked an unexpected constant.

files/class-api-client.php
```diff
@@ -37,10 +37,10 @@ public function __construct( $api_base, $files_site_id, $files_token, $cache ) {
 		$this->files_token   = $files_token;
 
 		// Add some context to the UA to simplify debugging issues
-		if ( defined( 'DOING_CRON' ) && DOING_CRON ) {
+		if ( defined( 'DOING_CRON' ) && constant( 'DOING_CRON' ) ) {
 			// current_filter may not be totally accurate but still better than nothing
 			$current_context = sprintf( 'Cron (%s)', current_filter() );
-		} elseif ( defined( 'WP_CLI' ) && WP_CLI ) {
+		} elseif ( defined( 'WP_CLI' ) && constant( 'WP_CLI' ) ) {
```
sjinks, line 43: With constant mocking, whenever we use `defined()`, we have to use `constant()`. Otherwise, when the constant was defined in the locker and then it is later accessed without `constant()`, PHP will spit an error because it knows nothing about the constant.

In PHP 8.1, undefined constants trigger fatal errors.

rebeccahum (approved):
🙇‍♀️

PR #4206, opened 2023-03-09
Add Feature::get_bucket method which returns the bucket the site is in based on ID and feature

rinatkhaziev:
## Description

This is just a tiny abstraction of existing logic that will unlock us to easier answer a question like "is site X and network site Y is included in certain bucket in %-based rollout?".

### Library Updated: Feature

We've introduced an API to get the bucket the current site is in for %-based feature rollouts.

## Checklist

## Steps to Test

1. Check out PR.
1. `wp shell`
2. `\Automattic\VIP\Feature::get_bucket( 31337, 'cool-feature' );` should return an integer in the range of 0-99

### DIFF
--- a/lib/feature/class-feature.php
+++ b/lib/feature/class-feature.php
@@ -159,9 +159,12 @@ class Feature {
 	}
 
 	/**
-	 * Given a site ID and a feature, return the bucket the site is in
+	 * Given a site ID and a feature, return the bucket the site is in.
 	 *
-	 * @param integer $site_id
+	 * Note: this is a lower-level API, it doesn't care if the feature is defined or whether it's a multisite or not.
+	 * Please use is_enabled()/is_enabled_by_ids()/is_enabled_by_percentage() instead.
+	 *
+	 * @param mixed $site_id
 	 * @param string $feature
 	 * @return integer
 	 */
--- a/lib/feature/class-feature.php
+++ b/lib/feature/class-feature.php
@@ -145,16 +145,31 @@ class Feature {
 
 		// Which bucket is the site in - 100 possibilites, one for each percentage. We run this through crc32 with
 		// the feature name so that the same sites aren't always the canaries
-		$bucket = crc32( $feature . '-' . constant( 'FILES_CLIENT_SITE_ID' ) ) % 100;
+		$bucket = static::get_bucket( constant( 'FILES_CLIENT_SITE_ID' ), $feature );
 
 		// Is the bucket enabled?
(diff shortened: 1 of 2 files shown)


lib/feature/class-feature.php
```diff
@@ -145,16 +145,31 @@ public static function is_enabled_by_percentage( string $feature ) {
 
 		// Which bucket is the site in - 100 possibilites, one for each percentage. We run this through crc32 with
 		// the feature name so that the same sites aren't always the canaries
-		$bucket = crc32( $feature . '-' . constant( 'FILES_CLIENT_SITE_ID' ) ) % 100;
+		$bucket = static::get_bucket( constant( 'FILES_CLIENT_SITE_ID' ), $feature );
 
 		// Is the bucket enabled?
 		$threshold = $percentage * 100; // $percentage is decimal
 
 		if ( is_multisite() && $bucket < $threshold ) {
```
rebeccahum, line 153: Should we move this logic inside `get_bucket` too? Since it's possible that in a multi-site, `$bucket` will be smaller than `$threshold` and we need to check the crc32 against the blog ID ...

for example:

```
\Automattic\VIP\Feature::get_bucket( 31337, 'cool-feature' ); 
```
Say that 31337 is a multi-site that has a threshold larger than bucket, then the real bucket actually is against the blog ID

rebeccahum (changes requested):

lib/feature/class-feature.php
rinatkhaziev: No I don't think so, this would make the method [impure](https://en.wikipedia.org/wiki/Pure_function), which is something I don't like and would also defeat the purpose of this abstraction :)

rebeccahum: OK fair enough. I guess as long as one knows that doing `\Automattic\VIP\Feature::get_bucket( 31337, 'cool-feature' );` might not be the most accurate for multi-sites due to the added processing logic 🙂 since there are cases where it might use the blog ID instead of the app ID. Maybe we can just update the PHPDocs with clarifying comments on that.

rinatkhaziev: @rebeccahum How would you phrase that comment?

rebeccahum: "Note: When using `get_bucket()` in a multisite, the blog ID may need to be used instead of app ID if `$threshold` exceeds the `$bucket`."

rebeccahum (dismissed):

PR #4218, opened 2023-03-14
Production release v20230314.0 🥧 

rebeccahum:

PR #4219, opened 2023-03-14
Production release v20230314.0 🥧

rebeccahum:
## Description
Same as https://github.com/Automattic/vip-go-mu-plugins/pull/4218 but removes https://github.com/Automattic/vip-go-mu-plugins/pull/4218/commits/ff665b1f24141b59b84ffb104e3c33f8076106e8 since we are deploying that on Thursday.


luiztiago (approved):
Looks great

PR #4244, opened 2023-03-17
Search: Add CLI command `wp vip-search health stop-validate-contents`

rebeccahum:
## Description
Would be nice to have a killswitch for the validate-contents CLI command since it can eat up a lot of resources.

Added CLI command `stop-validate-contents` to stop any validate_contents processing.

## Checklist

## Steps to Test
1) Kick off a validate contents `wp vip-search health validate-contents --max_diff_size=1000000`
2) While 1) is still running, run `wp vip-search health stop-validate-contents` and see process stop in 1)
3) Ensure that lock and transient is cleaned up via `wp option get vip_search_content_validation_process_post_id` and `wp cache get search_interrupt_validate_contents vip_search`

### DIFF
--- a/search/includes/classes/class-health.php
+++ b/search/includes/classes/class-health.php
@@ -48,7 +48,9 @@ class Health {
 
 	const REINDEX_JOB_DEFAULT_PRIORITY = 15;
 
-	const STOP_VALIDATE_CONTENTS_TRANSIENT = 'search_interrupt_validate_contents';
+	const STOP_VALIDATE_CONTENTS_KEY = 'search_interrupt_validate_contents';
+
+	const CACHE_GROUP = 'vip_search';
 
 	/**
 	 * Instance of Search class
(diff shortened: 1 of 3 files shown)


search/includes/classes/class-health.php
```diff
@@ -550,10 +558,16 @@ public function validate_index_posts_content( $options ) {
 				// phpcs:ignore WordPress.WP.AlternativeFunctions.rand_mt_rand
 				sleep( mt_rand( 2, 5 ) );
 			}
-		} while ( $start_post_id <= $last_post_id );
 
-		if ( $process_parallel_execution_lock ) {
+			$should_stop = get_transient( self::STOP_VALIDATE_CONTENTS_TRANSIENT );
```
rinatkhaziev: Are we sure that `get_transient` is the right approach here? Since it's a wrapper around `$value = wp_cache_get( $transient, 'transient' );` it'll just stay in memory.

https://github.com/WordPress/wordpress-develop/blob/6.1/src/wp-includes/option.php#L876-L878

Additionally, shouldn't it be in the do/while block above?  I.e. check that on every batch.

rinatkhaziev (changes requested):
Left some inline comments

search/includes/classes/class-health.php
rebeccahum: It is already in the do/while block above 🙂 . Is there another one I'm missing?

Re: using a transient, what do you recommend instead? An option for better longevity?

rinatkhaziev:
Thanks for the changes, the other thing i noticed is that some of the error codes. have `es_` prefix and others don't. Can we tidy this up as a part of this PR?

rinatkhaziev (approved):

PR #4266, opened 2023-03-24
Search: Improve get-last-indexed-post-id to include timestamp for debugging

rebeccahum:
## Description
If an index gets interrupted, we usually use `wp vip-search get-last-indexed-post-id` to check the last post ID. This change adds the timestamp and changes it to a JSON format:
```
wp vip-search get-last-indexed-post-id                                                                                             
{"post_id":979,"time":"2023-03-24 18:15:37"}
```

Add timestamp to `wp vip-search get-last-indexed-post-id`

## Checklist

## Steps to Test
1) Kick off an index with `wp vip-search index`
2) While the index is happening, in another window, do `wp vip-search get-last-indexed-post-id` and you should see a format similar to Description

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -675,9 +675,8 @@ class Search {
 
 		add_filter( 'ep_sync_indexable_kill', [ $this, 'do_not_sync_if_no_index' ], PHP_INT_MAX, 2 );
 
-		// Store last processed id into option and clean up before & after indexing command
+		// Store last processed id into option and clean up before indexing command
 		add_action( 'ep_cli_post_bulk_index', [ $this, 'update_last_processed_post_id_option' ], 10, 2 );
-		add_action( 'ep_wp_cli_after_index', [ $this, 'delete_last_processed_post_id_option' ] );
 		add_action( 'ep_wp_cli_pre_index', [ $this, 'delete_last_processed_post_id_option' ] );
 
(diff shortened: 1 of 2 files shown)


pschoffer (dismissed):
Works great!

should we maybe even drop the removal of the option?
https://github.com/Automattic/vip-go-mu-plugins/blob/search/add-timestamp-to-cli-get-last-indexed-post-id/search/includes/classes/class-search.php#L680


Now that we have the timestamp it might be useful to keep it in untill the next indexing. The only worry is that somebody might be using it to tell if they are indexing or not, but hopefully not?

The benefit of not dropping it is that it is easy to see when the last indexing finished + what was the last post ID it actually indexed. Although the ID is not that helpful as the indexing goes from newest (highest ID) to oldest (lowest ID).

rebeccahum:
@pschoffer Hm, good point! Updated the PR now.

> The only worry is that somebody might be using it to tell if they are indexing or not, but hopefully not?

We have `get-indexing-status` to tell though if an indexing is in progress, so I think we should be OK 🙂

rebeccahum:
@pschoffer After some thinking, I made the change and reverted it because if we want to have a `--resume` option on `wp vip-search index`, it might be good to keep it as-is. LMK what you think.

pschoffer (dismissed):
Looks great! Thanks for checking that removal.

rinatkhaziev (approved):

PR #4270, opened 2023-03-27
Adds Parsely configs in SDS and extended integration types with disabled options

mehmoodak:
## Description

Previously we only tracked basic Parse.ly attributes in SDS ( `active, integration_type, version` ) but now we also introduced an additional `configs` key which tracks more details about how Parse.ly plugin is integrated.

Example of possible config attributes:
```
{
        "is_pinned_version": false,
	"site_id": "example.com",
        "have_api_secret": true,
	"is_javascript_disabled": false,
	"is_autotracking_disabled": false,
	"should_track_logged_in_users": false,
	"tracked_post_types": [
		{ "type": "post", "track_type": "post" },
		{ "type": "page", "track_type": "non-post" },
		{ "type": "cpt_post", "track_type": "do-not-track" }
	]
}
```

Apart from this we have also extended the possible values in `integration_type` key and now also tracks disabled status of the plugin.

```
- DISABLED_MUPLUGINS_FILTER
- DISABLED_MUPLUGINS_SILENT_OPTION
```

## Changelog Description

Introduced `configs` key in `parsely` attributes which tracks more info about how the plugin is used and also extended possible integration types to also track the disable status of the plugin

## Checklist

- [x] This change has relevant tests (if applicable).

## Steps to Test

1. Pull branch

3. Test `config` attributes

    3.1. Enable `wp-parsely` by following [VIP Docs](https://docs.wpvip.com/how-tos/enabling-and-disabling-parsely/)

    3.2. Pull specifc branch from SDS that have these config changes (if not merged), run `docker-compose up` and make changes in `put_site_details` function so that it talks to the SDS running locally
    
    3.3. For verification we can either use `print_r` in `put_site_details` function and verify the body of `PUT` request or we can also verify the values from locally available SDS db
    
    3.4. Makes changes in plugin settings by going into `WP Admin > Settings > Parsely` and verify that `parsely.configs` key inside `PUT /site-details/sites` request have the expected values in it
    
    3.5. For testing `is_pinned_version` key inside `configs` you will have to use `wpvip_parsely_version` filter in `plugin-loader.php`

4. Test new values of `integration_type` 

    4.1. Use `wpvip_parsely_load_mu` filter to disable the plugin and verify that `parsely.integration_type` inside `PUT /site-details/sites` request have value `DISABLED_MUPLUGINS_FILTER`
    
    4.2. Use `_wpvip_parsely_mu` option to disable the plugin and verify that `parsely.integration_type` inside `PUT /site-details/sites` request have value `DISABLED_MUPLUGINS_SILENT_OPTION`


.github/workflows/parsely.yml
```diff
@@ -22,13 +22,28 @@ jobs:
       fail-fast: false
       matrix:
         config:
+          # Oldest version of the parsely plugin
           - { wp: latest, parsely: '3.1', mode: 'filter_enabled', php: '7.4' }
+          - { wp: latest, parsely: '3.1', mode: 'filter_disabled', php: '7.4' }
```
mehmoodak, line 27: - Added `filter_disabled`, `option_disabled`, `filter_and_option_disabled` modes.
- Previously we were using some random parsely versions for testing but now in this PR we updated `parsely` plugin version to use the oldest version, some mid version and latest available version (without pinning i.e. removed parsely key from config object)

config/class-site-details-index.php
```diff
@@ -20,11 +20,6 @@ class Site_Details_Index {
 	 */
 	private $timestamp = null;
 
-	/**
-	 * Name of the logstash feature to use for log2logstash call
-	 */
-	private const LOG_FEATURE_NAME = 'site_details';
```
mehmoodak, line 26: Not used so removed.

tests/bootstrap.php
```diff
@@ -106,42 +116,38 @@ function _configure_wp_parsely_specified_version() {
 	}
 }
 
-switch ( getenv( 'WPVIP_PARSELY_INTEGRATION_PLUGIN_VERSION' ) ) {
```
mehmoodak, line 109: No need of this `_configure_specify_wp_parsely_version` function is doing the same.

tests/parsely/test-mu-parsely-integration.php
```diff
@@ -70,7 +72,7 @@ public function test_parsely_class_existance() {
 
 		maybe_load_plugin();
 
-		if ( 'disabled' !== self::$test_mode ) {
+		if ( ! $this->is_plugin_disabled() ) {
```
mehmoodak, line 75: Apart from simple disabled mode now we have also filter and option values that represents disabled mode so added a function that checks for each possible config through which plugin can be disabled.

wp-parsely.php
```diff
@@ -28,18 +30,11 @@
  * Keep track of Parse.ly loading info in one place.
  */
 final class Parsely_Loader_Info {
-	// Describes how the parse.ly plugin is integrated/loaded.
-	const INTEGRATION_TYPE_MUPLUGINS        = 'MUPLUGINS';
-	const INTEGRATION_TYPE_MUPLUGINS_SILENT = 'MUPLUGINS_SILENT';
-	const INTEGRATION_TYPE_SELF_MANAGED     = 'SELF_MANAGED';
-	const INTEGRATION_TYPE_NONE             = 'NONE';
```
mehmoodak, line 35: - Added `Parsely_Integration_Type` class which act as enum and moved all of these constants there.
- Apart from this also updated const values which are now more clear.

wp-parsely.php
```diff
@@ -28,18 +30,11 @@
...
-
 	// Defaults for when detection was not possible.
 	const VERSION_UNKNOWN = 'UNKNOWN';
 
 	private static bool $active;
 	private static string $integration_type;
-	private static string $service_type;
```
mehmoodak, line 42: Not used anywhere so removed.

types/wp-parsely.php
```diff
@@ -0,0 +1,29 @@
+<?php
+/**
+ * This file contains types information which is needed by wp-parsely plugin.
```
mehmoodak: - Gives more info that helps in understanding of code.
- Helps IDE in autocompletion.

mehmoodak (commented):
Left some comments for easier review.

wp-parsely.php
```diff
@@ -67,6 +62,51 @@ public static function set_version( string $version ): void {
...
+		$configs->site_id                      = $options['apikey'] ?? '';
+		$configs->have_api_secret              = '' !== ( $options['api_secret'] ?? '' );
+		$configs->is_javascript_disabled       = (bool) ( $options['disable_javascript'] ?? false );
+		$configs->is_autotracking_disabled     = (bool) ( $options['disable_autotrack'] ?? false );
+		$configs->should_track_logged_in_users = (bool) ( $options['track_authenticated_users'] ?? false );
+
+		$configs->tracked_post_types = array();
+		$post_names                  = get_post_types( array( 'public' => true ) );
+		foreach ( $post_names as $post_name ) {
+			$tracked_post_type       = new stdClass();
```
WPprodigy: I think it'll be a bit more on par with the rest of the codebase to use an array here instead of an object.

```
$tracked_post_type = [ 'name'  => $post_name ];
...
$tracked_post_type['track_type'] = 'post';
```

mehmoodak: Using `stdClass` makes it easier to document via PHPDoc and gives IDE ability to autocomplete. So is it possible to have same benefits while using `array`? If yes then please suggest the recommended way because in codebase seems like we are just using `@return array` without any types.

Ideally there  should be type checking using tools like PHPStan but since it is not present so using PHPDoc as much as we can seems like the best way.

mehmoodak: I found this as a way to use PHPDoc to properly document the returned array but still using `stdClass` seems better because of better autocompletion in code editors.

```
	/**
	 * Returns configuration which represents how the plugin is configured in site.
	 *
	 * @return array{
	 *   is_pinned_version: bool,
	 *   site_id: string,
	 *   have_api_secret: bool,
	 *   is_javascript_disabled: bool,
	 *   is_autotracking_disabled: bool,
	 *   should_track_logged_in_users: bool,
	 *   tracked_post_types: array{
	 *     name: string,
	 *     track_type: string,
	 *   }[]
	 * }
	 */
	public static function get_configs()
```

Let me know which method do you think is most appropriate and I will implement that. Thanks

mjangda:
@mehmoodak I'm assuming the changes to search/elasticpress-next were not intentional?

https://github.com/Automattic/vip-go-mu-plugins/pull/4270/files#diff-5ffda7c543d918856d4eadd2b937af9ebded4111c2447e6bfc42303e25371c39

wp-parsely.php
```diff
@@ -67,6 +62,51 @@ public static function set_version( string $version ): void {
...
+			$tracked_post_type->name = $post_name;
+
+			if ( in_array( $post_name, ( $options['track_post_types'] ?? array() ) ) ) {
+				$tracked_post_type->track_type = 'post';
+			} elseif ( in_array( $post_name, ( $options['track_page_types'] ?? array() ) ) ) {
```
mjangda, line 98: I would assign `( $options['track_post_types'] ?? array() )` and `( $options['track_page_types'] ?? array() )` to variables outside to the loop. Improves readability and avoids the needs to create unnecessary arrays on every iteration of the loop.

mjangda: I would assign `( $options['track_post_types'] ?? array() )` and `( $options['track_page_types'] ?? array() )` to variables outside to the loop. Improves readability and avoids the needs to create unnecessary arrays on every iteration of the loop.

wp-parsely.php
mjangda: @Automattic/vip-platform-cantina curious to hear the team's thoughts on this, including the `types/wp-parsely.php` file.

mehmoodak:
> @mehmoodak I'm assuming the changes to search/elasticpress-next were not intentional?

@mjangda these were included on its own, I have taken this PR ([Add wp-parsely loader information to SDS](https://github.com/Automattic/vip-go-mu-plugins/pull/3542)) as reference and that does also have these changes so I kept them.

Not sure about the general practice here so please let me know whether to keep them or remove them. Thanks

**Update:** Some Enterprise Search related tests seems to be failing due to this reason so I have removed these changes which are caused by executing `git submodule update --init --recursive`

.github/workflows/parsely.yml
WPprodigy: This feels like a lot of test runs, not sure how beneficial all these cases are going to be vs how long they're going to be holding up every PR :/

- Are the `filter_and_option_enabled` & `filter_and_option_disabled` cases really needed for example?
- I think we should just do 1 version of parsely (latest)

types/wp-parsely.php
WPprodigy: I'm not sure about adding in a pattern like this and a new top-level folder into the broader mu-plugins repo. Could we drop this for now? Can put any helpful comments in the source files.

wp-parsely.php
WPprodigy: I'm leaning towards the array docblock to be honest, at least for this use case. PHP in particular is moving away from dynamically added class properties as well. And a full fledged class seems kinda overkill here.

I'm all for better typing (pretty sure we support Psalm already), but I think there's a larger conversation to be had and I'd rather not introduce a one-off pattern with the `types/wp-parsely.php` thing.

.github/workflows/parsely.yml
mehmoodak: Removed the tests of 1 version because I think we should have at least one oldest version also which should flag issues when we are expecting something to happen but it can't due to lack of filter.

mehmoodak:
@WPprodigy incorporated the feedback please have a look again. Thanks

wp-parsely.php
```diff
@@ -163,7 +211,12 @@ function maybe_load_plugin() {
 	// No integration: The site has not enabled parsely.
 	if ( ! $should_load || $should_prevent_loading ) {
 		Parsely_Loader_Info::set_active( false );
-		Parsely_Loader_Info::set_integration_type( Parsely_Loader_Info::INTEGRATION_TYPE_NONE );
+
+		if ( false === $filtered_load_status ) {
+			Parsely_Loader_Info::set_integration_type( Parsely_Integration_Type::DISABLED_MUPLUGINS_FILTER );
+		} elseif ( '0' === $option_load_status ) {
+			Parsely_Loader_Info::set_integration_type( Parsely_Integration_Type::DISABLED_MUPLUGINS_SILENT_OPTION );
+		}
```
WPprodigy, line 219: Are these the only two possibilities? Should there be a fallback to `INTEGRATION_TYPE_NONE` in case neither conditional happens?

WPprodigy (dismissed):
Looks good. Just left one quick comment.

wp-parsely.php
mehmoodak: In [get_integration_type](https://github.com/Automattic/vip-go-mu-plugins/blob/sends-updated-parsely-config-states-to-SDS/wp-parsely.php#L50) function there is already a fallback to select `NONE` if there is no integration type.

wp-parsely.php
```diff
@@ -67,6 +62,59 @@ public static function set_version( string $version ): void {
...
+		$post_names                    = get_post_types( array( 'public' => true ) );
+		$tracked_post_types            = $options['track_post_types'] ?? array();
+		$tracked_page_types            = $options['track_page_types'] ?? array();
+		foreach ( $post_names as $post_name ) {
+			$tracked_post_type         = array();
+			$tracked_post_type['name'] = $post_name;
+
+			if ( in_array( $post_name, $tracked_post_types ) ) {
+				$tracked_post_type['track_type'] = 'post';
+			} elseif ( in_array( $post_name, $tracked_page_types ) ) {
```
mjangda:
```suggestion
		$post_types                    = get_post_types( array( 'public' => true ) );
		$tracked_post_types            = $options['track_post_types'] ?? array();
		$tracked_page_types            = $options['track_page_types'] ?? array();
		foreach ( $post_types as $post_type ) {
			$tracked_post_type         = array();
			$tracked_post_type['name'] = $post_type;

			if ( in_array( $post_type, $tracked_post_types ) ) {
				$tracked_post_type['track_type'] = 'post';
			} elseif ( in_array( $post_type, $tracked_page_types ) ) {
```

wp-parsely.php
```diff
@@ -257,3 +310,21 @@ function maybe_disable_some_features() {
...
+/**
+ * Enum which represent all options to integrate `wp-parsely`.
+ */
+// phpcs:ignore Generic.Files.OneObjectStructurePerFile.MultipleFound
+abstract class Parsely_Integration_Type {
+	const ENABLED_MUPLUGINS_FILTER        = 'ENABLED_MUPLUGINS_FILTER';
+	const ENABLED_MUPLUGINS_SILENT_OPTION = 'ENABLED_MUPLUGINS_SILENT_OPTION';
+
+	const DISABLED_MUPLUGINS_FILTER        = 'DISABLED_MUPLUGINS_FILTER';
+	const DISABLED_MUPLUGINS_SILENT_OPTION = 'DISABLED_MUPLUGINS_SILENT_OPTION';
```
mjangda, line 323: Since the string value of these is changing (in a good way), we'll need to make sure that other systems are aware of the change. Most notably, we may need to update the reports in Gainsight to check the new values once they are syncing over.

mjangda (dismissed):
One minor variable name change; otherwise, this is good to go 👍

PR #4283, opened 2023-03-29
Search: Do not automatically swap index versions

rebeccahum:
## Description
In a case where manual commands are being run at the same time as the cron job to rebuild a new index version with an incorrect number of shards, there's a chance of the index being swapped accidentally in a scenario where the indexing has already completed on the cron job, but the cron job to swap it has not run yet. This allows a small window for a user to delete the already-completed index and create a new one in the background. While the new one is being built in the background, at any given chance when the cron job runs, it can be swapped since it was marked as "completed".

Since most indexes have the correct shards now, we should disable the automatic swap functionality.

Remove automatic swapping in cron job to rebuild index with incorrect number of shards.

## Checklist

## Steps to Test
0) Create a new index version `wp vip-search index-versions add post` and note results from `wp vip-search index-versions list post`
1)
```
$es->init();
$indexable = \ElasticPress\Indexables::factory()->get('post');
$hj->alert_to_swap_index_versions( \ElasticPress\Indexables::factory()->get($indexable->slug);
```
2) Check alert channel and check that `wp vip-search index-versions list post` remains the same

### DIFF
--- a/search/includes/classes/class-settingshealthjob.php
+++ b/search/includes/classes/class-settingshealthjob.php
@@ -326,7 +326,7 @@ class SettingsHealthJob {
 					}
 					break;
 				case 'swap':
-					$this->swap_index_versions( $indexable );
+					$this->alert_to_swap_index_versions( $indexable );
 					break;
 				case 'in-progress':
 				default:
@@ -423,6 +423,18 @@ class SettingsHealthJob {
 		$cmd->index( $args, $assoc_args );
 
 		update_option( self::LAST_PROCESSED_ID_OPTION, 'Indexing completed' );
+
+		\Automattic\VIP\Logstash\log2logstash(
+			array(
+				'severity' => 'info',
+				'feature'  => 'search_versioning',
+				'message'  => 'Built new index to meet shard requirements',
+				'extra'    => [
+					'homeurl'   => home_url(),
+					'indexable' => $indexable->slug,
+				],
+			)
+		);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


search/includes/classes/class-settingshealthjob.php
```diff
@@ -423,6 +423,18 @@ public function build_new_index( $indexable_slug, $last_processed_id = false ) {
...
 		update_option( self::LAST_PROCESSED_ID_OPTION, 'Indexing completed' );
+
+		\Automattic\VIP\Logstash\log2logstash(
+			array(
+				'severity' => 'info',
+				'feature'  => 'search_versioning',
+				'message'  => 'Built new index to meet shard requirements',
+				'extra'    => [
+					'homeurl'   => home_url(),
+					'indexable' => $indexable->slug,
```
rinatkhaziev, line 434: let's add index name here and `gethostname()` so that we can better figure out which container was responsible.

search/includes/classes/class-settingshealthjob.php
```diff
@@ -447,50 +459,18 @@ protected function check_process_build() {
...
-				'feature'  => 'search_versioning',
-				'message'  => 'Built new index to meet shard requirements',
-				'extra'    => [
-					'homeurl'   => home_url(),
-					'indexable' => $indexable->slug,
-				],
-			)
+	public function alert_to_swap_index_versions( $indexable ) {
+		$message = sprintf(
+			'Application %s: New index built to meet shard requirements. Please activate %s index to new version for %s after validating index health with `wp vip-search health validate-counts`.',
```
rinatkhaziev: Let's add old an new index name here.

rinatkhaziev (changes requested):
Thanks, left a few notes.

search/includes/classes/class-settingshealthjob.php
```diff
@@ -447,50 +461,35 @@ protected function check_process_build() {
...
 				home_url(),
-				$delete_version->get_error_message()
+				$new_version->get_error_message()
 			);
 			$this->send_alert( '#vip-go-es-alerts', $message, 2 );
 
 			return;
 		}
 
-		\Automattic\VIP\Logstash\log2logstash(
```
rinatkhaziev, line 483: What's the intent here with this removal? I think it's still good to have that info for audit.

rebeccahum: It's not being removed, it was just being moved up for right after the indexing completes

rinatkhaziev (approved):

PR #4291, opened 2023-03-29
Search: Change add_version() to account for failed put_mapping

rebeccahum:
## Description
Currently, `add_version()` assumes that `put_mapping()` will succeed. We should return a WP_Error if it doesn't though.

Search: Account for failed put_mapping call in add_version()
## Checklist

## Steps to Test
1) Go to `search/elasticpress-next/includes/classes/Elasticsearch.php` and change `put_mapping()` to return `false`
2) Attempt to create new index version `wp vip-search index-versions add post` and see failed error message:
```
Error: Unable to properly create new index with mapping
```


pschoffer (changes requested):
Great PR overall. I like all the new tests in particular.

One small thing I would raise is the move of `validate_post_index_mapping` from health to versioning functions. 

Was the main reason to colocate it with the only code that uses it? To me, that logic feels to fit more into the "health" class than "versioning" class. In case we would like to later do this check for other reason, it would be tough to find in versioning I think.

rebeccahum:
> Was the main reason to colocate it with the only code that uses it? 

@pschoffer Yep. You do bring a good point since it does technically belong in the health class semantically 🤔 . I've moved it back, thanks!

pschoffer (approved):

PR #4295, opened 2023-03-30
Remove SyncManager hooks if an indexable doesn't have an active index

rinatkhaziev:
## Description

To prevent Sync Manager from ever firing on non-existing indexes (thus forcing their creation with default mappings) this PR introduces a check that goes over each indexable and disables the sync hooks by calling `disable_sync_hooks`. That method has to be implemented in each Indexable's SyncManager.

## Checklist

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -2257,6 +2257,12 @@ class Search {
 
 	public function action__init() {
 		remove_action( 'wp_initialize_site', [ \ElasticPress\Indexables::factory()->get( 'post' )->sync_manager, 'action_create_blog_index' ] );
+		// Force disable sync hooks for indexables that don't have an index, this prevents the sync manager from creating the index with broken mappings
+		foreach ( \ElasticPress\Indexables::factory()->get_all() as $indexable ) {
+			if ( ! $indexable->index_exists() && is_callable( [ $indexable->sync_manager, 'disable_sync_hooks' ] ) ) {
+				$indexable->sync_manager->disable_sync_hooks();
+			}
+		}
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


rebeccahum:
How come you decided on this approach versus patching it in https://github.com/Automattic/ElasticPress/blob/e166010fc9f980cd7ce5f9889bbd267f8fd4042f/includes/classes/SyncManager.php#L203?

Something like:
```
	public function can_index_site() {
		$indexable = Indexables::factory()->get( $this->indexable_slug );
		if ( $indexable && ! $indexable->index_exists() ) {
			$this->tear_down();
			return false;
		}

		return true;
	}
```

rinatkhaziev:
@rebeccahum 

> How come you decided on this approach versus patching it in https://github.com/Automattic/ElasticPress/blob/e166010fc9f980cd7ce5f9889bbd267f8fd4042f/includes/classes/SyncManager.php#L203?
> 
> Something like:
> 
> ```
> 	public function can_index_site() {
> 		$indexable = Indexables::factory()->get( $this->indexable_slug );
> 		if ( $indexable && ! $indexable->index_exists() ) {
> 			$this->tear_down();
> 			return false;
> 		}
> 
> 		return true;
> 	}
> ```

That's what I tried to do initially, unfortunately it doesn't work. By the time of this call the indexable is not registered yet. It was a mess. If you can make it work like that - I have no quarrel.

rebeccahum:
Makes sense, thanks! Do we know how to reproduce the creating of an index w/ bad mapping on the newest EP version? This is what I tried:
1) Create subsite 2
2) `wp post generate --count=3`
3) `wp vip-search get-indexes` to check if there were any newly created indexes

JW if this PR is still necessary if it's not an occurrence anymore

rinatkhaziev:
@rebeccahum as long as those hooks are still getting added I think adding a second layer of protection is valuable?

rebeccahum:
Closing b/c redundancy from https://github.com/Automattic/vip-go-mu-plugins/blob/f85089c9fae7ef0eb1bf3b69ef1f5010582a056f/search/includes/classes/class-search.php#L2250

PR #4297, opened 2023-03-31
Search e2e: Wait for DOM elements to be loaded in openWidgetsPage()

rebeccahum:
## Description
I prefer this approach over #4296, where we are forcing a click, but in reality, all the DOM elements haven't loaded yet. 🙂 It does slow the test down by a tad, but better for accuracy-wise.

### DIFF
--- a/tests/search/e2e/support/commands.js
+++ b/tests/search/e2e/support/commands.js
@@ -52,6 +52,7 @@ Cypress.Commands.add('visitAdminPage', (page = 'index.php') => {
 Cypress.Commands.add('openWidgetsPage', () => {
 	cy.login();
 	cy.visitAdminPage('widgets.php');
+	cy.wait(3000); // Wait for all DOM elements to be loaded
 	cy.get('body').then(($body) => {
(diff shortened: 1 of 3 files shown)


tests/search/e2e/support/commands.js
```diff
@@ -52,6 +52,7 @@ Cypress.Commands.add('visitAdminPage', (page = 'index.php') => {
 Cypress.Commands.add('openWidgetsPage', () => {
 	cy.login();
 	cy.visitAdminPage('widgets.php');
+	cy.wait(3000); // Wait for all DOM elements to be loaded
```
sjinks, line 55: I am surprised that Cypress does not have a function to wait for DOMContentLoaded :-(

sjinks (approved):

tests/search/e2e/support/commands.js
rebeccahum: There is a way with chaining the `.should()` but that only works if that button element is there (which it may or may not be)

PR #4306, opened 2023-04-03
Fix fatal error in "widget updated" event

acicovic:
## Description
@bratvanov recently reported a fatal error that was occurring in our code in the `widget updated` event:

> We are getting a fatal error in a preproduction environment when attempting to save a custom widget. The stack trace shows `false` is passed to the `track_widget_updated` function but it expects an `array`.
> 
> This looks to be an issue with our code in mu plugins. The code didn’t account for the case where we could pass `false` to short-circuit the update. It just assumes the `$instance` is always an `array`.
> 
> https://developer.wordpress.org/reference/hooks/widget_update_callback/

Reading the docs above, although the `$instance` variable is typed as an array, the docs [explicitly state](https://developer.wordpress.org/reference/hooks/widget_update_callback/#description):
> Returning `false` will effectively short-circuit the widget’s ability to update settings.

Since our docs mention the `false` value as a way to short-circuit the update, we should be supporting it. This PR makes the function more flexible to receiving a `false` value, and returning if the `$instance` variable is not an array.

### Fixed error in the "widget updated" event

We fixed a fatal error that could occur in the "widget updated" event when attempting to save a custom widget.

## Checklist

### DIFF
--- a/vip-parsely/Telemetry/Events/track-widget-updated.php
+++ b/vip-parsely/Telemetry/Events/track-widget-updated.php
@@ -14,14 +14,18 @@ use WP_Widget;
 /**
  * Records an event using the given Telemetry System whenever a `parsely_recommended_widget` instance is updated.
  *
- * @param array $instance The current widget instance's settings.
+ * @param array|false $instance The current widget instance's settings.
  * @param array|null $new_instance Array of new widget settings.
  * @param array|null $old_instance Array of old widget settings.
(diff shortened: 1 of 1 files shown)


acicovic:
Let me know if you would like me to modify this so sonarcloud doesn't complain about 4 returns in the function.

rebeccahum (approved):

PR #4312, opened 2023-04-03
Don't add objects to queue if index doesn't exist

rinatkhaziev:
## Description

There's no particularly good reason to add anything to queue if an index doesn't exist. It means that the initial set up was not performed yet. It also has negative implications (a default mapping might be used which leaves the index in a broken state).

This PR addresses that by adding checks both to `queue_object` and `queue_objects` method to prevent pushing anything into queue if an index doesn't exist.

Improved handling of async queue in scenarios where an index doesn't exist yet.

## Checklist

## Steps to Test

1. Have Search active but don't do the set up steps
2. Use `wp post generate` to generate a few hundred of posts
3. Edit the `Uncategorized` term
4. Verify the queue is at `0` 
5. Run `wp vip-search index --setup` 
6. Repeat the steps and verify that the queue is not empty anymore.

### DIFF
--- a/search/includes/classes/class-queue.php
+++ b/search/includes/classes/class-queue.php
@@ -385,6 +385,12 @@ class Queue {
 			return;
 		}
 
+		$indexable = Indexables::factory()->get( $indexable_slug );
+
+		if ( ! $indexable || ! \ElasticPress\Elasticsearch::factory()->index_exists( $indexable->get_index_name() ) ) {
+			return;
+		}
+
 		foreach ( $object_ids as $object_id ) {
 			$this->queue_object( $object_id, $indexable_slug, $options );
 		}
--- a/search/includes/classes/class-queue.php
+++ b/search/includes/classes/class-queue.php
@@ -313,6 +313,10 @@ class Queue {
 			return new WP_Error( 'invalid-indexable', sprintf( 'Indexable not found for type %s', $indexable_slug ) );
 		}
 
+		if ( ! \ElasticPress\Elasticsearch::factory()->index_exists( $indexable->get_index_name() ) ) {
+			return new WP_Error( 'index-not-exists', sprintf( 'Index not found for type %s', $indexable_slug ) );
+		}
+
 		global $wpdb;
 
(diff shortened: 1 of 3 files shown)


rinatkhaziev:
Uhh, needs test fixes

search/includes/classes/class-queue.php
```diff
@@ -313,6 +313,10 @@ public function queue_object( $object_id, $indexable_slug = 'post', $options = a
 			return new WP_Error( 'invalid-indexable', sprintf( 'Indexable not found for type %s', $indexable_slug ) );
 		}
 
+		if ( ! \ElasticPress\Elasticsearch::factory()->index_exists( $indexable->get_index_name() ) ) {
```
rebeccahum: We can probably remove the check in `queue_objects()` since it calls `queue_object()`, yeah?

Edit: Nvm, I think this is probably an extra mechanism for "failing faster" since I'm seeing both called individually in different pieces of the code.

search/includes/classes/class-queue.php
```diff
@@ -381,6 +385,12 @@ public function queue_objects( $object_ids, $indexable_slug = 'post', $options =
 			return;
 		}
 
+		$indexable = Indexables::factory()->get( $indexable_slug );
+
+		if ( ! $indexable || ! \ElasticPress\Elasticsearch::factory()->index_exists( $indexable->get_index_name() ) ) {
```
rebeccahum:
```suggestion
		if ( ! $indexable || ! $indexable->index_exists() ) {
```

For simplification since it calls `\ElasticPress\Elasticsearch::factory()->index_exists( $indexable->get_index_name() )` already

search/includes/classes/class-queue.php
rebeccahum:
```suggestion
		if ( ! $indexable->index_exists() ) {
```

rebeccahum (changes requested):

sjinks (dismissed):

sjinks (approved):

PR #4325, opened 2023-04-06
Disables Parse.ly if VIP_PARSELY_SKIP_LOAD or VIP_IS_FEDRAMP constant is set

mehmoodak:
## Description
- We should disable the plugin based on `VIP_IS_FEDRAMP` constant which indicates that the current site is fedramp
- We should disable the plugin based on `VIP_PARSELY_SKIP_LOAD` constant which is a new opt-out mechanism
  - We will set this constant based on the meta attributes of client or site
  - After this PR we will have three ways of enabling/disabling Parse.ly ( via filter, option and this constant ) but in future we will only use this constant and migrate the users from other two approaches to make enabling/disabling simpler.

## Changelog Description

Disables Parse.ly if `VIP_PARSELY_SKIP_LOAD` or `VIP_IS_FEDRAMP` constant is set

## Checklist

## Steps to Test

1. Pull branch

3. Enable `wp-parsely` by following [VIP Docs](https://docs.wpvip.com/how-tos/enabling-and-disabling-parsely/)

4. Verify these cases separately

    - Open `vip-config.php` file, adds `define( 'VIP_PARSELY_SKIP_LOAD', true );` and verify that plugin is not loading.

    - Open `vip-config.php` file, adds `define( 'VIP_IS_FEDRAMP', true );` and verify that plugin is not loading.

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -186,11 +186,8 @@ function maybe_load_plugin() {
 		return;
 	}
 
-	$is_disabled_via_constant = defined( 'VIP_PARSELY_SKIP_LOAD' ) && true === constant( 'VIP_PARSELY_SKIP_LOAD' );
-	$is_fedramp               = defined( 'VIP_IS_FEDRAMP' ) && true === constant( 'VIP_IS_FEDRAMP' );
-
-	// Allow opting out via the VIP_PARSELY_SKIP_LOAD or VIP_IS_FEDRAMP constants.
-	if ( $is_disabled_via_constant || $is_fedramp ) {
+	// Allow opting out via the VIP_PARSELY_SKIP_LOAD constant
+	if ( defined( 'VIP_PARSELY_SKIP_LOAD' ) && true === constant( 'VIP_PARSELY_SKIP_LOAD' ) ) {
 		Parsely_Loader_Info::set_active( false );
 		Parsely_Loader_Info::set_integration_type( Parsely_Integration_Type::DISABLED_CONSTANT );
 
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -13,8 +13,6 @@
 
 namespace Automattic\VIP\WP_Parsely_Integration;
 
-use stdClass;
-
 // The default version is the first entry in the SUPPORTED_VERSIONS list.
 const SUPPORTED_VERSIONS = [
 	'3.8',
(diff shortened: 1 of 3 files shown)


wp-parsely.php
```diff
@@ -188,6 +186,17 @@ function maybe_load_plugin() {
 		return;
 	}
 
+	$is_disabled_via_constant = defined( 'VIP_PARSELY_SKIP_LOAD' ) && true === constant( 'VIP_PARSELY_SKIP_LOAD' );
+	$is_fedramp               = defined( 'VIP_IS_FEDRAMP' ) && true === constant( 'VIP_IS_FEDRAMP' );
```
mjangda: For the `VIP_IS_FEDRAMP` constant, we need to a take a similar approach to Jetpack: https://github.com/Automattic/vip-go-mu-plugins/blob/9fead4d530e81177b7efaec5acf32588ee37fd62/000-vip-init.php#L131-L134

I would put something like this next to the Jetpack one for now:

```
if ( Context::is_fedramp() ) {
    if ( ! defined( 'VIP_PARSELY_SKIP_LOAD' ) ) {
        define( 'VIP_PARSELY_SKIP_LOAD', true );
    }
}
```

By default, this will skip loading Parse.ly. But, FR customers can override the behaviour by defining the constant as false in their vip-config.php file (`define( 'VIP_PARSELY_SKIP_LOAD', false );`, which allows them to use the plugin.

mehmoodak: I was initially doing it this way but was not sure how to properly test this. Based on the feedback I am incorporating this change but please let me know how can I write test for it properly .... any existing example that can tests the logic we are going to put in `000-vip-init.php` file?

mehmoodak:
@mjangda  feedback incorporated. Please have a look again.

wp-parsely.php
mjangda: I think the a good approach would be to move the code to a function and then test the function. 

@sjinks might have some better suggestions as well.

mjangda (approved):
LGTM

wp-parsely.php
mehmoodak: `Context::is_fedramp()` is already a function ....... It is getting called by `000-vip-init.php` file but how can I mock it 🤔

mehmoodak: For now I am merging the PR, will add the test once I get the response or figure out the way myself.

mjangda: Take a look at `Constant_Mocker` (tests/files/acl/test-acl.php) or the `createMock` method (tests/admin-notice/test-class-admin-notice.php).

mehmoodak: I did looked into those but those didn't worked and most probably the reason is that in `000-vip-init.php` file is getting called very early (also in tests) which means this setting up `VIP_PARSELY_SKIP_LOAD` constant decision is taking place very early so these mocks or functions cannot help.

[https://github.com/Automattic/vip-go-mu-plugins/blob/develop/tests/bootstrap.php#L27](https://github.com/Automattic/vip-go-mu-plugins/blob/develop/tests/bootstrap.php#L27)

PR #4331, opened 2023-04-07
Don’t load new cache drop-in on legacy platform

WPprodigy:
It doesn't sound like we'll be getting the new PHP extension on the legacy platform, so we can just avoid loading the new drop-in for those remaining sites and they can continue using what they currently do.

### DIFF
--- a/drop-ins/object-cache.php
+++ b/drop-ins/object-cache.php
@@ -18,7 +18,7 @@ if ( ! defined( 'VIP_USE_NEXT_OBJECT_CACHE_DROPIN' ) ) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


rinatkhaziev:
@WPprodigy should we do the extension_loaded check as a part of this PR?

WPprodigy:
I'm thinking that part goes into the plugin itself. Though tbf, there's nothing of the sort for `Memcache` extension checks either.

It won't really be a concern moving forward platform-wise, but some bootstrapping checks are always pretty nice.

rinatkhaziev (approved):

PR #4342, opened 2023-04-11
Add VIP_PARSELY_ENABLED constant to enable / disable wp-parsely plugin

mehmoodak:
## Description

- Added `ENABLED_CONSTANT` integration_type (also renamed `BLOCKED_CONSTANT` to `DISABLED_CONSTANT` because it matches with `ENABLED_*` and `DISABLED_*` naming convention)

- Using `VIP_PARSELY_ENABLED` constant and possible scenarios of this constant are:

    - If not defined then it means parsely is neither enabled nor blocked (look enablement via filter or option)
    - If `true` then means parsely is enabled and integration_type should be `ENABLED_CONSTANT`
    - If not `true` (false or anything else) then means parsely is disabled and integration_type should be `DISABLED_CONSTANT`

## Changelog Description

Add VIP_PARSELY_ENABLED constant and remove VIP_PARSELY_SKIP_LOAD constant

## Checklist

## Steps to Test

1. Pull branch

3. Enable `wp-parsely` by following [VIP Docs](https://docs.wpvip.com/how-tos/enabling-and-disabling-parsely/)

4. Verify these cases separately

    - Open `vip-config.php` file, adds `define( 'VIP_PARSELY_ENABLED', false );` and verify that plugin is not loading.

    - Open `vip-config.php` file, adds `define( 'VIP_IS_FEDRAMP', true );` and verify that plugin is not loading.
    
    - Revert changes of point 3, open `vip-config.php` file, adds `define( 'VIP_PARSELY_ENABLED', true );` and verify that plugin is loading.

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -186,12 +186,21 @@ function maybe_load_plugin() {
 		return;
 	}
 
-	// Allow opting out via the VIP_PARSELY_SKIP_LOAD constant
-	if ( defined( 'VIP_PARSELY_SKIP_LOAD' ) && true === constant( 'VIP_PARSELY_SKIP_LOAD' ) ) {
-		Parsely_Loader_Info::set_active( false );
-		Parsely_Loader_Info::set_integration_type( Parsely_Integration_Type::BLOCKED_CONSTANT );
+	$parsely_enabled_constant = null; // Represents that the site doesn't have parsely enabled / blocked.
 
-		return;
+	if ( defined( 'VIP_PARSELY_ENABLED' ) ) {
+		$parsely_enabled_constant = constant( 'VIP_PARSELY_ENABLED' );
+
+		// Opt out if constant value isn't true
+		if ( true !== $parsely_enabled_constant ) {
+			Parsely_Loader_Info::set_active( false );
+			Parsely_Loader_Info::set_integration_type( Parsely_Integration_Type::DISABLED_CONSTANT );
+
+			return;
+		}
+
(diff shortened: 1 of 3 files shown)


rebeccahum:
Do we need to revert https://github.com/Automattic/vip-go-mu-plugins/pull/4325 in `staging`?

mehmoodak:
> Do we need to revert https://github.com/Automattic/vip-go-mu-plugins/pull/4325 in staging?

@rebeccahum 

1. We can keep it because there is no side effect.
2. Review is yet to happen on this PR (there could have been a possibility to keep both)

rebeccahum:
We just want to prevent anyone from adopting the `VIP_PARSELY_SKIP_LOAD` constant though -- so, I'm thinking:
1) If this PR is finalized before next Tuesday, it can go out to `staging`
2) If not 1), revert for the time being

mehmoodak:
Okay then Option 2 is fine i.e. revert for time being.

wp-parsely.php
```diff
@@ -205,26 +220,29 @@ function maybe_load_plugin() {
 	$option_load_status   = get_option( '_wpvip_parsely_mu', null );
 	$filtered_load_status = apply_filters( 'wpvip_parsely_load_mu', null );
 
-	$should_load            = true === $filtered_load_status || '1' === $option_load_status;
```
mehmoodak, line 208: Just some refactoring to move the code inside `true !== $parsely_enabled_constant` condition

mjangda (approved):
🚢

PR #4355, opened 2023-04-12
Search: Fix syntax error in confirming update active index version when --using-versions

rebeccahum:
## Description
Since `sprintf()` requires 2 arguments


sjinks (approved):
Nice catch!

PR #4363, opened 2023-04-14
Fix wp-parsely track info in self managed mode

mehmoodak:
## Description
If customer is self managing Parsely and we set `VIP_PARSELY_ENABLED` constant as `false` on our side then in this case we should track the plugin as active and in self managed mode which was not happening.

## Changelog Description
Fixed wp-parsely track info in self managed mode

## Checklist

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -186,6 +186,20 @@ function maybe_load_plugin() {
 		return;
 	}
 
+	// Self-managed integration: The plugin exists on the site and is being loaded already.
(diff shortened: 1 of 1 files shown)


mjangda (approved):
Good catch 👍 

Should we have a test case for this? (Can be in a separate PR)

mehmoodak:
> Should we have a test case for this? (Can be in a separate PR)

I tried adding it but there were lot of changes and some tricky code to make it work which wasn't feeling good so removed it. As of now we have multiple scenarios of enabling/disabling the plugin which was making it more difficult.

PR #4380, opened 2023-04-20
CI: Do not add Reverted PRs to changelog summary

rebeccahum:
## Description
For PRs that get reverted, we should add the `[Status] Reverted` label to it. Once we add it, the CI for the changelog summary should not pick it up.


sjinks (approved):
👍🏽

PR #4387, opened 2023-04-25
fix(wp-cli): Properly detect local environments

sjinks:
## Description

This PR fixes the detection of local environments. Because WP CLI initialized before `wp-config.php` gets loaded, `VIP_GO_APP_ENVIRONMENT`, `WP_ENVIRONMENT_TYPE`, `WPVIP_ENABLE_WP_DB`, and `WPVIP_ENABLE_WP_DB_WRITES` constants are not available in `Automattic\VIP\Helpers\WP_CLI_DB\Config::__construct()`.

### Plugin Updated: WP-CLI for VIP Go

Fix detection of local environments.

## Checklist

## Steps to Test

Create and start dev env:

```sh
vip dev-env create < /dev/null
vip dev-env start
```

Run a shell: `vip dev-env shell`

Alternatively, you can use a codespace instead of `vip dev-env create/start/shell`.

Try to run `wp db import`. It will fail with "Error: The db command is not currently supported in this environment."

Apply the patch and see the error go away.

### DIFF
--- a/lib/helpers/wp-cli-db/class-config.php
+++ b/lib/helpers/wp-cli-db/class-config.php
@@ -6,39 +6,49 @@ use Automattic\VIP\Environment;
 use Exception;
 
 class Config {
-	private bool $enabled      = false;
-	private bool $allow_writes = false;
-	private bool $is_sandbox   = false;
-	private bool $is_local     = false;
-	private bool $is_batch     = false;
-
-	public function __construct() {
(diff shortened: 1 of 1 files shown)


rebeccahum:
Hiya, tried to follow your reproduction steps and got:
```
╰─$ vip dev-env --slug vip-local shell -- wp db import
Error: Import file missing or not readable: wordpress.sql
```

sjinks:
Yeah, that's fine - you need to specify the name of the file if you really want to import something.

The point is that without this patch, you wouldn't get this far - the system would tell you that the db import command is not allowed, which should not happen in local dev environments or codespaces.

rebeccahum:
To clarify, that's the reproduction steps without the patch 🙂

PR #4405, opened 2023-04-26
Update qm-plugins to be compat with new query-monitor version

rebeccahum:
## Description
We updated Query Monitor in #4383.

However, Query Monitor has a new `QM_Data` DTO that must be used or else you get an error saying "Indirect modification of overloaded element has no effect". There should be no changes in functionality.

Also changes filenames and class names for more consistency. Added a `file_exists` check to account for non-atomic deployments.

Use new QM_Data DTO.

## Checklist

## Steps to Test
1) Test all the panels that qm-plugins adds to ensure that functionality remains the same


qm-plugins/qm-alloptions/class-qm-output-alloptions.php
```diff
@@ -4,7 +4,7 @@
  *
  * Class QM_Output_AllOptions
  */
-class QM_Output_AllOptions extends QM_Output_Html {
+class QM_Output_Html_AllOptions extends QM_Output_Html {
```
GaryJones, line 7: If the class name changed, does the filename also need to be updated?

(Here and for other files too)

sjinks (approved):

PR #4406, opened 2023-04-27
Add integration code for block data API

alecgeatches:
## Description

This adds a `vip-integrations.php` file and two constants for enabling the [VIP Block Data API](https://github.com/Automattic/vip-block-data-api/):

- `VIP_BLOCK_DATA_API_ENABLE` (`true` or `false`, default disabled)
- This was removed per @mehmoodak's comments in https://github.com/Automattic/vip-go-mu-plugins/pull/4406/commits/b73e9dfa26899ddc89f5e12455d8548d89d48100: ~`VIP_BLOCK_DATA_API_VERSION` (pinned version number from [vip-go-mu-plugins-ext](https://github.com/Automattic/vip-go-mu-plugins-ext), defaults to latest version of `vip-block-data-api`)~

These changes go along with [this PR in -ext](https://github.com/Automattic/vip-go-mu-plugins-ext/pull/30) to add the Block Data API to the `vip-integrations/` folder.

The `PluginLoadStrategy` used is largely to ensure that mu-plugins has the ability to decide load order if a plugin is both installed manually and enabled via constant. This is set to `PREFER_CUSTOMER_CODE_VERSION` currently which means if the Block Data API plugin is manually installed, that version will override the constant-enabled version.

Currently all of the code in `vip-integrations.php` is focused on loading vip-block-data-api, but this can later be generalized to handle other external plugin loading.

### Added constant to enable the VIP Block Data API plugin

Added support to enable the [VIP Block Data API plugin](https://github.com/Automattic/vip-block-data-api/) via the `VIP_BLOCK_DATA_API_ENABLE` constant.

## Checklist

- [ ] (?) This change has relevant unit tests (if applicable).
- [x] N/A - This change uses a rollout method to ease with deployment.
- [x] External documentation in progress - This change has relevant documentation additions / updates (if applicable).

## Steps to Test

1. Wait until a `vip-go-mu-plugins` deployment is made, which should pull [the `vip-integrations` folder from `-ext`](https://github.com/Automattic/vip-go-mu-plugins-ext/tree/trunk/vip-integrations) into [vip-go-mu-plugins-built](https://github.com/Automattic/vip-go-mu-plugins-built).
2. Clone [vip-go-mu-plugins-built](https://github.com/Automattic/vip-go-mu-plugins-built) locally set `vip dev-env`'s environment to use it, e.g.

    ```bash
    vip dev-env update --mu-plugins=~/projects/mu-plugins-built
    ```

3. Verify the block data API plugin is disabled by default. Visit the REST endpoint for an existing post ID:

    ```bash
    GET /wp-json/vip-block-data-api/v1/posts/<post_id>/blocks

    # e.g. GET http://my-site.vipdev.lndo.site/wp-json/vip-block-data-api/v1/posts/1/blocks
    ```

    Ensure you receive a `404`response with status code `rest_no_route`.

4. In `vip-config.php`, add:

    ```php
    define( 'VIP_BLOCK_DATA_API_ENABLE', true );
    ```

6. Visit the same URL as step 3, and ensure the block data API is now available:

    ```
    GET http://my-site.vipdev.lndo.site/wp-json/vip-block-data-api/v1/posts/1/blocks

    {"blocks":[{"name":"core/paragraph","attributes":{"content":"Welcome to WordPress. This is your first post. Edit or delete it, then start writing!","dropCap":false}}]}
    ```

Follow section removed via [this conversation](https://github.com/Automattic/vip-go-mu-plugins/pull/4406#discussion_r1179942240).

<details>

<summary>(removed) The `PluginLoadStrategy` code can be tested by:</summary>

1. Disable the mu-plugins version of the block data API if enabled:

    ```php
    define( 'VIP_BLOCK_DATA_API_ENABLE', false );
    ```

2. Install [vip-block-data-api](https://github.com/Automattic/vip-block-data-api/) manually in your site's plugins and enable it.
3. Modify the locally installed version of the plugin by adding an early return in [`src/rest/rest-api.php`'s `get_block_content()` function](https://github.com/Automattic/vip-block-data-api/blob/585e000/src/rest/rest-api.php#L83):

    ```php
    public static function get_block_content( $params ) {
        return ['local_plugin' => true];
        // ...
    }
    ```

4. Retrieve a post's block content via the rest API:

    ```
    GET http://my-site.vipdev.lndo.site/wp-json/vip-block-data-api/v1/posts/1/blocks
    ```

    Ensure this shows the local plugin `{'local_plugin' => true}` response.

5. Enable the mu-plugins constant for the plugin:

    ```php
    define( 'VIP_BLOCK_DATA_API_ENABLE', true );
    ```

7. Query the REST URL from step 4. Ensure that the `{'local_plugin' => true}` response is still present. This should be the case when `$plugin_load_strategy` is set to `PREFER_CUSTOMER_CODE_VERSION`. 
8. In `vip-integrations.php`, change `$plugin_load_strategy` to `PREFER_MU_PLUGINS_VERSION`:

    ```diff
    - $plugin_load_strategy = PluginLoadStrategy::PREFER_CUSTOMER_CODE_VERSION;
    + $plugin_load_strategy = PluginLoadStrategy::PREFER_MU_PLUGINS_VERSION;
    ```
9. Query the REST URL from step 4 again. This time, it should use the mu-plugins version of the REST API and return the block's contents.

</details>


mjangda:
> Currently all of the code in vip-integrations.php is focused on loading vip-block-data-api, but this can later be generalized to handle other external plugin loading.

Something I mentioned to @mehmoodak is to think about how we can abstract the logic into a `VIP_Integration` base class which can then be extended by each integration.

```
abstract class VIP_Integration_Base {
    abstract public function is_loaded() {}

    public function get_load_strategy() {
        // check status constant?
    }

    public function load() {
        // figure out version

        $file_path = __DIR__ . '/' . $this->slug . '-' . $plugin_version . '/' . $this->slug . '.php';
        if ( is_readable( $file_path ) ) {
            require_once( $file_path );
        }
    }
}

class Block_Data_Api_Integration extends VIP_Integration_Base {
    public $name = 'Block Data API';
    public $description = 'Your blocks, JSONified!';

    public function is_loaded() {
        return defined( 'VIP_BLOCK_DATA_API_LOADED' );
    }
}

class Parsely_Integration extends VIP_Integration_Base {
    public function is_loaded() {
        return class_exists( 'Parsely' );
    }
}

register_integration( 'block-data-api', new Block_Data_Api_Integration );
register_integration( 'parsely', new Parsely_Integration );

    foreach ( $registered_integrations as $integration ) {
        if ( $integration->is_loaded() ) {
            // Mark as active and loader strategy as customer managed for reporting
        }

         $load_strategy = $integration->get_load_strategy();
        if ( $load_strategy === 'blocked' ) {
            // Mark as inactive and loader strategy as blocked for reporting
            // return
        }

        $integration->load();
    }
} );
```

This is not a blocker for this PR, but something we should plan out.

vip-integrations.php
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+namespace Automattic\VIP\Integrations;
+
+if ( ! ( defined( 'VIP_BLOCK_DATA_API_ENABLE' ) && true === constant( 'VIP_BLOCK_DATA_API_ENABLE' ) ) ) {
+	// Do not automatically activate a plugin unless specified.
+	return;
+}
+
+const VIP_BLOCK_API_SUPPORTED_VERSIONS = [
```
mehmoodak: Instead of supporting multiple versions is this possible to only support one version? I think it will make things easier.

vip-integrations.php
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Plugin Name: VIP Integrations
+ * Description: Plugin loading for VIP plugin integrations.
+ * Author: Automattic
+ * License: GPL version 2 or later - http://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+ */
+
+namespace Automattic\VIP\Integrations;
+
+if ( ! ( defined( 'VIP_BLOCK_DATA_API_ENABLE' ) && true === constant( 'VIP_BLOCK_DATA_API_ENABLE' ) ) ) {
```
mehmoodak, line 12: The code in this file seems very specific to block data api so can we make it generic so that it will be easier to add our future integrations easier?

(Optinal) It will be great if we follow a strategy considering Parse.ly case also and follow consistent approach across all VIP Integrations.

mehmoodak (changes requested):
Done with first review and left some comments.

vip-integrations.php
alecgeatches: Would this mean dropping the `VIP_BLOCK_DATA_API_VERSION` configuration entirely and only using `VIP_BLOCK_DATA_API_ENABLE` as a on-off toggle for the latest version?

mehmoodak: Yes but that depends whether only supporting the latest version is a valid case for you or not.

vip-integrations.php
alecgeatches: @ingeniumed and I talked a lot about this internally. We'd definitely like to create a flexible system like @mjangda's [abstraction in this comment with `VIP_Integration_Base`](https://github.com/Automattic/vip-go-mu-plugins/pull/4406#issuecomment-1525010763), but we think this will take some time to get right.

I'm personally going to be out of office for two weeks as of tomorrow, and we know that this design step would likely take more than a little architecture and discussion. What we're aiming for right now is a design that (a) doesn't make too many assumptions about how the end architecture will work, and (b) is simple enough that it doesn't add make support complex for whatever architecture we choose.

Right now this is only focused on loading `vip-block-data-api`, but the plan is definitely to generalize when we have the pieces in place. We're specifically keeping things specific here to remove assumptions we'd make about a future system.

mehmoodak (dismissed):
Approving the PR but if it is possible to only support one version then please do so because it will make maintenance easier.

alecgeatches:
@mehmoodak @mjangda Thank you for the comments above! We agree that [something more generalized should be planned out](https://github.com/Automattic/vip-go-mu-plugins/pull/4406#issuecomment-1525010763) for integrations. For now, we wanted to release something small that does the trick (simple configuration to enable our plugin) and give enough time to work out the fine details of a system that will work for all `-ext` supported plugins.

I've made a couple of changes to the initial code:

1. [Per Mehmood's advice](https://github.com/Automattic/vip-go-mu-plugins/pull/4406#discussion_r1179841325), the code now only supports the latest version of the plugin. `VIP_BLOCK_DATA_API_VERSION` for version pinning is no longer supported. This has the benefit that when we transition to a completely new system we don't need to support versioning in a constant, and we can always add version pinning in the future if desired. For this simplified release plan I think that's a good idea.
2. Updated the plugin version to `1.0.0`. After an internal discussion we decided we'll need to do this in the near future anyway, so best to do it now. There's a [second `-ext` PR](https://github.com/Automattic/vip-go-mu-plugins-ext/pull/33) for the dependency update as well.

alecgeatches:
One more note on this PR: We're not looking to merge this until May 25th-ish, when we'll all be available and coordinated with VIP docs and announcements for a release. However, we'd like to get this in a ready state so we can ensure it's ready for deployment in mu-plugins at around the same time. Thank you!

vip-integrations.php
```diff
@@ -0,0 +1,38 @@
...
+}
+
+const VIP_BLOCK_API_LATEST_VERSION = '1.0.0';
+
+class PluginLoadStrategy {
+	const PREFER_MU_PLUGINS_VERSION    = 'PREFER_MU_PLUGINS_VERSION';
+	const PREFER_CUSTOMER_CODE_VERSION = 'PREFER_CUSTOMER_CODE_VERSION';
+}
+
+$plugin_load_strategy = PluginLoadStrategy::PREFER_CUSTOMER_CODE_VERSION;
```
mehmoodak: We aren't changing this variable now .... so do we need any changes in `PluginLoadStrategy`?

alecgeatches: Good question! I wrote a little about this in the issue description:

> The `PluginLoadStrategy` used is largely to ensure that mu-plugins has the ability to decide load order if a plugin is both installed manually and enabled via constant. This is set to `PREFER_CUSTOMER_CODE_VERSION` currently which means if the Block Data API plugin is manually installed, that version will override the constant-enabled version.

Said another way, I wanted to make sure the the design of our plugin would allow it to be loaded before or after client code, depending on how we want VIP integration dependencies to work in the future. This was a design goal in including this plugin, and I wanted to make sure it wasn't set up in a way that made choosing later impossible.

Keeping in this code allowed me to test that this was working properly, and ensure the design worked. You're correct that it's hardcoded right now and doesn't really add any value as-is, but it's useful for testing the loading of the plugin is flexible.

I'm open to removing it, however, if you think it's no longer useful here.

mehmoodak: Removing seems better because it will make lines from 32 to 37 simpler ..... and we can add this preference as a comment on line 27

alecgeatches: Okay! That makes sense to me, and the use of the `VIP_BLOCK_DATA_API_LOADED` constant makes it pretty clear how we're avoiding double-loads. I removed the hardcoded value and all of the surrounding unused logic in https://github.com/Automattic/vip-go-mu-plugins/pull/4406/commits/ced2eade22da937867d756fdaf5ed1a53ce3ac66 / https://github.com/Automattic/vip-go-mu-plugins/pull/4406/commits/9fc3e89a2c1a102a133e501586503ad9175b3926 (comment).

alecgeatches: It's now a ton simpler.

mehmoodak: Great. BTW now I think it is easier to make this generic but completely your call 🙂

alecgeatches: I'd prefer to generalize it when we have a second plugin to put in there. As long as it's flexible enough for generalization, I'd prefer it as-is.

vip-integrations.php
```diff
@@ -0,0 +1,24 @@
...
+	// Do not automatically activate a plugin unless specified.
+	return;
+}
+
+const VIP_BLOCK_API_LATEST_VERSION = '1.0.0';
+
+// Use the plugins_loaded filter to a ensure customer code version of the plugin overrides the mu-plugins version.
+add_action( 'plugins_loaded', function() {
+	if ( ! defined( 'VIP_BLOCK_DATA_API_LOADED' ) ) {
+		require __DIR__ . '/vip-integrations/vip-block-data-api-' . VIP_BLOCK_API_LATEST_VERSION . '/vip-block-data-api.php';
```
rinatkhaziev:
```suggestion
		require_once __DIR__ . '/vip-integrations/vip-block-data-api-' . VIP_BLOCK_API_LATEST_VERSION . '/vip-block-data-api.php';
```

We usually try to `require_once` unless there's a need to include a file multiple times

ingeniumed: Switched it over!

mehmoodak (dismissed):

ingeniumed:
@rinatkhaziev / @rebeccahum - I pushed up [3e9bec0](https://github.com/Automattic/vip-go-mu-plugins/pull/4406/commits/3e9bec0e01a10665ce203a9aa39c020a29c97b39) to ignore the `bootstrap80.php` files, so the mu-plugins sync doesn't fail due to linting.

rinatkhaziev (approved):

PR #4417, opened 2023-05-01
vip_strict_sanitize_username(): return empty string if not string

rebeccahum:
## Description
Uses a similar approach to `sanitize_text_field()` which calls `_sanitize_text_fields()`: https://github.com/WordPress/WordPress/blob/fe8c41f574bf10400af550e59d289dde45a1470a/wp-includes/formatting.php#L5531-L5534

### Function Updated: vip_strict_sanitize_username()

Return empty string if string not passed in

## Checklist

### DIFF
--- a/security.php
+++ b/security.php
@@ -33,6 +33,10 @@ if ( \Automattic\VIP\Security\Private_Sites::has_privacy_restrictions() ) {
  * @return string
  */
 function vip_strict_sanitize_username( $username ) {
+	if ( ! is_string( $username ) ) {
+		return '';
+	}
+
 	if ( is_email( $username ) ) {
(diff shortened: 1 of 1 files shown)


rinatkhaziev:
@rebeccahum Are we sure `is_string` is safe here? E.g. what if my username is `31337` - do we have a guarantee it's passed as a string vs integer?

security.php
```diff
@@ -33,6 +33,10 @@
  * @return string
  */
 function vip_strict_sanitize_username( $username ) {
+	if ( is_object( $username ) || is_array( $username ) ) {
```
sjinks:
```suggestion
	if ( ! is_scalar( $username ) ) {
```

sjinks (approved):

PR #4434, opened 2023-05-03
Search: Default state is no active index versions

rebeccahum:
## Description
Last missing piece of #3963.

Initial state of Enterprise Search is no active versions.

### DIFF
--- a/search/includes/classes/class-versioning.php
+++ b/search/includes/classes/class-versioning.php
@@ -220,10 +220,9 @@ class Versioning {
 	 * Retrieve details about available index versions
 	 *
 	 * @param \ElasticPress\Indexable $indexable The Indexable for which to retrieve index versions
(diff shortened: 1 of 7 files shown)


search/includes/classes/class-versioning.php
```diff
@@ -227,36 +227,22 @@ public function get_versions( Indexable $indexable, bool $provide_default = true
...
-						'number'         => 1,
-						'active'         => true,
-						'created_time'   => null, // We don't know when it was actually created
-						'activated_time' => null,
-					),
-				);
-			} else {
-				return [];
-			}
+		if ( ! is_array( $versions ) || ! isset( $versions[ $indexable->slug ] ) ||
```
rinatkhaziev, line 236: could we just use. the $provide_default code? We can change the default value to false there.

rinatkhaziev (changes requested):
Looks mostly good but I think there's potential to simplifying things a bit, see my inline comment.

PR #4438, opened 2023-05-08
CI: Add link at end of post to tag release for production

rebeccahum:


ci/changelog-summary.php
```diff
@@ -434,6 +434,9 @@ function build_changelog_and_update_prs() {
 
     $title = ucfirst( BRANCH ) . ' release - ' . date( 'Y/m/d' );
     $content = join( "\n<hr />\n", $changelog_entries );
+    if ( BRANCH === 'production' ) {
+        $content .= '<hr /><p>Please see the <a href="https://github.com/Automattic/vip-go-mu-plugins/releases/tag/v' . date( 'Ymd' ) . '.0">full release on GitHub</a>.</p>';
```
rinatkhaziev: hmm so this would be broken for any hotfix releases on the same day. Any ideas how we can address that?

rebeccahum: Maybe we can move the logic to do the actual tagging in the changelog summary generator as well so we have access to that information, as it's currently living in https://github.com/Automattic/vip-go-mu-plugins/blob/develop/.github/workflows/tag-release.yml.

Edit: On further research, we can use GitHub artifacts

sjinks (dismissed):

rinatkhaziev (approved):

PR #4447, opened 2023-05-09
Add support for SDS that sits in the customer namespace

abdullah-kasim:
## Description

We're doing some service migration currently. The `site-details` endpoint is being transferred to the `SDS_API_URL` endpoint. This PR helps get things ready for it - and all that'll be needed is to update the sites to have the `SDS_AUTH_TOKEN` and `SDS_API_URL`.

It's hard to add automated tests for this as we don't have [something like this](https://patchwork2.org/) in our repo except for Jetpack. I tried and gave up.

### Add site-details endpoint.

We added another constant for switching over to the new `site-details` service when it's ready.

## Checklist

- [x] This change has relevant unit tests (if applicable). **Not applicable because we can't mock global constants**
- [x] This change uses a rollout method to ease with deployment (if applicable - especially for large scale actions that require writes). **I guess it technically rolls out to one or two test sites**
- [x] This change has relevant documentation additions / updates (if applicable). **Not Applicable**

## Steps to Test

1. Check out PR.
2. Put some sort of `var_dump` for $url and $args at this [line](https://abdullah.github.com/Automattic/vip-go-mu-plugins/blob/d12a5e8eaf218afb74ed0147f7eda00aa7a02e25/config/class-site-details-index.php#L380-L380)
3. Define `SDS_AUTH_TOKEN` and `SDS_API_URL` to any url and token in `wp-config.php` - this actual service doesn't exist yet :)
4. Call `Site_Details_Index::put_site_details()` through `wp shell`
5. We should see `$args` and `$url` get populated with the payload and the url respectively.
6. Do the same thing from step 3 onwards again, but populate `SERVICES_AUTH_TOKEN` and `SERVICES_API_URL` instead to test backwards compatibility.


config/class-site-details-index.php
```diff
@@ -349,8 +349,21 @@ public function get_site_details() {
...
+			$url = rtrim( SDS_API_URL, '/' ) . '/sites';
+
+			$args = array(
+				'method'  => 'PUT',
+				'body'    => wp_json_encode( $site_details ),
+				'headers' => array(
+					'Authorization' => 'Bearer ' . SDS_AUTH_TOKEN,
+					'Content-Type'  => 'application/json',
+				),
+			);
```
pschoffer: The whole args array seems to be identical to the else branch except the token. What do you think of setting something like:
```
$auth_token = SDS_AUTH_TOKEN
```

Then constructing the args array only before `vip_safe_wp_remote_request`, so in shared code for both conditions.

pschoffer (changes requested):
Looks good, I left one nitpick :)

abdullah-kasim:
Cool, so the last test failed due to flakiness.

pschoffer (approved):
Very cool thanks!

pschoffer:
> Cool, so the last test failed due to flakiness.

Yeah, lets merge it I think. Feel free to merge.

abdullah-kasim:
Need to check tests. I saw it passing a while ago and it's somehow failing 3 times now.

abdullah-kasim:
Oh okay, it was codecov. Merging.

PR #4458, opened 2023-05-11
Add VIP Integrations framework

ingeniumed:
## Description

The previous integration was removed in [this PR](https://github.com/Automattic/vip-go-mu-plugins/pull/4451) to re-think the approach used to enable plugins. It's now up again with the logic being:

- Call the higher level method `load_integration` and pass in the correct slug along with any config options that are supported.

This makes it extendable in the future for other integrations, as well as provides the ability to change the way integrations are enabled as their implementation is hidden away.

In addition to that, the vendor folder for the block-data-api is now ignored by the linter. Without this, the linter would fail. Previously, the `dockerignore` file was updated with files to ignore. This reverses that, and use the lint ignore file instead.

There is a corresponding PR to reverse that dockerignore fix in the mu-plugins-ext: https://github.com/Automattic/vip-go-mu-plugins-ext/pull/35

This PR and the above PR should be merged at the same time to avoid build failures.

### Added high level method to enable VIP Integration Plugins

Added support to enable VIP Integration Plugins, starting with the [VIP Block Data API plugin](https://github.com/Automattic/vip-block-data-api/) via the `load_integration` method.

## Checklist

- [x] (?) This change has relevant unit tests (if applicable).
- [x] N/A - This change uses a rollout method to ease with deployment.
- [x] External documentation in progress - This change has relevant documentation additions / updates (if applicable).

## Steps to Test

Ensure you have [this PR's change](https://github.com/Automattic/vip-go-mu-plugins-ext/pull/35) in your vip-go-mu-plugins-ext.

1. Wait until a `vip-go-mu-plugins` deployment is made, which should pull [the `vip-integrations` folder from `-ext`](https://github.com/Automattic/vip-go-mu-plugins-ext/tree/trunk/vip-integrations) into [vip-go-mu-plugins-built](https://github.com/Automattic/vip-go-mu-plugins-built).
2. Clone [vip-go-mu-plugins-built](https://github.com/Automattic/vip-go-mu-plugins-built) locally set `vip dev-env`'s environment to use it, e.g.

    ```bash
    vip dev-env update --mu-plugins=~/projects/mu-plugins-built
    ```

3. Verify the block data API plugin is disabled by default. Visit the REST endpoint for an existing post ID:

    ```bash
    GET /wp-json/vip-block-data-api/v1/posts/<post_id>/blocks

    # e.g. GET http://my-site.vipdev.lndo.site/wp-json/vip-block-data-api/v1/posts/1/blocks
    ```

    Ensure you receive a `404`response with status code `rest_no_route`.

4. In `client-mu-plugins/plugin-loader.php`, add:

    ```php
    \Automattic\VIP\Integrations\activate( 'block-data-api' );
    ```

6. Visit the same URL as step 3, and ensure the block data API is now available:

    ```
    GET http://my-site.vipdev.lndo.site/wp-json/vip-block-data-api/v1/posts/1/blocks

    {"blocks":[{"name":"core/paragraph","attributes":{"content":"Welcome to WordPress. This is your first post. Edit or delete it, then start writing!","dropCap":false}}]}
    ```

---

### Unit tests

Run:

```
./bin/test.sh --filter VIP_Integrations_Test
```


mjangda:
Just pulling this out as the usage example for customers:

```
use function Automattic\VIP\Integrations\load_integration;
use const Automattic\VIP\Integrations\BLOCK_DATA_API_SLUG;
load_integration( BLOCK_DATA_API_SLUG );
```

In the future, this can be expanded to support other integrations and provide config options:

```
load_integration( Automattic\VIP\Integrations\PARSELY, array(
    'version' => '3.6',
    'meta' => array(
        'api_key' => 'example.com',
    ),
) );
```

This would go in `client-mu-plugins`.

vip-integrations.php
```diff
@@ -0,0 +1,38 @@
...
+const BLOCK_DATA_API_SLUG = 'block-data-api';
+
+/**
+ * Load the integration based on the slug provided.
+ *
+ * @param string $slug The slug of the integration to load.
+ * @param array $config The configuration for the integration, if supported.
+ */
+function load_integration( $slug, $config = array() ) {
+	if ( BLOCK_DATA_API_SLUG !== $slug ) {
```
mjangda: Let's a add a comment nothing that is temporary.

vip-integrations.php
```diff
@@ -0,0 +1,38 @@
...
+ * Plugin Name: VIP Integrations
+ * Description: Plugin loading for VIP plugin integrations.
+ * Author: Automattic
+ * License: GPL version 2 or later - http://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+ */
+
+namespace Automattic\VIP\Integrations;
+
+// New integrations names go here
+const BLOCK_DATA_API_SLUG = 'block-data-api';
```
mjangda: We can't use an enum here (beucase it's PHP 8+) but maybe a stub class would make sense?

```
class Integration {
    const BLOCK_DATA_API = 'block-data-api';
    // in the future, we can add wp-parsely and others here.
}
```

The usage becomes a little cleaner:

```
load_integration( Integration:BLOCK_DATA_API );
```

mjangda (dismissed):

vip-integrations.php
ingeniumed: Added it in

vip-integrations.php
ingeniumed: Good point, I turned it into an abstract class and used that instead.

ingeniumed:
Noting down the updated usage in `client-mu-plugins`:

```
use function Automattic\VIP\Integrations\load_integration;
use Automattic\VIP\Integrations\Integration;
load_integration( Integration::BLOCK_DATA_API );
```

vip-integrations.php
```diff
@@ -0,0 +1,44 @@
...
+	if ( Integration::BLOCK_DATA_API !== $slug ) {
+		return;
+	}
+
+	// Use the plugins_loaded filter to ensure customer code version of the plugin overrides the mu-plugins version.
+	add_action( 'plugins_loaded', function() {
+		// New integration requires go here
+
+		// Block Data API defines this when it is loaded, so it's a guard against loading twice.
+		if ( ! defined( 'VIP_BLOCK_DATA_API_LOADED' ) ) {
```
mehmoodak: Instead of defining if block above can we improve this condition as follow?

```php
if ( ! defined( 'VIP_BLOCK_DATA_API_LOADED' ) && Integration::BLOCK_DATA_API === $slug ) {
```

vip-integrations.php
```diff
@@ -0,0 +1,44 @@
...
+	// In the future, we can add wp-parsely and other integrations here.
+}
+
+/**
+ * Load the integration based on the name provided.
+ *
+ * @param string $slug The name of the integration to load.
+ * @param array $config The configuration for the integration, if supported.
+ */
+function load_integration( $slug, $config = array() ) {
```
mehmoodak: `$integration_name` seems like a better name than `$slug`

mehmoodak (dismissed):
Left some comments but approving the PR.

phpcs.xml.dist
```diff
@@ -41,6 +41,7 @@
 	<exclude-pattern>/wordpress-importer/*</exclude-pattern>
 	<exclude-pattern>/wp-parsely/*</exclude-pattern>
 	<exclude-pattern>*/build/*\.asset\.php</exclude-pattern>
+	<exclude-pattern>/vip-integrations/*</exclude-pattern>
```
GaryJones: Why is this being excluded?

GaryJones:
I'm not entirely sure of the purpose of this refactor, but I'd strongly encourage a look at https://github.com/Parsely/wp-parsely/tree/70a54f896fbae3e248e00c23809a77faa35a092e/src/Integrations for a more objection-orientated approach:

 - [Integrations (plural) class](https://github.com/Parsely/wp-parsely/blob/develop/src/Integrations/class-integrations.php): ~container~ collection which holds registered integrations in a private array property - takes class or object instances and a key. This can be done ([example](https://github.com/Parsely/wp-parsely/blob/70a54f896fbae3e248e00c23809a77faa35a092e/wp-parsely.php#L241-L266)) in a _composable root_ . The Integrations class itself knows nothing about any specific integration to avoid unnecessary coupling (better for testing). Can then trigger the [loading/integration process of all integrations](https://github.com/Parsely/wp-parsely/blob/70a54f896fbae3e248e00c23809a77faa35a092e/src/Integrations/class-integrations.php#L75-L80) by a [single method call](https://github.com/Parsely/wp-parsely/blob/70a54f896fbae3e248e00c23809a77faa35a092e/wp-parsely.php#L262-L263) (after optionally allowing the registered integrations to be filtered).
 - [Integration (singular) class](https://github.com/Parsely/wp-parsely/blob/develop/src/Integrations/class-integration.php): abstract class which provides the contract that all concrete integrations should adhere to - it knows nothing about any specific integration. This could be an interface (preferred), but for the wp-parsely situation, we make a `Parsely` instance available as it's typically needed for those integrations, hence it's an abstract class.
 - Concrete integration classes ([AMP example](https://github.com/Parsely/wp-parsely/blob/develop/src/Integrations/class-amp.php)): extends the Integration class, and provides the required abstract `integrate()` method from it to (optionally conditionally) hook the other public methods into hooks where needed. There can be as many of these concrete classes as necessary, and they cohesively isolate the domain-specific knowledge for that integration away from the mechanics of actually setting up an integration.

vip-integrations.php
ingeniumed: I had thought of that, but an advantage I saw with this approach was to exit out early. We don't need to even go into adding an action if the plugin names don't match.

Ex: Let's say its called twice, once for block data and the other for governance. I know the priority would take care of it, but why even let that hook get registered if we can exit out early from that function.

vip-integrations.php
ingeniumed: Good point, I'll tweak that.

phpcs.xml.dist
ingeniumed: I had initially had this targeted at just the specific vendor folder, but changed it to the whole folder. I'll go back to just the vendor folder instead.

ingeniumed: Pushed it up!

vip-integrations.php
ingeniumed: I went ahead and changed the config parameter as well.

alecgeatches:
@GaryJones @mehmoodak @mjangda @Automattic/vip-platform-cantina This is ready for review again! We've updated how integrations work based [on Gary's comments above](https://github.com/Automattic/vip-go-mu-plugins/pull/4458#issuecomment-1545667045). To activate an integration, run this code in `client-mu-plugins.php`:

```php
\Automattic\VIP\Integrations\activate_integration( 'block-data-api' );
```

Here were a few design decisions in the remake:

1. Integration registration, activation, and `integration()` calls [use the `muplugins_loaded` filter](https://github.com/Automattic/vip-go-mu-plugins/blob/9e25ebcda2d7c8766e4a45f46505d13c1a5a738b/vip-integrations.php#L17). This fires before `init` or `plugins_loaded`. This is useful because it allows concrete `Integration` classes to add hooks to any other action in their implementation. For the Block Data API we want to run our [plugin on `plugins_loaded`](https://github.com/Automattic/vip-go-mu-plugins/blob/9e25ebcda2d7c8766e4a45f46505d13c1a5a738b/vip-integration-helper/integrations/block-data-api.php#L15), so this gives us flexibility to determine how to load an integration on a per-integration basis. Using another filter like `init` for calling integration code would limit the filters still left to fire.

2. Added an activation step. The Parsely examples provided by @garyjones are all registered and activated automatically, but for mu-plugins this was changed to provide an explicit activation step for each feature. Integrations can be registered in `vip-integrations.php`, but will not be activated unless `activate_integration( $slug )` is called.
3. Added a `$config` variable provided to activated plugins. Configuration can be optionally provided in `activate_integration()`:

    ```php
    \Automattic\VIP\Integrations\activate_integration( 'block-data-api', ['config_key' => 'value' ] );
    ```

    This `$config` variable is passed to the `integrate()` function for each concrete `Integration` implementer. This isn't strictly necessary for our changes, but liked this part of the earlier design.
4. Using slugs (`'block-data-api'`) instead of constants. I sort of like we only need to set the slug during registration and don't need a separate structure to update with plugin names. However, I'm totally open to wrapping that part in a enum-like class.

Let us know what you think! Thank you.

vip-integration-helper/integrations/block-data-api.php
```diff
@@ -0,0 +1,27 @@
...
+		add_action( 'plugins_loaded', function() {
+			// Guard against loading the plugin twice
+			if ( defined( 'VIP_BLOCK_DATA_API_LOADED' ) ) {
+				return;
+			}
+
+			$load_path = __DIR__ . '/../../vip-integrations/vip-block-data-api-1.0.0/vip-block-data-api.php';
+			if ( file_exists( $load_path ) ) {
+				require_once $load_path;
+			}
```
GaryJones: Under what conditions might the plugin load twice?

Doesn't the use of `require_once` (rather than `require`) handle this case anyway?

vip-integration-helper/integrations/block-data-api.php
```diff
@@ -0,0 +1,27 @@
...
+		}, /* priority */ 1 );
```
GaryJones: Can you add more of a comment (in the code) as to why this should be loaded at an earlier priority than the default?

vip-integration-helper/integrations/integration.php
```diff
@@ -0,0 +1,47 @@
+<?php
+
+namespace Automattic\VIP\Integrations;
+
+defined( 'ABSPATH' ) || die();
```
GaryJones: Checks like this shouldn't be needed for class files. If an HTTP request is made directly to this file (which is what the `ABSPATH` check is looking to protect against), then all that will happen is the class (or interface, etc.) will be defined, and that's it - since there is (correctly) no side-effect code being called in this file, then there's nothing to actually protect.

vip-integration-helper/integrations/integration.php
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Abstract base class for all integration implementations.
+ */
+abstract class Integration {
+	/**
+	 * An optional configuration array for this integration, added during activation.
+	 *
+	 * @var array
+	 */
+	public $config = [];
```
GaryJones: I'd recommend making this `protected` and providing a method to retrieve it.

vip-integration-helper/integrations/integration.php
```diff
@@ -0,0 +1,47 @@
...
+	 * @var bool
+	 */
+	protected $active = false;
+
+	/**
+	 * Activates this integration with an optional configuration value.
+	 *
+	 * @param array  $config An associative array of configuration values for the integration.
+	 */
+	public function activate( $config = [] ): void {
```
GaryJones: If this is documented to be an array, then where's the type validation?

```suggestion
	public function activate( array $config = [] ): void {
```

Of course, it limits being able to use ArrayObject, or a custom configuration class per each integration, but the rest of the logic/docs makes assumptions that `$config` is expected to be an array anyway.

GaryJones: Actually, why is the configuration being set at activation time? Is it likely to change from the registration (`integrate`) time? Is this to allow customer-overrides of the config?

vip-integration-helper/integrations/integrations.php
```diff
@@ -0,0 +1,64 @@
+<?php
+
+namespace Automattic\VIP\Integrations;
+
+defined( 'ABSPATH' ) || die();
```
GaryJones: See a previous note.

vip-integration-helper/integrations/integrations.php
```diff
@@ -0,0 +1,64 @@
...
+
+require_once __DIR__ . '/integration.php';
```
GaryJones: This is side effect code in a class file (not good) - it should be moved out and into a place where other file-loading is happening.

vip-integration-helper/integrations/integration.php
GaryJones: This line uses a short array syntax, but other instances are using long array syntax - it should be consistent.

vip-integration-helper/integrations/integrations.php
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	/**
+	 * Registers an integration.
+	 *
+	 * @param string       $slug        A unique identifier for the integration.
+	 * @param class-string $class_name Fully-qualified class name that will be instantiated.
+	 */
+	public function register( string $slug, string $class_name ): void {
+		$integration = new $class_name();
```
GaryJones: By making the integration only be instantiated here, it restricts others from being able to instantiate elsewhere while passing it it's own parameters. As per wp-parsely, I'd recommend that `$class_name` accept a string or an object - if it's a string, instantiate it like you've done, but if it's an object, accept as is.

e.g. both should be acceptable:

```
$integrations->register( \My\Integration::class );

$foo = new Foo( new MyInjectedDependency() );
$integrations->register( $foo );
```

vip-integration-helper/integrations/integrations.php
```diff
@@ -0,0 +1,64 @@
...
+
+		$this->integrations[ $slug ] = $integration;
```
GaryJones: Any validation here needed to check if the slug already exists? With no validation, we'll be allowing overwrites to happen - which can be a good thing, but should be a known and documented feature.

vip-integration-helper/integrations/integrations.php
```diff
@@ -0,0 +1,64 @@
...
+		}
+
+		return $instance;
+	}
+
+	/**
+	 * Registers an integration.
+	 *
+	 * @param string       $slug        A unique identifier for the integration.
+	 * @param class-string $class_name Fully-qualified class name that will be instantiated.
```
GaryJones: `class-string` isn't a known PHP type.

vip-integrations.php
```diff
@@ -0,0 +1,48 @@
...
+ */
+function activate_integration( $integration_slug, $config = [] ) {
+	add_action( 'muplugins_loaded', function() use ( $integration_slug, $config ) {
+		$integration = Integrations::instance()->get_registered( $integration_slug );
+
+		if ( null !== $integration ) {
+			$integration->activate( $config );
+		}
+	}, /* priority */ 6 );
+}
```
GaryJones: So it's clear in my mind...

Currently:
 - `muplugins_loaded:4` - built-in integrations are registered.
 - `muplugins_loaded:6` - registered integrations activated via a call in `vip-config.php` happen at this point.
 - `muplugins_loaded:8` - active integrations have their `integrate()` method called, which, at least for block-data-api allows it to hook into `plugins_loaded` and do its thing.
 
That only gives someone a gap of `muplugins_loaded:<6` to register or unregister their Integration (any later and activate_integrations will be hooking into muplugins_loaded:6 when the new Integration hasn't been registered yet).

vip-integrations.php
```diff
@@ -0,0 +1,48 @@
...
+	Integrations::instance()->register( 'block-data-api', BlockDataApi::class );
+}
+
+/**
+ * Activates an integration with an optional configuration value.
+ *
+ * @param string $slug   A unique identifier for the integration.
+ * @param array  $config An associative array of configuration values for the integration.
+ */
+function activate_integration( $integration_slug, $config = [] ) {
```
GaryJones: This is already in the Integrations namespace, and it'll be cleaner to have this as `activate( 'block-data-api' )` etc.

```suggestion
function activate( $integration_slug, $config = [] ) {
```

vip-integrations.php
```diff
@@ -0,0 +1,48 @@
...
+
+require_once __DIR__ . '/vip-integration-helper/integrations/integrations.php';
+require_once __DIR__ . '/vip-integration-helper/integrations/block-data-api.php';
+
+add_action( 'muplugins_loaded', __NAMESPACE__ . '\\register_integrations', /* priority */ 4 );
+
+/**
+ * Register valid integrations.
+ */
+function register_integrations(): void {
```
GaryJones:
```suggestion
function register_vip_integrations(): void {
```

Just helps to differentiate it a little from `Integrations::register()`.

GaryJones:
I'd love to see some tests added for this as well. Off the top of my head...

**Registration**

- registering a FakeIntegration as a classname string.
- registering a FakeIntegration as an object.
- registering a FakeIntegration and passing in a config array.
- trying to register something that isn't a string or an object, but includes null (as a way someone might try to unregister an Integration) (error handling).
- demonstrate overriding an existing registered integration (slug) with a new Integration (i.e. register FakeIntegration with `block-data-api` slug) - maybe this behaviour should be blocked.
- demonstrate overriding an existing registered integration with the same Integration but a different config (since config can be overridden at activation time, perhaps this should be blocked).
- trying to pass in a non-array config parameter.

**Activation**
- activating a FakeIntegration.
- activating a FakeIntegration and passing an array config.
- trying to activate a FakeIntegration and passing a non-config array (error handling).
- trying to activate a non-registered integration (error handling).
- trying to activate a FakeIntegration that was registered after `muplugins_loaded` priority 6.
- having a FakeIntegrationWithValidation Integration demonstrate config validation (i.e. not accepting invalid keys or values) and then pass in a config array with invalid keys/values on activation.

vip-integration-helper/integrations/block-data-api.php
alecgeatches: This plugin can be loaded twice if the customer has a copy of the plugin installed in the `plugins/` folder and activated manually. In that case, `require_once` wouldn't work since there are two plugin paths. We're using `plugins_loaded` here specifically to allow customer code to override the mu-plugins version if desired. I'll update the comment here to be more specific!

vip-integration-helper/integrations/block-data-api.php
alecgeatches: Good catch, I don't think this is necessary. I'll remove the `1` priority.

vip-integration-helper/integrations/integration.php
alecgeatches: > Actually, why is the configuration being set at activation time?
> ...
> Is this to allow customer-overrides of the config?

Yep, that's the idea. If we'd like to provide user-facing integration config, we can provide options like:

```php
\Automattic\VIP\Integrations\activate( 'some-integration', [
    'API_KEY' => '974f2acb...',
] );
```

These provide a way to connect configuration in the same place as activation, rather than using constants or something in addition.

vip-integration-helper/integrations/integrations.php
alecgeatches: I stole this from [wp-parsely's `Integrations` implementation](https://github.com/Parsely/wp-parsely/blob/2f5c3428c7c11c6177c5622a2477022072e5c2b3/src/Integrations/class-integrations.php#L53):

```php
/**
 * ...
 * @param class-string|Integrations $class_or_object Fully-qualified class name, or an instantiated object.
 *                                                   If a class name is passed, it will be instantiated.
```

but admittedly it's the first time I've seen it used. It appears to be [a PHPStan/PHPDoc type](https://phpstan.org/writing-php-code/phpdoc-types#class-string) and we also use it in a few places in mu-plugins code as well as wp-parsely code. Should I still replace this with `string`?

vip-integrations.php
alecgeatches: That's correct - there's a bit of space (`4`, `6`, `8`) to allow injecting between the setup priorities, but it's not designed for that explicitly. Each VIP `Integration` should be registered by us via `register_vip_integrations()` which will ensure activation occurs afterward. As long as our VIP integrations are registered there, the priorities can be ignored by integration implementations.

The priorities are using sub-`10` values so that integrations can register with the default `muplugins_loaded` (`10`) if desired.

vip-integration-helper/integrations/integration.php
alecgeatches: Good point! I've removed those checks everywhere but `vip-integrations.php` where our side-effects live.

vip-integration-helper/integrations/integration.php
alecgeatches: Changed `$config` to `protected` with a `get_config()` method instead. I've also updated all array literals to the short-array syntax.

vip-integration-helper/integrations/integrations.php
alecgeatches: Updated `register()` to take an instantiated object or class name, like `wp-parsely`'s integrations.

vip-integration-helper/integrations/integrations.php
GaryJones: If PHPStan isn't being used against this code, then yes. Otherwise leave it as is (I'm curious how IDEs interpret it).

vip-integration-helper/integrations/integrations.php
alecgeatches: Good idea. I added a `throw` since I don't think overridding integrations should be expected:

https://github.com/Automattic/vip-go-mu-plugins/blob/5819a1867a7d42988b2dcf85923f7f39b8155f9b/vip-integration-helper/integrations/integrations.php#L38-L40

We can change this to be allowable in the future if we need it for some reason.

alecgeatches:
@GaryJones Thank you for all of the feedback! I've provided changes for your current code comments, and I'll have a go at adding some tests later. Thank you!

vip-integration-helper/integrations/integrations.php
alecgeatches: Changed this to `string` since we're not using PHPStan directly in mu-plugins.

alecgeatches:
This is ready for one more review! @GaryJones - in the interest of getting this released sooner, we'd like to cover extensive testing in a separate PR. We've basic tests in [test-vip-integrations.php](https://github.com/Automattic/vip-go-mu-plugins/blob/9427259a0e132c21b8190ec5dcc88be9f29d288b/tests/vip-integrations/test-vip-integrations.php). Building out these tests required us to re-wire how we integrate with `muplugins_loaded` so that we could more easily test these functions outside the normal WordPress loop:

https://github.com/Automattic/vip-go-mu-plugins/blob/9427259a0e132c21b8190ec5dcc88be9f29d288b/vip-integrations.php#L21-L25

...

https://github.com/Automattic/vip-go-mu-plugins/blob/9427259a0e132c21b8190ec5dcc88be9f29d288b/vip-integrations.php#L29-L36

Rather than having our integrations/activations directly tied to `muplugins_loaded:<some-priority>`, they're tied to `<action>_vip_integrations` hooks instead. I think this makes the timeline of loading and hooking into integrations much clearer.

Please let me know what you think, and if you're okay with us extending testing in a later PR. Thank you!

GaryJones:
I think there are further opportunities to improve the code here, such as:

- Making the collection class not a Singelton pattern (though it currently doesn't have a private `__construct()`, so it's not actually enforcing the singleton-ness anyway).
- Make the collection class implement `\Countable`, `\Iterator`, and `\ArrayAccess`, or just extend `\ArrayObject` and override methods as needed.
- Have the tests be more Unit test in nature i.e. testing the registering of an integration doesn't need to be wrapped in an `add_action()` callback. It should know nothing about WordPress, since it's just a collection class (pure PHP object).

... but these don't need to be blockers for now.

alecgeatches:
@GaryJones Thank you for the suggestions and guidance on the PR! Because we've hit a point where improvements on the code are no longer blocking, we're happy with getting the block data API in now, and improving code as we expand integrations.

@Automattic/vip-platform-cantina @mehmoodak @mjangda is anyone ready to give approval for this merge? Thank you!

alecgeatches:
Note WP Core tests are failing, but [this appears to be true across all branches](https://github.com/Automattic/vip-go-mu-plugins/actions/workflows/core-tests.yml) and is not specific to our code.

vip-integration-helper/integrations/integration.php
```diff
@@ -0,0 +1,52 @@
...
+	 * @var array
+	 */
+	protected array $config = [];
+
+	/**
+	 * A boolean indicating if this integration should be loaded. Defaults to false.
+	 *
+	 * @var bool
+	 */
+	protected bool $active = false;
```
mjangda: Minor: I find that booleans that start with is/has/etc make it more obvious it's a boolean (i.e. `is_active`)?

vip-integration-helper/integrations/integrations.php
```diff
@@ -0,0 +1,78 @@
+<?php
```
mjangda: Should we skip the `vip-` prefix and `-helper` suffix? Seems redundant.

I would just call the folder `integrations`?

mjangda:
Reviewing and testing this today.

vip-integration-helper/integrations/integrations.php
mjangda: Aside: should add `@private` to the docblocks for the various classes to provide a signal to customers not to use these directly?

Alternatively, could just add a disclaimer stating that all access is `private` unless specified.

vip-integration-helper/integrations/integrations.php
```diff
@@ -0,0 +1,78 @@
...
+	 * @param string|Integration $class_or_object Fully-qualified class or instantiated Integration object.
+	 */
+	public function register( string $slug, $class_or_object ): void {
+		if ( isset( $this->integrations[ $slug ] ) ) {
+			throw new InvalidArgumentException( sprintf( 'Integration with slug "%s" is already registered.', $slug ) );
+		}
+
+		if ( ! is_object( $class_or_object ) ) {
+			$class_or_object = new $class_or_object();
+		}
```
mjangda: Minor (not blocker): should we validate that the class_exists and also that it's of the right type (inherits from the base integration)?

vip-integration-helper/integrations/integrations.php
```diff
@@ -0,0 +1,78 @@
...
+
+		$this->integrations[ $slug ] = $class_or_object;
+	}
+	/**
+	 * Returns a registered integration for a key, or null if not found.
+	 *
+	 * @param string $slug A unique identifier for the integration.
+	 */
+	public function get_registered( string $slug ): ?Integration {
```
mjangda: Minor: we could just call this `get`?

vip-integration-helper/integrations/integrations.php
```diff
@@ -0,0 +1,78 @@
...
+	 * Call integrate() for each registered and activated integration.
+	 */
+	public function integrate(): void {
+		foreach ( $this->integrations as $slug => $integration ) {
+			$is_active = $integration->is_active() ? 'yes' : 'no';
+
+			if ( $integration->is_active() ) {
+				$integration->integrate( $integration->get_config() );
+
+				unset( $this->integrations[ $slug ] );
```
mjangda: Why do we unset here?

vip-integration-helper/integrations/integrations.php
```diff
@@ -0,0 +1,78 @@
...
+	public function get_registered( string $slug ): ?Integration {
+		return $this->integrations[ $slug ] ?? null;
+	}
+
+	/**
+	 * Call integrate() for each registered and activated integration.
+	 */
+	public function integrate(): void {
+		foreach ( $this->integrations as $slug => $integration ) {
+			$is_active = $integration->is_active() ? 'yes' : 'no';
```
mjangda: Unused?

vip-integrations.php
```diff
@@ -0,0 +1,70 @@
...
+		} );
+
+		do_action( 'register_vip_integrations' );
+	}
+
+	public function activate(): void {
+		do_action( 'activate_vip_integrations' );
+	}
+
+	public function integrate(): void {
```
mjangda: Should we call this `load`? Not totally obvious was `integrate` is supposed to mean.

vip-integrations.php
```diff
@@ -0,0 +1,70 @@
...
+
+	// Actions
+
+	public function register(): void {
+		add_action( 'register_vip_integrations', function() {
+			// Register VIP integrations here
+			Integrations::instance()->register( 'block-data-api', BlockDataApi::class );
+		} );
+
+		do_action( 'register_vip_integrations' );
```
mjangda: I think the naming needs to be reversed:

`register_vip_integrations` => `vip_integrations_register`
`activate_vip_integrations` => `vip_integrations_activate`

vip-integration-helper/integrations/block-data-api.php
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Integrates VIP Block Data REST API.
+ */
+class BlockDataApi extends Integration {
+	/**
+	 * Applies hooks to integrate Block Data API plugin.
+	 */
+	public function integrate( array $config ): void {
+		add_action( 'plugins_loaded', function() {
```
mjangda: Minor: What's the reason we need to wait until `plugins_loaded`?

vip-integrations.php
```diff
@@ -0,0 +1,70 @@
...
+		add_action( 'muplugins_loaded', [ $this, 'integrate' ], /* priority */ 8 );
+	}
+
+	// Actions
+
+	public function register(): void {
+		add_action( 'register_vip_integrations', function() {
+			// Register VIP integrations here
+			Integrations::instance()->register( 'block-data-api', BlockDataApi::class );
+		} );
```
mjangda:
```suggestion
		// Register VIP integrations here
		Integrations::instance()->register( 'block-data-api', BlockDataApi::class );
```

The `add_action` seems unnecessary here?

vip-integrations.php
```diff
@@ -0,0 +1,70 @@
...
+
+require_once __DIR__ . '/vip-integration-helper/integrations/integration.php';
+require_once __DIR__ . '/vip-integration-helper/integrations/integrations.php';
+require_once __DIR__ . '/vip-integration-helper/integrations/block-data-api.php';
+
+class VipIntegrations {
+	public function setup() {
+		add_action( 'muplugins_loaded', [ $this, 'register' ], /* priority */ 4 );
+		add_action( 'muplugins_loaded', [ $this, 'activate' ], /* priority */ 6 );
+		add_action( 'muplugins_loaded', [ $this, 'integrate' ], /* priority */ 8 );
```
mjangda: What's the reason for staggering these across different priorities? If there's no good reason, I would just call one method and trigger the custom hooks there:

```
add_action( 'muplugins_loaded', [ $this, 'run' ] );

public function run() {
    do_action( 'vip_integrations_register' );
    do_action( 'vip_integrations_activate' );
...
}
```

vip-integrations.php
```diff
@@ -0,0 +1,70 @@
...
+ */
+function activate( string $integration_slug, array $config = [] ) {
+	add_action( 'activate_vip_integrations', function() use ( $integration_slug, $config ) {
+		$integration = Integrations::instance()->get_registered( $integration_slug );
+
+		if ( null === $integration ) {
+			throw new InvalidArgumentException( sprintf( 'VIP Integration with slug "%s" is not a registered integration.', $integration_slug ) );
+		} else {
+			$integration->activate( $config );
+		}
```
mjangda: Can we simplify this by just doing something like `Integrations::instance()->activate( $slug );` (this logic would move into that method)

vip-integration-helper/integrations/integrations.php
```diff
@@ -0,0 +1,78 @@
...
+			if ( $integration->is_active() ) {
+				$integration->integrate( $integration->get_config() );
+
+				unset( $this->integrations[ $slug ] );
+			}
+		}
+	}
+
+	/**
+	 * Remove all registered integrations
```
mjangda: Let's add a comment noting that this is just for tests.

vip-integrations.php
```diff
@@ -0,0 +1,70 @@
...
+
+use InvalidArgumentException;
+
+defined( 'ABSPATH' ) || die();
+
+require_once __DIR__ . '/vip-integration-helper/integrations/integration.php';
+require_once __DIR__ . '/vip-integration-helper/integrations/integrations.php';
+require_once __DIR__ . '/vip-integration-helper/integrations/block-data-api.php';
+
+class VipIntegrations {
```
mjangda: What's the benefit is having this class separate from the `Integrations` class? Feels like they could be combined?

vip-integrations.php
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	public function activate(): void {
+		do_action( 'activate_vip_integrations' );
+	}
+
+	public function integrate(): void {
+		add_action( 'integrate_vip_integrations', function() {
+			Integrations::instance()->integrate();
+		} );
```
mjangda:
```suggestion
		Integrations::instance()->integrate();
```

Similar to `register`.

vip-integrations.php
GaryJones: The `Integrations` class is just a collection - it should know nothing about WordPress.

vip-integration-helper/integrations/integration.php
alecgeatches: Agreed! Changed this to `$is_active`.

vip-integration-helper/integrations/integrations.php
alecgeatches: Done on both of these - moved integrations code to just `integrations/`, `@private`ed all of the internal integration classes.

vip-integration-helper/integrations/integrations.php
alecgeatches: Good idea! I added exceptions for these and tests.

vip-integration-helper/integrations/integrations.php
alecgeatches: This was extraneous and has been removed.

vip-integration-helper/integrations/integrations.php
alecgeatches: Good catch! Removed.

vip-integrations.php
alecgeatches: Great idea, changed all the `integrate` verbiage into `load`. Much clearer.

vip-integrations.php
alecgeatches: You're right! These have been switched to be hierarchical.

vip-integration-helper/integrations/block-data-api.php
alecgeatches: If the plugin is installed and activated manually in customer code, and also activated via integrations, we want the customer code version to win. `plugins_loaded` is so we can check for a regular plugin load before integration.

vip-integrations.php
alecgeatches: I made some changes to how we handle integrations by removing a lot of these extraneous filters. I'll leave a comment below with more information.

vip-integrations.php
alecgeatches: Good idea! I moved `activate()` into the `Integrations` class. There is still a wrapper `activate()` method in `vip-integrations.php` that's used to call into the main instance's `activate()`, but the logic now lives in the class.

vip-integration-helper/integrations/integrations.php
alecgeatches: By unsingletoning `Integrations` this method was no longer necessary.

vip-integrations.php
alecgeatches: ^ Following some of @GaryJones previous advice, we did some more work to make most integrations code dumber and unaware of WordPress. This meant we were able to [remove slow test isolation](https://github.com/Automattic/vip-go-mu-plugins/pull/4458/commits/78ddb24506af230f1f1673fa9a7df3a975a21869#diff-64b428d68bfd6773d14920ee388f49ef7f4807037d8fb196be4ab8e1930d9961L9-L12) and generally made tests a lot cleaner.

alecgeatches:
@mjangda @GaryJones Pushed some changes to address Mo's feedback, and ultimately implemented some of [Gary's design suggestions](https://github.com/Automattic/vip-go-mu-plugins/pull/4458#issuecomment-1561143586) like:

1. The main `Integrations` collection is no longer a singleton, which makes test code much easier:

    https://github.com/Automattic/vip-go-mu-plugins/blob/56eb4475d502937f5fe2c5ed93c4ec95d0bf33f4/tests/vip-integrations/test-vip-integrations.php#L20-L28

2. Removed most of the filters we were using for a more pure, unit testable core. [The `vip-integrations.php` file](https://github.com/Automattic/vip-go-mu-plugins/blob/56eb4475d502937f5fe2c5ed93c4ec95d0bf33f4/vip-integrations.php) is now a thin wrapper that creates a `Integrations` instance, exposes an `activate()` function, and loads integrations using a single filter. All of the other filters were removed, separating plugin activation from the WP lifecycle. Ideally this file stays very small and wraps WordPress side-effects, and testing can be focused on the `Integrations` class.

Between these two changes I think the resulting code is easier to understand, has better tests, and can execute those tests faster without worrying about requiring test isolation.

Please take another look when you get a chance. Thank you!

tests/vip-integrations/test-vip-integrations.php
```diff
@@ -3,52 +3,85 @@
...
- * @preserveGlobalState disabled
- */
 class VIP_Integrations_Test extends WP_UnitTestCase {
 	public function setUp(): void {
-		// Remove existing default integrations to avoid conflicts when re-registering
-		Integrations::instance()->remove_registered();
+		parent::setUp();
+
+		// Reset static is_loaded status on integration so it can be reloaded
+		FakeIntegration::$is_loaded = false;
```
GaryJones: Why make the `FakeIntegration::is_loaded` property static? Why not leave it as a regular property? That way, each new test that starts by registering a new instance of FakeIntegration will automatically have the property set to false.

That way, you can also drop this `setUp()` method as well from the test case.

tests/vip-integrations/test-vip-integrations.php
```diff
@@ -3,52 +3,85 @@
...
-			Integrations::instance()->register( 'fake-integration', FakeIntegration::class );
-		});
+	// Registration
+
+	public function test_register_integration_as_class_name_loads() {
+		$integrations = new Integrations();
+		$integrations->register( 'fake-integration-class', FakeIntegration::class );
 
-		activate( 'fake-integration' );
+		$integrations->activate( 'fake-integration-class' );
```
GaryJones, line 22:
```suggestion
		// Only activated integrations are loaded by the collection class, so activate FakeIntegration.
		$integrations->activate( 'fake-integration-class' );
```

tests/vip-integrations/test-vip-integrations.php
```diff
@@ -3,52 +3,85 @@
...
 
-		$this->run_integration_actions();
-		$this->assertTrue( FakeIntegration::$is_integrated );
+		$integrations->load_active();
+		$this->assertTrue( FakeIntegration::$is_loaded );
 	}
 
-	public function test_register_integration_as_instantiated_class() {
+	public function test_register_integration_as_instantiated_class_loads() {
 		$fake_integration = new class() extends Integration {
```
GaryJones: Why are you repeating the definition of FakeIntegration as an anonymous class? Just register an object of the FakeIntegrations class.

```php
// Create the collection.
$integrations = new Integrations();
// Register an instantiated object.
$integrations->register( 'fake-integration-instance', new FakeIntegration() );
```

The only thing that should be different in this test compared to the previous one, is the use of `new FakeIntegration()` instead of `FakeIntegration::class`.

tests/vip-integrations/test-vip-integrations.php
```diff
@@ -3,52 +3,85 @@
...
+	public function test_double_slug_registration_throws() {
+		$this->expectException( InvalidArgumentException::class );
+
+		$an_integration = new class() extends Integration {
+			public function load( array $config ): void {}
+		};
+
+		$integrations = new Integrations();
+		$integrations->register( 'non-unique-slug', FakeIntegration::class );
+		$integrations->register( 'non-unique-slug', $an_integration );
```
GaryJones: No need to create an anonymous class - just try to register the same Integration twice:

```suggestion
		$integrations = new Integrations();
		$integrations->register( 'fake-integration', FakeIntegration::class );
		$integrations->register( 'fake-integration', FakeIntegration::class );
```

tests/vip-integrations/test-vip-integrations.php
```diff
@@ -3,52 +3,85 @@
...
+		$integrations = new Integrations();
+		$integrations->register( 'fake-integration', FakeIntegration::class );
+
+		$integrations->load_active();
+		$this->assertFalse( FakeIntegration::$is_loaded );
+	}
+
+	// Errors
+
+	public function test_double_slug_registration_throws() {
```
GaryJones: Consider adding the exception name to the test name:

```suggestion
	public function test_double_slug_registration_throws_invalidArgumentException() {
```

Tip: Run tests with `--testdox` to see how nicer test method names can help make them more descriptive.

tests/vip-integrations/test-vip-integrations.php
```diff
@@ -3,52 +3,85 @@
...
+		$integrations = new Integrations();
+		$integrations->register( 'non-unique-slug', FakeIntegration::class );
+		$integrations->register( 'non-unique-slug', $an_integration );
+	}
+
+	public function test_invalid_class_name_throws() {
+		$this->expectException( InvalidArgumentException::class );
+
+		$integrations = new Integrations();
+		$integrations->register( 'fake-integration', FakeIntegration::class . '-does-not-exist' );
```
GaryJones: No need to make PHP find and reference a real class here.

```suggestion
		$integrations->register( 'fake-integration', 'not-an-integration' );
```

tests/vip-integrations/test-vip-integrations.php
```diff
@@ -3,52 +3,85 @@
...
+	}
+
+	public function test_non_integration_subclass_throws() {
+		$this->expectException( InvalidArgumentException::class );
+
+		$random_class = new class() {};
+
+		$integrations = new Integrations();
+		$integrations->register( 'fake-integration', $random_class );
```
GaryJones: Could be simplified with a standard class instance instead of defining a new class:

```suggestion
		$integrations = new Integrations();
		$integrations->register( 'fake-integration', new stdClass() );
```

GaryJones (commented):
Some suggestions on the tests.

tests/vip-integrations/test-vip-integrations.php
alecgeatches: I agree, this is a bit awkward and I'll try to improve.

> Why make the `FakeIntegration::is_loaded` property static?

This class was originally created for class-based registration, where we don't have access to an instance:

```php
$integrations = new Integrations();
$integrations->register( 'fake-integration-class', FakeIntegration::class );

// Only activated integrations are loaded by the collection class, so activate FakeIntegration.
$integrations->activate( 'fake-integration-class' );

$integrations->load_active();
// Is FakeIntegration::class loaded?
```

tests/vip-integrations/test-vip-integrations.php
alecgeatches: The idea here was to demonstrate that the repeated slug is the problem, not the repeated class, but it adds some unnecessary complexity to the test. I'll remove it.

tests/vip-integrations/test-vip-integrations.php
alecgeatches: Good tip! I haven't used that before. I'll use a more verbose `test__<action>__<expected-outcome>` format for these.

alecgeatches:
@GaryJones Thank you for the test review! Made some changes based on your feedback in https://github.com/Automattic/vip-go-mu-plugins/pull/4458/commits/173dfa59e1bef99d8833a5b234a075a9530971aa. The main changes were:

1. Add `$is_loaded_class` and `$is_loaded_instance` to `FakeIntegration` class:

    https://github.com/Automattic/vip-go-mu-plugins/blob/173dfa59e1bef99d8833a5b234a075a9530971aa/tests/vip-integrations/fake-integration.php#L5-L17

    This will allow us to use a simpler `$fake_integration->is_loaded_instance` for most of our tests, but still cover class registration [where an instance isn't available](https://github.com/Automattic/vip-go-mu-plugins/pull/4458#discussion_r1214745978). Still not perfect, but allows more flexibility and brevity in tests.

2. Removed the `setUp()` integration flag reset and instead added it as a setup in the only test that cares about it:

    https://github.com/Automattic/vip-go-mu-plugins/blob/173dfa59e1bef99d8833a5b234a075a9530971aa/tests/vip-integrations/test-vip-integrations.php#L14-L19

    This prevents test reordering from causing issues and keeps other tests from doing unnecessary setup.

3. For all other tests, use `new FakeIntegration()` rather than anonymous classes. Along with the instance-specific `$is_loaded_instance`, makes test logic simple.

Thank you for all of your feedback! At this point we're looking to merge ASAP, so unless you spot any major problems we'd appreciate the goahead from you and the @Automattic/vip-platform-cantina team. Thank you!

GaryJones (dismissed):
👍🏻

vip-integrations.php
```diff
@@ -0,0 +1,38 @@
...
+function activate( string $integration_slug, array $config = [] ): void {
+	global $vip_integrations;
+	$vip_integrations->activate( $integration_slug, $config );
+}
+
+// Load integrations in muplugins_loaded:5 to allow integrations to hook
+// muplugins_loaded:10 or any later action
+add_action( 'muplugins_loaded', function() use ( $vip_integrations ) {
+	$vip_integrations->load_active();
+}, /* priority */ 5 );
```
mjangda:
```suggestion
}, 5 );
```

integrations/integrations.php
```diff
@@ -0,0 +1,91 @@
...
+	 * @private
+	 */
+	public function activate( string $integration_slug, array $config = [] ): void {
+		$integration = $this->get( $integration_slug );
+
+		if ( null === $integration ) {
+			throw new InvalidArgumentException( sprintf( 'VIP Integration with slug "%s" is not a registered integration.', $integration_slug ) );
+		} else {
+			$integration->activate( $config );
+		}
```
mjangda: Minor: the `else` is not necessary:

```suggestion
		$integration->activate( $config );
```

integrations/block-data-api.php
```diff
@@ -0,0 +1,29 @@
...
+	 * @private
+	 */
+	public function load( array $config ): void {
+		add_action( 'plugins_loaded', function() {
+			// Do not load plugin if already loaded by customer code
+			if ( defined( 'VIP_BLOCK_DATA_API_LOADED' ) ) {
+				return;
+			}
+
+			$load_path = __DIR__ . '/../../vip-integrations/vip-block-data-api-1.0.0/vip-block-data-api.php';
```
mjangda:
```suggestion
			$load_path = WPMU_PLUGIN_DIR . '/vip-integrations/vip-block-data-api-1.0.0/vip-block-data-api.php';
```

integrations/block-data-api.php
```diff
@@ -0,0 +1,29 @@
...
+ * @private
+ */
+class BlockDataApi extends Integration {
+	/**
+	 * Applies hooks to load Block Data API plugin.
+	 *
+	 * @private
+	 */
+	public function load( array $config ): void {
+		add_action( 'plugins_loaded', function() {
```
mjangda, line 27: One slight worry I have is that we'll forget to include this for a future integration and end up breaking some sites. We could handle this elsewhere, but I worry that could be hard to detangle in the future so let's keep it as-is for now. Let's just document it.

```suggestion
		// Wait until plugins_loaded to give precedence to the plugin in the customer repo
		add_action( 'plugins_loaded', function() {
```

integrations/integration.php
```diff
@@ -0,0 +1,62 @@
...
+	 *
+	 * @private
+	 */
+	public function get_config(): array {
+		return $this->config;
+	}
+
+	/**
+	 * Abstract base for integration functionality.
+	 * Implement custom action and filter calls to load integration here.
```
mjangda, line 57:
```suggestion
	 * Implement custom action and filter calls to load integration here.
	 *
	 * For plugins / integrations that can be added to customer repos, the implementation should hook into plugins_loaded and check if the plugin is already loaded first.
```

vip-integrations.php
```diff
@@ -0,0 +1,38 @@
...
+ */
+function activate( string $integration_slug, array $config = [] ): void {
+	global $vip_integrations;
+	$vip_integrations->activate( $integration_slug, $config );
+}
+
+// Load integrations in muplugins_loaded:5 to allow integrations to hook
+// muplugins_loaded:10 or any later action
+add_action( 'muplugins_loaded', function() use ( $vip_integrations ) {
+	$vip_integrations->load_active();
```
mjangda, line 37: @alecgeatches @ingeniumed @mehmoodak for integrations that are platform-managed, would we add that logic here or elsewhere? For example:

```
foreach ( $vip_integrations->get_registered() as $slug => $integration ) {
    if ( /* integration as marked as enabled in integrations config */ ) {
        $config = /* get config from integrations config */
        activate( $slug, $config );
    }
}

$vip_integrations->load_active();
```

mjangda (dismissed):
Some minor tweaks and questions. Other than that, we're looking good 👍

alecgeatches:
Note: Added a small update to use the [minor `1.0.1` release](https://github.com/Automattic/vip-block-data-api/releases/tag/1.0.1) which is also ready to go in [our vip-go-mu-plugins-ext PR](https://github.com/Automattic/vip-go-mu-plugins-ext/pull/35).

vip-integrations.php
mehmoodak: Theoretically it seems like we can add this logic but @alecgeatches or @ingeniumed knows best.

integrations/block-data-api.php
```diff
@@ -0,0 +1,30 @@
...
+	 */
+	public function load( array $config ): void {
+		// Wait until plugins_loaded to give precedence to the plugin in the customer repo
+		add_action( 'plugins_loaded', function() {
+			// Do not load plugin if already loaded by customer code
+			if ( defined( 'VIP_BLOCK_DATA_API_LOADED' ) ) {
+				return;
+			}
+
+			$load_path = WPMU_PLUGIN_DIR . '/vip-integrations/vip-block-data-api-1.0.1/vip-block-data-api.php';
```
rinatkhaziev: Should this really be hardcoded? 

I.e. what happens when automation removes 1.0.1 (which currently doesn't exist because the sibling PR is not merged yet)?

alecgeatches: > Should this really be hardcoded?

@rinatkhaziev Not a big fan of this being hardcoded either. This implementation was somewhat inspired by the wp-parsely implementation that [loops through version paths](https://github.com/Automattic/vip-go-mu-plugins/blob/develop/wp-parsely.php#L271) based on the hardcoded [`SUPPORTED_VERSIONS`](https://github.com/Automattic/vip-go-mu-plugins/blob/develop/wp-parsely.php#L19-L27), which is a bit abstracted but still requires version string hardcoding.

Here are some ideas @ingeniumed and I have to solve this:

1. Move the `1.0.1` to a separate function or constant in `integrations/block-data-api.php`. This doesn't fix the hardcoding but makes the separation a bit cleaner.
2. Use [`glob()`](https://www.php.net/manual/en/function.glob.php) to match version paths and select the latest one. This may be a bit tricky as we'd need to parse out some sort of semantic versioning and comparison. I think there may be potential for performance penalties as well, but because this code is only run when the plugin is activated it's lower risk.

We'd also like to adjust the code to set the `active` flag to `false` when the plugin path can't be loaded and add a test for that. It won't fix the hardcoded version but will allow us to test around it better.

Any suggestions on how we can avoid hardcoding the path here? It would be great if we can remove that.

rinatkhaziev: What if we pass the version to activate as a part of `$config` on `activate`?

ingeniumed: Good question, we didn't go down that path because:

- We want to be the ones managing the version. We don't want to support version pinning via a constant, or via what someone else calling this method passes in. So calling this method would mean you get a version we define - which will currently be the latest one but in the future can be changed including the implementation for how that is managed ([Alec posted this here, that touches on it](https://github.com/Automattic/vip-go-mu-plugins/pull/4406#issuecomment-1526996573)).
- We want this code to be generic, and not be directed at a specific plugin. In order to set the version [here in the config when `activate` is called](https://github.com/Automattic/vip-go-mu-plugins/blob/0efda5285de6b057d3f335468afabdfd0f4ae517/vip-integrations.php#L31), we would need to check what the plugin name is and set the version based on that. At the moment, there isn't really any specific code targeted at block data api within that linked code.

Given that, would option 1 be a good way to solve this? 

What that would involve:

- We define the version as a separate function/constant within [the block-data-api implementation](https://github.com/Automattic/vip-go-mu-plugins/blob/0efda5285de6b057d3f335468afabdfd0f4ae517/integrations/block-data-api.php#L10). We would then call that in the load function. That would ensure that the version is exposed within the plugin itself, in a way that's more visible. We also add a note above it to ensure that this version is carefully evaluated against the lowestVersion in mu-plugins-ext, and link the specific config.json.
- Currently, if the file is not found the plugin is not loaded. So if the version is wrong, the plugin itself won't load to be safe. We take that a step further and mark the active flag as false if the file is not found.

Would that be better? Thanks for the feedback!

integrations/block-data-api.php
```diff
@@ -0,0 +1,42 @@
...
+ */
+class BlockDataApi extends Integration {
+
+	/**
+	 * The version of the Block Data API plugin to load, that's set to the latest version.
+	 * This should be higher than the lowestVersion set in https://github.com/Automattic/vip-go-mu-plugins-ext/blob/trunk/config.json#L63
+	 * 
+	 * @var string
+	 */
+	protected string $version = '1.0';
```
rinatkhaziev, line 18: This now mismatches the path in ext (it.s currently there as 1.0.0 (and 1.0.1 in the unmerged PR).

alecgeatches: Tested locally with the latest changes in https://github.com/Automattic/vip-go-mu-plugins-ext/pull/35, -ext should pull `vip-integrations/vip-block-data-api-1.0` for any `1.0.x` tag. The -ext PR will need to be merged first.

Issue #4193, opened 2023-03-08
PHPUnit Deprecation Warnings


sjinks:
```
1) Automattic\VIP\Search\Queue_Test::test_requeue_with_different_priority
assertObjectHasAttribute() is deprecated and will be removed in PHPUnit 10.

2) Automattic\VIP\Search\Search_Test::test__maybe_alert_for_prolonged_query_limiting with data set #3 (7201, true)
Expecting E_WARNING and E_USER_WARNING is deprecated and will no longer be possible in PHPUnit 10.

3) Automattic\VIP\Prometheus\Test_Prometheus::test_create_registry_wrong_backend
Expecting E_WARNING and E_USER_WARNING is deprecated and will no longer be possible in PHPUnit 10.

4) Automattic\VIP\Cache\Vary_Cache_Test::test__register_groups__invalid with data set "invalid-group-name" (array('dev-group---__'), 'invalid_vary_group_name')
Expecting E_WARNING and E_USER_WARNING is deprecated and will no longer be possible in PHPUnit 10.

5) Automattic\VIP\Cache\Vary_Cache_Test::test__register_groups__invalid with data set "invalid-group-array" (array('dev-group', 'dev-group---__'), 'invalid_vary_group_name')
Expecting E_WARNING and E_USER_WARNING is deprecated and will no longer be possible in PHPUnit 10.

6) Automattic\VIP\Cache\Vary_Cache_Test::test__register_groups__did_send_headers
Expecting E_WARNING and E_USER_WARNING is deprecated and will no longer be possible in PHPUnit 10.

7) Automattic\VIP\Files\Acl\VIP_Files_Acl_Test::test__maybe_load_restrictions__no_constant_and_with_one_option
Expecting E_WARNING and E_USER_WARNING is deprecated and will no longer be possible in PHPUnit 10.

8) Automattic\VIP\Files\Acl\VIP_Files_Acl_Test::test__send_visibility_headers__invalid_visibility
Expecting E_WARNING and E_USER_WARNING is deprecated and will no longer be possible in PHPUnit 10.

9) Automattic\VIP\Files\WP_Filesystem_VIP_Test::test__get_transport_for_path__disallowed_write
Expecting E_ERROR and E_USER_ERROR is deprecated and will no longer be possible in PHPUnit 10.

[...]

PR #4463, opened 2023-05-16
tests: Fix 'Expecting E_WARNING and E_USER_WARNING' deprecation

sjinks:
This PR fixes a missed test case with a deprecation warning.

Related: #4193
Related: #4393

Ref: https://github.com/Automattic/vip-go-mu-plugins/actions/runs/4990322583/jobs/8935370475?pr=4454

### DIFF
--- a/tests/test-vip-cache-manager.php
+++ b/tests/test-vip-cache-manager.php
@@ -1,6 +1,7 @@
 <?php
 
 class VIP_Go_Cache_Manager_Test extends WP_UnitTestCase {
+	/** @var WPCOM_VIP_Cache_Manager */
 	public $cache_manager;
 
 	public function setUp(): void {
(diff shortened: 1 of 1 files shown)


tests/test-vip-cache-manager.php
```diff
@@ -66,12 +77,14 @@ public function test__valid__queue_purge_url( $queue_url, $expected_urls ) {
 	 * @dataProvider get_data_for_invalid_queue_purge_url_test
 	 */
 	public function test__invalid__queue_purge_url( $queue_url ) {
-		$this->expectWarning();
-
-		$actual_output = $this->cache_manager->queue_purge_url( $queue_url );
-
-		$this->assertFalse( $actual_output, 'Return value from `queue_purge_url` should be false.' );
-		$this->assertEmpty( $this->cache_manager->get_queued_purge_urls(), 'List of queued purge urls should be empty' );
```
sjinks, line 74: When we `expectWarning()`/`expectException()`/…, the code that follows the statement that is supposed to pit a warning or throw an exception, does not run. Therefore, if we really need to run some checks, we have to catch the exception ourselves.

sjinks: See #4464

PR #4464, opened 2023-05-16
tests: Fix `expectException()` usage

sjinks:
When we `expectException()`, the code that follows the statement that throws an exception, does not run.

For example,

```php
$this->expectException( \Exception::class );

$actual_result = Vary_Cache::register_groups( [
	'dev-group',
	'design-group',
] );

$this->assertFalse( $actual_result, 'register_groups after send_headers did not return false' );
$this->assertEquals( [], Vary_Cache::get_groups(), 'Registered groups are not empty.' );
```

`Vary_Cache::register_groups()` spits a warning that gets converted into an `Exception` (we do this with our custom error handler; PHPUnit employs a similar technique). Assertions that follow `Vary_Cache::register_groups()`, are not run.

The same thing happens here:

```php
function send_visibility_headers( $file_visibility, $file_path ) {
	// ...
	switch ( $file_visibility ) {
		// ...
		default:
			trigger_error( sprintf( 'Invalid file visibility (%s) ACL set for %s', $file_visibility, $file_path ), E_USER_WARNING );
			break;
	}

	http_response_code( $status_code );

	if ( null !== $is_private ) {
		$private_header_value = $is_private ? 'true' : 'false';
		header( sprintf( 'X-Private: %s', $private_header_value ) );
	}
}
```

```php
	public function test__send_visibility_headers__invalid_visibility() {
		$this->expectException( \Exception::class );
		$this->expectExceptionMessage( 'Invalid file visibility (NOT_A_VISIBILITY) ACL set for /wp-content/uploads/invalid.jpg' );

		send_visibility_headers( 'NOT_A_VISIBILITY', '/wp-content/uploads/invalid.jpg' );

// The checks below will not run:
		$this->assertEquals( 500, http_response_code(), 'Status code does not match expected' );

		$headers = headers_list();
		$this->assertNotContains( 'X-Private: true', $headers, 'Sent headers include X-Private: true header but should not.', true );
		$this->assertNotContains( 'X-Private: false', $headers, 'Sent headers include X-Private:false header but should not.', true );
	}
```

When the code spits a warning that we convert to an exception, no code that follows the guilty statement is run; in this case, we leave the function right after `trigger_error()`; no headers are sent.

### DIFF
--- a/tests/files/acl/test-acl.php
+++ b/tests/files/acl/test-acl.php
@@ -3,6 +3,8 @@
 namespace Automattic\VIP\Files\Acl;
 
 use Automattic\Test\Constant_Mocker;
+use Exception;
+use Throwable;
 use WP_UnitTest_Factory;
 use WP_UnitTestCase;
 
@@ -44,12 +46,14 @@ class VIP_Files_Acl_Test extends WP_UnitTestCase {
 	public function test__maybe_load_restrictions__no_constant_and_with_one_option() {
 		update_option( 'vip_files_acl_restrict_all_enabled', 1 );
 
-		$this->expectException( \Exception::class );
-		$this->expectExceptionMessage( 'File ACL restrictions are enabled without server configs' );
-
-		maybe_load_restrictions();
-
-		$this->assertEquals( false, has_filter( 'vip_files_acl_file_visibility' ) );
+		try {
+			maybe_load_restrictions();
+			self::assertTrue( false );
+		} catch ( Throwable $e ) {
+			self::assertInstanceOf( Exception::class, $e );
+			self::assertStringContainsString( 'File ACL restrictions are enabled without server configs', $e->getMessage() );
+			self::assertFalse( has_filter( 'vip_files_acl_file_visibility' ) );
+		}
 	}
 
 	public function test__maybe_load_restrictions__constant_and_restrict_all_option() {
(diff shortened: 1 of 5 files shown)


tests/files/acl/test-acl.php
```diff
@@ -44,12 +46,14 @@ public function test__maybe_load_restrictions__no_constant_and_no_options() {
...
 
-		$this->expectException( \Exception::class );
-		$this->expectExceptionMessage( 'File ACL restrictions are enabled without server configs' );
-
-		maybe_load_restrictions();
-
-		$this->assertEquals( false, has_filter( 'vip_files_acl_file_visibility' ) );
+		try {
+			maybe_load_restrictions();
+			self::assertTrue( false );
```
rebeccahum: I'm not sure what this is doing exactly.

rebeccahum (changes requested):
Just left one comment on one thing I don't fully understand.

Also, I'm wondering if we need to figure out a different way to test for the ones we've removed after the `expectException()`.

tests/files/acl/test-acl.php
sjinks: This code makes sure that `maybe_load_restrictions();` spits a warning (that gets converted into an exception by our helper error handler). If exception is not thrown, `assertTrue( false )` makes the test fail. This tests the case when `maybe_load_exception()` runs without warnings.

rebeccahum: Can we just add a comment there for future reference to why we are doing it this way? Makes sense now, thank you!

rebeccahum: Also in #4463 too

rebeccahum (approved):

PR #4465, opened 2023-05-16
tests: Do not contact external sites in Site_Details_Index_Test

sjinks:
When our tests contact `wordpress.org`, a request may fail; that causes tests to fail.

Ideally, unit tests should not contact any external resources.

This PR fixes `Site_Details_Index_Test`; `vip_site_details_index_data` indirectly calls `wp_update_themes()`, which, in turn, tries to access `https://api.wordpress.org/themes/update-check/1.1/`.

Ref: https://github.com/Automattic/vip-go-mu-plugins/actions/runs/4990322583/jobs/8935370475?pr=4454


rebeccahum (approved):
Cooool. Thank you for this! It's been happening quite a bit recently.

PR #4494, opened 2023-05-24
Filter blog_public sooner

pschoffer:
## Description
Filtering `blog_public` on sync could cause an issue on different syncs the JP is doing that is not invoking filters we used before.

This change moves the filter sooner to leverage core's option filter which then gets used before any access to that option.

### Plugin Updated: Jetpack

On non-production environments, we enforce `blog_public` option in a more robust way.

## Checklist

## Steps to Test

This command works for me on local dev-env:
```
dev-env exec --slug vip-local -- wp eval '$opt = new Automattic\Jetpack\Sync\Modules\Options(); $opt->set_defaults(); $all_options = $opt->get_all_options(); var_dump($all_options["blog_public"]);'
```

### DIFF
--- a/security/class-private-sites.php
+++ b/security/class-private-sites.php
@@ -54,7 +54,6 @@ class Private_Sites {
 
 		add_filter( 'jetpack_active_modules', array( $this, 'filter_jetpack_active_modules' ) );
 		add_filter( 'jetpack_get_available_modules', array( $this, 'filter_jetpack_get_available_modules' ) );
-		add_filter( 'jetpack_sync_before_send_jetpack_full_sync_options', array( $this, 'filter_blog_public_option_for_full_sync' ), 11 );
 		add_filter( 'pre_option_blog_public', fn() => '-1' );
 
 		$this->disable_core_feeds();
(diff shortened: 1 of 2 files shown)


pschoffer:
I went through the uses of `blog_public` in core - https://github.com/search?q=repo%3AWordPress%2FWordPress%20blog_public&type=code

There is only one code section that I found where hardcoding `-1` as oppose to `1` which is [default](https://github.com/WordPress/WordPress/blob/2cd27b23495856227ed3f34aadb53d2ce702a830/wp-admin/includes/schema.php#L473) would make any impact.

https://github.com/WordPress/WordPress/blob/2cd27b23495856227ed3f34aadb53d2ce702a830/wp-admin/options-writing.php#L209-L210
That code shows msg about notifying or not some downstream services. Looking at the notification itself the `blog_public` seems to be not used at all so the line is a bit misleading.

In any case, I don't think that is a big issue for non-prod environments.


---- 

There are also a few places in core that check for the truthiness of `get_option( 'blog_public' );` where our change won't make it better nor worse. Those code snippets are aimed to discourage search indexers, so ideally, we would like it to be false, but that to me feels like an independent issue.

sjinks (approved):

PR #4508, opened 2023-05-25
ci: Workaround for failing core tests

sjinks:
Core tests (namely, `test_package_json`) fail because of this bug:

https://github.com/WordPress/wordpress-develop/blob/6.2.2/package.json#L3

The version specified in `package.json` (6.2.3) must match the actual WP version (6.2.2): https://github.com/WordPress/wordpress-develop/blob/6.2.2/tests/phpunit/tests/basic.php#L54

We fix the test by patching `package.json` to set the correct version.

### DIFF
--- a/.github/workflows/core-tests.yml
+++ b/.github/workflows/core-tests.yml
@@ -50,6 +50,11 @@ jobs:
           submodules: recursive
           path: wordpress/src/wp-content/mu-plugins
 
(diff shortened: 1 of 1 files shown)


rebeccahum:
~@rinatkhaziev What's the likelihood of getting Core to fix this?~

Edit, apparently 0% lol

rebeccahum (approved):

PR #4561, opened 2023-06-08
fix: Type error when callable is not a method

### LINKED ISSUE #4542: Some filters may cause fatal error in class-search.php
I encountered a fatal error in `/search/includes/classes/class-search.php` when filtering `ep_weighting_fields_for_post_type`. I can work around this by saving a weight config which bypasses the code in question, however I believe there's a bug in the code which I'll outline below.

The relevant code is here:

https://github.com/Automattic/vip-go-mu-plugins/blob/a7db1ed7d74054340f8923ad1f7bf8feb94e92f1/search/includes/classes/class-search.php#L2344-L2347

## Expected/Desired Behavior

It should be possible to add filters without a fatal error.

## Actual Behavior

Filtering any of the following may cause a fatal error when [this code](https://github.com/Automattic/vip-go-mu-plugins/blob/a7db1ed7d74054340f8923ad1f7bf8feb94e92f1/search/includes/classes/class-search.php#L2344-L2347) is executed.

* 'ep_weighting_configuration_for_search',
* 'ep_query_weighting_fields',
* 'ep_weighting_configuration',
* 'ep_weighting_fields_for_post_type',

This happens because the code assumes that the filter callback will be either a closure or an object, but it may also be a string representing a callable function. The fatal error is as follows:

```
Fatal error: Uncaught Error: get_class(): Argument #1 ($object) must be of type object, string given
in /var/www/cacm/wp-content/mu-plugins/search/includes/classes/class-search.php on line 2344
```

The code in question expects the callback within the `$wp_filter` global to be in this format, where the `function` key contains an object:

```
array(1) {
      ["00000000000005420000000000000000weighting_fields_for_post_type"]=>
      array(2) {
        ["function"]=>
        array(2) {
          [0]=>
          object(ElasticPress\Feature\SearchOrdering\SearchOrdering)#1346 (9) {
            ["slug"]=>
            string(14) "searchordering"
            ["title"]=>
            string(21) "Custom Search Results"
            ["summary"]=>
            string(70) "Insert specific posts into search results for specific search queries."
            ["docs_url"]=>
            string(135) "https://elasticpress.zendesk.com/hc/en-us/articles/360050447492-Configuring-ElasticPress-via-the-Plugin-Dashboard#custom-search-results"
            ["default_settings"]=>
            array(0) {
            }
            ["requires_install_reindex"]=>
            bool(false)
            ["order"]=>
            NULL
            ["group_order"]=>
            NULL
            ["available_during_installation"]=>
            bool(false)
          }
          [1]=>
          string(30) "weighting_fields_for_post_type"
        }
        ["accepted_args"]=>
        int(2)
      }
    }
```

However, [if you follow the documentation for that filter in from elasticpress](https://www.elasticpress.io/blog/2020/06/custom-fields-and-weighted-search-with-elasticpress-part-2/), the callback will be a string and the code in question will pass the first character of the string to the `get_class()` function. In that scenario, the callback might look like this:

```
array(1) {
      ["epio_blog_add_custom_field_ep_weighting"]=>
      array(2) {
        ["function"]=>
        string(39) "epio_blog_add_custom_field_ep_weighting"
        ["accepted_args"]=>
        int(2)
      }
    }
```

1. Unset the `elasticpress_weighting` option. If it's set, the code in question won't execute due to [this condition](https://github.com/Automattic/vip-go-mu-plugins/blob/a7db1ed7d74054340f8923ad1f7bf8feb94e92f1/search/includes/classes/class-search.php#L2311-L2313).
1. Add a callable function (not an object) to one of the listed filters. I've done my testing with `'ep_weighting_fields_for_post_type'`
1. Perform a search and you'll see a fatal error, as indicated above.

Note that saving the weighting configuration page bypasses this error, but I think that's just a bandaid. In my opinion, the custom code in VIP's search implementation which iterates through the listed filters should be updated to handle a third case, when the callback function is a callable instead of a closure or object, gracefully.

sjinks:
## Description

Fixes: #4542

Fix type error when a callable passed to `ep_weighting_configuration_for_search`, `ep_query_weighting_fields`, `ep_weighting_configuration`, or `ep_weighting_fields_for_post_type` is not a method.

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

See #4542

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -2338,7 +2338,7 @@ class Search {
 			if ( isset( $wp_filter[ $ep_filter ] ) ) {
 				foreach ( $wp_filter[ $ep_filter ]->callbacks as $callback ) {
 					foreach ( $callback as $el ) {
-						if ( $el['function'] instanceof \Closure ) {
+						if ( ! is_array( $el['function'] ) || ! is_object( $el['function'][0] ) ) {
 							return $should_do_weighting;
 						} else {
 							$class = get_class( $el['function'][0] );


rebeccahum (approved):
Thanks!

PR #4579, opened 2023-06-12
Update invalid login/forgotten password locking logic

pschoffer:
## Description

We want to make security behaviour around account locking during unsuccessful login more consistent as well as more configurable. This change refactors code and implements more filters so that sites can tweak the working of this more.

In order to disconnect the window of when the failed attempts are recorded from the period of how long the account is locked, we introduced a set of new cache keys. 

We were and continue using `<ip>`, `<username>` and `<ip>|<username>` for holding the count. These keys' expiration time serves as the event recording window for when we need to "collect" enough failed attempts. They used to also serve for tracking down if the account/ip is locked or not.

This changed and we are now using `locked_<ip>`, `locked_<username>` and `locked_<ip>|<username>` to keep a separate expiration period for the lock itself.

### Plugin Updated: Security

We have a plugin that temporarily locks an account upon a certain number of failed login attempts as well as password recovery attempts.

This change changes defaults for the most common scenario of attempts from the same IP address from 5 failed attempts within 5 minutes to 3 failed attempts within 15 minutes. We also made the lock time 30 minutes.

We also made this plugin more configurable via filters. For example following configuration will change the behaviour to lock an account after 5 attempts within 2 minutes. The account then stays locked for 10 minutes.

```php
add_filter( 'wpcom_vip_ip_username_login_threshold', fn() => 5 );
add_filter( 'wpcom_vip_invalid_login_window', fn() => MINUTE_IN_SECONDS * 2 );
add_filter( 'wpcom_vip_invalid_login_lock_period', fn() => MINUTE_IN_SECONDS * 10 );
```

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test
1. Login with bad password 3 times
2. Should get following error:
image

### DIFF
--- a/security.php
+++ b/security.php
@@ -58,9 +58,9 @@ function wpcom_vip_is_restricted_username( $username ) {
 }
 
 /**
- * PRIVATE: This function is not intended to be used outside of this file and is subject to change without notice.
+ * Return the defaults for login and password reset authorization limits, windows, and lockout periods.
  *
- * Returns the defualts for limits.
+ * @internal The values are subject to change without notice.
  */
 function _vip_get_auth_count_limit_defaults() {
 	if ( Context::is_fedramp() ) {
@@ -102,9 +102,9 @@ function _vip_get_auth_count_limit_defaults() {
 }
 
 /**
- * PRIVATE: This function is not intended to be used outside of this file and is subject to change without notice.
+ * Return the cache keys used for login and forgotten password rate limiting.
  *
- * Returns the cache keys used for login and forgotten passoword rate limiting.
+ * @internal
  *
  * @param string $raw_username The username to use for the cache keys.
  * @return array Associative array of cache keys.
@@ -123,10 +123,10 @@ function _vip_login_cache_keys( $raw_username ) {
 }
 
 /**
- * PRIVATE: This function is not intended to be used outside of this file and is subject to change without notice.
- *
- * Checks if counts for the given username and IP address are above the given thresholds.
+ * Check if counts for the given username and IP address are above the given thresholds.
  * If so, the account is locked for a period of time.
+ *
+ * @internal
  *
  * @param string $username The username to check.
  * @param string $cache_group The cache group to use for the rate limiting.
--- a/security.php
+++ b/security.php
@@ -65,33 +65,39 @@ function wpcom_vip_is_restricted_username( $username ) {
 function _vip_get_auth_count_limit_defaults() {
 	if ( Context::is_fedramp() ) {
 		return [
-			'ip_username'         => 3,
-			'ip'                  => 5,
-			'username'            => 10,
+			'ip_username_login'          => 3,
+			'ip_login'                   => 5,
+			'username_login'             => 10,
+			'ip_username_password_reset' => 3,
+			'ip_password_reset'          => 3,
+			'username_password_reset'    => 15,
 
-			'ip_username_window'  => MINUTE_IN_SECONDS * 15,
-			'ip_window'           => MINUTE_IN_SECONDS * 15,
-			'username_window'     => MINUTE_IN_SECONDS * 15,
+			'ip_username_window'         => MINUTE_IN_SECONDS * 15,
+			'ip_window'                  => MINUTE_IN_SECONDS * 15,
+			'username_window'            => MINUTE_IN_SECONDS * 15,
 
-			'ip_username_lockout' => MINUTE_IN_SECONDS * 30,
-			'ip_lockout'          => MINUTE_IN_SECONDS * 30,
-			'username_lockout'    => MINUTE_IN_SECONDS * 30,
+			'ip_username_lockout'        => MINUTE_IN_SECONDS * 30,
+			'ip_lockout'                 => MINUTE_IN_SECONDS * 30,
+			'username_lockout'           => MINUTE_IN_SECONDS * 30,
 		];
 	}
 
 
 	return [
-		'ip_username'         => 5,
-		'ip'                  => 50,
-		'username'            => 25,
+		'ip_username_login'          => 5,
+		'ip_login'                   => 50,
+		'username_login'             => 25,
+		'ip_username_password_reset' => 3,
+		'ip_password_reset'          => 3,
+		'username_password_reset'    => 15,
 
-		'ip_username_window'  => MINUTE_IN_SECONDS * 5,
-		'ip_window'           => MINUTE_IN_SECONDS * 60,
-		'username_window'     => MINUTE_IN_SECONDS * 25,
+		'ip_username_window'         => MINUTE_IN_SECONDS * 5,
+		'ip_window'                  => MINUTE_IN_SECONDS * 60,
+		'username_window'            => MINUTE_IN_SECONDS * 25,
 
-		'ip_username_lockout' => MINUTE_IN_SECONDS * 5,
-		'ip_lockout'          => MINUTE_IN_SECONDS * 60,
-		'username_lockout'    => MINUTE_IN_SECONDS * 25,
+		'ip_username_lockout'        => MINUTE_IN_SECONDS * 5,
+		'ip_lockout'                 => MINUTE_IN_SECONDS * 60,
+		'username_lockout'           => MINUTE_IN_SECONDS * 25,
 	];
 }
 
@@ -142,7 +148,7 @@ function _vip_maybe_temporary_lock_account( $username, $cache_group ) {
 	 * @param string $ip        IP address of the login request.
 	 * @param string $username  Username of the login request.
 	 */
-	$ip_username_threshold = apply_filters( "wpcom_vip_ip_username_{$event_type}_threshold", $defaults['ip_username'], $ip, $username );
+	$ip_username_threshold = apply_filters( "wpcom_vip_ip_username_{$event_type}_threshold", $defaults[ "ip_username_{$event_type}" ], $ip, $username );
 
 	/**
 	 * Filters the threshold for limiting logins by IP.
@@ -150,7 +156,7 @@ function _vip_maybe_temporary_lock_account( $username, $cache_group ) {
 	 * @param int    $threshold IP login threshold.
 	 * @param string $ip        IP address of the login request.
 	 */
-	$ip_threshold = apply_filters( "wpcom_vip_ip_{$event_type}_threshold", $defaults['ip'], $ip );
+	$ip_threshold = apply_filters( "wpcom_vip_ip_{$event_type}_threshold", $defaults[ "ip_{$event_type}" ], $ip );
 
 	/**
 	 * Filters the threshold for limiting logins by username.
@@ -158,7 +164,7 @@ function _vip_maybe_temporary_lock_account( $username, $cache_group ) {
 	 * @param int    $threshold Username login threshold.
 	 * @param string $username  Username of the login request.
 	 */
-	$username_threshold = apply_filters( "wpcom_vip_username_{$event_type}_threshold", $defaults['username'], $username );
+	$username_threshold = apply_filters( "wpcom_vip_username_{$event_type}_threshold", $defaults[ "username_{$event_type}" ], $username );
 
 	$is_restricted_username = wpcom_vip_is_restricted_username( $username );
 	if ( $is_restricted_username ) {
--- a/security.php
+++ b/security.php
@@ -54,6 +54,19 @@ function wpcom_vip_is_restricted_username( $username ) {
 		|| WPCOM_VIP_MACHINE_USER_EMAIL === $username;
 }
 
+function _wpcom_vip_login_cache_keys( $raw_username ) {
+	$username = vip_strict_sanitize_username( $raw_username );
+
+	// phpcs:ignore WordPressVIPMinimum.Variables.ServerVariables.UserControlledHeaders
+	$ip = filter_var( $_SERVER['REMOTE_ADDR'] ?? '', FILTER_VALIDATE_IP, [ 'options' => [ 'default' => '' ] ] );
+
+	return [
+		'ip_username_cache_key' => $ip . '|' . $username,
+		'ip_cache_key'          => $ip,
+		'username_cache_key'    => $username,
+	];
+}
+
 /**
  * Tracks and caches IP, IP|Username events, and Username events.
  * We're tracking IP, and IP|Username events for both login attempts and
@@ -64,11 +77,7 @@ function wpcom_vip_is_restricted_username( $username ) {
  * @param int $cache_expiry The number in seconds of the cache expiry.
  */
 function wpcom_vip_track_auth_attempt( $username, $cache_group, $cache_expiry ) {
-	// phpcs:ignore WordPressVIPMinimum.Variables.ServerVariables.UserControlledHeaders
-	$ip                    = filter_var( $_SERVER['REMOTE_ADDR'] ?? '', FILTER_VALIDATE_IP, [ 'options' => [ 'default' => '' ] ] );
-	$ip_username_cache_key = $ip . '|' . $username; // IP + username
-	$ip_cache_key          = $ip; // IP only
-	$username_cache_key    = $username; // Username only
+	$cache_keys = _wpcom_vip_login_cache_keys( $username );
 
 	// Longer TTL when logging in as admin, which we don't allow on WP.com
 	$is_restricted_username = wpcom_vip_is_restricted_username( $username );
@@ -77,33 +86,24 @@ function wpcom_vip_track_auth_attempt( $username, $cache_group, $cache_expiry ) 
 		$cache_expiry = HOUR_IN_SECONDS + $cache_expiry;
 	}
 
-	wp_cache_add( $ip_username_cache_key, 0, $cache_group, $cache_expiry ); // phpcs:ignore WordPressVIPMinimum.Performance.LowExpiryCacheTime.CacheTimeUndetermined
-	wp_cache_add( $ip_cache_key, 0, $cache_group, HOUR_IN_SECONDS );
-	wp_cache_add( $username_cache_key, 0, $cache_group, MINUTE_IN_SECONDS * 15 );
-	wp_cache_incr( $ip_username_cache_key, 1, $cache_group );
-	wp_cache_incr( $ip_cache_key, 1, $cache_group );
-	wp_cache_incr( $username_cache_key, 1, $cache_group );
+	wp_cache_add( $cache_keys['ip_username_cache_key'], 0, $cache_group, $cache_expiry ); // phpcs:ignore WordPressVIPMinimum.Performance.LowExpiryCacheTime.CacheTimeUndetermined
+	wp_cache_add( $cache_keys['ip_cache_key'], 0, $cache_group, HOUR_IN_SECONDS );
+	wp_cache_add( $cache_keys['username_cache_key'], 0, $cache_group, MINUTE_IN_SECONDS * 15 );
+	wp_cache_incr( $cache_keys['ip_username_cache_key'], 1, $cache_group );
+	wp_cache_incr( $cache_keys['ip_cache_key'], 1, $cache_group );
+	wp_cache_incr( $cache_keys['username_cache_key'], 1, $cache_group );
 }
 
 function wpcom_vip_login_limiter( $username ) {
-	// Do some extra sanitization on the username.
-	$username = vip_strict_sanitize_username( $username );
-
 	wpcom_vip_track_auth_attempt( $username, CACHE_GROUP_LOGIN_LIMIT, MINUTE_IN_SECONDS * 5 );
 }
 add_action( 'wp_login_failed', 'wpcom_vip_login_limiter' );
 
 function wpcom_vip_login_limiter_on_success( $username ) {
-	// Do some extra sanitization on the username.
-	$username = vip_strict_sanitize_username( $username );
+	$cache_keys = _wpcom_vip_login_cache_keys( $username );
 
-	// phpcs:ignore WordPressVIPMinimum.Variables.ServerVariables.UserControlledHeaders
-	$ip                    = filter_var( $_SERVER['REMOTE_ADDR'] ?? '', FILTER_VALIDATE_IP, [ 'options' => [ 'default' => '' ] ] );
-	$ip_username_cache_key = $ip . '|' . $username; // IP + username
-	$ip_cache_key          = $ip; // IP only
-
-	wp_cache_decr( $ip_username_cache_key, 1, CACHE_GROUP_LOGIN_LIMIT );
-	wp_cache_decr( $ip_cache_key, 1, CACHE_GROUP_LOGIN_LIMIT );
+	wp_cache_decr( $cache_keys['ip_username_cache_key'], 1, CACHE_GROUP_LOGIN_LIMIT );
+	wp_cache_decr( $cache_keys['ip_cache_key'], 1, CACHE_GROUP_LOGIN_LIMIT );
 }
 add_action( 'wp_login', 'wpcom_vip_login_limiter_on_success' );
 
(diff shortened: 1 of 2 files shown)


security.php
```diff
@@ -54,56 +55,148 @@ function wpcom_vip_is_restricted_username( $username ) {
...
+		$lock_interval = apply_filters( 'wpcom_vip_invalid_login_lock_period', MINUTE_IN_SECONDS * 30  );
+		if ( $cache_group === CACHE_GROUP_LOGIN_LIMIT ) {
+			do_action( 'login_limit_exceeded', $username );
+		} else {
+			do_action( 'password_reset_limit_exceeded', $username );
+		}
+
+		if ( $ip_username_count >= $ip_username_threshold ) {
+			wp_cache_set( CACHE_KEY_LOCK_PREFIX . $cache_keys['ip_username_cache_key'], true, $cache_group, $lock_interval );
+		} else if ( $ip_count >= $ip_threshold ) {
```
sjinks:
```suggestion
		} elseif ( $ip_count >= $ip_threshold ) {
```

SonarCloud insists on this 🤷🏽‍♂️

security.php
```diff
@@ -18,6 +18,7 @@
 
 define( 'CACHE_GROUP_LOGIN_LIMIT', 'login_limit' );
 define( 'CACHE_GROUP_LOST_PASSWORD_LIMIT', 'lost_password_limit' );
+define( 'CACHE_KEY_LOCK_PREFIX', 'locked_' );
```
GaryJones, line 23:
```suggestion
define( 'CACHE_KEY_LOCK_PREFIX', 'vip_locked_' );
```

I wonder if adding a VIP prefix to this would help namespace / deliminate it from other plugins who may cache values  with a `locked_` prefix.

security.php
```diff
@@ -54,56 +55,148 @@ function wpcom_vip_is_restricted_username( $username ) {
 		|| WPCOM_VIP_MACHINE_USER_EMAIL === $username;
 }
 
+function _wpcom_vip_login_cache_keys( $raw_username ) {
```
GaryJones: Instead of using an underscore prefix to (presumably) indicate privateness, stick to WP code standards and avoid the underscore but document the function instead.

GaryJones: As an aside, we may want to consider the prefix for newly introduced functions - I would guess that you've gone for consistency with other functions in the file, but even the next line shows we have functions defined with just a `vip_` prefix as well. We've not had the WPCOM branding for a while now :D

security.php
```diff
@@ -54,56 +55,148 @@ function wpcom_vip_is_restricted_username( $username ) {
...
+	$username = vip_strict_sanitize_username( $raw_username );
+
+	// phpcs:ignore WordPressVIPMinimum.Variables.ServerVariables.UserControlledHeaders
+	$ip = filter_var( $_SERVER['REMOTE_ADDR'] ?? '', FILTER_VALIDATE_IP, [ 'options' => [ 'default' => '' ] ] );
+
+	return [
+		'ip_username_cache_key' => $ip . '|' . $username,
```
GaryJones, line 119: If the `$ip` is empty, can cache keys start with a pipe character?

security.php
```diff
@@ -54,56 +55,148 @@ function wpcom_vip_is_restricted_username( $username ) {
...
+		'ip_cache_key'          => $ip,
```
GaryJones, line 120: If the `$ip` is empty, will the cache key be empty?

security.php
```diff
@@ -54,56 +55,148 @@ function wpcom_vip_is_restricted_username( $username ) {
...
+		'username_cache_key'    => $username,
+	];
+}
+
+function _wpcom_vip_maybe_temporary_lock_account( $username, $cache_group ) {
+	$cache_keys = _wpcom_vip_login_cache_keys( $username );
+
+	$ip_username_count = wp_cache_get( $cache_keys['ip_username_cache_key'], $cache_group );
+	$ip_count          = wp_cache_get( $cache_keys['ip_cache_key'], $cache_group );
+	$username_count    = wp_cache_get( $cache_keys['username_cache_key'], $cache_group );
```
GaryJones, line 140: Could `wp_cache_get_multiple()` be used here yet? cc @WPprodigy

security.php
```diff
@@ -54,56 +55,148 @@ function wpcom_vip_is_restricted_username( $username ) {
...
+	 * Filters the threshold for limiting logins by username.
+	 *
+	 * @param int    $threshold Username login threshold.
+	 * @param string $username  Username of the login request.
+	 */
+	$username_threshold = apply_filters( 'wpcom_vip_username_login_threshold', 10, $username );
+
+	/**
+	 * Change the thresholds for Password Resets
+	 */
```
GaryJones:
```suggestion
	// Change the thresholds for Password Resets.
```

security.php
```diff
@@ -54,56 +55,148 @@ function wpcom_vip_is_restricted_username( $username ) {
...
+		 * @param int    $threshold Username password reset threshold. Default is 5 times the IP + username combination threshold.
+		 * @param string $username  Username of the password reset request
+		 */
+		$username_threshold = apply_filters( 'wpcom_vip_username_password_reset_threshold', 5 * $ip_username_threshold, $username );
+	} elseif ( $is_restricted_username ) {
+		$ip_username_threshold = 2;
+	}
+
+	if ( $ip_username_count >= $ip_username_threshold || $ip_count >= $ip_threshold || $username_count >= $username_threshold ) {
+		$lock_interval = apply_filters( 'wpcom_vip_invalid_login_lock_period', MINUTE_IN_SECONDS * 30  );
```
GaryJones: I wonder if there's a case when someone would want a 30 minute lockout period for passing username threshold, but 10 minutes for a username + IP threshold, for instance?

If the passing counts and thresholds were passed in as another filter argument, then developers have the opportunity to fine-tune the lockout period per the type of threshold that was triggered.

security.php
GaryJones: Also, shouldn't `CACHE_KEY_LOCK_PREFIX` be used here, or in `_wpcom_vip_login_cache_keys()`? It's used with `wp_cache_set()` in line 152/154.

security.php
```diff
@@ -54,56 +55,148 @@ function wpcom_vip_is_restricted_username( $username ) {
...
-	$ip                    = filter_var( $_SERVER['REMOTE_ADDR'] ?? '', FILTER_VALIDATE_IP, [ 'options' => [ 'default' => '' ] ] );
-	$ip_username_cache_key = $ip . '|' . $username; // IP + username
-	$ip_cache_key          = $ip; // IP only
-	$username_cache_key    = $username; // Username only
+function wpcom_vip_track_auth_attempt( $username, $cache_group ) {
+	$cache_keys = _wpcom_vip_login_cache_keys( $username );
+
+	$base_event_window = apply_filters( 'wpcom_vip_invalid_login_window', MINUTE_IN_SECONDS * 15  );
+	if ( $cache_group === CACHE_GROUP_LOST_PASSWORD_LIMIT ) {
+		$base_event_window = apply_filters( 'wpcom_vip_lost_password_window', MINUTE_IN_SECONDS * 30  );
```
GaryJones: Docs needed for these filters.

security.php
GaryJones: Docs needed for the filter.

security.php
```diff
@@ -54,56 +55,148 @@ function wpcom_vip_is_restricted_username( $username ) {
...
+	} elseif ( $is_restricted_username ) {
+		$ip_username_threshold = 2;
+	}
+
+	if ( $ip_username_count >= $ip_username_threshold || $ip_count >= $ip_threshold || $username_count >= $username_threshold ) {
+		$lock_interval = apply_filters( 'wpcom_vip_invalid_login_lock_period', MINUTE_IN_SECONDS * 30  );
+		if ( $cache_group === CACHE_GROUP_LOGIN_LIMIT ) {
+			do_action( 'login_limit_exceeded', $username );
+		} else {
+			do_action( 'password_reset_limit_exceeded', $username );
```
GaryJones: Docs needed for the actions.

security.php
```diff
@@ -54,56 +55,148 @@ function wpcom_vip_is_restricted_username( $username ) {
...
-	$username_cache_key    = $username; // Username only
+function wpcom_vip_track_auth_attempt( $username, $cache_group ) {
+	$cache_keys = _wpcom_vip_login_cache_keys( $username );
+
+	$base_event_window = apply_filters( 'wpcom_vip_invalid_login_window', MINUTE_IN_SECONDS * 15  );
+	if ( $cache_group === CACHE_GROUP_LOST_PASSWORD_LIMIT ) {
+		$base_event_window = apply_filters( 'wpcom_vip_lost_password_window', MINUTE_IN_SECONDS * 30  );
+	}
 
 	// Longer TTL when logging in as admin, which we don't allow on WP.com
```
GaryJones: Comment needs updating, since no VIP sites are on WP.com any more.

security.php
```diff
@@ -54,56 +55,148 @@ function wpcom_vip_is_restricted_username( $username ) {
...
 
-	if ( $is_restricted_username ) {
-		$cache_expiry = HOUR_IN_SECONDS + $cache_expiry;
-	}
+	wp_cache_add( $cache_keys['ip_username_cache_key'], 0, $cache_group, $ip_username_event_window ); // phpcs:ignore WordPressVIPMinimum.Performance.LowExpiryCacheTime.CacheTimeUndetermined
+	wp_cache_add( $cache_keys['ip_cache_key'], 0, $cache_group, $base_event_window );
+	wp_cache_add( $cache_keys['username_cache_key'], 0, $cache_group, $base_event_window );
+	wp_cache_incr( $cache_keys['ip_username_cache_key'], 1, $cache_group );
+	wp_cache_incr( $cache_keys['ip_cache_key'], 1, $cache_group );
+	wp_cache_incr( $cache_keys['username_cache_key'], 1, $cache_group );
```
GaryJones: Prefix constant needed here on cache key names?

tests/test-security.php
```diff
@@ -118,6 +120,71 @@ public function test__lost_password_limit() {
 
 	}
 
+	public function test__wpcom_vip_track_auth_attempt__defaults() {
```
GaryJones, line 141: Tip: Use `--testdox` when running tests and you'll see how a well-written test method can help better describe what it is that is actually being tested.

Adding `@covers` [annotations](https://docs.phpunit.de/en/9.6/annotations.html#covers) (needed for strict code coverage) also means the inclusion of the function under test is not needed in the test method name:

```suggestion
	/**
	 * @covers ::wpcom_vip_track_auth_attempt
	 */
	public function test_auth_attempt_tracked() {
```

tests/test-security.php
```diff
@@ -118,6 +120,71 @@ public function test__lost_password_limit() {
...
+
+		$this->assertEquals( true, is_wp_error( $result ) );
+	}
+
+	private function _clean_event_window_cache() {
+		wp_cache_delete( $this->test_username, CACHE_GROUP_LOGIN_LIMIT );
+		wp_cache_delete( $this->test_ip, CACHE_GROUP_LOGIN_LIMIT );
+		wp_cache_delete( $this->test_ip . '|' . $this->test_username, CACHE_GROUP_LOGIN_LIMIT );
+	}
+
```
GaryJones, line 210: Move the setUp and tearDown methods (below this line) to the top of the test class - they add important context to how each test method runs.

security.php
GaryJones: It's also required per WPCS as well.

tests/test-security.php
```diff
@@ -118,6 +120,71 @@ public function test__lost_password_limit() {
 
 	}
 
+	public function test__wpcom_vip_track_auth_attempt__defaults() {
+		wpcom_vip_track_auth_attempt( $this->test_username, CACHE_GROUP_LOGIN_LIMIT );
+
+		$username_count = wp_cache_get( $this->test_username, CACHE_GROUP_LOGIN_LIMIT );
+		$this->assertEquals( 1, $username_count );
```
GaryJones: These "check the count matches" assertions could be done with `assertSame()` so that the type and value are matched, not just an equivalent value.

security.php
pschoffer: I guess it would. But it would break the convention of the existing count keys that don't have `vip_` prefix. Please see description which I now filled for details.

security.php
pschoffer: Is there a way to make the function unavailable in php outside this file? cc @sjinks 

I am fine with dropping the `wpcom_`

security.php
pschoffer: Yep it would:
```
<?php
// Enter your code here, enjoy!
$ip = "";
$username = "Batman";
echo( $ip . '|' . $username );
```
Returns `|Batman`

security.php
pschoffer: Yep

security.php
pschoffer: There are two sets of cache keys. One is prefixed with `CACHE_KEY_LOCK_PREFIX` and then there are the "count" ones that are without the prefix. Please check the description of the PR that I now filled in.

security.php
pschoffer: Oooh does that mean we can drop that whole extended fail time for these usernames? I would rather not test it. At least not in the same PR. 

But if you know more about why this exception exists there I would be happy with extending the comment.

security.php
pschoffer: There are two sets of cache keys. One is prefixed with CACHE_KEY_LOCK_PREFIX and then there are the "count" ones that are without the prefix. Please check the description of the PR that I now filled in.

tests/test-security.php
pschoffer: Removing the name form the test case would make it harded for developer to run all tests related that method. What I used to do and our README recommend is to do:

```
./bin/test.sh --filter test__wpcom_vip_username_is_limited
``` 

Is there a way to leverage those annotations instead?

tests/test-security.php
pschoffer: Done, thanks!

tests/test-security.php
pschoffer: Done, thanks!

security.php
```diff
@@ -54,56 +55,148 @@ function wpcom_vip_is_restricted_username( $username ) {
...
+		 * Filters the threshold for limiting password resets by IP + username combination.
+		 *
+		 * Note that changing this value will also change the default value for the username password reset threshold as well,
+		 * which is set as 5 times this value.
+		 *
+		 * @param int    $threshold IP + Username combination password reset threshold. Default 3.
+		 * @param string $ip        IP address of the password reset request.
+		 * @param string $username  Username of the password reset request.
+		 */
+		$ip_username_threshold = apply_filters( 'wpcom_vip_ip_username_password_reset_threshold', 3, $ip, $username );
```
rebeccahum: Do we want to make it configurable for everyone? Uncertain if requirements want certain groups to not be able to configure the thresholds. Also, should we enforce minimum thresholds?

security.php
GaryJones: > Is there a way to make the function unavailable in php outside this file?

No - it would have to be a non-public method in a class for it to be inaccessible / privately scoped.

security.php
GaryJones: Is an empty cache key legit? Is that checked for somewhere?

tests/test-security.php
GaryJones: https://docs.phpunit.de/en/9.6/annotations.html#group is probably what you could use.

Not sure I knew about using test method prefixes to run a group of tests though! I know you can also specify a file path as well.

security.php
pschoffer: I don't know. But seeing as it worked until this point (at least [years](https://github.com/Automattic/vip-go-mu-plugins/blame/develop/security.php#L70)) I think we are OK to keep it as is. Or do you think changing it is worth the increased risk of breaking things? If yes what do you suggest?

GaryJones: No need to change in this PR, but would be worth investigating deeper so that you do know, and the code can be made more obvious / tests added around this.

security.php
GaryJones: What's the downside of switching? A period of up to 30 minutes where the count may be off (people get a few extra tries to log in when the new key name gets deployed), and a need to delete the old key names on a request (code can be removed in the next deploy).

(The PHP constant should also ideally be namespace / prefixed to reduce the risk of clashing with another plugin.)

security.php
GaryJones: Seems to work fine setting and getting from the cache:

Screenshot 2023-06-13 at 11 45 51

security.php
GaryJones: Screenshot 2023-06-13 at 11 47 18

It seems that an empty cache key can't be used, so this may be "working" in the surrounding logic by accident instead of on purpose.

security.php
GaryJones: We don't create sites with `admin` anyway, but the [function](https://github.com/Automattic/vip-go-mu-plugins/blob/b3ac61e7e73c7b33b24a82b7c2a3ab06715d6995/security.php#L51) also covers a couple of bot credentials too, so the exception is worth keeping in - you just need to update the comment to remove the reference to "WP.com".

security.php
pschoffer: Good, thanks for checking.

security.php
pschoffer: > But it would break the convention of the existing count keys that don't have vip_ prefix

So we would have `127.0.0.1|batman` and `vip_locked_127.0.0.1|batman` cache keys. One with a prefix and the other without.

However, I now prefixed the cache groups. Is that covering your worry of the naming clash?

security.php
pschoffer: I changed the comment

security.php
```diff
@@ -54,56 +57,196 @@ function wpcom_vip_is_restricted_username( $username ) {
 		|| WPCOM_VIP_MACHINE_USER_EMAIL === $username;
 }
 
+/**
+ * PRIVATE: This function is not intended to be used outside of this file and is subject to change without notice.
+ *
+ * Returns the defualts for limits.
+ */
+function _vip_get_auth_count_limit_defaults() {
```
pschoffer, line 65: @rinatkhaziev this solution combines both the default value, which we could later sourced in a different manner as well as a set of filters for code adjustment.

Even if we decide not to implement different sourcing, it's still handy to have these values here in one place instead of spread out over the code below.

What do you think?

security.php
pschoffer: I have no strong opinion on this. @rinatkhaziev ?

tests/test-security.php
```diff
@@ -118,17 +138,129 @@ public function test__lost_password_limit() {
...
+		wp_cache_set( $this->test_ip, $original_count, CACHE_GROUP_LOGIN_LIMIT );
+		wp_cache_set( $this->test_ip . '|' . $this->test_username, $original_count, CACHE_GROUP_LOGIN_LIMIT );
 
-		parent::tearDown();
+		wpcom_vip_login_limiter_on_success( $this->test_username );
+
+		$username_count = wp_cache_get( $this->test_username, CACHE_GROUP_LOGIN_LIMIT );
+		$this->assertSame( $original_count, $username_count ); // username is NOT decreased
+
+		$ip_count = wp_cache_get( $this->test_ip, CACHE_GROUP_LOGIN_LIMIT );
```
pschoffer, line 166: It seems these cache assertions are a bit flaky. Is there any other way I could do this?

security.php
```diff
@@ -54,56 +57,202 @@ function wpcom_vip_is_restricted_username( $username ) {
 		|| WPCOM_VIP_MACHINE_USER_EMAIL === $username;
 }
 
+/**
+ * PRIVATE: This function is not intended to be used outside of this file and is subject to change without notice.
+ *
+ * Returns the defualts for limits.
```
rinatkhaziev:
```suggestion
 * Returns the defaults for limits.
```

security.php
GaryJones: Yes ✅

security.php
GaryJones:
```suggestion
 * Return the defaults for login and password reset authorization limits, windows, and lockout periods.
 *
 * @internal The values are subject to change without notice.
```

ref: https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/internal.html#internal

security.php
```diff
@@ -54,56 +57,202 @@ function wpcom_vip_is_restricted_username( $username ) {
...
+	];
+}
+
+/**
+ * PRIVATE: This function is not intended to be used outside of this file and is subject to change without notice.
+ *
+ * Returns the cache keys used for login and forgotten passoword rate limiting.
+ *
+ * @param string $raw_username The username to use for the cache keys.
+ * @return array Associative array of cache keys.
```
GaryJones:
```suggestion
 * Return the cache keys used for login and forgotten password rate limiting.
 *
 * @internal
 *
 * @param string $raw_username The username to use for the cache keys.
 * @return array Associative array of cache keys.
```

Any particular reason this needs to be considered private/internal?

security.php
```diff
@@ -54,56 +57,202 @@ function wpcom_vip_is_restricted_username( $username ) {
...
+}
+
+/**
+ * PRIVATE: This function is not intended to be used outside of this file and is subject to change without notice.
+ *
+ * Checks if counts for the given username and IP address are above the given thresholds.
+ * If so, the account is locked for a period of time.
+ *
+ * @param string $username The username to check.
+ * @param string $cache_group The cache group to use for the rate limiting.
```
GaryJones:
```suggestion
 * Check if counts for the given username and IP address are above the given thresholds.
 * If so, the account is locked for a period of time.
 *
 * @internal
 *
 * @param string $username The username to check.
 * @param string $cache_group The cache group to use for the rate limiting.
```

Why would this be considered private/internal?

tests/test-security.php
GaryJones: Flaky how? 

Does the tests architecture mock the object cache (with, an in-memory cache, say) or does it just default to the wp_options table, since we don't want to be testing the actual functionality of wp_cache_set() and wp_cache_get() themselves, only that they've been called with the right arguments. This is where BrainMonkey could also help.

GaryJones: If it's flaky per PHP / PHPUnit version, then [Yoast's WPTestUtils](https://github.com/Yoast/wp-test-utils) (or at least [PHPUnit-Polyfills](https://packagist.org/packages/yoast/phpunit-polyfills)) can help, and that requires using a slightly different version of `tearDown()` method name that works across different PHPUnit versions.

tests/test-security.php
pschoffer: It seems testsdox is happy even without the comments:

```
./bin/test.sh --filter test__wpcom_vip_username_is_limited --testdox
VIP_Go_Security_
 ✔ wpcom vip username is limited  should not limit by default
 ✔ wpcom vip username is limited  should be limit after few tries
 ✔ wpcom vip username is limited  should be limit even after the event window
 ✔ wpcom vip username is limited  should be limit after 3 attempts fedramp
 ```

File: security.php (hunk 10 above)
Comment (pschoffer) [reply security.php]: I added a comment there.

File: security.php (hunk 19 above)
Comment (pschoffer) [reply security.php]: Fixed thanks

File: tests/test-security.php (hunk 18 above)
Comment (pschoffer) [reply tests/test-security.php:166]: Oh I tracked it down. It was due to the sanitization of usernames somehow working differently on multisite (lowercasing text).

File: tests/test-security.php (hunk 13 above)
Comment (GaryJones) [reply tests/test-security.php:141]: Right, but that could be arranged as something like:

```php
/**
 * @group login_limits
 * /
function test_username_should_not_be_limited_by_default() {...}
```
...which would then give an output like:
```sh
$ ./bin/test.sh --group login_limits --testdox
VIP_Go_Security
 ✔ Username should not be limited by default
 ✔ Username should be limited after few tries
 ✔ Username should be limited even after the event window
 ✔ Username should be limited after 3 attempts for Fedramp
```

You could see how the test method names can match up to acceptance criteria.

Not a blocker, and low priority, but [nicer tests output](https://peakd.com/hive-168588/@crell/advanced-phpunit-shenanigans).

rinatkhaziev (approved):
Thanks @pschoffer

PR #4589, opened 2023-06-13
A8C_Files: Move logic out in the open

rebeccahum:
## Description
We currently have this logic applied elsewhere. It would be good to just move it out in the open.

There should be no functionality change and no sites are using this constant actively.

### DIFF
--- a/a8c-files.php
+++ b/a8c-files.php
@@ -93,7 +93,7 @@ class A8C_Files {
 		add_filter( 'option_upload_url_path', array( $this, 'upload_url_path' ) );
 
 		// Conditionally schedule the attachment filesize metadata update job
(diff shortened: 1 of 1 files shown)


pschoffer:
> There should be no functionality change and no sites are using this constant actively.

I think this change will make prod sites to schedule that update cron job, while before no site had the job unless they had that specific ENV var, which they didn't. correct?

From the title of this PR I think this is what we are after, but I am just making sure.

Also no Changelog description.

pschoffer (approved):
VIP_FILESYSTEM_SCHEDULE_FILESIZE_UPDATE - was set in prior logic.

This looks great now

PR #4595, opened 2023-06-14
Streamwrapper: Move logic out into the open

rebeccahum:
## Description
Similar to #4589, we should move this logic out into the open since it is already enabled for everyone. There is still the option to disable it with the constant `VIP_FILESYSTEM_USE_STREAM_WRAPPER` though.

### DIFF
--- a/a8c-files.php
+++ b/a8c-files.php
@@ -78,7 +78,7 @@ class A8C_Files {
 			return GB_IN_BYTES * 3; 
 		} );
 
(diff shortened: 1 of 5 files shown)


rinatkhaziev:
This flips the logic and may result in breakage in Local Dev Env which didn't have the stream wrapper constant. This is being addressed here: https://github.com/Automattic/vip-container-images/pull/517

But even when that's merged we don't update wp-config for existing environments.

PR #4617, opened 2023-06-20
Add auth tokens through VIP_SERVICES_AUTH_TOKENS instead, if available

abdullah-kasim:
## Description

Updates where we're getting the token from.

### Site Details Service

We added an alternative endpoint for sending data to the Site Details Service

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

1. Get a local dev-env running with local mu-plugins
2. Get local xdebug running with `vip dev-env shell --root --slug=yourlocalslug`, then type in `phpenmod xdebug`, then `exit` and then run `vip dev-env shell`.
3. Run `XDEBUG_TRIGGER=xdebug PHP_IDE_CONFIG="serverName=server-name-in-ide" wp shell`. Alternatively, just run `wp shell` if you don't want to use xdebug.
4. `var_dump()` away/breakpoint away around the methods for this change.
5. Then, in `wp shell`, run `require_once '/wp/wp-content/mu-plugins/config/class-site-details-index.php'; $clz = new Automattic\VIP\Config\Site_Details_Index();`
6. Run `$encoded = base64_encode(json_encode(['site' => ['vip-site-details' => ['token' => 'someToken', 'url' => 'http://localhost:4002/' ]]])); define('VIP_SERVICES_AUTH_TOKENS', $encoded)` so that we set VIP_SERVICES_AUTH_TOKENS for us to test with.
7. Finally, run `$clz->put_site_details()`.

### DIFF
--- a/config/class-site-details-index.php
+++ b/config/class-site-details-index.php
@@ -348,18 +348,21 @@ class Site_Details_Index {
 	 * and sends it to the site details service
 	 */
 	public function put_site_details() {
-		$site_details       = $this->get_site_details();
-		$url                = null;
-		$token              = null;
-		$auth_token_details = null;
+		$site_details      = $this->get_site_details();
+		$url               = null;
+		$token             = null;
+		$new_service_url   = null;
+		$new_service_token = null;
 
 		if ( defined( 'VIP_SERVICES_AUTH_TOKENS' ) && ! empty( VIP_SERVICES_AUTH_TOKENS ) ) {
 			$auth_token_details = json_decode( base64_decode( VIP_SERVICES_AUTH_TOKENS ), true );
+			$new_service_url    = $auth_token_details['site']['vip-site-details']['url'] ?? null;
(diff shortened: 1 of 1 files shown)


aagam-shah (approved):

saroshaga (approved):

abdullah-kasim:
The manual test will have to make do. There's no existing test though I do agree with making one.

However, the test is a tremendous effort as it requires a refactor before the code becomes testable as I'm not sure if we have a good way to override defined variables and so on and I haven't opened any discussions on importing such libraries.

Might just be me not knowing about something though because thinking back, we could always use a different `wp-config.php` for different test scenarios.

sjinks:
@abdullah-kasim we actually have a class to mock `define()`/`defined()`/`constant()` calls (see https://github.com/Automattic/vip-go-mu-plugins/blob/develop/tests/mock-constants.php).

To fake the HTTP response, you can use the `pre_http_request` filter (there is an example in `Site_Details_Index_Test`).

PR #4628, opened 2023-06-20
BB8-10403: Use updated sha256 for Gravatar

rebeccahum:
## Description
Now that Gravatar supports SHA256, let's update our references to it.


sjinks:
You are fast, I was about to take it :-)

sjinks (approved):

PR #4640, opened 2023-06-23
VIP Request Block: remove fastcgi_finish_request and don't block in CLI context

rinatkhaziev:
## Description

1. fastcgi_finish_request is not needed here because we bail via `exit` immediately after.
2. Killing off requests running in WP-CLI due to faulty logical expression (i.e. an unintended match) is not desirable either, so rather than killing the request we bail early. 

### Library updated: VIP Request Block

We've updated VIP Request Block library not run in WP-CLI context as that potentially may lead to inability to run CLI commands in case an expression evaluated to `true` in CLI context.

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

1. Check out the PR, add something like 
```php
VIP_Request_Block::block_and_log( 'test', 'test' );
```

2. Open the browser verify that request is terminated
3. Spawn a CLI command and verify it works.

### DIFF
--- a/lib/class-vip-request-block.php
+++ b/lib/class-vip-request-block.php
@@ -133,7 +133,10 @@ class VIP_Request_Block {
 			header( 'Expires: Wed, 11 Jan 1984 05:00:00 GMT' );
 			header( 'Cache-Control: no-cache, must-revalidate, max-age=0' );
 
-			fastcgi_finish_request();
+			if ( function_exists( 'fastcgi_finish_request' ) ) {
+				fastcgi_finish_request();
+			}
+
 			// phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
 			error_log( sprintf( 'VIP Request Block: request was blocked based on "%s" with value of "%s"', $criteria, $value ) );
 			exit;
--- a/lib/class-vip-request-block.php
+++ b/lib/class-vip-request-block.php
@@ -128,14 +128,16 @@ class VIP_Request_Block {
 			newrelic_ignore_transaction();
 		}
 
+		if ( defined( 'WP_CLI' ) && constant( 'WP_CLI' ) ) {
+			// phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
+			error_log( sprintf( 'VIP Request Block: would block based on "%s" with value of "%s" but it\'s a CLI request', $criteria, $value ) );
+			return true;
+		}
+
 		if ( ! defined( 'WP_TESTS_DOMAIN' ) ) {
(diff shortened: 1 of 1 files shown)


sjinks (approved):
Does it make sense to invoke `fastcgi_finish_request()` at all? We are aborting the request almost immediately.

lib/class-vip-request-block.php
```diff
@@ -128,15 +128,17 @@ public static function block_and_log( string $value, string $criteria ) {
 			newrelic_ignore_transaction();
 		}
 
+		if ( defined( 'WP_CLI' ) && constant( 'WP_CLI' ) ) {
+			// phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
+			error_log( sprintf( 'VIP Request Block: would block based on "%s" with value of "%s" but it\'s a CLI request', $criteria, $value ) );
```
mjangda: The error log seems unnecessary; doesn't really add much.

rinatkhaziev: @mjangda why not? my reasoning was that if someone does this unintentionally they wouldn't know unless they'd look at their logs, at which point they might correct their mistake.

mjangda: If we're in a context where the request block stuff shouldn't run, we should just bail early and do nothing. That's better than asking customers to remember to check the context when implementing:

```
// Annoying
if ( ! defined( 'WP_CLI' ) && ! defined( 'DOING_CRON' ) ) {
    VIP_Request_Block::block( ... );
}

// Much better
VIP_Request_Block::block( ... );
```

lib/class-vip-request-block.php
```diff
@@ -128,12 +128,17 @@ public static function block_and_log( string $value, string $criteria ) {
 			newrelic_ignore_transaction();
 		}
 
+		if ( defined( 'WP_CLI' ) && constant( 'WP_CLI' ) ) {
```
mjangda, line 127: https://make.wordpress.org/cli/handbook/guides/commands-cookbook/#include-in-a-plugin-or-theme

```suggestion
		if ( defined( 'WP_CLI' ) && WP_CLI ) {
```

rinatkhaziev: we've been using `constant` for quite a while because it's better testable that way.

mjangda: Ah, TIL. Thanks! Should we suggest an update to WP-CLI docs?

rinatkhaziev:
@mjangda @sjinks I addressed y'alls feedback, can you take another look please

mjangda (approved):

PR #4660, opened 2023-07-05
Adds Parsely as VIP Integration

mehmoodak:
## Description

- Adds parsely in the list of supported VIP integrations and integrate it using `ParselyIntegration`  class.
- In `Integration` class adds the functionality of activating integration based on the config file provided by VIP.

## Changelog Description

Adds Parse.ly as VIP Integration

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

1. `git checkout vip-parsely-41/adds-wp-parsely-vip-integration`
1. Create integrations config file inside `php` container and place configs (we have to do it manually because as of now `vip dev-env` doesn't have support to auto create this folder/file while creating new envs)
  - `touch /wp/config/integrations-config/parsely-config.php` and adds following content
       - `parsely` can be replaced with any other slug of the integration
       - You may have to restart vip dev-env for any change in file.
  
  ```
<?php

return array(
	'env' => array(
		'status' => 'disabled',
		'config' => array (
			'site_id' => 'site.com',
			'api_secret' => 'site_api_secret',
			'metadata_secret' => 'site_metadata_secret',
		)
	),
	'network_sites' => array (
		'1' => array (
			'status' => 'disabled',
			'config' => array (
				'site_id' => 'site1.com',
				'api_secret' => 'site_api_secret_1',
				'metadata_secret' => 'site_metadata_secret_1',
			)
		),
		'2' => array (
			'status' => 'enabled',
			'config' => array (
				'site_id' => 'site2.com',
				'api_secret' => 'site_api_secret_2',
				'metadata_secret' => 'site_metadata_secret_2',
			)
		),
	)
);
  ```

4. Verify integration enablement with different combinations of vip configs which in our case means `wp-parsely` plugin is activated with the SiteID and secrets which are provided by VIP (credentails support will be available on `v3.9` and above).
4. Apart from enabling the plugin via VIP config customer can also enables it by placing `\Automattic\VIP\Integrations\activate( 'parsely' );` in `plugin-loader.php` file so better to also test this flow.


integrations/block-data-api.php
```diff
@@ -1,4 +1,9 @@
 <?php
+/**
+ * Integration: Block Data API.
+ *
+ * @package Automattic\VIP\Integrations
+ */
```
mehmoodak, line 6: `phpcs` is showing a lot of warnings through out all the files so address those in this PR (mostly comments).

integrations/block-data-api.php
```diff
@@ -7,35 +12,33 @@
  *
  * @private
  */
-class BlockDataApi extends Integration {
+class BlockDataApiIntegration extends Integration {
```
mehmoodak, line 15: Added `Integration` in name which I think is meaningful.

integrations/block-data-api.php
```diff
@@ -7,35 +12,33 @@
...
 
 	/**
 	 * The version of the Block Data API plugin to load, that's set to the latest version.
-	 * This should be higher than the lowestVersion set in https://github.com/Automattic/vip-go-mu-plugins-ext/blob/trunk/config.json#L63
-	 * 
+	 * This should be higher than the lowestVersion set in "vip-block-data-api" config (https://github.com/Automattic/vip-go-mu-plugins-ext/blob/trunk/config.json)
```
mehmoodak, line 19: Remove `#L63` because that can change in future and instead updated description for better understanding.

integrations/integrations.php
```diff
@@ -20,48 +25,35 @@ class Integrations {
...
 	 * Registers an integration.
 	 *
-	 * @param string             $slug            A unique identifier for the integration.
-	 * @param string|Integration $class_or_object Fully-qualified class or instantiated Integration object.
+	 * @param Integration $integration Instantiated Integration object.
 	 *
-	 * @private
+	 * @throws InvalidArgumentException Excpetion if invalid argument are passed.
 	 */
-	public function register( string $slug, $class_or_object ): void {
```
mehmoodak, line 28: - Added `slug` property in `Integration` class and removed it from param (because we can get it from obj)
- For simplicity just supporting instantiated obj in param instead of supporting both obj and class name

integrations/parsely.php
```diff
@@ -0,0 +1,51 @@
...
+class ParselyIntegration extends Integration {
+
+	/**
+	 * The version of the parsely plugin to load.
+	 *
+	 * This should be the latest version available in `parsely` config https://github.com/Automattic/vip-go-mu-plugins-ext/blob/trunk/config.json
+	 *
+	 * @var string
+	 */
+	private string $version = '3.8'; // If updated then make sure to update SUPPORTED_VERSIONS in /wp-parsely.php (legacy support).
```
mehmoodak: I tried using `SUPPORTED_VERSIONS[0]` to limit changes on upgrade but didn't succeed. If you have any idea how its possible then please let me know. Thanks

mehmoodak (commented):
Added comments for easier review.

integrations/integration.php
```diff
@@ -1,57 +1,261 @@
...
-		return $this->is_active;
+		if ( $this->is_active_by_customer ) {
+			return true;
+		}
+
+		if ( $this->is_active_by_vip() ) {
+			return true;
+		}
+
+		return false;
```
ingeniumed: Would simplifying this down to

```suggestion
		return ( $this->is_active_by_customer ) || ( $this->is_active_by_vip() );
```

work?

integrations/integrations.php
```diff
@@ -77,13 +73,15 @@ public function load_active(): void {
 	 * @param string $slug   A unique identifier for the integration.
 	 * @param array  $config An associative array of configuration values for the integration.
 	 *
+	 * @throws InvalidArgumentException Excpetion if invalid argument are passed.
```
ingeniumed:
```suggestion
	 * @throws InvalidArgumentException Exception if invalid argument are passed.
```

integrations/integrations.php
```diff
@@ -20,41 +25,32 @@ class Integrations {
 	/**
 	 * Registers an integration.
 	 *
-	 * @param string             $slug            A unique identifier for the integration.
-	 * @param string|Integration $class_or_object Fully-qualified class or instantiated Integration object.
+	 * @param Integration $integration Instantiated integration object.
+	 *
+	 * @throws InvalidArgumentException Excpetion if invalid argument are passed.
```
ingeniumed:
```suggestion
	 * @throws InvalidArgumentException Exception if invalid argument are passed.
```

ingeniumed (commented):
Your comments are wonderful, thanks for adding them in. I really like the renaming changes you've done, as well as the helper methods that you have added in. Awesome work 💯

integrations/integration.php
mehmoodak: In my recent PR's mo always suggested to not have conditions in same line to avoid complexity and makes code easier to read which I somewhat agrees (not making this change)

integrations/integrations.php
```diff
@@ -77,13 +73,15 @@ public function load_active(): void {
 	 * @param string $slug   A unique identifier for the integration.
 	 * @param array  $config An associative array of configuration values for the integration.
 	 *
+	 * @throws InvalidArgumentException Exception if invalid argument are passed.
+	 *
 	 * @private
 	 */
-	public function activate( string $integration_slug, array $config = [] ): void {
-		$integration = $this->get( $integration_slug );
+	public function activate( string $slug, array $config = [] ): void {
```
mehmoodak: Matching it with Doc comment.

tests/bootstrap.php
```diff
@@ -174,17 +174,22 @@ function _configure_specify_wp_parsely_version() {
 
 tests_add_filter( 'muplugins_loaded', function () {
 	echo "[WP_PARSELY_INTEGRATION] Removing autoload (so we can manually test)\n";
-	$removed = remove_action( 'plugins_loaded', 'Automattic\VIP\WP_Parsely_Integration\maybe_load_plugin', 1 );
-	if ( ! $removed ) {
-		throw new Exception( '[WP_PARSELY_INTEGRATION] Failed to remove autoload' );
+
+	if ( has_action( 'plugins_loaded', 'Automattic\VIP\WP_Parsely_Integration\maybe_load_plugin' ) ) {
```
mehmoodak, line 178: Added `has_action` check before removal to avoid the error during some test.

tests/parsely/test-mu-parsely-integration.php
```diff
@@ -6,10 +6,8 @@
 use Parsely\UI\Row_Actions;
 use WP_UnitTestCase;
 
-function test_mode() {
-	$mode = getenv( 'WPVIP_PARSELY_INTEGRATION_TEST_MODE' );
-	return $mode ?: 'disabled';
-}
```
mehmoodak, line 12: Created a common `utils` folder and add common functions there which we are using more than once and across multiple tests.

tests/vip-integrations/fake-integration.php
mehmoodak, line 1: Moved files from `tests/vip-integrations` to `tests/integrations` to match the file structure with source code.

mehmoodak (commented):
Added comments for easier review.

integrations/integration.php
```diff
@@ -1,55 +1,260 @@
...
+	 */
+	private string $slug;
+
 	/**
 	 * An optional configuration array for this integration, added during activation.
 	 *
 	 * @var array
 	 */
-	protected array $config = [];
+	private array $customer_config = [];
```
ingeniumed: Is there a way we could combine this, and `vip_config` into one array? That way, each integration implementation has only a single config to consume and work on. In the case where the integration is implemented via `mu-plugins`, the config would be the same as what $customer_config would be. If it's via the other route, then it would realistically just be vip_config and not customer_config as well.

ingeniumed (approved):
Just got the one comment, otherwise it looks good to me.

integrations/integration.php
mehmoodak: This is what my understanding is:

- `customer_config` are the configs provided by customer for plugin configuration and it can include things like versioning and custom plugin settings.

- `vip_config` are the configs which are provided by VIP and it can include things like status/secrets of the integration (for client, site and network sites) and in future may include custom plugin settings. It have a relatively complex structure than `customer_config`

So I think for now it is a better decision to separate them because of different attributes and different structures of these configs. As of now from code we can easily make the decision to handle `customer_config` easily and handle `vip_config` with care because it contains sensitive info but if we mix these then I think it can introduce complexity which will be difficult to understand and maintain.

integrations/integration.php
```diff
@@ -1,55 +1,260 @@
...
+		if ( Site_Integration_Status::BLOCKED === $site_status ) {
+			return false;
+		}
+
+		// If enabled on network site then set credentials via filter and return true.
+		if ( is_multisite() && $this->get_value_from_vip_config( 'network_sites', 'status' ) === Site_Integration_Status::ENABLED ) {
+			$have_config = $this->get_value_from_vip_config( 'network_sites', 'config' ) !== '';
+
+			if ( '' !== $this->vip_config_filter_name && $have_config ) {
+				add_filter( $this->vip_config_filter_name, function() {
```
mjangda: Allowing for variable filter names makes things really hard to follow and debug. Where possible, it's better to use an explicitly defined and generic filter name.

In this case, I'm not sure I see the benefit to this either. Why not just have each integration handle the config overrides their own way?

WPprodigy: +1

mehmoodak: > Why not just have each integration handle the config overrides their own way?

@mjangda Can you please explain this a bit more?

mehmoodak: Thanks for nice feedback, things look a bit cleaner now

Implemented this feedback in https://github.com/Automattic/vip-go-mu-plugins/pull/4660/commits/53733f9a6ad8f29f8ca552b12731fc03ab13878a, please do let me know if I got it wrong or it needs some more work.

integrations/integrations.php
```diff
@@ -77,13 +73,15 @@ public function load_active(): void {
...
+		$integration = $this->get( $slug );
 
 		if ( null === $integration ) {
-			throw new InvalidArgumentException( sprintf( 'VIP Integration with slug "%s" is not a registered integration.', $integration_slug ) );
+			throw new InvalidArgumentException( sprintf( 'VIP Integration with slug "%s" is not a registered integration.', $integration ) );
```
mjangda: `$integration` will be `null`.

```suggestion
			throw new InvalidArgumentException( sprintf( 'VIP Integration with slug "%s" is not a registered integration.', $slug ) );
```

integrations/parsely.php
```diff
@@ -0,0 +1,50 @@
...
+	 * @return void
+	 */
+	public function setup_config(): void {
+		add_filter( 'wp_parsely_credentials', function() {
+			if ( is_multisite() ) {
+				return $this->get_vip_config_of_current_network_site();
+			}
+
+			return $this->get_vip_config_of_site();
+		} );
```
mjangda: Do we want to validate that `get_vip_config_of_current_network_site` / `get_vip_config_of_site` return the expected values? Or should we just leave that to the parsely plugin?

One use case we need to figure out: I have the plugin active but not configured (i.e. at the platform-level, we have not provisioned the keys yet so the config does not have site ID / secret). We may still need a way to force off the inputs on the settings page. Currently, this will pass empty values to `wp_parsely_credentials` which will mean that the plugin is not put into "managed" mode and P100 users would be able to edit the site ID and secret themselves.

integrations/parsely.php
```diff
@@ -0,0 +1,50 @@
...
+		// Do not load plugin if already loaded by customer code.
+		if ( class_exists( 'Parsely\Parsely' ) ) {
+			return;
+		}
+
+		// Activating the integration via setting constant whose implementation is already
+		// handled via Automattic\VIP\WP_Parsely_Integration (ideally we should move
+		// all the implementation here such that there will be only one way of managing
+		// the plugin).
+		define( 'VIP_PARSELY_ENABLED', true );
```
mjangda, line 31: Won't this cause our activation stats to be skewed because everything activated through this integrations framework will show up as `ENABLED_CONSTANT`?

integrations/integration.php
```diff
@@ -1,66 +1,298 @@
...
+/**
+ * Enum which represent all possible statuses for the site integration via VIP.
+ *
+ * These should be in sync with the statuses available on the backend.
+ */
+abstract class Site_Integration_Status {
+	const ENABLED  = 'enabled';
+	const DISABLED = 'disabled';
+	const BLOCKED  = 'blocked';
+}
```
mjangda: Should we just simplify these two enums as `Integration_Status`?

integrations/integration.php
```diff
@@ -1,66 +1,298 @@
...
+		if ( is_multisite() ) {
+			$network_site_status = $this->get_value_from_vip_config( 'network_sites', 'status' );
+			
+			// Setup config and return true If enabled on network site.
+			if ( Site_Integration_Status::ENABLED === $network_site_status ) {
+				$have_config = $this->get_vip_config_of_current_network_site() !== '';
+
+				if ( $have_config ) {
+					$this->setup_config();
+				}
```
mjangda: This is a side effect inside a boolean method, which is not ideal (can lead to bugs and unexpected issues).

It would be better if this was triggered separately.

integrations/integration.php
```diff
@@ -1,66 +1,298 @@
...
 	 * @private
 	 */
 	public function is_active(): bool {
-		return $this->is_active;
+		if ( $this->is_active_by_customer ) {
+			return true;
+		}
+
+		return $this->is_active_by_vip();
+	}
```
mjangda: I think we should change the terminology here:

- `by_customer ` => `via_code`
- `by_vip`  => `via_config`

integrations/integration.php
mjangda: In the original Integrations PR, we'd talked about potentially keeping things simple and combining both activation paths, which I think we should do.

See:
- https://github.com/Automattic/vip-go-mu-plugins/pull/4458#discussion_r1220373639
- https://github.com/Automattic/vip-go-mu-plugins/pull/4458#issuecomment-1544102329

Here's one way we could do this:

```
diff --git a/integrations/integrations.php b/integrations/integrations.php
index 35bc43a6a..5712486f5 100644
--- a/integrations/integrations.php
+++ b/integrations/integrations.php
@@ -54,6 +54,15 @@ class Integrations {
 		return $this->integrations[ $slug ] ?? null;
 	}
 
+	/**
+	 * Returns all registered integrations.
+	 *
+	 * @return array<Integration>
+	 */
+	public function get_registered(): array {
+		return $this->integrations;
+	}
+
 	/**
 	 * Call load() for each registered and activated integration.
 	 *
diff --git a/vip-integrations.php b/vip-integrations.php
index 0728a6326..6e9ff5127 100644
--- a/vip-integrations.php
+++ b/vip-integrations.php
@@ -34,6 +34,17 @@ function activate( string $slug, array $config = [] ): void {
 // Load integrations in muplugins_loaded:5 to allow integrations to hook
 // muplugins_loaded:10 or any later action.
 add_action( 'muplugins_loaded', function() {
+	foreach ( IntegrationsSingleton::instance()->get_registered() as $slug => $integration ) {
+		if ( $integration->is_active_via_config_file() ) {
+			$config = $integration->get_config_from_config_file();
+
+			activate( $slug, array(
+				'loaded_via' => 'platform-config',
+				'config' => $config,
+			) );
+		}
+	}
+
 	IntegrationsSingleton::instance()->load_active();
 }, 5 );
```

The nice thing is that we have a consistent pathway for activation. I think there's some limits here in that customers calling `activate` directly can't as easily set configs for network site, etc. but we could make that an explicit limitation.

integrations/parsely.php
mehmoodak: As far as I know this is what we wanted? i.e. we introduced this constant at the start of P100 project.

integrations/integration.php
mehmoodak: I think that will cause confusion while reading code because client only have support for `BLOCKED` status or nothing. The current implementation of enums aligns with that.

integrations/integration.php
mehmoodak: Apologies, didn't get it can you please explain a bit more :)

integrations/parsely.php
mehmoodak: > Do we want to validate that get_vip_config_of_current_network_site / get_vip_config_of_site return the expected values? Or should we just leave that to the parsely plugin?

By validation you mean? currently we do have tests to make sure that we are passing right config. Other than that I think each integration should take care of the validations and shouldn't trust anything provided by platform .... for now we can keep things simple and add validations in future if necessary.

> Currently, this will pass empty values to wp_parsely_credentials which will mean that the plugin is not put into "managed" mode and P100 users would be able to edit the site ID and secret themselves.

Thanks for pointing it out, I was planning to do this work in separate ticket but now will do it in current PR.

integrations/integration.php
mehmoodak: Currently keys of `$customer_config` param aren't defined yet but most probably it will include `$version` and maybe some more things and `$vip_config` support following params

**Note:** As of now versioning of `wp-parsely` is supported via`wpvip_parsely_version` filter and we are not changing it.

```
{
  "client": {
    "status": ""
  },
  "site": {
    "status": "",
    "config": {}
  },
  "network_sites": {
    "1": {
      "status": "",
      "config": {}
    }
  }
}
```

So can you please suggest how should I extend this? I am thinking of extending the config provided by platform with the customer configs and then forward it it to the integration (i.e. plugin) via `setup_config` method .... in this case customer configs will take priority .... is it okay?

It will be helpful if we can decides what kind of values we could expect from `customer_configs` ..... I think if its generic across all integrations then it does make sense to use one config and one filter but if could going to be different for each integration then maybe its better to keep this separate (personally I prefers to keep it generic across all integrations but let me know about your thoughts)

integrations/integration.php
mjangda: https://en.wikipedia.org/wiki/Side_effect_(computer_science)

Basically, the goal of the function is to give you a true or false value. However, the function does more than that by changing the state of the object (this is the "side effect"), which is a bit unexpected. And because it's unexpected, that can lead to other unexpected behaviours and bugs.

mehmoodak: Got it. Thanks for explaning, will take care of this.

integrations/parsely.php
mjangda: We should still be able to differentiate between the different ways the plugin will be activated like whether it's through platform config or through customer code.

We can leave this for now though and handle it separately when we deprecate the `WP_Parsely_Integration` implementation.

integrations/integration.php
mehmoodak: Addressed in https://github.com/Automattic/vip-go-mu-plugins/pull/4660/commits/0f6f3967c49b9cd9e84a6432566b2a1d81a14417 (tests are updated seprately)

integrations/integration.php
mehmoodak: Addressed in https://github.com/Automattic/vip-go-mu-plugins/pull/4660/commits/0f6f3967c49b9cd9e84a6432566b2a1d81a14417

integrations/parsely.php
mehmoodak: > Currently, this will pass empty values to wp_parsely_credentials which will mean that the plugin is not put into "managed" mode and P100 users would be able to edit the site ID and secret themselves.

On second thought I think plugin should just check the existance of `wp_parsely_credentials` filter to set the plugin in managed mode (ticket: https://github.com/Automattic/vip-parsely/issues/53)

Let me know if there are any more changes needed here.

integrations/block-data-api.php
```diff
@@ -22,15 +27,15 @@ class BlockDataApi extends Integration {
 	 *
 	 * @private
 	 */
-	public function load( array $config ): void {
```
mehmoodak, line 25: We can use `$this->get_config` so no need to pass the configs as param.

integrations/integration.php
```diff
@@ -22,14 +36,29 @@ abstract class Integration {
...
-	 * @param array  $config An associative array of configuration values for the integration.
+	 * @param array $config An associative array of configuration values for the integration.
+	 *
+	 * @throws LogicException Exception if integration is already activated.
 	 *
 	 * @private
 	 */
 	public function activate( array $config = [] ): void {
+		if ( $this->is_active() ) {
+			throw new LogicException( sprintf( 'VIP Integration with slug "%s" is already activated.', $this->get_slug() ) );
```
mjangda: This can result in fatal errors on customer sites which is not ideal. We may want to handle it a bit more gracefully: log warning and skip the double-activation request.

mehmoodak: Addressed in https://github.com/Automattic/vip-go-mu-plugins/pull/4660/commits/266b1587da401cef0ce8d60ae702cf630bdf446c (just skipping silently without error_log, let me know if its compulsory)

integrations/integration-config.php
```diff
@@ -0,0 +1,176 @@
...
+	 * @param string $config_type Type of the config whose data is needed i.e. client, site, network-sites etc.
+	 * @param string $key Key of the config from which we have to extract the data.
+	 *
+	 * @return null|array
+	 *
+	 * @throws InvalidArgumentException Exception if invalid argument is passed.
+	 */
+	protected function get_value_from_config( string $config_type, string $key ) {
+		if ( ! in_array( $config_type, [ 'client', 'site', 'network_sites' ], true ) ) {
+			throw new InvalidArgumentException( 'Config type must be one of client, site and network_sites.' );
```
mjangda: Avoid fatals:

```suggestion
			trigger_error( '$config_type param must be one of client, site and network_sites.', E_USER_WARNING );
			return null;
```

Aside: It may be nice to have a helper function that can throw the exception locally but avoid fatals in production.

integrations/integrations.php
```diff
@@ -13,49 +18,72 @@ class Integrations {
...
+	 *
+	 * @throws InvalidArgumentException Exception if invalid argument are passed.
 	 *
 	 * @private
 	 */
-	public function register( string $slug, $class_or_object ): void {
-		if ( isset( $this->integrations[ $slug ] ) ) {
+	public function register( $integration ): void {
+		if ( ! is_subclass_of( $integration, Integration::class ) ) {
+			throw new InvalidArgumentException( sprintf( 'Integration class "%s" must extend %s.', get_class( $integration ), Integration::class ) );
```
mjangda: Same note about avoiding fatals.

integrations/integrations.php
```diff
@@ -13,49 +18,72 @@ class Integrations {
...
+		}
+
+		$slug = $integration->get_slug();
+
+		if ( null !== $this->get_integration( $slug ) ) {
 			throw new InvalidArgumentException( sprintf( 'Integration with slug "%s" is already registered.', $slug ) );
```
mjangda: Same note about avoiding fatals.

integrations/integrations.php
```diff
@@ -66,26 +94,28 @@ public function get( string $slug ): ?Integration {
...
 	 * @private
 	 */
-	public function activate( string $integration_slug, array $config = [] ): void {
-		$integration = $this->get( $integration_slug );
+	public function activate( string $slug, array $config = [] ): void {
+		$integration = $this->get_integration( $slug );
 
 		if ( null === $integration ) {
-			throw new InvalidArgumentException( sprintf( 'VIP Integration with slug "%s" is not a registered integration.', $integration_slug ) );
+			throw new InvalidArgumentException( sprintf( 'VIP Integration with slug "%s" is not a registered integration.', $slug ) );
```
mjangda: Same note about avoiding fatals.

integrations/integrations.php
mjangda: Much cleaner!

integrations/parsely.php
```diff
@@ -0,0 +1,42 @@
...
+
+		add_filter( 'wp_parsely_credentials', array( $this, 'wp_parsely_credentials_callback' ) );
+	}
+
+	/**
+	 * Callback for `wp_parsely_credentials` filter.
+	 */
+	protected function wp_parsely_credentials_callback(): array {
+		return $this->get_config();
```
mjangda: We should only return the data that this filter needs:

- `site_id`
- `api_secret`
- `is_managed = true`

integrations/integrations.php
```diff
@@ -13,49 +18,72 @@ class Integrations {
...
+		return $this->integrations[ $slug ] ?? null;
+	}
+
+	/**
+	 * Activates integrations based on the configuration provided by VIP
+	 * (only if not already activated by customer).
+	 *
+	 * @return void
+	 */
+	public function activate_integrations_via_vip_config() {
```
mjangda: What do you think about renaming this to `activate_platform_managed_integrations` or even just `activate_managed_integrations`?

integrations/integrations.php
```diff
@@ -13,49 +18,72 @@ class Integrations {
...
 			}
 
-			$integration_object = new $class_or_object();
-		}
+			$vip_config = $this->get_integration_config( $slug );
 
-		if ( ! is_subclass_of( $integration_object, Integration::class ) ) {
-			throw new InvalidArgumentException( sprintf( 'Integration class "%s" must extend %s.', get_class( $integration_object ), Integration::class ) );
+			if ( $vip_config->is_active_via_vip() ) {
+				$this->activate( $slug, $vip_config->get_site_config() );
```
mjangda: Here's an idea: let's change the param name for `Integration::activate` and `Integrations::activate` to `options` and `config` becomes one of the supported keys of that.

```
public function activate( string $slug, array $options = [] ): void {
```

`$options` would accept common configuration parameters across all integrations (e.g. `version`) and `$options['config']` would be the integration-specific configs.

```suggestion
				$this->activate( $slug, [
					'config' => $vip_config->get_site_config(),
				] );
```

In the future, we could even do something like this (where `get_integration_opts()` could return those generic configs stored at the platform-level):

```
				$this->activate( $slug, array_merge( $vip_config->get_integration_opts(), [
					'config' => $vip_config->get_site_config(),
				] ) );
```

@alecgeatches curious to hear your perspective on this as well.

integrations/integration-config.php
```diff
@@ -0,0 +1,176 @@
...
+		// Look for key inside client or site config.
+		if ( 'network_sites' !== $config_type && isset( $this->config[ $config_type ][ $key ] ) ) {
+			return $this->config[ $config_type ][ $key ];
+		}
+
+		// Look for key inside network-sites config.
+		$blog_id = get_current_blog_id();
+		if ( 'network_sites' === $config_type && isset( $this->config[ $config_type ][ $blog_id ] ) ) {
+			if ( isset( $this->config[ $config_type ][ $blog_id ][ $key ] ) ) {
+				return $this->config[ $config_type ][ $blog_id ][ $key ];
```
mjangda:
```suggestion
		$network_site_id = get_current_blog_id();
		if ( 'network_sites' === $config_type && isset( $this->config[ $config_type ][ $network_site_id ] ) ) {
			if ( isset( $this->config[ $config_type ][ $network_site_id ][ $key ] ) ) {
				return $this->config[ $config_type ][ $network_site_id ][ $key ];
```

integrations/parsely.php
mehmoodak: Added a ticket to add `is_managed` attribute on plugins (https://github.com/Automattic/vip-parsely/issues/61).

integrations/parsely.php
mehmoodak: When `activate` is triggered from platform side then we only set the site related data in `config` property not the whole integration configuration provided by VIP ([src](https://github.com/Automattic/vip-go-mu-plugins/blob/266b1587da401cef0ce8d60ae702cf630bdf446c/integrations/integrations.php#L73)) so it seems like it is already behaving  like you are suggesting.

integrations/integrations.php
mehmoodak: Renamed to `activate_platform_integrations`

integrations/integration.php
mehmoodak: Added `trigger_error`

integrations/integrations.php
mehmoodak: Feedback implemented in https://github.com/Automattic/vip-go-mu-plugins/pull/4660/commits/2052aadf18938305de1f026468984550bbf96778

mehmoodak:
@mjangda all of the feedback has been incorporated please have a look again. Thanks

integrations/parsely.php
mjangda: I  was thinking something like this:

```suggestion
	protected function wp_parsely_credentials_callback( $original_credentials ): array {
		$config = $this->get_config();
		
		if ( empty( $config ) ) {
			return $original_credentials;
		}

		$new_credentials = array(
			'is_managed' => true,
			'site_id' => $config['site_id'] ?? null,
			'api_secret' => $config['api_secret'] ?? null,
		};

		return $new_credentials;
```

mjangda (approved):
Testing this and it looks good.

Added one suggestion on how to pass through the parsely credentials.

integrations/parsely.php
mehmoodak: This looks easy to maintain (thanks for the suggestion), will incorporate this before merging but without `$original_credentials` param because `wp_parsely_credentials` filter on plugin doesn't pass any parameter.

mjangda: Interesting. Filters should always be passing something (even if it's a null value). I think we should assume that's the case, and update the filter in the plugin to do so as well.

mehmoodak: I think I have mistaken the filter is passing `false`. Okay will update it as suggested.

mehmoodak: Added in https://github.com/Automattic/vip-go-mu-plugins/pull/4660/commits/9839b7907cc6cfd36ecb57ebcfb1ff40bf5ea928, will add `is_managed` key later (have a separate ticket for it)

mehmoodak: Done in https://github.com/Automattic/vip-go-mu-plugins/pull/4660/commits/9839b7907cc6cfd36ecb57ebcfb1ff40bf5ea928

PR #4711, opened 2023-07-20
Add extra protections around wp_load_alloptions()

WPprodigy:
## Description
Re-implements `wp_load_alloptions()` using the new filter available in WP 6.2. It's pretty much a line for line replication ([seen here](https://github.com/WordPress/WordPress/blob/44629e628672a77edb1d3100546b458973854015/wp-includes/option.php#L289-L361)), except:

- Caches alloptions locally in the event where memcached can't add() or get().
- Doesn't fallback to the secondary alloptions query without accounting for autoload (seems unnecessary?)
- Prepares for killing the request if the alloptions query fails (is dangerous to continue as if there are no plugins, rewrite_rules, etc).

### Performance Improvements: alloptions > 1mb

Setup: Add a really large autoloaded option on the site:

```
update_option( 'vip_testing_large_option', 'longvalue_' . bin2hex( random_bytes( 1024 * 1024 ) ) );
```

#### Before mitigation
- 282 second (!!!) page generation time.
- 11.6 second db query time, 1201 alloptions queries
- 262 seconds memcached query time
    - 261.1s spent in adds() for options group (1252 opts)
    - 1.3s spent for gets() in options group (1255 opts)
    
#### After mitigation
- 1.82s page generation time.
- .0588 db query time, 2 alloptions queries.
- 748.7ms memcached query time
    - 679.1ms spent in adds() for options group (3 opts)
    - 1.9ms spent in gets() for options group (8 opts)
    
### Performance Improvements: memcached unavailable

Setup: hack object cache drop-in to mimic a memcached server outage just for the alloptions cache key (returning false from get() and add() whilst still performing the remote action though). Deleted the massive option added previously, this test is a normal size alloptions on a simple test site.

#### Before mitigation
- 5.2857s page generation time
- 1.04s db query time, 1200 alloptions queries
- 2.62s memcached query time
    - 2.04s spent in adds() for options group (1251 opts)
    - 572.6ms spent in gets() for options group (1254 opts)

#### After mitigation
- 0.3281s page generation time.
- .0018s db query time, 2 alloptions queries
- 18.1ms memcached query time
    - 3.7ms spent in adds() for options group (3 opts)
    - 2.3ms spent in gets() for options group (8 opts)
    
### Performance Improvements: database unavailable

While this patch currently does help this scenario by not doing the fallback query, ultimately I think we'll end up killing the request so it won't matter too much.

### Extra protections around the "alloptions" functionality

Increases performance and helps ensure stability.

## Pre-review checklist

### DIFF
--- a/001-core/options-api.php
+++ b/001-core/options-api.php
@@ -1,0 +1,65 @@
+<?php
+
+namespace Automattic\VIP\Core\OptionsAPI;
+
+/**
+ * Add additional protections around the alloptions functionality.
+ *
+ * Note that there is one (unavoidable) core call to get_option() before this filter is registered (in wp_plugin_directory_constants()),
+ * So by the time this starts filtering, there's already been one occurance of wp_load_alloptions().
+ *
+ * Here we re-implement most of what core does in wp_load_alloptions(), with some notable adjustments:
+ * - 1) Prevent spamming the DB if memcached is unable to add() the key to cache.
+ * - 2) (TBD) Kill the request if options cannot be retrieved from the database (or cache).
+ */
+add_filter( 'pre_wp_load_alloptions', __NAMESPACE__ . '\pre_wp_load_alloptions_protections', 2, 999 );
+function pre_wp_load_alloptions_protections( $pre_loaded_alloptions, $force_cache ) {
+	global $wpdb;
+	static $fallback_cache = [];
+
+	// Allow other filters the chance to early return (before priority 999).
+	// And abort from this special handling during installations.
+	if ( is_array( $pre_loaded_alloptions ) || wp_installing() ) {
+		return $pre_loaded_alloptions;
+	}
+
+	// 1) If successfully fetched from cache, return right away.
+	$alloptions_from_cache = wp_cache_get( 'alloptions', 'options', $force_cache );
+	if ( ! empty( $alloptions_from_cache ) && is_array( $alloptions_from_cache ) ) {
+		return apply_filters( 'alloptions', $alloptions_from_cache );
+	}
+
+	// 2) Return from the local fallback cache if available, helping prevent excess queries in cases where memcached is unable to save the results.
+	$blog_id = get_current_blog_id();
+	if ( ! $force_cache && isset( $fallback_cache[ $blog_id ] ) ) {
+		return apply_filters( 'alloptions', $fallback_cache[ $blog_id ] );
+	}
+
+	// 3) Otherwise query the DB for fresh results.
+	$suppress      = $wpdb->suppress_errors();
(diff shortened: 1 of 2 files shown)


WPprodigy:
Noting that this probably means we can start being less strict in https://github.com/Automattic/vip-go-mu-plugins/blob/c3be5f418a23223b89931d000528f5c312f55f09/alloptions-limit.php, and perhaps not force killing the request there. In fact, all of the mitigations there might be ripe for removal if/when we switch to individual cache keys for options.

001-core/options-api.php
```diff
@@ -0,0 +1,79 @@
...
+
+// phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery
+
+namespace Automattic\VIP\Core\OptionsAPI;
+
+/**
+ * Add additional protections around the alloptions functionality.
+ *
+ * Note that there is one (unavoidable) core call to get_option() before this filter is registered (in wp_plugin_directory_constants()),
+ * So by the time this starts filtering, there's already been one occurance of wp_load_alloptions().
```
rinatkhaziev:
```suggestion
 * So by the time this starts filtering, there's already been one occurrence of wp_load_alloptions().
```

rinatkhaziev:
Thanks, LGTM but cc @mjangda for any additional feedback/considerations.

rinatkhaziev (approved):

mjangda:
Haven't had a chance to review this properly yet but can we make sure this is feature flagged so we can selectively test it before rolling out everywhere?

WPprodigy:
The `_VIP_DIE_ON_ALLOPTIONS_FAILURE` piece will be feature flagged and very selectively rolled out at the start. Can chat more about the other piece, but in short it's more or less what already happens with most other cache writes - in that it doesn't actually require success before the rest of the request starts serving it from runtime cache: https://github.com/Automattic/wp-cache-memcached/blob/develop/includes/wp-object-cache.php#L293-L301

The `add()` is just a bit different since it comes with the extra contract of having to be "first". In this alloptions case though, there are sets() that already happen after say an update_option(), so that contract isn't really in place. More specifically though, [because of this logic](https://github.com/Automattic/vip-go-mu-plugins/blob/develop/misc.php#L79-L96), any request could really be the "winner" for the add() race - including requests that are coming from a read replica. I have plans of solving for this too, but separately.

For what's in this PR, it's really just edge case protection to stop the DB and memcached from being spammed into an outage. And the divergent behavior introduced here only comes into play after two failed attempts to add() to memcached (so something is quite wrong). And the behavior isn't really different, because core would just keep querying the DB each time, attempting to add(), and then continuing with the results. So nothing changes with where or how it would have gotten the data, just saves lots of trouble.

PR #4737, opened 2023-08-01
Force QM when we should enable Search Dev Tools

pschoffer:
## Description

On Multisite we would not enable Search Dev Tools for non Super Admins. The reason is that the `ep_get_query_log` function would not be available. 

That function is [loaded](https://github.com/Automattic/vip-go-mu-plugins/blob/develop/search/includes/classes/class-search.php#L513) if  `wpcom_vip_qm_enable` is true. 

`wpcom_vip_qm_enable` would check the same permission as we (`manage_options`) BUT only for [singlesite](https://github.com/Automattic/vip-go-mu-plugins/blob/develop/query-monitor/classes/QueryMonitor.php#L115)

This change makes it so that `ep_get_query_log` is always available.

This change enables Search Dev Tools for all Administrators on multisite installations.

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

1) Create multisite dev-env
2) Create a new subsite test
3) Create a new Admin user - vip dev-env exec --slug multisite -- wp user create --url=http://test.multisite.vipdev.lndo.site/ test2 test2@gmail.com --role='administrator' --user_pass=testtest
4) Log out and login as the new user test2
5) Disable debug mode by making [this method](https://github.com/Automattic/vip-go-mu-plugins/blob/b6eae073c07b2f4683929fd398f393e4ca6ffbf3/000-debug/debug-mode.php#L39) return false 
6) Search dev tools should still be visible

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -497,11 +497,10 @@ class Search {
 	 */
 	protected function maybe_enable_ep_query_logging() {
 		add_action( 'plugins_loaded', [ $this, 'enable_ep_query_logging_if_query_monitor_enabled' ] );
-		add_action( 'plugins_loaded', [ $this, 'load_ep_debug_bar_panel' ] );
+		add_action( 'plugins_loaded', [ $this, 'load_ep_get_query_log_function' ] );
 	}
 
-	public function load_ep_debug_bar_panel() {
-		// Load query log override function to remove Authorization header from requests
+	public function load_ep_get_query_log_function() {
 		require_once __DIR__ . '/../functions/ep-get-query-log.php';
 	}
 
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -497,20 +497,25 @@ class Search {
 	 */
 	protected function maybe_enable_ep_query_logging() {
 		add_action( 'plugins_loaded', [ $this, 'enable_ep_query_logging_if_query_monitor_enabled' ] );
+		add_action( 'plugins_loaded', [ $this, 'load_ep_debug_bar_panel' ] );
+	}
+
+	public function load_ep_debug_bar_panel() {
+		// Load query log override function to remove Authorization header from requests
+		require_once __DIR__ . '/../functions/ep-get-query-log.php';
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


rinatkhaziev (changes requested):
As discussed in Slack we should approach this from a different angle.

search/includes/classes/class-search.php
```diff
@@ -497,20 +497,25 @@ protected function setup_constants() {
 	 */
 	protected function maybe_enable_ep_query_logging() {
 		add_action( 'plugins_loaded', [ $this, 'enable_ep_query_logging_if_query_monitor_enabled' ] );
+		add_action( 'plugins_loaded', [ $this, 'load_ep_debug_bar_panel' ] );
+	}
+
+	public function load_ep_debug_bar_panel() {
```
rinatkhaziev: This is incorrectly named - we don't have the Debug Bar anymore + it's not loading the panel but a single helper function.

pschoffer: I renamed it to `load_ep_get_query_log_function`, what do you think. Feel free to drop name suggestions.

search/search-dev-tools/search-dev-tools.php
```diff
@@ -100,7 +100,7 @@ function rest_callback( \WP_REST_Request $request ) {
  * @return boolean
  */
 function should_enable_search_dev_tools(): bool {
-	return ( current_user_can( SEARCH_DEV_TOOLS_CAP ) || is_debug_mode_enabled() ) && function_exists( 'ep_get_query_log' );
+	return ( current_user_can( SEARCH_DEV_TOOLS_CAP ) || is_debug_mode_enabled() );
```
rinatkhaziev: Let's rollback this change - checking for function_exists is cheap but provides extra safety (in case something will change in the future)

rinatkhaziev (changes requested):
thanks, left a bit of inline feedback.

rinatkhaziev:
Other than that I tested, seems to be working!

search/search-dev-tools/search-dev-tools.php
pschoffer: Done thanks

sjinks (approved):

PR #4740, opened 2023-08-01
Trigger warning for alloptions fallback cache

WPprodigy:
Help debug the sites struggling with alloptions

### DIFF
--- a/001-core/options-api.php
+++ b/001-core/options-api.php
@@ -64,6 +64,8 @@ function pre_wp_load_alloptions_protections( $pre_loaded_alloptions, $force_cach
 	$add_result = wp_cache_add( 'alloptions', $alloptions, 'options' );
 
 	if ( false === $add_result && false === wp_cache_get( 'alloptions', 'options', true ) ) {
+		trigger_error( 'VIP: Saving to alloptions fallback cache.', E_USER_WARNING );
+
 		// Prevent memory issues in case something is looping over thousands of subsites.
 		if ( count( $fallback_cache ) > 10 ) {
 			$fallback_cache = [];


renatonascalves:
One inconvenience of this change is that it will exit with status code other than 0. Particularly in a CI environment (Github Action, Buddy, etc).

It also pollutes the logs with those warnings.

```
Warning: VIP: Saving to alloptions fallback cache. in /buddy/site.com/.buddy-tests/wordpress/wp-content/mu-plugins/001-core/options-api.php on line 67
```

WPprodigy:
Ah, wasn't aware of that effect.

I've been auditing the errors and figuring out the next steps towards resolving the common causes (looking at you, rewrite_rules and user_roles). Have also found this error to be quite helpful when issues are ongoing as it helps you see where constant (and usually problematic) option writes are coming from.

Will probably remove it eventually, but for now would it be sufficient to hide it behind a check like this?

```
if ( defined( 'WPCOM_IS_VIP_ENV' ) && true === WPCOM_IS_VIP_ENV ) {
  trigger_error( 'VIP: Saving to alloptions fallback cache.', E_USER_WARNING );
}
```

renatonascalves:
@WPprodigy That would be ideal. 

This constant is not set in a CI context and would only work on VIP environments.

renatonascalves:
On a related note: apparently the problem was/is being caused by the `object-cache.php` file. More specifically, the location of the file.

We install that file from [here](https://github.com/alleyinteractive/mantle-ci/blob/develop/install-wp-tests.sh#L390-L393).

There might be some differences between this one and the version VIP falls back to. I'd still recommend adding the conditional though, so that this error is only logged in a VIP environment.

bartoszgadomski:
Hi @WPprodigy, thank you for your work on this.

After inspecting the CI logs for unit tests, we saw thousands of occurrences of the mentioned warning (`Warning: VIP: Saving to alloptions fallback cache. in /var/www/html/wp-content/plugins/vendor/automattic/vip-go-mu-plugins/001-core/options-api.php on line 67`). If that could be triggered on a VIP environments only, that would be great 🙂

WPprodigy:
Added here: https://github.com/Automattic/vip-go-mu-plugins/pull/4759

PR #4745, opened 2023-08-04
Fix deprecation warnings

anubisthejackle:
## Description
Fixes a deprecation warning in `qm-object-cache`.

## Changelog Description
Updated `QM_Output_Html_Object_Cache_Group_Stats::output_table_cell` `$weight` parameter to type hint `float` instead of `int`.
 
## Pre-review checklist

- [X] This change uses a rollout method to ease with deployment (if applicable - especially for large scale actions that require writes).

## Pre-deploy checklist

## Steps to Test
Load a site using PHP 8.2 and confirm there are no deprecation warnings like `Deprecated: Implicit conversion from float 1.2345678908765432E-5 to int loses precision`

### DIFF
--- a/qm-plugins/qm-object-cache/html/class-qm-output-html-object-cache-group-stats.php
+++ b/qm-plugins/qm-object-cache/html/class-qm-output-html-object-cache-group-stats.php
@@ -132,7 +132,7 @@ class QM_Output_Html_Object_Cache_Group_Stats extends QM_Output_Html {
 	 *
 	 * @param $size int|null Raw size
 	 *
-	 * @return $size string Human readable size format
+	 * @return string Human readable size format
 	 */
 	public function process_size( ?int $size ) {
 		return size_format( (int) $size, 2 );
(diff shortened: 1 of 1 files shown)


qm-plugins/qm-object-cache/html/class-qm-output-html-object-cache-group-stats.php
```diff
@@ -166,11 +166,11 @@ public function output_sortable_table_col( string $title ) {
 	 * Outputs a table cell.
 	 *
 	 * @param string $value Value to be outputted in table cell
-	 * @param int|null $weight Weight by sorting priority
+	 * @param float|null $weight Weight by sorting priority
 	 *
 	 * @return void
 	 */
-	public function output_table_cell( ?string $value, int $weight = null ) {
+	public function output_table_cell( ?string $value, ?float $weight = null ) {
```
sjinks, line 173: If `$weight` is `float` now, shouldn't we explicitly convert it to `int` when printing the `data-qm-sort-weight` attribute?

anubisthejackle: Its my understanding that this data attribute is used for sorting the table in JS. I don't see any issues with leaving it as a float for display purposes. I'd argue it adds a higher level of accuracy to the sort, but I doubt that adds up to any difference realistically.

sjinks (approved):

PR #4759, opened 2023-08-08
Conditionally trigger warning for alloptions fallback

WPprodigy:
Conditionally trigger the warning only on VIP environments, helping out with CI things as mentioned here https://github.com/Automattic/vip-go-mu-plugins/pull/4740#issuecomment-1664366754

### DIFF
--- a/001-core/options-api.php
+++ b/001-core/options-api.php
@@ -64,7 +64,9 @@ function pre_wp_load_alloptions_protections( $pre_loaded_alloptions, $force_cach
 	$add_result = wp_cache_add( 'alloptions', $alloptions, 'options' );
 
 	if ( false === $add_result && false === wp_cache_get( 'alloptions', 'options', true ) ) {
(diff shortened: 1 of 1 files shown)


001-core/options-api.php
```diff
@@ -64,7 +64,9 @@ function pre_wp_load_alloptions_protections( $pre_loaded_alloptions, $force_cach
 	$add_result = wp_cache_add( 'alloptions', $alloptions, 'options' );
 
 	if ( false === $add_result && false === wp_cache_get( 'alloptions', 'options', true ) ) {
-		trigger_error( 'VIP: Saving to alloptions fallback cache.', E_USER_WARNING );
+		if ( defined( 'WPCOM_IS_VIP_ENV' ) && true === WPCOM_IS_VIP_ENV ) {
```
rinatkhaziev:
```suggestion
		if ( defined( 'WPCOM_IS_VIP_ENV' ) && true === constant( 'WPCOM_IS_VIP_ENV' ) ) {
```

PR #4765, opened 2023-08-09
Add user last seen information

luismulinari:
## Description

This PR introduces a `Last Seen` user meta to the users, which is going to be used to report and block inactive users, improving security by reducing the surface attack.

## Inactive users 
```
curl -X GET --user "example@example.org:key" "http://vip-testing-luismulinari-multi.vipdev.lndo.site/wp-json/wp/v2/users?context=edit"
{"code":"rest_forbidden_context","message":"Sorry, you are not allowed to list users.","data":{"status":401}}%
```

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test
1. Check out PR.
2. Add the following constants:
```
define( 'VIP_SECURITY_INACTIVE_USERS_ACTION', 'BLOCK' );
define( 'VIP_SECURITY_CONSIDER_USERS_INACTIVE_AFTER_DAYS', 30 );
```
3. Go to `wp-admin` > `Users`


security/user-last-seen.php
```diff
@@ -0,0 +1,60 @@
...
+	$cache_key = LAST_SEEN_UPDATE_USER_META_CACHE_KEY_PREFIX . $current_user->ID;
+
+	try {
+		if ( wp_cache_get( $cache_key ) ) {
+			return;
+		}
+
+		update_user_meta( $current_user->ID, 'vip_last_seen', time() );
+
+		wp_cache_set( $cache_key, true, null, LAST_SEEN_UPDATE_USER_META_CACHE_TTL );
```
WPprodigy: This would be less race-condition prone:

```
if ( wp_cache_add( $cache_key, true, 'vip', LAST_SEEN_UPDATE_USER_META_CACHE_TTL ) ) {
	update_user_meta( $current_user->ID, 'vip_last_seen', time() );
}
```

security/user-last-seen.php
```diff
@@ -0,0 +1,60 @@
...
+const LAST_SEEN_UPDATE_USER_META_CACHE_KEY_PREFIX = 'vip_last_seen_update_user_meta_cache_key';
+
+add_action( 'set_current_user', 'update_user_last_seen', 10, 0 );
+add_filter('manage_users_columns', 'users_columns' ) ;
+add_filter('manage_users_custom_column', 'users_custom_column', 10, 3 ) ;
+
+function update_user_last_seen() {
+	global $current_user;
+
+	$cache_key = LAST_SEEN_UPDATE_USER_META_CACHE_KEY_PREFIX . $current_user->ID;
```
WPprodigy: Not 100% sure, but might be possible for the current user ID to be `0` if some code does `wp_set_current_user( 0 );`.  So might want to safeguard that and make sure we actually have an ID first, and that `$current_user` is set / not null / has that ID property.

security/user-last-seen.php
```diff
@@ -0,0 +1,60 @@
...
+		}
+
+		update_user_meta( $current_user->ID, 'vip_last_seen', time() );
+
+		wp_cache_set( $cache_key, true, null, LAST_SEEN_UPDATE_USER_META_CACHE_TTL );
+	} catch ( \Exception $e ) {
+		trigger_error(
+			sprintf( 'failed to update user last seen meta', esc_html( $e->getMessage() ) ),
+			E_USER_WARNING
+		);
```
WPprodigy: I like error handling, but unfortunately WP doesn't really throw when things fail like DB writes. So no need here for the try/catch, as there's nothing above that will throw.

security/user-last-seen.php
WPprodigy, line 1: Let's namespace this file to avoid global function name conflicts. Example: https://github.com/Automattic/vip-go-mu-plugins/blob/e1227255d3b0c48c62fbcce37e8159263d9f2d10/security/login-error.php

luismulinari:
Thanks for such a thoughtful review @WPprodigy  ❤️

security/user-last-seen.php
```diff
@@ -0,0 +1,148 @@
...
+		// Last seen meta was checked recently
+		return $user_id;
+	}
+
+	$is_api_request = ( ( defined( 'XMLRPC_REQUEST' ) && XMLRPC_REQUEST ) || ( defined( 'REST_REQUEST' ) && REST_REQUEST ) );
+
+	if ( $is_api_request && is_considered_inactive( $user_id ) ) {
+		// To block API requests for inactive requests, we need to return a WP_Error object here
+		return new \WP_Error( 'inactive_account', __( '<strong>Error</strong>: Your account has been flagged as inactive. Please contact your site administrator.', 'inactive-account-lockdown' ) );
+	}
```
mjangda: Is there a reason we're not blocking non-API requests here?

security/user-last-seen.php
```diff
@@ -0,0 +1,148 @@
...
+		date_i18n( get_option('date_format'), $last_seen_timestamp ),
+		date_i18n( get_option('time_format'), $last_seen_timestamp )
+	);
+
+	if ( ! is_considered_inactive( $user_id ) ) {
+		return sprintf( '<span>%s</span>', esc_html__( $formatted_date ) );
+	}
+
+	$unblock_link = '';
+	if ( current_user_can( 'edit_user', array() ) ) {
```
mjangda:
```suggestion
	if ( current_user_can( 'edit_user', $user_id ) ) {
```

security/user-last-seen.php
```diff
@@ -0,0 +1,148 @@
...
+		) );
+
+		$unblock_link = "<div class='row-actions'><span>User blocked due to inactivity. <a class='reset_last_seen_action' href='" . esc_url( $url ) . "'>" . __( 'Unblock' ) . "</a></span></div>";
+	}
+	return sprintf( '<span class="wp-ui-text-notification">%s</span>' . $unblock_link, esc_html__( $formatted_date ) );
+}, 10, 3 );
+
+add_action( 'user_row_actions', function ( $actions  ) {
+	if( isset($_GET['action'] ) && $_GET['action'] === 'reset_last_seen' ){
+		$user_id = $_GET['user_id'];
```
mjangda:
```suggestion
		$user_id = absint( $_GET['user_id'] );
```

security/user-last-seen.php
```diff
@@ -0,0 +1,148 @@
...
+		delete_user_meta( $user_id, LAST_SEEN_META_KEY );
```
mjangda: This handling should probably be moved to an earlier hook like `admin_init`.

We also need to:

- validate the user exists
- check `current_user_can( 'edit_user', $user_id )`
- do a nonce check to prevent CORS

And then, after deleting, it would be nice to redirect back the users page with a success message.

security/user-last-seen.php
```diff
@@ -0,0 +1,148 @@
...
+}, 10, 1 );
+
+add_filter( 'views_users', function ( $views ) {
+	global $wpdb;
+
+	if ( ! defined( 'VIP_CONSIDER_USERS_INACTIVE_AFTER_DAYS' ) ) {
+		return $views;
+	}
+
+	$count = $wpdb->get_var( 'SELECT COUNT(meta_key) FROM ' . $wpdb->usermeta . ' WHERE meta_key = "' . LAST_SEEN_META_KEY . '" AND meta_value < ' . get_inactivity_timestamp() );
```
mjangda:
```suggestion
	$count = $wpdb->get_var( $wpdb->prepare( 'SELECT COUNT(meta_key) FROM ' . $wpdb->usermeta . ' WHERE meta_key = %s AND meta_value < %s', LAST_SEEN_META_KEY, get_inactivity_timestamp() ) );
```

security/user-last-seen.php
```diff
@@ -0,0 +1,148 @@
...
+
+	$view = __( 'Blocked Users' );
+	if ( $count ) {
+		$class = isset( $_REQUEST[ 'last_seen_filter' ] ) ? 'current' : '';
+		$view = '<a class="' . $class . '" href="users.php?last_seen_filter=blocked">' . $view . '</a>';
```
mjangda:
```suggestion
		$view = '<a class="' . esc_attr( $class ) . '" href="users.php?last_seen_filter=blocked">' . esc_html( $view ) . '</a>';
```

security/user-last-seen.php
```diff
@@ -0,0 +1,148 @@
...
+
+function is_considered_inactive( $user_id ) {
+	if ( ! defined( 'VIP_CONSIDER_USERS_INACTIVE_AFTER_DAYS' ) ) {
+		return false;
+	}
+
+	$last_seen_timestamp = get_user_meta( $user_id, LAST_SEEN_META_KEY, true );
+	if ( ! $last_seen_timestamp ) {
+		return false;
+	}
```
mjangda: If a user never logs in to their account, we should probably still deactivate them after the threshold.

security/user-last-seen.php
```diff
@@ -0,0 +1,148 @@
+<?php
+namespace Automattic\VIP\Security;
+
+const LAST_SEEN_META_KEY = 'wpvip_last_seen';
+const LAST_SEEN_UPDATE_USER_META_CACHE_TTL = MINUTE_IN_SECONDS * 5; // Store last seen once every five minute to avoid too many write DB operations
+const LAST_SEEN_UPDATE_USER_META_CACHE_KEY_PREFIX = 'wpvip_last_seen_update_user_meta_cache_key';
+const GROUP = 'wpvip';
```
mjangda: An alternate way to do this would be set the group as `user_last_seen` and the key can just be the `$user_id`.

We may also want to change the group to a global one because one user accessing 5 different network sites at the same time would trigger 5 DB writes (instead of 1).

security/user-last-seen.php
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	$cache_key = LAST_SEEN_UPDATE_USER_META_CACHE_KEY_PREFIX . $user_id;
+
+	if ( wp_cache_get( $cache_key, GROUP ) ) {
+		// Last seen meta was checked recently
+		return $user_id;
+	}
+
+	$is_api_request = ( ( defined( 'XMLRPC_REQUEST' ) && XMLRPC_REQUEST ) || ( defined( 'REST_REQUEST' ) && REST_REQUEST ) );
```
mjangda: If we add `use Automattic\VIP\Utils\Context` to the top, we can do this:

```suggestion
	$is_api_request = Context::is_rest_api() || Context::is_xmlrpc_api();
```

mjangda (changes requested):
Spotted a couple of things I wanted to flag.

security/user-last-seen.php
```diff
@@ -0,0 +1,273 @@
...
+			return;
+		}
+
+		// Use a global cache group to avoid having to the data for each site
+		wp_cache_add_global_groups( array( self::LAST_SEEN_CACHE_GROUP ) );
+
+		add_action( 'admin_init', array( $this, 'register_release_date' ) );
+
+		add_filter( 'determine_current_user', array( $this, 'determine_current_user' ), 30, 1 );
+		add_filter( 'authenticate', array( $this, 'authenticate' ), 20, 1 );
```
WPprodigy: Can this be brought down inside the conditional hooks below? `authenticate` looks like it can move down to being inside `$this->is_block_action_enabled()` rather than checking that condition inside the method itself.

security/user-last-seen.php
```diff
@@ -0,0 +1,273 @@
...
+		if ( wp_cache_get( $user_id, self::LAST_SEEN_CACHE_GROUP ) ) {
+			// Last seen meta was checked recently
+			return $user_id;
+		}
+
+		if ( $this->is_block_action_enabled() && $this->is_considered_inactive( $user_id ) ) {
+			// Force current user to 0 to avoid recursive calls to this filter
+			wp_set_current_user( 0 );
+
+			return new \WP_Error( 'inactive_account', __( '<strong>Error</strong>: Your account has been flagged as inactive. Please contact your site administrator.', 'inactive-account-lockdown' ) );
```
WPprodigy: For the translation text domain, I think maybe we should stick with `wpvip` to keep mu-plugins easier to translate all around. But we've not really kept that up super great.

If this intends to ever be a standalone plugin though, then keeping a specific text domain does make sense.

security/user-last-seen.php
```diff
@@ -0,0 +1,273 @@
...
+				'fields'       => 'ID',
+				'meta_key'     => self::LAST_SEEN_META_KEY,
+				// phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_value
+				'meta_value'   => $this->get_inactivity_timestamp(),
+				'meta_type'    => 'NUMERIC',
+				'meta_compare' => '<',
+				'count_total'  => true,
+			),
+		);
+		$count       = (int) $users_query->get_total();
```
WPprodigy: This could potentially be very expensive, and is just for some UI text that isn't super required. Some thoughts:

1) Maybe we just query w/ a limit of 1 user, and if it finds one then we show the filters/buttons.
2) Query with a limit of 101, and show 100+ if it returns the full amount.
3) Drop the query altogether and just show a button without a count.

security/user-last-seen.php
```diff
@@ -0,0 +1,273 @@
...
+
+		wp_safe_redirect( $url );
+		exit();
+	}
+
+	public function register_release_date() {
+		if ( ! get_option( self::LAST_SEEN_RELEASE_DATE_TIMESTAMP_OPTION_KEY ) ) {
+			// Right after the first admin_init, set the release date timestamp
+			// to be used as a fallback for users that never logged in before.
+			add_option( self::LAST_SEEN_RELEASE_DATE_TIMESTAMP_OPTION_KEY, time() );
```
WPprodigy: Let's make this a non-autoloaded option, since it's not something needed on every request.

security/user-last-seen.php
```diff
@@ -0,0 +1,273 @@
...
+		if ( $release_date_timestamp ) {
+			return $release_date_timestamp < $this->get_inactivity_timestamp();
+		}
+
+		// Release date is not defined yed, so we can't consider the user inactive.
+		return false;
+	}
+
+	public function get_inactivity_timestamp() {
+		return strtotime( sprintf( '-%d days', constant( 'VIP_SECURITY_CONSIDER_USERS_INACTIVE_AFTER_DAYS' ) ) ) + self::LAST_SEEN_UPDATE_USER_META_CACHE_TTL;
```
WPprodigy: Do we need to validate/sanitize `VIP_SECURITY_CONSIDER_USERS_INACTIVE_AFTER_DAYS` first here? Make sure it's a number string and if not fallback to some sane default?

WPprodigy (commented):
Left a few thoughts. And then one larger question is whether we want to be setting this meta on all user roles or maybe just the ones with elevated permissions?

For example, subscriber (core) and customer (woocommerce) don't seem like account types we should be blocking. Returning to an eCommerce site a year or two later and being blocked form ordering or checking receipts is not good.

It additionally would increase the metadata in the DB and make the user queries here more painful.

Rather than one-by-one allow-listing certain roles, maybe we look for specific caps like manage_options, publish_posts, edit_posts, etc.

security/user-last-seen.php
luismulinari: It makes total sense!

Implemented on https://github.com/Automattic/vip-go-mu-plugins/pull/4765/commits/bb49b0cb066c6e26aff2f395733629ecd457c1fb

security/user-last-seen.php
luismulinari: Fixed all the translation domains to `wpvip` on https://github.com/Automattic/vip-go-mu-plugins/pull/4765/commits/bb49b0cb066c6e26aff2f395733629ecd457c1fb

security/user-last-seen.php
luismulinari: Good idea!

I ended up choosing `1` since it seems to be the best cost/benefit for now.

Implemented on https://github.com/Automattic/vip-go-mu-plugins/pull/4765/commits/bb49b0cb066c6e26aff2f395733629ecd457c1fb

security/user-last-seen.php
luismulinari: TIL about autoloaded options!

Implemented on https://github.com/Automattic/vip-go-mu-plugins/pull/4765/commits/bb49b0cb066c6e26aff2f395733629ecd457c1fb

security/user-last-seen.php
luismulinari: Good idea.

Implemented on https://github.com/Automattic/vip-go-mu-plugins/pull/4765/commits/bb49b0cb066c6e26aff2f395733629ecd457c1fb

luismulinari:
Hey @WPprodigy, thanks for this awesome review once again. It's been a joy to learn about WordPress from you :)

> one larger question is whether we want to be setting this meta on all user roles or maybe just the ones with elevated permissions?

It makes total sense. I implemented that as you suggested and also added a filter `vip_security_last_seen_elevated_capabilities` to be able to customize the elevated capabilities on the fly.

What do you think?

security/user-last-seen.php
```diff
@@ -0,0 +1,330 @@
...
+	public function add_blocked_users_filter( $views ) {
+		$blog_id = is_network_admin() ? null : get_current_blog_id();
+
+		$users_query = new \WP_User_Query(
+			array(
+				'blog_id'      => $blog_id,
+				'fields'       => 'ID',
+				'meta_key'     => self::LAST_SEEN_META_KEY,
+				// phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_value
+				'meta_value'   => $this->get_inactivity_timestamp(),
```
WPprodigy: Because this number always changes, it's not able to take advantage of any query caching 😬 . TBD on performance for this on larger sites, but might want to re-visit layer and add a simple caching layer around it.

security.php
```diff
@@ -16,6 +16,13 @@
 require_once __DIR__ . '/security/login-error.php';
 require_once __DIR__ . '/security/password.php';
 
+if ( defined( 'VIP_SECURITY_INACTIVE_USERS_ACTION' ) ) {
+	require_once __DIR__ . '/security/user-last-seen.php';
+
+	$last_seen = new \Automattic\VIP\Security\User_Last_Seen();
+	$last_seen->init();
+}
```
WPprodigy: I think maybe we should load all this on the `plugins_loaded` hook, to allow for sites to have more time to conditionally setup the constants and use the filters.

That said, need to make sure it's still early enough to catch any of the authenticate calls.

```
	if ( defined( 'VIP_SECURITY_INACTIVE_USERS_ACTION' ) ) {
		require_once __DIR__ . '/security/user-last-seen.php';

		$last_seen = new \Automattic\VIP\Security\User_Last_Seen();
		$last_seen->init();
	}
} );
```

WPprodigy (commented):
The latest changes look good, I altered the capabilities part a bit.

Additionally, when a new user is added or a user is promoted and we’re past the LAST_SEEN_RELEASE_DATE_TIMESTAMP_OPTION_KEY period, then they are gonna start out as being blocked I think?

security/class-user-last-seen.php
```diff
@@ -0,0 +1,388 @@
...
+		}
+
+		$this->release_date = get_option( self::LAST_SEEN_RELEASE_DATE_TIMESTAMP_OPTION_KEY );
+
+		// Use a global cache group since users are shared among network sites.
+		wp_cache_add_global_groups( array( self::LAST_SEEN_CACHE_GROUP ) );
+
+		add_action( 'admin_init', array( $this, 'register_release_date' ) );
+		add_action( 'set_user_role', array( $this, 'user_promoted' ) );
+		add_action( 'vip_support_user_added', function( $user_id ) {
```
rebeccahum: Since we already have an existing job that clears out the `vip_support` users (https://github.com/Automattic/vip-go-mu-plugins/blob/66912f3ca4e270870250f35ab96d81c192b95a1a/vip-support/class-vip-support-user.php#L90) after a timeperiod of being added, do you think we can skip checking those types of users with that role for inactivity altogether?

security/class-user-last-seen.php
```diff
@@ -0,0 +1,388 @@
...
+	const LAST_SEEN_CACHE_GROUP                            = 'wpvip_last_seen';
+	const LAST_SEEN_UPDATE_USER_META_CACHE_TTL             = MINUTE_IN_SECONDS * 5; // Store last seen once every five minute to avoid too many write DB operations
+	const LAST_SEEN_RELEASE_DATE_TIMESTAMP_OPTION_KEY      = 'wpvip_last_seen_release_date_timestamp';
+
+	public function init() {
+		if ( ! defined( 'VIP_SECURITY_INACTIVE_USERS_ACTION' ) || constant( 'VIP_SECURITY_INACTIVE_USERS_ACTION' ) === 'NO_ACTION' ) {
+			return;
+		}
+
+		$this->release_date = get_option( self::LAST_SEEN_RELEASE_DATE_TIMESTAMP_OPTION_KEY );
```
rebeccahum, line 25: Are we using this property anywhere or just directly checking the option?

security/class-user-last-seen.php
```diff
@@ -0,0 +1,388 @@
...
+		$error = null;
+		if ( ! isset( $_GET['reset_last_seen_nonce'] ) || ! wp_verify_nonce( sanitize_text_field( $_GET['reset_last_seen_nonce'] ), 'reset_last_seen_action' ) ) {
+			$error = __( 'Unable to verify your request', 'wpvip' );
+		}
+
+		if ( ! get_userdata( $user_id ) ) {
+			$error = __( 'User not found.', 'wpvip' );
+		}
+
+		if ( ! current_user_can( 'edit_user', $user_id ) ) {
```
rebeccahum, line 267: Minor, but can we move this permissions check up earlier?

rebeccahum (changes requested):

security/class-user-last-seen.php
WPprodigy: The options main purpose is for: https://github.com/Automattic/vip-go-mu-plugins/pull/4765/files#diff-d6cd56ee4fe98abf6b5d384fc94675502abd9ad6a5d412954a4ccccc35cb2bb4R311-R314

But I agree, let's just call get_option() when it's needed. There is runtime caching in place for it, and validation is handled internally already. So no need for the class property to maintain it's own state. As-is, the current version will run this DB query/memcache get on every request rather unnecessarily.

security/class-user-last-seen.php
```diff
@@ -94,23 +99,35 @@ public function authenticate( $user ) {
...
+		return $status;
+	}
+
+	/**
+	 * @param bool $available True if application password is available, false otherwise.
+	 * @param \WP_User $user The user to check.
+	 * @return bool
+	 */
+	public function application_password_authentication( $available, $user ) {
+		global $wp_last_seen_application_password_error;
```
WPprodigy: Let's use a class property rather than a global

luismulinari: Good call. Done: 182dc0d33

nickdaugherty:
I have a higher level thought on this ... can we reduce the instances of VIP-specific naming inside the class (meta/option keys, filter names) and make it more generic? This functionality is something that has wider appeal outside of VIP and by removing VIP references, re create an easier path to open sourcing this as a standalone plugin one day. It's always a goal to contribute back when possible and this seems like it can be a great standalone plugin in the future.

security/class-user-last-seen.php
```diff
@@ -0,0 +1,420 @@
...
+		return defined( 'VIP_SECURITY_INACTIVE_USERS_ACTION' ) && constant( 'VIP_SECURITY_INACTIVE_USERS_ACTION' ) === 'BLOCK';
+	}
+
+	private function should_check_user_last_seen( $user_id ) {
+		/**
+		 * Filters the users that should be skipped when checking/recording the last seen.
+		 *
+		 * @param array $skip_users The list of user IDs to skip.
+		 */
+		$skip_users = apply_filters( 'vip_security_last_seen_skip_users', array() );
```
nickdaugherty, line 368: What do you think about moving this filter to `is_considered_inactive()`? That gives more flexibility for filtering the results of the check and might be more predictable.

security/class-user-last-seen.php
```diff
@@ -0,0 +1,420 @@
...
+			'reset_last_seen_success' => 1,
+		), $url );
+
+		wp_safe_redirect( $url );
+		exit();
+	}
+
+	public function ignore_inactivity_check_for_user( $user_id, $until_timestamp = null ) {
+		if ( ! $until_timestamp ) {
+			$until_timestamp = strtotime( '+2 days' );
```
nickdaugherty, line 301: The default of `+2 days` feels a little too magic and is hard to predict. Maybe we should just require `$until_timestamp` be specified?

nickdaugherty: Minor, but I'd also consider naming the function `ignore_inactivity_check_for_user_until( $user_id, $until_timestamp )`. The `_until` makes it obvious that it's a limited time action, vs. permanent (which is what I thought it would do before I read the code)

PR #4771, opened 2023-08-11
Adjust error logging in Jetpack Connection Pilot

WPprodigy:
Mostly just altering the error logging and cleaning up the way errors are passed around so that we'll have a better idea of the full execution path when something goes wrong.

Also removed some unnecessary filtering, and instead just performed the actions inline.


vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
```diff
@@ -288,6 +302,8 @@ private static function maybe_create_user() {
 	 * This helps prevent cache issues for times where the database was directly updated.
 	 */
 	private static function refresh_options_cache() {
+		wp_cache_flush_runtime();
```
WPprodigy, line 305: Small addition just to help ensure the request doesn't carry over some other stale object cache values that may have been updated via the roundtrips with Jetpack.

vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
```diff
@@ -168,31 +167,31 @@ public static function connect_site( bool $skip_connection_tests = false, bool $
...
-
-			if ( is_akismet_key_invalid() ) {
-				$original_user = wp_get_current_user();
-				// Getting wpcomvip user, since it's the owner of the Jetpack connection
-				$vip_user = get_user_by( 'login', 'wpcomvip' );
-				if ( ! $original_user || ! $vip_user ) {
-					return false;
-				}
+		if ( ! class_exists( 'Akismet_Admin' ) || ! method_exists( 'Akismet_Admin', 'connect_jetpack_user' ) || ! function_exists( 'is_akismet_key_invalid' ) ) {
+			return new WP_Error( 'jp-cxn-pilot-akismet-dependencies-missing', 'Akismet is missing required functions/methods to perform the connection.' );
```
rinatkhaziev, line 174: This doesn't satisfy `bool` return type.

vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
```diff
@@ -168,31 +167,31 @@ public static function connect_site( bool $skip_connection_tests = false, bool $
...
 
-				return $result;
-			}
+		// Put the machine user in charge of things.
+		wp_set_current_user( $user->ID );
 
-			return true;
+		$result = \Akismet_Admin::connect_jetpack_user();
+		if ( ! $result ) {
+			return new WP_Error( 'jp-cxn-pilot-akismet-connection-failed', 'Akismet could not be connected.' );
```
rinatkhaziev, line 191: same here

rinatkhaziev (changes requested):
I think there's possibility of fatals now because return type might not match in connect_akismet - left inline notes.

vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
WPprodigy: nice catch! fixed

PR #4788, opened 2023-08-16
Sunrise: Allow setting WP-CLI context via blog_id

mjangda:
Adds a new param `--load-blog-id` which allows loading blog context via a blog_id value instead of a URL.

Based on https://github.com/wp-cli/wp-cli/issues/5649#issuecomment-1227814680

### WP-CLI: Add new --load-blog-id global param

We've added a new global param (`--load-blog-id`) to WP-CLI, which allows loading blog context via a blog_id value instead of a URL. This is helpful in cases where you have an ID but not a URL.

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

TODO

### DIFF
--- a/lib/sunrise/sunrise.php
+++ b/lib/sunrise/sunrise.php
@@ -25,6 +25,44 @@ unset( $mu_plugin_dir );
 use Automattic\VIP\Utils\Context;
 
 /**
+ * Adds support for loading blog context via the `--load-blog-id` param.
+ *
+ * Borrowed from WP.com. This does a reverse lookup for the URL based on the specified ID and passes that to WP-CLI.
+ * This ensures that the correct context is loaded. It's helpful in cases where you have an ID but not a URL.
+ *
+ * Usage: `wp --load-blog-id=3 option get home`
+ *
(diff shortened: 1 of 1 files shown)


GaryJones:
I think `--site-id` may work better than references to "blog" - I know for VIP we have (well @yolih has) tried to standardise terminology around "site" including on multisite applications. The use of the core `get_site()` function already points in this direction as well.

And some customer applications aren't considered blogs, so having an interface that tries to reference it as one seems odd.

lib/sunrise/sunrise.php
```diff
@@ -97,6 +135,11 @@ function handle_not_found_error( $error_type ) {
 	}
 }
 
+// Overrides related to WP-CLI
+if ( Context::is_wp_cli() ) {
```
andrea-sdl, line 143: @mjangda Do you think checking for `is_multisite()` too would be helpful here?

GaryJones: Does line 8 at the top of this file already cover that?

andrea-sdl: @GaryJones you're correct! I didn't notice that

PR #4410, opened 2023-04-28
Fixing bug where Super Admins cannot setup Time Based One-Time Password as first option Two Factor 

spenserhale:
## Changelog Description
I am setting the controller parameter configuration to have user_id as an integer instead of a number because strict comparison requires matching types. Please have a look at #4409 for more information.

Fixes #4409

## Checklist

Please make sure the items below have been covered before requesting a review:

- [Y] This change works and has been tested locally (or has an appropriate fallback).
- [Y] This change works and has been tested on a Go sandbox.
  - I used filters to apply this change in our WP VIP production.
- [Y] This change has relevant unit tests.
  - No tests were found
- [Y] This change uses a rollout method to ease with deployment
  - Not applicable
- [Y] This change has relevant documentation additions/updates.
  - I do not believe applicable
- [Y] I've created a changelog description that aligns with the provided examples.
  - See #4409 

## Hotfix Code
```php
function hooks_filter_two_factor_endpoint_configuration(array $endpoints): array {
    if ( isset($endpoints['/two-factor/1.0/totp']) ) {
        $endpoints['/two-factor/1.0/totp'][0]['args']['user_id']['type'] = 'integer';
        $endpoints['/two-factor/1.0/totp'][1]['args']['user_id']['type'] = 'integer';
    }

    if ( isset($endpoints['/two-factor/1.0/generate-backup-codes']) ) {
        $endpoints['/two-factor/1.0/totp'][0]['args']['user_id']['type'] = 'integer';
    }

    return $endpoints;
}
add_filter( 'rest_endpoints', 'hooks_filter_two_factor_endpoint_configuration' );
```

## Steps to Test

1. Setup a multisite
2. Create a new user that is a super_admin
3. Log in as the new super_admin
4. Go to `wp-admin` > `Edit Profile`
5. submit an authentication code

PR #4797, opened 2023-08-17
Fix 2FA setup process when TOTP is the first method being activated

WPprodigy:
## Description

When TOTP is the first 2FA method being activated, it fails to allow the code to be validated due to the capability checks failing in the rest endpoint: https://github.com/Automattic/vip-go-mu-plugins-built/blob/09e2cde1496a173a694ed383ea329eb185d3b26f/shared-plugins/two-factor/providers/class-two-factor-totp.php#L94C1-L95C1

This is because between `$user_id` and `$args[0]`, one is an integer and one is a double. And the triple equals doesn't appreciate that.

### Fixed bug when setting up TOTP as the first 2FA method on an account

Was previously not possible to activate TOTP as the first method due to a bug in the capabilities logic.

## Pre-review checklist

## Steps to Test

- Make a new account, enforce 2fa, and try to setup TOTP.
- More details in BB8-10578

### DIFF
--- a/two-factor.php
+++ b/two-factor.php
@@ -191,7 +191,7 @@ function wpcom_vip_two_factor_filter_caps( $caps, $cap, $user_id, $args ) {
 		];
 
 		// You can edit your own user account (required to set up 2FA)
-		if ( 'edit_user' === $cap && ! empty( $args ) && $user_id === $args[0] ) {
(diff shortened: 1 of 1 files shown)


mjangda:
Related: https://github.com/Automattic/vip-go-mu-plugins/pull/4410

sjinks:
I believe that #4410 is fixed in #4684

sjinks (approved):

PR #4802, opened 2023-08-21
Adds "is_managed" key in Parse.ly Integration if integration is managed by VIP

mehmoodak:
## Description

By default we shows a banner on the plugin which let customers know to add credentials but for integration enabled via VIP we don't want to show this banner because if the site is yet to launch then we don't have its domain and without domain we cannot setup Parse.ly credentials so enabling the integration properly is a two step process:

1. Enable the integration when environment is init so that customers are already aware of Parse.ly and don't have surprises or unexpected behavior during launch.

2. Configure the integration ( i.e. setup credentials ) after the launch.

## Changes Made

- While forwarding credentials from platform we have added `is_managed' key which will give signal to the plugin that platform is managing this integration so don't show any banner.

## Changelog Description

Adds "is_managed" key in Parse.ly Integration if managed by VIP

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

1. Check out PR
1. Clone wp-parsely which is ready for v3.9 and `wp_parsely_credentials` filter (`git clone git@github.com:Parsely/wp-parsely.git wp-parsely-3.8`)
1. Add Parse.ly integration config as mentioned in this PR (https://github.com/Automattic/vip-go-mu-plugins/pull/4660) but have some without config / credentials
1. Make sure that the Parse.ly warning banner is not appearing on sites that are enabled but does not have credentials

### DIFF
--- a/integrations/parsely.php
+++ b/integrations/parsely.php
@@ -41,15 +41,21 @@ class ParselyIntegration extends Integration {
 	 * @return array
 	 */
 	public function wp_parsely_credentials_callback( $original_credentials ) {
-		$config = $this->get_config();
+		$config      = $this->get_config();
+		$credentials = array();
 
+		// If config provided by VIP is empty then take original credentials else take config.
 		if ( empty( $config ) ) {
-			return $original_credentials;
+			$credentials = $original_credentials;
+		} else {
+			$credentials = array(
+				'site_id'    => $config['site_id'] ?? null,
+				'api_secret' => $config['api_secret'] ?? null,
+			);
 		}
 
-		return array(
-			'site_id'    => $config['site_id'] ?? null,
-			'api_secret' => $config['api_secret'] ?? null,
-		);
+		// Adds `is_managed` flag to indicate that platform is managing this integration
+		// and we have to hide the credential banner warning or more.
+		return array_merge( array( 'is_managed' => true ), $credentials );
 	}
 }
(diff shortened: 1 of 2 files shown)


integrations/parsely.php
```diff
@@ -41,15 +41,18 @@ public function load(): void {
...
+		$credentials = array();
 
-		if ( empty( $config ) ) {
-			return $original_credentials;
-		}
-
-		return array(
+		// If config provided by VIP is empty then take original credentials else take
+		// credentials from config.
+		$credentials = empty( $config ) ? $original_credentials : array(
```
alecgeatches, line 49: Small suggestion - may be better to check for the keys you need explicitly (`site_id` and `api_secret`) rather than any `$config` value. If `$config` currently or in the future accepts other configuration keys, this code would set both values to `null` if any other configuration key is set.

```suggestion
		$credentials = ( isset( $config[ 'site_id' ] ) && isset( $config[ 'api_secret' ] ) ? array( ... ) : $original_credentials;
```

(note this implementation switches the order of the array values and `$original_credentials`)

alecgeatches (approved):
Small suggestion on `$config` value testing, otherwise good to go!

integrations/parsely.php
mehmoodak: `$original_credentials` means credentials that we got from plugin so I think its fine to return as it is without any extra checks to keep things simple. Later we can improve if needed.

PR #4810, opened 2023-08-24
Use wpcf7_upload_dir filter to set Contact Form 7 uploads directory

raamdev:
## Description

In 2021, Contact Form 7 changed the way it handles the `WPCF7_UPLOADS_TMP_DIR` constant (see https://github.com/rocklobster-in/contact-form-7/commit/90e87ab044ce6b4ff4e4df755576795908340181), which resulted in [our fix](https://github.com/Automattic/vip-go-mu-plugins/blob/524a46de14a6e5b0bc733348c1866b8a344468a5/plugin-fixes.php#L10-L19) no longer working.

The issue is that the plugin only returns `WPCF7_UPLOADS_TMP_DIR` [when the value is inside `WP_CONTENT_DIR`](https://github.com/rocklobster-in/contact-form-7/blob/2cfaa472fa485c6d3366fcdd80701fdaf7f9e425/includes/file.php#L322). When it's not (as in our case where we set it to `/tmp/cf7`), the plugin ignores the value of `WPCF7_UPLOADS_TMP_DIR` and [instead uses its own `wpcf7_upload_dir()` function](https://github.com/rocklobster-in/contact-form-7/blob/2cfaa472fa485c6d3366fcdd80701fdaf7f9e425/includes/file.php#L327) to build the uploads path, setting it to a path inside `wp-content/` and trying to write there, which fails on the VIP platform due to that path being an object store.

So, setting `WPCF7_UPLOADS_TMP_DIR` to `/tmp/cf7` is ineffective. Instead, we need to filter `wpcf7_upload_dir` to ensure that Contact Form 7 only attempts to write to `/tmp/cf7`.

This PR filters `wpcf7_upload_dir` and sets the directory to `/tmp/cf7/`, ensuring that Contact Form 7 will only use that temp path for uploads and not try to write to `wp-content/`.

This fix should allow us to remove [all of the existing Contact Form 7 fixes](https://github.com/Automattic/vip-go-mu-plugins/blob/524a46de14a6e5b0bc733348c1866b8a344468a5/plugin-fixes.php#L10-L42) we currently have in the VIP MU plugins:

- [`WPCF7_UPLOADS_TMP_DIR` does not need to be set](https://github.com/Automattic/vip-go-mu-plugins/blob/524a46de14a6e5b0bc733348c1866b8a344468a5/plugin-fixes.php#L10C4-L19) because the plugin ignores it anyway when we set it to a `/tmp/ path and if we're filtering `wpcf7_upload_dir`, there's no need to also define `WPCF7_UPLOADS_TMP_DIR`.
- [The fix for cleaning up uploaded attachments](https://github.com/Automattic/vip-go-mu-plugins/blob/524a46de14a6e5b0bc733348c1866b8a344468a5/plugin-fixes.php#L21-L42) can also be removed, since [`wpcf7_upload_tmp_dir()` will always return our temp directory](https://github.com/rocklobster-in/contact-form-7/blob/2cfaa472fa485c6d3366fcdd80701fdaf7f9e425/includes/file.php#L347) if we're filtering `wpcf7_upload_dir`, so there should be no need to remove the `wpcf7_cleanup_upload_files` action (this fix was described here: https://github.com/Automattic/vip-go-mu-plugins/pull/4044)

I have not removed those fixes in this PR. They are redundant with this filter in place, but they do not cause any conflict with added filter.

### Plugin fixes updated to improve Contact Form 7 upload compatibility 

Added a filter to set the Contact Form 7 uploads directory to a temp directory.

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test
1. Install a recent Contact Form 7 (e.g. 5.8) on a VIP test site
2. Create a contact form that includes a field for uploading an attachment
3. Submit the form with the attachment and notice how the form fails to submit due to CF7 trying to write to `wp-content/`
4. Checkout this PR and switch the site to use the appropriate version of mu-plugins
5. Perform the same test with submitting a form that includes an attachment and observe that the form submits as expected as CF7 writes to the temp directory instead of `wp-content/`


raamdev:
**DRAFT:** I'm marking this PR as a draft for now because during testing I discovered that while this fixes the issue with Contact Form 7 forms that include an attachment hanging on submission on the VIP platform, CF7's mailing routines that email those attachments to the site owner [does not honor the `wpcf7_upload_dir` filter](https://github.com/rocklobster-in/contact-form-7/blob/2cfaa472fa485c6d3366fcdd80701fdaf7f9e425/includes/mail.php#L220-L235).

This bug in the plugin results in the form attachments not being included in the form submission email when the `wpcf7_upload_dir` filter is used to set an uploads path outside of `WP_CONTENT_DIR`. 

I'm going to reach out to the plugin author to propose a fix and will come back to this PR whenever the plugin is patched.

makbeta:
@raamdev Seems like you are solving for a similar issue, a few others has brought up previously. Sadly, similar solutions to allow for the file uploads outside the wp-content directory has been rejected. :-( 
I have created and maintain a fork of the code with the solution to this problem, which is different from yours, but solves the problem. The links to all the similar issues previously brought up to the plugin authors are included in the read me file of my repository.
https://github.com/makbeta/contact-form-7

I hope it is helpful in getting this plugin working in your site. I'm subscribing to your requests to see if the author will finally accept some kind of contribution that would allow uploads outside wp-content directory.

raamdev:
@makbeta Thanks for the comment and the additional context.

I forgot to leave a reference here last week on this Pull Request, but I opened a GitHub issue (https://github.com/rocklobster-in/contact-form-7/issues/1282) and a Pull Request (https://github.com/rocklobster-in/contact-form-7/pull/1283) with the plugin developer. 

If those changes are merged upstream, this PR for the WordPress VIP platform that uses the `wpcf7_upload_dir` filter to set the upload directory to `get_temp_dir()` will fix the incompatibility with the VIP platform and allow Contact Form 7 form attachments to work as expected.

raamdev:
The plugin developer [has agreed](https://github.com/rocklobster-in/contact-form-7/issues/1282#issuecomment-1703657785) to allow a path defined in `WP_TEMP_DIR` to be considered a content path for storing temporary uploads (see https://github.com/rocklobster-in/contact-form-7/commit/42b605abd6560a5bf72c5fabcf4346f90d1ad805), so I'm going to close this PR in favor of #4901.

PR #4839, opened 2023-09-05
Adds wp_parsely_managed_options filter in Parse.ly integration

mehmoodak:
## Description

In `wp-parsely v3.9` we introduced `wp_parsely_managed_options` filter which will disable some plugin settings options and show `Update` call to action link, now we need to use that filter in our Parse.ly integration to customise the settings if the plugin is managed by VIP.

## Changes Made

- Added `wp_parsely_managed_options` filter in parsely integration.
- After adding filter I noticed that it is getting called even if the integration is not enabled via VIP so to resolve this issue following things has been done.
  - Added `configure_for_vip` function in our base integration class which only get called if the integration is enabled via VIP config.
  - Added `is_integration_already_available_via_customer` function in our base class and used it inside `activate` method to avoid activating the plugin if the plugin is already loaded via customer code.

## Changelog Description

Adds `wp_parsely_managed_options` filter in Parse.ly integration

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

1. Check out PR
1. Clone wp-parsely (`git clone git@github.com:Parsely/wp-parsely.git wp-parsely-3.8`)
1. Add VIP config of Parse.ly integration with only status attribute as mentioned in this PR (https://github.com/Automattic/vip-go-mu-plugins/pull/4660) 
1. Verify `wp-parsely` is enabled if status is `enabled` and UI is showing upgrade links
1. Verify `wp-parsely` is disbaled if status is not `enabled` and UI not showing any upgrade links
1. Verify that status attribute of VIP config have no effect if the integration is activated via customer manually or already loaded via customer code.

### DIFF
--- a/integrations/block-data-api.php
+++ b/integrations/block-data-api.php
@@ -23,7 +23,7 @@ class BlockDataApiIntegration extends Integration {
 	protected string $version = '1.0';
 
 	/**
-	 * Returns `true` if `Block Data API` is already available via customer code. We will use
+	 * Returns `true` if `Block Data API` is already available e.g. via customer code. We will use
 	 * this function to prevent activating of integration from platform side.
 	 */
 	public function is_loaded(): bool {
@@ -38,6 +38,14 @@ class BlockDataApiIntegration extends Integration {
 	public function load(): void {
 		// Wait until plugins_loaded to give precedence to the plugin in the customer repo.
 		add_action( 'plugins_loaded', function() {
+			// Return if the integration is already loaded.
+			//
+			// In activate() method we do make sure to not activate the integration if its already loaded
+			// but still adding it here as a safety measure i.e. if load() is called directly.
+			if ( $this->is_loaded() ) {
+				return;
+			}
+
 			// Load the version of the plugin that should be set to the latest version, otherwise if it's not found deactivate the integration.
 			$load_path = WPMU_PLUGIN_DIR . '/vip-integrations/vip-block-data-api-' . $this->version . '/vip-block-data-api.php';
 			if ( file_exists( $load_path ) ) {
--- a/integrations/block-data-api.php
+++ b/integrations/block-data-api.php
@@ -26,7 +26,7 @@ class BlockDataApiIntegration extends Integration {
 	 * Returns `true` if `Block Data API` is already available via customer code. We will use
 	 * this function to prevent activating of integration from platform side.
 	 */
-	public function is_integration_already_available_via_customer(): bool {
+	public function is_loaded(): bool {
 		return defined( 'VIP_BLOCK_DATA_API_LOADED' );
 	}
 
@@ -51,5 +51,5 @@ class BlockDataApiIntegration extends Integration {
 	/**
 	 * Configure `Block Data API` for VIP Platform.
 	 */
-	public function configure_for_vip(): void {}
+	public function configure(): void {}
 }
--- a/integrations/block-data-api.php
+++ b/integrations/block-data-api.php
@@ -23,6 +23,14 @@ class BlockDataApiIntegration extends Integration {
 	protected string $version = '1.0';
 
 	/**
+	 * Returns `true` if `Block Data API` is already available via customer code. We will use
+	 * this function to prevent activating of integration from platform side.
+	 */
+	public function is_integration_already_available_via_customer(): bool {
+		return defined( 'VIP_BLOCK_DATA_API_LOADED' );
+	}
+
+	/**
 	 * Applies hooks to load Block Data API plugin.
 	 *
 	 * @private
@@ -30,11 +38,6 @@ class BlockDataApiIntegration extends Integration {
 	public function load(): void {
 		// Wait until plugins_loaded to give precedence to the plugin in the customer repo.
 		add_action( 'plugins_loaded', function() {
-			// Do not load plugin if already loaded by customer code.
-			if ( defined( 'VIP_BLOCK_DATA_API_LOADED' ) ) {
-				return;
-			}
-
 			// Load the version of the plugin that should be set to the latest version, otherwise if it's not found deactivate the integration.
 			$load_path = WPMU_PLUGIN_DIR . '/vip-integrations/vip-block-data-api-' . $this->version . '/vip-block-data-api.php';
 			if ( file_exists( $load_path ) ) {
@@ -44,4 +47,9 @@ class BlockDataApiIntegration extends Integration {
 			}
 		} );
 	}
+
+	/**
+	 * Configure `Block Data API` for VIP Platform.
+	 */
+	public function configure_for_vip(): void {}
 }
--- a/integrations/integration.php
+++ b/integrations/integration.php
@@ -63,6 +63,15 @@ abstract class Integration {
 	 * @private
 	 */
 	public function activate( array $options = [] ): void {
+		// If integration is already available in customer code then don't activate it from platform side.
+		if ( $this->is_integration_already_available_via_customer() ) {
+			trigger_error( // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_trigger_error
+				sprintf( 'Prevented activating of integration with slug "%s" because it is already available via customer code.', esc_html( $this->slug ) ),
+				E_USER_WARNING
+			);
+			return;
+		}
+
 		// Don't do anything if integration is already activated.
 		if ( $this->is_active() ) {
 			trigger_error( sprintf( 'VIP Integration with slug "%s" is already activated.', esc_html( $this->get_slug() ) ), E_USER_WARNING ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_trigger_error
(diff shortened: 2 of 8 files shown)


integrations/block-data-api.php
```diff
@@ -22,6 +22,14 @@ class BlockDataApiIntegration extends Integration {
 	 */
 	protected string $version = '1.0';
 
+	/**
+	 * Returns `true` if `Block Data API` is already available via customer code. We will use
+	 * this function to prevent activating of integration from platform side.
+	 */
+	public function is_integration_already_available_via_customer(): bool {
+		return defined( 'VIP_BLOCK_DATA_API_LOADED' );
+	}
```
mehmoodak, line 31: Added `is_integration_already_available_via_customer` function in our base Integration class to make sure that it get defined by each integration and apart from that we can reuse this check if its a function.

integrations/block-data-api.php
```diff
@@ -44,4 +47,9 @@ public function load(): void {
 			}
 		} );
 	}
+
+	/**
+	 * Configure `Block Data API` for VIP Platform.
+	 */
+	public function configure_for_vip(): void {}
```
mehmoodak: Modify the integration base class to add `configure_for_vip` function which will only get called if the integration is enabled via VIP Config and configure the integration as needed by VIP. We could have done this via some simple if condition but this approach seems more declarative and easy to reuse.

integrations/integration.php
```diff
@@ -102,6 +109,14 @@ public function get_slug(): string {
 		return $this->slug;
```
acicovic, line 109: I haven't looked in depth, but this function seems a bit strange because it is supposed to be void but returns a slug. I also see that you removed the only void return that existed. I know this isn't your code, just flagging this in case something needs to be done here in the future.

tests/integrations/test-integration.php
```diff
@@ -44,6 +61,8 @@ public function test__calling_activate_twice_on_same_integration_does_not_activa
 
 		$integration->activate();
 		$integration->activate();
+
+		$this->assertFalse( $integration->is_active() );
```
acicovic, line 65: If I understand correctly, this will be `false` due to the second `load()` call failing?

acicovic (approved):
I'm not an authority on the Block Data API, but the code looks good to me.

tests/integrations/test-integration.php
mehmoodak: Its false due to second `activate` I have just added it to make sure that after calling `trigger_error` we don't continue with the execution of the function.

acicovic: Right. Thanks!

integrations/integration.php
mehmoodak: I think you are mistaken by collapsing of code because `get_slug` function returns `string` not `void`, please have a look again and let me know if you still think there is any issue with the code.

acicovic: Oh, right. Sorry for that! 😁

integrations/block-data-api.php
mjangda: Can we just call this `is_loaded()` since there's nothing in the method that's actually checking if it's loaded by a customer or not.

integrations/block-data-api.php
```diff
@@ -30,11 +38,6 @@ class BlockDataApiIntegration extends Integration {
 	public function load(): void {
 		// Wait until plugins_loaded to give precedence to the plugin in the customer repo.
 		add_action( 'plugins_loaded', function() {
-			// Do not load plugin if already loaded by customer code.
```
mjangda, line 33: I wonder if we should keep the `is_loaded` check here just as a safety check?

integrations/integration.php
```diff
@@ -102,6 +109,14 @@ public function get_slug(): string {
...
 	}
 
+	/**
+	 * Returns `true` if the integration is already available via customer code. We will use
+	 * this function to prevent activating of integration from platform side.
+	 *
+	 * @private
+	 */
+	abstract public function is_integration_already_available_via_customer(): bool;
```
mjangda:
```suggestion
	abstract public function is_loaded(): bool;
```

integrations/parsely.php
```diff
@@ -13,24 +13,37 @@
...
+		return class_exists( 'Parsely' ) || class_exists( 'Parsely\Parsely' );
+	}
+
 	/**
 	 * Loads the plugin.
 	 *
 	 * @private
 	 */
 	public function load(): void {
-		// Do not load plugin if already loaded by customer code.
```
mjangda, line 22: I wonder if we should keep the `is_loaded` check here just as a safety check?

integrations/integration.php
```diff
@@ -112,4 +127,14 @@ public function get_slug(): string {
...
+
+	/**
+	 * Configure the integration for VIP platform.
+	 *
+	 * If we want to implement functionality only if the integration is enabled via VIP
+	 * then we will use this function.
+	 *
+	 * @private
+	 */
+	abstract public function configure_for_vip(): void;
```
mjangda: Can we just call this `configure` to keep things simple? Even though the intention of the method is to load the platform-level configs, we're only creating this for internal use.

```suggestion
	abstract public function configure(): void;
```

mjangda (commented):
Had some suggestions around naming and hardening; otherwise this looks good.

integrations/integration.php
mehmoodak: Done in https://github.com/Automattic/vip-go-mu-plugins/pull/4839/commits/e8c247c88c604fb36ca47228a423f97b9cfdb1cb

integrations/integration.php
mehmoodak: Done in https://github.com/Automattic/vip-go-mu-plugins/pull/4839/commits/e8c247c88c604fb36ca47228a423f97b9cfdb1cb

integrations/block-data-api.php
mehmoodak: After this refactoring now our integrations must provide `is_loaded` function and based on this we will call the `load` function and there is also a [test](https://github.com/Automattic/vip-go-mu-plugins/blob/e8c247c88c604fb36ca47228a423f97b9cfdb1cb/tests/integrations/test-integration.php#L40C2-L54C3) present as safeguard.

So I think keeping current implementation is better because its cleaner.

integrations/parsely.php
mehmoodak: Same as above.

integrations/integration.php
```diff
@@ -63,10 +63,17 @@ public function __construct( string $slug ) {
 	 * @private
 	 */
 	public function activate( array $options = [] ): void {
+		// If integration is already available in customer code then don't activate it from platform side.
+		if ( $this->is_loaded() ) {
+			trigger_error( // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_trigger_error
+				sprintf( 'Prevented activating of integration with slug "%s" because it is already available via customer code.', esc_html( $this->slug ) ),
```
mjangda:
```suggestion
				sprintf( 'Prevented activating of integration with slug "%s" because it is already loaded.', esc_html( $this->slug ) ),
```

integrations/block-data-api.php
```diff
@@ -30,11 +38,6 @@ class BlockDataApiIntegration extends Integration {
...
-			if ( defined( 'VIP_BLOCK_DATA_API_LOADED' ) ) {
-				return;
-			}
-
 			// Load the version of the plugin that should be set to the latest version, otherwise if it's not found deactivate the integration.
```
mjangda: Even though it feels duplicative, the concern would be if we (or other code) unintentionally called `load()` multiple times, we could end up with problems.

```suggestion
			if ( $this->is_loaded() ) {
				return;
			}

			// Load the version of the plugin that should be set to the latest version, otherwise if it's not found deactivate the integration.
```

integrations/parsely.php
```diff
@@ -13,24 +13,37 @@
...
-		if ( class_exists( 'Parsely\Parsely' ) ) {
-			return;
-		}
-
 		// Activating the integration via setting constant whose implementation is already
```
mjangda, line 38:
```suggestion
		if ( $this->is_loaded() ) {
			return;
		}

		// Activating the integration via setting constant whose implementation is already
```

integrations/block-data-api.php
mjangda: See new comment with proposed fix.

integrations/parsely.php
mjangda: See new comment.

mjangda (approved):
Some small suggestions after which this is good to go 🚀

integrations/block-data-api.php
mehmoodak: Okay should we do same for `configure` because that can also be called directly?

mehmoodak: I think we should avoid duplication so removing this `is_loaded` from integration code and adding it in each function where needed.

mehmoodak: Ignore the above comment ... it makes sense to keep `is_loaded()` inside `activate` function of base integration class.

mehmoodak: Added in https://github.com/Automattic/vip-go-mu-plugins/pull/4839/commits/4bc3fdeda3547e91dcbc6c8656666e72049b7e2f

integrations/parsely.php
mehmoodak: Added in https://github.com/Automattic/vip-go-mu-plugins/pull/4839/commits/4bc3fdeda3547e91dcbc6c8656666e72049b7e2f

integrations/integration.php
mehmoodak: Addressed in https://github.com/Automattic/vip-go-mu-plugins/pull/4839/commits/4bc3fdeda3547e91dcbc6c8656666e72049b7e2f

PR #4845, opened 2023-09-06
FORNO-1668: Add Prometheus user stats collector

t-wright:
## Description

Adds a Prometheus collector to count users by role and 2FA status. Initially limited to `develop` and `staging` environments.

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

1. Check out PR
1. Enable the `prom-user-collection` feature for your test environment (e.g. by adding `local` [here](https://github.com/Automattic/vip-go-mu-plugins/blob/e324084b849ceae193f45fe6a1e383ad21c3f3b4/lib/feature/class-feature.php#L60-L63))
1. Run the cron to ensure metrics have been processed `wp cron event run vip_aggregated_cron_hourly`
1. Visit the site to trigger collection
2. Vist `<siteurl>/.vip-prom-metrics` to ensure the metrics are present


prometheus-collectors/class-user-stats-collector.php
```diff
@@ -0,0 +1,154 @@
...
+	private string $blog_id;
+	private const MAX_USERS_FOR_COUNT     = 1_000_000;
+	private const MAX_USERS_FOR_2FA_STATS = 5_000;
+	private const METRIC_OPTION           = 'vip-prom-users';
+	public const TFA_STATUS_ENABLED       = 'enabled';
+	public const TFA_STATUS_DISABLED      = 'disabled';
+	public const TFA_STATUS_DEFAULT       = 'unknown';
+
+	public function initialize( RegistryInterface $registry ): void {
+		$this->blog_id = Plugin::get_instance()->get_site_label();
```
saroshaga, line 23: Isn't there a chance that `get_site_label()` [returns](https://github.com/Automattic/vip-go-mu-plugins/blob/e5550234092366e8e17c9dbca5a843d6778ab525/prometheus/inc/class-plugin.php#L165) a string?

prometheus-collectors/class-user-stats-collector.php
```diff
@@ -0,0 +1,154 @@
...
+		}
+
+		// Since we bunch everything together under a single label this logic wouldn't make any sense.
+		if ( is_multisite() && wp_count_sites()['all'] > Plugin::MAX_NETWORK_SITES ) {
+			return;
+		}
+
+		// count_users() is expensive so skip collecting this metric if there are too many users.
+		// https://developer.wordpress.org/reference/functions/count_users/
+		if ( get_user_count() > self::MAX_USERS_FOR_COUNT ) {
```
saroshaga, line 61: So the expense comes from the `avail_roles` calculation in the `count_users()` function? Is that why we're using the `get_user_count()` ?

prometheus-collectors/class-user-stats-collector.php
```diff
@@ -0,0 +1,154 @@
...
+	public const TFA_STATUS_DEFAULT       = 'unknown';
+
+	public function initialize( RegistryInterface $registry ): void {
+		$this->blog_id = Plugin::get_instance()->get_site_label();
+
+		$this->users_gauge = $registry->getOrRegisterGauge(
+			'user',
+			'count',
+			'Number of users by role and 2FA status',
+			[ 'site_id', 'role', 'tfa_status' ]
```
saroshaga: What's the site_id here? is it the network site or the env ID?

sjinks: Network site

sjinks: `get_site_label()` takes care of it.

saroshaga: Maybe we should name it such then?

prometheus-collectors/class-user-stats-collector.php
sjinks: `get_site_label()` always returns a string. But `blog_id` is a [string](https://github.com/Automattic/vip-go-mu-plugins/pull/4845/files#diff-9781da76b2e5bc7227af83167daeb047143f8175ee81e8eae54d0a2dd794dc86R14) either.

prometheus-collectors/class-user-stats-collector.php
t-wright: You mean call it something like `network_site_id`? That makes sense, though `site_id` is the label name used by a number of the other existing collectors so perhaps better to follow suit.

prometheus-collectors/class-user-stats-collector.php
t-wright: Yep, `get_user_count()` is a cached count of all users on the network so we can avoid collecting this metric altogether if it has an huge number of users.

prometheus-collectors/class-user-stats-collector.php
saroshaga: > That makes sense, though site_id is the label name used by a number of the other existing collectors so perhaps better to follow suit.

for network sites?? Hmm.. I don't prefer that because it is easy to confuse with GOOP site IDs which are also used a lot on Prom. @sjinks what do you feel?

prometheus-collectors/class-user-stats-collector.php
saroshaga: Sorry I should have been clearer... looking at the code I linked above, isn't there a chance that the method returns the string `network` ?

prometheus-collectors/class-user-stats-collector.php
mjangda: We should try to use the newer terminology wherever possible (i.e. `network_site_id`), including updating the existing the collectors.

t-wright: > I don't prefer that because it is easy to confuse with GOOP site IDs which are also used a lot on Prom

I think the GOOP site IDs are mostly limited to the namespace label in Prom? But I see what you mean - would definitely be less ambiguous. I'll update here and can open a PR to switch the others.

prometheus-collectors/class-user-stats-collector.php
t-wright: Yep, that is possible. We don't collect the metric in that scenario: https://github.com/Automattic/vip-go-mu-plugins/pull/4845/files#diff-9781da76b2e5bc7227af83167daeb047143f8175ee81e8eae54d0a2dd794dc86R54-R57

sjinks (approved):

sjinks (changes requested):
There are a couple of complaints from phpcs:

https://github.com/Automattic/vip-go-mu-plugins/actions/runs/6261490517/job/17001686419

But other than that, LGTM

t-wright:
> There are a couple of complaints from phpcs

Thanks! Have fixed those now.

sjinks (approved):

PR #4857, opened 2023-09-12
Fixes fatal error while reading config file of the integration

mehmoodak:
## Description

When we update integration via VIP Internal CLI then first it update the k8s configmap and then after few seconds changes are reflected in our config file, this update under the hood doesn't replace the file but pointed the file to the new symlink (this is how configmap works in k8s) but PHP uses file status cache which can become stale if symlink is updated.

In our particular case fatal error was occuring because `is_readable()` check was returning `true` because of cache but the file itself was not there i.e. symlink got updated.

## Changes Made

- Uses `clearstatcache` to clear the cache and make sure we read from latest available config file.

## Changelog Description

Fixes fatal error while reading config file of the integration

## Pre-review checklist

## Steps to Test

No change in functionality, just fixing the stale cache issue.

### DIFF
--- a/integrations/integration-vip-config.php
+++ b/integrations/integration-vip-config.php
@@ -78,25 +78,29 @@ class IntegrationVipConfig {
 	 * @return null|mixed
 	 */
 	protected function get_vip_config_from_file( string $slug ) {
-		$config_file_path = ABSPATH . 'config/integrations-config/' . $slug . '-config.php';
+		$config_file_directory = ABSPATH . 'config/integrations-config';
+		$config_file_name      = $slug . '-config.php';
+		$config_file_path      = $config_file_directory . '/' . $config_file_name;
 
 		/**
-		 * Clear realpath cache to get new path else we can get cached path which can be non-existant
-		 * because k8s configmaps updates the file via symlink instead of actually replacing the file and
-		 * PHP cache can hold a reference to the old symlink.
+		 * Clear cache to always read data from latest config file.
 		 *
-		 * Did tried using `clearstatcache( true, $config_file_path )` but it was giving old symlink,
-		 * `is_readable()` does return correct value (false) for this case but we prefer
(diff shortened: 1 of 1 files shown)


integrations/integration-vip-config.php
```diff
@@ -80,11 +80,23 @@ private function set_config( string $slug ): void {
...
+		 */
+		clearstatcache( true );
+		$config_real_path = realpath( $config_file_path );
+
+		if ( ! is_readable( $config_real_path ) ) {
 			return null;
 		}
 
-		return require_once $config_file_path;
+		return require $config_real_path;
```
mehmoodak: Using require because `require_once` returns `true` if the file is already loaded but here we always want content of the file.

integrations/integration-vip-config.php
```diff
@@ -80,11 +80,23 @@ private function set_config( string $slug ): void {
...
+		/**
+		 * Clear realpath cache to get new path else we can get cached path which can be non-existant
+		 * because k8s configmaps updates the file via symlink instead of actually replacing the file and
+		 * PHP cache can hold a reference to the old symlink.
+		 *
+		 * Did tried using `clearstatcache( true, $config_file_path )` but it was giving old symlink,
+		 * `is_readable()` does return correct value (false) for this case but we prefer
+		 * to get the config which `clearstatcache( true )` does return because it gives us new symlink.
+		 */
+		clearstatcache( true );
```
mehmoodak: Clearing the whole file status cache doesn't seems like a good idea but none of the other solution were working like I have also mentioned in comment so please let me know if there is something else I should try.

mjangda:
Related: https://pracucci.com/php-realpath-cache-and-kubernetes-secrets-configmap-updates.html

integrations/integration-vip-config.php
mehmoodak: Added changes to clear cache of a specific file.

WPprodigy (approved):

PR #4868, opened 2023-09-15
PIE-3749 Adds support for temporary redirects to resolve Jetpack IDC

andrea-sdl:
Explore a new site option to have temporary redirects in place to resolve Jetpack IDC.


000-vip-init.php
```diff
@@ -303,6 +304,11 @@
 
 	$encloseme_cleaner = new VIP_Encloseme_Cleanup();
 	$encloseme_cleaner->init();
+
+	// redirects url for jetpack network launches, if needed
+	require_once __DIR__ . '/config/class-vip-jetpack-network-launch-redirects.php';
+
+	add_action( 'init', [ VIP_Jetpack_Network_Launch_Redirects::class, 'maybe_redirect_jetpack_network_launches' ] );
```
WPprodigy: For safety during the initial rollout, will need to have this look more like:

```
if ( file_exists( __DIR__ . '/config/class-vip-jetpack-network-launch-redirects.php' ) ) {
	require_once __DIR__ . '/config/class-vip-jetpack-network-launch-redirects.php';
	add_action( 'init', [ VIP_Jetpack_Network_Launch_Redirects::class, 'maybe_redirect_jetpack_network_launches' ] );
}
```

Can remove the file_exists() on a subsequent deploy.

Would also prefer to move this functionality into more JP-specific area of MU. Let's put it into `vip-jetpack`.

config/class-vip-jetpack-network-launch-redirects.php
```diff
@@ -0,0 +1,92 @@
...
+	const REDIRECT_TTL_MINUTES               = 60;
+	const LOG_FEATURE_NAME                   = 'vip_network_launch_redirects';
+	// This is used in front of the source domain to avoid the S&R to replace it
+	const URL_REPLACE_PREFIX = '##';
+
+	public static function maybe_redirect_jetpack_network_launches() {
+		// applies redirects only in the frontend and for multisites
+		if ( ! is_multisite() || is_admin() || ! Context::is_web_request() ) {
+			return;
+		}
```
WPprodigy: `Context::is_web_request()` will fail for xmlrpc requests, which I think it what we're also wanting to target here for JP requests. So may need to get a bit more specific here on what we want to early return on.

WPprodigy (commented):
One thing I'm unsure about is making sure the request ends up routed to the right subsite in the first place.

For example, a request for a specific domain comes to our routing and we decide with application has that domain mapped to it. Then WP gets the request and decides which subsite it maps to based on the global tables. Then the subsite is loaded up, at which point you'd have access to reading this new `vip_network_temp_redirects` option.

But you have to first get WP to route the request towards the correct subsite. That's probably the first problem to solve for, hooking into that routing process and finding a way to interject these custom/temporary redirects at that point. So we might not be looking at a per-subsite option but rather something more global.

andrea-sdl:
@WPprodigy closing this in favor of exploring other options to resolve the IDC in a more performant/less invasive way.

PR #4861, opened 2023-09-14
Don't skip wpcom_vip_load_plugin for wp-activate.php under subdirectories

tangrufus:
## Description

#4247 introduced a regression that resulted in `client-mu-plugins` not being loaded in `wp-activate.php`.

#4328 addressed the regression by explicitly checking whether current page is `/wp-activate.php` or not.

This pull request continues #4328's spirit and handle `wp-activate.php` under subdirectories, e.g: `https://example.com/foo/wp-activate.php`

## Pre-review checklist

Please make sure the items below have been covered before requesting a review:

- [x] This change works and has been tested locally (or has an appropriate fallback).
- [ ] This change works and has been tested on a Go sandbox.
- [ ] This change has relevant unit tests (if applicable).
- [ ] This change uses a rollout method to ease with deployment (if applicable - especially for large scale actions that require writes).
- [ ] This change has relevant documentation additions / updates (if applicable).
- [ ] I've created a changelog description that aligns with the provided examples.

## Steps to Test

1. Check out the PR
1. Add a dummy plugin under `client-mu-plugins` to make sure it's loaded during the further testing steps
1. On a multisite with **subdirectory** setup and open registration, go to register page `https://example.com/foo/wp-login.php?action=register`
1. Create a new account
1. In the verification email, click on the link to `wp-activate.php`, e.g. `https://example.com/foo/wp-activate.php?key=123456789` - make sure the dummy plugin is loaded
1. Also, visit a link directly at `/wp-activate.php` and make sure the plugin is loaded

---

Props to @gmazzap

PR #4869, opened 2023-09-18
fix(vip-helpers): load mu-plugins for wp-activate.php in subdireсtories

sjinks:
## Description

Supersedes: #4861

Load mu-plugins for `wp-activate.php` located in a subdirectory (or for a subdirectory-based multisite installation).

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

This change has unit tests.

### DIFF
--- a/vip-helpers/vip-utils.php
+++ b/vip-helpers/vip-utils.php
@@ -1283,17 +1283,17 @@ function wpcom_vip_can_use_shared_plugin( $plugin ) {
 /**
  * Helper function to check if we can load plugins or not.
  */
-function wpcom_vip_should_load_plugins() {
+function wpcom_vip_should_load_plugins( $_reset = false ) {
 	static $should_load_plugins;
 
-	if ( isset( $should_load_plugins ) ) {
+	if ( ! $_reset && isset( $should_load_plugins ) ) {
 		return $should_load_plugins;
 	}
 
 	$should_load_plugins = true;
 
 	// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
-	if ( wp_installing() && strtok( $_SERVER['REQUEST_URI'] ?? '', '?' ) !== '/wp-activate.php' ) {
+	if ( wp_installing() && ! str_ends_with( strtok( $_SERVER['REQUEST_URI'] ?? '', '?' ), '/wp-activate.php' ) ) {
 		$should_load_plugins = false;
 	} elseif ( defined( 'WP_CLI' ) && WP_CLI ) {
 		// WP-CLI loaded with --skip-plugins flag
--- a/tests/vip-helpers/vip-utils/test-wpcom-vip-should-load-plugins.php
+++ b/tests/vip-helpers/vip-utils/test-wpcom-vip-should-load-plugins.php
@@ -1,0 +1,40 @@
+<?php
+
+class WPCOM_VIP_Should_Load_Plugins_Test extends WP_UnitTestCase {
+	public function setUp(): void {
+		parent::setUp();
+		wp_installing( true );
+	}
+
+	public function tearDown(): void {
+		wp_installing( false );
+		parent::tearDown();
+	}
+
+	/**
+	 * @dataProvider data_load_in_wp_activate
+	 */
+	public function test_load_in_wp_activate( string $request_uri, bool $expected ): void {
+		// phpcs:ignore WordPress.Security.ValidatedSanitizedInput
+		$saved = $_SERVER['REQUEST_URI'] ?? null;
+		try {
+			$_SERVER['REQUEST_URI'] = $request_uri;
+			$actual                 = wpcom_vip_should_load_plugins( true );
+
+			self::assertSame( $expected, $actual );
+		} finally {
+			$_SERVER['REQUEST_URI'] = $saved;
+		}
+	}
+
+	public function data_load_in_wp_activate(): iterable {
+		return [
+			[ '/wp-activate.php', true ],
+			[ '/wp-activate.php?something', true ],
+			[ '/wp-activate.php/path', false ],
+			[ '/?/wp-activate.php', false ],
+			[ '/dir/wp-activate.php', true ],
+			[ '/dir/subdir/wp-activate.php', true ],
+		];
+	}
+}


vip-helpers/vip-utils.php
```diff
@@ -1283,17 +1283,17 @@ function wpcom_vip_can_use_shared_plugin( $plugin ) {
...
-	if ( isset( $should_load_plugins ) ) {
+	if ( ! $_reset && isset( $should_load_plugins ) ) {
 		return $should_load_plugins;
 	}
 
 	$should_load_plugins = true;
 
 	// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
-	if ( wp_installing() && strtok( $_SERVER['REQUEST_URI'] ?? '', '?' ) !== '/wp-activate.php' ) {
+	if ( wp_installing() && ! str_ends_with( strtok( $_SERVER['REQUEST_URI'] ?? '', '?' ), '/wp-activate.php' ) ) {
```
gmazzap, line 1296: Why not using `basename()` which basically exists for cases like this and allows you to replace two function calls (`strtok` + ` str_ends_with`) with one?

Besides, can you guarantee that you'll never use this utility function before WordPress loads `compat.php` which defines `str_ends_with()` for PHP < 8.0? 

It would not be better to keep such utilities callable as early as possible and rely on "bare PHP" as much as possible?

For example, inside an `advanced-cache.php` dropin file, on PHP < 8.0, `str_ends_with` could not be defined, because `compat.php` is loaded _after_ that file [up to the pretty recent WP 6.2.2](https://github.com/WordPress/WordPress/blob/9e0f2faa28f1aa04a69e0b4eaa410a38adcd2e1e/wp-settings.php#L95-L107).

Moreover, the shim for that PHP 8.0 function was added in [WP 5.9](https://github.com/WordPress/WordPress/blob/90a29fbc4698e2959c7c35de9f4327fd98717bf7/wp-includes/compat.php#L474), are you absolutely sure you have no WP installation with a version older that that? I mean, we are talking here about a fatal error that can turn a producion site down.

Considering the alternative to avoid `str_ends_with()` is actually _less code_, and likely _more perfomant_ (because 1 VS 2 function calls), what's the reason of preferring it?

sjinks: @gmazzap I wish it were that easy. `REQUEST_URI` may contain a query string; `basename` will happily accept it. The original PR suggested to use [`parse_url()`](https://github.com/Automattic/vip-go-mu-plugins/pull/4861/files#diff-e3e0b5e7ec15b54034b133dc38c47759865fe8ef51de53b2cca48e81d2217cedR1295); it is much slower than `strtok()`.

Once we get rid of the query string, we have `basename()` + string comparison vs `str_ends_with()`

> Besides, can you guarantee that you'll never use this utility function before WordPress loads compat.php which defines str_ends_with() for PHP < 8.0?

Of course. We define `str_ends_with()` in `lib/helpers/php-compat.php`, which is loaded from `000-vip-init.php`. `lib/helpers/php-compat.php` is loaded before `vip-helpers/vip-utils.php`:

```php
// Polyfills
require_once __DIR__ . '/lib/helpers/php-compat.php';

// Load our development and environment helpers
require_once __DIR__ . '/vip-helpers/vip-notoptions-mitigation.php';
require_once __DIR__ . '/vip-helpers/vip-utils.php';
```

So when it comes to `wpcom_vip_should_load_plugins()`, we do have `str_ends_with()` defined.

> For example, inside an advanced-cache.php dropin file

`wpcom_vip_should_load_plugins()` won't be available to drop-ins because they are loaded prior to mu-plugins. And if/when `wpcom_vip_should_load_plugins()` is available, `str_ends_with()` is also available.

> Moreover, the shim for that PHP 8.0 function was added in [WP 5.9](https://github.com/WordPress/WordPress/blob/90a29fbc4698e2959c7c35de9f4327fd98717bf7/wp-includes/compat.php#L474), are you absolutely sure you have no WP installation with a version older that that?

Yes. WP 5.9 is the oldest supported version. Moreover, we define `str_ends_with()` even if WordPress does not.

> Considering the alternative to avoid str_ends_with() is actually less code, and likely more perfomant (because 1 VS 2 function calls), what's the reason of preferring it?

I would agree if it were `basename()` vs `strtok()` + `str_ends_with()`, but `basename()` alone unfortunately does not work. It needs either `strtok()` or `(wp_)parse_url()`.

And it turns out that `basename()` + string comparison is not faster than `str_ends_with()`: `basename()` has to scan for the directory separator while `str_ends_with()` already knows the length of the pattern. Then, you will have to compare the returned result with the pattern, and `str_ends_with()` already does that. For PHP 8+, this happens in the C code; this will be a bit faster than comparison in the userland.

And because this code runs on every request, we try to squeeze the tiniest bits of performance.

WPprodigy (approved):

vip-helpers/vip-utils.php
gmazzap: Thanks @sjinks I did not know that you have yet another polyfill for PHP on top of what WP has.

I'm always a bit vary of using pure strings manipulation for URLs instead of functions designed for URLs. You know, URLs are strings, but not _just_ strings.
(For example, an URL like `x/y/wp-activate.php/` which is a valid URL would for `wp-activate.php`does _not_ ends with `wp-activate.php`)

Anyway, as long as this works, fine by me.

gmazzap: Also, regarding

> wpcom_vip_should_load_plugins() won't be available to drop-ins because they are loaded prior to mu-plugins. 

Do I remember wrongly or the "lib" folder is loaded prior to mu-plugins as well? This is unrelevant for this topic because of your `php-compat.php`, more a curiosity on my end.

sjinks: > Do I remember wrongly or the "lib" folder is loaded prior to mu-plugins as well?

In `wp-config.php`:

```php
// Dependencies that should be loaded prior to VIP Config (custom error handler, Environment helpers, VIP_Request_Block)
if ( file_exists( __DIR__ . '/wp-content/mu-plugins/000-pre-vip-config/requires.php' ) ) {
        require_once __DIR__ . '/wp-content/mu-plugins/000-pre-vip-config/requires.php';
}

require( dirname( __FILE__ ) . '/config/wp-config.php' );
```

[`000-pre-vip-config/requires.php`](https://github.com/Automattic/vip-go-mu-plugins/blob/develop/000-pre-vip-config/requires.php) loads the following files:
```php
$files = [
	'/lib/wpcom-error-handler/wpcom-error-handler.php',
	'/lib/class-vip-request-block.php',
	'/lib/environment/class-environment.php',
	'/lib/helpers/environment.php',
	'/lib/utils/class-context.php',
];

$cli_files = [
	'/lib/helpers/wp-cli-db.php',
];
```

(all paths are relative to `wp-content/mu-plugins`).

PR #4882, opened 2023-09-21
Fixes warning while reading integration config for PHP 8+

### LINKED ISSUE #4878: `file_get_contents()` repeatedly called on not existing file
## Expected/Desired Behavior

`file_get_contents` at [this](https://github.com/Automattic/vip-go-mu-plugins/blob/9cea8f3ed04e0130f1470013745ef600fb8d08b4/integrations/integration-vip-config.php#L92) line should stop to produce Warnings if the file checked doesn't exist (whatever PHP version is being used).

## Actual Behavior

Right now, that line produces a warning in PHP 8+ if the file doesn't exists. This is probably caused by the fact the the Error Control Operator (`@`) changed its behavior from PHP 8.

Specifically the errors are triggered for both Parsely and Block Data Api integrations:

`file_get_contents(/var/www/config/integrations-config/parsely-config.php): Failed to open stream: No such file or directory`
`file_get_contents(/var/www/config/integrations-config/block-data-api-config.php): Failed to open stream: No such file or directory`

Just loading any page in an environment where Warnings are visible should be enough (Probably also having the integrations disabled is required?).

## Additional notes

Is invalidating the cache of the config file before checking for its existence a doable solution for you? This way `file_get_content` can be just avoided.

Otherwise is there any chance to not reach that part of code at all if the integrations are not active? So the part would be just skipped for whose are not supposed to use it.

mehmoodak:
## Description

In VIP Integrations we used `@file_get_contents` to read config file of each integration. We expected that `@` will suppress all warnings but for PHP 8+ it is still producing a warning if file doesn't exist.

Closes: https://github.com/Automattic/vip-go-mu-plugins/issues/4878

## Changelog Description

Fixes warning while reading integration config file for PHP 8+

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

1. Check out PR.
1. If needed setup new environment via `vip dev-env create`
1. Open dev site on browser
1. Use `vip dev-env logs --slug vip-single-site -S php -d -f` to verify that no warning is generated while reloading or navigating dev site.

### DIFF
--- a/integrations/integration-vip-config.php
+++ b/integrations/integration-vip-config.php
@@ -83,29 +83,16 @@ class IntegrationVipConfig {
 		$config_file_path      = $config_file_directory . '/' . $config_file_name;
 
 		/**
-		 * Temporarily disable warnings by clearing the E_WARNING bit.
-		 *
-		 * Reading config file via `file_get_contents()` even with `@` operator generate warnings on PHP 8+.
-		 */
-		$current_error_reporting = error_reporting(); // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.runtime_configuration_error_reporting
-		error_reporting( $current_error_reporting & ~E_WARNING ); // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.runtime_configuration_error_reporting
-
-		/**
 		 * Clear cache to always read data from latest config file.
 		 *
 		 * Kubernetes ConfigMap updates the file via symlink instead of actually replacing the file and
 		 * PHP cache can hold a reference to the old symlink that can cause fatal if we use require
 		 * on it.
 		 */
-		if ( false === file_get_contents( $config_file_path ) ) {
-			clearstatcache( true, $config_file_directory . '/' . $config_file_name );
-			// Clears cache for files created by k8s ConfigMap.
-			clearstatcache( true, $config_file_directory . '/..data' );
-			clearstatcache( true, $config_file_directory . '/..data/' . $config_file_name );
-		}
-
-		// Restore the original error reporting.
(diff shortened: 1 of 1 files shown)


integrations/integration-vip-config.php
```diff
@@ -82,20 +82,31 @@ protected function get_vip_config_from_file( string $slug ) {
...
+
 		/**
 		 * Clear cache to always read data from latest config file.
 		 *
 		 * Kubernetes ConfigMap updates the file via symlink instead of actually replacing the file and
 		 * PHP cache can hold a reference to the old symlink that can cause fatal if we use require
 		 * on it.
 		 */
-		if ( false === @file_get_contents( $config_file_path ) ) { // phpcs:ignore WordPress.PHP.NoSilencedErrors.Discouraged
+		if ( false === file_get_contents( $config_file_path ) ) {
```
mehmoodak: Another solution is to avoid this `if` statement and always read latest config from file.

Personally I am more inclined toward this solution because its simpler and always reading simple configs from file doesn't seems like an expensive operation.

Please do let me know about your thoughts on this solution.

sjinks (approved):

integrations/integration-vip-config.php
mjangda: > always read latest config from file

Do you mean just running `clearstatcache` and then `require`? If so, we should do a quick benchmark to see the impact before doing this.

mehmoodak: Did some quick benchmarking on local machine and here are the stats:

1. `clearstatcache()` with disable warning and if statement is on average taking 2.5 * E-4 seconds
1. `clearstatcache()` without if statement is on average taking 4.0 * E-5 seconds 


```php
$start_time = microtime(true);

// Code here

$end_time = microtime(true);
$execution_time = $end_time - $start_time;

echo "Execution Time: {$execution_time} seconds";
```

Based on these benchmarks removing the if statement and making the code simpler.

PR #4890, opened 2023-09-22
CANTINA-944: For private sites, force disable blog_public UI toggle buttons in Reading Settings to avoid confusion

rebeccahum:
## Description
For private sites, force disable blog_public UI toggle buttons in Reading Settings to avoid confusion, as we already force it to -1 in https://github.com/Automattic/vip-go-mu-plugins/blame/c3be5f418a23223b89931d000528f5c312f55f09/security/class-private-sites.php#L57

Screenshot 2023-09-22 at 3 14 34 PM

When hooked onto `blog_privacy_selector`:

image

### Plugin Updated: Private Sites

Force disable blog_public UI toggle buttons in Reading Settings to avoid confusion

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test
1) If constant not set already, set it `define( 'VIP_JETPACK_IS_PRIVATE', true );` in vip-config
2) Go to wp-admin > Settings > Reading Settings and scroll down to ensure box is grayed out like in screenshot
3) Add `add_filter( 'blog_privacy_selector', '__return_true' );` and repeat step 2)

### DIFF
--- a/security/class-private-sites.php
+++ b/security/class-private-sites.php
@@ -85,12 +85,22 @@ class Private_Sites {
 				}
 			}
 
+			function updateDescription(message) {
+				var description = document.querySelector('tr.option-site-visibility p.description');
+				if (description) {
+					description.textContent = message;
+				}
+			}
+
 			var checkbox = 'tr.option-site-visibility input#blog_public[type=\"checkbox\"]';
+			var description = document.querySelector('tr.option-site-visibility p.description');
 			if (document.querySelector(checkbox)) {
 				disableButton(checkbox);
+				updateDescription('This option is disabled when the constant \"VIP_JETPACK_IS_PRIVATE\" is enabled.');
 			} else {
 				disableButton('tr.option-site-visibility input#blog-public[type=\"radio\"]');
 				disableButton('tr.option-site-visibility input#blog-norobots[type=\"radio\"]');
+				updateDescription('These options are disabled when \"VIP_JETPACK_IS_PRIVATE\" is enabled.');
 			}
 		});" );
 	}
--- a/security/class-private-sites.php
+++ b/security/class-private-sites.php
@@ -71,11 +71,27 @@ class Private_Sites {
 	}
 
 	/**
-	 * Force the blog_public option to be -1 and disable checkbox/radio UI in Reading Settings
+	 * Force the blog_public option to be -1 and disable UI
 	 */
 	public function force_blog_public_option() {
 		add_filter( 'blog_privacy_selector', '__return_true' );
 		add_filter( 'option_blog_public', array( $this, 'filter_restrict_blog_public' ) );
+		add_action( 'admin_enqueue_scripts', array( $this, 'disable_blog_public_ui' ) );
+	}
+
+	/**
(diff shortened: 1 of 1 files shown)


security/class-private-sites.php
```diff
@@ -71,6 +70,41 @@ public function disable_core_feeds() {
...
+					element.disabled = true;
+				}
+			}
+
+			function updateDescription(message) {
+				var description = document.querySelector('tr.option-site-visibility p.description');
+				if (description) {
+					description.textContent = message;
+				}
+			}
```
sjinks: Not sure if it makes sense now, but...

```javascript
function updateProperty(selector, property, value) {
    const element = document.querySelector(selector);
    if (element) {
        element[property] = value;
    }
}

const disableButton = (selector) => updateProperty(selector, 'disabled', true);
const updateDescription = (message) => updateProperty('tr.option-site-visibility p.description', textContent, message);
```

security/class-private-sites.php
```diff
@@ -71,6 +70,54 @@ public function disable_core_feeds() {
...
+
+			var checkbox = 'tr.option-site-visibility input#blog_public[type="checkbox"]';
+			if (document.querySelector(checkbox)) {
+				updateProperty(checkbox, 'disabled', true);
+			} else {
+				updateProperty('tr.option-site-visibility input#blog-public[type="radio"]', 'disabled', true);
+				updateProperty('tr.option-site-visibility input#blog-norobots[type="radio"]', 'disabled', true);
+			}
+			updateProperty('tr.option-site-visibility p.description', 'textContent', '%s');
+		});
```
sjinks:
```suggestion
		function onContentLoaded(callback) {
			if (document.readyState !== 'loading') {
				callback();
			} else {
				document.addEventListener('DOMContentLoaded', callback);
			}
		}

		onContentLoaded(function() {
			function updateProperty(selector, property, value) {
				const element = document.querySelector(selector);
				if (element) {
					element[property] = value;
				}
			}
			var checkbox = 'tr.option-site-visibility input#blog_public[type="checkbox"]';
			if (document.querySelector(checkbox)) {
				updateProperty(checkbox, 'disabled', true);
			} else {
				updateProperty('tr.option-site-visibility input#blog-public[type="radio"]', 'disabled', true);
				updateProperty('tr.option-site-visibility input#blog-norobots[type="radio"]', 'disabled', true);
			}
			updateProperty('tr.option-site-visibility p.description', 'textContent', '%s');
		});
```

sjinks: I don't know if our clients use Cloudflare or any kind of site load optimization tools, but if they do, `DOMContentLoaded` may not fire.

For example, Cloudflare's RocketLoader loads script on `DOMContentLoaded` event. This means that if you set another event listener for `DOMContentLoaded`, it is too late; it won't fire.

The solution here is to use [`document.readyState`](https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState). When its value is `loading`, setting our `DOMContentLoaded` handler is OK. If it is set to something else, this means that the `DOMContentLoaded` (`interactive`) or `load` (`complete`) event has already fired and we need to invoke our handler immediately.

WPprodigy (approved):

PR #4906, opened 2023-09-27
Search e2e Tests: Fix b0rked WooCommerce test

rebeccahum:
## Description
Looks like a WooCommerce update added some extra queries, so we need to take that into account in the tests and check the third one. Follows up on https://github.com/Automattic/vip-go-mu-plugins/pull/4729/files#r1279557202.


sjinks (approved):

sjinks:
Thank you!

PR #4907, opened 2023-09-28
Add the governance plugin

ingeniumed:
## Description

This is meant to add the [VIP governance plugin](https://github.com/Automattic/vip-governance-plugin/tree/trunk) as an integration in mu-plugins, alongside the Block Data API. It's using the existing VIP integrations framework that's already in place for the Block Data API and Parsely to do this.

I have also added 2 test for the Block Data API and the VIP Governance plugin that simply tries to load them, and then verifies if they are inactive. It also calls their `is_loaded` function which checks if their constant is defined which should be false. Unlike the parsely plugin, their code isn't there so the code that can be tested is limited unfortunately.

The corresponding PR in mu-plugins-ext has been made [here](https://github.com/Automattic/vip-go-mu-plugins-ext/pull/44). That should be merged first.

### Plugin Added: VIP Governance 1.0

We added VIP Governance Plugin 1.0 for VIP customers.

This is a plugin that adds additional governance capabilities to the block editor.

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

Example:

1. Check out PR, along with [this](https://github.com/Automattic/vip-go-mu-plugins-ext/pull/44) PR as well.
2. Use this as your mu-plugins in a local dev-env instance
3. Activate using the following:
```php
// client-mu-plugins/plugin-loader.php

\Automattic\VIP\Integrations\activate( 'vip-governance' );
```
4. Go to `wp-admin` > `Plugin` and ensure VIP Governance shows up
5. Go to `wp-admin` > `VIP Governance` and ensure that loads correctly
6. Go to the a new post page and ensure everything loads fine.

### DIFF
--- a/tests/integrations/test-vip-governance.php
+++ b/tests/integrations/test-vip-governance.php
@@ -22,25 +22,4 @@ class VIP_Governance_Integration_Test extends WP_UnitTestCase {
 
 		$this->assertFalse( $vip_governance_integration->is_active() );
 	}
-
-	public function test__load_call_returns_without_setting_constant_if_vip_governance_is_already_loaded(): void {
-		/**
-		 * Integration mock.
-		 *
-		 * @var MockObject|VipGovernanceIntegration
-		 */
-		$vip_governance_integration_mock = $this->getMockBuilder( VipGovernanceIntegration::class )->setConstructorArgs( [ 'vip-governance' ] )->setMethods( [ 'is_loaded' ] )->getMock();
-		$vip_governance_integration_mock->expects( $this->once() )->method( 'is_loaded' )->willReturn( true );
-		$preload_state = defined( 'VIP_GOVERNANCE_LOADED' );
-
-		$vip_governance_integration_mock->load();
-
-		$this->assertEquals( $preload_state, defined( 'VIP_GOVERNANCE_LOADED' ) );
-	}
-
-	public function test__configure_is_doing_nothing_for_configuration_on_vip_platform(): void {
-		$vip_governance_integration = new VipGovernanceIntegration( $this->slug );
-
-		$vip_governance_integration->configure();
-	}
 }
--- a/tests/integrations/test-vip-governance.php
+++ b/tests/integrations/test-vip-governance.php
@@ -21,4 +21,18 @@ class VIP_Governance_Integration_Test extends WP_UnitTestCase {
 
 		$this->assertFalse( $vip_governance_integration->is_active() );
 	}
+
+	public function test__if_is_loaded_gives_back_true_when_loaded(): void {
+		$vip_governance_integration = new VipGovernanceIntegration( $this->slug );
+
+		define( 'VIP_GOVERNANCE_LOADED', true );
+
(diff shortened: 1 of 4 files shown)


vip-integrations.php
```diff
@@ -14,16 +14,22 @@
 
 defined( 'ABSPATH' ) || die();
 
+// @codeCoverageIgnoreStart - the actual code here is tested individually in the unit tests.
```
ingeniumed, line 17: I ended up adding this here as this is where the registrations happen for the actual plugins. The underlying logic with the real plugins is better tested with the fake-integration that's in the unit tests. Instead, I added two tests for two of the integrations that didn't have unit tests and attempted to add what I could in there.

So for next time if an integration is added, what would be done is:

- Add the integration here
- Add the integration under integrations, and then add a test for it as well.

So that way as much of the new code is covered via the unit test.

Hope that makes sense, and makes it clear why I did this.

vip-integrations.php
```diff
@@ -14,16 +14,22 @@
...
 require_once __DIR__ . '/integrations/enums.php';
 require_once __DIR__ . '/integrations/integration-vip-config.php';
 require_once __DIR__ . '/integrations/block-data-api.php';
 require_once __DIR__ . '/integrations/parsely.php';
+require_once __DIR__ . '/integrations/vip-governance.php';
 
 // Register VIP integrations here.
 IntegrationsSingleton::instance()->register( new BlockDataApiIntegration( 'block-data-api' ) );
 IntegrationsSingleton::instance()->register( new ParselyIntegration( 'parsely' ) );
+IntegrationsSingleton::instance()->register( new VipGovernanceIntegration( 'vip-governance' ) );
```
WPprodigy: Due to deployments not being atomic, we need to safeguard this just for the initial deploy. Something like this:

```
if ( file_exists( __DIR__ . '/integrations/vip-governance.php' ) ) {
	require_once __DIR__ . '/integrations/vip-governance.php';
	IntegrationsSingleton::instance()->register( new VipGovernanceIntegration( 'vip-governance' ) );
}
```

Then can remove the next time the file is touched.

ingeniumed: I moved that line, and the initial require under the if check. Hopefully that should safeguard the code for the initial deploy. Lemme know if that doesn't look right.

WPprodigy (approved):

vip-integrations.php
WPprodigy: Looks good now!

tests/integrations/test-vip-governance.php
```diff
@@ -0,0 +1,30 @@
...
+
+	public function test__load_call_returns_inactive_because_no_governance_plugin_loaded(): void {
+		$vip_governance_integration = new VipGovernanceIntegration( $this->slug );
+
+		$vip_governance_integration->load();
+
+		$this->assertFalse( $vip_governance_integration->is_active() );
+	}
+
+	public function test__if_is_loaded_gives_back_true_when_loaded(): void {
```
mehmoodak, line 25: It seems like the function name is not matching with assert statement.

PR #4910, opened 2023-09-29
CANTINA-958: Add better filtering in `wp vip two-factor report` CLI command

rebeccahum:
## Description

Adds filtering by whether 2fa is enabled, 2fa provider, user role, and user.

Example usages:
	 *      wp vip two-factor report
	 *      wp vip two-factor report --2fa-enabled=true
	 *      wp vip two-factor report --2fa-provider=email
	 *      wp vip two-factor report --role=administrator
	 *      wp vip two-factor report --role=administrator --2fa-enabled=false
	 *      wp vip two-factor report --role=administrator --2fa-enabled=true --2fa-provider=email
	 *      wp vip two-factor report --user_login=wpvip
	
### WP-CLI Command Updated: wp vip two-factor report

Add filtering by whether 2fa is enabled, 2fa provider, user role, and user.

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test
1) Create users with different roles, administrative users with 2FA and administrative users without 2FA
2) Do the examples in description and ensure it returns the expected results

### DIFF
--- a/wp-cli/vip-two-factor.php
+++ b/wp-cli/vip-two-factor.php
@@ -30,7 +30,6 @@ class VIP_Two_Factor_Command extends WPCOM_VIP_CLI_Command {
 	 *      wp vip two-factor report --role=administrator --2fa-enabled=true --2fa-provider=email
 	 *      wp vip two-factor report --user_login=wpvip
 	 *
-	 * @subcommand report
 	 * @synopsis [--2fa-enabled=<true|false>] [--role=<string>] [--2fa-provider=<string>] [--user_login=<login|id|email>]
 	 */
 	public function report( $args, $assoc_args ) {
--- a/wp-cli/vip-two-factor.php
+++ b/wp-cli/vip-two-factor.php
@@ -22,13 +22,26 @@ class VIP_Two_Factor_Command extends WPCOM_VIP_CLI_Command {
 	 *
 	 * ## EXAMPLES
 	 *
-	 *      wp vip two-factor report
-	 *      wp vip two-factor report --2fa-enabled=true
-	 *      wp vip two-factor report --2fa-provider=email
-	 *      wp vip two-factor report --role=administrator
-	 *      wp vip two-factor report --role=administrator --2fa-enabled=false
-	 *      wp vip two-factor report --role=administrator --2fa-enabled=true --2fa-provider=email
-	 *      wp vip two-factor report --user_login=wpvip
+	 *     # List 2FA status for all users.
+	 *     $ wp vip two-factor report
+	 *
+	 *     # List users with 2FA enabled. 
+	 *     $ wp vip two-factor report --2fa-enabled=true
+	 *
+	 *     # List users who use email as their authentication factor. 
+	 *     $ wp vip two-factor report --2fa-provider=email
+	 *
+	 *     # List 2FA status for administrators. 
+	 *     $ wp vip two-factor report --role=administrator
+	 *
+	 *     # List administrators who don't have 2FA enabled. 
+	 *     $ wp vip two-factor report --role=administrator --2fa-enabled=false
+	 *
+	 *     # List 2FA status for administrators using the email authentication factor. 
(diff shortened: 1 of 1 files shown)


WPprodigy (approved):

rebeccahum:
@GaryJones Can you plz traumatize me with a code review thx

wp-cli/vip-two-factor.php
```diff
@@ -1,35 +1,130 @@
...
+	 *
+	 * ## EXAMPLES
+	 *
+	 *      wp vip two-factor report
+	 *      wp vip two-factor report --2fa-enabled=true
+	 *      wp vip two-factor report --2fa-provider=email
+	 *      wp vip two-factor report --role=administrator
+	 *      wp vip two-factor report --role=administrator --2fa-enabled=false
+	 *      wp vip two-factor report --role=administrator --2fa-enabled=true --2fa-provider=email
+	 *      wp vip two-factor report --user_login=wpvip
```
GaryJones:
```suggestion
	 *     # List 2FA status for all users.
	 *     $ wp vip two-factor report
	 *
	 *     # List users with 2FA enabled. 
	 *     $ wp vip two-factor report --2fa-enabled=true
	 *
	 *     # List users who use email as their authentication factor. 
	 *     $ wp vip two-factor report --2fa-provider=email
	 *
	 *     # List 2FA status for administrators. 
	 *     $ wp vip two-factor report --role=administrator
	 *
	 *     # List administrators who don't have 2FA enabled. 
	 *     $ wp vip two-factor report --role=administrator --2fa-enabled=false
	 *
	 *     # List 2FA status for administrators using the email authentication factor. 
	 *     $ wp vip two-factor report --role=administrator --2fa-enabled=true --2fa-provider=email
	 *
	 *     # List 2FA status only for the wpvip user. 
	 *     $ wp vip two-factor report --user_login=wpvip
```

Question:
> List users who use email as their authentication factor. 

Is this just primary factor, or does it include secondary providers as well?

wp-cli/vip-two-factor.php
```diff
@@ -1,35 +1,130 @@
...
+	 *
+	 * @subcommand report
```
GaryJones:
```suggestion
```

Only needed when the command doesn't match the method name (i.e. hyphens vs underscore).

wp-cli/vip-two-factor.php
```diff
@@ -1,35 +1,130 @@
...
 			}
+
+			if ( $twofa_enabled_flag && $user->two_factor_enabled !== $twofa_enabled_flag ) {
+				unset( $users[ $idx ] );
+			}
 		}
 
+		$fields   = array( 'ID', 'display_name', 'roles' );
 		$fields[] = 'two_factor_enabled';
 		$fields[] = 'two_factor_providers';
```
GaryJones, line 141:
```suggestion
		$fields   = array( 'ID', 'display_name', 'roles', 'two_factor_enabled', 'two_factor_providers' );
```

Further, looking at the output from ` wp user list --format=table`:

```sh
+----+------------------+-------------------+-----------------------------------------+---------------------+---------------+
| ID | user_login       | display_name      | user_email                              | user_registered     | roles         |
+----+------------------+-------------------+-----------------------------------------+---------------------+---------------+
```

...then maybe this report output should also include the user_login, after the ID? 
(May be outside the scope of this PR)

GaryJones (commented):
A couple of minor bits, but overall ✅ 

Would be good to add some Behat tests to check the guard clauses you have in here, as well successful output. ([example](https://github.com/Automattic/Co-Authors-Plus/blob/develop/features/rename-coauthor.feature))

Do you have any thoughts about getting this into https://github.com/WordPress/two-factor instead?

wp-cli/vip-two-factor.php
rebeccahum: It includes secondary providers 🙂

wp-cli/vip-two-factor.php
rebeccahum: Nah, I don't think it's necessary given the user lookup can be login/id/email.

PR #4918, opened 2023-10-04
Update upload name to include timestamp if file already exists.

bsessions85:
## Description

This PR changes the naming of uploaded files. If a file already exists, we will change the filename to the format `my-file-{timestamp}.ext`. Previously, WP would rename the file to `my-file-{next-available-number}.ext` which could cause big delays when the same file name was used often.

###  Modified Upload Filename Format

When a file exists, rename the file with appended timestamp rather than adding incremental digit.

## Pre-review checklist

- [ n/a] This change works and has been tested locally (or has an appropriate fallback).
- [X ] This change works and has been tested on a Go sandbox.
- [X ] This change has relevant unit tests (if applicable).
- [n/a ] This change uses a rollout method to ease with deployment (if applicable - especially for large scale actions that require writes).
- [ x] I've created a changelog description that aligns with the provided examples.

## Pre-deploy checklist

- [n/a ] VIP staff: Ensure any alerts added/updated conform to internal standards (see internal documentation). 

## Steps to Test

Test in Sandbox
1. Check out PR.
1. Go to media library
1. Upload file
1. Upload file again.
1. Verify that first file is saved with original filename and second filename has timestamp added to original filename.


tests/files/test-vip-filesystem.php
```diff
@@ -280,14 +284,41 @@ public function test__filter_validate_file__valid_file() {
...
+		$basepath = $this->get_upload_path();
+
+		$stub = $this->getMockBuilder( VIP_Filesystem::class )
+				->setMethods( [ 'validate_file_unique_name', 'validate_file_type' ] )
+				->getMock();
+
+		$stub->expects( $this->once() )
+				->method( 'validate_file_unique_name' )
+				->with( $file['name'], $basepath . '/' )
+				->will( $this->returnValue( 'testfile_1696439069.txt' ) );
```
GaryJones: The logic uses a hyphen, but this is expecting an underscore to be used?

bsessions85: Good catch, fixed the test. Thanks!

files/class-vip-filesystem.php
```diff
@@ -197,6 +202,24 @@ public function filter_validate_file( $file ) {
...
+	 *
+	 * @param  string  Path starting with /wp-content/uploads
+	 */
+	protected function validate_file_unique_name( $file_name, $upload_path ) {
+		$result = $this->stream_wrapper->client->is_file( $upload_path . $file_name );
+		
+		if ( true === $result ) {
+			$file_root = pathinfo( $file_name, PATHINFO_FILENAME );
+			$ext       = pathinfo( $file_name, PATHINFO_EXTENSION );
+			$timestamp = time();
```
mjangda, line 216: We may want to use `microtime` here and either create a random suffix based on that (or append an additional randomized string) to avoid the rare edge case collision where two files with the same name are uploaded at the exact same time.

mjangda: Another option I realized is to re-use the `validate_file_type` call. That calls `get_unique_filename` on the API, which returns a unique name if a file already exists with the provided name. That way, we don't need to introduce an additional API call the upload flow.

We'll just want to verify the the API endpoint actually does return a unique name :)

bsessions85: I did think about using `get_unique_filename` but when I looked at it in [wpvip-files service](https://github.com/Automattic/wpvip-files-service/blob/1d752be7cf906f2a670516cf8b1b646b8b69685f/src/class-files-service.php#L164), it seemed like it was basically doing the same thing that WP  does.

PR #4920, opened 2023-10-04
Add potential multi-dataset logger

luismulinari:
## Description

This PR adds a log entry containing the request information and backtrace when a potential multi-dataset query is detected.

## Changelog Description

N/A

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

1. Check out PR.
2. Perform a SQL query across multiple datasets
3. Observe logstash logs

### DIFF
--- a/prometheus-collectors/class-potential-multi-dataset-queries-collector.php
+++ b/prometheus-collectors/class-potential-multi-dataset-queries-collector.php
@@ -59,7 +59,7 @@ class Potential_Multi_Dataset_Queries_Collector implements CollectorInterface {
 					'message'  => 'Potential multi dataset query detected',
 					'blog_id'  => get_current_blog_id(),
 					'extra'    => [
-						'uri'       => isset( $_SERVER['REQUEST_URI'] ) ? sanitize_url( $_SERVER['REQUEST_URI'] ) : '',
+						'uri'       => isset( $_SERVER['REQUEST_URI'] ) ? esc_url_raw( $_SERVER['REQUEST_URI'] ) : '',
 						'method'    => isset( $_SERVER['REQUEST_METHOD'] ) ? sanitize_text_field( $_SERVER['REQUEST_METHOD'] ) : '',
 						'backtrace' => $backtrace,
 					],
--- a/prometheus-collectors/class-potential-multi-dataset-queries-collector.php
+++ b/prometheus-collectors/class-potential-multi-dataset-queries-collector.php
@@ -60,7 +60,7 @@ class Potential_Multi_Dataset_Queries_Collector implements CollectorInterface {
 					'blog_id'  => get_current_blog_id(),
 					'extra'    => [
 						'uri'       => isset( $_SERVER['REQUEST_URI'] ) ? esc_url_raw( $_SERVER['REQUEST_URI'] ) : '',
-						'method'    => isset( $_SERVER['REQUEST_METHOD'] ) ? sanitize_text_field( $_SERVER['REQUEST_METHOD'] ) : '',
+						'http_method'    => isset( $_SERVER['REQUEST_METHOD'] ) ? sanitize_text_field( $_SERVER['REQUEST_METHOD'] ) : '',
 						'backtrace' => $backtrace,
 					],
 				]
--- a/prometheus-collectors/class-potential-multi-dataset-queries-collector.php
+++ b/prometheus-collectors/class-potential-multi-dataset-queries-collector.php
@@ -4,6 +4,7 @@ namespace Automattic\VIP\Prometheus;
 
 use Prometheus\Counter;
 use Prometheus\RegistryInterface;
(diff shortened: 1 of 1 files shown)


prometheus-collectors/class-potential-multi-dataset-queries-collector.php
```diff
@@ -49,6 +50,21 @@ public function sql_query_log( $query ): void {
...
+			$backtrace = function_exists( 'wp_debug_backtrace_summary' ) ? wp_debug_backtrace_summary( null, 4, false ) : []; // phpcs:ignore
+			\Automattic\VIP\Logstash\log2logstash(
+				[
+					'severity' => 'debug',
+					'feature'  => 'potential_multi_dataset_queries',
+					'message'  => 'Potential multi dataset query detected',
+					'blog_id'  => get_current_blog_id(),
+					'extra'    => [
+						'uri'       => isset( $_SERVER['REQUEST_URI'] ) ? esc_url_raw( $_SERVER['REQUEST_URI'] ) : '',
+						'method'    => isset( $_SERVER['REQUEST_METHOD'] ) ? sanitize_text_field( $_SERVER['REQUEST_METHOD'] ) : '',
```
mjangda:
```suggestion
						'http_method'    => isset( $_SERVER['REQUEST_METHOD'] ) ? sanitize_text_field( $_SERVER['REQUEST_METHOD'] ) : '',
```

prometheus-collectors/class-potential-multi-dataset-queries-collector.php
```diff
@@ -49,6 +50,21 @@ public function sql_query_log( $query ): void {
...
+						'backtrace' => $backtrace,
```
mjangda: It may be helpful to include `$last_global_table`, `$last_blog_table` and `blog_ids_count`.

prometheus-collectors/class-potential-multi-dataset-queries-collector.php
```diff
@@ -49,6 +50,21 @@ public function sql_query_log( $query ): void {
...
 				]
 			);
+
+			$backtrace = function_exists( 'wp_debug_backtrace_summary' ) ? wp_debug_backtrace_summary( null, 4, false ) : []; // phpcs:ignore
+			\Automattic\VIP\Logstash\log2logstash(
+				[
+					'severity' => 'debug',
+					'feature'  => 'potential_multi_dataset_queries',
+					'message'  => 'Potential multi dataset query detected',
+					'blog_id'  => get_current_blog_id(),
```
mjangda: I believe `log2logstash` adds the `blog_id` for us?

prometheus-collectors/class-potential-multi-dataset-queries-collector.php
```diff
@@ -49,6 +50,21 @@ public function sql_query_log( $query ): void {
 					$blog_ids_count >= 3 ? '3+' : (string) $blog_ids_count,
 				]
 			);
+
+			$backtrace = function_exists( 'wp_debug_backtrace_summary' ) ? wp_debug_backtrace_summary( null, 4, false ) : []; // phpcs:ignore
+			\Automattic\VIP\Logstash\log2logstash(
```
mjangda, line 59: I just realized that the logstash code is loaded during mu-plugins but we can have queries run before that which will cause a fatal here.

We should make sure that the function exists before using it (I can't remember how that works with namespaces, and if we just need it here or if the `use` above is at risk as well).

(I guess this is where having an autoloader for library functions like this would be nice :( )

mjangda (changes requested):
Left some notes.

prometheus-collectors/class-potential-multi-dataset-queries-collector.php
luismulinari: Ohh good idea!

prometheus-collectors/class-potential-multi-dataset-queries-collector.php
luismulinari: You are right. We do include a few default params.

prometheus-collectors/class-potential-multi-dataset-queries-collector.php
luismulinari: Since this collector is also part of the mu-plugins, and the files (`prometheus.php` and `logstash.php`) are required at `000-vip-init.php`, I think everything is going to be loaded properly when we call `Potential_Multi_Dataset_Queries_Collector->initialize() -> add_action( 'sql_query_log' ...)`.

Just to make it risk-free, I added the check you suggested. 

What do you think?

mjangda: Good point. But, I think it's worth it to be safe, especially if load order is moved around.

mjangda (approved):
😎

WPprodigy (approved):

Issue #1141, opened 2019-02-21
WPcom specific rewrite rules logic should be removed


WPprodigy:
### 1) Flushing rewrite rules from wp-admin forces them to essentially flush twice.

This bit of code in `rri_wpcom_flush_rules ()` should no longer be necessary and should be removed: https://github.com/Automattic/vip-go-mu-plugins/blob/master/rewrite-rules-inspector.php#L57-L69

The call to `flush_rewrite_rules( false );` that the plugin makes before this hook is called is sufficient on VIP Go: https://github.com/Automattic/Rewrite-Rules-Inspector/blob/master/rewrite-rules-inspector.php#L282-L283

`flush_rewrite_rules( false );` boils down to this (which you'll notice is essentially the same thing we duplicate in our addon plugin):

```
// In wp_rewrite->flush_rules()
update_option( 'rewrite_rules', '' );
$this->wp_rewrite_rules();

// In wp_rewrite->wp_rewrite_rules()
$this->rules = get_option('rewrite_rules');
if ( empty($this->rules) ) {
  $this->matches = 'matches';
  $this->rewrite_rules(); // Reconstructs rules from scratch. Remember this is what the plugin does to find missing rules.
  update_option('rewrite_rules', $this->rules);
}
```

### 2) Flushing rewrites can behave differently when a theme is using the older WPcom ways of declaring permalink structures.

If a site is using `wpcom_vip_load_permastruct()`, `wpcom_vip_load_tag_base()`, etc, then you'll see in the same function as above (`rri_wpcom_flush_rules()`) that it calls `wpcom_vip_refresh_wp_rewrite()`.

It's in this function that I believe we have some error-some logic that might be causing some hard to catch bugs. This portion in particular feels like it should not be needed: https://github.com/Automattic/vip-go-mu-plugins/blob/da7fcd24b773e1957687c6eb56f0b56ddd402d9e/vip-helpers/vip-permastructs.php#L145-L158

I believe the `$wp_rewrite->init();` might be interferring with things. Though I'm not 100% sure how this all works so a second opinion here would be very much appreciated.

sboisvert:
I'm pretty sure you're right about point 1. Point 2 I'm less sure, but one thing that came to mind is, why are we still supporting this? Why do we want to have this dark magic that we copy pasted from wpcom? Does it help our clients?
[...]

PR #4922, opened 2023-10-05
Fix broken category links when removing category base

### LINKED ISSUE #4541: Removing category base using wpcom_vip_load_category_base() forces other functions to return a relative URL
If you use the [wpcom_vip_load_category_base() function](https://github.com/Automattic/vip-go-mu-plugins/blame/c18071981baad1c3163a3f0d04a3f85c6d35323b/vip-helpers/vip-permastructs.php#L39) to remove the category base then it forces a relative URL to be returned via the [category_link](https://github.com/Automattic/vip-go-mu-plugins/blob/c18071981baad1c3163a3f0d04a3f85c6d35323b/vip-helpers/vip-permastructs.php#L48) filter

This impacts calls to[ get_term_link()](https://developer.wordpress.org/reference/functions/get_term_link/) and [get_category_link()](https://developer.wordpress.org/reference/functions/get_category_link/), making them return relative URLs instead of absolute URLs.

This then has further knock-on impacts on code that relies on the normal absolute URL behaviour of those functions, including, for example, our own [vip-cache-Manager](https://github.com/Automattic/vip-go-mu-plugins/blob/c18071981baad1c3163a3f0d04a3f85c6d35323b/vip-cache-manager/vip-cache-manager.php#L703) when it tries to get a URL to purge the cache.

## Expected/Desired Behavior

Ideally, this should be changed so that calls to get_term_url and get_category_link will both return absolute URLs even if you have used wpcom_vip_load_category_base. This is the normal behaviour when `wpcom_vip_load_category_base()` has not been used.

However, there is a risk that this may have other impacts as the issue has been present for at least 8 years.

```
wp> get_category_link(1);
=> string(46) "http://localtest.local/uncategorized/"
```

## Actual Behavior

If you have removed the category base you will get relative URLs returned.

```
wp> get_category_link(1);
=> string(15) "/uncategorized/"
```

1. Using WP-CLI or theme code, call `get_category_link(CATEGORYID)` with a valid category ID and note the absolute URL.
2. Add `wpcom_vip_load_category_base('');` to the themes function file or an mu-plugin.
3. Repeat step 1 and note that you now get a relative URL returned instead of an absolute one.

philipjohn:
## Description

Calls to get_term_link() and get_category_link() should always return a full URL, but wpcom_vip_load_category_base() has been breaking that behaviour. This fix restores the correct functioning by using `home_url()` to ensure there is a full URL.

Fixes #4541

### Fix broken category links when removing category base

Calls to get_term_link() and get_category_link() should always return a full URL, but wpcom_vip_load_category_base() has been breaking that behaviour. This fix restores the correct functioning by using `home_url()` to ensure there is a full URL.

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

1. Using WP-CLI or theme code, call `get_category_link(CATEGORYID)` with a valid category ID and note the absolute URL.
2. Add `wpcom_vip_load_category_base('');` to the themes function file or an mu-plugin.
3. Repeat step 1 and note that you now get a relative URL returned instead of an absolute one.
4. Apply this patch and repeat step 1 and observe you now have a full URL again

### DIFF
--- a/vip-helpers/vip-permastructs.php
+++ b/vip-helpers/vip-permastructs.php
@@ -46,7 +46,7 @@ function wpcom_vip_load_category_base( $new_category_base ) {
 	// For empty category base, remove the '/category/' from the base, but include the parent category if it's a child
 	if ( '' === $new_category_base ) {
 		add_filter( 'category_link', function ( $link, $term_id ) {
-			return '/' . get_category_parents( $term_id, false, '/', true );
+			return home_url( '/' . get_category_parents( $term_id, false, '/', true ) );
 		}, 9, 2 );
 	}
 }


WPprodigy:
Hey Phillip 😃 !

Thanks for the patch. I'll give this some testing just to make sure there won't be any unexpected side-effects.

As an aside, I rather would like to deprecate the wpcomvip-era custom rewrite rules eventually. Some context in this issue, such as: https://github.com/Automattic/vip-go-mu-plugins/issues/1141#issuecomment-761245093

WPprodigy:
Setting an empty string as the category base via `wpcom_vip_load_category_base( '' )` doesn't seem to really work for me as far as actually loading a category page - WP doesn't seem to find the route. Same with `wpcom_vip_load_tag_base( '' );`, except it won't even respect the empty string and still defaults to `/tag`.

That said, I can see the issue with direct function calls like ` get_category_link(10, 'category')`. Tested both child and parent categories and the patch does seem sound. Confirmed that `get_term_link()` uses `home_url()` in the same way.

WPprodigy (approved):

PR #4935, opened 2023-10-11
Removes hardcoded VIP_PARSELY_ENABLED constant for FedRAMP sites

mehmoodak:
## Description

Currently we are hardcoding `VIP_PARSELY_ENABLED` constant as `false` for FedRAMP sites but now we are enforcing it from GOOP so we can remove it from here and only have the GOOP as source of truth.

### Update `VIP_PARSELY_ENABLED` constant for FedRAMP sites

Removes hardcoded `VIP_PARSELY_ENABLED` constant for FedRAMP sites because validations are available on backend now.

## Pre-review checklist

## Pre-deploy checklist

### DIFF
--- a/integrations/integration-vip-config.php
+++ b/integrations/integration-vip-config.php
@@ -109,32 +109,24 @@ class IntegrationVipConfig {
 	 * @private
 	 */
 	public function is_active_via_vip(): bool {
-		// Return false if blocked on org.
+		return Env_Integration_Status::ENABLED === $this->get_site_status();
+	}
+
+	/**
+	 * Get site status.
+	 *
+	 * @return string|null
+	 *
+	 * @private
+	 */
+	public function get_site_status() {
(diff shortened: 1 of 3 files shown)


integrations/integration-vip-config.php
```diff
@@ -109,32 +109,24 @@ protected function get_vip_config_from_file( string $slug ) {
...
-		if ( Env_Integration_Status::BLOCKED === $env_status ) {
-			return false;
-		}
-
-		// Look into network_sites config before because if not present we will fallback to env config.
-		$network_site_status = $this->get_value_from_config( 'network_sites', 'status' );
-
-		if ( Env_Integration_Status::ENABLED === $network_site_status ) {
-			return true;
-		}
```
mehmoodak, line 129: I noticed that this function can be simplified via refactoring so added `get_site_status()` function which will also help us in future by giving the actual status of the integration instead of just boolean.

Note: there is no change in tests which ensure that everything is working as expected.

rebeccahum (approved):

PR #4953, opened 2023-10-18
Add/mime type caching

rbcorrales:
## Description
This PR adds the logic for caching the MIME types by relying in the `pre_get_available_post_mime_types()` filter that will be added in 6.4 (https://core.trac.wordpress.org/ticket/52759).

**The code here is not final and needs additional testing for adding, updating, and removing attachments. I'm submitting is as a draft PR to start the conversation.**

Notes:
- The code will only run if the WP version is `>=` than `MINIMUM_WORDPRESS_VERSION`, which is set to 6.4.
- The effects of the code (and the caching) can be disabled altogether by ~~adding a `vip_cache_mime_types` filter and returning `false`~~ removing the `vip_cache_mime_types` action from `init`.
- Since the MIME type query could actually timeout for sites with a huge number of assets, I've included the constant `MAX_POSTS_TO_QUERY_DEFAULT` that will be used to determine if the querying/caching should be enabled or if it should fall back to the default WordPress MIME types, bringing the dropdown content to parity with the grid view. This can also be filtered via `vip_max_posts_to_query_for_mime_type_caching`.

To discuss:
- ~~Another simplified approach that would reduce the complexity would be to simply delete the cache on adding, updating, and deleting attachments and let it rebuild by itself the next time.~~ Edit: The cache is now being deleted.
- ~~If updating the caches as the code is doing it right now, we could also avoid impacting the attachment update and delete times (since an additional query might need to happen) by refactoring it to run the cache updates on shutdown.~~ Edit: This no longer applies as the cache is now being deleted on delete/update.
- ~~We could also leverage the cache updates to a cron process so we remove the update logic from the web.~~ Edit: This could be revisited later if needed.

### Feature added: MIME types caching for the media library

Cache MIME types via the `pre_get_available_post_mime_types()` filter introduced in WordPress 6.4.

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

1. Check out PR.
1. Go to `wp-admin` > `Media Library`
1. Click on the list view.
1. Verify that the attachment filter dropdown populates correctly for a site with a number of attachments less or equal to the `MAX_POSTS_TO_QUERY_DEFAULT` constant.
1. If the number of attachments exceeds the constant or the filtered value from `vip_max_posts_to_query_for_mime_type_caching`, then validate that the default attachment types are loaded. Its contents should match the dropdown from the grid view.
1. Validate that the items available in the dropdown also update accordingly when adding, updating, or deleting attachments.

### DIFF
--- a/performance.php
+++ b/performance.php
@@ -14,3 +14,4 @@ require_once __DIR__ . '/performance/vip-tweaks.php';
 require_once __DIR__ . '/performance/do-pings.php';
 require_once __DIR__ . '/performance/media.php';
 require_once __DIR__ . '/performance/plugins.php';
+require_once __DIR__ . '/performance/class-media-library-caching.php';
--- a/performance.php
+++ b/performance.php
@@ -14,4 +14,3 @@ require_once __DIR__ . '/performance/vip-tweaks.php';
 require_once __DIR__ . '/performance/do-pings.php';
 require_once __DIR__ . '/performance/media.php';
 require_once __DIR__ . '/performance/plugins.php';
-require_once __DIR__ . '/performance/class-media-library-caching.php';
--- a/performance/class-media-library-caching.php
+++ b/performance/class-media-library-caching.php
@@ -14,6 +14,7 @@ namespace Automattic\VIP\Performance;
  */
 class Media_Library_Caching {
 	public const MINIMUM_WORDPRESS_VERSION          = '6.4';
+	public const CACHE_GROUP                        = 'mime_types';
 	public const AVAILABLE_MIME_TYPES_CACHE_KEY     = 'vip_available_mime_types_';
 	public const USING_DEFAULT_MIME_TYPES_CACHE_KEY = 'vip_using_default_mime_types_';
 	public const MAX_POSTS_TO_QUERY                 = 100000;
@@ -59,7 +60,7 @@ class Media_Library_Caching {
 		global $wpdb;
 
 		$cache_key  = self::AVAILABLE_MIME_TYPES_CACHE_KEY . $type;
-		$mime_types = wp_cache_get( $cache_key );
+		$mime_types = wp_cache_get( $cache_key, self::CACHE_GROUP );
 
 		if ( false === $mime_types ) {
 			$attachment_count = $wpdb->get_var( $wpdb->prepare( "SELECT COUNT(1) FROM {$wpdb->posts} WHERE post_type = %s", $type ) );
@@ -74,11 +75,11 @@ class Media_Library_Caching {
 			}
 
 			// Cache the results.
-			wp_cache_set( $cache_key, $mime_types );
-			wp_cache_set( self::USING_DEFAULT_MIME_TYPES_CACHE_KEY . $type, $use_defaults );
+			wp_cache_set( $cache_key, $mime_types, self::CACHE_GROUP );
+			wp_cache_set( self::USING_DEFAULT_MIME_TYPES_CACHE_KEY . $type, $use_defaults, self::CACHE_GROUP );
 		}
 
-		// If there were any previous mime types, merge them with the cached mime types.
+		// If there were any previous MIME types, merge them with the cached MIME types.
 		if ( is_array( $filtered_mime_types ) ) {
 			$mime_types = array_unique( array_merge( $filtered_mime_types, $mime_types ) );
 		}
@@ -110,7 +111,7 @@ class Media_Library_Caching {
 	public static function update_post_mime_types_cache_on_add( $post_id ) {
 		$type = get_post_type( $post_id );
 
-		if ( wp_cache_get( self::USING_DEFAULT_MIME_TYPES_CACHE_KEY . $type ) ) {
+		if ( wp_cache_get( self::USING_DEFAULT_MIME_TYPES_CACHE_KEY . $type, self::CACHE_GROUP ) ) {
 			return;
 		}
 
@@ -136,11 +137,11 @@ class Media_Library_Caching {
 			return;
 		}
 
-		if ( ! wp_cache_get( self::USING_DEFAULT_MIME_TYPES_CACHE_KEY . $old_post_type ) ) {
-			self::remove_mime_type_from_cache( $old_mime_type, $old_post_type, $post_id );
+		if ( ! wp_cache_get( self::USING_DEFAULT_MIME_TYPES_CACHE_KEY . $old_post_type, self::CACHE_GROUP ) ) {
+			self::delete_mime_type_cache( $old_post_type );
 		}
 
-		if ( ! wp_cache_get( self::USING_DEFAULT_MIME_TYPES_CACHE_KEY . $new_post_type ) ) {
+		if ( ! wp_cache_get( self::USING_DEFAULT_MIME_TYPES_CACHE_KEY . $new_post_type, self::CACHE_GROUP ) ) {
 			self::add_mime_type_to_cache( $new_mime_type, $new_post_type );
 		}
 	}
@@ -154,63 +155,42 @@ class Media_Library_Caching {
 	public static function update_post_mime_types_cache_on_delete( $post_id, $post ) {
 		$type = $post->post_type;
 
-		if ( wp_cache_get( self::USING_DEFAULT_MIME_TYPES_CACHE_KEY . $type ) ) {
+		if ( wp_cache_get( self::USING_DEFAULT_MIME_TYPES_CACHE_KEY . $type, self::CACHE_GROUP ) ) {
 			return;
 		}
 
-		$mime_type = $post->post_mime_type;
-		self::remove_mime_type_from_cache( $mime_type, $type, $post_id );
+		self::delete_mime_type_cache( $type );
 	}
 
 	/**
 	 * Add a MIME type to the cache.
 	 *
-	 * @param string $mime_type The mime type to add.
+	 * @param string $mime_type The MIME type to add.
 	 * @param string $type      The post type name.
 	 */
 	private static function add_mime_type_to_cache( $mime_type, $type ) {
 		if ( false !== $mime_type ) {
 			$cache_key  = self::AVAILABLE_MIME_TYPES_CACHE_KEY . $type;
-			$mime_types = wp_cache_get( $cache_key );
+			$mime_types = wp_cache_get( $cache_key, self::CACHE_GROUP );
 
 			if ( false !== $mime_types ) {
 				// Add the new mime type to the cache if not present.
 				if ( ! in_array( $mime_type, $mime_types, true ) ) {
 					$mime_types[] = $mime_type;
-					wp_cache_set( $cache_key, $mime_types );
+					wp_cache_set( $cache_key, $mime_types, self::CACHE_GROUP );
 				}
 			}
 		}
 	}
 
 	/**
-	 * Remove a MIME type from the cache.
+	 * Delete the MIME type the cache.
 	 *
-	 * @param string   $mime_type       The mime type to remove.
-	 * @param string   $type            The post type name.
-	 * @param int|null $exclude_post_id The post ID to exclude from the query. Default null.
+	 * @param string $type The post type name.
 	 */
-	private static function remove_mime_type_from_cache( $mime_type, $type, $exclude_post_id = null ) {
-		global $wpdb;
-
-		if ( false !== $mime_type ) {
-			$cache_key  = self::AVAILABLE_MIME_TYPES_CACHE_KEY . $type;
-			$mime_types = wp_cache_get( $cache_key );
-
-			if ( false !== $mime_types ) {
-				// Check if there are any posts left with the mime type before removing it from the cache.
-				// phpcs:ignore WordPress.DB.DirectDatabaseQuery.NoCaching
-				$count = $wpdb->get_col( $wpdb->prepare( "SELECT 1 FROM $wpdb->posts WHERE ID != %d AND post_type = %s AND post_mime_type = %s LIMIT 1", $exclude_post_id, $type, $mime_type ) );
-
-				if ( $count < 1 ) {
-					// Remove the mime type from the cache if present.
-					if ( in_array( $mime_type, $mime_types, true ) ) {
-						$mime_types = array_diff( $mime_types, array( $mime_type ) );
-						wp_cache_set( $cache_key, $mime_types );
-					}
-				}
-			}
-		}
+	private static function delete_mime_type_cache( $type ) {
+		$cache_key = self::AVAILABLE_MIME_TYPES_CACHE_KEY . $type;
+		wp_cache_delete( $cache_key, self::CACHE_GROUP );
 	}
 }
 
--- a/performance/class-media-library-caching.php
+++ b/performance/class-media-library-caching.php
@@ -17,7 +17,7 @@ class Media_Library_Caching {
 	public const CACHE_GROUP                        = 'mime_types';
 	public const AVAILABLE_MIME_TYPES_CACHE_KEY     = 'vip_available_mime_types_';
 	public const USING_DEFAULT_MIME_TYPES_CACHE_KEY = 'vip_using_default_mime_types_';
-	public const MAX_POSTS_TO_QUERY                 = 100000;
+	public const MAX_POSTS_TO_QUERY_DEFAULT         = 100000; // TODO: benchmark this value.
 
 	/**
 	 * Class initialization.
@@ -63,8 +63,16 @@ class Media_Library_Caching {
 		$mime_types = wp_cache_get( $cache_key, self::CACHE_GROUP );
 
 		if ( false === $mime_types ) {
-			$attachment_count = $wpdb->get_var( $wpdb->prepare( "SELECT COUNT(1) FROM {$wpdb->posts} WHERE post_type = %s", $type ) );
-			$use_defaults     = $attachment_count > self::MAX_POSTS_TO_QUERY;
+
+			/**
+			 * Filters the max number of posts to query for dynamic MIME type caching.
+			 *
+			 * @param bool $max_posts_to_query Max number of posts to query.
+			 */
+			$max_posts_to_query = apply_filters( 'vip_max_posts_to_query_for_mime_type_caching', self::MAX_POSTS_TO_QUERY_DEFAULT );
+
+			$attachment_count = $max_posts_to_query > 0 ? $wpdb->get_var( $wpdb->prepare( "SELECT COUNT(1) FROM {$wpdb->posts} WHERE post_type = %s", $type ) ) : 1;
+			$use_defaults     = $attachment_count > $max_posts_to_query;
 
 			if ( $use_defaults ) {
 				// If there are too many posts to query, use the default mime types.
--- a/performance/class-media-library-caching.php
+++ b/performance/class-media-library-caching.php
@@ -57,8 +57,6 @@ class Media_Library_Caching {
 	 * @return array An array of MIME types.
 	 */
 	public static function get_cached_post_mime_types( $filtered_mime_types, $type ) {
-		global $wpdb;
-
 		if ( 'attachment' !== $type ) {
 			return $filtered_mime_types;
 		}
@@ -74,6 +72,7 @@ class Media_Library_Caching {
 			 */
 			$max_posts_to_query = apply_filters( 'vip_max_posts_to_query_for_mime_type_caching', self::MAX_POSTS_TO_QUERY_DEFAULT );
 
+			global $wpdb;
 			$attachment_count = $max_posts_to_query > 0 ? $wpdb->get_var( $wpdb->prepare( "SELECT COUNT(1) FROM {$wpdb->posts} WHERE post_type = %s", $type ) ) : 1;
 			$use_defaults     = $attachment_count > $max_posts_to_query;
 
--- a/performance/class-media-library-caching.php
+++ b/performance/class-media-library-caching.php
@@ -15,9 +15,9 @@ namespace Automattic\VIP\Performance;
 class Media_Library_Caching {
 	public const MINIMUM_WORDPRESS_VERSION          = '6.4';
 	public const CACHE_GROUP                        = 'mime_types';
-	public const AVAILABLE_MIME_TYPES_CACHE_KEY     = 'vip_available_mime_types_';
-	public const USING_DEFAULT_MIME_TYPES_CACHE_KEY = 'vip_using_default_mime_types_';
-	public const MAX_POSTS_TO_QUERY_DEFAULT         = 100000; // TODO: benchmark this value.
+	public const AVAILABLE_MIME_TYPES_CACHE_KEY     = 'vip_available_mime_types';
+	public const USING_DEFAULT_MIME_TYPES_CACHE_KEY = 'vip_using_default_mime_types';
+	public const MAX_POSTS_TO_QUERY_DEFAULT         = 100000;
 
 	/**
 	 * Class initialization.
@@ -44,23 +44,26 @@ class Media_Library_Caching {
 	 */
 	private static function enable_post_mime_types_caching() {
 		add_filter( 'pre_get_available_post_mime_types', array( __CLASS__, 'get_cached_post_mime_types' ), 10, 2 );
-		add_action( 'add_attachment', array( __CLASS__, 'update_post_mime_types_cache_on_add', 10, 1 ) );
+		add_action( 'add_attachment', array( __CLASS__, 'update_post_mime_types_cache_on_add' ), 10, 1 );
 		add_action( 'attachment_updated', array( __CLASS__, 'update_post_mime_types_cache_on_edit' ), 10, 3 );
-		add_action( 'delete_attachment', array( __CLASS__, 'update_post_mime_types_cache_on_delete' ), 10, 2 );
+		add_action( 'delete_attachment', array( __CLASS__, 'update_post_mime_types_cache_on_delete' ), 10 );
 	}
 
 	/**
 	 * Get cached results for get_available_post_mime_types() to avoid a query on every page load.
 	 *
 	 * @param string[]|null $filtered_mime_types An array of MIME types. Default null.
-	 * @param string        $type                The post type name. Usually 'attachment' but can be any post type.
+	 * @param string        $type                The post type name.
 	 * @return array An array of MIME types.
 	 */
 	public static function get_cached_post_mime_types( $filtered_mime_types, $type ) {
 		global $wpdb;
 
-		$cache_key  = self::AVAILABLE_MIME_TYPES_CACHE_KEY . $type;
-		$mime_types = wp_cache_get( $cache_key, self::CACHE_GROUP );
+		if ( 'attachment' !== $type ) {
+			return $filtered_mime_types;
+		}
+
+		$mime_types = wp_cache_get( self::AVAILABLE_MIME_TYPES_CACHE_KEY, self::CACHE_GROUP );
 
 		if ( false === $mime_types ) {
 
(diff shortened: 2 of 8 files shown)


rbcorrales:
Rebased my branch from production to develop.

vip-helpers/class-vip-media-library-caching.php
```diff
@@ -0,0 +1,210 @@
+<?php
```
rebeccahum: Minor nitpick - but can we put this in the performance folder @ https://github.com/Automattic/vip-go-mu-plugins/tree/develop/performance?

vip-helpers/class-vip-media-library-caching.php
```diff
@@ -0,0 +1,210 @@
...
+	public const USING_DEFAULT_MIME_TYPES_CACHE_KEY = 'vip_using_default_mime_types_';
+	public const MAX_POSTS_TO_QUERY                 = 100000;
+
+	/**
+	 * Class constructor.
+	 */
+	public function __construct() {
+		global $wp_version;
+
+		if ( ! defined( 'VIP_DISABLE_MIME_TYPE_CACHING' ) &&
```
rebeccahum: Maybe we could use a filter instead to opt-out, e.g. https://github.com/Automattic/vip-go-mu-plugins/blob/aa74a6c36b63aaa9ecccafd16d892a66ec374c6d/001-core/privacy.php#L21-L22

vip-helpers/class-vip-media-library-caching.php
```diff
@@ -0,0 +1,210 @@
...
+namespace Automattic\VIP\Helpers;
+
+/**
+ * VIP Media Library Caching class.
+ */
+class VIP_Media_Library_Caching {
+	public const MINIMUM_WORDPRESS_VERSION          = '6.4';
+	public const AVAILABLE_MIME_TYPES_CACHE_KEY     = 'vip_available_mime_types_';
+	public const USING_DEFAULT_MIME_TYPES_CACHE_KEY = 'vip_using_default_mime_types_';
+	public const MAX_POSTS_TO_QUERY                 = 100000;
```
rebeccahum: If possible, would be nice to do some benchmarking to see at what point in the count of the posts table it really starts to slow down

vip-helpers/class-vip-media-library-caching.php
```diff
@@ -0,0 +1,210 @@
...
+
+			if ( false !== $mime_types ) {
+				// Check if there are any posts left with the mime type before removing it from the cache.
+				// phpcs:ignore WordPress.DB.DirectDatabaseQuery.NoCaching
+				$count = $wpdb->get_col( $wpdb->prepare( "SELECT 1 FROM $wpdb->posts WHERE ID != %d AND post_type = %s AND post_mime_type = %s LIMIT 1", $exclude_post_id, $type, $mime_type ) );
+
+				if ( $count < 1 ) {
+					// Remove the mime type from the cache if present.
+					if ( in_array( $mime_type, $mime_types, true ) ) {
+						$mime_types = array_diff( $mime_types, array( $mime_type ) );
```
rebeccahum: `array_diff()` can be a bit slow, what are your thoughts on using `unset()` instead?

vip-helpers/class-vip-media-library-caching.php
```diff
@@ -0,0 +1,210 @@
...
+		global $wpdb;
+
+		if ( false !== $mime_type ) {
+			$cache_key  = self::AVAILABLE_MIME_TYPES_CACHE_KEY . $type;
+			$mime_types = wp_cache_get( $cache_key );
+
+			if ( false !== $mime_types ) {
+				// Check if there are any posts left with the mime type before removing it from the cache.
+				// phpcs:ignore WordPress.DB.DirectDatabaseQuery.NoCaching
+				$count = $wpdb->get_col( $wpdb->prepare( "SELECT 1 FROM $wpdb->posts WHERE ID != %d AND post_type = %s AND post_mime_type = %s LIMIT 1", $exclude_post_id, $type, $mime_type ) );
```
rebeccahum: Rather than executing an extra query to check everytime an attachment is deleted, what about just invalidating the cache instead? This way, we can just repopulate as needed.

vip-helpers/class-vip-media-library-caching.php
```diff
@@ -0,0 +1,210 @@
...
+		$new_mime_type = $post_after->post_mime_type;
+		$new_post_type = $post_after->post_type;
+
+		// Do nothing if the mime type didn't change.
+		if ( $new_mime_type === $old_mime_type ) {
+			return;
+		}
+
+		if ( ! wp_cache_get( self::USING_DEFAULT_MIME_TYPES_CACHE_KEY . $old_post_type ) ) {
+			$this->remove_mime_type_from_cache( $old_mime_type, $old_post_type, $post_id );
```
rebeccahum: Same as above ^^.

vip-helpers/class-vip-media-library-caching.php
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	/**
+	 * Get cached results for get_available_post_mime_types() to avoid a query on every page load.
+	 * 
+	 * @param string[]|null $filtered_mime_types An array of MIME types. Default null.
+	 * @param string        $type                The post type name. Usually 'attachment' but can be any post type.
+	 * @return array An array of MIME types.
+	 */
+	public function get_cached_post_mime_types( $filtered_mime_types, $type ) {
```
rebeccahum: Do we care about running this filter if `$type` isn't `attachment`? My understanding is that the expensive query stems from https://github.com/WordPress/WordPress/blob/b0da2934b925077ab95cd36cad4217ed214665b1/wp-admin/includes/post.php#L1390

vip-helpers/class-vip-media-library-caching.php
```diff
@@ -0,0 +1,210 @@
...
+		global $wpdb;
+
+		$cache_key  = self::AVAILABLE_MIME_TYPES_CACHE_KEY . $type;
+		$mime_types = wp_cache_get( $cache_key );
+
+		if ( false === $mime_types ) {
+			$attachment_count = $wpdb->get_var( $wpdb->prepare( "SELECT COUNT(1) FROM {$wpdb->posts} WHERE post_type = %s", $type ) );
+			$use_defaults     = $attachment_count > self::MAX_POSTS_TO_QUERY;
+
+			if ( $use_defaults ) {
```
rebeccahum, line 61: What are your thoughts on caching this value with an expiry time to prevent it from running a query every time if we don't have anything returned in `$mime_types`?

rebeccahum (dismissed):

vip-helpers/class-vip-media-library-caching.php
emrikol: What if instead of running this long query in the admin interface, we offload it to a cron job? This could also have the added benefit of using the core `get_available_mime_types()` (after removing the custom caching filter) so that we don't need to build and maintain a custom SQL query. And in cron, the slow query shouldn't matter as much.

vip-helpers/class-vip-media-library-caching.php
```diff
@@ -0,0 +1,210 @@
...
+	 * 
+	 * @param string[]|null $filtered_mime_types An array of MIME types. Default null.
+	 * @param string        $type                The post type name. Usually 'attachment' but can be any post type.
+	 * @return array An array of MIME types.
+	 */
+	public function get_cached_post_mime_types( $filtered_mime_types, $type ) {
+		global $wpdb;
+
+		$cache_key  = self::AVAILABLE_MIME_TYPES_CACHE_KEY . $type;
+		$mime_types = wp_cache_get( $cache_key );
```
emrikol: The amount of data stored for mime types should be very low. Instead of storing this in object cache, why don't we store it in a non-autoloaded option?

This way the data is always there in the database, and on a cold cache start, or random eviction, the data won't need to be rebuilt?

vip-helpers/class-vip-media-library-caching.php
emrikol: When might this happen? Would the cache invalidation during add/edit/delete be enough to clear it out?

vip-helpers/class-vip-media-library-caching.php
```diff
@@ -0,0 +1,210 @@
...
+ */
+
+ // phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery
+
+namespace Automattic\VIP\Helpers;
+
+/**
+ * VIP Media Library Caching class.
+ */
+class VIP_Media_Library_Caching {
```
emrikol: If we use object cache, it would be good to use a non-default cache group as well so we can better track stats, and future-proofing if we ever support `wp_cache_flush_group()`

vip-helpers/class-vip-media-library-caching.php
rbcorrales: Sure, I'll do that.

vip-helpers/class-vip-media-library-caching.php
rbcorrales: Sounds good.

vip-helpers/class-vip-media-library-caching.php
rbcorrales: Yeah, that's a good idea. I put that arbitrary number initially based on the issues I saw with a customer with a few million attachments, making the query take more than a minute (which would likely be killed). We can do additional benchmarks but also add the possibility of overriding it via a filter for some custom scenarios.

vip-helpers/class-vip-media-library-caching.php
rbcorrales: I considered the idea (it's on the PR notes at the top), but I wasn't 100% sure, as running the query for regenerating this cache for a large site might be still costly. But I agree we can remove a lot of the complexity by simply invalidating it.

vip-helpers/class-vip-media-library-caching.php
rbcorrales: I added `$type` to preserve the method signature and filter definition from:
https://github.com/WordPress/WordPress/blob/b0da2934b925077ab95cd36cad4217ed214665b1/wp-includes/post.php#L7937

What you pointed out is the only place where that function is called in core, and 'attachment' is hard coded. But I still preferred to make the distinction to make the code more resilient if, for some (unlikely) reason, core decided to call that function for a different post type at some point.

vip-helpers/class-vip-media-library-caching.php
rbcorrales: `$use_defaults` is being cached below, so the other filters are aware, but here, it's gated inside the `$mime_types` cache generation in case we have a change in the number of attachments that crosses the `MAX_POSTS_TO_QUERY` threshold, which should effectively change the expected behavior. But maybe I'm overthinking it for some edge cases.

vip-helpers/class-vip-media-library-caching.php
rebeccahum: Sorry for any confusion, to clarify -- I meant can we bail early if `$type` isn't attachment? Don't need the logic to run for other types.

vip-helpers/class-vip-media-library-caching.php
rbcorrales: Let me think about it, but I think the only way this could be a db option is if we completely defer the update behavior to a cron, given the risk of having a bulk attachment import/delete triggering a massive number of db writes.

@rebeccahum thoughts?

rebeccahum: Mmm, I'm in preference for storing it in object cache due to the nature of the data, which _could_ change based on uploads/updates to the media. I also have concerns with the triggering of DB writes as well.

vip-helpers/class-vip-media-library-caching.php
rbcorrales: Ok, I get that. It's only that the current structure would allow caching of MIME types for attachments OR any other post types that might come in the future through that filter. If we restrict it to only attachments, a lot of the logic involving variable cache buckets according to the `$type` variable could also be removed from the other places, and `attachment` could be hard coded. But I leave the final call up to you. 🙂

vip-helpers/class-vip-media-library-caching.php
rbcorrales: Done.

vip-helpers/class-vip-media-library-caching.php
rbcorrales: Done. Added a new `vip_cache_mime_types` filter. It's enabled by default.

vip-helpers/class-vip-media-library-caching.php
rbcorrales: I made this value filterable via `vip_max_posts_to_query_for_mime_type_caching`, but the benchmark for the default is pending.

vip-helpers/class-vip-media-library-caching.php
rbcorrales: The logic for deletions was simplified, and this was removed.

vip-helpers/class-vip-media-library-caching.php
rbcorrales: The cache is now being deleted on attachment deletions.

vip-helpers/class-vip-media-library-caching.php
rbcorrales: This was implemented, and a `mime_types` cache group is being used across this class.

vip-helpers/class-vip-media-library-caching.php
rebeccahum: Given that the hooks we are using are attachment specific, it'd make sense to only cache for attachments, yeah?

rebeccahum: I think cron seems overkill for this. We can revisit later if we need it.

rbcorrales: I removed all the post-type-based logic, and it will only cache attachments now.

performance/class-media-library-caching.php
```diff
@@ -0,0 +1,164 @@
...
+
+	/**
+	 * Get cached results for get_available_post_mime_types() to avoid a query on every page load.
+	 *
+	 * @param string[]|null $filtered_mime_types An array of MIME types. Default null.
+	 * @param string        $type                The post type name.
+	 * @return array An array of MIME types.
+	 */
+	public static function get_cached_post_mime_types( $filtered_mime_types, $type ) {
+		global $wpdb;
```
rebeccahum: Minor nitpick - let's move this down since we don't need to call the global yet.

vip-helpers/class-vip-media-library-caching.php
rbcorrales: I tested the query from WP core in the following scenarios:

| Attachments | Total Posts | Time taken |
|---|---|---|
| 10M | 28M | killed |
| 4.4M | 9M | killed |
| 2.5M | 10M | 53s / killed |
| 615K | 2.6M | 56s / killed |
| 530K | 1.4M | 13s |
| 180K | 1M | 2.6s |
| 35K | 330K | 0.6s |

@rebeccahum What if we up the default max query from 100K to 500K attachments?

rbcorrales:
All the feedback has been discussed/taken care of. Test cases cover all the functions, but the Codecov report is currently at 0% because there's no 6.4 environment, so the tests are skipped until that's released.

performance/class-media-library-caching.php
```diff
@@ -0,0 +1,163 @@
...
+		if ( $post_before->post_mime_type !== $post_after->post_mime_type ) {
+			wp_cache_delete( self::AVAILABLE_MIME_TYPES_CACHE_KEY, self::CACHE_GROUP );
+		}
+	}
+
+	/**
+	 * Update the MIME types cache when a post is deleted.
+	 */
+	public static function update_post_mime_types_cache_on_delete() {
+		wp_cache_delete( self::AVAILABLE_MIME_TYPES_CACHE_KEY, self::CACHE_GROUP );
```
WPprodigy: I feel like this might lead to a bit more invalidation than really needed.

Maybe we could do a follow-up query and see if we can find just 1 attachment with the mime type that is being deleted. And if so, there's no need to invalidate. That may be overkill though.

Also, if `USING_DEFAULT_MIME_TYPES_CACHE_KEY` is true, we shouldn't invalidate right?

rbcorrales: @WPprodigy That was precisely my initial approach to this, but I refactored it after the PR feedback, so it wasn't overkill and we could avoid the extra query 😄 

You can see the previous state here:
https://github.com/Automattic/vip-go-mu-plugins/blob/95eaa57dca82cd1e18e49013f1bcbc14ad002807/performance/class-media-library-caching.php#L203-L212

rbcorrales: > Also, if USING_DEFAULT_MIME_TYPES_CACHE_KEY is true, we shouldn't invalidate right?

That's a tricky question because there's the edge case where adding/removing attachments might go over or below the threshold for executing the query or going with the defaults.

So, for example, if the `MAX_POSTS_TO_QUERY_DEFAULT` variable is set to 500K, and you currently have 500,001 attachments but delete one (or a bunch of them), then in theory, you should stop showing the defaults and do the DB query, but that wouldn't happen if you don't invalidate the cache so that the threshold could be rechecked.

WPprodigy: I guess it depends on how expensive the `SELECT COUNT(1) FROM {$wpdb->posts} WHERE post_type = %s` query could be, as invalidating here causes that to run again at a minimum.

If that gets pricey, then I'd say it's worth just aborting if already using default mime types. I'm not really worried about being EXACT on the `max_posts_to_query`. But if that stays cheap enough, then invalidating here is fine :)

rbcorrales: Based on the tests I performed while checking [this](https://github.com/Automattic/vip-go-mu-plugins/pull/4953#discussion_r1377025323), the count attachment select was relatively slow on very large tables, but still only took about 1/20th of the time that the distinct query takes, as it will only read the index. In a scenario with about 10M attachments, it took less than 10 seconds, while the distinct query could take up to minutes.

Regarding the previous delete check that I referenced above, that one is better than running a new distinct query but still slower than the attachment count check, as it needs to read the table.

Having said that, I think I'll avoid invalidating the cache if using defaults then.

rbcorrales: A check was added to the update/delete filters to check if it's using defaults.

performance.php
```diff
@@ -14,3 +14,4 @@
 require_once __DIR__ . '/performance/do-pings.php';
 require_once __DIR__ . '/performance/media.php';
 require_once __DIR__ . '/performance/plugins.php';
+require_once __DIR__ . '/performance/class-media-library-caching.php';
```
WPprodigy: Let's do some slight re-organization of the files

- Call the class  `Mime_Types_Caching`
- Store it at `/performance/class-mime-types-caching.php`
- But remove the require from this `performance.php` file and instead:
  - Load this new class file in the existing `/performance/media.php`
  - Move the add_action() that initiates this class inside `/performance/media.php`, perhaps with a new function declared in this file that loads the class instance and runs init(). This will make it simpler for customers to just unhook it all, and we can do away with the `vip_cache_mime_types` filter then. The class won't need to be concerned with that level of detail.

performance/class-media-library-caching.php
```diff
@@ -0,0 +1,174 @@
...
+				// If there are too many posts to query, use the default MIME types.
+				$mime_types = self::get_default_mime_types();
+			} else {
+				// Otherwise, use the same query from core.
+				$mime_types = $wpdb->get_col( $wpdb->prepare( "SELECT DISTINCT post_mime_type FROM $wpdb->posts WHERE post_type = %s", $type ) );
+			}
+
+			// Cache the results.
+			wp_cache_set( self::AVAILABLE_MIME_TYPES_CACHE_KEY, $mime_types, self::CACHE_GROUP );
+			wp_cache_set( self::USING_DEFAULT_MIME_TYPES_CACHE_KEY, $use_defaults, self::CACHE_GROUP );
```
WPprodigy: May end up in a confusing scenario where one gets evicted but the other is not. I'd prefer we just keep this data all inside one cache key I think, better for slight performance gains as well. The cache value can just be like:

```
[
 'available_mime_types' => [...]
 'using_default_mime_types' => true|false
]
```

Then we just fetch one cache key and can check for the specific values when needed.

performance.php
rbcorrales: These changes were implemented.

performance/class-media-library-caching.php
rbcorrales: Went with a `vip_mime_types` cache key, which contains an array with the following keys (to avoid name redundancy):
```
[
 'available_types' => [...]
 'using_defaults' => true|false
]
```
And refactored everything to handle a single cache entry.

performance/class-mime-types-caching.php
```diff
@@ -0,0 +1,160 @@
...
+
+		return $available_types;
+	}
+
+	/**
+	 * Get the default MIME types.
+	 *
+	 * @return string[] An array of default MIME types.
+	 */
+	public static function get_default_mime_types() {
```
rinatkhaziev, line 103: I have performance concerns about this method. `get_post_mime_types()` is a slow function, and this method is essentially partially undoing its work.  So first, get_post_mime_types gets flat mime types and runs regex over available mime types to filter some out and group others by arbitrary type, and then we come in and do the flattening.

Consider this on 10k iterations:
> get_post_mime_types Elapsed: 1.9381890296936

> wp_get_mime_types: 0.00063705444335938


Furthermore, the functional approach is almost 3 times slower than simplistic explode/join on the same 10k iterations:

```php
	$pmt =  array_keys( get_post_mime_types() );
	$i = 0;
	$m = microtime( true );
	while ( $i++ < 10000 ) {
		$functional = array_reduce(
			$pmt,
			function ( $carry, $mime_type ) {
				return array_merge( $carry, explode( ',', $mime_type ) );
			},
			array()
		);
	}
	echo "array_reduce Elapsed: " . microtime( true ) - $m;


	$i = 0;
	$m = microtime( true );
	while ( $i++ < 10000 ) {
		$explode_join = explode( ',', join( ',', $pmt ) );
	}

	echo "explode/join elapsed: " . microtime(true) - $m;
} );
```

> array_reduce Elapsed: 0.0083451271057129
> explode/join elapsed: 0.003554105758667

rinatkhaziev (dismissed):

performance/class-mime-types-caching.php
rbcorrales: Thanks for the feedback, @rinatkhaziev. The join/explode approach seems to be more performant indeed. Code updated.

In regards to the `get_post_mime_types()` call, I utilized it because the grid view populates the media type dropdown based on the results from calling that function too, and I wanted the list view to be in parity to ensure consistency in the future:
https://github.com/WordPress/WordPress/blob/6.4.1/wp-includes/media.php#L4702

In our current version of `get_default_mime_types()`, we end up with an array of 26 items (in a vanilla WP install) after manipulating the results from `get_post_mime_types()`. While doing `array_values( wp_get_mime_types() )` in this function would be quicker, it returns 96 items corresponding to the unfiltered data.

If used as is, the end result in the dropdown will be similar, but there are at least two side effects:
1. More than triple the data would be cached/read every time (3.68KB vs 1.07KB when serialized)
2. Having more items would mean the caller functions at the upper levels (like `wp_match_mime_types()` coming from `media_upload_library_form()`) would need to perform more iterations.

rinatkhaziev: I clearly lack context here :) Will this method only be called in wp-admin?

rbcorrales: Correct, only when visiting the list view in the media library on a site with a large number of attachments and if the results are not already cached.

rbcorrales:
Hey @rinatkhaziev, could this be merged, or are there any outstanding items (apart from rebasing)? All the feedback was addressed two months ago but was just marked as stale today for lack of activity.

rinatkhaziev:
Hey @rbcorrales, sorry about that! We're going to take another look and hopefully deploy it to staging the next week.

performance/class-mime-types-caching.php
```diff
@@ -0,0 +1,163 @@
...
+
+	/**
+	 * Class initialization.
+	 */
+	public static function init() {
+		global $wp_version;
+
+		if ( isset( $wp_version ) &&
+			version_compare( $wp_version, self::MINIMUM_WORDPRESS_VERSION, '>=' ) ) {
+			static::enable_post_mime_types_caching();
```
rebeccahum, line 29: Is it possible we want an easy mechanism for our users to disable this altogether (just incase)?

performance/class-mime-types-caching.php
```diff
@@ -0,0 +1,163 @@
...
+		$mime_types = wp_cache_get( self::MIME_TYPES_CACHE_KEY, self::CACHE_GROUP );
+
+		if ( false === $mime_types ) {
+
+			/**
+			 * Filters the max number of posts to query for dynamic MIME type caching.
+			 *
+			 * @param bool $max_posts_to_query Max number of posts to query.
+			 */
+			$max_posts_to_query = apply_filters( 'vip_max_posts_to_query_for_mime_type_caching', self::MAX_POSTS_TO_QUERY_DEFAULT );
```
rebeccahum, line 66:
```suggestion
			$max_posts_to_query = absint( apply_filters( 'vip_max_posts_to_query_for_mime_type_caching', self::MAX_POSTS_TO_QUERY_DEFAULT ) );
```

Then we can remove the check on the off-chance that it evaluates to less than 0.

performance/class-mime-types-caching.php
```diff
@@ -0,0 +1,163 @@
...
+
+			global $wpdb;
+			$attachment_count = $max_posts_to_query > 0 ? $wpdb->get_var( $wpdb->prepare( "SELECT COUNT(1) FROM {$wpdb->posts} WHERE post_type = %s", $type ) ) : null;
```
rebeccahum, line 69:
```suggestion
			$attachment_count = $wpdb->get_var( $wpdb->prepare( "SELECT COUNT(1) FROM {$wpdb->posts} WHERE post_type = %s", $type ) );
```

rebeccahum (changes requested):
Minor, but everything else is looking good!

performance/class-mime-types-caching.php
rbcorrales: The mechanism to turn this off would be removing the action from `init`:
```
remove_action( 'init', 'vip_cache_mime_types' );
```
This approach was Caleb's suggestion from a [previous review](https://github.com/Automattic/vip-go-mu-plugins/pull/4953#discussion_r1382281936) instead of using a new filter.

performance/class-mime-types-caching.php
rbcorrales: I structured it like that to have a way to skip the query in case a user always wanted to use the defaults. Then they can just set `MAX_POSTS_TO_QUERY_DEFAULT` to zero or a negative value, like `-1`, and we can skip the `COUNT()` query altogether.

This approach could be helpful for users who wouldn't ever be near the threshold for running the full query in the DB, like someone with millions of attachments.

It should probably be documented, but does it make sense?

performance/class-mime-types-caching.php
rbcorrales: Please check my other comment.

performance/class-mime-types-caching.php
rebeccahum: Makes sense. Can you add a note in that filter PHPDoc?

rbcorrales: Added with https://github.com/Automattic/vip-go-mu-plugins/pull/4953/commits/c654f6bc77fed90f4cec6fa1bba6bc8369846889.

rebeccahum (approved):

performance/class-mime-types-caching.php
```diff
@@ -0,0 +1,165 @@
...
+	public const MAX_POSTS_TO_QUERY_DEFAULT = 500000;
+
+	/**
+	 * Class initialization.
+	 */
+	public static function init() {
+		global $wp_version;
+
+		if ( isset( $wp_version ) &&
+			version_compare( $wp_version, self::MINIMUM_WORDPRESS_VERSION, '>=' ) ) {
```
GaryJones, line 28: https://developer.wordpress.org/reference/functions/is_wp_version_compatible/ would have been slightly better to use here, avoiding the need to pull in the global variable and account for version suffixes.

PR #4955, opened 2023-10-18
Fix connection pilot’s backoff logic

WPprodigy:
DateTime->diff() gives a result like this:

```
$diff = $dt_now->diff( $dt_heartbeat, true );
=> object(DateInterval)#8069 (16) {
  ["y"]=>
  int(0)
  ["m"]=>
  int(1)
  ["d"]=>
  int(17)
  ["h"]=>
  int(2)
  ["i"]=>
  int(20)
  ["s"]=>
  int(3)
```

So checking only `$diff->h` is not sufficient once days, or even months/years have passed. Checking hours alone for sure will never work after the backoff is greater than 24h. And the max backoff period currently set is 7 days.

Also removed an extra `update_heartbeat()` call that kind of gets in the way of the backoff logic and could refresh the timestamp and cause inaccurate backoff checks to follow. Instead, can just enforce the `MAX_BACKOFF_FACTOR` in the `update_backoff_factor()` method, while also ensuring a call to that method always refreshes the timestamp appropriately.

### DIFF
--- a/vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
+++ b/vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
@@ -180,12 +180,7 @@ class Connection_Pilot {
 	 */
 	private function should_back_off(): bool {
 		if ( ! empty( $this->last_heartbeat['backoff_factor'] ) && ! empty( $this->last_heartbeat['timestamp'] ) ) {
-			$backoff_factor = $this->last_heartbeat['backoff_factor'];
-
-			// Ensure we don't go past the max, and support future decreases should they occur.
-			if ( $backoff_factor > self::MAX_BACKOFF_FACTOR ) {
-				$backoff_factor = self::MAX_BACKOFF_FACTOR;
-			}
+			$backoff_factor = min( $this->last_heartbeat['backoff_factor'], self::MAX_BACKOFF_FACTOR );
 
 			if ( $backoff_factor > 0 ) {
 				$seconds_elapsed = time() - $this->last_heartbeat['timestamp'];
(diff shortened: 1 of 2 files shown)


vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
```diff
@@ -212,11 +210,15 @@ private function update_backoff_factor(): void {
 		$backoff_factor = isset( $this->last_heartbeat['backoff_factor'] ) ? (int) $this->last_heartbeat['backoff_factor'] : 0;
 
 		if ( $backoff_factor >= self::MAX_BACKOFF_FACTOR ) {
-			return;
+			$backoff_factor = self::MAX_BACKOFF_FACTOR;
 		} elseif ( $backoff_factor <= 0 ) {
 			$backoff_factor = 1;
 		} else {
 			$backoff_factor = $backoff_factor * 2;
```
rebeccahum: mek one liner for me brain:
```suggestion
			$backoff_factor = min( $backoff_factor * 2, self::MAX_BACKOFF_FACTOR );
```

vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
```diff
@@ -212,11 +210,15 @@ private function update_backoff_factor(): void {
...
+
+			if ( $backoff_factor > self::MAX_BACKOFF_FACTOR ) {
```
rebeccahum:
```suggestion

```

vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
```diff
@@ -212,11 +210,15 @@ private function update_backoff_factor(): void {
...
+				$backoff_factor = self::MAX_BACKOFF_FACTOR;
```
rebeccahum:
```suggestion

```

vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
```diff
@@ -212,11 +210,15 @@ private function update_backoff_factor(): void {
 		$backoff_factor = isset( $this->last_heartbeat['backoff_factor'] ) ? (int) $this->last_heartbeat['backoff_factor'] : 0;
 
 		if ( $backoff_factor >= self::MAX_BACKOFF_FACTOR ) {
-			return;
+			$backoff_factor = self::MAX_BACKOFF_FACTOR;
 		} elseif ( $backoff_factor <= 0 ) {
 			$backoff_factor = 1;
 		} else {
 			$backoff_factor = $backoff_factor * 2;
+
```
rebeccahum:
```suggestion

```

vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
WPprodigy: `+4 -18` xD

PR #4973, opened 2023-10-24
CANTINA-954: Security: Use ambiguous error message in forgot password

rebeccahum:
## Description
We should use an ambiguous error message in the Forgot Password, since the messages confirm if a user has entered a correct username or email in the "Forgot Password" form.

### Updated: Forgot Password

When users hit the Forgot Password in wp-login, it will no longer confirm if their email or username exists.

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test
1) Go to wp-admin > Forgot password
2) Test with an invalid email e.g. `test@test.com`
3) Test with a valid username e.g. `vipgo`
4) Both should display the same error message


sjinks (approved):

security/login-error.php
```diff
@@ -1,13 +1,17 @@
 <?php
 namespace Automattic\VIP\Security;
 
+use WP_Error;
+
+const FORGET_PWD_MESSAGE = 'If there is an account associated with the username/email address, you will receive an email with a link to reset your password.';
```
WPprodigy, line 6: This needs to be a translatable string :)

PR #5021, opened 2023-11-14
fix(files): account for instantiation of `WP_Filesystem_VIP` without arguments

sjinks:
## Description

See CANTINA-997.

When `WP_Filesystem()` is invoked without arguments, it instantiates `WP_Filesystem_VIP` without passing arguments it expects. As a result, the reference to `WP_Filesystem_Direct` becomes `null`, and all FS operations fail.

### Plugin Updated: VIP File Service

Fix a bug during the instantiation of `WP_Filesystem_VIP` without arguments.

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

See CANTINA-997.

### DIFF
--- a/files/class-wp-filesystem-vip.php
+++ b/files/class-wp-filesystem-vip.php
@@ -25,6 +25,10 @@ class WP_Filesystem_VIP extends \WP_Filesystem_Base {
 		$this->method = 'vip';
 		$this->errors = new WP_Error();
 
+		if ( ! is_array( $dependencies ) || 2 !== count( $dependencies ) ) {
+			$dependencies = request_filesystem_credentials( site_url() );
+		}
+
 		list( $filesystem_uploads, $filesystem_direct ) = $dependencies;
 
 		$this->uploads = $filesystem_uploads;


sjinks:
Related: #4970

GaryJones (commented):
Confirmed the following now works:

```
vip dev-env --slug=... shell
```
then
```
wp scaffold plugin foo
```

What doesn't yet work is `vip dev-env --slug=... exec -- wp scaffold plugin foo`.

files/class-wp-filesystem-vip.php
```diff
@@ -25,6 +25,10 @@ public function __construct( $dependencies ) {
 		$this->method = 'vip';
 		$this->errors = new WP_Error();
 
+		if ( ! is_array( $dependencies ) || 2 !== count( $dependencies ) ) {
+			$dependencies = request_filesystem_credentials( site_url() );
```
WPprodigy, line 29: https://developer.wordpress.org/reference/functions/request_filesystem_credentials/

> bool|array True if no filesystem credentials are required, false if they are required but have not been provided, array of credentials if they are required and have been provided.

Do we need to protect against boolean returns?

sjinks:
```sh
php -r 'error_reporting(-1); list($a, $b) = false;'
```

does not generate any warnings; we are probably safe.

WPprodigy (approved):

PR #5037, opened 2023-11-21
PIE-3814 Put Jetpack in offline mode when we're launching a site in the network

andrea-sdl:
## Description
This PR adds some changes to put Jetpack in offline mode when we're going through the launch flow.

To do so we're using the `launching` flag in the `vip_launch_tools` cache group.

When that is set to true, we'll put Jetpack in offline mode. This is to avoid sending wrong data to Jetpack while we're doing the Search&Replace.

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

1. Check out PR.
1. On a multisite, with Jetpack active, select a network site and run ` wp cache set launching true vip_launch_tools 0 --url=<networkSiteURL>`
1. For that site you should see Jetpack go into offline mode.
image
1. Other sites in the network should not be affected.
1. Remove the cache with `wp cache delete launching vip_launch_tools --url=<networkSiteURL>`. Jetpack should be active again on that specific site.


vip-jetpack/vip-jetpack.php
```diff
@@ -409,3 +409,36 @@ function vip_filter_plugin_version_jetpack( $plugin_meta, $plugin_file ) {
...
+	// If not multisite, return the offline mode value.
+	if (! is_multisite() ){
+		return $offline_mode;
+	}
+
+	// We want to change the offline mode only if the customer enrolls into the IDC resolution
+	if ( ! defined( 'JETPACK_SYNC_IDC_OPTIN' ) || true !== constant( 'JETPACK_SYNC_IDC_OPTIN' ) ) {
+		return $offline_mode;
+	}
+	if ( wp_cache_get( 'launching', 'vip_launch_tools' ) ){
```
andrea-sdl: @WPprodigy I'm using the `wp_cache_get` here as you suggested. I wonder, though, aside from interacting with the WP CLI cache commands, could this option be cleared out by some other tools we have? If so, which ones?

I want to be sure we don't enter in online mode by accident when we're launching a site.

andrea-sdl:
@WPprodigy This PR is ready for review. I wonder if we should also add an admin notice for when we intentionally enable offline mode, so that customers aren't surprised when we enable it and they go to the JP page.

vip-jetpack/vip-jetpack.php
```diff
@@ -409,3 +409,30 @@ function vip_filter_plugin_version_jetpack( $plugin_meta, $plugin_file ) {
...
+	$vip_site_launching = wp_cache_get( 'launching', 'vip_launch_tools' );
+	if ( 'true' === $vip_site_launching || true === $vip_site_launching ) {
+		// enables jetpack offline mode
+		return true;
+	}
+	// keep the offline mode as it was.
+	return $offline_mode;
+}
+
+add_action( 'jetpack_offline_mode', 'vip_filter_jetpack_offline_mode_on_site_launch', PHP_INT_MAX, 1 );
```
rinatkhaziev:
```suggestion
add_filter( 'jetpack_offline_mode', 'vip_filter_jetpack_offline_mode_on_site_launch', PHP_INT_MAX, 1 );
```

rinatkhaziev (approved):

PR #5051, opened 2023-11-27
Parsely: Fix config values of Site Details Sync

mehmoodak:
## Description

If the plugin is managed via Platform then Site Details Service was not receiving correct values because `wp_parsely_credentials` filter was not executing, this has been fixed now.

## Changelog Description

Parsely: Fix config values of SIte Details Sync (if the plugin is managed via Platform)

## Pre-review checklist

## Pre-deploy checklist

- [n/a] VIP staff: Ensure any alerts added/updated conform to internal standards (see internal documentation). 

## Steps to Test

1. Check out PR.
1. There is no visible effect of this change so kindly verify through unit tests.

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -17,6 +17,8 @@
 
 namespace Automattic\VIP\WP_Parsely_Integration;
 
+use Parsely\Parsely;
+
 /**
  * The default version is the first entry in the SUPPORTED_VERSIONS list.
  */
@@ -60,13 +62,6 @@ final class Parsely_Loader_Info {
 	 * @var string
 	 */
 	private static string $version;
-
-	/**
-	 * Options of the plugin.
-	 *
-	 * @var array
-	 */
-	private static array $parsely_options;
 
 	/**
 	 * Check if the plugin is active.
@@ -144,7 +139,7 @@ final class Parsely_Loader_Info {
 		}
 
 		$configs = array();
-		$options = self::get_parsely_options() ?: [];
+		$options = self::get_parsely_options();
 
 		$configs['is_pinned_version']            = has_filter( 'wpvip_parsely_version' );
 		$configs['site_id']                      = $options['apikey'] ?? '';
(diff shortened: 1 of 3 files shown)


wp-parsely.php
```diff
@@ -179,11 +174,20 @@ public static function get_configs() {
...
+		$parsely = new Parsely();
+
+		/**
+		 * Parse.ly options.
+		 *
+		 * @var array
+		 */
+		$parsely_options = $parsely->get_options();
+
+		return $parsely_options;
```
mjangda: I think we can just use `$_GLOBALS['parsely']`?

```suggestion
		/**
		 * Parse.ly options.
		 *
		 * @var array
		 */
		$parsely_options = array();
		if ( isset( $GLOBALS['parsely'] ) && is_a( $GLOBALS['parsely'], 'Parsely\Parsely' ) ) {
			$parsely_options = $parsely->get_options();
		}
		
		return $parsely_options;
```

wp-parsely.php
```diff
@@ -17,6 +17,8 @@
 
 namespace Automattic\VIP\WP_Parsely_Integration;
 
+use Parsely\Parsely;
+
```
mjangda: This can be removed with my other suggested change.

```suggestion
```

mjangda (changes requested):

wp-parsely.php
mehmoodak: Addressed in https://github.com/Automattic/vip-go-mu-plugins/pull/5051/commits/053a7d6895695604ad842bf72e735ebdb13917a1

wp-parsely.php
mehmoodak: Addressed in https://github.com/Automattic/vip-go-mu-plugins/pull/5051/commits/053a7d6895695604ad842bf72e735ebdb13917a1

mjangda (approved):

PR #5098, opened 2023-12-13
Add retry limit to the connection pilot

WPprodigy:
Starting tracking the amount of retries, and implement a retry limit.

 Also tracking two different timestamps now instead of conflating the same one. `failure_timestamp` vs `success_timestamp`


rinatkhaziev (approved):
We've discussed replacing `$retry` with `$attempt` for better clarity, but otherwise looks good.

rebeccahum (approved):

PR #5107, opened 2023-12-19
(CANTINA-1006) Search: For CLI commands, use composition instead of inheritance

rebeccahum:
## Description
On a maintenance aspect, it makes sense to use composition instead of inheriting all of the commands into `vip-search` & overriding the ones we don't need (as they will still appear in the list when we do `wp vip-search`). 

This should not change anything functionally, except for:
- the subcommand, `wp vip-search get-algorithm-version`, which will return the actual number rather than `default` as a value
- for multi-sites using `--network-mode` and `--setup`, fixes a minor bug where it prompts via `confirm_destructive_operation()` twice (once before the command is run at all, and then, again for each subsite) to only prompt once per each subsite.

Remove inheritance from ElasticPress commands and fix CLI command `get-algorithm-version` to return a numeric value.

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test
1) Test all commands listed below and ensure they work the same as before:
```
─$ vip dev-env --slug es-ms-site exec -- wp vip-search
Running validation steps...
✓ Check for Docker installation
✓ Check Docker connectivity
✓ Check DNS resolution
usage: wp vip-search activate-feature <feature-slug>
   or: wp vip-search clean-ep-feature-settings
   or: wp vip-search clear-index
   or: wp vip-search deactivate-feature <feature-slug>
   or: wp vip-search delete-index
   or: wp vip-search documents <command>
   or: wp vip-search get-algorithm-version
   or: wp vip-search get-index-settings <type> [--version] [--format=<string>]
   or: wp vip-search get-indexes
   or: wp vip-search get-indexing-status [--pretty]
   or: wp vip-search get-last-index [--pretty]
   or: wp vip-search get-last-indexed-post-id
   or: wp vip-search get-mapping
   or: wp vip-search health <command>
   or: wp vip-search index [--setup] [--network-wide] [--blog-ids] [--per-page] [--nobulk] [--show-errors] [--offset] [--upper-limit-object-id] [--lower-limit-object-id] [--indexables] [--show-bulk-errors] [--show-nobulk-errors] [--post-type] [--include] [--post-ids] [--version] [--skip-confirm] [--using-versions]
   or: wp vip-search index-versions <command>
   or: wp vip-search list-features [--all]
   or: wp vip-search put-mapping [--network-wide] [--indexables] [--ep-host] [--ep-prefix] [--skip-confirm]
   or: wp vip-search queue <command>
   or: wp vip-search recreate-network-alias
   or: wp vip-search stop-indexing

See 'wp help vip-search <command>' for more information on a specific command.
```

### DIFF
--- a/search/includes/classes/commands/class-corecommand.php
+++ b/search/includes/classes/commands/class-corecommand.php
@@ -6,14 +6,18 @@ use WP_CLI;
 use WP_CLI\Utils;
 use ElasticPress\Elasticsearch;
 
-use function Automattic\VIP\Logstash\log2logstash;
-
 /**
  * Core commands for interacting with VIP Search
  *
  * @package Automattic\VIP\Search
  */
-class CoreCommand extends \ElasticPress\Command {
+class CoreCommand {
+	private $ep_command;
+
+	public function __construct() {
+		$this->ep_command = new \ElasticPress\Command();
+	}
+
 	private function verify_arguments_compatibility( $assoc_args ) {
 		if ( array_key_exists( 'version', $assoc_args ) && array_key_exists( 'using-versions', $assoc_args ) ) {
 			WP_CLI::error( 'The --version argument is not allowed when specifying --using-versions' );
@@ -285,9 +289,7 @@ class CoreCommand extends \ElasticPress\Command {
 				switch_to_blog( (int) $blog_id );
 				$assoc_args['url'] = home_url();
 				WP_CLI::line( "* Indexing blog {$blog_id}: {$assoc_args['url']}" );
-				WP_CLI::runcommand( 'vip-search index ' . Utils\assoc_args_to_str( $assoc_args ), [
-					'exit_error' => false,
-				] );
+				$this->ep_command->index( $args, $assoc_args );
 				Utils\wp_clear_object_cache();
 				restore_current_blog();
 			}
@@ -315,8 +317,7 @@ class CoreCommand extends \ElasticPress\Command {
 				]
 			);
 
-			array_unshift( $args, 'elasticpress', 'index' );
-			WP_CLI::run_command( $args, $assoc_args );
+			$this->ep_command->index( $args, $assoc_args );
 		}
 
 		if ( $using_versions ) {
@@ -337,7 +338,7 @@ class CoreCommand extends \ElasticPress\Command {
 	 */
 	public function put_mapping( $args, $assoc_args ) {
 		self::confirm_destructive_operation( $assoc_args );
-		parent::put_mapping( $args, $assoc_args );
+		$this->ep_command->put_mapping( $args, $assoc_args );
 	}
 
 	/**
@@ -383,11 +384,8 @@ class CoreCommand extends \ElasticPress\Command {
 	 * Throw error when delete-index command is attempted to be used.
 	 *
 	 * @subcommand delete-index
-	 *
-	 * @param array $args Positional CLI args.
-	 * @param array $assoc_args Associative CLI args.
 	 */
-	public function delete_index( $args, $assoc_args ) {
+	public function delete_index() {
 		WP_CLI::error( 'Please use index versioning to manage your indices: https://docs.wpvip.com/how-tos/vip-search/version-with-enterprise-search/' );
 	}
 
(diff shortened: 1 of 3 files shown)


search/includes/classes/commands/class-corecommand.php
```diff
@@ -532,7 +526,7 @@ public function get_last_indexed_post_id( $args, $assoc_args ) {
 	 * @param array $args Positional CLI args.
 	 * @param array $assoc_args Associative CLI args.
```
GaryJones: These param docs need deleting?

GaryJones (commented):
I'm in favour of this change in general. It looks like a classic use of the [Decorator pattern](https://refactoring.guru/design-patterns/decorator/php/example#lang-features) or [Proxy pattern](https://refactoring.guru/design-patterns/proxy/php/example), and as such, I'd suggest a couple of changes:

- Instead of hard-coding the instantiating of the wrapped object in the constructor, pass an already-instantiated `\ElasticPress\Command()` object in via the constructor parameter. 

```php
	public function __construct( \ElasticPress\Command $ep_command ) {
		$this->ep_command = $ep_command;
	}
```
The typing will be against a third-party class, not an interface, but that's fine here (especially for a Proxy). This composition-by-param-type is slightly better than composition-by-newing when it comes to testing, etc.

As such, [this line](https://github.com/Automattic/vip-go-mu-plugins/blob/694236607fa29180a573c8e76cc4c4538fc98cdd/search/includes/classes/class-search.php#L678C4-L678C81) would change to pass in an object rather than a callable:

```php
$vip_search_core_command = new \Automattic\VIP\Search\Commands\CoreCommand( new \ElasticPress\Command() );
WP_CLI::add_command( 'vip-search', $vip_search_core_command );
```

- Don't change the signature (removed params) from the public methods - it should be consistent with the wrapped object, even if the params are not used in our implementation (decorating/proxy class).

search/includes/classes/commands/class-corecommand.php
rebeccahum: Looks like it, good catch!

search/includes/classes/commands/class-corecommand.php
```diff
@@ -383,11 +388,8 @@ public function get_index_settings( $args, $assoc_args ) {
 	 * Throw error when delete-index command is attempted to be used.
 	 *
 	 * @subcommand delete-index
-	 *
-	 * @param array $args Positional CLI args.
-	 * @param array $assoc_args Associative CLI args.
```
GaryJones: Now that the  method args have been kept, then these (multiple) DocBlocks no longer need the `@param` tags removing.

search/includes/classes/commands/class-corecommand.php
```diff
@@ -399,7 +401,7 @@ public function delete_index( $args, $assoc_args ) {
 	 * @param array $assoc_args arguments that were passed to the caller command.
 	 * @return void
 	 */
-	public static function confirm_destructive_operation( array $assoc_args ) {
```
GaryJones: Why remove the `array` type? The next line tries to access the argument as though it's an array.

tests/search/e2e/integration/wp-cli.spec.js
```diff
@@ -176,34 +176,10 @@ describe('WP-CLI Commands', () => {
...
-		cy.wpCli('wp vip-search set-algorithm-version --version=1.0.0')
-			.its('stdout')
-			.should('contain', 'Done');
-
-		cy.wpCli('wp vip-search get-algorithm-version').its('stdout').should('contain', '1.0.0');
-
-		cy.wpCli('wp vip-search set-algorithm-version', true)
-			.its('stderr')
-			.should('contain', 'This command expects a version number or the --default flag');
-	});
```
GaryJones, line 197: Why remove these tests?

rebeccahum: Because we've disabled the `set-algorithm-version` command. If the algorithm version needs to be set, it should be done by filter.

GaryJones (approved):
Not tested, but suggestions on architectural changes look to have been incorporated ✅

sjinks (approved):

PR #5182, opened 2024-01-24
FORNO-1745: Handle SMTP mail failure when sending mail from unmapped domains

aagam-shah:
## Description

This PR adds handling for mail failures from VIP SMTP Server if they are not allowed to send mail from domains registered in the application. We are temporarily adding a fallback which will rewrite the From mail to `donotreply@wpvip.com` if mail is sent from unmapped domains.


vip-mail.php
```diff
@@ -139,6 +140,35 @@ public function filter_wp_mail_from() {
...
+
+	/**
+	* This function is used to handle the wp_mail failures if the sender address is rejected by the SMTP server
+	* @param WP_Error $error The WP_Error object passed by reference
+	 */
+	public function handle_wp_mail_failures( $error ) {
+		if ( isset( $error->error_data['wp_mail_failed'] ) && isset( $error->error_data['wp_mail_failed']['phpmailer_exception_code'] ) && isset( $error->errors['wp_mail_failed'] ) ) {
+			$error_data = $error->error_data['wp_mail_failed'];
+
+			// The phpmailer exception code for Sender Address rejection is 1 and we also are validating the message is matching to the one that's expected
```
saroshaga, line 152: so we don't need to read the 553 error code?

aagam-shah: Correct, unfortunately that exact error code is not passed to WP Error

saroshaga: so we need to rely on the `Sender address rejected: not owned by` error string?

aagam-shah: That message and also additional `phpmailer_exception_code`

saroshaga (approved):
Thanks for working on this @aagam-shah & @t-wright ❤️

vip-mail.php
```diff
@@ -139,6 +140,35 @@ public function filter_wp_mail_from() {
...
+	* This function is used to handle the wp_mail failures if the sender address is rejected by the SMTP server
+	* @param WP_Error $error The WP_Error object passed by reference
+	 */
+	public function handle_wp_mail_failures( $error ) {
+		if ( defined( 'VIP_SMTP_ENABLED' ) && true === constant( 'VIP_SMTP_ENABLED' ) && isset( $error->error_data['wp_mail_failed'] ) && isset( $error->error_data['wp_mail_failed']['phpmailer_exception_code'] ) && isset( $error->errors['wp_mail_failed'] ) ) {
+			$error_data = $error->error_data['wp_mail_failed'];
+
+			// The phpmailer exception code for Sender Address rejection is 1 and we also are validating the message is matching to the one that's expected
+			if ( 1 === $error_data['phpmailer_exception_code'] && false !== strpos( $error->errors['wp_mail_failed'][0], 'Sender address rejected: not owned by user' ) ) {
+					$to          = is_array( $error_data['to'] ) ? array_map( 'esc_html', $error_data['to'] ) : esc_html( $error_data['to'] );
```
chriszarate, line 154: @aagam-shah @t-wright Are you sure you want to apply `esc_html` here, and for what purpose? There are characters like `&` that are perfectly valid in email addresses that will get escaped.

https://en.wikipedia.org/wiki/Email_address#Local-part

vip-mail.php
```diff
@@ -139,6 +140,35 @@ public function filter_wp_mail_from() {
...
+					$subject     = esc_html( $error_data['subject'] );
+					$message     = esc_html( $error_data['message'] );
+					$headers     = is_array( $error_data['headers'] ) ? array_map( 'esc_html', $error_data['headers'] ) : esc_html( $error_data['headers'] );
```
chriszarate, line 157: All of these really, not sure what `esc_html` is guarding against.

rinatkhaziev: Yeah that was my feedback, but on top of it, it'll also mangle the original message.

saroshaga: @chriszarate : the usage of `esc_html` on email body itself might have been an attempt to enforce the following

> VIP’s infrastructure for message delivery is not intended for the distribution of HTML emails
- https://docs.wpvip.com/wordpress-on-vip/email/#h-limitations

But yeah we probably didn't need to do it for the other things.

Running `esc_url` on the attachments also should be fine I guess? But Todd can speak to the exact motivation here.

rinatkhaziev: I don't think "not intended to" means "we prohibit you from" cause that's not true. I just sent something to you as a test, not a phishing attempt!

PR #5202, opened 2024-01-30
FORNO-1745: Remove use of `esc_html()` when handling mail failures

t-wright:
## Description

In https://github.com/Automattic/vip-go-mu-plugins/pull/5182 we added handling for failures when sending mail from unmapped domains. In this PR we're removing the use of `esc_html` on the email body and other fields which was added in that PR since the use of this function could lead to undesirable escaping in this context.

### DIFF
--- a/vip-mail.php
+++ b/vip-mail.php
@@ -142,8 +142,9 @@ final class VIP_SMTP {
 
 
 	/**
-	* This function is used to handle the wp_mail failures if the sender address is rejected by the SMTP server
-	* @param WP_Error $error The WP_Error object passed by reference
+	 * This function is used to handle the wp_mail failures if the sender address is rejected by the SMTP server
+	 *
+	 * @param WP_Error $error The WP_Error object passed by reference
 	 */
 	public function handle_wp_mail_failures( $error ) {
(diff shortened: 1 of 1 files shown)


sanmai (approved):
Reviewed the code, looks sane.

vip-mail.php
```diff
@@ -142,28 +142,34 @@ public function filter_wp_mail_from() {
...
-					$to          = is_array( $error_data['to'] ) ? array_map( 'esc_html', $error_data['to'] ) : esc_html( $error_data['to'] );
-					$subject     = esc_html( $error_data['subject'] );
-					$message     = esc_html( $error_data['message'] );
-					$headers     = is_array( $error_data['headers'] ) ? array_map( 'esc_html', $error_data['headers'] ) : esc_html( $error_data['headers'] );
-					$attachments = is_array( $error_data['attachments'] ) ? array_map( 'esc_url', $error_data['attachments'] ) : esc_url( $error_data['attachments'] );
+				$to          = $error_data['to'] ?? null;
+				$subject     = $error_data['subject'] ?? null;
+				$message     = $error_data['message'] ?? null;
+				$headers     = $error_data['headers'] ?? null;
+				$attachments = $error_data['attachments'] ?? null;
```
saroshaga, line 159: We're removing `esc_url` from here

t-wright: Yes, I think that one should have been okay, but probably unnecessary.

saroshaga: got it! thanks!

saroshaga (approved):

PR #5265, opened 2024-02-21
Filter the font directory to work with the VIP Filesystem

noahtallen:
## Description
WordPress 6.5 [introduces the Font Library](https://make.wordpress.org/test/2023/10/03/help-test-the-font-library/), which allows customers to upload custom fonts and use them with global styles.

We need to use the new `font_dir` filter to ensure this will work in the VIP FS, since the default font dir is not writable on our platform.

This ultimately changes the font dir and URL to:

Normal/root site:
- `vip://wp-content/uploads/fonts`
- `https://$site_url/wp-content/uploads/fonts`

Network site:
- `vip://wp-content/uploads/sites/$network_site_id/fonts`
- `https://$site_url/$network_site_path/wp-content/uploads/sites/$network_site_id/fonts`

### Use a custom font directory for the new Font Library
WordPress 6.5 [introduces the Font Library](https://make.wordpress.org/test/2023/10/03/help-test-the-font-library/), which allows you to upload and manage your own font files from the site editor. We have used the new `font_dir` filter to set the font upload directory to `wp-content/uploads/fonts` to get this feature working with the VIP Filesystem.

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test
1. Sandbox a VIP site and copy these changes to mu-plugins on that site.
2. From the site editor, choose the global styles sidebar, and then typography.
3. Click the settings icon.
4. Upload a font that you don't have installed on your computer.
5. Use the font for, say, site headings in global styles.
6. Reload the editor, and the new font should display there.
7. View the frontend of the site and the new font should display there.

This recording shows the process:

https://github.com/Automattic/vip-go-mu-plugins/assets/6265975/54cad5ef-e199-4d35-af26-c6847369a556

### DIFF
--- a/files/class-vip-filesystem.php
+++ b/files/class-vip-filesystem.php
@@ -468,13 +468,7 @@ class VIP_Filesystem {
 	 * Changes the Font Library directory to work with the VIP Filesystem.
 	 */
 	public function filter_wp_font_dir( $defaults ) {
-		// Without removing the filter, there will be an infinite loop, because
-		// calling wp_upload_dir will trigger upload_dir, which includes wp_get_font_dir
-		// in the context of uploading fonts. That ultimately calls this function
-		// again through the font_dir filter.
-		remove_filter( 'upload_dir', 'wp_get_font_dir' );
 		$upload_dir = wp_upload_dir();
-		add_filter( 'upload_dir', 'wp_get_font_dir' );
 	
 		$defaults['basedir'] = $upload_dir['basedir'] . '/fonts';
 		$defaults['baseurl'] = $upload_dir['baseurl'] . '/fonts';
--- a/files/class-vip-filesystem.php
+++ b/files/class-vip-filesystem.php
@@ -100,6 +100,7 @@ class VIP_Filesystem {
 		add_filter( 'get_attached_file', [ $this, 'filter_get_attached_file' ], 20 );
 		add_filter( 'wp_generate_attachment_metadata', [ $this, 'filter_wp_generate_attachment_metadata' ], 10, 2 );
 		add_filter( 'wp_read_image_metadata', [ $this, 'filter_wp_read_image_metadata' ], 10, 2 );
+		add_filter( 'font_dir', [ $this, 'filter_wp_font_dir' ], 10, 1 );
 
 		/**
 		 * The core's function recurse_dirsize would call to opendir() which is not supported by the
@@ -124,6 +125,7 @@ class VIP_Filesystem {
 		remove_filter( 'get_attached_file', [ $this, 'filter_get_attached_file' ], 20 );
 		remove_filter( 'wp_generate_attachment_metadata', [ $this, 'filter_wp_generate_attachment_metadata' ] );
 		remove_filter( 'wp_read_image_metadata', [ $this, 'filter_wp_read_image_metadata' ], 10, 2 );
+		remove_filter( 'font_dir', [ $this, 'filter_wp_font_dir' ], 10, 1 );
 		remove_filter( 'pre_recurse_dirsize', '__return_zero' );
 	}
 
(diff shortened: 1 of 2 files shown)


noahtallen:
I think codecov might not realize the test ultimately executes the code via a filter :)

files/class-vip-filesystem.php
```diff
@@ -461,4 +463,24 @@ public function filter_wp_read_image_metadata( $meta, $file ) {
...
+	/**
+	 * Changes the Font Library directory to work with the VIP Filesystem.
+	 */
+	public function filter_wp_font_dir( $defaults ) {
+		// Without removing the filter, there will be an infinite loop, because
+		// calling wp_upload_dir will trigger upload_dir, which includes wp_get_font_dir
+		// in the context of uploading fonts. That ultimately calls this function
+		// again through the font_dir filter.
+		remove_filter( 'upload_dir', 'wp_get_font_dir' );
+		$upload_dir = wp_upload_dir();
```
noahtallen, line 471: May be worth switching to `doing_filter` instead, though that may get rolled into core to fix the infinite loop (see https://wordpress.slack.com/archives/C02RQBWTW/p1709082902708339?thread_ts=1708987810.984969&cid=C02RQBWTW)

noahtallen: Actually, this infinite loop bug will be fixed here: https://github.com/WordPress/gutenberg/pull/58839, so this won't be needed in the near future.

tests/files/test-vip-filesystem.php
```diff
@@ -386,4 +386,23 @@ public function data_get_transport_for_path(): iterable {
...
 	}
+
+	public function test_wp_font_dir() {
+		// Only available in WP 6.5 and newer:
+		if ( ! function_exists( '\wp_get_font_dir' ) ) {
+			$this->assertEquals( true, true );
+			return;
+		}
+
+		$font_dir = \wp_get_font_dir();
```
noahtallen, line 397: We need to follow this PR: https://github.com/WordPress/wordpress-develop/pull/6211

In case the function is renamed/changed. (Or test it by doing the filter instead.)

peterwilsoncc: > We need to follow this PR: [WordPress/wordpress-develop#6211](https://github.com/WordPress/wordpress-develop/pull/6211)
> 
> In case the function is renamed/changed. (Or test it by doing the filter instead.)

I'll try to remember to come back and let you know.

I've been doing some expreiments on my local build and don't intend to change the name of the function or the filter but will confirm.

noahtallen: Thanks!

tests/files/test-vip-filesystem.php
```diff
@@ -386,4 +386,23 @@ public function data_get_transport_for_path(): iterable {
...
+		// Only available in WP 6.5 and newer:
+		if ( ! function_exists( '\wp_get_font_dir' ) ) {
+			$this->markTestSkipped( 'test_wp_font_dir does not need to run for WP < 6.5.' );
+			return;
+		}
+
+		$font_dir = \wp_get_font_dir();
+
+		$this->assertEquals( $font_dir, [
+			'path'    => 'vip://wp-contentt/uploads/fonts',
```
rinatkhaziev: @sjinks do you happen to know where this double `t` in `wp-content` coming from?

noahtallen: I was able to verify it happens here:

https://github.com/Automattic/vip-go-mu-plugins/blob/9cbad3ef1c15019c1ed5c0edf8a31dac6f88f1ae/files/class-vip-filesystem.php#L159-L167

If both `strlen( constant( 'WP_CONTENT_DIR' ) )` are changed to `strlen( constant( 'WP_CONTENT_DIR' ) ) + 1`, then it goes back to single t

noahtallen: The root cause here is that WP_CONTENT_DIR constant can be set to a couple different things. So in `$params['path']`, it's `/tmp/...`, but in the constant itself, it's `/wp/...`. Since that's one character shorter, we see the off-by-one issue in the final string.

Not sure exactly why two different constants come into play, but this test file does set the `/wp` version, whereas most of the other tests have the `/tmp` version.

rinatkhaziev (approved):

tests/files/test-vip-filesystem.php
noahtallen: for posterity, I fixed this by changing the WP_CONTENT_DIR  constant used by the vip-filesystem test to the `/tmp` version, which matches the constant elsewhere in the test suite

PR #5272, opened 2024-02-23
Use a relative WPMU_PLUGIN_DIR config for VIP mu-plugins

luismulinari:
## Description

Since VIP mu-plugins can live in a non-standard directory (`wp-content/mu-plugins`), this PR enforces VIP mu-plugins to be loaded from a custom mu-plugins directory config called `WPVIP_MU_PLUGIN_DIR`, which is auto-defined to be the directory where VIP mu-plugins are.

## Pre-review checklist

## Pre-deploy checklist

### DIFF
--- a/cache/README.md
+++ b/cache/README.md
@@ -33,7 +33,7 @@ Let's add this to a new file in `client-mu-plugins` called `beta.php` where we'l
 <?php
 
 // Load the VIP Vary_Cache class
-require_once( WPMU_PLUGIN_DIR . '/cache/class-vary-cache.php' );
+require_once( WPVIP_MU_PLUGIN_DIR . '/cache/class-vary-cache.php' );
 
 use Automattic\VIP\Cache\Vary_Cache;
 ```
--- a/lib/sunrise/sunrise.php
+++ b/lib/sunrise/sunrise.php
@@ -7,7 +7,7 @@ if ( ! defined( 'ABSPATH' ) || ! is_multisite() ) {
 	return;
 }
 
-$mu_plugin_dir = defined( 'WPVIP_MU_PLUGIN_DIR' ) ? constant( 'WPVIP_MU_PLUGIN_DIR' ) : constant( 'WP_CONTENT_DIR' ) . '/mu-plugins';
+$mu_plugin_dir = realpath( __DIR__ . '/../..' );
 
 if ( defined( 'VIP_IS_MULTIPLE_DATASETS_ENABLED' ) && VIP_IS_MULTIPLE_DATASETS_ENABLED ) {
 	require_once $mu_plugin_dir . '/lib/db-multiple-datasets-config.php';
@@ -75,7 +75,7 @@ function handle_not_found_error( $error_type ) {
 
 	$is_web_request = Context::is_web_request();
 	if ( $is_web_request ) {
-		$mu_plugin_dir       = defined( 'WPVIP_MU_PLUGIN_DIR' ) ? constant( 'WPVIP_MU_PLUGIN_DIR' ) : constant( 'WP_CONTENT_DIR' ) . '/mu-plugins';
+		$mu_plugin_dir       = realpath( __DIR__ . '/../..' );
 		$is_maintenance_mode = Context::is_maintenance_mode();
 		if ( $is_maintenance_mode ) {
 			// 503 prevents page from being cached.
--- a/lib/sunrise/sunrise.php
+++ b/lib/sunrise/sunrise.php
@@ -75,6 +75,8 @@ function handle_not_found_error( $error_type ) {
 
 	$is_web_request = Context::is_web_request();
 	if ( $is_web_request ) {
+		// Sunrise is loaded before any mu plugins, including 000-vip-init.php
+		// So we can't use the WPVIP_MU_PLUGIN_DIR constant here
 		$mu_plugin_dir       = realpath( __DIR__ . '/../..' );
 		$is_maintenance_mode = Context::is_maintenance_mode();
 		if ( $is_maintenance_mode ) {
--- a/lib/sunrise/sunrise.php
+++ b/lib/sunrise/sunrise.php
@@ -7,7 +7,7 @@ if ( ! defined( 'ABSPATH' ) || ! is_multisite() ) {
 	return;
 }
 
-$mu_plugin_dir = realpath( __DIR__ . '/../..' );
+$mu_plugin_dir = defined( 'WPVIP_MU_PLUGIN_DIR' ) ? constant( 'WPVIP_MU_PLUGIN_DIR' ) : constant( 'WP_CONTENT_DIR' ) . '/mu-plugins';
 
 if ( defined( 'VIP_IS_MULTIPLE_DATASETS_ENABLED' ) && VIP_IS_MULTIPLE_DATASETS_ENABLED ) {
 	require_once $mu_plugin_dir . '/lib/db-multiple-datasets-config.php';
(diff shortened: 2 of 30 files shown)


vaultpress/cron-tasks.php
```diff
@@ -84,9 +84,9 @@ function _scan_site() {
 			if ( is_dir( WP_PLUGIN_DIR ) && strpos( realpath( WP_PLUGIN_DIR ), realpath( WP_CONTENT_DIR ) . DIRECTORY_SEPARATOR ) !== 0 && strpos( realpath( WP_PLUGIN_DIR ), realpath( ABSPATH ) . DIRECTORY_SEPARATOR ) !== 0 )
 				$paths['plugins'] = new VP_FileScan( WP_PLUGIN_DIR, $ignore_symlinks );
 
-			// Is WPMU_PLUGIN_DIR inside ABSPATH or WP_CONTENT_DIR?
-			if ( is_dir( WPMU_PLUGIN_DIR ) && strpos( realpath( WPMU_PLUGIN_DIR ), realpath( WP_CONTENT_DIR ) . DIRECTORY_SEPARATOR ) !== 0 && strpos( realpath( WPMU_PLUGIN_DIR ), realpath( ABSPATH ) . DIRECTORY_SEPARATOR ) !== 0 )
-				$paths['mu-plugins'] = new VP_FileScan( WPMU_PLUGIN_DIR, $ignore_symlinks );
+			// Is WPVIP_MU_PLUGIN_DIR inside ABSPATH or WP_CONTENT_DIR?
+			if ( is_dir(  WPVIP_MU_PLUGIN_DIR ) && strpos( realpath( WPVIP_MU_PLUGIN_DIR ), realpath( WP_CONTENT_DIR ) . DIRECTORY_SEPARATOR ) !== 0 && strpos( realpath( WPVIP_MU_PLUGIN_DIR ), realpath( ABSPATH ) . DIRECTORY_SEPARATOR ) !== 0 )
+				$paths['mu-plugins'] = new VP_FileScan( WPVIP_MU_PLUGIN_DIR, $ignore_symlinks );
```
WPprodigy: We shouldn't change this one, it's VaultPress wanting to scan site code, of which we should let it keep looking into `WPMU_PLUGIN_DIR` as that's more important than scanning our future custom place for vip mu plugins.

Also, VP should maybe be gone by now anyway 🙃

cache/README.md
```diff
@@ -33,7 +33,7 @@ Let's add this to a new file in `client-mu-plugins` called `beta.php` where we'l
 <?php
 
 // Load the VIP Vary_Cache class
-require_once( WPMU_PLUGIN_DIR . '/cache/class-vary-cache.php' );
+require_once( WPVIP_MU_PLUGIN_DIR . '/cache/class-vary-cache.php' );
```
WPprodigy, line 37: Please note that this is going to have to be something we account for in the migration. Any customer code currently doing things like this will break.

lib/sunrise/sunrise.php
```diff
@@ -7,7 +7,7 @@
 	return;
 }
 
-$mu_plugin_dir = defined( 'WPMU_PLUGIN_DIR' ) ? constant( 'WPMU_PLUGIN_DIR' ) : constant( 'WP_CONTENT_DIR' ) . '/mu-plugins';
+$mu_plugin_dir = defined( 'WPVIP_MU_PLUGIN_DIR' ) ? constant( 'WPVIP_MU_PLUGIN_DIR' ) : constant( 'WP_CONTENT_DIR' ) . '/mu-plugins';
```
WPprodigy, line 10: This will not work. Sunrise is loaded before any mu plugins, including `000-vip-init.php` where the constant is first defined.

We'll want to define this constant earlier, in wp-config or something.

query-monitor/classes/Util.php
```diff
@@ -118,9 +118,9 @@ public static function get_file_dirs() {
 			self::$file_dirs = apply_filters( 'qm/component_dirs', self::$file_dirs );
 
 			self::$file_dirs['plugin'] = WP_PLUGIN_DIR;
-			self::$file_dirs['mu-vendor'] = WPMU_PLUGIN_DIR . '/vendor';
-			self::$file_dirs['go-plugin'] = WPMU_PLUGIN_DIR . '/shared-plugins';
-			self::$file_dirs['mu-plugin'] = WPMU_PLUGIN_DIR;
+			self::$file_dirs['mu-vendor'] = WPVIP_MU_PLUGIN_DIR . '/vendor';
+			self::$file_dirs['go-plugin'] = WPVIP_MU_PLUGIN_DIR . '/shared-plugins';
+			self::$file_dirs['mu-plugin'] = WPVIP_MU_PLUGIN_DIR;
```
WPprodigy: This might be misleading, since we intend to support customer mu plugins at `WPMU_PLUGIN_DIR`.  So we should call this one `vip-mu-plugin` I think and leave `self::$file_dirs['mu-plugin'] = WPMU_PLUGIN_DIR;`.

cache/README.md
luismulinari: Good point.

lib/sunrise/sunrise.php
luismulinari: I fixed it on https://github.com/Automattic/vip-go-mu-plugins/pull/5272/commits/a4cc270ac25cb7580713cd0c9fa686010a97e168 by using a relative path to it. WDYT?

query-monitor/classes/Util.php
luismulinari: Nice catch!

I fixed it on https://github.com/Automattic/vip-go-mu-plugins/pull/5272/commits/a4cc270ac25cb7580713cd0c9fa686010a97e168 and just added the `self::$file_dirs['mu-plugin'] = WPMU_PLUGIN_DIR;` when `WPVIP_MU_PLUGIN_DIR !== WPVIP_MU_PLUGIN_DIR` to make sure we are not adding the directory twice when the are the same.

luismulinari:
Hey @WPprodigy ! Thanks a lot for the review. When you have some time, could you take a look again?

lib/sunrise/sunrise.php
```diff
@@ -7,7 +7,7 @@
 	return;
 }
 
-$mu_plugin_dir = defined( 'WPMU_PLUGIN_DIR' ) ? constant( 'WPMU_PLUGIN_DIR' ) : constant( 'WP_CONTENT_DIR' ) . '/mu-plugins';
+$mu_plugin_dir = realpath( __DIR__ . '/../..' );
```
sanmai: I suppose load-ordering issues prevent us from using `WPVIP_MU_PLUGIN_DIR` here. It could be helpful to add a comment to explain the rationale.

query-monitor/classes/Util.php
```diff
@@ -118,11 +118,15 @@ public static function get_file_dirs() {
...
-			self::$file_dirs['mu-vendor'] = WPMU_PLUGIN_DIR . '/vendor';
-			self::$file_dirs['go-plugin'] = WPMU_PLUGIN_DIR . '/shared-plugins';
-			self::$file_dirs['mu-plugin'] = WPMU_PLUGIN_DIR;
+			self::$file_dirs['vip-mu-vendor'] = WPVIP_MU_PLUGIN_DIR . '/vendor';
+			self::$file_dirs['vip-go-plugin'] = WPVIP_MU_PLUGIN_DIR . '/shared-plugins';
+			self::$file_dirs['vip-mu-plugin'] = WPVIP_MU_PLUGIN_DIR;
 			self::$file_dirs['vip-plugin'] = get_theme_root() . '/vip/plugins';
 
+			if ( WPVIP_MU_PLUGIN_DIR !== WPMU_PLUGIN_DIR ) {
+				self::$file_dirs['mu-plugin'] = WPMU_PLUGIN_DIR;
```
sanmai: It also looks like an opportunity for a comment with an explanation.

sanmai (approved):
Sounds like we are making mu-plugins work from a relative to the installation root directory, so we can have them installed in the non-default location.

Comments are non-blocking.

lib/sunrise/sunrise.php
WPprodigy: Mm, it doesn't feel super great but maybe is less flimsy than it first appears. If we were to define `WPVIP_MU_PLUGIN_DIR` earlier on though I do think that might be better.

luismulinari: Just added `WPVIP_MU_PLUGIN_DIR`.

We are going to set this constant on `wp-config.php`. Does it work?

WPprodigy: Yeah, i like that

PR #5297, opened 2024-02-29
Instead of adding user to superadmins option, filter

nickdaugherty:
## Description

It's easier to control which support users are super admins with a filter, rather than updating the site_admins option via grant_super_admin()

### Update: Improve how VIP Support Users are created

Tweak to VIP's internal support system to make management easier.

## Pre-review checklist

- n/a This change uses a rollout method to ease with deployment (if applicable - especially for large scale actions that require writes).

## Pre-deploy checklist

### DIFF
--- a/tests/vip-support/test-user-proxied.php
+++ b/tests/vip-support/test-user-proxied.php
@@ -7,6 +7,7 @@
 namespace Automattic\VIP\Support_User\Tests;
 
 use Automattic\Test\Constant_Mocker;
+use Automattic\VIP\Support_User\Role;
 use Automattic\VIP\Support_User\User;
 use WP_UnitTestCase;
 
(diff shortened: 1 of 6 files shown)


sanmai (dismissed):
It looks pretty cool to me, but let's get approval from someone else as well 🤞

brunobasto (dismissed):
Looks great! Thanks for working on this @nickdaugherty and @sanmai

WPprodigy:
I found that the user was still ending up in the `site_admins` site option due to being promoted to a super admin in other locations still. 

I removed the promotions to super admin in https://github.com/Automattic/vip-go-mu-plugins/pull/5297/commits/298c4b9c78e6d500bbc83c2aea0b1f8fd3cff4eb, but now it seems the filter is falling short of being enough here. The user appears as a super admin when in the network admin, but is unable to view individual subsite dashboards unless the user is directly added onto the site via network admin. From what I can see, this is because the user fails the `user_can_access_admin_page()` check due to the `$_wp_submenu_nopriv` global.

PR #5318, opened 2024-03-11
Introduce Attendant class for Jetpack user management

WPprodigy:
## Description

 Introduces new `Attendant` class that will be responsible for handling the jetpack connection user.

By default, sites will continue using the same `wpcomvip` user for connections. Upon enabling via the `vip_jetpack_connection_pilot_use_org_owner` filter, the new customer-specific user will be used as the connection owner. Once the filter is enabled, existing connections will fail the connection test with the `jp-cxn-pilot-has-legacy-vip-owner` error which will trigger a (re)connection attempt that will make the new user the owner. But until the filter is enabled, all will stay the same.

Additionally using this Attendant class to lock down the bot user with additional security restrictions. In the future, after all sites have switched over, we'll be able to phase out all the various `wpcomvip`-specific controls throughout the codebase and can even work towards removing that user from sites as it is no longer used for anything else.

### Plugin Updated: VIP Jetpack

Introduce new `Attendant` class for managing the Jetpack connection's primary user.

## Deploy Notes

There is a small change that needs to take place in the `Provisioner` class as well that must go out alongside this change.

## Steps to Test

Default User:
1) Create a new subsite
2) Run `wp jetpack-start connect --url=example.com/test1`
3) Ensure JP is connected underneath the `wpcomvip` user.

New User:
1) Add `add_filter( 'vip_jetpack_connection_pilot_use_org_owner', '__return_true' )` to the site (before plugins loaded priority 25)
2) Create another subsite
3) Run `wp jetpack-start connect --url=example.com/test2`
4) Ensure JP is connected underneath the new user.

Migration:
1) On the first subsite, ensure the filter is now present.
2) Disconnect JP: `wp jetpack disconnect blog --url=example.com/test1`
3) Re-connect JP: `wp jetpack-start connect --url=example.com/test1`
4) Ensure the new user is now the connection owner.


vip-jetpack/connection-pilot/class-jetpack-connection-attendant.php
```diff
@@ -0,0 +1,309 @@
...
+			] );
+
+			if ( is_wp_error( $result ) ) {
+				return new WP_Error( 'jp-cxn-pilot-owner-creation-failed', 'Failed to create new user.' );
+			}
+
+			$user = get_userdata( $result );
+			if ( ! $user ) {
+				return new WP_Error( 'jp-cxn-pilot-owner-creation-failed', 'Failed to create new user.' );
+			}
```
rinatkhaziev, line 114: It feels that the  second call (`get_userdata`) is a bit redundant? Do we expect some filters being in play? Cause otherwise we can just use the returned user id and the rest of the data from the original payload.

vip-jetpack/connection-pilot/class-jetpack-connection-attendant.php
```diff
@@ -0,0 +1,309 @@
...
+
+		return $this->get_org_owner_details();
+	}
+
+	private function get_org_owner_details() {
+		$org_id = defined( 'VIP_ORG_ID' ) ? constant( 'VIP_ORG_ID' ) : 0;
+
+		return [
+			'email'        => "vip-jetpack-owner+{$org_id}@wpvip.com",
+			'display_name' => 'VIP Jetpack Connection Owner',
```
rinatkhaziev:
```suggestion
			'display_name' => 'WPVIP Jetpack Connection Owner',
```

vip-jetpack/connection-pilot/class-jetpack-connection-attendant.php
```diff
@@ -0,0 +1,309 @@
...
+		}
+
+		return $this->get_org_owner_details();
+	}
+
+	private function get_org_owner_details() {
+		$org_id = defined( 'VIP_ORG_ID' ) ? constant( 'VIP_ORG_ID' ) : 0;
+
+		return [
+			'email'        => "vip-jetpack-owner+{$org_id}@wpvip.com",
```
rinatkhaziev, line 170: Do we really need `vip-` here? since we're already `@wpvip.com`

vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
```diff
@@ -373,6 +382,6 @@ protected function send_alert( $message = '', $wp_error = null ) {
 	}
 }
 
-add_action( 'init', function () {
+add_action( 'plugins_loaded', function () {
 	Connection_Pilot::instance();
 }, 25 );
```
rinatkhaziev, line 389: Any particular reason for it to be on `25`?

vip-jetpack/connection-pilot/class-jetpack-connection-attendant.php
```diff
@@ -0,0 +1,309 @@
...
+
+		// Add user to blog if needed, and ensure they are a super admin.
+		if ( is_multisite() ) {
+			$blog_id = get_current_blog_id();
+
+			if ( ! is_user_member_of_blog( $user->ID, $blog_id ) ) {
+				$added_to_blog = add_user_to_blog( $blog_id, $user->ID, $details['role'] );
+
+				if ( is_wp_error( $added_to_blog ) ) {
+					return new WP_Error( 'jp-cxn-pilot-owner-ms-add-failed', 'Failed to add user to blog.' );
```
rinatkhaziev: Should we pass through the explanation from the WP_Error? Looking at code it can be either `user_does_not_exist` or user_cannot_be_added (based on a filter return value)

vip-jetpack/connection-pilot/class-jetpack-connection-attendant.php
WPprodigy: `wp_insert_user()` only gives back an ID on success, whereas the rest of the function wants to utilize the user object (for `$user->user_email`, but mainly `$user->roles`).

It is nice as an extra check to make sure filters didn't break the user creation though.

vip-jetpack/connection-pilot/class-jetpack-connection-attendant.php
WPprodigy: Makes it extra VIP! 😆

vip-jetpack/connection-pilot/class-jetpack-connection-pilot.php
WPprodigy: Past guy forgot to mention why the 25 https://github.com/Automattic/vip-go-mu-plugins/pull/2119#discussion_r633858903, no idea 👀 

This PR moves the loading earlier from `init` to `plugins_loaded` though so that some of the security controls for the bot account are registered early enough. But that does mean that checking of the `VIP_JETPACK_AUTO_MANAGE_CONNECTION` constant and associated filter inside `should_run_connection_pilot()` will be happening quite earlier and could be confusing for users trying to change that in a theme/plugin. So gonna move that part of the logic back onto `init` at the same priority.

There's another `plugins_loaded` hook related to the attendant that I have going at priority 30. So no particular reason to be on 25 for the initiation of connection pilot, but I rather like just defining a priority for things like this rather than grouping together with tons of other things on the default. Helps make everything being purposely hooked after it a bit more clear, and also just leaves less up to chance on who registered what first.

vip-jetpack/connection-pilot/class-jetpack-connection-attendant.php
WPprodigy: Passing through the error code now 👍

vip-jetpack/connection-pilot/class-jetpack-connection-controls.php
```diff
@@ -124,8 +125,8 @@ private static function test_jetpack_owner_connection(): bool {
 	 * @return mixed bool|WP_Error True if JP was (re)connected, WP_Error otherwise.
 	 */
 	public static function connect_site( bool $skip_connection_tests = false, bool $disconnect = false ) {
-		if ( ! self::validate_constants() ) {
-			return new WP_Error( 'jp-cxn-pilot-missing-constants', 'This is not a valid VIP Go environment or some constants are missing.' );
+		if ( ! self::validate_environment() ) {
+			return new WP_Error( 'jp-cxn-pilot-invalid-environment', 'This is not a valid VIP Go environment.' );
```
rinatkhaziev:
```suggestion
			return new WP_Error( 'jp-cxn-pilot-invalid-environment', 'This is not a valid WPVIP environment.' );
```

rinatkhaziev (approved):

vip-jetpack/connection-pilot/class-jetpack-connection-attendant.php
```diff
@@ -0,0 +1,309 @@
...
+
+	/**
+	 * Check if the current Jetpack connection owner is the VIP user.
+	 *
+	 * @return string 'is_vip' | 'is_legacy_vip' | 'not_vip'
+	 */
+	public function check_connection_owner_validity() {
+		$expected_owner_details = $this->get_expected_owner_details();
+		$legacy_owner_details   = $this->get_legacy_owner_details();
+		$org_owner_details      = $this->get_org_owner_details();
```
sjinks, line 44: @WPprodigy `$org_owner_details` is not used by the code. Is it safe to remove this line?

PR #5341, opened 2024-03-18
BB8-11121: Prevent enabling debug mode on local env for logged out users

aswasif007:
## Description

We are updating the code to prevent debug mode from enabling automatically on local env for logged out users.

## Pre-review checklist

## Pre-deploy checklist


aswasif007:
I am not sure if this has other implications. @Automattic/vip-platform-cantina please advise!

rebeccahum:
@aswasif007 Per the BB8 comments, we shouldn't be removing this entirely locally, but only forcing it for users logged in locally I think.

rinatkhaziev (changes requested):
Needs further consideration.

WPprodigy:
It's a handy feature to be able to get QM for logged out requests. I don't think we should take that away. But at the same time, I agree that we shouldn't start QM for logged out by default since that's not normal behavior.

Maybe should be this instead:

```
if ( $is_nocache && $is_debug && ( $is_proxied || $is_local ) ) {
```

PR #5381, opened 2024-03-23
ci: use exact versions of actions

sjinks:
This is to avoid situations like #5365, when a workflow suddenly fails because one of the actions it depends upon has changed.

### DIFF
--- a/.github/workflows/core-tests.yml
+++ b/.github/workflows/core-tests.yml
@@ -42,14 +42,14 @@ jobs:
           echo "PHP_FPM_GID=$(id -g)" >> "${GITHUB_ENV}"
 
       - name: Checkout WordPress
-        uses: actions/checkout@v4.1.2 # v4.0.0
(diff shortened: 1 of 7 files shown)


.github/workflows/core-tests.yml
```diff
@@ -80,9 +80,9 @@ jobs:
         run: composer install -n
         working-directory: wordpress
 
-      - name: Downgrade PHPUnit
-        run: composer require --dev -n phpunit/phpunit:9.6.12
-        working-directory: wordpress
+      # - name: Downgrade PHPUnit
```
noahtallen, line 83: Do we require the correct phpunit version by default now?

noahtallen (approved):
Looks good!

.github/workflows/core-tests.yml
sjinks: I think so; 9.6.13 had some kind of a bug. The tests seem to work w/the latest version.

PR #5396, opened 2024-03-28
Update dependencies

mikebendt:
## Description

I noticed a few out-of-date dependencies with `npm outdated` - this brings them up-to-date.

### Dependencies

Update npm dependencies

## Pre-review checklist

- [x] This change works and has been tested locally (or has an appropriate fallback).
- [ ] This change works and has been tested on a Go sandbox.
- n/a This change has relevant unit tests (if applicable).
- n/a This change uses a rollout method to ease with deployment (if applicable - especially for large scale actions that require writes).
- n/a This change has relevant documentation additions / updates (if applicable).
- [x] I've created a changelog description that aligns with the provided examples.

## Pre-deploy checklist

- [ ] VIP staff: Ensure any alerts added/updated conform to internal standards (see internal documentation). 

## Steps to Test

Running `npm i` and then testing WP Admin functionality

PR #5397, opened 2024-03-31
[Do Not Merge] Attempt to Combat Chain-Supply Attacks

sjinks:
This PR is a proof-of-concept: the idea is to detect changes to the lock files and verify them with `composer install` or `npm install`. If those commands generate a different lock file, shout out loud.


sjinks:
5492d82 introduces a change to a `composer.lock`: it covertly modifies the download URL for one of the packages we use. Because GitHub by default does not show diffs for lock files, this change can easily go unnoticed.

See #5396

PR #5410, opened 2024-04-04
Update WordPress Importer plugin to include post_type in the query for existing post

dhusakovic:
## Description

The post import process uses [post_exists](https://developer.wordpress.org/reference/functions/post_exists/) function to check whether each of the importing posts already exists. The query utilises the title and date parameters for this search, which will run through all the posts to try and find a match.  

In cases where we have a large number of posts this process times out and the import fails. 

Passing the **post_type** to **post_exists** function allows us to use the **type_status_date** key, which significantly drops the number of rows examined and the overall query time in certain instances, depending on how many posts of that post_type we have.

## Changelog Description

Update to WordPress Importer plugin to improve the post_exists query time during post import process.

## Pre-review checklist

### DIFF
--- a/wordpress-importer/class-wp-import.php
+++ b/wordpress-importer/class-wp-import.php
@@ -656,7 +656,7 @@ class WP_Import extends WP_Importer {
 
 			$post_type_object = get_post_type_object( $post['post_type'] );
 
-			$post_exists = post_exists( $post['post_title'], '', $post['post_date'] );
+			$post_exists = post_exists( $post['post_title'], '', $post['post_date'], $post['post_type'] );
 
 			/**
 			* Filter ID of the existing post corresponding to post currently importing.


sjinks (approved):

rebeccahum (changes requested):
Hi @dhusakovic, thanks for the PR! Would it be possible to do the change upstream so that when we update it to the latest version, the change doesn't get overwritten? https://github.com/WordPress/wordpress-importer

dhusakovic:
@rebeccahum  I've created an issue and suggested the change here https://github.com/WordPress/wordpress-importer/issues/162. Can we still push the update here while we wait for the upstream changes?

rebeccahum:
@dhusakovic I'd prefer to keep it in sync with upstream to ensure it doesn't get overwritten with the way we have many moving parts in MU-plugins! However, I looked at the upstream issue and it looks like your suggestion [would be a welcome change in a PR](https://github.com/WordPress/wordpress-importer/issues/162#issuecomment-2044115739) 🙂.

rebeccahum:
Closing b/c it's in now per #5967

PR #5474, opened 2024-04-16
Fix changelog generator so that all PR commits are checked

noahtallen:
## Description
When working through a recent mu-plugins release, a changelog was not generated for production. After some digging, I realized the script was only fetching 30 of the many PRs associated, and so some PRs/commits were not being checked for chaneglogs. (So if the first 30 commits had no changelog entry, no post would be generated.)

Thankfully, this endpoint can be paginated with `page` and `per_page`, so we simply request with page++ until the page is empty. At that point, we have all the commits for the request.

## Steps to Test
First, create a `test.php` file, and copy both the `curl_get_all` and `curl_get` functions from this branch into that file. Then, add this test code:

```php
$commit_url = 'https://api.github.com/repos/Automattic/vip-go-mu-plugins/pulls/5472/commits';

$some_data = curl_get($commit_url);
print_r($some_data[0]);
echo( "\nPartial commits count: " . count( $some_data ) );

$all_data = curl_get_all( $commit_url );
print_r($all_data[0]);
echo( "\nAll commits count: " . count( $all_data ) );
```

You can compare the two commits to see that they are the same, and verify that more commits are requested using the new approach.

### DIFF
--- a/ci/changelog-summary.php
+++ b/ci/changelog-summary.php
@@ -329,7 +329,7 @@ function get_changelog_tags( $pr ) {
  * @return array $pr_ids The IDs pulled from the commits
  */
 function get_pr_ids_from_commits( $commit_url ) {
-    $commits = curl_get( $commit_url );
+    $commits = curl_get_all( $commit_url );
     $pr_ids = [];
 
     foreach( $commits as $commit ) {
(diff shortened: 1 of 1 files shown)


rinatkhaziev (approved):
thank you for the fix!

PR #5478, opened 2024-04-19
Add log suppression to VIP_Request_Block

maxschmeling:
## Description

This PR introduces a new method that allows suppression of the logs. In case it's too noisy a new helper can be used in vip-config.php

```php
/* vip-config.php */
// Supress the logs
VIP_Request_Block::should_log( false );

// Resume the logs
VIP_Request_Block::should_log( true );
```

### VIP_Request_Block Log Output Control

Added an option to suppress logging for requests blocked by `VIP_Request_Block`:

```php
/* In vip-config.php */

// Supress the logs
VIP_Request_Block::should_log( false );

// Resume the logs
VIP_Request_Block::should_log( true );
```

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test
1. Check out the PR
2. Spin up a dev environment
3. Add `VIP_Request_Block::should_log( false );` to `vip-config.php`
4. Add a block for a local request (I used user-agent, but there's probably a simpler option)
5. Make a request, verify it was blocked, then verify no log entry for the block
6. Remove the `VIP_Request_Block::should_log( false );` line, repeat step 5, verify the block was logged.

### DIFF
--- a/lib/class-vip-request-block.php
+++ b/lib/class-vip-request-block.php
@@ -17,6 +17,13 @@
  */
 class VIP_Request_Block {
 	/**
+	 * Suppress logging of blocked requests to reduce log noise.
+	 *
+	 * @var bool
+	 */
+	public static $suppressLog = false;
+
+	/**
 	 * Block a specific IP based either on true-client-ip, falling back to x-forwarded-for
 	 *
 	 * 🛑 BE CAREFUL: blocking a reverse proxy IP instead of the client's IP will result in legitimate traffic being blocked!!!
@@ -137,8 +144,11 @@ class VIP_Request_Block {
 			header( 'Expires: Wed, 11 Jan 1984 05:00:00 GMT' );
 			header( 'Cache-Control: no-cache, must-revalidate, max-age=0' );
 
-			// phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
-			error_log( sprintf( 'VIP Request Block: request was blocked based on "%s" with value of "%s"', $criteria, $value ) );
+			if ( ! self::$suppressLog ) {
+				// phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
+				error_log( sprintf( 'VIP Request Block: request was blocked based on "%s" with value of "%s"', $criteria, $value ) );
+			}
+
 			exit;
 		}
 
--- a/lib/class-vip-request-block.php
+++ b/lib/class-vip-request-block.php
@@ -166,7 +166,25 @@ class VIP_Request_Block {
 	 * @param bool $should_log whether to log blocked requests.
 	 * @return void
 	 */
-	public static function should_log( bool $should_log = true ) {
+	public static function toggle_logging( bool $should_log = true ) {
 		static::$should_log = $should_log;
+	}
+
+	/**
+	 * Enable logging of blocked requests.
+	 *
+	 * @return void
+	 */
+	public static function enable_logging() {
+		static::toggle_logging( true );
+	}
+
+	/**
+	 * Disable logging of blocked requests.
+	 *
+	 * @return void
+	 */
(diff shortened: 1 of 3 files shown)


maxschmeling:
I wanted to add a test for this change, but there currently isn't a test for the `error_log` call and from what Copilot told me, I'd have to introduce a wrapper for it that can be mocked and then pass that into `VIP_Request_Block `. I can do that, but wanted to get feedback first.

sjinks:
Yes, that is correct.

sjinks:
Maybe smth like

```php
public static function log( string $criteria, string $value ): void {
    // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
    error_log( sprintf( 'VIP Request Block: request was blocked based on "%s" with value of "%s"', $criteria, $value ) );
}
```

```diff

-				// phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
-				error_log( sprintf( 'VIP Request Block: request was blocked based on "%s" with value of "%s"', $criteria, $value ) );
+				static::log( $criteria, $value );
```

Then in the test:

```php
class MyRequestBlock extends VIP_Request_Block {
    static $log_called = false;

    public static function log( string $criteria, string $value ): void {
        self::$log_called = true;
    }
}
```

and in the test case in `setUp()` (or wherever is appropriate) you set `MyRequestBlock::$log_called = false;` and in a test case you do `static::assertTrue( MyRequestBlock::$log_called );` or `static::assertFalse( MyRequestBlock::$log_called );`

This is probably easier than mocking.

maxschmeling:
@sjinks I've tried a few different things but couldn't get the test to work. I haz no PHP skills.

If you could give me a little help getting that across the finish line, that would be much appreciated. Thanks

rinatkhaziev:
@yolih for awareness

lib/helpers/wp-cli-db/class-wp-cli-db.php
```diff
@@ -49,7 +49,6 @@ public function validate_subcommand( array $command ) {
 			'query',
 			'prefix',
 			'columns',
-			'size',
```
rebeccahum: I think this snuck in

lib/class-vip-request-block.php
```diff
@@ -137,11 +144,29 @@ public static function block_and_log( string $value, string $criteria ) {
...
+		error_log( sprintf( 'VIP Request Block: request was blocked based on "%s" with value of "%s"', $criteria, $value ) );
+	}
+
+	/**
+	 * Control logging of blocked requests.
+	 *
+	 * @param bool $should_log whether to log blocked requests.
+	 * @return void
+	 */
+	public static function should_log( bool $should_log = true ) {
```
rebeccahum: Minor nitpick, but I'm not a huge fan of the function name...I feel like we should rename it, since at first glance, `should_log()` reads to me like a function that returns on what the current logging status is, rather than that it controls it. Maybe a bit more tedious, but how would you feel about having two explicit functions, `enable_logging()` and `disable_logging()`?

rinatkhaziev: How about `toggle_logging`?

yolih: I like the suggestion for `enable_logging()` and `disable_logging()` because it's so clear and human-readable. It's also similar to how PHPCS scanning is enabled and disabled for sections of code (`phpcs:enable` and `phpcs:disable`).

lib/helpers/wp-cli-db/class-wp-cli-db.php
rinatkhaziev: wait, is it supposed to be there or no?

rebeccahum: no

rebeccahum (approved):

PR #5504, opened 2024-04-23
Integrations: Add Enterprise Search

rebeccahum:
## Description
This PR does two things:
- Adds ES as an integration
- Adds cascadingConfig ability to configs on multisites

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test
1) Using dev-env, create a multisite
2) Create another subsite to be site ID 2
3) Inside the PHP container, create `/wp/config/integrations-config/enterprise-search-config.php` (you will probably need to create the `integrations-config` folder) with contents:
```
<?php

return array(
  'env' => array(
  	'status' => 'disabled',
  	'config' => array (
  		'username' => 'test-username',
  		'password' => 'test-password',
  	),
	'cascade_config' => true,
  ),
  'network_sites' => array (
  	'1' => array (
  		'status' => 'enabled',
  		'config' => array (
  		)
  	),
  )
);
```
3) Verify integration being enabled on site 1, but not on 2 with the existence of CLI command `vip dev-env exec -- wp --url=<siteUrl> vip-search`

### DIFF
--- a/integrations/enterprise-search.php
+++ b/integrations/enterprise-search.php
@@ -33,4 +33,32 @@ class EnterpriseSearchIntegration extends Integration {
 
 		require_once __DIR__ . '/../search/search.php';
 	}
+
+	/**
+	 * Configure `Enterprise Search` for VIP Platform.
+	 */
+	public function configure(): void {
+		if ( $this->is_es_credentials_set() ) {
+			return;
+		}
+
+		add_action( 'vip_search_loaded', array( $this, 'vip_set_es_credentials' ) );
+	}
+
+	/**
+	 * Set the Elasticsearch credentials.
+	 */
+	public function vip_set_es_credentials(): void {
+		$config = $this->get_config();
+		if ( isset( $config['username'] ) && isset( $config['password'] ) ) {
+			define( 'VIP_ELASTICSEARCH_USERNAME', $config['username'] );
+			define( 'VIP_ELASTICSEARCH_PASSWORD', $config['password'] );
+		}
+	}
+
+	private function is_es_credentials_set(): bool {
+		$username_defined = defined( 'VIP_ELASTICSEARCH_USERNAME' ) && constant( 'VIP_ELASTICSEARCH_USERNAME' );
+		$password_defined = defined( 'VIP_ELASTICSEARCH_PASSWORD' ) && constant( 'VIP_ELASTICSEARCH_PASSWORD' );
+		return $username_defined && $password_defined;
+	}
 }
--- a/integrations/enterprise-search.php
+++ b/integrations/enterprise-search.php
@@ -16,7 +16,7 @@ class EnterpriseSearchIntegration extends Integration {
 	 * this function to prevent loading of integration again from platform side.
 	 */
 	public function is_loaded(): bool {
-		return class_exists( '\Automattic\VIP\Search\Search' );
+		return class_exists( \Automattic\VIP\Search\Search::class );
 	}
 
 	/**
--- a/integrations/enterprise-search.php
+++ b/integrations/enterprise-search.php
@@ -1,0 +1,40 @@
+<?php
+/**
+ * Integration: Enterprise Search.
+ *
+ * @package Automattic\VIP\Integrations
+ */
+
+namespace Automattic\VIP\Integrations;
+
+/**
+ * Loads Enterprise Search VIP Integration.
+ *
+ * @private
+ */
+class EnterpriseSearchIntegration extends Integration {
+	/**
+	 * Returns `true` if Enterprise Search is already available e.g. customer code. We will use
+	 * this function to prevent loading of integration again from platform side.
+	 */
+	public function is_loaded(): bool {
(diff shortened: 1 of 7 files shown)


integrations/enterprise-search.php
```diff
@@ -0,0 +1,40 @@
...
+	 * this function to prevent loading of integration again from platform side.
+	 */
+	public function is_loaded(): bool {
+		return class_exists( '\Automattic\VIP\Search\Search' );
+	}
+
+	/**
+	 * Loads the plugin.
+	 *
+	 * @private
```
rinatkhaziev: annotation says `@private` but it's actually public

integrations/enterprise-search.php
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Loads Enterprise Search VIP Integration.
+ */
+class EnterpriseSearchIntegration extends Integration {
+	/**
+	 * Returns `true` if Enterprise Search is already available e.g. customer code. We will use
+	 * this function to prevent loading of integration again from platform side.
+	 */
+	public function is_loaded(): bool {
+		return class_exists( '\Automattic\VIP\Search\Search' );
```
GaryJones: Would something like:
```suggestion
		return class_exists( \Automattic\VIP\Search\Search::class );
```
...work? IDEs can usually help autocomplete and better syntax highlight this than a regular string that can have typos.

integrations/integration-vip-config.php
```diff
@@ -135,12 +135,14 @@ public function get_site_status() {
...
 	 *
 	 * @return array
-	 *
-	 * @private
 	 */
 	public function get_site_config() {
 		if ( is_multisite() ) {
 			$config = $this->get_value_from_config( 'network_sites', 'config' );
+			// If network site config is not found then fallback to env config if it exists
+			if ( empty( $config ) && true === $this->get_value_from_config( 'env', 'cascadeConfig' ) ) {
```
mehmoodak: Should we change `cascadeConfig` to `snake_case`?

integrations/enterprise-search.php
```diff
@@ -0,0 +1,64 @@
...
+		if ( isset( $config['username'] ) && isset( $config['password'] ) ) {
+			define( 'VIP_ELASTICSEARCH_USERNAME', $config['username'] );
+			define( 'VIP_ELASTICSEARCH_PASSWORD', $config['password'] );
+		}
+	}
+
+	private function is_es_credentials_set(): bool {
+		$username_defined = defined( 'VIP_ELASTICSEARCH_USERNAME' ) && constant( 'VIP_ELASTICSEARCH_USERNAME' );
+		$password_defined = defined( 'VIP_ELASTICSEARCH_PASSWORD' ) && constant( 'VIP_ELASTICSEARCH_PASSWORD' );
+		return $username_defined && $password_defined;
```
mehmoodak, line 77: `OR` seems better here else `vip_set_es_credentials` can throw an error if any of this is already defined which is very unlikely.

```suggestion
		return $username_defined || $password_defined;
```

tests/integrations/test-enterprise-search.php
```diff
@@ -0,0 +1,77 @@
...
+		$existing_value = class_exists( '\Automattic\VIP\Search\Search' );
+
+		$es_integration_mock->load();
+
+		if ( ! $existing_value ) {
+			$this->assertTrue( class_exists( '\Automattic\VIP\Search\Search' ) );
+		}
+	}
+
+	public function test__configure_action(): void {
```
mehmoodak, line 66: We should also add the test case for `should_not_configure_if_es_constants_are_already_present`

mehmoodak (approved):

integrations/enterprise-search.php
rebeccahum: In the case that one is defined but the other isn't, I'd still rather not `is_es_credentials_set()` return true.

integrations/integration-vip-config.php
rebeccahum: Yes, I'll need to update the GOOP end (which I'll shortly do).

integrations/integration-vip-config.php
```diff
@@ -135,12 +135,14 @@ public function get_site_status() {
 	 * Get site config.
 	 *
 	 * @return array
-	 *
-	 * @private
 	 */
 	public function get_site_config() {
 		if ( is_multisite() ) {
 			$config = $this->get_value_from_config( 'network_sites', 'config' );
+			// If network site config is not found then fallback to env config if it exists
```
noahtallen, line 142: maybe not applicable to this line. But in the GOOP PR, I see that when both the network site and main site config exist, we merge them:

https://github.com/Automattic/vip-go-api/pull/5172/files#diff-8e75ccc1dc1ca94f37c3361246e36acd19c64eee88d78eafe15ed8b040734ce1R392

Do we need to do anything to keep that behavior in mu-plugins, or does that get encoded in the configmap for that network site??

rebeccahum: @noahtallen That gets encoded in the ConfigMap for the network site!

PR #5538, opened 2024-05-03
2FA: Add pattern to validate phone number format as per E.164

saroshaga:
## Description

The SMS provider that the two-factor authentication plugin uses rejects any phone number that isn't in [E.164 standards](https://www.twilio.com/docs/glossary/what-e164).

Here we're applying some client-side validation using HTML to ensure that incorrect phone numbers aren't configured.

### Changed

- WPCOM VIP 2FA: Add pattern to validate phone number format as per E.164

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

1. Check out PR.
1. Go to `wp-admin` > `Tools` > `Bakery`
1. Click on "Bake Cookies" button.
1. Verify cookies are delicious.


brunobasto (approved):
Looks good to me! Thank you for handling this!

rebeccahum (approved):

rebeccahum (changes requested):
Hmm, I just tested this and it's still accepting phone numbers such as 780-123-4567. I think it might be missing the `<form>` tags around it for the validation?

wpcom-vip-two-factor/sms-provider.php
```diff
@@ -201,12 +201,15 @@ public function user_options( $user ) {
 			<?php else : ?>
 				<label>Phone Number
 					<input name="vip-two-factor-phone" type="tel"
+						pattern="^\+[1-9]\d{1,14}$"
```
sjinks, line 204: > ITU-T specifies that a phone number cannot be longer than fifteen digits, with one to three digits reserved for the country calling code, but valid numbers in Germany have been assigned that are longer than this.
>
> -- https://github.com/google/libphonenumber/blob/master/FALSEHOODS.md

sjinks (changes requested):

mjangda:
@andrea-sdl how did we solve this for VIP Auth? Can we implement similar validation here?

andrea-sdl:
@mjangda for VIP Auth we used https://github.com/jackocnr/intl-tel-input on the frontend, and then validated the phone number via the Twilio API https://github.com/Automattic/vip-auth/blob/d414a80f657954c84026d7fe9dd4799d423ee6be/client-mu-plugins/vip-auth/src/mfa/providers/class-sms.php#L75

PR #5597, opened 2024-05-20
Disable Divi caching

kat3samsin:
## Description

Added disabling of Divi Caching by default. VIP already supports page and object caching Divi Caching can cause bloated media directory.
[Docs](https://docs.wpvip.com/plugins/incompatibilities/#divi)

### Disables Divi plugin caching
Added disabling of Divi Caching by default. 

## Pre-review checklist

## Pre-deploy checklist


WPprodigy (commented):
1) We should prevent a slew of php warnings for customer's currently defining these. So like:

```
if ( ! defined( 'ET_DISABLE_FILE_BASED_CACHE' ) ) {
	define( 'ET_DISABLE_FILE_BASED_CACHE', true );
}
```

2) Should we also be including the rest of the logic mentioned here (with additional safeguards to not register and constant that is already set): https://docs.wpvip.com/plugins/incompatibilities/#divi ?

3) Lastly, let's try to limit the scope of the code to only running when we're sure it is needed - i.e. Divi is present on the site. So wrapping all of the logic inside of something like `if defined( 'ET_CORE_VERSION' ) { ... }`. Load order will be important though, might have to hook it all onto `after_setup_theme` for example.

kat3samsin:
Thanks for the detailed feedback @WPprodigy! I've applied changes in e7d167bf2b66adb0c6e8cc88d900dff69c8d98a0

rinatkhaziev:
@kat3samsin thanks! 

I'm not entirely sure how Divi caching works, is it appropriate for it to use VIP filesystem at all?

kat3samsin:
@rinatkhaziev thanks for looking.
> I'm not entirely sure how Divi caching works, is it appropriate for it to use VIP filesystem at all?

From what I understand, the dynamic loading (dynamic css) feature of Divi is trying to write in `wp_content` which is not allowed on the VIP filesystem. This PR disables the dynamic loading feature if the plugin is installed so that customers would not need to manually apply the [instructions](https://docs.wpvip.com/plugins/incompatibilities/#divi) from the docs.

rinatkhaziev (approved):

PR #5629, opened 2024-06-05
Force secure_logged_in_cookie to always be true

rebeccahum:
## Description
Forces `secure_logged_in_cookie` to be always `true`

### Changed
- Force `secure_logged_in_cookie` to always return true

## Pre-review checklist

## Pre-deploy checklist

### DIFF
--- a/security.php
+++ b/security.php
@@ -397,3 +397,13 @@ function vip_send_wplogin_header( $user ) {
 	return $user;
 }
 add_filter( 'determine_current_user', 'vip_send_wplogin_header', 10000 );
+
+/**
+ * Force secure_logged_in_cookie to be always secure no matter what.
+ * 
+ * @return true Always return true to force secure_logged_in_cookie to be secure.
+ */
+function vip_force_secure_logged_in_cookie() {
+	return true;
+}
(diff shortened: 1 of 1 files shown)


nickdaugherty (approved):
💥 looks great to me. This shouldn't have any negative impact since all HTTP requests redirect to HTTPS on the platform, so there should be no existing reliance on the old behavior.

noahtallen (approved):
Looks great!

rebeccahum:
@nickdaugherty This was breaking things locally (and on our tests) since those envs aren't SSL. I've adjusted the filter to only return true for non-local VIP environments.
I tested the image on an application and it looks OK, but I'm wondering if you think this change warrants a rollout?

nickdaugherty:
@rebeccahum good catch. This can be part of the regular release cycle.

BrookeDot:
If a customer has the "HTTPS Admin/HTTP Frontend" option set, will this cause an issue for them? https://docs.wpvip.com/tls/#h-https-redirection

If so, should we add a check before forcing the cookie to be secure? Alternatively, I'm assuming customers can filter `secure_logged_in_cookie` themselves if in the rare cases where this needs to be `false`. 

cc:// @rebeccahum @nickdaugherty

PR #5643, opened 2024-06-13
VIP Integrations: Improves reading of config files and add multiple configs support of same integration

mehmoodak:
## Description

- Refactored "Integrations Framework" to make reading of config files generic i.e. read config by integration type available inside data instead of slug available in file name.
- Updated `Integrations Framework" to add the ability of configuring multiple instances of same integration.

### Improved VIP Integrations
- Improves reading of config files to make the implementation more generic.
- Adds the ability of configuring multiple instances of same integration.

## Pre-review checklist

- [x] This change works and has been tested locally or in Codespaces.

## Pre-deploy checklist

- [x] Update ConfigMaps generating from backend
- [x] Reload ConfigMaps of all envs to make sure that config files have `type` attribute. We do have this reload ability available in backend API.

## Steps to Test

1. Checkout this PR

1. Create integration config files inside `php` container and place configs (we have to do it manually because as of now `vip dev-env` doesn't have support to auto create this folder/file while creating new envs)
    - `mkdir /wp/config/integrations-config`
    - `touch /wp/config/integrations-config/parsely-config.php` and adds following content
        - `parsely` can be replaced with any other slug of the integration

    ```
    return array(
      'type' => 'parsely',
      'env' => array(
      	'status' => 'enabled',
      	'config' => array (
      		'site_id' => 'site.com',
      		'api_secret' => 'site_api_secret',
      		'metadata_secret' => 'site_metadata_secret',
      	)
      ),
      'network_sites' => array (
      	'1' => array (
      		'status' => 'enabled',
      		'config' => array (
      			'site_id' => 'site1.com',
      			'api_secret' => 'site_api_secret_1',
      			'metadata_secret' => 'site_metadata_secret_1',
      		)
      	),
      	'2' => array (
      		'status' => 'enabled',
      		'config' => array (
      			'site_id' => 'site2.com',
      			'api_secret' => 'site_api_secret_2',
      			'metadata_secret' => 'site_metadata_secret_2',
      		)
      	),
      )
    );
    ```

1. Create config files for multiple integrations.
    - `touch /wp/config/integrations-config/[SLUG]-123-config.php` and adds following content
    -  Replace slug with anything

    ```
    <?php
    
    return array(
      'type' => 'slug',
      'label' => 'Slug 123',
      'env' => array(
      	'status' => 'enabled',
            'config' => array (
      		'key' => 'slug-123',
      	)
      )
    );
    ```

    - `touch /wp/config/integrations-config/[SLUG]-456-config.php` and adds following content

    ```
    <?php
    
    return array(
      'type' => 'slug',
      'label' => 'Slug 456',
      'env' => array(
      	'status' => 'enabled',
            'config' => array (
      		'key' => 'slug-456',
      	)
      )
    );
    ```
1. Restart vip dev-env to apply changes in these config files.

1. Verify integration enablement with different combinations of vip configs e.g. `wp-parsely` plugin is activated with the SiteID and secrets which are provided by VIP (credentails support will be available on `v3.9` and above).

1. Apart from enabling the plugin via VIP config, customer can also enables it by placing `\Automattic\VIP\Integrations\activate( 'parsely' );` in `plugin-loader.php` file so better to also test this flow also.


integrations/integration.php
```diff
@@ -135,18 +156,116 @@ public function get_slug(): string {
...
+	public function set_vip_configs( array $vip_configs ): void {
+		$this->vip_configs = $vip_configs;
+	}
+
+	/**
+	 * Get statuses of the integration in context of current site.
+	 *
+	 * @return array<Env_Integration_Status>
+	 */
+	private function get_site_statuses() {
```
mehmoodak, line 176: Method moved from `IntegrationVipConfig` file and added multi status support.

integrations/integration.php
```diff
@@ -135,18 +156,116 @@ public function get_slug(): string {
...
+	 * Get configs of the integration in context of current site.
+	 *
+	 * @return array<array<mixed>> Returns an array of configs if the integration have multiple configs else single config object.
+	 *
+	 * @private
 	 */
-	public function set_vip_config( IntegrationVipConfig $vip_config ): void {
-		if ( ! $this->is_active() ) {
-			trigger_error( sprintf( 'Configuration info can only assigned if integration is active.' ), E_USER_WARNING ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_trigger_error
+	public function get_site_configs() {
```
mehmoodak, line 204: Method moved from `IntegrationVipConfig` file and added multi configs support.

integrations/integrations.php
```diff
@@ -54,42 +54,34 @@ private function get_integration( string $slug ): ?Integration {
 	 * Activates integrations based on the configuration provided by VIP
 	 * (only if not already activated by customer).
 	 *
+	 * @param VipIntegrationsConfig $vip_integrations_config Object which contains all configurations.
+	 *
 	 * @return void
 	 */
-	public function activate_platform_integrations() {
+	public function activate_platform_integrations( VipIntegrationsConfig $vip_integrations_config ) {
```
mehmoodak, line 61: For easier mocking providing config as a param.

tests/integrations/fake-multi-config-integration.php
mehmoodak, line 1: `FakeMultiConfigIntegration` class for multi config integration.

tests/integrations/test-integration.php
```diff
@@ -8,37 +8,36 @@
...
-		$this->original_error_reporting = error_reporting();
-		set_error_handler( static function ( int $errno, string $errstr ) {
-			if ( error_reporting() & $errno ) {
-				// phpcs:ignore WordPress.Security.EscapeOutput.ExceptionNotEscaped -- CLI
-				throw new ErrorException( $errstr, $errno ); // NOSONAR
-			}
-
-			return false;
-		}, E_USER_WARNING );
+		$this->original_error_reporting = setup_custom_error_reporting();
```
mehmoodak, line 36: Avoided duplication across files and added following methods:

- setup_custom_error_reporting
- reset_custom_error_reporting

tests/integrations/test-integration.php
```diff
@@ -147,19 +141,435 @@ public function test__is_active_returns_false_when_integration_is_not_active():
 		$this->assertFalse( $integration->is_active() );
 	}
 
-	public function test__set_vip_config_function_throws_error_if_integration_is_not_active(): void {
+	public function test__is_active_via_vip_returns_false_if_empty_config_is_provided(): void {
```
mehmoodak, line 144: - Moved methods from `test-integration-config.php` class to here.
- Updated methods to allow multiple configs.

integrations/integration.php
```diff
@@ -48,15 +49,22 @@ abstract class Integration {
...
 	 */
-	private IntegrationVipConfig $vip_config;
+	private array $vip_configs = [];
+
+	/**
+	 * A boolean indicating if the integration have multiple configs.
+	 *
+	 * @var bool
+	 */
+	protected bool $have_multiple_configs = false;
```
rebeccahum: For backwards-compat, should we default this to `true`? It looks like a lot of the integrations seem to have more than one config.

mehmoodak: Seems like we should change the variable name to `$have_multiple_instances` because here I am not referring about the config keys of the integration but multiple configurations of the same integration.

Does the new name make sense? feel free to let me know if you have a better suggestion :)

rebeccahum: Ahhhh, then yes. Let's change the name because that definitely confused me! Maybe even `$has_multiple_setups`?

mehmoodak: Fixed [251976d](https://github.com/Automattic/vip-go-mu-plugins/pull/5643/commits/251976d22dd5e4774453f0197b053eea5c0e8fa6)

WPprodigy:
Pausing on this for now. Will revisit when we're adding in this functionality.

mehmoodak:
Had some discussions with Cantina team and we decided to reopen this Pull Request while keeping the existing proposal as is.

integrations/integration-vip-config.php
mehmoodak, line 1: - Changed the name to vip-integrations-config.php and refactored the code.

integrations/vip-integrations-config.php
mehmoodak, line 1: - Changed the name from `integration-vip-config.php`
- Refactored `get_vip_config_from_file` to read all files at once and rename the function to `read_config_files`
- Moved `is_active_via_vip`, `get_site_status`, `get_site_config` and `get_value_from_config` to `Integration` class because it makes more sense there. Apart from this added support for multiple configs.

tests/integrations/fake-multi-config-integration.php
mehmoodak, line 1: - `FakeMultiConfigIntegration` class for multi config integration.

integrations/integration.php
```diff
@@ -114,6 +124,17 @@ public function is_active(): bool {
...
 
+	/**
+	 * Returns `true` if the integration is enabled in VIP configs else `false`.
+	 *
+	 * @return bool
+	 *
+	 * @private
+	 */
+	public function is_active_via_vip(): bool {
+		return in_array( Env_Integration_Status::ENABLED, $this->get_site_statuses() );
```
mehmoodak, line 135: For multiple setups enable the integration if any of the source is `ENABLED`.

tests/utils/utils.php
```diff
@@ -41,18 +42,51 @@ function get_class_method_as_public( $class_name, $method ) {
...
-	$property->setAccessible( true ); // NOSONAR
+	$property->setAccessible( true ); // NOSONAR.
 
 	return $property;
 }
+
+/**
+ * Setup custom error handler and returns current level of error reporting.
+ */
+function setup_custom_error_reporting(): int {
```
mehmoodak, line 69: Moved duplicated code to following methods

- setup_custom_error_reporting
- reset_custom_error_reporting

mehmoodak:
Made some minor changes and now this PR is good for review.

integrations/integration.php
```diff
@@ -7,7 +7,8 @@
 
 namespace Automattic\VIP\Integrations;
 
-use Automattic\VIP\Integrations\IntegrationVipConfig;
+use Org_Integration_Status;
```
rebeccahum, line 10: I believe we're actually moving away from differing the statuses and want it all in one.

integrations/integration.php
```diff
@@ -135,18 +156,114 @@ public function get_slug(): string {
...
 	 *
 	 * @return void
+	 *
+	 * @private
 	 */
-	public function set_vip_config( IntegrationVipConfig $vip_config ): void {
-		if ( ! $this->is_active() ) {
-			trigger_error( sprintf( 'Configuration info can only assigned if integration is active.' ), E_USER_WARNING ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_trigger_error
+	public function set_vip_configs( array $vip_configs ): void {
+		$this->vip_configs = $vip_configs;
```
rebeccahum, line 168: How come we're removing the active check?

integrations/integration.php
```diff
@@ -135,18 +156,114 @@ public function get_slug(): string {
...
+				$config = $this->get_value_from_config( $vip_config, 'env', 'config' );
+			}
+
+			if ( isset( $config ) ) {
+				$configs[] = $config;
+			}
+		}
+
+		// Return config object if integration have only one config else return all configs.
+		return ( ! $this->have_multiple_setups && isset( $configs[0] ) ) ? $configs[0] : $configs;
```
rebeccahum, line 231: Besides Composable, how many other integrations will have multiple set-ups? Just wondering, but is this something that can be handled elsewhere? :/

integrations/integration.php
WPprodigy: yes please, let's normalize and just have one "statuses" enum.

integrations/integration.php
```diff
@@ -135,18 +156,114 @@ public function get_slug(): string {
...
+		 * Statuses of the integration.
+		 *
+		 * @var array<string>
+		 */
+		$statuses = [];
+
+		foreach ( $this->vip_configs as $vip_config ) {
+			if ( $this->get_value_from_config( $vip_config, 'org', 'status' ) === Org_Integration_Status::BLOCKED ) {
+				return [ Org_Integration_Status::BLOCKED ];
+			}
```
WPprodigy, line 187: Can we instead make these decisions in goop before setting the configmap?

Anything that doesn't belong on a site config (based on inheritance from the client level) shouldn't end up in the configmap in the first place.

integrations/vip-integrations-config.php
```diff
@@ -0,0 +1,172 @@
...
+	 *       ),
+	 *       'network_sites' => array (
+	 *         1 => array (
+	 *           'status' => 'disabled',
+	 *           'config'  => array(),
+	 *         ),
+	 *         2 => array (
+	 *           'status' => 'enabled',
+	 *           'config'  => array(),
+	 *         ),
```
WPprodigy: I know it's just an example, but I think this is surfacing in the code as a result so wanted to point out here too.

Nothing blocked at the org level should ever make it to the WP site. There's no scenario where we will block on org but enable on env. If blocked at org level or site level, then that should mean the configmap will never see an entry for that integration. It doesn't exist at all.

integrations/integration.php
```diff
@@ -135,18 +156,114 @@ public function get_slug(): string {
...
+		$value = null;
+
+		if ( ! in_array( $config_type, [ 'org', 'env', 'network_sites' ], true ) ) {
+			trigger_error( 'config_type param (' . esc_html( $config_type ) . ') must be one of org, env or network_sites.', E_USER_WARNING ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_trigger_error
+		} elseif ( isset( $vip_config[ $config_type ] ) ) {
+
+			// Look for key inside org or env config.
+			if ( 'network_sites' !== $config_type && isset( $vip_config[ $config_type ][ $key ] ) ) {
+				return $vip_config[ $config_type ][ $key ];
+			}
```
WPprodigy, line 253: I'm not even sure if we need to concept of org configs down at this level either.

When generating configmaps, we're already dealing with only one specific environment. So we can decide what the configs are for that particular environment (based on higher levels if needed), and pass down the resulting env configs along with network site configs if it's a MS.

integrations/integration.php
WPprodigy: There are some nice benefits to having individual sources be able to cascade down with their own configs like the other integrations. Helps prevent duplication for network sites for example, and keeps things logically operating the same way.

integrations/integration.php
WPprodigy: I still think we absolutely need the concept of a parent integration present here. The parent should be the one deciding if the whole thing should be enabled (and at what version) for the env/subsite.

mehmoodak: In #5644 we are checking enablement on parent before getting any status from child ([src](https://github.com/Automattic/vip-go-mu-plugins/blob/f001a56d85eab2e5bbd1c60f20502597062c43a1/integrations/remote-data-blocks.php#L63)). This PR is mainly just a refactoring PR that supports composable cases.

integrations/integration.php
mehmoodak: As of now we don't support all statuses on `ORG` and `ENV` so I think the distinction makes sense and explaining the code.

integrations/integration.php
mehmoodak: We are only using [set_vip_configs](https://github.com/Automattic/vip-go-mu-plugins/blob/45da1ae78305caedcf93c84ede21be81418c79e6/integrations/integrations.php#L70) in one place which is already handling the [is_active](https://github.com/Automattic/vip-go-mu-plugins/blob/45da1ae78305caedcf93c84ede21be81418c79e6/integrations/integrations.php#L64) case, therefore removed it.

integrations/integration.php
mehmoodak: 1. Currently we can get the info about why a particular integration doesn't have any data in ConfigMap by just looking at the file. If we move it to GOOP then we loses this visibility which effect debugging if user doesn't have enough permissions.

2. In future we can have integrations that want to show some notification in WP Dashboard if the plugin is `blocked` on client level. So it's useful to have this info available.

If we decide on making any change then I will make these changes in the next PR because this PR is just doing refactoring.

integrations/vip-integrations-config.php
mehmoodak: Thanks for pointing this out. Fixed the example in https://github.com/Automattic/vip-go-mu-plugins/pull/5643/commits/fe14e065a2a00eabedf89d70a61b4582cb40a87d

As per current implementation if the integration is blocked on `org` level then we don't include any data from site level integration even if there is data.

```php
// Example

<?php

return array(
  'parsely' => array(
    array(
      'org'  => array( 'status' => 'blocked' ),
    ),
  ),
);
```

integrations/integration.php
mehmoodak: Currently from code point of view we do support having config on `org` level but in actual we don't have any config from `ConfigMap` so feel free to let me know if you like me to add a condition here which throws the error if config is provided on `org` level. Personally I think it is fine as it is.

PR #5644, opened 2024-06-13
Adds Remote Data Blocks integration

mehmoodak:
## Description

Adds `Remote Data Blocks` integration. This integration is dependent on data sources which is handled via `data_source_integrations` property.

### Added
- Added "Remote Data Blocks" integration

## Pre-review checklist

- [x] This change works and has been tested locally or in Codespaces.

## Pre-deploy checklist

- [ ] Confirm the plugin version before merging.

## Steps to Test

1. Checkout this PR

1. Create integration config files inside `php` container and place configs (we have to do it manually because as of now `vip dev-env` doesn't have support to auto create this folder/file while creating new envs)
    - `mkdir /wp/config/integrations-config`
    - `touch /wp/config/integrations-config/[SLUG]-123-config.php` and adds following content (Replace slug with anything)

    ```
    <?php
    
    return array(
      'type' => 'slug',
      'label' => 'Slug 123',
      'env' => array(
      	'status' => 'enabled',
            'config' => array (
      		'key' => 'slug-123',
      	)
      )
    );
    ```

    - `touch /wp/config/integrations-config/[SLUG]-456-config.php` and adds following content

    ```
    <?php
    
    return array(
      'type' => 'slug',
      'label' => 'Slug 456',
      'env' => array(
      	'status' => 'enabled',
            'config' => array (
      		'key' => 'slug-456',
      	)
      )
    );
    ```
1. Restart vip dev-env to apply changes in these config files.

1. Use `vip_integration_composable_blocks_config` filter to make sure that expected data is returning.

1. Apart from enabling the plugin via VIP config, customer can also enables it by placing `\Automattic\VIP\Integrations\activate( 'remote-data-blocks' );` in `plugin-loader.php` so please test this also and verify the configs.


integrations/vip-composable-blocks.php
```diff
@@ -0,0 +1,82 @@
...
+ * @private
+ */
+class VipComposableBlocksIntegration extends Integration {
+	/**
+	 * The version of the "VIP Composable Blocks" plugin to load, that's set to the latest version.
+	 * This should be higher than the `lowestVersion` set in "vip-composable-blocks" config (https://github.com/Automattic/vip-go-mu-plugins-ext/blob/trunk/config.json)
+	 *
+	 * @var string
+	 */
+	protected string $version = '1.0';
```
mehmoodak: Is `v1.0` okay here? cc: @chriszarate @maxschmeling

chriszarate: 1.0 is fine with me, but we might want to sync this with a version slug in the plugin repo

integrations/integration.php
```diff
@@ -74,6 +87,11 @@ abstract class Integration {
 	public function __construct( string $slug ) {
 		$this->slug = $slug;
 
+		// Registers child integrations if any.
+		foreach ( $this->child_integrations as $integration ) {
+			IntegrationsSingleton::instance()->register( $integration );
+		}
```
chriszarate: Do we expect this child logic to be reused by other integrations? If not, maybe it should live in overridden methods of `VipComposableBlocksIntegration` to keep the base class simple.

integrations/vip-composable-blocks/airtable.php
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+namespace Automattic\VIP\Integrations;
+
+/**
+ * Loads Airtable Integration.
+ *
+ * @private
+ */
+class AirtableIntegration extends Integration {
```
chriszarate: Do we need to enumerate the different types of child integrations here? Or can a single generic datasource integration suffice? Ideally we don't need to release mu-plugins in order to add a new datasource type.

integrations/integration.php
mehmoodak: Addressed in [5ae46f4](https://github.com/Automattic/vip-go-mu-plugins/pull/5644/commits/5ae46f4b0d1fd68da0c5593843313804905247c7)

integrations/vip-composable-blocks/airtable.php
mehmoodak: Addressed in [5ae46f4](https://github.com/Automattic/vip-go-mu-plugins/pull/5644/commits/5ae46f4b0d1fd68da0c5593843313804905247c7)

integrations/vip-composable-blocks.php
mehmoodak: Changed to `0.1.0` [06e0307](https://github.com/Automattic/vip-go-mu-plugins/pull/5644/commits/06e030791bb657ef68f4f8eda88d535bccc7b1de)

integrations/vip-composable-blocks/airtable.php
mehmoodak: > Ideally we don't need to release mu-plugins in order to add a new datasource type.

As per current implementation we do need to release MU Plugins in order to add a new data source because the data_sources are hard coded in `VipComposableBlocksIntegration`.

If this is a must have requirement then we can update ConfigMap template to include `parent_type` attribute and then get data_sources based on this attribute. Please do let me know about your thoughts on this.

chriszarate: Fine with me to hardcode possible data sources as long as we have a pain-free path to not hardcode it in the future

mehmoodak: Okay so for now I am going with hard coded approach. In future we can update the implementation if needed.

integrations/integration.php
```diff
@@ -222,7 +222,13 @@ public function get_site_configs() {
 				$config = $this->get_value_from_config( $vip_config, 'env', 'config' );
 			}
 
+			if ( isset( $vip_config['label'] ) ) {
+				$config['label'] = $vip_config['label']; // Useful to differentiate between multiple configs of same integration.
+			}
+
 			if ( isset( $config ) ) {
+				$config['type'] = $this->slug; // Useful specially when integration have multiple setups.
```
rebeccahum, line 230: Should we check it's set before assigning it?

rebeccahum (changes requested):

integrations/integration.php
mehmoodak: We set it in constructor so it should always be set.

rebeccahum (approved):

PR #5727, opened 2024-07-19
Fixes `log2logstash` debugging on local dev-envs

brunobasto:
## Description

I noticed calls to `Logger::log2logstash` didn't output to either `stdout` or `debug.log` on local `vip dev-env`s . Some checks were not accounting for the case where `VIP_GO_ENV` is set to `local` - which is the case on the local dev-env.

### Fixed
- Fixed debug logs with `Logger::log2logstash` on local dev envs.

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

1. Check out PR.
1. Create a dev-env with `vip dev-env create` that points to your local mu-plugins folder.
1. Add a call to 

```php
Logger::log2logstash( [
	'severity' => 'error',
	'feature'  => 'test',
	'message'  => 'Test message'
] );
```

somewhere on your php files

4. Check that logs are written to `/tmp/debug.log` on your dev-env container.

### DIFF
--- a/logstash/class-logger.php
+++ b/logstash/class-logger.php
@@ -565,11 +565,6 @@ class Logger {
 	 * @param array $entry Data.
 	 */
 	public static function wp_debug_log( array $entry ): void {
-		if ( defined( 'VIP_GO_ENV' ) && constant( 'VIP_GO_ENV' ) !== 'local' ) {
-			// Don't run this on VIP Go
-			return;
-		}
-
 		$log_path = WP_CONTENT_DIR . '/debug.log';
 		$log_path = is_string( WP_DEBUG_LOG ) && WP_DEBUG_LOG ? WP_DEBUG_LOG : $log_path;
 
--- a/logstash/class-logger.php
+++ b/logstash/class-logger.php
@@ -524,7 +524,7 @@ class Logger {
 
 		// Process all entries.
 		foreach ( static::$entries as $entry ) {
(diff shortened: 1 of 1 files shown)


logstash/class-logger.php
```diff
@@ -565,7 +565,7 @@ public static function maybe_wp_debug_log_entries( array $entry ): void {
 	 * @param array $entry Data.
 	 */
 	public static function wp_debug_log( array $entry ): void {
-		if ( defined( 'VIP_GO_ENV' ) && VIP_GO_ENV ) {
+		if ( defined( 'VIP_GO_ENV' ) && constant( 'VIP_GO_ENV' ) !== 'local' ) {
```
rebeccahum: I think we can remove this check since we already do the checking above.

brunobasto: Good point. Removed here: https://github.com/Automattic/vip-go-mu-plugins/pull/5727/commits/aa4ca76060ff113777195cf46df5f9f7c3db19cf

Note that now we'll log if people call `wp_debug_log` manually, and before it would skip it if it wasn't in the local env.

rebeccahum: I'm not seeing usages of it outside of vip-go-mu-plugins, so we should be all set here.

rebeccahum (approved):

brunobasto:
Thank you @rebeccahum

PR #5737, opened 2024-07-23
Adds support for the TWILIO_MESSAGING_SERVICE_SID when sending SMS

andrea-sdl:
## Description

This PR adds support for the `MessagingServiceSid` parameter in Twilio Calls. This is an alternative approach to sending SMS where all the logic around the sender is located in Twilio, to benefit from their knowledge about if/when to use an alphanumeric ID instead of a number.

## Pre-review checklist

- [X] This change works and has been tested on a sandbox.

## Pre-deploy checklist

## Steps to Test

- In a local test site, using this change, define the variable `define( 'TWILIO_MESSAGING_SERVICE_SID', '<redacted>' );` with the correct `MessagingServiceId` from Twilio.
- Log in, and go into your user profile and setup SMS as a 2FA.
- The SMS should arrive.
- Confirm, from the [Twilio Messaging logs](https://console.twilio.com/us1/monitor/logs/sms), that the message has been sent and that the "Service" column is now set.

### DIFF
--- a/lib/sms.php
+++ b/lib/sms.php
@@ -43,10 +43,18 @@ function send_sms( $to, $message ) {
 			}
 
 			$body = array(
-				'From' => SMS_FROM_NUMBER,
 				'To'   => $to_number,
 				'Body' => $message_split,
 			);
+			/**
+			 * If defined, we want to use the MessagingServiceSid to leverage all the automatic logic of the messaging service to route the SMS from the right "From" number.
+			 * For example in some countries, the "From" number should be a local number to the recipient or an alphanumeric sender ID.
+			 */
+			if ( defined( 'TWILIO_MESSAGING_SERVICE_SID' ) ) {
+				$body['MessagingServiceSid'] = TWILIO_MESSAGING_SERVICE_SID;
+			} else {
+				$body['From'] = SMS_FROM_NUMBER;
+			}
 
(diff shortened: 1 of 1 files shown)


rebeccahum (approved):

lib/sms.php
```diff
@@ -43,10 +43,18 @@ function send_sms( $to, $message ) {
...
 			$body = array(
-				'From' => SMS_FROM_NUMBER,
 				'To'   => $to_number,
 				'Body' => $message_split,
 			);
+			/**
+			 * If defined, we want to use the MessagingServiceSid to leverage all the automatic logic of the messaging service to route the SMS from the right "From" number.
+			 * For example in some countries, the "From" number should be a local number to the recipient or an alphanumeric sender ID.
+			 */
+			if ( defined( 'TWILIO_MESSAGING_SERVICE_SID' ) ) {
```
mjangda, line 53: @rebeccahum @andrea-sdl can we actually prefix this with `VIP_`?

I know there are other constants in this file that aren't prefixed, but it would be a good idea to do so for any new and safe to change constants.

andrea-sdl: @mjangda added in https://github.com/Automattic/vip-go-mu-plugins/pull/5737

PR #5769, opened 2024-08-05
test(e2e): add generic checks

sjinks:
This PR adds generic checks to our E2E test suite. So far, there are only three checks, but suggestions are welcome:
* check that the home page contains the closing `</html>` tag, no `wp_die()` message, and its response code is less than 500. This check ensures that our code does not accidentally break the frontend;
* check that the REST API works, responds with 200 OK, and provides a valid JSON that contains some expected fields;
* check that XML RPC works, responds with 200 OK, and responds to the `demo.sayHello` RPC call.


rinatkhaziev:
Also would be great to check the REST API?

rebeccahum (approved):

PR #5770, opened 2024-08-05
Add base telemetry library

hanifn:
## Description

This PR adds a new generic Telemetry library consisting of base classes for Automattic's Tracks system integration. This is based heavily on the existing `Telemetry` package under the `vip-parsely` directory but meant to be more generic and can be used by other plugins.

### Added
- Added new Telemetry lib for A8C Tracks integration

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

This is a base library so no concrete implementation to test yet.


telemetry/Tracks/class-tracks-client.php
```diff
@@ -0,0 +1,187 @@
...
+	 * @var array<Tracks_Event>
+	 */
+	protected $events = array();
+
+	/**
+	 * Constructor.
+	 */
+	public function __construct() {
+		// Track events asynchronously (inject pixels in the footer).
+		add_action( 'admin_footer', array( $this, 'render_tracking_pixels' ) );
```
rinatkhaziev: This only tracks WP Admin requests, is that intentional? 
for frontend we'd need to use `wp_footer`

telemetry/Tracks/class-tracks-client.php
```diff
@@ -0,0 +1,187 @@
...
+	 * Events queued to be sent to the Tracks pixel.
+	 *
+	 * @var array<Tracks_Event>
+	 */
+	protected $events = array();
+
+	/**
+	 * Constructor.
+	 */
+	public function __construct() {
```
rinatkhaziev: since it's a singleton this should be private.

rinatkhaziev:
1. Besides the inline feedback, let's drop `go` from `vipgo`. 
2. Needs tests

telemetry/Tracks/class-tracks-client.php
hanifn: Yes that is intentional. We're limiting this to WP admin for now. I'm also still trying to figure out a good way to make this part flexible so that where it renders the pixel is determined by the plugin implementation 🤔

telemetry/Tracks/class-tracks-client.php
```diff
@@ -0,0 +1,187 @@
...
+			return $request;
+		}
+
+		return true;
+	}
+
+	/**
+	 * Outputs a Tracks pixel for every registered event.
+	 */
+	public function render_tracking_pixels(): void {
```
alecgeatches: @hanifn How are the actual pixel-style tracking pixels used? I'm more used to the server-side only pixel requests, so just want to understand. Is it:

1. Detect I'm on a page where I want to track something e.g. `$pagenow === 'edit.php'`
2. Call `register_events( [ new Tracks_Event( 'edit_loaded', ... ) ] )` during page loads
3. Pixels automatically fire off on browser render

Is that correct?

alecgeatches (approved):
Looks good to me! Agree with Rinat that tests would be great, especially as a demo of how to build an event and fire it off.

telemetry/Tracks/class-tracks-client.php
```diff
@@ -0,0 +1,187 @@
+<?php
```
mjangda: Unless we're using uppercase already elsewhere, we should rename the folder to `tracks` (lowercase `t`).

telemetry/Tracks/class-tracks-client.php
```diff
@@ -0,0 +1,187 @@
...
+		$is_event_recordable = $event->is_recordable();
+
+		if ( true !== $is_event_recordable ) {
+			return $is_event_recordable;
+		}
+
+		$pixel_url = static::instance()->generate_pixel_url( $event );
+
+		if ( null === $pixel_url ) {
+			return new WP_Error(
```
mjangda: How will know about these errors? We may want to use log2logstash to track them.

telemetry/Tracks/class-tracks-client.php
```diff
@@ -0,0 +1,187 @@
...
+		);
+	}
+
+	/**
+	 * Records any remaining events synchronously.
+	 */
+	public function record_remaining_events(): void {
+		foreach ( $this->events as $event ) {
+			static::instance()->record_event_synchronously( $event );
+		}
```
mjangda: The REST API supports bulk events which would be better since we could just make one remote request instead of multiples: https://github.com/Automattic/vip-cli/blob/2a8c90afa9ed2e5d784e820c5da20183094b6711/src/lib/analytics/clients/tracks.ts#L84-L105

telemetry/Tracks/class-tracks-event.php
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	/**
+	 * Processes the event's properties to get them ready for validation.
+	 *
+	 * @param string $event_name The event's name.
+	 * @param array<string, mixed>|array<empty> $event_properties Any event properties to be processed.
+	 * @return stdClass The resulting event object with processed properties.
+	 */
+	protected static function process_properties(
```
mjangda: Minor: the use of `static` methods for this is a bit odd.

telemetry/Tracks/class-tracks-event.php
```diff
@@ -0,0 +1,275 @@
...
+	): stdClass {
+		$event = (object) static::sanitize_properties_array( $event_properties );
+		$event = static::set_user_properties( $event );
+
+		// Set event name.
+		$event->_en = preg_replace(
+			'/^(?:' . static::EVENT_NAME_PREFIX . ')?(.*)/',
+			static::EVENT_NAME_PREFIX . '\1',
+			$event_name
+		) ?? '';
```
mjangda: It's not totally clear what the `preg_replace` is doing here.

We should also set a different prefix for local environments (e.g. `vip_dev_`.

telemetry/Tracks/class-tracks-event.php
```diff
@@ -0,0 +1,275 @@
...
+
+		// Users in the VIP environment.
+		if ( defined( 'VIP_GO_APP_ID' ) ) {
+			$app_id = constant( 'VIP_GO_APP_ID' );
+			if ( is_integer( $app_id ) && 0 < $app_id ) {
+				$event->_ui = $app_id . '_' . $wp_user_id;
+				$event->_ut = 'vip_go_app_wp_user';
+
+				return $event;
+			}
```
mjangda: We need to use an identifier that is pseudo-anonymized but still allows tracking the user across different environments (e.g. if I'm working on wpvip.com and docs.wpvip.com, I should not be counted as two different users).

Maybe a hash of the email?

```
$salt = constant( 'VIP_TELEMETRY_SALT' ); // this would be private but globally shared across the platform
$tracks_user_id = hash_hmac( 'sha256', $user->user_email, $salt );
```

telemetry/Tracks/class-tracks-event.php
```diff
@@ -0,0 +1,275 @@
...
+		}
+
+		// Set VIP environment if it exists.
+		if ( defined( 'VIP_GO_APP_ENVIRONMENT' ) ) {
+			$app_environment = constant( 'VIP_GO_APP_ENVIRONMENT' );
+			if ( is_string( $app_environment ) && '' !== $app_environment ) {
+				$event->vipgo_env = $app_environment;
+			}
+		}
+
```
mjangda: We should also add:

- org id
- if the user is a VIP user

telemetry/Tracks/class-tracks-event.php
```diff
@@ -0,0 +1,275 @@
...
+/**
+ * Class that creates and validates Tracks events.
+ *
+ * @since 3.12.0
+ */
+class Tracks_Event {
+	/**
+	 * Event name prefix.
+	 */
+	protected const EVENT_NAME_PREFIX = 'viptelemetry_';
```
mjangda: We want `vip_` as the prefix.

We may also want to think about how each of our plugins could register and append their own namespace (e.g. the workflows plugin may want all events to be `vip_workflows_`)

telemetry/Tracks/class-tracks-event.php
```diff
@@ -0,0 +1,275 @@
...
+	 */
+	public function is_recordable() {
+		if ( ! Telemetry_System::is_wpadmin_telemetry_allowed() ) {
+			return false;
+		}
+
+		// Don't record events during unit tests and CI runs.
+		if ( 'wptests_capabilities' === wp_get_current_user()->cap_key ) {
+			return false;
+		}
```
mjangda: This feels odd...

telemetry/Tracks/class-tracks.php
```diff
@@ -0,0 +1,79 @@
...
+	 *                                                            Key names must be lowercase and snake_case.
+	 * @return bool|WP_Error True if recording the event succeeded.
+	 *                       False if telemetry is disabled.
+	 *                       WP_Error if recording the event failed.
+	 */
+	public function record_event(
+		string $event_name,
+		array $event_properties = array()
+	) {
+		$event = new Tracks_Event( $event_name, $event_properties );
```
mjangda: If the Tracks Event had some errors, we should log those somewhere so we have visibility into failures.

telemetry/class-telemetry-system.php
```diff
@@ -0,0 +1,73 @@
...
+	abstract protected function activate_tracking(): void;
+
+	/**
+	 * Returns whether wp-admin telemetry is allowed to be enabled. This is off
+	 * by default.
+	 *
+	 * @return bool Whether wp-admin telemetry is allowed to be enabled.
+	 */
+	public static function is_wpadmin_telemetry_allowed(): bool {
+		return apply_filters( 'wp_parsely_enable_wpadmin_telemetry', false );
```
mjangda: Let's remove this filter.

telemetry/Tracks/class-tracks.php
```diff
@@ -0,0 +1,79 @@
...
+		}
+
+		return $pixel->record_event_asynchronously( $event );
+	}
+
+	/**
+	 * Registers the events into their respective WordPress hooks, so they
+	 * can be recorded when the hook fires.
+	 */
+	protected function activate_tracking(): void {
```
mjangda: I think we should get rid of this abstraction and make all our our tracking explicit via `record_event()` calls.

telemetry/Tracks/class-tracks-event.php
```diff
@@ -0,0 +1,275 @@
...
+			$wp_base_url = get_option( 'siteurl' );
+		}
+
+		/**
+		 * The base URL of the site.
+		 *
+		 * @var string $wp_base_url
+		 */
+		$event->_ui = wp_hash( sprintf( '%s|%s', $wp_base_url, $wp_user_id ) );
+		$event->_ut = 'wpparsely:user_id';
```
mjangda: Remove the wpparsely reference.

mjangda (commented):
Thanks for working on this @hanifn. I've added some comments / questions.

telemetry/Tracks/class-tracks-client.php
hanifn: These errors would eventually be returned to the callbacks that are hooked into the filters we're using to track events. The idea is to let the plugins implementing this handle the errors however they want to.

telemetry/Tracks/class-tracks-client.php
hanifn: Oh good idea! I'll look into that 👍

telemetry/Tracks/class-tracks-event.php
hanifn: ~~This was copied from the Parsely plugin so I'm not entirely sure the reason for it.~~

Okay I figured it out. This is because the method is called in the constructor when the object has not been instantiated yet.

telemetry/Tracks/class-tracks-event.php
hanifn: Yeah I've been thinking about that myself. An idea I have is to set the prefix as a constructor argument for the `Tracks` class and the `Tracks_Event` class.

telemetry/Tracks/class-tracks-event.php
hanifn: This is to stop events from being recorded during tests, is this not the correct way to do it? Or is it just unnecessary?

telemetry/Tracks/class-tracks.php
hanifn: We would still need a way register all the callbacks to the appropriate filters and `record_event()` is specifically used by the callbacks to send the events to the Tracks system so I'm not sure to achieve what you're asking 🤔

telemetry/Tracks/class-tracks-event.php
hanifn: Added these in 2ebe3ad40e25314db8768e48ca88fd16a10109c0 but I'm not so sure about the VIP user check. I vaguely recall a helper function for this but the only thing I can find is `is_automattician()` which doesn't seem quite correct 🤔

telemetry/Tracks/class-tracks-client.php
mjangda: I think we should handle this in the library rather than asking each plugin to handle the error.

telemetry/Tracks/class-tracks-event.php
mjangda: The `user_has_vip_support_role()` method is probably the best one, although I'm surprised we don't have a helper function for that.

https://github.com/Automattic/vip-go-mu-plugins/blob/44e6293555331be77646e30ee2df97c961fcd34c/vip-support/class-vip-support-user.php#L833

Maybe we can add a new `is_proxied_vip_support_user()` helper function that adds on to `is_proxied_automattician()`?

telemetry/Tracks/class-tracks-event.php
mjangda: More just that it's happening in an unusual spot in the code. Feels like we should do this as late as possible (i.e. before the request is sent).

telemetry/Tracks/class-tracks.php
mjangda: What I mean is that rather than abstracting like so: https://github.com/Automattic/vip-go-mu-plugins/blob/44e6293555331be77646e30ee2df97c961fcd34c/vip-parsely/vip-parsely.php#L37-L55

With this approach the registration of the hook is separated too far away the actual tracking code which makes it harder to debug.

We should let plugins do the extra work needed to record what they need:

```
add_action( 'transition_post_status', function( $new, $old, $post ) {
    Automattic\VIP\Telemetry::get_instance()->record_event( 'post_status_changed', [ ... ] );
}, 10, 3 );
```

This requires more code but keeps the code more explicit and easier to read / follow / debug.

mjangda: @Automattic/vip-platform-cantina what do y'all think?

telemetry/Tracks/class-tracks-client.php
rinatkhaziev: Why would it be consequential to the plugins?

telemetry/tracks/class-tracks-client.php
```diff
@@ -0,0 +1,187 @@
...
+	 */
+	public function render_tracking_pixels(): void {
+		foreach ( $this->events as $event ) {
+			$pixel_url = $this->generate_pixel_url( $event );
+
+			if ( null === $pixel_url ) {
+				continue;
+			}
+
+			echo '<img style="position: fixed;" src="', esc_url( $pixel_url ), '" />';
```
rinatkhaziev: 1. Is there a benefit of doing this vs simply running on shutdown? Suppose we go overboard and track 20 different things during the same page load.
2. Such as it is this is not async, so the distinction between this method and `record_event_synchronously` is kind of misnomer.

telemetry/Tracks/class-tracks.php
rinatkhaziev: I agree with @mjangda , and I don't even think this will be more code? 

And if a particular plugin wants to enrich the data somehow it can do something like where it would make sure that the data has the necessary event properies

```php
Automattic\VIP\Telemetry::get_instance()->record_event( 'post_status_changed', enrich_track_event( [ ... ] ) );
```

telemetry/Tracks/class-tracks-client.php
rinatkhaziev: > We may want to use log2logstash to track them.
I second that, we can set feature `telemetry` and then pass the some of the properties from `$event` as `$extra`. 
Here's an example:
https://github.com/Automattic/vip-go-mu-plugins/blob/4d236a4d90a6e6e58dffb8899b9a494813585f46/config/class-sync.php#L222-L228

telemetry/Tracks/class-tracks-client.php
ingeniumed: It'd be worth expanding this to the frontend as well imo. If we wanted, we could add a marker in the event to say when it's frontend or WP Admin and target the event recording based on that but I don't think it's worth it?

telemetry/Tracks/class-tracks-client.php
ingeniumed: Agreed with @rinatkhaziev and @mjangda, we should handle this in the library so errors are raised uniformly across all implementations if they originate from within the library

telemetry/Tracks/class-tracks-event.php
ingeniumed: It'd be great to allow plugins to set some sort of a prefix so looking up those logs based on that would be easier

telemetry/Tracks/class-tracks.php
ingeniumed: Agreed on the above.

This also leads to a good point - we should add a README for this library with an example of how to use it. That'll help to simplify it down once you see how a plugin would implement it. IMO, it should be as easy to follow as the `send_pixel` method.

telemetry/tracks/class-tracks-client.php
ingeniumed: Agreed on this, I'm not fully grasping why this would be better over running on shutdown 🤔. We want the site's experience to not be impacted, even if that may come at the expense of telemetry imo.

telemetry/Tracks/class-tracks-client.php
hanifn: Added `log2logstash` in 1b308c7646759ea9ad36c94fac528bfd669cf294

telemetry/Tracks/class-tracks-client.php
hanifn: Changed in 7ae681a528aaf3678c307e68e0c56c7f134b2748

telemetry/Tracks/class-tracks.php
hanifn: I think my changes in 1b308c7646759ea9ad36c94fac528bfd669cf294 should cover this

telemetry/Tracks/class-tracks.php
hanifn: Removed in 637891038f65b5a66d5bff9840f68d36e6d1ffd3

mjangda:
Thanks for the fixes, @hanifn. Based on the latest changes, is this how a typical implementation would work?

```php
$tracker = new Tracks( 'vip_bakery_' );
$tracker->record_event( 'bakery_opened' );
```

If we want to abstract a bit for easier re-usability, that could look something like this:

```php
namespace Automattic\VIP\Bakery\Analytics;

use Automattic\VIP\Telemetry\Tracks;

function record_event( $event_name, $event_props = [] ) {
    static $tracker = new Tracks( 'vip_bakery_' ); // Our prefix for the events will be 'vip_bakery_'

    static $global_props = [
        'delicious' => 'always',
    ];


    $merged_props = array_merge( $global_props, event_props ); // We should probably allow passing global props to the `Tracks` constructor and handle this there.

    $tracker->record_event( $event_name, $merged_props );
}
```

Then, elsewhere in our code:

```php
use function Automattic\VIP\Bakery\Analytics\record_event;

record_event( 'bakery_open' );

function bake_cookies( $number ) {
    // ...

    record_event( 'cookies_baked', [ 'number' => $number ] );
}

add_action( 'vip_bakery_eat_cookie', function( $cookie_type ) {
    record_event( 'cookies_eaten', [ 'cookie_type' => $cookie_type ] );
}, 10 );
```

Every plugin will need the same boilerplate but it's not so bad.

telemetry/Tracks/class-tracks-event.php
hanifn: Added this in 89b0248fc7bd0dc74b6995ef8a64f4731551d916

telemetry/Tracks/class-tracks-event.php
hanifn: Changed in e6a6436f60695cf401336128bc956eda060bea53

telemetry/Tracks/class-tracks-event.php
hanifn: Changed this in d9697a67ce5395ae291a52447316419ac84d2043
The prefix can now be passed to the `Tracks` class

telemetry/Tracks/class-tracks-client.php
hanifn: @alecgeatches Yes that is about how it works although the `register_events` method no longer exist now. Any events that are triggered after the pixel rendered will be handled during the `shutdown` action by the `record_remaining_events` method

telemetry/Tracks/class-tracks-event.php
hanifn: The `is_recordable` method is being called by methods like `record_event_synchronously` which are the methods that makes the request. This seems pretty late to me though 🤔

telemetry/tracks/class-tracks-client.php
hanifn: 1. A benefit for having this on top of the shutdown part is that any events that are recorded before the pixel are rendered will be sent first so if we have 20 different events to track, there is a change those events could be sent in 2 batches. One for events before the pixel is rendered and one for events after.
2. Agreed it is kind of a misnomer but the naming convention is consistent between the Parsely plugin, WPCOM and Woo

hanifn:
> Based on the latest changes, is this how a typical implementation would work?

@mjangda Yes that looks about right. The client plugin would need to have appropriate functions or methods to hook into the events they want to track and those functions or methods has to call the `Tracks->record_event()` method. How they achieve that is entirely up to them since this new approach is pretty flexible

telemetry/Tracks/class-tracks-event.php
ingeniumed: [This](https://github.com/Automattic/vip-block-data-api/blob/trunk/src/analytics/analytics.php#L99) is one way that we limited such tracking from happening (for MC Stats) to only production sites. This would be worth considering so it's done pretty early on.

telemetry/tracks/class-tracks-client.php
```diff
@@ -0,0 +1,258 @@
...
+
+		if ( true !== $is_event_recordable ) {
+			return $is_event_recordable;
+		}
+
+		$pixel_url = static::instance()->generate_pixel_url( $event );
+
+		if ( null === $pixel_url ) {
+			log2logstash( [
+				'severity' => 'error',
```
ingeniumed, line 104: Should this be an error, or warning? Considering the status code is 400 wondering if that's better? Just don't want the customer's logs filled up with errors because the plugin didn't implement an event correctly.

telemetry/tracks/class-tracks-client.php
ingeniumed: I'm leaning towards starting with just doing this on shutdown, rather than the pixel approach and shutdown approach. I think we should try to minimize our impact on the page as much as possible. Considering this is telemetry, having it go on shutdown at the very should be fine imo.

ingeniumed (commented):
Got a minor comment about the errors, but a bigger one about the dual pixel and shutdown approach that I think we should tackle.

telemetry/tracks/class-tracks-client.php
hanifn: If the library can't generate the proper pixel URL then the site won't be able to send events for AJAX requests. I think that's a pretty serious problem that should be logged IMHO

telemetry/README.md
```diff
@@ -0,0 +1,59 @@
...
+			'post'       => (array) $post,
+		] );
+	}
+}
+```
+
+With the class above, you can then initiate event tracking in the main plugin file with these lines:
+
+```php
+$tracker = new Tracker();
```
sanmai, line 57:
```suggestion
$tracker = new MyPluginTracker();
```

PR #5776, opened 2024-08-05
Fix incorrect width and height in core/image block

itsTallulah:
## Description

Addresses incorrect `img` tag `width`|`height` values in `core/image` blocks by filtering `render_block`, where it can get the attachment ID and intended size from the classes of the `img` and `figure` tags.
From there it looks for the intended size's w|h values in the attachment's postmeta, and add them to the `img` tag for the return.
Also fixes incorrect srcsets for block images without having to touch them, since it's upstream of them and supplies the values they need.
Returns original safely and silently on failure. Matches narrowly, and can operate on already-filtered blocks as long as they have the required size and ID classes.
Optional switch for `resize` args can be omitted, but it could be nice to give folks the choice. I defer to plat as to how to make this switchable (or if we should just lose it for simplicity and brevity).

### Fixed
- Fixed incorrect image sizes and srcsets for `core/image` blocks on frontend requests

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

1. Check out PR / use relevant file as generic mu-plugin
2. Examine any page/post with core/image block/s
3. Examine src of any core/image block (other than size:full)
4. Its `width` and `height` values and srcset should all be congruent with its intended size
5. It should also visually look the same size in the frontend and the editor
6. Repeat as required for any/all registered sizes
7. Optional: register some custom sizes and repeat


fix-block-img-tag-sizes.php
```diff
@@ -0,0 +1,67 @@
...
+
+                // Replace the original URL with the new URL in the img tag
+                $block_content = str_replace($original_url, $new_url, $block_content);
+            }
+        }
+
+        // Add the width and height attributes to the img tag
+        $block_content = preg_replace(
+            '/<img ([^>]+)>/',
+            '<img foo="bar" $1 width="' . esc_attr($width) . '" height="' . esc_attr($height) . '">',
```
sjinks: Is this `foo="bar"` thing really necessary? :-)

sjinks: Also, can the `` tag contain other attributes (say, `alt`)? If so, is it possible to avoid overwriting them?

fix-block-img-tag-sizes.php
```diff
@@ -0,0 +1,67 @@
...
+        if (!$metadata || !isset($metadata['sizes'][$size_name])) {
+            return $block_content; // No metadata found for this size, return original content
+        }
+
+        // Get the width, height, and file from the metadata
+        $width = $metadata['sizes'][$size_name]['width'];
+        $height = $metadata['sizes'][$size_name]['height'];
+        $new_file = $metadata['sizes'][$size_name]['file'];
+
+        if ($resizearg) {
```
sjinks: This condition is always `false` because we don't seem to set `$resizearg` to `true` anywhere.

fix-block-img-tag-sizes.php
```diff
@@ -0,0 +1,67 @@
...
+        // Add the width and height attributes to the img tag
+        $block_content = preg_replace(
+            '/<img ([^>]+)>/',
+            '<img foo="bar" $1 width="' . esc_attr($width) . '" height="' . esc_attr($height) . '">',
+            $block_content
+        );
+    }
+    return $block_content;
+}
+add_filter('render_block', 'fix_img_block_sizes', 10, 3);
```
t-wright: Since we're only handling `core/image` blocks, perhaps we can hook into `render_block_core/image` here? https://github.com/WordPress/wordpress-develop/blob/50af37a9083f003f8e98d089091d2cc428797cc5/src/wp-includes/class-wp-block.php#L581

fix-block-img-tag-sizes.php
itsTallulah: Oops, that foo is a testing leftover! Thanks for the spot. And yeah, the img and figure tags can contain other values and classes too.

fix-block-img-tag-sizes.php
itsTallulah: Indeed, we don't set it true anywhere - yet. I'm deferring to y'all as to how we should let the user set this (if we include it at all).

fix-block-img-tag-sizes.php
itsTallulah: Good point - Should behave exactly the same on that filter, since it's right after the first one and uses the same values. Will give that a try too.

fix-block-img-tag-sizes.php
itsTallulah: See the first commit: it has the now-removed `add_block_bits`, which I was using for as an informal test for extra values in the tags by filtering a bunch of extra stuff into the block markup just upstream.

itsTallulah:
Looking at the `resizearg` option I had in here, it would arguably make more sense to have that as a separate filter combo that can hit all contexts where image scaler args would otherwise be `w=` style, so I'm going to remove it from this. I'll pitch that as a separate PR, since it's far lower prio than this fix.

fix-block-img-tag-sizes.php
itsTallulah: Works well and makes the function smaller. Nice, thx!

itsTallulah:
Removed `resizearg` section, added a skip for fullsize image size, changed filter to `render_block_core/image`, renamed function to `vip_fix_img_block_sizes` for specificity and to avoid collisions (since pervious versions of this fix have been shipped to some customers as a stopgap).

itsTallulah:
Bumping out of stale to discuss

itsTallulah:
Refactored to use HTML API as discussed in Slack. Verbose with comments on purpose, since I haven't used the HTML API much and I wanted to make sure the flow was followable. Thoughts?

a8c-files.php
```diff
@@ -104,6 +104,9 @@ public function __construct() {
 				$this->schedule_update_job();
 			}
 		}
+
+		// Add filter to fix image block sizes
+		add_filter( 'render_block_core/image', array( $this, 'fix_img_block_sizes' ), 10, 3 );
```
sjinks:
```suggestion
		if ( ! is_admin() ) {
			add_filter( 'render_block_core/image', array( $this, 'fix_img_block_sizes' ), 10, 3 );
		}
```

and remove `is_admin()` check in `fix_img_block_sizes()`.

a8c-files.php
```diff
@@ -511,6 +514,88 @@ public function update_attachment_meta() {
...
+	 * @param array  $block The full block, including name and attributes.
+	 * @param object $instance The block instance.
+	 * @return string Modified block content.
+	 */
+	public function fix_img_block_sizes( $block_content, $block, $instance ) {
+		// Don't fire in wp-admin, image blocks are fine there
+		if ( is_admin() ) {
+			return $block_content;
+		}
+
```
sjinks:
```suggestion
```

a8c-files.php
```diff
@@ -511,6 +514,88 @@ public function update_attachment_meta() {
...
+		// Get values ready
+		$width                = $metadata['sizes'][ $size_name ]['width'];
+		$height               = $metadata['sizes'][ $size_name ]['height'];
+		$update_tag_processor = new WP_HTML_Tag_Processor( $block_content );
+
+		// Set width/height for img
+		if ( $update_tag_processor->next_tag( 'img' ) ) {
+			$update_tag_processor->set_attribute( 'width', esc_attr( $width ) );
+			$update_tag_processor->set_attribute( 'height', esc_attr( $height ) );
```
sjinks:
```suggestion
			$update_tag_processor->set_attribute( 'width', $width );
			$update_tag_processor->set_attribute( 'height', $height );
```

See https://developer.wordpress.org/reference/classes/wp_html_tag_processor/set_attribute/:
> For string attributes, the value is escaped using the `esc_attr` function.

See https://github.com/WordPress/wordpress-develop/blob/50fb4086b7afbfa012c5d1f2eeff79b1bae3b00e/src/wp-includes/html-api/class-wp-html-tag-processor.php#L3875

a8c-files.php
```diff
@@ -511,6 +514,88 @@ public function update_attachment_meta() {
...
+		}
+
+		// Get the result ready
+		$block_content = $update_tag_processor->get_updated_html();
```
sjinks: Do you think we can reuse `$img_tag_processor` here?

tests/files/test-fix-block-img-tag-sizes.php
```diff
@@ -0,0 +1,110 @@
...
+	 * @param string $expected      The expected output.
+	 * @param array  $metadata      Optional metadata for the image.
+	 * @return void
+	 */
+	public function test__fix_img_block_sizes( $is_admin, $block_content, $expected, $metadata = null ) {
+		
+		if ( $is_admin && ! defined( 'WP_ADMIN' ) ) {
+			define( 'WP_ADMIN', true );
+		}
+
```
sjinks: I don't like it because it might affect other tests.

I suggest moving the `is_admin()` check around the `add_filter()` call.

tests/files/test-fix-block-img-tag-sizes.php
```diff
@@ -0,0 +1,110 @@
...
+		// For the successful-update case, add more detailed assertions
+		if ( $metadata ) {
+			$this->assertStringContainsString( 'width="300"', $actual, 'Width attribute should be present' );
+			$this->assertStringContainsString( 'height="200"', $actual, 'Height attribute should be present' );
+			$this->assertStringContainsString( 'class="wp-image-123"', $actual, 'Image class should be preserved' );
+			$this->assertStringContainsString( 'class="wp-block-image size-medium"', $actual, 'Figure class should be preserved' );
+			remove_filter( 'get_post_metadata', $filter, 10 );
+		} else {
+			$this->assertEquals( $expected, $actual );
+		}
```
sjinks: Maybe we can make `$expected` a `string` or an `array`? If `$expected` is a string, we call `assertEquals()`. Otherwise, we invoke `assertStringContainsString()` for every array item. This will make the code cleaner.

tests/files/test-fix-block-img-tag-sizes.php
```diff
@@ -0,0 +1,110 @@
...
+		if ( $metadata ) {
+			$attachment_id = 123; // Matches the wp-image-123 in test data
+			$filter        = function ( $meta_value, $post_id, $meta_key ) use ( $metadata, $attachment_id ) {
+				if ( '_wp_attachment_metadata' === $meta_key && $attachment_id == $post_id ) {
+					return [ $metadata ];
+				}
+				return $meta_value;
+			};
+			add_filter( 'get_post_metadata', $filter, 10, 3 );
+		}
```
sjinks: Should this go to `setUp()` and `tearDown()`?

tests/files/test-fix-block-img-tag-sizes.php
rinatkhaziev: @sjinks there's no guarantee order of arguments will stay in the same order (hence the THIS IS FLAKY comment). With this approach it's less flaky

sjinks: @rinatkhaziev we will have `[ 'width="300"', 'height="200"', ... ]` etc and then

```php
if ( is_array(‌ $expected ) ) {
    foreach ( $expected as $substring ) {
        self::assertStringContainsString( $substring, $actual );
    }
}
```

This will move all hardcoded checks out of the test into the data.

ThirdEyeRose:
Hi folks - we have a customer who is interested in using this solution when we release it. I see we've moved into a reviewing stage but I know we're still in active development. Is there a rough estimate for when this is expected to be in the pipeline to production?

The customer has been trying to adapt this concept to their own custom plugin, but if our version will be available in a few weeks, they'd prefer to wait for our version, so I wanted to check in.

tests/files/test-fix-block-img-tag-sizes.php
sjinks: https://github.com/Automattic/vip-go-mu-plugins/blob/4ddac9fb0ad018436c51be39d1617f5806671ac6/tests/files/test-fix-block-img-tag-sizes.php#L91-L97

rinatkhaziev (approved):

PR #5827, opened 2024-08-28
Allow Overwriting SMTP Host in VIP Mail

luismulinari:
## Description

This PR introduces a new feature in the VIP Mail plugin that allows for the SMTP host to be overwritten. This is particularly useful in cases where the default SMTP host needs to be changed for specific mailing requirements.

The changes include:

1. A new constant `VIP_SMTP_HOST_OVERWRITE_ALLOW_LIST` has been defined. This constant holds a comma-separated list of hosts that are allowed to overwrite the default SMTP host.

2. In the `phpmailer_init` method of the `VIP_SMTP` class, we check if the `VIP_SMTP_HOST_OVERWRITE_ALLOW_LIST` constant is defined. If it is, we explode it into an array of allowed hosts.

3. We then check if the `Mailer` property of the `phpmailer` object is set to 'smtp' and if the `Host` property of the `phpmailer` object is not in the array of allowed hosts. If both conditions are met, we return early from the method, effectively preventing the host from being overwritten.

This feature provides more flexibility in configuring the SMTP settings for the VIP Mail plugin. It allows developers to specify a list of hosts that can overwrite the default SMTP host, thereby catering to a wider range of mailing requirements.

## Pre-review checklist

## Pre-deploy checklist

### DIFF
--- a/vip-mail.php
+++ b/vip-mail.php
@@ -103,7 +103,7 @@ final class VIP_SMTP {
 		}
 
 		$host_overwrite_allow_list = defined( 'VIP_SMTP_HOST_OVERWRITE_ALLOW_LIST' )
-			? explode( ',', constant( 'VIP_SMTP_HOST_OVERWRITE_ALLOW_LIST' ) )
+			? array_map( 'trim', explode( ',', constant( 'VIP_SMTP_HOST_OVERWRITE_ALLOW_LIST' ) ) )
 			: [];
 
 		if ( 'smtp' === $phpmailer->Mailer && ! in_array( $phpmailer->Host, $host_overwrite_allow_list, true ) ) {
--- a/vip-mail.php
+++ b/vip-mail.php
@@ -106,7 +106,7 @@ final class VIP_SMTP {
 			? explode( ',', constant( 'VIP_SMTP_HOST_OVERWRITE_ALLOW_LIST' ) )
 			: [];
 
-		if ( 'smtp' === $phpmailer->Mailer && ! in_array( $phpmailer->Host, $host_overwrite_allow_list ) ) {
+		if ( 'smtp' === $phpmailer->Mailer && ! in_array( $phpmailer->Host, $host_overwrite_allow_list, true ) ) {
 			return;
 		}
 
--- a/vip-mail.php
+++ b/vip-mail.php
@@ -102,7 +102,11 @@ final class VIP_SMTP {
 			return;
 		}
 
-		if ( 'smtp' === $phpmailer->Mailer ) {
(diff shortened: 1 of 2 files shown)


abdullah-kasim (approved):
LGTM

vip-mail.php
```diff
@@ -102,7 +102,11 @@ public function phpmailer_init( &$phpmailer ): void {
 			return;
 		}
 
-		if ( 'smtp' === $phpmailer->Mailer ) {
+		$host_overwrite_allow_list = defined( 'VIP_SMTP_HOST_OVERWRITE_ALLOW_LIST' )
+			? explode( ',', constant( 'VIP_SMTP_HOST_OVERWRITE_ALLOW_LIST' ) )
```
sanmai: Non-blocking: I'm curious if you would consider `str_replace` to remove any odd spaces. It should be very fast if there are no spaces, and also guard against human error.

vip-mail.php
```diff
@@ -102,7 +102,11 @@ public function phpmailer_init( &$phpmailer ): void {
...
+			: [];
+
+		if ( 'smtp' === $phpmailer->Mailer && ! in_array( $phpmailer->Host, $host_overwrite_allow_list ) ) {
```
sanmai:
```suggestion
		if ( 'smtp' === $phpmailer->Mailer && ! in_array( $phpmailer->Host, $host_overwrite_allow_list, true ) ) {
```

In this context, there are very few scenarios where the use of strict comparison with `in_array` (`true` as the third parameter) would cause issues. So it could make sense to set it up to avoid any doubts this code does something you don't expect.

sanmai (approved):
Before this change, we will return early if `'smtp' === $phpmailer->Mailer`.

In the same situation if we consider that `VIP_SMTP_HOST_OVERWRITE_ALLOW_LIST` is not defined, then `'smtp' === $phpmailer->Mailer && ! in_array( $phpmailer->Host, [] )` will evaluate to `true && ! false` which is the same `true` hence this change is non-BC breaking.

vip-mail.php
rebeccahum: Might be better to use `array_map()` with `trim()` for the whitespaces. Something like:

```
$host_overwrite_allow_list = defined('VIP_SMTP_HOST_OVERWRITE_ALLOW_LIST')
    ? array_map('trim', explode(',', constant('VIP_SMTP_HOST_OVERWRITE_ALLOW_LIST')))
    : [];
```

luismulinari: Awesome! Thanks for the ideas <3

rebeccahum (approved):

PR #5863, opened 2024-09-16
Simplify Parse.ly loader

rinatkhaziev:
## Description

We're reducing the number of ways the plugin can be loaded. Eventually there will be only one way (via Integrations) and the rest will be gradually phased out.

Additionally, this fixes SDS reporting an `UNKNOWN` version by relying on raw `parsely` option rather than checking for the global instance of Parse.ly which may not exist at the moment of the version check call. 

### Removed
- Removed ability to load Parse.ly via an option.

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

1. Check out the PR
2. Set filter `add_filter( 'wpvip_parsely_load_mu', '__return_true' );` 
3. Activate the plugin from `plugins` folder
4. Make sure nothing fatals and the plugin is loaded through the core plugin system.
5. Deactivate the plugin and set `add_filter( 'wpvip_parsely_load_mu', '__return_false' );` 
6. Make sure the Parse.ly is not loaded

### DIFF
--- a/wp-parsely.php
+++ b/wp-parsely.php
@@ -281,22 +281,19 @@ function maybe_load_plugin() {
 		Parsely_Loader_Info::set_integration_type( Parsely_Integration_Type::ENABLED_CONSTANT );
 	}
 
-	$option_load_status   = get_option( '_wpvip_parsely_mu', null );
 	$filtered_load_status = apply_filters( 'wpvip_parsely_load_mu', null );
 
 	// If plugin isn't enabled via constant then check for filter and option status.
 	if ( true !== $parsely_enabled_constant ) {
-		$should_load            = true === $filtered_load_status || '1' === $option_load_status;
-		$should_prevent_loading = false === $filtered_load_status || '0' === $option_load_status;
+		$should_load            = true === $filtered_load_status;
+		$should_prevent_loading = false === $filtered_load_status;
 
 		// No integration: The site has not enabled parsely.
 		if ( ! $should_load || $should_prevent_loading ) {
 			Parsely_Loader_Info::set_active( false );
 
-			if ( false === $filtered_load_status ) {
+			if ( $should_prevent_loading ) {
 				Parsely_Loader_Info::set_integration_type( Parsely_Integration_Type::DISABLED_MUPLUGINS_FILTER );
-			} elseif ( '0' === $option_load_status ) {
-				Parsely_Loader_Info::set_integration_type( Parsely_Integration_Type::DISABLED_MUPLUGINS_SILENT_OPTION );
 			}
 
 			return;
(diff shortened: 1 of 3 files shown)


wp-parsely.php
```diff
@@ -281,30 +281,23 @@ function maybe_load_plugin() {
...
 
-	$option_load_status   = get_option( '_wpvip_parsely_mu', null );
 	$filtered_load_status = apply_filters( 'wpvip_parsely_load_mu', null );
 
 	// If plugin isn't enabled via constant then check for filter and option status.
 	if ( true !== $parsely_enabled_constant ) {
-		$should_load            = true === $filtered_load_status || '1' === $option_load_status;
-		$should_prevent_loading = false === $filtered_load_status || '0' === $option_load_status;
+		$should_load            = true === $filtered_load_status;
+		$should_prevent_loading = false === $filtered_load_status;
```
rinatkhaziev: This now reads ... weird. But I didn't want to introduce any extra changes to the logic.

wp-parsely.php
```diff
@@ -281,30 +281,23 @@ function maybe_load_plugin() {
...
 
 		// No integration: The site has not enabled parsely.
 		if ( ! $should_load || $should_prevent_loading ) {
```
rebeccahum: Then we can just do 
```suggestion
		if ( null === $filtered_load_status || false === $filtered_load_status ) {
```

wp-parsely.php
rebeccahum: I think we can just check directly for the `$filtered_load_status` because it does read weird af now.

wp-parsely.php
```diff
@@ -281,30 +281,23 @@ function maybe_load_plugin() {
...
 			Parsely_Loader_Info::set_active( false );
 
-			if ( false === $filtered_load_status ) {
+			if ( $should_prevent_loading ) {
```
rebeccahum, line 295: And then we can remove this line to simplify it.

rebeccahum (approved):

rinatkhaziev:
I've tested in all possible permutations, seems like it works :)

wp-parsely.php
```diff
@@ -250,61 +251,46 @@ function maybe_load_plugin() {
...
+			Parsely_Loader_Info::set_active( true );
+			Parsely_Loader_Info::set_integration_type( Parsely_Integration_Type::SELF_MANAGED );
+			$parsely_options = Parsely_Loader_Info::get_parsely_options();
+			if ( array_key_exists( 'plugin_version', $parsely_options ) ) {
+				Parsely_Loader_Info::set_version( $parsely_options['plugin_version'] );
 			}
+			break;
+		// Integrations-managed
+		case defined( 'VIP_PARSELY_ENABLED' ):
+			Parsely_Loader_Info::set_active( constant( 'VIP_PARSELY_ENABLED' ) );
```
rebeccahum:
```suggestion
			Parsely_Loader_Info::set_active( true === constant( 'VIP_PARSELY_ENABLED' ) );
```

rebeccahum (approved):

rebeccahum (approved):

PR #5869, opened 2024-09-17
Add base telemetry library (take 2)

sanmai:
## Description

> Builds on https://github.com/Automattic/vip-go-mu-plugins/pull/5770

This PR adds a new generic Telemetry library consisting of base classes for Automattic's Tracks system integration. This is based heavily on the [existing `Telemetry` package under the `vip-parsely` directory](https://github.com/Automattic/vip-go-mu-plugins/tree/406c53d89bd9877eac0b174d0967bc3af3ed98ec/vip-parsely/Telemetry) but meant to be more generic and can be used by other plugins.

Current API:

```php
use Automattic\VIP\Telemetry\Tracks;

$tracks = new Tracks( 'vip_workflow_', [ 'plugin_version' => '1.2.3' ] );
$tracks->record_event( 'post_custom_status_changed', [
    'new_status' => $new_status,
    'old_status' => $old_status,
    'post_id'    => $post->ID,
] );
```

Internals:

- `Tracks` is the main entry point; it constructs `Tracks_Client` and `Tracks_Event_Queue` initially and builds `Tracks_Event` as necessary.
- `Tracks_Event_Queue` registers itself in the shutdown hook (so no singleton is necessary) and maintains a list of pending events. 
- `Tracks_Client` filters out bogus events and sends valid events in a POST request to the API. 
- And `Tracks_Event` builds `Tracks_Event_DTO` from the input array, validating and extending what is necessary.

So there are no singletons, most classes have fixed responsibilities, and we should be able to test with DI most things.

This is not the final form though: I'm concerned `Tracks_Event` does too much, so I might consider moving some of the responsibilities to `Tracks_Event_DTO` but not before I have it covered by unit tests.

#### Development Loop
```
php vendor/bin/phpcbf telemetry tests/telemetry 000-vip-init.php; ./bin/test.sh --filter Telemetry
```

### Added

- Added a telemetry library consisting of base classes for Automattic's Tracks system integration.

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

I added the following:
```php
function record_telemetry_event_in_footer() {
    $tracks = new \Automattic\VIP\Telemetry\Tracks();
    $tracks->record_event( 'local_event', [
        'hello' => 'vip',
    ] );
}

add_action('admin_footer', 'record_telemetry_event_in_footer');
```
Then I loaded the admin and waited for some time to see the event appear in Tracks.


telemetry/tracks/class-tracks-event-builder.php
```diff
@@ -0,0 +1,332 @@
...
+use JsonSerializable;
+use Automattic\VIP\Support_User\User as Support_User;
+use function Automattic\VIP\Logstash\log2logstash;
+
+/**
+ * Class that creates and validates Tracks events.
+ *
+ * @since 3.12.0
+ */
+class Tracks_Event_Builder implements JsonSerializable {
```
sanmai: The name is suboptimal.

telemetry/tracks/class-tracks-client.php
```diff
@@ -0,0 +1,86 @@
...
+ * Handles all operations related to the Tracks pixel.
+ */
+class Tracks_Client {
+	/**
+	 * Tracks REST API endpoint for post requests
+	 */
+	protected const TRACKS_ENDPOINT = 'https://public-api.wordpress.com/rest/v1.1/tracks/record';
+
+	/**
+	 * @var Tracks_Client
```
sanmai:
```suggestion
	 * @var WP_Http
```

telemetry/tracks/class-tracks-client.php
```diff
@@ -0,0 +1,86 @@
...
+	 * Record a batch of events using the Tracks REST API
+	 * 
+	 * @param Tracks_Event[] $events Array of Tracks_Event objects to record
+	 * @return bool|WP_Error True if batch recording succeeded.
+	 *                       WP_Error is any error occured.
+	 */
+	public function batch_record_events( array $events, array $common_props = [] ) {
+		// Don't record events during unit tests and CI runs.
+		if ( 'wptests_capabilities' === wp_get_current_user()->cap_key ) {
+			// TODO we should not need this at all
```
sanmai: I'm not sure this early return is really necessary.

telemetry/tracks/class-tracks-event.php
```diff
@@ -0,0 +1,327 @@
...
+	 * Returns the event's data.
+	 *
+	 * @return Tracks_Event|WP_Error Event object if the event was created successfully, WP_Error otherwise.
+	 */
+	public function get_data() {
+		return $this->data;
+	}
+
+	#[\ReturnTypeWillChange]
+	public function jsonSerialize() {
```
sanmai: Might remove this later.

telemetry/tracks/class-tracks-event.php
```diff
@@ -0,0 +1,327 @@
...
+		$event->_ui = wp_hash( sprintf( '%s|%s', $wp_base_url, $wp_user->ID ) );
+		$event->_ut = 'anon'; // Same as the default value in the original code.
+
+		return $event;
+	}
+
+	/**
+	 * Validates the event object.
+	 *
+	 * @param Tracks_Event_DTO|self $event Event object to validate.
```
sanmai:
```suggestion
	 * @param Tracks_Event_DTO $event Event object to validate.
```

telemetry/tracks/class-tracks-event.php
```diff
@@ -0,0 +1,327 @@
...
+			$wp_base_url = get_option( 'siteurl' );
+		}
+
+		/**
+		 * The base URL of the site.
+		 *
+		 * @var string $wp_base_url
+		 */
+		$event->_ui = wp_hash( sprintf( '%s|%s', $wp_base_url, $wp_user->ID ) );
+		$event->_ut = 'anon'; // Same as the default value in the original code.
```
sanmai, line 222: https://github.com/Automattic/vip-go-mu-plugins/blob/f18c1d1930fff1dd27b32be3703807930cbc9b5a/vip-parsely/Telemetry/Tracks/class-tracks-event.php#L62

telemetry/tracks/class-tracks-event.php
```diff
@@ -0,0 +1,327 @@
+<?php
+/**
+ * Telemetry: Tracks Event Builder class
```
sanmai:
```suggestion
 * Telemetry: Tracks Event class
```

telemetry/class-tracks.php
```diff
@@ -0,0 +1,85 @@
...
+	 * If the event doesn't pass validation, it gets silently discarded.
+	 *
+	 * @param string                            $event_name The event name. Must be snake_case.
+	 * @param array<string, mixed>|array<empty> $event_properties Any additional properties to include with the event.
+	 *                                                            Key names must be lowercase and snake_case.
+	 * @return bool|WP_Error True if recording the event succeeded.
+	 *                       False if telemetry is disabled.
+	 *                       WP_Error if recording the event failed.
+	 */
+	public function record_event(
```
ingeniumed, line 68: The doc for this method needs to be updated to account for the fact that pixel approach has been dropped, and that only the async approach is being used.

telemetry/class-tracks.php
```diff
@@ -0,0 +1,85 @@
...
+		string $event_name,
+		array $event_properties = array()
+	) {
+		if ( [] !== $this->global_event_properties ) {
```
ingeniumed, line 72: Could you do an empty check instead here?

```suggestion
		if ( ! empty( $this->global_event_properties ) ) {
```

telemetry/tracks/class-tracks-client.php
```diff
@@ -0,0 +1,84 @@
+<?php
+/**
+ * Telemetry: Tracks Pixel class
```
ingeniumed: Would adjust this to show that it's going to be an async call to the tracks endpoint. There's more instances of this below as well.

telemetry/tracks/class-tracks-event-queue.php
```diff
@@ -0,0 +1,72 @@
...
+
+	/**
+	 * Records all queued events synchronously.
+	 */
+	public function record_events(): void {
+		if ( [] === $this->events ) {
+			return;
+		}
+
+		$this->client->batch_record_events( $this->events );
```
ingeniumed: If there's an error during the recording operation, all the events would be dropped. Wondering if it's worth re-trying based on the status code? If it's a 4xx then no, but if it's a 5xx it might be?

ingeniumed (approved):
Don't have anything major to change, just minor documentation changes and one question around re-trying a request

telemetry/class-tracks.php
sanmai: I'd prefer that we use a comparison that is as explicit as possible. 

For example, `empty()` will return true for `"0"`, so if someone uses `empty()`, you will have to double-check that they are not using it with strings, which adds to review burden and maintenance. There are no such issues with `!==` or `===`: you see what it is, and nothing more.

telemetry/class-tracks.php
sanmai: Removed all references to pixels I could find in 77aa90429

telemetry/tracks/class-tracks-event-queue.php
sanmai: What is the cost of a missed event? We need to discuss this because good retries require a back-off strategy, which complicates many things. Even with back-off, the Tracks system might not appreciate retries because they could load the system more when it is already struggling. So, if we want to complicate the system with retries, we have to have a very good reason to do that.

So if the cost of missed singular event is negligible, it might be better to skip retries.

I'm looking at VIP CLI implementation, and there are no retries:

https://github.com/Automattic/vip-cli/blob/2b4f7da6d6e80ecb98a99e7e643ba91d38f8a307/src/lib/analytics/clients/tracks.ts#L119-L136

telemetry/class-tracks.php
ingeniumed: Great point, happy to stick with this approach then 👍🏾

telemetry/tracks/class-tracks-event-queue.php
ingeniumed: I should have added this in my comment, but I'm 50/50 on it for this reason. Considering it's telemetry, I'd argue that the cost of a missed event is not really worth it. As is we are doing this on shutdown to ensure that there's minimal disruption to a site's operations.

Perhaps it's worth adding this as a comment so the decision behind it is known.

telemetry/tracks/class-tracks-client.php
```diff
@@ -0,0 +1,91 @@
...
+		// no events - nothing to do
+		if ( [] === $valid_events ) {
+			return true;
+		}
+
+		// @codeCoverageIgnoreStart
+		if ( _wp_http_get_object() === $this->http && 'wptests_capabilities' === wp_get_current_user()->cap_key ) {
+			throw new \Exception( 'WP_Http should be mocked in unit tests' );
+		}
+		// @codeCoverageIgnoreEnd
```
sanmai: I'm down to drop this whole part, unless there are objections.

telemetry/tracks/class-tracks-event.php
```diff
@@ -0,0 +1,336 @@
...
+	/**
+	 * Constructor.
+	 *
+	 * @param string                            $prefix The event's prefix.
+	 * @param string                            $event_name The event's name.
+	 * @param array<string, mixed>|array<empty> $event_properties Any properties included in the event.
+	 */
+	public function __construct( string $prefix, string $event_name, array $event_properties = [] ) {
+		$event_data        = static::process_properties( $prefix, $event_name, $event_properties );
+		$validation_result = static::get_event_validation_result( $event_data );
```
sanmai: There's very little benefit to static methods here (they don't get to access `$this` and thus somewhat faster), so if there are strong feelings against using them, we can switch to instance methods instead here or in a follow up.

My rationale for leaving them is that there's no harm.

telemetry/tracks/class-tracks-event-queue.php
sanmai: Did that 👍

telemetry/tracks/class-tracks-client.php
ingeniumed: Good with me

ingeniumed (approved):

telemetry/README.md
```diff
@@ -0,0 +1,67 @@
...
+## Tracks
+
+Tracks is an event tracking tool used to understand user behaviour within Automattic. This library provides a way for plugins to interact with the Tracks system and start recording events.
+
+### How to use
+
+Example:
+
+```php
+use Automattic\VIP\Telemetry\Tracks
```
alecgeatches: Missing semicolon here:

```suggestion
use Automattic\VIP\Telemetry\Tracks;
```

telemetry/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+function track_post_status( $new_status, $old_status, $post ) {
+	$tracks = new Tracks( 'myplugin_' );
+
+	$tracks->record_event( 'post_status_changed', [
+		'new_status' => $new_status,
+		'old_status' => $old_status,
+		'post'       => (array) $post,
```
alecgeatches: Love this easy example. A whole `$post` [may be overkill](https://github.com/Automattic/vip-workflow-plugin/pull/21#discussion_r1733516494) for something we'd want to recommend sending to telemetry. How about:

```suggestion
		'post_id'       => $post->ID,
```

or similar instead?

telemetry/tracks/class-tracks-event.php
```diff
@@ -0,0 +1,336 @@
...
+
+		return $this->data;
+	}
+
+	/**
+	 * Returns whether the event can be recorded.
+	 *
+	 * @return bool|WP_Error True if the event is recordable.
+	 */
+	public function is_recordable() {
```
alecgeatches: Is there a reason why we can't encode the `bool|WP_Error` return type in PHP types here? Also in some other places in this class.

telemetry/tracks/class-tracks-event.php
```diff
@@ -0,0 +1,336 @@
...
+		$wp_user = wp_get_current_user();
+
+		// Only track logged-in users.
+		if ( 0 === $wp_user->ID ) {
+			return $event;
+		}
+
+		// Set anonymized event user ID; it should be consistent across environments.
+		if ( defined( 'VIP_TELEMETRY_SALT' ) ) {
+			$salt           = constant( 'VIP_TELEMETRY_SALT' ); // this would be private but globally shared across the platform
```
alecgeatches: The comment here makes it sound like `VIP_TELEMETRY_SALT` is not yet implemented. Is that correct, or can we update the comment to make it present-tense?

telemetry/tracks/class-tracks-event.php
```diff
@@ -0,0 +1,336 @@
...
+			$event->_ui = $tracks_user_id;
+			$event->_ut = 'vip:user_email';
+
+			return $event;
+		}
+
+		// Users in the VIP environment.
+		if ( defined( 'VIP_GO_APP_ID' ) ) {
+			$app_id = constant( 'VIP_GO_APP_ID' );
+			if ( is_integer( $app_id ) && 0 < $app_id ) {
```
alecgeatches: Super minor nit, I find yoda-style checks for `<`/`>` hard to parse, and I need to transpose them in my head to read code. Could we:

```suggestion
			if ( is_integer( $app_id ) && $app_id > 0 ) {
```

telemetry/tracks/class-tracks-event.php
```diff
@@ -0,0 +1,336 @@
...
+				'feature'  => 'telemetry',
+				'message'  => $msg,
+				'extra'    => [
+					'event' => (array) $event,
+				],
+			] );
+			return new WP_Error(
+				'invalid_event',
+				$msg,
+				array( 'status' => 400 )
```
alecgeatches: Do these statuses have a functional use?

alecgeatches (approved):
I think this is a robust implementation with solid testing. Some minor comments, but looks great to me overall.

mjangda:
Cross-referencing some work done by @andrea-sdl for VIP Auth: https://github.com/Automattic/vip-auth/blob/a1d6a42c7be669117d1f6239b8ad46c521fc11e2/client-mu-plugins/vip-auth/src/class-tracking.php

000-vip-init.php
```diff
@@ -233,6 +233,15 @@
 require_once __DIR__ . '/vip-helpers/class-user-cleanup.php';
 require_once __DIR__ . '/vip-helpers/class-wpcomvip-restrictions.php';
 
+// Load the Telemetry files
+require_once __DIR__ . '/telemetry/class-telemetry-system.php';
+require_once __DIR__ . '/telemetry/class-tracks.php';
+require_once __DIR__ . '/telemetry/tracks/class-tracks-event-dto.php';
+require_once __DIR__ . '/telemetry/tracks/class-tracks-event.php';
+require_once __DIR__ . '/telemetry/tracks/class-tracks-event-queue.php';
+require_once __DIR__ . '/telemetry/tracks/class-tracks-client.php';
```
rinatkhaziev: This will likely lead to intermittent fatals on the initial deploy as this file will be updated sooner than the subdirectories.

telemetry/tracks/class-tracks-event.php
```diff
@@ -0,0 +1,336 @@
...
+
+	/**
+	 * Sets the Tracks User ID and User ID Type depending on the current
+	 * environment.
+	 *
+	 * @param Tracks_Event_DTO $event The event to annotate with identity information.
+	 * @return Tracks_Event_DTO The new event object including identity information.
+	 */
+	protected static function set_user_properties( Tracks_Event_DTO $event ): Tracks_Event_DTO {
+		$wp_user = wp_get_current_user();
```
rinatkhaziev, line 186: This is unsafe, depending on the context the `wp_get_current_user` may not be present. E.g. this would fatal:

```php
add_action( 'muplugins_loaded', fn() => var_dump( wp_get_current_user() ) );
```

telemetry/tracks/class-tracks-event.php
```diff
@@ -0,0 +1,336 @@
...
+				$event->_ui = sprintf( '%s_%s', $app_id, $wp_user->ID );
+				$event->_ut = 'vip_go_app_wp_user';
+
+				return $event;
+			}
+		}
+
+		// All other environments.
+		$wp_base_url = get_option( 'home' );
+		if ( ! is_string( $wp_base_url ) || '' === $wp_base_url ) {
```
rinatkhaziev: Under which conditions this would've matched?

telemetry/tracks/class-tracks-event.php
sanmai: It may note be implemented, but we can make it present-tense anyway.

telemetry/tracks/class-tracks-event.php
sanmai: Sure, I find it hard to parse myself.

telemetry/tracks/class-tracks-event.php
sanmai: I don't recall. I left them in because there's no harm, but if you think otherwise, let's cut them out.

telemetry/tracks/class-tracks-event.php
sanmai: I've seen this fatal during testing. How would you like to proceed here?

I'm thinking of moving all validation to the write time at shutdown. Any objections?

telemetry/tracks/class-tracks-event.php
sanmai: No good reason other than me being defensive. Since we have good test coverage, it is a good time to add them all around which I did 👍

000-vip-init.php
sanmai: How would you like this part changed? Would a follow-up PR with these lines work?

telemetry/tracks/class-tracks-event.php
sanmai: Did that in 275bddd616872420ee28e763c6a0f17c70d3c8e0; please let me know what you think is best.

telemetry/tracks/class-tracks-event.php
sanmai: What would you like to see here instead?

It sounds like it could happen in case of a major fault, or deliberate malice. 

```
function set_home_option_empty( $value ) {
    return '';
}

add_filter( 'option_home', 'set_home_option_empty' );
```


For more context, this is from Parsely's implementation:

https://github.com/Automattic/vip-go-mu-plugins/blob/5bf617632fbcb65b95e1957a19139e1091f96470/vip-parsely/Telemetry/Tracks/class-tracks-event.php#L53-L56

telemetry/tracks/class-tracks-event.php
```diff
@@ -0,0 +1,362 @@
...
+			$this->data = $validation_result ?? $event_data;
+		}
+
+		return $this->data;
+	}
+
+	/**
+	 * Returns the event's data for JSON representation.
+	 */
+	#[\ReturnTypeWillChange]
```
sanmai:
```suggestion
```

000-vip-init.php
sanmai: Please let me know if this approach works:  https://github.com/Automattic/vip-go-mu-plugins/pull/5869/commits/c95a993f6199ac614d11c1089ed49203db95969e

rinatkhaziev: I'm not sure if this works. Are there any cross-dependencies? I.e. where telemetry/class-tracks.php trying to include `telemetry/tracks/class-tracks-client.php`? I.e. the situation is - the file A was already copied and is passing file_exists check. The file A wants something from file C, which is not copied yet.

telemetry/tracks/class-tracks-event.php
alecgeatches: I usually see these in the context of a REST response, but in here they're all consumed by other code. I think they should be remove unless we're actually using these for HTTP status responses, as they imply something misleading.

000-vip-init.php
sanmai: These files do not have side effects, only class definitions, so it should not be an issue.
And neither of these files includes or requires other files directly.

Please let us know if you have something particular on your mind to resolve this.

telemetry/tracks/class-tracks-event.php
sanmai: Makes sense, they are gone in f70cc5543

mehmoodak:
1. Is there an estimate for when this PR will be merged?
2. This PR currently adds the telemetry library for PHP so are there any discussions of doing something similar for JavaScript/TypeScript codebases?

sanmai:
@mehmoodak 
1. We plan to merge it once we resolve all concerns identified during the review.
2. There is a [pair of similar clients in VIP CLI](https://github.com/Automattic/vip-cli/tree/2b4f7da6d6e80ecb98a99e7e643ba91d38f8a307/src/lib/analytics/clients).

telemetry/tracks/class-tracks-event.php
sanmai: Since I don't have a strong opinion about this fallback (I'm following in the footsteps of the author of https://github.com/Automattic/vip-go-mu-plugins/pull/5770), I removed it in 8ffe069. Please let me know if you want it back.

000-vip-init.php
sanmai: I changed this part to only require files if all of them are present. So if there's a missing file, none will be required.

telemetry/class-telemetry-event-queue.php
```diff
@@ -0,0 +1,74 @@
...
+
+	/**
+	 * Enqueues an event to be recorded asynchronously.
+	 *
+	 * @param Telemetry_Event $event The event to record.
+	 * @return bool|WP_Error True if the event was enqueued for recording.
+	 *                       False if the event is not recordable.
+	 *                       WP_Error if the event is generating an error.
+	 */
+	public function record_event_asynchronously( Telemetry_Event $event ): bool|WP_Error {
```
sjinks, line 49: Intersection types are a PHP 8.1 feature; do we still have sites running PHP 8.0?

telemetry/tracks/class-tracks-client.php
```diff
@@ -0,0 +1,86 @@
...
+			'events'      => $valid_events,
+			'commonProps' => $common_props,
+		];
+
+		$response = $this->http->post(
+			static::TRACKS_ENDPOINT,
+			array(
+				'body'       => wp_json_encode( $body ),
+				'user-agent' => 'viptelemetry',
+				'headers'    => array(
```
sjinks, line 66: Nitpick: could we please use a uniform syntax for arrays? Sometimes we use `array()`, sometimes `[]`.

telemetry/class-telemetry-system.php
```diff
@@ -0,0 +1,32 @@
...
+	 * @param string               $event_name The event's name.
+	 * @param array<string, mixed> $event_properties Any additional properties
+	 *                                               to include with the event.
+	 * @return bool|WP_Error True if recording the event succeeded.
+	 *                       False if telemetry is disabled.
+	 *                       WP_Error if recording the event failed.
+	 */
+	abstract public function record_event(
+		string $event_name,
+		array $event_properties = array()
```
sjinks: Same

telemetry/class-tracks.php
```diff
@@ -0,0 +1,80 @@
...
+	 * Event queue.
+	 *
+	 * @var Telemetry_Event_Queue
+	 */
+	private Telemetry_Event_Queue $queue;
+
+	/**
+	 * @param array<string, mixed> The global event properties to be included with every event.
+	 */
+	private array $global_event_properties = array();
```
sjinks: Same

sjinks (approved):
LGTM except for https://github.com/Automattic/vip-go-mu-plugins/pull/5869/files#r1767343471

93g8kq6v40541

telemetry/class-telemetry-event-queue.php
sanmai: It is a 8.0 feature: https://3v4l.org/rgSko#v8.0.30

telemetry/tracks/class-tracks-event.php
sanmai: @sjinks Can you please have another look to see if this approach has a different fault? It is now deferred: we are not calling `wp_get_current_user()` from the constructor.

telemetry/tracks/class-tracks-client.php
sanmai: Done!

mehmoodak:
Thanks for working on this.

PR #5902, opened 2024-10-01
fix(two-factor): move `load_plugin_textdomain()` to `init`

sjinks:
## Description

WordPress Nightly now spits a [doing-it-wrong deprecation warning](https://github.com/WordPress/wordpress-develop/blob/522a9091c3a48553e61800de1967c7d77ff26e1c/src/wp-includes/l10n.php#L1002-L1013) when `load_plugin_textdomain()` is invoked too early (like in `plugins_loaded`).

This breaks our tests, such as E2E tests in VIP CLI.

For now, we fix the affected plugins locally and then create a PR to the upstream.

Ref: https://github.com/Automattic/vip-cli/actions/runs/11130390882/job/30929621217?pr=2045

### Fixed
- Two Factor plugin: address a deprecation warning about `load_plugin_textdomain()`

## Pre-review checklist

- [ ] This change uses a rollout method to ease deployment (if applicable - especially for large-scale actions requiring writes).
- [ ] This change has relevant documentation additions/updates (if applicable).

## Pre-deploy checklist

## Steps to Test

Tests must pass.

### DIFF
--- a/shared-plugins/two-factor/class-two-factor-core.php
+++ b/shared-plugins/two-factor/class-two-factor-core.php
@@ -93,7 +93,7 @@ class Two_Factor_Core {
 	 * @since 0.1-dev
 	 */
 	public static function add_hooks( $compat ) {
-		add_action( 'plugins_loaded', array( __CLASS__, 'load_textdomain' ) );
+		add_action( 'init', array( __CLASS__, 'load_textdomain' ) );
 		add_action( 'init', array( __CLASS__, 'get_providers' ) );
 		add_action( 'wp_login', array( __CLASS__, 'wp_login' ), 10, 2 );
 		add_filter( 'wp_login_errors', array( __CLASS__, 'maybe_show_reset_password_notice' ) );


shared-plugins/two-factor/class-two-factor-core.php
```diff
@@ -93,7 +93,7 @@ class Two_Factor_Core {
 	 * @since 0.1-dev
 	 */
 	public static function add_hooks( $compat ) {
-		add_action( 'plugins_loaded', array( __CLASS__, 'load_textdomain' ) );
+		add_action( 'init', array( __CLASS__, 'load_textdomain' ) );
```
sjinks, line 96: Technically, this could go into `after_setup_theme` but according to

https://github.com/WordPress/wordpress-develop/blob/522a9091c3a48553e61800de1967c7d77ff26e1c/src/wp-includes/l10n.php#L1002-L1013

> Attempted to load translations for the `xxx` domain too early. Translations should be loaded after the `after_setup_theme` action has fired, to ensure that the current user is already set up.

Current user is set after `after_setup_theme`; that's why I have chosen `init`.

rinatkhaziev (approved):

PR #5920, opened 2024-10-09
Update integration's config management

WPprodigy:
There are two ways for an integration is be activated:

1) When an integration is activated via the platform mechanisms, then we'll always use the configs that come from the injected configmap.
2) When an integration is activated manually, we'll use whatever config is passed in during the code-activation. Notably, these manual activations will not pull their config from the configmap.

Behavior-wise, this PR doesn't really change the above. It does change some of the underlining ways that configs are retrieved though. Notably, there are now two methods for fetching the config: `Integration->get_env_config()` and `Integration->get_network_site_config()` rather than just one that attempts to merge those two.

This allows integrations to make decisions on their own about what data they want to find at the env level and what data they want to check for at the network_site level (if a multisite). So instead of having to cascade "all or nothing", integrations will have the ability to pick and choose how to interpret their own cascading needs.

### DIFF
--- a/integrations/integration-vip-config.php
+++ b/integrations/integration-vip-config.php
@@ -130,13 +130,11 @@ class IntegrationVipConfig {
 			return Org_Integration_Status::BLOCKED;
 		}
 
-		$env_status = $this->get_value_from_config( 'env', 'status' );
-
-		if ( is_multisite() ) {
-			return $this->get_value_from_config( 'network_sites', 'status' ) ? : $env_status;
+		if ( ! is_multisite() ) {
+			return $this->get_value_from_config( 'env', 'status' );
 		}
 
-		return $env_status;
+		return $this->get_value_from_config( 'network_sites', 'status' ) ? : $this->get_value_from_config( 'env', 'status' );
 	}
 
 	public function get_env_config() {
--- a/integrations/integration-vip-config.php
+++ b/integrations/integration-vip-config.php
@@ -132,7 +132,7 @@ class IntegrationVipConfig {
 	}
 
 	public function get_env_config() {
-		return $this->get_value_from_config( 'env', 'config' ) ?? [];
+		return $this->get_value_from_config( 'env', 'config' );
 	}
 
 	public function get_network_site_config() {
@@ -140,7 +140,7 @@ class IntegrationVipConfig {
 			return [];
 		}
 
-		return $this->get_value_from_config( 'network_sites', 'config' ) ?? [];
+		return $this->get_value_from_config( 'network_sites', 'config' );
 	}
 
 	/**
--- a/integrations/integration-vip-config.php
+++ b/integrations/integration-vip-config.php
@@ -126,28 +126,21 @@ class IntegrationVipConfig {
 			return Org_Integration_Status::BLOCKED;
 		}
 
-		// Look into network_sites config before and then fallback to env config.
-		return $this->get_value_from_config( 'network_sites', 'status' ) ??
-			$this->get_value_from_config( 'env', 'status' );
+		// Look into network_sites config first, and then fallback to env config.
+		$network_site_status = is_multisite() ? $this->get_value_from_config( 'network_sites', 'status' ) : null;
+		return $network_site_status ? $network_site_status : $this->get_value_from_config( 'env', 'status' );
 	}
 
-	/**
-	 * Get site config.
-	 *
-	 * @return array
-	 */
-	public function get_site_config() {
-		if ( is_multisite() ) {
-			$config = $this->get_value_from_config( 'network_sites', 'config' );
-			// If network site config is not found then fallback to env config if it exists
-			if ( empty( $config ) && true === $this->get_value_from_config( 'env', 'cascade_config' ) ) {
-				$config = $this->get_value_from_config( 'env', 'config' );
-			}
-		} else {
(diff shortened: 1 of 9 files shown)


integrations/integration.php
```diff
@@ -78,33 +76,15 @@ public function __construct( string $slug ) {
...
 
-	/**
-	 * Callback for `switch_blog` filter.
-	 */
-	public function switch_blog_callback(): void {
-		// Updating config to make sure `get_config()` returns config of current blog instead of main site.
-		if ( isset( $this->vip_config ) ) {
-			$this->options['config'] = $this->vip_config->get_site_config();
-		}
-	}
```
WPprodigy, line 106: This is no longer needed, because `get_network_site_config()` will always pull the config from the configmap based on the current network site's ID. It's not stored in this intermediary `$options['config']` anymore -- that is only used when integrations are manually activated w/ a custom config array.

integrations/integrations.php
```diff
@@ -66,14 +68,12 @@ public function activate_platform_integrations() {
 			$vip_config = $this->get_integration_vip_config( $slug );
 
 			if ( $vip_config->is_active_via_vip() ) {
-				$this->activate( $slug, [
-					'config' => $vip_config->get_site_config(),
-				] );
+				$this->activate( $slug );
```
WPprodigy, line 71: No longer need to be passing in a "cached" version of `config` during platform activation, instead we'll just be pulling directly from the vip_config for these.

integrations/integration-vip-config.php
```diff
@@ -126,28 +126,21 @@ public function get_site_status() {
 			return Org_Integration_Status::BLOCKED;
 		}
 
-		// Look into network_sites config before and then fallback to env config.
-		return $this->get_value_from_config( 'network_sites', 'status' ) ??
-			$this->get_value_from_config( 'env', 'status' );
+		// Look into network_sites config first, and then fallback to env config.
+		$network_site_status = is_multisite() ? $this->get_value_from_config( 'network_sites', 'status' ) : null;
+		return $network_site_status ? $network_site_status : $this->get_value_from_config( 'env', 'status' );
```
rinatkhaziev: two ternaries in the row read a bit silly.

integrations/integration-vip-config.php
```diff
@@ -126,28 +126,21 @@ public function get_site_status() {
...
+		return $this->get_value_from_config( 'env', 'config' ) ?? [];
+	}
+
+	public function get_network_site_config() {
+		if ( ! is_multisite() ) {
+			return [];
 		}
 
-		return $config ?? array(); // To keep function signature consistent.
+		return $this->get_value_from_config( 'network_sites', 'config' ) ?? [];
```
rinatkhaziev, line 143: missing docblock

integrations/integration-vip-config.php
```diff
@@ -126,28 +126,21 @@ public function get_site_status() {
...
-	public function get_site_config() {
-		if ( is_multisite() ) {
-			$config = $this->get_value_from_config( 'network_sites', 'config' );
-			// If network site config is not found then fallback to env config if it exists
-			if ( empty( $config ) && true === $this->get_value_from_config( 'env', 'cascade_config' ) ) {
-				$config = $this->get_value_from_config( 'env', 'config' );
-			}
-		} else {
-			$config = $this->get_value_from_config( 'env', 'config' );
+	public function get_env_config() {
```
rinatkhaziev: missing docblock

integrations/integration-vip-config.php
```diff
@@ -126,28 +126,21 @@ public function get_site_status() {
...
+		return $this->get_value_from_config( 'env', 'config' ) ?? [];
+	}
+
+	public function get_network_site_config() {
```
rinatkhaziev: missing docblock

integrations/integration-vip-config.php
```diff
@@ -159,26 +152,19 @@ public function get_site_config() {
 	 * @return null|string|array Returns `null` if key is not found, `string` if key is "status" and `array` if key is "config".
```
rinatkhaziev, line 162: This seems confusing and not entirely accurate? There's no special handling for `config` vs `status`

WPprodigy: No special handling, but in the configmap the `config` is an array and `status` is a string. So the return type does vary based on that.

integrations/integration-vip-config.php
```diff
@@ -115,39 +115,42 @@ public function is_active_via_vip(): bool {
...
+	 * For single sites simply return global status.
+	 * For multisites, try to get status based on current blog id,
+	 * if not found then fallback to global environment status.
 	 *
-	 * @private
+	 * @return string|null
 	 */
 	public function get_site_status() {
 		if ( $this->get_value_from_config( 'org', 'status' ) === Org_Integration_Status::BLOCKED ) {
 			return Org_Integration_Status::BLOCKED;
```
rebeccahum, line 128: Should we get rid of the status level differentiation here as enums while we're at it?

rebeccahum (approved):
Tested and works, just one minor thing:

integrations/integration-vip-config.php
WPprodigy: I think probably going to end up removing org information entirely. Isn't really a level of hierarchy/inheritance that the application should need to worry about.

PR #5968, opened 2024-11-01
FORNO-1632: Add job to clean up redundant VaultPress options

t-wright:
## Description

Adds a job to remove redundant `_vp_ai_ping` options [added by VaultPress](https://github.com/Automattic/vaultpress/blob/958c3d2ea6ca638105a1e6a85f58726b8911be6a/vaultpress.php#L1434). 

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

1. Check out PR.
1. Add an option `wp option add _vp_ai_ping_12345`
1. Set `ENABLE_VIP_VAULTPRESS_PING_CLEANUP` to `true`
1. `wp cron-control events list` and verify `vip_vaultpress_ping_cleanup` is scheduled
2. After the event runs, `wp option get _vp_ai_ping_12345` should return an error
2. Run the event again `wp cron event run vip_vaultpress_ping_cleanup`
1. Verify completion option is set `wp option get vip_vaultpress_ping_cleanup_complete`
2. Verify `vip_vaultpress_ping_cleanup` is no longer scheduled

### DIFF
--- a/lib/class-vip-vaultpress-ping-cleanup.php
+++ b/lib/class-vip-vaultpress-ping-cleanup.php
@@ -6,6 +6,9 @@ class VIP_VaultPress_Ping_Cleanup {
 	const CRON_HOOK           = 'vip_vaultpress_ping_cleanup';
 	const CRON_INTERVAL       = 'hourly';
 
+	// How many options to delete per cron run.
+	const QUERY_LIMIT = 100;
+
 	public static function init(): void {
 		if ( ! class_exists( 'VaultPress' ) && defined( 'ENABLE_VIP_VAULTPRESS_PING_CLEANUP' ) && true === ENABLE_VIP_VAULTPRESS_PING_CLEANUP ) {
 			add_action( 'init', array( __CLASS__, 'schedule_cron' ), 99999 );
@@ -44,7 +47,13 @@ class VIP_VaultPress_Ping_Cleanup {
 		global $wpdb;
 
 		// phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching
-		return $wpdb->query( $wpdb->prepare( "DELETE FROM $wpdb->options WHERE option_name LIKE %s", self::VP_PING_OPTION_NAME ) );
+		return $wpdb->query(
+			$wpdb->prepare(
+				"DELETE FROM $wpdb->options WHERE option_name LIKE %s LIMIT %d",
+				self::VP_PING_OPTION_NAME,
+				self::QUERY_LIMIT
+			)
+		);
 	}
 
 	private static function requires_cleanup(): bool {
--- a/lib/class-vip-vaultpress-ping-cleanup.php
+++ b/lib/class-vip-vaultpress-ping-cleanup.php
@@ -1,0 +1,69 @@
+<?php
+
+class VIP_VaultPress_Ping_Cleanup {
+	const VP_PING_OPTION_NAME = '_vp_ai_ping%';
+	const OPTION_NAME         = 'vip_vaultpress_ping_cleanup_complete';
+	const CRON_HOOK           = 'vip_vaultpress_ping_cleanup';
+	const CRON_INTERVAL       = 'hourly';
+
+	// How many options to delete per cron run.
+	const QUERY_LIMIT = 500;
+
+	public static function init(): void {
+		if ( ! class_exists( 'VaultPress' ) && defined( 'ENABLE_VIP_VAULTPRESS_PING_CLEANUP' ) && true === ENABLE_VIP_VAULTPRESS_PING_CLEANUP ) {
+			add_action( 'init', array( __CLASS__, 'schedule_cron' ), 99999 );
+			add_action( self::CRON_HOOK, array( __CLASS__, 'do_cron' ) );
+		}
+	}
+
+	public static function schedule_cron(): void {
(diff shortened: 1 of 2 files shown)


lib/class-vip-vaultpress-ping-cleanup.php
```diff
@@ -0,0 +1,69 @@
...
+		return $wpdb->query( $wpdb->prepare( "DELETE FROM $wpdb->options WHERE option_name IN ( $option_name_placeholders )", $option_names ) );
+	}
+
+	private static function get_vaultpress_option_names(): array {
+		global $wpdb;
+
+		// phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching
+		return $wpdb->get_col(
+			$wpdb->prepare(
+				"SELECT option_name FROM $wpdb->options WHERE option_name LIKE %s LIMIT %d",
```
rebeccahum: Wondering if we should just straight up delete the ones matching the pattern since this saves a DB fetch of ~500 rows. I think a fetch is only good if we are doing any extra checking of the option name (which it looks like we aren't)? However, if you want to check for the existence of it before marking it as completed, it might be better to just do a limit of 1 in this sense -- and for that, we can probably just use `SELECT 1` rather than `SELECT option_name` for faster row retrieval in the DB.

t-wright: That sounds good - thanks for the suggestion. I wondered whether there might be cases where there were so many of these options that it would be safer to delete in batches, but seems [it was limited to 100](https://github.com/Automattic/vaultpress/commit/14ea02389e38f981885c80b02cdbec568c3b16b6) in 2018 so I suppose that's unlikely.

lib/class-vip-vaultpress-ping-cleanup.php
```diff
@@ -0,0 +1,63 @@
...
+		}
+
+		self::delete_options();
+	}
+
+	private static function delete_options(): int|bool {
+		global $wpdb;
+
+		// phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching
+		return $wpdb->query( $wpdb->prepare( "DELETE FROM $wpdb->options WHERE option_name LIKE %s", self::VP_PING_OPTION_NAME ) );
```
rebeccahum: I still think we should keep the limit on it to be extra safe. What if there was an import that went haywire?

t-wright: Good call! [83ae640](https://github.com/Automattic/vip-go-mu-plugins/pull/5968/commits/83ae64096ccf2aa8528784858497620dfe0171dc)

rinatkhaziev:
I don't think this needs to be a part of MU-plugins, seems like a job for a one-off script in _that_ repo.

t-wright:
> I don't think this needs to be a part of MU-plugins, seems like a job for a one-off script in _that_ repo.

@rinatkhaziev 👋 the goal here is to clean up the options left over by VaultPress since its [removal from mu-plugins](https://github.com/Automattic/vip-go-mu-plugins/pull/5767). 

There can be a bunch of these leftover:

```
select option_name, round(length(option_value)/1024/1024, 2) as 'MB'
from wp_options
where option_name like '%ai_ping%';

+-------------------+------+
| option_name       | MB   |
+-------------------+------+
| _vp_ai_ping_21910 | 0.85 |
| _vp_ai_ping_21915 | 0.41 |
| _vp_ai_ping_21982 | 1.27 |
| _vp_ai_ping_21984 | 1.27 |
| _vp_ai_ping_21986 | 1.27 |
| _vp_ai_ping_21988 | 2.02 |
| _vp_ai_ping_22060 | 0.85 |
| _vp_ai_ping_22064 | 2.02 |
| _vp_ai_ping_22083 | 1.27 |
| _vp_ai_ping_22101 | 1.27 |
| _vp_ai_ping_22105 | 1.27 |
...
```

Open to other options for performing the cleanup if you folks have a preferred method.

lib/class-vip-vaultpress-ping-cleanup.php
```diff
@@ -0,0 +1,72 @@
...
+
+		if ( $completed && wp_next_scheduled( self::CRON_HOOK ) ) {
+			wp_clear_scheduled_hook( self::CRON_HOOK );
+			return;
+		}
+
+		if ( false === $completed && ! wp_next_scheduled( self::CRON_HOOK ) ) {
+			wp_schedule_event( time(), self::CRON_INTERVAL, self::CRON_HOOK );
+		}
+	}
```
WPprodigy, line 30: It would be best to not have to run this code on every request. Rather than managing a separate cron job for this task, what if we instead hooked onto an existing one, maybe `a8c_cron_control_clean_legacy_data`?

Can hook onto the hook and have it check to ensure VaultPress is not currently active. And if not, then delete all the excess legacy VP data. I'm also not so sure we need to store the results in a new option called `vip_vaultpress_ping_cleanup_complete`, as that doesn't seem very friendly if the data were ever to come back around (after a data sync or sql import for example). If Vaultpress isn't active and `requires_cleanup()` returns false (which is a light query), then we're good.

Since that job runs once a day, let's do more than 100 cleanups per run. Can loop over a 100 at a time, but I don't see any problem with just getting them all purged out.

rebeccahum:
Closing in favour of what was discussed in Slack.

PR #6016, opened 2024-11-26
VIP Parse.ly: Check for Support User class exists in the vip-parsely plugin

vaurdan:
## Description

This PR adds a new defensive check in the VIP Parse.ly plugin, that prevents the VIP-specific customizations of the WP Parse.ly plugin to load when `WP_INSTALLING` is set and true. 

This fixes a fatal error being thrown when accessing the `example.com/wp-activate.php` URL. Since the `vip-support` plugin is not loaded when `WP_INSTALLING` is defined, this extra validation short circuits the `wp_parsely_current_user_can_use_pch_feature` callback to return the original value, if the `Automattic\VIP\Support_User\User` class does not exists.

```
PHP Fatal error:  Uncaught Error: Class "Automattic\VIP\Support_User\User" not found in /var/www/wp-content/mu-plugins/vip-parsely/vip-parsely.php:93
Stack trace:
#0 /var/www/wp-includes/class-wp-hook.php(312): {closure}(NULL, 'smart_linking', Object(WP_User))
#1 /var/www/wp-includes/plugin.php(205): WP_Hook->apply_filters(NULL, Array)
(...)
```

### Fixed
- VIP Parse.ly: Fixed a fatal error that happened in certain conditions within the VIP Parse.ly plugin.

## Pre-review checklist

## Pre-deploy checklist


sjinks (approved):

mjangda:
This still isn't quite the best fix and likely bound to cause other problems in the future. It suggests that someone (customer code or our code) is doing something in a context they shouldn't.

If our short-term goal is to just stop the fatals from happening, I would just do a `class_exists` check inside the callback for `wp_parsely_current_user_can_use_pch_feature` before we use the `Automattic\VIP\Support_User\User` class. (It's a bit of anti-pattern since most of mu-plugins code runs on the assumption that key VIP code like the Support User functionality is correctly loaded inside late hooks / filters).

vaurdan:
Thank you for the feedback @mjangda!

> It suggests that someone (customer code or our code) is doing something in a context they shouldn't.

I don't think that is anything wrong per-se in the customer code that triggers this issue: there is an API preload that runs on `wp_footer` and uses `rest_preload_api_request`. This API preloading consequently initializes the REST API (via `rest_api_init`), which then also initializes the Parse.ly REST API (and all its dependencies). 

Since `wp-activate.php` calls the `wp_footer` action, it will also run on this context, loading the REST API, then Parse.ly, and then the callback - `wp_parsely_current_user_can_use_pch_feature` - that triggers the fatal. 

I think part of the problem is having the [VIP Support plugin not loading on this specific scenario](https://github.com/Automattic/vip-go-mu-plugins/blob/develop/vip-support.php#L10) (`WP_INSTALLING` defined and true). I'm sure there's a reason for that, however a safer alternative would be to validate if the constant is not true only in the necessary blocks of the code that shouldn't run during WP installs/upgrades.

> It's a bit of anti-pattern since most of mu-plugins code runs on the assumption that key VIP code like the Support User functionality is correctly loaded inside late hooks / filters

Yes, that's partly the reason why I tried to keep that logic outside the `wp_parsely_current_user_can_use_pch_feature` callback. But I do see why you saying that it might cause problems in the future - right now, nothing on the Parse.ly plugin needs to run when `WP_INSTALLING` is set, however if that changes this could cause unexpected issues.

I'm going to change it to check for the class inside the callback as suggested, so we can stop the fatals that are currently happening.

mjangda (approved):

sjinks (approved):

acicovic:
@sjinks, the Copilot review doesn't seem to have worked. Are we OK to merge this? If yes, feel free to do so.

PR #6048, opened 2024-12-10
exploring basic Pendo tracking

smithjw1:
## Description

This is a jumping off point to adding Pendo tracking to the WordPress admin. When you wanted to include Pendo, you could insert the tracking code with a hook like:

```
use Automattic\VIP\Telemetry\Pendo;
function insert_pendo() {
	$pendo = new Pendo();
	$pendo->include_script();
}
add_action( 'admin_enqueue_scripts', 'insert_pendo', 10, 3 );
```

The plan is to enable Pendo tracking when a customer is using PCH or Composable to start, with future integrations. We only want to enable this once, so we may want to invoke it outside of those integrations and check to see if they are running.

There are some missing pieces, so this draft PR shouldn't be merged as is, but is posted here as a way to further the implementation discussion.

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

The logic will not fire Pendo anywhere but on production, so local testing isn't possible.


smithjw1:
Superseeded by: https://github.com/Automattic/vip-go-mu-plugins/pull/6140

PR #6101, opened 2025-01-17
Add Jetpack integration loader, add the way to load config in local dev env

rinatkhaziev:
## Description

There are two things happening:

Added JetpackIntegration class - very basic, but it does set up the version respecting both integrations or the legacy way.

Added ability to override config values via two filters:

`apply_filters( 'vip_integrations_config_file_path', $config_file_path_orig, $slug );`

This one allows us to override the file path in non-VIP environments (dev env/Codespaces)

`apply_filters( 'vip_integrations_pre_load_config', null, $config_file_path, $slug );`
Following core WP pattern, adding the `pre_` filter will short-circuit the prod logic and will allow to inject the values as needed.

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

1. Add `integrations.json` into the root of dev env project with the following content:

```
{
	"jetpack": {
		"status": "enabled",
		"config": {
			"version": "14.0"
		}
	},
	"parsely": {
		"status": "disabled",
		"config": {
			"apikey": "myexampledomain.test"
		}
	}
}
```
2. In jetpack.php comment out the call to `vip_jetpack_load`
3. Refresh a page and verify that Jetpack is indeed loaded.
4. Flip the status and verify Jetpack is not loaded anymore.

### DIFF
--- a/integrations/integration-vip-config.php
+++ b/integrations/integration-vip-config.php
@@ -61,11 +61,7 @@ class IntegrationVipConfig {
 	 * @param string $slug A unique identifier for the integration.
 	 */
 	private function set_config( string $slug ): void {
-		if ( defined( 'VIP_GO_APP_ENVIRONMENT' ) && 'local' === constant( 'VIP_GO_APP_ENVIRONMENT' ) ) {
-			$config = $this->get_local_config_from_file( $slug );
-		} else {
-			$config = $this->get_vip_config_from_file( $slug );
-		}
+		$config = $this->get_vip_config_from_file( $slug );
 
 		if ( ! is_array( $config ) ) {
 			return;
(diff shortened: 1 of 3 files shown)


integrations/integration-vip-config.php
```diff
@@ -61,7 +61,11 @@ public function __construct( string $slug ) {
 	 * @param string $slug A unique identifier for the integration.
 	 */
 	private function set_config( string $slug ): void {
-		$config = $this->get_vip_config_from_file( $slug );
+		if ( defined( 'VIP_GO_APP_ENVIRONMENT' ) && 'local' === constant( 'VIP_GO_APP_ENVIRONMENT' ) ) {
+			$config = $this->get_local_config_from_file( $slug );
+		} else {
```
sjinks: Do we need this? We can define the `WPVIP_INTEGRATIONS_CONFIG_DIR` in Dev Environments as well. This will make it possible to reuse this code in Codespaces.

sjinks: Or maybe something like this (WordPress way):

```php
$config = apply_filters( 'integrations_load_config', null, $slug );
if ( is_null ( $config ) ) {
    $config = $this->get_vip_config_from_file( $slug );
}

$config = apply_filters( 'integrations_post_process_config', $config );
```

sjinks: Filter names are, of course, to be discussed :-)

If we use this approach, we can delegate the loading logic to the `dev-env-plugin.php`: this will work both in dev environments and codespaces (the current `get_local_config_from_file()` implementation will not work with codespaces).

The `post_process` filter may be used to filter out some data (I am not sure this is necessary, but I like to have such a possibility).

rinatkhaziev: Well, each config is split per file and is a PHP snippet, I don't think we want that in local env?

sjinks (approved):

PR #6140, opened 2025-02-12
Add support for Pendo Track events

chriszarate:
## Description

Provide telemetry for Pendo "Track events." These events are conceptually similar to Automattic Tracks events, but have a different payload structure. See an explanation of ["Event types" on this page](https://support.pendo.io/hc/en-us/articles/23335876011803-Events-overview).

Usage is almost identical to that of Automattic Tracks. This PR introduces a generic `Telemetry` class that will send events to both Automattic Tracks and Pendo. Going forward, we will update our plugins to use the `Telemetry` class.

In general, I've tried to follow the implementation pattern provided by the existing Tracks implementation and the generic Telemetry base classes and interfaces. I did make some minor changes to the abstract class `Telemetry_Event` and the extending `Tracks_Event` class in order to improve the abstraction and reduce code duplication.

New constants will need to be provided in order to configure this functionality automatically:

- `VIP_PENDO_TRACK_INTEGRATION_KEY`: The integration API key used to authenticate with Pendo and send Track events. This key can only be used to write (append) events to Pendo's API.
- `VIP_SF_ACCOUNT_ID`: The account ID added to event payloads and used to correlate Pendo events.

> [!NOTE]
> Client-side Pendo Page/Feature events has moved to https://github.com/Automattic/vip-go-mu-plugins/pull/6150

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

1. Check out PR.
1. Follow README to use the `Telemetry` class and record events.


chriszarate:
I've added guards against requiring missing files during rollout in https://github.com/Automattic/vip-go-mu-plugins/pull/6140/commits/436d3771ee7d9153daba963019f5187e6873803f. There are no external consumers of these files, so nothing to guard on the usage side.

000-vip-init.php
```diff
@@ -234,15 +234,35 @@
...
+// Make sure all telemetry files are present before loading them.
+$safe_to_load_telemetry = array_reduce( $telemetry_files, function ( bool $carry, string $file ): bool {
+	return $carry && file_exists( $file );
+}, true );
+
+if ( true === $safe_to_load_telemetry ) {
+	foreach ( $telemetry_files as $file ) {
+		require_once $file;
+	}
+}
```
rinatkhaziev, line 265: Clever

telemetry/class-telemetry-event.php
```diff
@@ -16,12 +16,59 @@
...
+			return $data;
+		}
+
+		return true;
+	}
+
+	/**
+	 * Returns the event's data for JSON representation.
+	 */
+	public function jsonSerialize(): mixed {
```
rinatkhaziev, line 65: nitpick but violating WP naming conventions for functions and methods (snake_case over camelCase)

telemetry/pendo/class-pendo-track-event.php
rinatkhaziev, line 1: Do we want the snake case validation/erroring? Maybe simply auto-converting would be more practical approach? We can still flag for user if need be.

telemetry/class-telemetry-event.php
chriszarate: Good call out. However, this method is implementing the `JsonSerializable` interface from PHP core so that we can customize the behavior of `json_encode`. Also, this is relocated code from `Tracks_Event`.

https://www.php.net/manual/en/class.jsonserializable.php

telemetry/pendo/pendo-utils.php
```diff
@@ -0,0 +1,77 @@
...
+	$is_vip_user = $event_props['is_vip_user'];
+	$user_id     = $is_vip_user ? 'vip-' . $user_props['_ui'] : $user_props['_ui'];
+	$vip_org_id  = $event_props['vip_org'] ?? 'unknown';
+	$sf_org_id   = defined( 'VIP_SF_ACCOUNT_ID' ) ? constant( 'VIP_SF_ACCOUNT_ID' ) : sprintf( 'nosfid_wordpress_%s', $vip_org_id );
+
+	return [
+		'account_id'     => (string) $sf_org_id,
+		'country_code'   => sanitize_text_field( $_SERVER['GEOIP_COUNTRY_CODE'] ?? 'unknown' ),
+		'org_id'         => (string) $vip_org_id,
+		'role_wordpress' => $wp_user->roles[0], // The suffix helps prevent collisions with other contexts like the VIP Dashboard.
```
rinatkhaziev: Let's guard against any extra weirdness in case there are no roles.
```suggestion
		'role_wordpress' => $wp_user->roles[0] ?? '', // The suffix helps prevent collisions with other contexts like the VIP Dashboard.
```

telemetry/pendo/class-pendo-track-event.php
chriszarate: Generally I followed the behavior of the existing Tracks implementation, which also returns an error for non-snake-case property names. I don't have a strong opinion, but all calling code will be first-party / VIP, so compliance should be easy. Also, these errors are just for signaling and logging; they are never surfaced to the UI. When an event fails validation it is logged and ignored. Let me know what you think and I can address.

https://github.com/Automattic/vip-go-mu-plugins/blob/436d3771ee7d9153daba963019f5187e6873803f/telemetry/tracks/class-tracks-event.php#L219-L222

telemetry/pendo/pendo-utils.php
chriszarate: Done in https://github.com/Automattic/vip-go-mu-plugins/pull/6140/commits/832a255b1d9d70b2d142c76ee9ec849b74f1b4b9 with test. Thanks!

rinatkhaziev (approved):

PR #6150, opened 2025-02-20
Implement Pendo JavaScript library (Page/Feature events)

chriszarate:
## Description

Provide Pendo-based telemetry for client-side Pendo Page/Feature events. These events require a client library to be enqueued and configured. "Track events" can also be sent client-side, but we should avoid this and prefer the server-side implementation.

Usage requires a single static function call to enable the client-side library (`Pendo::enable_javascript_library()`).

## Pre-review checklist

## Pre-deploy checklist


smithjw1 (commented):
Question about how enablement is controlled. I see the `edit_posts` restriction, but not the other items [mentioned here](https://vipproductp2.wordpress.com/2024/11/07/project-thread-integrating-pendo-into-wordpress/#comment-1480). As well a check to a org level opt out (synced from GOOP).

chriszarate:
> I see the `edit_posts` restriction, but not the other items [mentioned here](https://vipproductp2.wordpress.com/2024/11/07/project-thread-integrating-pendo-into-wordpress/#comment-1480).

@smithjw1 These checks were initially separated. I updated the code and added tests In a way that should make it more obvious and reviewable.

- ✅ Nonproduction Environments. Covered by `Pendo::is_pendo_enabled_for_environment()`
- ✅ Where our plugin is installed. Not enforced by mu-plugins, but instead de facto covered by the absence of plugin code calling `Pendo::enable_javascript_library()`
- ✅ Only for users with `edit_posts` capability
- ⛔ On certain URLs. Per [this comment](https://wp.me/pcCPcL-1hs#comment-1749) I did not implement this restriction.
- ✅ Outside of VIP. Covered by `Pendo::is_pendo_enabled_for_environment()`

> As well a check to a org level opt out (synced from GOOP).

That is implemented in two ways:

1. The presence of `VIP_DISABLE_PENDO_TELEMETRY=true` which disables all Pendo telemetry.
2. The absence of `VIP_PENDO_SNIPPET_API_KEY` which disables just the snippet enqueuing.

000-vip-init.php
```diff
@@ -245,6 +245,7 @@
 require_once __DIR__ . '/telemetry/tracks/class-tracks-client.php';
 require_once __DIR__ . '/telemetry/tracks/tracks-utils.php';
 require_once __DIR__ . '/telemetry/pendo/class-pendo.php';
+require_once __DIR__ . '/telemetry/pendo/class-pendo-javascript-library.php';
```
rinatkhaziev: This may trigger intermittent deploy error due to how deploy system works, needs to be guarded against with `file_exists` checks and if this class is a hard dependency somewhere that'd need to be wrapped as well. The checks can be removed after the initial deploy. Ping me in Slack if you need more details about it. @chriszarate

telemetry/pendo/class-pendo-javascript-library.php
```diff
@@ -0,0 +1,167 @@
...
+
+		// A null response indicates that the user is not logged-in or we otherwise
+		// have no meaningful way to attribute the event.
+		if ( empty( $user_properties ) ) {
+			$message = __( 'User is not logged-in', 'vip-telemetry' );
+			log2logstash( [
+				'severity' => 'error',
+				'feature'  => 'telemetry',
+				'message'  => $message,
+			] );
```
rinatkhaziev, line 153: why do we want to log this?

chriszarate: All of the `log2logstash` calls highlight problems that should not be possible if the code is being called correctly. I mostly following the pattern of existing "doing it wrong" logging in the Tracks classes ([example](https://github.com/Automattic/vip-go-mu-plugins/blob/f666cb9c0e899a8c970562a1fad75b6e5433b83a/telemetry/tracks/class-tracks-event.php#L192)).

chriszarate: To actually answer your question: Pendo Telemetry is restricted to users logged in to the WP Admin dashboard. This error likely indicates that the library is being called in a context that it shouldn't be, or attribution is unexpectedly not working.

rinatkhaziev (approved):

chriszarate:
@rinatkhaziev I made a little bit of a mess on this PR after your review, but the changes are just:

- Merging `develop`
- https://github.com/Automattic/vip-go-mu-plugins/pull/6150/commits/6bbfbcf9e29a3d0396d8cdbebbf26347b67bcb23
   - Fixing a property name
- https://github.com/Automattic/vip-go-mu-plugins/pull/6150/commits/34793345f9edaee1587a2143855b84d43b51ae44
   - Removing an unactionable error log so that we can easily disable this functionality by withholding the constant
- https://github.com/Automattic/vip-go-mu-plugins/pull/6150/commits/e3f851e78c1248c96b9271a3b75e8fa7844f2f5d
   - Initializing the class that enqueues the JS. Has no effect until the constant is supplied.
- Additional tests

PR #6204, opened 2025-03-24
two-factor: Compat with Jetpack Account Protection

mjangda:
Ensure that the two-factor plugin and Jetpack Account Protection work well together.

See https://wp.me/peb6dq-3Cr#comment-2273

See https://github.com/Automattic/jetpack/pull/42657

### Added

- We've made sure that the upcoming Jetpack Account Protection feature works well with the Two Factor plugin.

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

1. Check out PR.
1. Install Jetpack 14.5 Beta
1. Log in with a user with a compromised password and no MFA. Should be forced to go through the Jetpack verification flow
2. Log in with a user with a compromised password and MFA active. Should be required to do MFA step and then be logged in.

### DIFF
--- a/two-factor.php
+++ b/two-factor.php
@@ -276,3 +276,17 @@ function wpcom_vip_two_factor_admin_notice() {
 	</div>
 	<?php
 }
+
+/**
+ * Jetpack Account Protection does not factor in a user's MFA status.
+ *
+ * If a user has MFA active, we should skip the Account Protection check.
+ */
+add_filter( 'jetpack_account_protection_user_requires_protection', 'wpcom_vip_two_factor_bypass_jetpack_account_protection', 10, 2 );
+function wpcom_vip_two_factor_bypass_jetpack_account_protection( $user_requires_protection, $user ) {
+	if ( Two_Factor_Core::is_user_using_two_factor( $user->ID ) ) {
+		return false;
+	}
+
+	return $user_requires_protection;
+}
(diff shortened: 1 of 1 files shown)


sjinks (approved):

two-factor.php
```diff
@@ -276,3 +276,17 @@ function wpcom_vip_two_factor_admin_notice() {
...
 }
+
+/**
+ * Jetpack Account Protection does not factor in a user's MFA status.
+ *
+ * If a user has MFA active, we should skip the Account Protection check.
+ */
+add_filter( 'jetpack_account_protection_user_requires_protection', 'wpcom_vip_two_factor_bypass_jetpack_account_protection', 10, 2 );
+function wpcom_vip_two_factor_bypass_jetpack_account_protection( $user_requires_protection, $user ) {
+	if ( Two_Factor_Core::is_user_using_two_factor( $user->ID ) ) {
```
mjangda: Do we need a class/method exists check here? Is `Two_Factor_Core` always available?

andrea-sdl: In theory it should be available, but I'm not sure if there are edge cases when this could not be available (broken sites?)

mjangda: Looks like a filter can be used to skip loading:

https://github.com/Automattic/vip-go-mu-plugins/blob/070adaeb2b77919659e411b4317e1a05ba23e3d2/two-factor.php#L168-L171

mjangda: Implemented here: 8ab371d90c74b290162817540cfd307ad497a69f

andrea-sdl (approved):

PR #6215, opened 2025-03-26
Config: Add heartbeat syncs

WPprodigy:
## Description

Rather than sending the full site details payload every time, this will opt for sending a heartbeat checkin instead when the data hasn't changed. So there are now two sync types: full sync and heartbeat sync.

1) Full syncs will trigger whenever the data has changed, or at a minimum of once per day.
2) Heartbeat syncs will trigger every 25 minutes when the data hasn't changed.

## Pre-deploy checklist

- Ensure new `/sites/heartbeat` endpoint has been deployed.

### DIFF
--- a/config/class-site-details-index.php
+++ b/config/class-site-details-index.php
@@ -20,6 +20,11 @@ class Site_Details_Index {
 	 */
 	private $timestamp = null;
 
+	const SYNC_DATA_OPTION = 'vip_config_sync_data';
+
+	const MINUTE_IN_MS = MINUTE_IN_SECONDS * 1000;
+	const DAY_IN_MS    = DAY_IN_SECONDS * 1000;
+
 	/**
 	 * Standard singleton except accept a timestamp for mocking purposes.
 	 *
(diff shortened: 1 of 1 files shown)


config/class-site-details-index.php
```diff
@@ -349,32 +354,127 @@ public function get_site_details() {
...
+			return 'full';
+		}
 
-			vip_safe_wp_remote_request( rtrim( $url, '/' ) . '/sites', false, 3, 5, 10, $args );
+		// Send a heartbeat if it's been more than 25 minutes (1/3 of the stale threshold).
+		if ( $current_timestamp - $last_sync_timestamp > self::MINUTE_IN_MS * 25 ) {
+			return 'heartbeat';
 		}
+
+		return 'none';
```
rebeccahum, line 435: Is this necessary if we aren't handling the value anywhere? Does it need to be handled?

config/class-site-details-index.php
```diff
@@ -349,32 +354,127 @@ public function get_site_details() {
...
-					'Authorization' => 'Bearer ' . $token,
-					'Content-Type'  => 'application/json',
-				),
-			);
+	/**
+	 * Determine if we need a full sync, a heartbeat, or none at all.
+	 */
+	private function determine_sync_type( $site_details, $sync_data ) {
+		$current_timestamp        = $site_details['timestamp'];
+		$last_sync_timestamp      = $sync_data['last_synced'] ?? 0;
```
rebeccahum, line 407: I'm seeing this same statement twice, once declared here and one below the comment `// Send a heartbeat if it's been more than 25 minutes (1/3 of the stale threshold).`

config/class-site-details-index.php
WPprodigy: For return consistency. There’s nothing to do for this case right now given the current control flow, but that’s okay.

config/class-site-details-index.php
WPprodigy: Hm, that should be fixed by a prev commit already

rebeccahum (approved):

PR #6246, opened 2025-04-16
Load the latest version for RDB

ingeniumed:
## Description

The purpose of this PR is to tweak the version that's loaded for RDB, so it's always the latest one found in mu-plugins. This way, we don't need to keep updating the version for the plugin in mu-plugins and instead can have it dynamically be updated by what's deployed via mu-plugins-ext.

### Changed
- Dynamically load the latest version of the RDB plugin

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

1. Check out the PR
2. Create a dev-env with this PR, as well as with latest mu-plugins-ext
3. Activate remote-data-blocks
4. Ensure the latest version of the plugin is loaded. I made a tweak to the settings UI to verify this.

### DIFF
--- a/integrations/remote-data-blocks.php
+++ b/integrations/remote-data-blocks.php
@@ -43,7 +43,8 @@ class RemoteDataBlocksIntegration extends Integration {
 				return;
 			}
 
-			// Get all the entries in the path of WPVIP_MU_PLUGIN_DIR/vip-integrations/remote-data-blocks-<version>/ and check what versions are available.
+			// Get all the entries in the path of WPVIP_MU_PLUGIN_DIR/vip-integrations/remote-data-blocks-<version>/
+			// and check what versions are available.
 			$versions = $this->get_versions();
 
 			// if no versions are found, return early.
@@ -51,11 +52,6 @@ class RemoteDataBlocksIntegration extends Integration {
 				$this->is_active = false;
 				return;
 			}
-
-			// Sort the versions in descending order.
-			uksort($versions, function ( $a, $b ) use ( $versions ) {
-				return version_compare( $versions[ $b ], $versions[ $a ] );
-			});
 
 			// Load the latest version of the plugin.
 			$latest_directory = array_key_first( $versions );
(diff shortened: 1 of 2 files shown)


shekharnwagh (dismissed):
I tested this locally with `vip dev-env` (copied `/vip-integrations` from `vip-go-mu-plugins-ext`) and it works. The implementation uses file system operations (`is_dir()`, `scandir()`) on every request - would this be a performance concern? Would it be worth adding caching here? The entries in that directory would be limited, so maybe not, but interested in hearing your thoughts.

maxschmeling:
> I tested this locally with `vip dev-env` (copied `/vip-integrations` from `vip-go-mu-plugins-ext`) and it works. The implementation uses file system operations (`is_dir()`, `scandir()`) on every request - would this be a performance concern? Would it be worth adding caching here? The entries in that directory would be limited, so maybe not, but interested in hearing your thoughts.

I think this is ok for now. The scanning should go away once we start basing the version on config.

integrations/remote-data-blocks.php
```diff
@@ -48,14 +43,52 @@ public function load(): void {
...
+		if ( is_dir( $dir ) ) {
+			$scan_entries = scandir( $dir );
+			foreach ( $scan_entries as $entry ) {
+				if ( preg_match( '/^remote-data-blocks-(\d+\.\d+)$/', $entry, $matches ) && is_dir( $dir . $entry ) && file_exists( $dir . $entry . '/remote-data-blocks.php' ) ) {
+					$versions[ $entry ] = $matches[1]; // Store directory name as key and version as value
+				}
+			}
+		}
+
+		return $versions;
```
rinatkhaziev, line 100: Few things here. 
1) Rather than doing scandir and then sorting in the loader you can simply  `scandir( '/var/www/wp-content/mu-plugins/vip-integrations', SCANDIR_SORT_DESCENDING );`

```php
=> array(11) {
  [7]=>
  string(23) "remote-data-blocks-0.12"
  [8]=>
  string(23) "remote-data-blocks-0.11"
```

2) That `preg_match` seems redundant. It can be simplified with a `str_contains` check:  `str_contains( $entry, 'remote-data-blocks-' )`
Then to get a version you'd just need to 
```php
explode( 'remote-data-blocks-', 'remote-data-blocks-0.11' );
=> array(2) {
  [0]=>
  string(0) ""
  [1]=>
  string(4) "0.11"
```
 
3) All in all it's a bit hard to read, better update the `is_dir` check to return `$versions` early rather than wrapping the whole block in it.

integrations/remote-data-blocks.php
```diff
@@ -48,14 +43,52 @@ public function load(): void {
...
+			}
+
+			// Sort the versions in descending order.
+			uksort($versions, function ( $a, $b ) use ( $versions ) {
+				return version_compare( $versions[ $b ], $versions[ $a ] );
+			});
+
+			// Load the latest version of the plugin.
+			$latest_directory = array_key_first( $versions );
+			$load_path        = WPVIP_MU_PLUGIN_DIR . '/vip-integrations/' . $latest_directory . '/remote-data-blocks.php';
```
rinatkhaziev: See the feedback in `get_versions` re: how we can simplify this

rinatkhaziev (changes requested):
Thanks for the PR, I left some feedback inline.

integrations/remote-data-blocks.php
shekharnwagh: `scandir` with `SCANDIR_SORT_DESCENDING` didn't work well with versions - it returns `0.9 > 0.11` since it uses alphabetical sort. I ended up incorporating suggestions 2 and 3 and moved the sorting to `get_versions` from `load` in 4a804fdf8.

rinatkhaziev (approved):

integrations/remote-data-blocks.php
ingeniumed: I should have noted this as a comment in the PR but I had tried `scandir` with `SCANDIR_SORT_DESCENDING`, and it didn't work for this very reason. 

2 & 3 are good points, thanks for mentioning that and thanks @shekharnwagh for making those changes 👍🏾

PR #6267, opened 2025-04-23
CI (WP Core tests): Remove new test added by 6.8 (N/A to VIP)

rebeccahum:
## Description
We disable it at https://github.com/Automattic/vip-go-mu-plugins/blob/b8db989b922fec48af1c27999aa792cd28c3fe5c/a8c-files.php#L701-L704

The test was added per https://github.com/WordPress/wordpress-develop/commit/7daa963e8061308b3bccb681e494487851371a9d. It will never scale down because it will never reach the "big threshold".

### DIFF
--- a/.github/workflows/core-tests.yml
+++ b/.github/workflows/core-tests.yml
@@ -106,7 +106,10 @@ jobs:
         working-directory: wordpress
 
       - name: Run PHPUnit tests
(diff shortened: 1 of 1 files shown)


sjinks:
Does [this thing](https://github.com/Automattic/vip-go-mu-plugins/blob/16682e8a3882430abae6045ecb2f57062f0d9b28/a8c-files.php#L146) interfere with the tests?

rebeccahum:
@sjinks no that’s not the cause

sjinks (approved):

PR #6275, opened 2025-04-29
Adds stats tracking for XML RPC usage

brunobasto:
## Description

We want to gather stats about which types of password are used when authenticating xmlrpc requests. This will give us insights to guide our decisions around security controls and recommendations.

### Added
- A filter on `authenticate` to track the type of password used to authenticate: `user_pass`, `app_ass` or `cookie`. This filter will only track authenticated XML RPC requests.

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

Perform some authenticated XML-RPC requests, like:

```bash
curl -i \
  -H 'Content-Type: text/xml' \
  --data '<?xml version="1.0"?>
<methodCall>
  <methodName>wp.getUsersBlogs</methodName>
  <params>
    <param><value><string>vipgo</string></value></param>
    <param><value><string>user_pass|app_pass</string></value></param>
  </params>
</methodCall>' \
  http://vip-security-boost.vipdev.lndo.site/xmlrpc.php
```

And check that the appropriate pixel is sent.


rinatkhaziev:
@brunobasto I think ideally we want to use the new Telemetry functionality https://github.com/Automattic/vip-go-mu-plugins/tree/develop/telemetry#using-tracks-and-pendo-individually

brunobasto:
> @brunobasto I think ideally we want to use the new Telemetry functionality https://github.com/Automattic/vip-go-mu-plugins/tree/develop/telemetry#using-tracks-and-pendo-individually

Oh nice! I'll refactor. I think we'll want to send it just to Tracks (not Pendo), but I see the API allows for that. Thanks @rinatkhaziev

rinatkhaziev:
It _should_ handle that, but I don't think we never used it just yet 😬

brunobasto:
> It _should_ handle that, but I don't think we never used it just yet 😬

It does, but I ran into a problem:

The `authenticate` filter is too soon for the Telemetry framework. It relies on `wp_get_current_user()` and that won't be set at that time.

so I went with a two-step approach that stores the password type used to authenticate on a static variable `XML_RPC_Auth_Tracker::xmlrpc_password_type` and then, on the `xmlrpc_call` action, we send the event.

brunobasto:
> [5 Security Hotspots](https://sonarcloud.io/project/security_hotspots?id=Automattic_vip-go-mu-plugins&pullRequest=6275&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)

It's complaining about variables named `password`, but that's the accurate way of calling them AFAIK

stats.php
```diff
@@ -109,6 +117,85 @@ function track_file_delete() {
 	] );
 }
 
+/**
+ * Tracks the authentication type used during successful XML-RPC requests.
+ */
+function determine_xmlrpc_password_type( $user, $username, $password ) {
```
andrea-sdl: I wanted to test this behavior and I started digging into the WordPress core logic for the [xmlrpc-server](https://github.com/WordPress/WordPress/blob/3906e3e9d6d25225b41cda97f8e6f89f24e31211/wp-includes/class-wp-xmlrpc-server.php) to see how the authentication was handled.

One thing I noticed when testing this is that `WP_Application_Passwords::is_in_use()` will return `true`  even if we are using the user password.
This is caused by the `determine_xmlrpc_password_type`, I suspect some of the functions we call set this.

I don't think we want that that logic to change, we should find a way to introduce this tracking without changing any behavior.

stats.php
```diff
@@ -7,15 +7,23 @@
...
+
 // Limit tracking to production
 if ( true === WPCOM_IS_VIP_ENV && false === WPCOM_SANDBOXED ) {
 	add_action( 'transition_post_status', __NAMESPACE__ . '\track_publish_post', 9999, 2 );
 	add_filter( 'wp_handle_upload', __NAMESPACE__ . '\handle_file_upload', 9999 );
 	// Hook early because overrides in a8c-files and stream wrapper return empty.
 	// Which makes it hard to differentiate between full size and thumbs.
 	add_action( 'wp_delete_file', __NAMESPACE__ . '\handle_file_delete', -1, 1 );
+	// Determine the password type and store it in XML_RPC_Auth_Tracker
+	add_filter( 'authenticate', __NAMESPACE__ . '\determine_xmlrpc_password_type', 30, 3 ); // core authenticates on 20
```
andrea-sdl:
```
	add_filter( 'authenticate', __NAMESPACE__ . '\determine_xmlrpc_password_type', 30, 3 ); // core authenticates on 20
```

@brunobasto instead of this, is there a reason you didn't go for adding the `app_pass` only if the password was used?

For example we could use the `application_password_did_authenticate` ([source](https://github.com/WordPress/WordPress/blob/f5f3764a9feb2fdfde23a1498e2e07a3b5964e60/wp-includes/user.php#L497)) to detect when it's used, and default to `user_pass` if it's not being called.

Here's a rough example I tested locally (it requires removing the `authenticate` filter you added)
```
add_filter( 'application_password_did_authenticate', __NAMESPACE__ . '\application_password_did_authenticate' );
function application_password_did_authenticate( $user ) {
	// Only proceed if it's an XML-RPC request
	if ( ! ( defined( 'XMLRPC_REQUEST' ) && XMLRPC_REQUEST ) ) {
		return $user;
	}

	// We are only interested in successful authentication events.
	if ( is_wp_error( $user ) || ! ( $user instanceof \WP_User ) ) {
		return $user;
	}

	// Skip tracking for Jetpack requests.
	if ( vip_is_jetpack_request() ) {
		return;
	}
	XML_RPC_Auth_Tracker::$xmlrpc_password_type = 'app_pass';
}
```

andrea-sdl (changes requested):
Left some comments ;)

stats.php
brunobasto: > is there a reason you didn't go for adding the app_pass only if the password was used?

I didn't think of other way to do it. I wasn't aware of `application_password_did_authenticate`. I think I overlooked it when looking at https://github.com/WordPress/wordpress-develop/blob/6.8.1/src/wp-includes/user.php#L459

That's perfect, thank you.

brunobasto: Updated in https://github.com/Automattic/vip-go-mu-plugins/pull/6275/commits/c26c948c3935a3f28f307cf383bc8d2da51a7e66

stats.php
```diff
@@ -109,6 +117,46 @@ function track_file_delete() {
...
+		return;
+	}
+
+	XML_RPC_Auth_Tracker::track( $xmlrpc_method );
+}
+
+function application_password_did_authenticate( $user ) {
+	// Only proceed if it's an XML-RPC request
+	if ( ! ( defined( 'XMLRPC_REQUEST' ) && XMLRPC_REQUEST ) ) {
+		return $user;
```
andrea-sdl:
```suggestion
		return;
```
@brunobasto the `application_password_did_authenticate` is an [action hook](https://developer.wordpress.org/reference/hooks/application_password_did_authenticate/) and not a filter. Therefore the `return $user` is not used anywhere, we can just `return`

stats.php
```diff
@@ -109,6 +117,46 @@ function track_file_delete() {
...
+
+	// Skip tracking for Jetpack requests.
+	if ( vip_is_jetpack_request() ) {
+		return;
+	}
+
+	XML_RPC_Auth_Tracker::track( $xmlrpc_method );
+}
+
+function application_password_did_authenticate( $user ) {
```
andrea-sdl:
```suggestion
function maybe_set_xml_rpc_auth_tracker_type( $user ) {
```
nitpick, not a blocker, but I think it would make it more clear when reading it.

andrea-sdl (commented):
@brunobasto thank you for this! PR works fine, I only left couple of suggestions on the `application_password_did_authenticate` part :)

andrea-sdl (approved):

PR #6279, opened 2025-05-04
Improve/ignore platform required paths in new relic

mchanDev:
## Description

By default, New Relic runs for every request, including internal metrics endpoints which get hit often. Unfortunately, this has a negative impact on the accuracy because these internal endpoints are very fast, leading to skewing of stats like avg response time.  

### Added

- Added `newrelic_ignore_transaction` to some internal endpoints that should not be tracked in New Relic to ensure accuracy of the app stats.

## Pre-review checklist

## Pre-deploy checklist

### DIFF
--- a/prometheus/endpoint.php
+++ b/prometheus/endpoint.php
@@ -10,6 +10,10 @@ use Prometheus\Storage\APCng;
 use Prometheus\Storage\InMemory;
 
 // @codeCoverageIgnoreStart -- this is a standalone endpoint which doens't run in the context of the WP tests
+if ( extension_loaded( 'newrelic' ) ) { // Ensure PHP agent is available
+	newrelic_ignore_transaction();
+}
+
 require_once __DIR__ . '/vendor/autoload.php';
 require_once dirname( __DIR__ ) . '/lib/utils/class-context.php';
 


walthowd (dismissed):

sjinks (dismissed):
LGTM but would like a second option from @rinatkhaziev

http-concat.php
```diff
@@ -6,6 +6,9 @@
  * Author: Automattic
  * Version: 1.0
  */
+if ( extension_loaded( 'newrelic' ) ) { // Ensure PHP agent is available
+	newrelic_ignore_transaction();
+}
```
rinatkhaziev: This will not work. Requests for actual concat are served from a different endpoint (https://github.com/Automattic/nginx-http-concat/blob/master/ngx-http-concat.php)

prometheus.php
```diff
@@ -10,6 +10,10 @@
 use Automattic\VIP\Prometheus\Potential_Multi_Dataset_Queries_Collector;
 use Automattic\VIP\Prometheus\Multisite_Stats_Collector;
 // @codeCoverageIgnoreStart -- this file is loaded before tests start
+if ( extension_loaded( 'newrelic' ) ) { // Ensure PHP agent is available
+	newrelic_ignore_transaction();
+}
+
```
rinatkhaziev: Same as the above, but the endpoint is here: 

https://github.com/Automattic/vip-go-mu-plugins/blob/develop/prometheus/endpoint.php

rinatkhaziev (changes requested):
Thanks for the PR, unfortunately as-is is not going to work, I left inline feedback.

mchanDev:
As per feedback from @rinatkhaziev 
The PR has had the newrelic_ignore_transaction moved as noted in the PR comments

And an additional PR opened against the nginx-http-concat repo
https://github.com/Automattic/nginx-http-concat/pull/84

rinatkhaziev (approved):
looks good, thanks!

rinatkhaziev (approved):

PR #6301, opened 2025-05-13
skip loading remote data blocks on unsupported WP versions

shekharnwagh:
## Description

Adds version compatibility check before loading the remote data blocks integration. The integration will only load if the current WordPress version meets or exceeds the minimum_wp_version specified in config. If no minimum version is configured, the integration loads normally. This check is required as currently Remote Data Blocks plugin relies on functionality not available in older versions of WordPress (<6.7).

### Added
- Added a WordPress version compatibility check before loading the remote data blocks integration.

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

1. Check out PR.
2. Setup local dev environment with MU Plugins pointed to this checked out branch.

	```sh
	# Create a dev environment from a VIP deployed app with integrations configured and VIP GO MU Plugins pointed to the checked out branch
	vip dev-env create @app_id.environment_type -s <local_dev_environment_slug> -u <path_to_checked_out_branch>

	# Start the dev environment
	vip dev-env start <local_dev_environment_slug> --vscode
	```

3. When the VS Code window opens up, edit the `integrations/integrations-config/integrations.json` file to add a `remote-data-blocks` integration config -

	```json
	"remote-data-blocks": {
		"type": "remote-data-blocks",
		"org": {
			"status": "enabled"
		},
		"env": {
			"status": "enabled",
			"config": {
				"version": "latest",
				"minimum_wp_version": "6.7"
			}
		}
	}
	```

4. Verify that RDB integration is loaded as default WP version is greater than 6.7. 
5. Try changing the local dev environment's WordPress version to to different version and verify that RDB integration is loaded or not loaded based on the WordPress version.

	```sh
	# Change the WordPress version of the local dev environment
	vip dev-env update -s <local_dev_environment_slug> -w <wordpress_version>

	# Restart the local dev environment
	vip dev-env start <local_dev_environment_slug>
	```

### DIFF
--- a/integrations/remote-data-blocks.php
+++ b/integrations/remote-data-blocks.php
@@ -25,6 +25,11 @@ class RemoteDataBlocksIntegration extends Integration {
 		return defined( 'REMOTE_DATA_BLOCKS__LOADED' );
 	}
 
+	public function get_wp_version(): string {
+		global $wp_version;
+		return $wp_version;
+	}
+
 	/**
 	 * Returns `true` if the current WordPress version is supported by Remote Data Blocks.
 	 *
@@ -33,7 +38,7 @@ class RemoteDataBlocksIntegration extends Integration {
 	 *              version is defined, `false` otherwise.
 	 */
 	public function is_supported_wp_version(): bool {
-		global $wp_version;
+		$wp_version = $this->get_wp_version();
 
 		$config = $this->get_env_config();
 		if (
--- a/integrations/remote-data-blocks.php
+++ b/integrations/remote-data-blocks.php
@@ -25,6 +25,11 @@ class RemoteDataBlocksIntegration extends Integration {
 		return defined( 'REMOTE_DATA_BLOCKS__LOADED' );
 	}
 
+	/**
+	 * Returns the current WordPress version.
+	 *
+	 * Wraps the global `$wp_version` variable in a function to make it easier to mock in tests.
+	 */
 	public function get_wp_version(): string {
 		global $wp_version;
 		return $wp_version;
--- a/integrations/remote-data-blocks.php
+++ b/integrations/remote-data-blocks.php
@@ -26,35 +26,28 @@ class RemoteDataBlocksIntegration extends Integration {
 	}
 
 	/**
-	 * Returns the current WordPress version.
-	 *
-	 * Wraps the global `$wp_version` variable in a function to make it easier to mock in tests.
-	 */
-	public function get_wp_version(): string {
-		global $wp_version;
-		return $wp_version;
-	}
-
-	/**
 	 * Returns `true` if the current WordPress version is supported by Remote Data Blocks.
 	 *
 	 * @return bool `true` if the current WordPress version greater than or equal to the minimum
-	 *              WordPress version defined in the environment config or if no minimum WordPress
-	 *              version is defined, `false` otherwise.
+	 *              WordPress version defined in the environment config, `false` otherwise.
 	 */
 	public function is_supported_wp_version(): bool {
-		$wp_version = $this->get_wp_version();
+		$wp_version = get_bloginfo( 'version' );
 
 		$config = $this->get_env_config();
 		if (
 			isset( $config['minimum_wp_version'] ) &&
 			is_string( $config['minimum_wp_version'] ) &&
-			version_compare( $wp_version, $config['minimum_wp_version'], '<' )
+			version_compare( $wp_version, $config['minimum_wp_version'], '>=' )
 		) {
-			return false;
+			return true;
 		}
 
-		return true;
+		/**
+		 * Default to false if the minimum WordPress version is not set to avoid fatals due to
+		 * Remote Data Blocks being loaded on unsupported WordPress versions.
+		 */
+		return false;
 	}
 
 	/**
--- a/integrations/remote-data-blocks.php
+++ b/integrations/remote-data-blocks.php
@@ -26,6 +26,28 @@ class RemoteDataBlocksIntegration extends Integration {
 	}
 
 	/**
+	 * Returns `true` if the current WordPress version is supported by Remote Data Blocks.
+	 *
+	 * @return bool `true` if the current WordPress version greater than or equal to the minimum
+	 *              WordPress version defined in the environment config or if no minimum WordPress
+	 *              version is defined, `false` otherwise.
+	 */
+	public function is_supported_wp_version(): bool {
+		global $wp_version;
+
+		$config = $this->get_env_config();
+		if (
+			isset( $config['minimum_wp_version'] ) &&
+			is_string( $config['minimum_wp_version'] ) &&
+			version_compare( $wp_version, $config['minimum_wp_version'], '<' )
+		) {
+			return false;
+		}
+
+		return true;
+	}
+
+	/**
 	 * Loads the plugin.
 	 *
 	 * This is called after the integration is activated and configured.
@@ -40,6 +62,11 @@ class RemoteDataBlocksIntegration extends Integration {
 			// In activate() method we do make sure to not activate the integration if its already loaded
 			// but still adding it here as a safety measure i.e. if load() is called directly.
 			if ( $this->is_loaded() ) {
+				return;
+			}
+
+			if ( ! $this->is_supported_wp_version() ) {
+				$this->is_active = false;
 				return;
 			}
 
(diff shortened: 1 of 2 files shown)


integrations/remote-data-blocks.php
```diff
@@ -25,6 +25,33 @@ public function is_loaded(): bool {
...
+	}
+
+	/**
+	 * Returns `true` if the current WordPress version is supported by Remote Data Blocks.
+	 *
+	 * @return bool `true` if the current WordPress version greater than or equal to the minimum
+	 *              WordPress version defined in the environment config or if no minimum WordPress
+	 *              version is defined, `false` otherwise.
+	 */
+	public function is_supported_wp_version(): bool {
```
ingeniumed, line 34: I would move this to the base `integration` class instead, because this isn't going to be the only integration that would use it. That way, every integration only needs line 73 to actually prevent loading the integration.

integrations/remote-data-blocks.php
```diff
@@ -25,6 +25,33 @@ public function is_loaded(): bool {
 		return defined( 'REMOTE_DATA_BLOCKS__LOADED' );
 	}
 
+	public function get_wp_version(): string {
```
ingeniumed: It would be good to add a comment here as to why this is implemented this way.

integrations/remote-data-blocks.php
ingeniumed: I understand why this is done this way, ignore this for now.

integrations/remote-data-blocks.php
shekharnwagh: a12fdbf53

integrations/remote-data-blocks.php
```diff
@@ -25,6 +25,38 @@ public function is_loaded(): bool {
 		return defined( 'REMOTE_DATA_BLOCKS__LOADED' );
 	}
 
+	/**
+	 * Returns the current WordPress version.
+	 *
+	 * Wraps the global `$wp_version` variable in a function to make it easier to mock in tests.
+	 */
+	public function get_wp_version(): string {
+		global $wp_version;
```
chriszarate: In this codebase, I more commonly see `get_bloginfo( 'version' )` to obtain the current WP version.

integrations/remote-data-blocks.php
```diff
@@ -25,6 +25,38 @@ public function is_loaded(): bool {
...
+	 *              WordPress version defined in the environment config or if no minimum WordPress
+	 *              version is defined, `false` otherwise.
+	 */
+	public function is_supported_wp_version(): bool {
+		$wp_version = $this->get_wp_version();
+
+		$config = $this->get_env_config();
+		if (
+			isset( $config['minimum_wp_version'] ) &&
+			is_string( $config['minimum_wp_version'] ) &&
```
chriszarate, line 40: Shouldn't this method return `false` if `$config['minimum_wp_version']` is not correctly set, instead of defaulting to `true`?

integrations/remote-data-blocks.php
```diff
@@ -43,6 +75,11 @@ public function load(): void {
 				return;
 			}
 
+			if ( ! $this->is_supported_wp_version() ) {
+				$this->is_active = false;
```
chriszarate, line 72: What is the point of `$this->is_active`? It is never read, nor is it declared as a public class var.

tests/integrations/test-remote-data-blocks.php
```diff
@@ -215,4 +215,67 @@ public function test_get_versions_ignores_non_matching_directories(): void {
...
+		$integration_mock->method( 'get_wp_version' )->willReturn( $current_wp_version );
+		$this->assertEquals( $expected_result, $integration_mock->is_supported_wp_version() );
+	}
+
+	public static function wp_version_support_provider(): array {
+		return [
+			'no minimum_wp_version in config' => [ '6.7', [], true ],
+			'wp_version less than minimum'    => [ '5.6', [ 'minimum_wp_version' => '6.7' ], false ],
+			'wp_version equal to minimum'     => [ '6.7.1', [ 'minimum_wp_version' => '6.7' ], true ],
+			'wp_version greater than minimum' => [ '6.8.2', [ 'minimum_wp_version' => '6.7' ], true ],
```
chriszarate: Could use some tests with patch versions and equalities
```suggestion
			'wp_version greater than minimum' => [ '6.8.2', [ 'minimum_wp_version' => '6.7' ], true ],
			'wp_version equal to minimum' => [ '6.7', [ 'minimum_wp_version' => '6.7' ], true ],
			'wp_version equal to minimum but more specific' => [ '6.7.0', [ 'minimum_wp_version' => '6.7' ], true ],
			'wp_version equal to minimum but less specific' => [ '6.7', [ 'minimum_wp_version' => '6.7.0' ], true ],
```

tests/integrations/test-remote-data-blocks.php
```diff
@@ -215,4 +215,67 @@ public function test_get_versions_ignores_non_matching_directories(): void {
...
+		$integration_mock->method( 'get_env_config' )->willReturn( $config ?? [] );
+		$integration_mock->method( 'get_wp_version' )->willReturn( $current_wp_version );
+		$this->assertEquals( $expected_result, $integration_mock->is_supported_wp_version() );
+	}
+
+	public static function wp_version_support_provider(): array {
+		return [
+			'no minimum_wp_version in config' => [ '6.7', [], true ],
+			'wp_version less than minimum'    => [ '5.6', [ 'minimum_wp_version' => '6.7' ], false ],
+			'wp_version equal to minimum'     => [ '6.7.1', [ 'minimum_wp_version' => '6.7' ], true ],
```
chriszarate: they are not equal

integrations/remote-data-blocks.php
shekharnwagh: Made it return `true` to account for case where GOOP change was deployed after this or the configmap changes weren't propagated. Now that GOOP is deployed, I think we could default to being strict where if `minimum_wp_version` is not in config we don't load the plugin. 

Though `minimum_wp_version` acts like an additional filter, and wouldn't absence of a filter indicate that needn't be applied ?

tests/integrations/test-remote-data-blocks.php
shekharnwagh: > 'wp_version equal to minimum but less specific' => [ '6.7', [ 'minimum_wp_version' => '6.7.0' ], true ]

This fails since `version_compare` treats `6.7` as less than `6.7.0`. Is this something which might happen with wordpress versions ? If yes, we could patch the logic to handle such cases. When I check the WP releases page - https://wordpress.org/download/releases/, the versions were like `6.7`, `6.7.1`, `6.7.2` with `6.7.0` not mentioned. That might indicate we don't get cases like `6.7.0` but I don't see anything wrong with that. If we decide to patch it should we only worry about patch version being absent/present or extend that for other cases where versions contains texts like `release`, `rc`, `beta` ?

integrations/remote-data-blocks.php
shekharnwagh: I found `global $wp_version` being used in 38 files vs `get_bloginfo( 'version' )` used in 15 files, if that matters. TBH I only searched for the first and it, being so widely used, settled on that without checking if there is any other way. The `get_bloginfo( 'version' )` does seem better, given that it's a function which might be easier to mock.

tests/integrations/test-remote-data-blocks.php
chriszarate: I don't know the answer, kinda why I wanted more test cases. I bet @rinatkhaziev has good advice here

rinatkhaziev: WP doesn't follow semver. So it's '6.7' => '6.7.1'. pre-releases are `X.Y-beta1', 'X.Y-rc1`, etc.

integrations/remote-data-blocks.php
maxschmeling: > Though minimum_wp_version acts like an additional filter, and wouldn't absence of a filter indicate that needn't be applied ?

This is how I would look at it. If the value isn't there, don't apply the logic and let it load like normal.

chriszarate: I see that logic. But since this is designed to prevent fatals on customer sites, I was thinking more from a security / protective stance. If we expect this guard to always be in place, do we want it to "fail open" (permissive) or "fail closed" (restrictive)?

chriszarate: Ie, what happens if a bug in the code that provides this config accidentally stops providing `minimum_wp_version`? What do we want to happen?

tests/integrations/test-remote-data-blocks.php
shekharnwagh: Added few more tests for pre-release versions, but the way it is now, works for above scenarios

integrations/remote-data-blocks.php
shekharnwagh: It is exposed via this public method in the parent - https://github.com/Automattic/vip-go-mu-plugins/blob/ddc0ff307245ae183e9a56f984115fba3d49bf3a/integrations/integration.php#L93-L95

maxschmeling (approved):
I'm marking this as approved since I tested it locally and it's working. Still need to address open discussion items.

integrations/remote-data-blocks.php
shekharnwagh: Moved to skip loading RDB when `minimum_wp_version` is not avaiable

shekharnwagh:
> I'm marking this as approved since I tested it locally and it's working. Still need to address open discussion items.

I think I've replied to or addressed all the comments. Waiting for review and an approval from someone from Cantina.

PR #6302, opened 2025-05-16
Adds log entry when user is locked out due to rate limits

brunobasto:
## Description

This pull request enhances the logging functionality of the `_vip_maybe_temporary_lock_account` function in `security.php` to provide more detailed information about user lockout events.

### Added
- Added a log entry using `log2logstash` to record detailed information whenever a user is temporarily locked out. This includes the username, lockout duration, lockout reason, IP address, and additional context such as the request URI and HTTP method.

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

1. Fail to login a few times until you get locked out
2. When you do get locked out, check your debug log files and make sure it contains a log entry similar to the following:

```json
Automattic\VIP\Logstash\Logger: {
    "site_id": 1,
    "blog_id": 1,
    "http_host": "vip-security-boost.vipdev.lndo.site",
    "severity": "warning",
    "feature": "a8c_vip_security",
    "message": "User vipgo locked out for 300 seconds due to ip_username from IP 172.17.2.1",
    "user_id": 0,
    "extra": "{\n    \"uri\": \"\\\/wp-login.php\",\n    \"http_method\": \"POST\",\n    \"username\": \"vipgo\",\n    \"lockout_type\": \"ip_username\",\n    \"lockout_duration\": 300,\n    \"ip_address\": \"172.17.2.1\"\n}",
    "timestamp": "2025-05-16 13:11:22",
    "index": "log2logstash",
    "http_user_agent": "Mozilla\/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit\/537.36 (KHTML, like Gecko) Chrome\/135.0.0.0 Safari\/537.36",
    "file": "\/wp\/wp-content\/mu-plugins\/security.php",
    "line": "208"
}
```

### DIFF
--- a/security.php
+++ b/security.php
@@ -203,7 +203,7 @@ function _vip_maybe_temporary_lock_account( $username, $cache_group ) {
 		\Automattic\VIP\Logstash\log2logstash( array(
 			'severity' => 'warning',
 			'feature'  => 'security',
-			'message'  => sprintf( 'User %s locked out for %d seconds due to %s from IP %s', $username, $lock_interval, $lock_reason, $ip ),
+			'message'  => sprintf( 'User locked out due to $lock_reason. Username: %s, IP: %s, Interval: %d', $username, $ip, $lock_interval ),
 			'extra'    => [
 				'uri'              => isset( $_SERVER['REQUEST_URI'] ) ? esc_url_raw( $_SERVER['REQUEST_URI'] ) : '',
 				'http_method'      => isset( $_SERVER['REQUEST_METHOD'] ) ? sanitize_text_field( $_SERVER['REQUEST_METHOD'] ) : '',
--- a/security.php
+++ b/security.php
@@ -197,8 +197,22 @@ function _vip_maybe_temporary_lock_account( $username, $cache_group ) {
 		 */
 		$lock_interval = apply_filters( "vip_{$event_type}_{$lock_reason}_lockout", $default_lockout, $username );
 
-
 		wp_cache_set( CACHE_KEY_LOCK_PREFIX . $cache_keys[ "{$lock_reason}_cache_key" ], true, $cache_group, $lock_interval ); // phpcs:ignore WordPressVIPMinimum.Performance.LowExpiryCacheTime.CacheTimeUndetermined
+
+		// Log which user was locked out, what type of lockout, the lockout duration, and the IP address
+		\Automattic\VIP\Logstash\log2logstash( array(
+			'severity' => 'warning',
+			'feature'  => 'security',
+			'message'  => sprintf( 'User %s locked out for %d seconds due to %s from IP %s', $username, $lock_interval, $lock_reason, $ip ),
+			'extra'    => [
+				'uri'              => isset( $_SERVER['REQUEST_URI'] ) ? esc_url_raw( $_SERVER['REQUEST_URI'] ) : '',
+				'http_method'      => isset( $_SERVER['REQUEST_METHOD'] ) ? sanitize_text_field( $_SERVER['REQUEST_METHOD'] ) : '',
+				'username'         => $username,
+				'lockout_type'     => $lock_reason,
+				'lockout_duration' => $lock_interval,
+				'ip_address'       => $ip,
+			],
+		) );
 	}
 }
 
--- a/security.php
+++ b/security.php
@@ -207,7 +207,7 @@ function _vip_maybe_temporary_lock_account( $username, $cache_group ) {
 			'extra'    => [
 				'uri'              => isset( $_SERVER['REQUEST_URI'] ) ? esc_url_raw( $_SERVER['REQUEST_URI'] ) : '',
 				'http_method'      => isset( $_SERVER['REQUEST_METHOD'] ) ? sanitize_text_field( $_SERVER['REQUEST_METHOD'] ) : '',
-				'username'         => $username,
+				'username'         => vip_strict_sanitize_username( $username ),
 				'lockout_type'     => $lock_reason,
 				'lockout_duration' => $lock_interval,
 				'ip_address'       => $ip,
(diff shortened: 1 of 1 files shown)


security.php
```diff
@@ -197,8 +197,22 @@ function _vip_maybe_temporary_lock_account( $username, $cache_group ) {
...
 		$lock_interval = apply_filters( "vip_{$event_type}_{$lock_reason}_lockout", $default_lockout, $username );
 
-
 		wp_cache_set( CACHE_KEY_LOCK_PREFIX . $cache_keys[ "{$lock_reason}_cache_key" ], true, $cache_group, $lock_interval ); // phpcs:ignore WordPressVIPMinimum.Performance.LowExpiryCacheTime.CacheTimeUndetermined
+
+		// Log which user was locked out, what type of lockout, the lockout duration, and the IP address
+		\Automattic\VIP\Logstash\log2logstash( array(
+			'severity' => 'warning',
+			'feature'  => 'security',
+			'message'  => sprintf( 'User %s locked out for %d seconds due to %s from IP %s', $username, $lock_interval, $lock_reason, $ip ),
```
pandah3: Can we sanitize the username and other fields too?
It looks like there's [`vip_strict_sanitize_username()`](https://github.com/Automattic/vip-go-mu-plugins/blob/develop/security.php#L38) for sanitizing usernames.

security.php
```diff
@@ -197,8 +197,22 @@ function _vip_maybe_temporary_lock_account( $username, $cache_group ) {
...
+			'extra'    => [
+				'uri'              => isset( $_SERVER['REQUEST_URI'] ) ? esc_url_raw( $_SERVER['REQUEST_URI'] ) : '',
+				'http_method'      => isset( $_SERVER['REQUEST_METHOD'] ) ? sanitize_text_field( $_SERVER['REQUEST_METHOD'] ) : '',
+				'username'         => $username,
+				'lockout_type'     => $lock_reason,
+				'lockout_duration' => $lock_interval,
+				'ip_address'       => $ip,
```
pandah3: Let's also sanitize these fields ^

pandah3 (commented):
Where are you seeing the logs? I'm running `vip dev-env logs --slug vip-local` but I don't see any Logstash logs. I apparently didn't have the mu-plugins repo locally (😅 ), so I had to go through the full local dev setup again. I might've missed something.

security.php
andrea-sdl: I wonder if we should use a separate "feature" for each of the login locks, so they are easier too search.

security.php
andrea-sdl: @pandah3 Dunno if that helps but we're eventually converting everything to a JSON here https://github.com/Automattic/vip-go-mu-plugins/blob/4e2c5b955cf477235002c532ce516184669fb5fb/logstash/class-logger.php#L541 so I think there might still be some sanitization going on. I think it's still good to sanitize data that comes from user-input, but noting this just in case.

security.php
rebeccahum: Maybe just a general message that can be grepped at the beginning: "User locked out due to $lock_reason. Username: <username>, IP: <ip>, Interval: <interval>"

security.php
brunobasto: Added sanitization of username here: https://github.com/Automattic/vip-go-mu-plugins/pull/6302/commits/642e783d1e82e18e926040092026430cdeadcf5f

I'm not sure we need to sanitize the other fields, as they are not user-input, they are fixed values:.

- lock_reason is either `username`, `ip_username` or `ip`
- lockout_duration is just a number
- ip_address also just an ip

security.php
brunobasto: > Can we sanitize the username and other fields too?

See https://github.com/Automattic/vip-go-mu-plugins/pull/6302#discussion_r2098849130

brunobasto: > I wonder if we should use a separate "feature" for each of the login locks, so they are easier too search.

Yeah, it's an option. I can update. @pandah3, objections?

brunobasto:
> Where are you seeing the logs? I'm running `vip dev-env logs --slug vip-local` but I don't see any Logstash logs. I apparently didn't have the mu-plugins repo locally (😅 ), so I had to go through the full local dev setup again. I might've missed something.

I usually open the `debug.log` file on the dev-env folder. Mine is at `/Users/bruno/.local/share/vip/dev-environment/my-dev-env/log/`

security.php
pandah3: I personally like keeping everything under a single `feature` to group all the security logs together. Can we move `lockout_type` to a top level field instead of keeping it under `extra` or is the log schema locked down? I think that + the more greppable message format that Rebecca mentioned would make it easier to search.

security.php
pandah3: Not sure if we get the IP from the header for reverse proxy sites, so it might good to sanitize for that, but yeah, probably not necessary otherwise.

security.php
pandah3: Hmm doesn't look like it 🤔 :
https://github.com/Automattic/vip-go-mu-plugins/blob/develop/logstash/class-logger.php#L88

I think making the message more greppable would still be good though!

pandah3 (approved):
Thank you 🙏 I see the logs as expected in `debug.log`!

security.php
brunobasto: > I think making the message more greppable would still be good though!

Suggestions?

Maybe

```php
'message'  => sprintf( '[user_lockout] User %s locked out for %d seconds due to %s from IP %s', $username, $lock_interval, $lock_reason, $ip ),
```

?

brunobasto: I think my GH UI wasn't updated, and I completely missed the message from Rebecca. I can update to the suggested format.

brunobasto: Applied suggestion here: https://github.com/Automattic/vip-go-mu-plugins/pull/6302/commits/b5c50b81e8e8ae1ae0c33a15f22cb783fc8c8f47

pandah3: @brunobasto Andrea noted that Jetpack uses separate `feature` values for similar cases, so we can follow that pattern here too.

PR #6303, opened 2025-05-16
feat: handle typejuggling error for ninja forms

daniel-ayvar:
## Description

Patches a [type juggling bug with ninjaforms](https://wordpress.org/support/topic/fix-type-juggling-error/) that causes a lot of db writes to occur on every wp-admin request. This resolves this by simply setting the value to a string to [avoid unnecessary db calls.](https://github.com/WordPress/WordPress/blob/c08c2e28cc77324f17f23ac0fcbfc786fd98524a/wp-includes/option.php#L906-L917)

### Added
- Patch that reduces db calls for "ninja_forms_needs_updates" updates in ninjaforms 

## Pre-review checklist

## Pre-deploy checklist


plugin-fixes.php
```diff
@@ -173,3 +173,26 @@ function vip_proxy_twitter_embed( $provider ) {
...
+* on every wp-admin request.
+*
+* This can lead to excessive `options` table churn and potential `alloptions` bloat,
+* especially on high-traffic editorial sites, triggering performance issues related to roles and rewrites.
+*
+* @see: https://plugins.trac.wordpress.org/browser/ninja-forms/trunk/ninja-forms.php#L541
+* @param mixed $value The new value to be saved.
+* @param string $old_value The old value to be replaced.
+*/
+function vip_ninja_forms_update_option( $value, $old_value ) {
```
rebeccahum, line 195: > Valid as of Ninja Forms v3.10.1

Is there a way we can only do this for affected versions?

rebeccahum (changes requested):

plugin-fixes.php
daniel-ayvar: Searched a bit and it seems this was introduced v3.4.34.2. Ill update this and only apply if the version >= 3.4.34.2.

plugin-fixes.php
```diff
@@ -173,3 +173,35 @@ function vip_proxy_twitter_embed( $provider ) {
...
+* @param mixed $value The new value to be saved.
+* @param string $old_value The old value to be replaced.
+* @return mixed The value to be saved.
+*/
+function vip_ninja_forms_update_option( $value, $old_value ) {
+	return ( 0 === $value && '0' === $old_value ) ? '0' : $value;
+}
+
+function vip_ninja_forms_setup() {
+	if (
```
rebeccahum, line 196: Nitpick, but since this function would need to be called all the time, I think we can just add the checks directly inside the filter since the filter will only run when updating that specific Ninja Forms option -- this will avoids needing to conditionally hook it at load time.

daniel-ayvar: Thanks for the feedback. I went ahead and implemented your feedback. Still learning what works with filters when it comes to performance.

rebeccahum (approved):
Thank you!

PR #6328, opened 2025-05-28
Speed up `process_entries_on_shutdown`

bor0:
Part of WCCOM-1276

## Description

As part of [Desire's recent FAR](https://alphamattic.wordpress.com/2025/04/17/desires-focus-area-review-2025-04-performance-and-reliability/), we noticed that the `process_entries_on_shutdown` event can sometimes take 15 seconds. The proposals here are microoptimizations, and a bit difficult to measure until we merge and test them out.

ℹ️ Reviewing this commit by commit, with whitespace disabled can make the review easier.

## Changelog Description

- Speed up `process_entries_on_shutdown` by applying some microoptimizations, and introducing a new filter to reduce I/O operations.

### Added
- Added a new filter `wpcom_vip_logstash_in_memory` that will store all logs in memory and dump them once, performing only single I/O operation, at the cost of memory increase.

## Pre-review checklist

## Pre-deploy checklist


logstash/class-logger.php
```diff
@@ -523,12 +523,35 @@ public static function process_entries_on_shutdown(): void {
...
+		if ( ( ! defined( 'VIP_GO_ENV' ) || ! VIP_GO_ENV || constant( 'VIP_GO_ENV' ) === 'local' ) && static::should_wp_debug_log_entries() ) {
+			foreach ( static::$entries as $entry ) {
+				static::wp_debug_log( $entry );
+			}
+
+			return;
+		}
+		$path = ( is_dir( '/chroot' ) ? '/chroot' : '' ) . '/tmp/logstash.log';
+
+		if ( apply_filters( 'wpcom_vip_logstash_in_memory', false ) ) {
```
rebeccahum, line 535: Wondering if we should just switch to the new way... @sjinks what do you think??

bor0: That would surely simplify this PR, but I wasn't sure if this could cause out of memory issues, depending on how large the logstash entries are...

sjinks: My only concern is how much memory this will require.

Typically, `error_log()` opens the file, writes data, and closes the file. These operations are pretty fast. If write operations take 15 seconds, we must be writing a lot of data.

And, IIRC, if PHP runs out of memory at some point, we will lose the buffered data (it won't be written to the error log). Is this acceptable for us?

bor0: > If write operations take 15 seconds

I'm not sure if writing is the bottleneck. We only got `process_entries_on_shutdown` from New Relic as a piece of info. Any other things that could cause a bottleneck there?

> Is this acceptable for us?

This is why I added a filter so that we can experiment and see the difference on WCCOM

sjinks (approved):

logstash/class-logger.php
sjinks: OK, let's test :-)

rinatkhaziev: @bor0 we can deploy this PR on a site of your choice so that you can test in real world scenario, but rather than trying microoptimizations we should look at the root cause. If we suspect that the we're adding a lot of logs this would be the first thing to prove/disprove.

bor0: @rinatkhaziev do you have any suggestions on how we can gather this data? This is as far as I got ([link](https://one.newrelic.com/nr1-core/apm-reports/performance/MjA0MjY2N3xBUE18QVBQTElDQVRJT058MTUxNDg0MjM2?duration=259200000&filters=selectedInstance%20IN%20%28%29&state=eea4c008-02f1-6613-f0af-c209c65b5a6a)):

Screenshot 2025-06-04 at 20 08 06

Another alternative is to check the codebase, but that might not be easy.

```
bor0:~/dev/woocommerce.com$ rg wc_get_logger | wc -l
     677
bor0:~/dev/woocommerce.com$ rg WC_Logger | wc -l
     201
```

rebeccahum: Did you want to look at the Sift-related logic possibly?

bor0: Yeah, we are already looking into it in parallel [here](https://github.com/Automattic/woocommerce.com/pull/23963), but figured we could try `vip-go-mu-plugins` at the same time ;)

bor0:
Closing for now as they'll be looking into https://linear.app/a8c/issue/SHILL-844/ first.

PR #6398, opened 2025-07-10
Fix fatal error during WordPress multisite activation by deferring VIP Support initialization

luismulinari:
## Description

The VIP Support plugin was causing fatal errors during WordPress multisite activation when accessing `/wp-activate.php` route:

```
PHP Fatal error: Uncaught Error: Class "Automattic\VIP\Support_User\User" not found in /wordpress/mu-plugins/telemetry/tracks/tracks-utils.php:22
```

The error occurred because during WordPress multisite activation, the constant `WP_INSTALLING` is `true`, which was preventing the `Automattic\VIP\Support_User\*` classes from being loaded in `vip-support.php`, making them unavailable for downstream code.

This PR fixes the issue by always loading the support user classes/files and deferring the initialization of VIP Support classes when `WP_INSTALLING` is true.

Changes made:
* Removed the `WP_INSTALLING` check from `vip-support.php` to ensure classes are always loaded;
* Added `WP_INSTALLING` guards around Role::init() and User::init() to prevent premature initialization during WordPress installation.

### DIFF
--- a/vip-support.php
+++ b/vip-support.php
@@ -7,6 +7,4 @@
  * License:      GPLv2 or later
  */
 
-if ( ! defined( 'WP_INSTALLING' ) ) {
-	require_once __DIR__ . '/vip-support/vip-support.php';
-}
+require_once __DIR__ . '/vip-support/vip-support.php';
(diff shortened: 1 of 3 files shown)


maxschmeling (approved):

chriszarate (approved):
Thanks!

rebeccahum (approved):

PR #6424, opened 2025-07-23
Adds an internal filter to better detect when sites are overriding 2FA rules

andrea-sdl:
## Description

The current `wpcom_vip_is_two_factor_forced` is used both by customers and also in our codebase. 

This makes it difficult to detect a site is using and leveraging this filter. This PR adds an extra filter (`wpcom_vip_internal_is_two_factor_forced`) outside of `wpcom_vip_is_two_factor_forced`, that's only used internally and is not meant for "public" usage.

Similarly, we changed Jetpack connection pilot to ensure it uses this new internal filter instead of the old one. 

Note: There is an extra usage of the`wpcom_vip_is_two_factor_forced` filter inside the codebase that we intentionally did not change because that would override the customer behavior if they have used it with a different priority. 

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

* Have a test site that implements https://docs.wpvip.com/security/two-factor-authentication/#h-enforce-two-factor-authentication-for-user-roles-and-capabilities as an example.
* Create a contributor/editor (since the capability required there user and check that 2FA is still required and it's working.

### DIFF
--- a/two-factor.php
+++ b/two-factor.php
@@ -130,8 +130,9 @@ function wpcom_vip_is_two_factor_forced() {
 	if ( ! wpcom_vip_should_force_two_factor() ) {
 		return false;
 	}
-
-	return apply_filters( 'wpcom_vip_is_two_factor_forced', false );
+	$is_enforced = apply_filters( 'wpcom_vip_is_two_factor_forced', false );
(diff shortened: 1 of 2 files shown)


brunobasto (approved):
Looks good! 🚀

sjinks (approved):

PR #6467, opened 2025-08-18
Add filter for default max image width

seanlanglands:
## Description

This pull request introduces a filter to make the default maximum width for image resizing configurable in the `image_resize` method, which originally filter's WordPress core's `image_downsize` filter. This allows users to override the default value of 1024 pixels via the `vip_go_default_image_max_width` filter.

Here's a niche use case where one would want to change this default value:

WordPress [intentionally omits generating a transformed image](https://github.com/WordPress/WordPress/blob/f63d38c15107276aaa97cf293508e0db12044538/wp-includes/media.php#L661) if the image's full size is the same as a custom image size. Separate transformed images are not generated on the VIP File System, but this WordPress core behavior does affect how registered custom image sizes are made available.

Say we create a new image size (e.g., `add_image_size( 'header-bg-1920', 1920, 1920)`), upload an image with the exact width (`1920`), and then choose to use size `full` — instead of the `header-bg-1920` size. The various helper functions (e.g., `wp_get_attachment_image_src`, `wp_get_attachment_image`, etc) will use the set `1024` value for various HTML output attributues.

Programmatically, you could pass `header-bg-1920` in your calling functions, but if there is code in a 3rd party plugin that generates size options via attachment meta, the `header-bg-1920` size won't be available to select, as WordPress core excludes it because of [wp_image_resize_identical_dimensions](https://github.com/WordPress/WordPress/blob/f63d38c15107276aaa97cf293508e0db12044538/wp-includes/media.php#L661).

The `image_resize` filter does take into account the ["content_width" if defined](https://github.com/Automattic/vip-go-mu-plugins/blob/develop/a8c-files.php#L289), but setting a custom [`content_width`](https://codex.wordpress.org/Content_Width) shouldn't be a solution here, as "full width" header images wouldn't be analogous to the desired main content width.

### Added
- Added filter to make the default maximum width for image resizing filterable.

## Pre-review checklist

## Pre-deploy checklist


a8c-files.php
```diff
@@ -289,7 +289,14 @@ public function image_resize( $ignore, $id, $size ) {
...
 		} else {
-			$_max_w = 1024;
+			/**
+			 * Filter the default maximum width for image resizing when no other constraints are available.
+			 *
+			 * @param int $max_width Default maximum width. Default 1024.
+			 * @param int $id Attachment ID for image.
+			 * @param array|string $size Size of image, either array or string.
+			 */
+			$_max_w = apply_filters( 'vip_go_default_image_max_width', 1024, $id, $size );
```
rebeccahum, line 299: Let's rename this since we're moving away from all vip go references

rebeccahum (changes requested):
I'm wondering if we should also filter out height if we're allowing width to be filtered out. Note - probably better to do it in a single filter too

PR #6217, opened 2025-03-28
add remote-data-blocks integration loader

shekharnwagh:
## Description

This PR is meant to add Remote Data Blocks as a VIP Integration, that can be activated in the same way as VIP Block Data API. The only difference is that, it has configs that can be provided as well. This has been specified, along with the necessary safety checks.

Please note that there is a corresponding PR with this to update the latest version of the plugin to 0.11.

https://github.com/Automattic/vip-go-mu-plugins-ext/pull/68

### Added
- Add Remote Data Blocks as a new VIP integration

## Pre-review checklist

Please make sure the items below have been covered before requesting a review:

- [x] This change works and has been tested locally or in Codespaces (or has an appropriate fallback).
- [x] This change has relevant unit tests (if applicable).
- [x] This change has relevant documentation additions / updates (if applicable).
- [x] I've created a changelog description that aligns with the provided examples.

## Pre-deploy checklist

- [x] VIP staff: Ensure any alerts added/updated conform to internal standards (see internal documentation). 

## Steps to Test

- Clone [RDB](https://github.com/Automattic/remote-data-blocks)
- Activate Remote Data Blocks
- Spin up a wp environment
- Ensure RDB is available in this plugin

PR #6478, opened 2025-08-21
[VIP-1715] Add real-time collaboration integration loader

shekharnwagh:
## Description

Implements `RealTimeCollaborationIntegration` class that loads custom build of `gutenberg` along with latest version of `vip-real-time-collaboration` plugin. Configures the `VIP_RTC_WS_AUTH_SECRET` and `VIP_RTC_WS_URL` constants from the `web_socket_auth_secret` and `web_socket_url` integration config values.

The custom build of gutenberg is built from https://github.com/Automattic/gutenberg/tree/add/experimental-collaborative-editing. It has the code required to support collaborative editing feature which is enabled by `vip-real-time-collaboration` plugin. This is temporary solution until the code from that branch makes it way to Gutenberg release.

It currently assumes only one version of the gutenberg custom build is available in `/vip-integrations/gutenberg` dir unlike the vip-real-time-collaboration plugin which can have multiple versions in `/vip-integrations/vip-real-time-collaboration-<version>/` dirs.

### Added
- Add support for Real-Time Collaboration integration.

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

1. Check out PR.
2. Add built code of gutenberg to `/vip-integrations/gutenberg` dir.
3. Add built code of `vip-real-time-collaboration` to `/vip-integrations/vip-real-time-collaboration-<version>/` dir.
4. Run `composer install` and `npm install` to install dependencies.
5. Activate the 'Real-Time Collaboration' integration on one of your test sites and note down its application ID.
6. Spin up a local dev for that site using command - `vip dev-env create @<application_id>.production -s <slug> -a <path_to_app_code> -u <path_to_local_vip-go-mu-plugins_dir>`
7. Start it using the command - `vip dev-env start --slug <slug> --editor vscode`
8. That should open a VS Code window. Modify the `<app_name>/integrations-config/integrations.json` to add following entry -

	```json
	"real-time-collaboration": {
		"type": "real-time-collaboration",
		"org": {
			"status": "enabled"
		},
		"env": {
			"status": "enabled",
			"config": {
				"web_socket_auth_secret": "test-secret-key",
				"web_socket_url": "wss://test.example.com/_ws"
			}
		}
	}
	```
9. Spin up the websocket server by running `npm run dev` in the `websocket-server` directory of `vip-real-time-collaboration` plugin local cloned directory. Make sure to set the `VIP_RTC_WS_AUTH_SECRET` env variable with same value as above.
10. Verify that the collaborative editing feature is working. 
11. Try modifying the `integrations.json` and see if the changes are reflected in the local dev.

### DIFF
--- a/integrations/real-time-collaboration.php
+++ b/integrations/real-time-collaboration.php
@@ -1,0 +1,111 @@
+<?php
+
+/**
+ * Integration: Real-Time Collaboration.
+ *
+ * @package Automattic\VIP\Integrations
+ */
+
+namespace Automattic\VIP\Integrations;
+
+/**
+ * Loads Real-Time Collaboration VIP Integration.
+ *
+ * @private
+ */
+class RealTimeCollaborationIntegration extends Integration {
+
+	public function is_loaded(): bool {
+		// Check for the existence of the plugin version constant defined in the main plugin file.
+		return defined( 'VIP_REAL_TIME_COLLABORATION__LOADED' );
+	}
+
+	/**
+	 * Loads the plugin.
+	 *
+	 * This is called after the integration is activated and configured.
+	 *
+	 * @private
+	 */
+	public function load(): void {
+		/**
+		 * Return if the integration is already loaded.
+		 *
+		 * In activate() method we do make sure to not activate the integration if its already loaded
+		 * but still adding it here as a safety measure i.e. if load() is called directly.
+		 */
+		if ( $this->is_loaded() ) {
+			return;
+		}
+
+		/**
+		 * Load the custom build of Gutenberg from vip-integrations.
+		 *
+		 * Built from - https://github.com/Automattic/gutenberg/tree/add/experimental-collaborative-editing
+		 * This has the code required to support collaborative editing.
+		 */
+		$gutenberg_path = WPVIP_MU_PLUGIN_DIR . '/vip-integrations/gutenberg/gutenberg.php';
+		if ( file_exists( $gutenberg_path ) ) {
+			require_once $gutenberg_path;
+		}
+
+		/**
+		 * Get all the entries in the path of WPVIP_MU_PLUGIN_DIR/vip-integrations/vip-real-time-collaboration-<version>/
+		 * and check what versions are available.
+		 */
+		$versions = $this->get_versions();
+
+		// if no versions are found, return early.
+		if ( empty( $versions ) ) {
+			$this->is_active = false;
+			return;
+		}
+
+		// Load the latest version of the plugin.
+		$latest_directory = array_key_first( $versions );
+		$load_path        = WPVIP_MU_PLUGIN_DIR . '/vip-integrations/' . $latest_directory . '/vip-real-time-collaboration.php';
+
+		// This check isn't strictly necessary, but better safe than sorry.
+		if ( file_exists( $load_path ) ) {
+			require_once $load_path;
+		} else {
+			$this->is_active = false;
+		}
+	}
+
+	/**
+	 * Get the available versions of Real-Time Collaboration in descending order.
+	 *
+	 * @return array<string, string> An associative array of available versions, where the key is the
+	 *                               directory name and the value is the version number. The versions
+	 *                               are sorted in descending order.
+	 */
+	public function get_versions() {
+		return get_available_versions(
+			WPVIP_MU_PLUGIN_DIR . '/vip-integrations/',
+			'vip-real-time-collaboration',
+			'vip-real-time-collaboration.php'
+		);
+	}
+
+	/**
+	 * Configure Real-Time Collaboration for VIP Platform.
+	 *
+	 * This is called after the integration is activated but before the plugin is loaded.
+	 *
+	 * @private
+	 */
+	public function configure(): void {
+		$env_config = $this->get_env_config();
+
+		// Set up WebSocket authentication secret constant
+		if ( isset( $env_config['web_socket_auth_secret'] ) && ! defined( 'VIP_RTC_WS_AUTH_SECRET' ) ) {
+			define( 'VIP_RTC_WS_AUTH_SECRET', $env_config['web_socket_auth_secret'] );
+		}
+
+		// Set up WebSocket URL constant with dummy value
+		if ( isset( $env_config['web_socket_url'] ) && ! defined( 'VIP_RTC_WS_URL' ) ) {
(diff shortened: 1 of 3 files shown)


tests/integrations/test-real-time-collaboration.php
```diff
@@ -0,0 +1,154 @@
...
+
+		$integration_mock->expects( $this->once() )
+			->method( 'is_loaded' )
+			->willReturn( true );
+
+		// If early return works, get_versions should never be called
+		$integration_mock->expects( $this->never() )
+			->method( 'get_versions' );
+
+		$integration_mock->load();
```
acicovic, line 52: We don't have an assertion here. Is this expected?

tests/integrations/test-real-time-collaboration.php
```diff
@@ -0,0 +1,154 @@
...
+
+	public function test_configure_defines_websocket_url_constant(): void {
+		/** @var MockObject|RealTimeCollaborationIntegration $integration_mock */
+		$integration_mock = $this->getMockBuilder( RealTimeCollaborationIntegration::class )
+			->setConstructorArgs( [ $this->slug ] )
+			->onlyMethods( [ 'get_env_config' ] )
+			->getMock();
+
+		$integration_mock->method( 'get_env_config' )->willReturn( [
+			'web_socket_url' => 'wss://test.example.com/ws',
```
acicovic: Unimportant, but if we want the `_` in `/ws`, let's add it.

acicovic (approved):

vip-integrations.php
```diff
@@ -34,6 +34,10 @@
 	require_once __DIR__ . '/integrations/jetpack.php';
 }
 
+if ( file_exists( __DIR__ . '/integrations/real-time-collaboration.php' ) ) {
```
maxschmeling, line 37: Why do some integrations have an `if ( file_exists )` check?

integrations/real-time-collaboration.php
```diff
@@ -0,0 +1,111 @@
...
+
+		/**
+		 * Load the custom build of Gutenberg from vip-integrations.
+		 *
+		 * Built from - https://github.com/Automattic/gutenberg/tree/add/experimental-collaborative-editing
+		 * This has the code required to support collaborative editing.
+		 */
+		$gutenberg_path = WPVIP_MU_PLUGIN_DIR . '/vip-integrations/gutenberg/gutenberg.php';
+		if ( file_exists( $gutenberg_path ) ) {
+			require_once $gutenberg_path;
```
maxschmeling: Is this the right way to load the plugin over something like [`activate_plugin`](https://developer.wordpress.org/reference/functions/activate_plugin/)?

maxschmeling (approved):

vip-integrations.php
shekharnwagh: It came from @rinatkhaziev 's feedback on previous PR where we added RDB - https://github.com/Automattic/vip-go-mu-plugins/pull/6217#discussion_r2027614042

integrations/real-time-collaboration.php
shekharnwagh: I am not entirely sure, but all the integrations are activated this way. My guess (I might be wrong) is, we don't want to expose this as plugins in WP-Admin. @rinatkhaziev Do you the answer here ?

ingeniumed: If this ensures that we don't expose Gutenberg in WP-Admin, I wonder if we might run into conflicts with those sites that have Gutenberg already enabled and want to use our plugin?

integrations/real-time-collaboration.php
```diff
@@ -0,0 +1,111 @@
...
+		}
+
+		/**
+		 * Get all the entries in the path of WPVIP_MU_PLUGIN_DIR/vip-integrations/vip-real-time-collaboration-<version>/
+		 * and check what versions are available.
+		 */
+		$versions = $this->get_versions();
```
ingeniumed, line 56: We should consider pulling this out into a utility, as the RDB plugin uses this as well.

integrations/real-time-collaboration.php
```diff
@@ -0,0 +1,111 @@
...
+	 *
+	 * This is called after the integration is activated but before the plugin is loaded.
+	 *
+	 * @private
+	 */
+	public function configure(): void {
+		$env_config = $this->get_env_config();
+
+		// Set up WebSocket authentication secret constant
+		if ( isset( $env_config['web_socket_auth_secret'] ) && ! defined( 'VIP_RTC_WS_AUTH_SECRET' ) ) {
```
ingeniumed, line 137: I'm wondering if we should add in a check to make sure our plugin doesn't load if these aren't defined? I know we have a check already in our code, but this might be a good guard to have in place as well.

vip-integrations.php
rinatkhaziev: @maxschmeling because deploys are not-atomic and not properly gating this produces intermittent fatal errors.

integrations/real-time-collaboration.php
rinatkhaziev: It would and this needs to be properly guarded against. Easy to test though :)

integrations/real-time-collaboration.php
shekharnwagh: Pushed a commit to add `can_load` function which does such checks.

integrations/real-time-collaboration.php
shekharnwagh: Added logic to avoid loading the integration if Gutenberg is activated as a plugin - 7d1bf5fdd

integrations/real-time-collaboration.php
```diff
@@ -0,0 +1,146 @@
...
+			define( 'VIP_RTC_WS_AUTH_SECRET', $env_config['web_socket_auth_secret'] );
+		}
+
+		// Set up WebSocket URL constant with dummy value
```
ingeniumed: Minor: is it a dummy value or the actual value?

ingeniumed (approved):

integrations/real-time-collaboration.php
shekharnwagh: Its actual value. It was using a dummy value as when the PR was opened the config didn't pass the WebSocket URL through. Fixed the comment.

shekharnwagh: Thank you for catching this !

rinatkhaziev (approved):

PR #6545, opened 2025-10-01
[VIP-1939] Display integrations in plugins list in WP-Admin

shekharnwagh:
## Description

This PR adds functionality to display plugins loaded by VIP Integrations in the WordPress admin plugins list, providing better visibility.

Plugins loaded by VIP Integrations are loaded via `require_once` from the `mu-plugins/vip-integrations/` directory. Since WordPress's `get_plugins()` only scans the standard `wp-content/plugins/` directory, these integration plugins are invisible in the wp-admin plugins list.

To solve this, created a reusable `IntegrationPluginDisplayTrait` that:
- Injects integration plugins into WordPress's plugins list via the `all_plugins` filter
- Shows plugins as active with "Enabled via WPVIP Integrations" status
- Removes activate/deactivate links (like code-activated plugins)
- Prevents accidental database writes for integration plugins
- Suppresses "Plugin file does not exist" error messages using output buffering

This approach is similar to how `wpcom_vip_load_plugin()` displays code-activated plugins, but tweaked for the plugin code being in `mu-plugins/vip-integrations` dir.

Screenshot 2025-10-03 at 12 17 33 AM

### Added
- VIP Integrations: Added a reusable `IntegrationPluginDisplayTrait` that displays integration plugins in the WordPress admin plugins list
- VIP Integrations: When Real-Time Collaboration integration is loaded, the custom build of Gutenberg and the latest version of the Real-Time Collaboration plugin activated via VIP Integrations are displayed in the WordPress admin plugins list

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

1. Check out PR.
2. Copy the `/vip-integrations` directory from the [vip-go-mu-plugins-built](https://github.com/Automattic/vip-go-mu-plugins-built) repo to `/vip-integrations` directory of this repo.
3. Run `composer install` and `npm install` to install dependencies.
4. Activate the 'Real-Time Collaboration' integration on one of your test sites and note down its application ID.
5. Spin up a local dev for that site using command - `vip dev-env create @<application_id>.production -s <slug> -a <path_to_app_code> -u <path_to_local_vip-go-mu-plugins_dir>`
6. Start it using the command - `vip dev-env start --slug <slug> --editor vscode`
7. That should open a VS Code window. Modify the `<app_name>/integrations-config/integrations.json` to add following entry -

	```json
	"real-time-collaboration": {
		"type": "real-time-collaboration",
		"org": {
			"status": "enabled"
		},
		"env": {
			"status": "enabled",
			"config": {
				"web_socket_auth_secret": "test-secret-key",
				"web_socket_url": "wss://test.example.com/_ws"
			}
		}
	}
	```
8. Navigate to the `wp-admin` > `Plugins` page and verify that the custom build of Gutenberg and the latest version of the Real-Time Collaboration plugin activated via VIP Integrations are displayed in the WordPress admin plugins list.


integrations/integration-plugin-display-trait.php
```diff
@@ -0,0 +1,386 @@
...
+	 * @var bool
+	 */
+	private static $buffer_started = false;
+
+	/**
+	 * Custom message to display for integration plugins.
+	 *
+	 * @var string
+	 */
+	private static $plugin_display_message = 'Enabled via WPVIP Integrations';
```
pkevan, line 80: is there something more concise we can use here - feels a little long, particularly when it spans two lines.

shekharnwagh: Open to ideas. Worded it so it's similar to badge - `Enabled via code` shown when we [activate plugin from code](https://docs.wpvip.com/plugins/activate-plugins-through-code/) using `wpcom_vip_load_plugin`.

What do we think using just `WPVIP Integration` ? But that might be slightly confusing when that shows up for `Gutenberg` plugin.

pkevan: Potentially something as simple as a link to the integrations center via a link `Settings`?

pkevan: Screenshot 2025-10-10 at 09 09 46

shekharnwagh: Just took a quick look at the code in this repo and I couldn't find any references to `dashboard.wpvip.com`. If we have add it, it'll be some work constructing the url and maintaining it in unlikely scenario the routing in dashboard changes.

pkevan: perhaps we get away with a link to the docs? like: https://docs.wpvip.com/integrations/

ingeniumed: How about `Enabled via integrations` and integrations links to https://docs.wpvip.com/integrations/?

rinatkhaziev (changes requested):
So using regex and ob doesn't seem entirely safe. There are better ways, for example:

https://developer.wordpress.org/reference/hooks/plugin_action_links/

integrations/integration-plugin-display-trait.php
```diff
@@ -0,0 +1,386 @@
...
+		add_filter( 'option_active_plugins', [ $this, 'filter_active_plugins_for_display' ] );
+		add_filter( 'site_option_active_sitewide_plugins', [ $this, 'filter_network_active_plugins_for_display' ] );
+
+		// Prevent accidental database writes
+		add_filter( 'pre_update_option_active_plugins', [ $this, 'filter_update_active_plugins' ] );
+		add_filter( 'pre_update_site_option_active_sitewide_plugins', [ $this, 'filter_update_network_active_plugins' ] );
+
+		// Cleanup and error suppression
+		add_action( 'admin_init', [ $this, 'cleanup_recently_activated' ], 1 );
+		add_action( 'admin_notices', [ $this, 'suppress_deactivation_notices' ], 1 );
```
pkevan, line 133: curious as to when these get triggered, as it looks like they would only appear when the plugin activation/deactivation links are present and a user clicks on them?

integrations/integration-plugin-display-trait.php
maxschmeling: I wouldn't be against adding a link or something better down the road, but the simple but clear text that you already have in there is fine with me even though it splits to two lines.

integrations/integration-plugin-display-trait.php
```diff
@@ -0,0 +1,386 @@
...
+		$integration_plugins = static::get_loaded_integration_plugins();
+
+		if ( empty( $integration_plugins ) || static::$buffer_started ) {
+			return;
+		}
+
+		static::$buffer_started = true;
+
+		// Use output buffering to filter out error notices about integration plugins
+		ob_start( function ( $buffer ) use ( $integration_plugins ) {
```
maxschmeling:
```suggestion
		// Use output buffering to filter out error notices about integration plugins
		static::$buffer_started = ob_start( function ( $buffer ) use ( $integration_plugins ) {
```

maxschmeling (approved):

integrations/integration-plugin-display-trait.php
shekharnwagh: Thinking through this again, yeah we won't need this since we've already removed the links to activate/deactivate plugin and the activation happens via code. Removed in a38323ebf

shekharnwagh:
> So using regex and ob doesn't seem entirely safe. There are better ways, for example:
> 
> https://developer.wordpress.org/reference/hooks/plugin_action_links/

The `plugin_action_links` only customizes the action buttons in each plugin row. The "Plugin file does not exist" error is generated during WordPress's plugin validation. We are already hooking into that filer to remove the activation and deactivation links and adding the `Enabled via WPVIP Integrations` message.
That's why I ended up using output buffering to suppress the error notice.

I'm open to better solutions. Is there a specific WordPress hook or filter that prevents the validation check for plugins in non-standard directories? @rinatkhaziev

rinatkhaziev:
> I'm open to better solutions. Is there a specific WordPress hook or filter that prevents the validation check for plugins in non-standard directories? @rinatkhaziev

OIC, I misunderstood the diff.

Okay, in this particular case simply removing `filter_active_plugins_for_display` gets rid of the notice.

```diff
diff --git a/integrations/integration-plugin-display-trait.php b/integrations/integration-plugin-display-trait.php
index 9376cfb55..ce872a255 100644
--- a/integrations/integration-plugin-display-trait.php
+++ b/integrations/integration-plugin-display-trait.php
@@ -122,7 +122,7 @@ trait IntegrationPluginDisplayTrait {
 		add_filter( 'network_admin_plugin_action_links', [ $this, 'filter_plugin_action_links' ], 10, 2 );
 
 		// Active plugin management (display only)
-		add_filter( 'option_active_plugins', [ $this, 'filter_active_plugins_for_display' ] );
+		// add_filter( 'option_active_plugins', [ $this, 'filter_active_plugins_for_display' ] );
 		add_filter( 'site_option_active_sitewide_plugins', [ $this, 'filter_network_active_plugins_for_display' ] );
 
 		// Prevent accidental database writes
@@ -130,8 +130,8 @@ trait IntegrationPluginDisplayTrait {
 		add_filter( 'pre_update_site_option_active_sitewide_plugins', [ $this, 'filter_update_network_active_plugins' ] );
 
 		// Cleanup and error suppression
-		add_action( 'admin_notices', [ $this, 'suppress_deactivation_notices' ], 1 );
-		add_action( 'network_admin_notices', [ $this, 'suppress_deactivation_notices' ], 1 );
+		// add_action( 'admin_notices', [ $this, 'suppress_deactivation_notices' ], 1 );
+		// add_action( 'network_admin_notices', [ $this, 'suppress_deactivation_notices' ], 1 );
 	}
 
 	/**

```
CleanShot 2025-10-24 at 18 13 53@2x

shekharnwagh:
> Okay, in this particular case simply removing filter_active_plugins_for_display gets rid of the notice.

Removing `filter_active_plugins_for_display` means they now show up in inactive plugins list which isn't what we'd want. We do have a text saying "Enabled via WPVIP Integrations", but they not being in active list might indicate they aren't loaded.

Screenshot 2025-10-25 at 10 40 00 AM

shekharnwagh:
> > Okay, in this particular case simply removing filter_active_plugins_for_display gets rid of the notice.
> 
> Removing `filter_active_plugins_for_display` means they now show up in inactive plugins list which isn't what we'd want. We do have a text saying "Enabled via WPVIP Integrations", but they not being in active list might indicate they aren't loaded.
> Screenshot 2025-10-25 at 10 40 00 AM

Bringing this on your radar in case it slipped through - @rinatkhaziev

rinatkhaziev:
Still very much on the radar, I asked @sjinks to look at it to provide ideas/suggestions. @shekharnwagh

sjinks:
I am exploring how to (ab)use the `plugins_list` filter: https://github.com/WordPress/WordPress/blob/master/wp-admin/includes/class-wp-plugins-list-table.php#L309

The idea is that we can add the "Integrations" tab to the plugins list and display our integrations there.

I hope that this approach is less fragile and does not require changes to integrations' code.

shekharnwagh:
> I am exploring how to (ab)use the `plugins_list` filter: https://github.com/WordPress/WordPress/blob/master/wp-admin/includes/class-wp-plugins-list-table.php#L309
> 
> The idea is that we can add the "Integrations" tab to the plugins list and display our integrations there.
> 
> I hope that this approach is less fragile and does not require changes to integrations' code.

@sjinks did your exploration of the `plugins_list` filter yield any results?

PR #6585, opened 2025-10-22
adding better publish tracking

smithjw1:
## Description

This PR moves publish telemetry from a simple counter approach to a more detailed approach with a few key metrics. The new approach is composed of three functions.

* `record_created_time` - is designed to run once per post, when it is first explicitly saved to record the creation time.
* `record_published_time`- is designed to run once per post, when it is first published, to record the publish time, and to schedule an asynchronous telemetry call.
* `send_publish_telemetry` - is designed to send the telemetry data. It is called asynchronously for performance and usability purposes.

This data will allow us to answer questions like:

* How long does it take to publish on VIP?
* How many posts are published a day?
* Does using real-time collaboration change the average time to publish?

### Changed
- More nuanced telemetry around post-publish events.

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

1. Check out PR.
1. Go to `wp-admin`
1. Create a new post
1. Verify that `_vip_edtel_created_time` and `_vip_edtel_published_time` meta are not set
1. Save the post
2. Verify that `_vip_edtel_created_time` meta is set and the value is sensible, and `_vip_edtel_published_time` is not set
3. Make more changes and save the post
4. Verify that `_vip_edtel_created_time` is still set and the value has not changed
5. Publish the post
6. Verify that `_vip_edtel_published_time` is set, and that the value is sensible
7. Verify that `_vip_edtel_created_time` is still set and the value has not changed
8. Make changes and click publish again
9. Verify that `_vip_edtel_published_time` is still set and the value has not changed
10. Verify that `_vip_edtel_created_time` is still set and the value has not changed


smithjw1:
@chriszarate I requested your review because we've worked together on other telemetry pieces. I've added specific questions/comments in the diff.

stats.php
```diff
@@ -27,20 +34,95 @@
...
+
+	// Add published time (only if not already set due to $unique = true)
+	add_post_meta( $post->ID, '_vip_edtel_published_time', gmdate( 'c' ), true );
+
+	// Defensive check for VIP Telemetry class
+	if ( ! class_exists( 'Automattic\\VIP\\Telemetry\\Tracks' ) ) {
+		return;
+	}
+
+	wp_schedule_single_event( ( time() + 30 ), 'viptel_send_publish_telemetry', array( $post->ID ) );
```
smithjw1: I wanted to pass this of to Cron so that we don't slow down the publish event. 

I don't really care about a delay of even 24 hours, we don't need these stats in anything close to real time, but I think we want them coming in randomly, and publish times aren't always random, since customers schedule posts.

I'm open to any thoughts about a sensible delay so we don't overload cron.

stats.php
```diff
@@ -27,20 +34,95 @@
...
+}
+
+function send_publish_telemetry( $post_id ) {
+	$post = get_post( $post_id );
+	if ( ! $post ) {
+		return;
+	}
+
+	// Get RTC state
+	$rtc_state = get_post_meta( $post_id, '_vip_rtc_state', true );
```
smithjw1, line 97: AI was trying to get me to use `get_post_meta( $post_id, '', true)` and then filter the results to only do one DB call.

My gut is that for most of our customers, getting all the meta will be slower, but I'm open to other ideas.

stats.php
```diff
@@ -27,20 +34,95 @@
...
+
+	// Prepare event properties
+	$properties = array(
+		'post_type'  => sanitize_text_field( $post->post_type ),
+		'word_count' => absint( str_word_count( wp_strip_all_tags( $post->post_content ) ) ),
```
smithjw1, line 102: I know this is hefty, and I'm open to other ideas. Since posts can be just about anything, I want a way to filter out posts that don't have content.

I don't know where I would set this without seeing data. In the future, we may be able to put a filter here to not send events.

stats.php
```diff
@@ -27,20 +34,95 @@
...
+		}
+	}
+	$properties['created_to_published_seconds'] = $created_to_published_seconds;
+
+	// Send telemetry event
+	try {
+		$tracks = new Tracks( 'vip' );
+		$tracks::record_event( 'post_published', $properties );
+	} catch ( Exception $e ) { // phpcs:ignore Generic.CodeAnalysis.EmptyStatement.DetectedCatch
+		// fail silently because we don't want to fill up error logs if our internal tracking fails
```
smithjw1, line 126: Open to ideas here, maybe the try/catch block doesn't even need to exist.

stats.php
```diff
@@ -15,7 +15,14 @@
 
 // Limit tracking to production
 if ( true === WPCOM_IS_VIP_ENV && false === WPCOM_SANDBOXED ) {
-	add_action( 'transition_post_status', __NAMESPACE__ . '\track_publish_post', 9999, 2 );
+	
+	// Record created time of a post
+	add_action( 'save_post', __NAMESPACE__ . '\record_created_time', 9999, 3 );
```
chriszarate, line 20: I don't think you need this since `transition_post_status` fires on initial save also. Can just check if post meta exists and set it if it doesn't. This can also help catch edge cases like imported or syndicated content?

stats.php
```diff
@@ -27,20 +34,101 @@
...
 		return;
 	}
 
+	// Skip revisions
+	if ( wp_is_post_revision( $post ) ) {
+		return;
+	}
+
+	// Add published time (only if not already set due to $unique = true)
+	add_post_meta( $post->ID, '_vip_edtel_published_time', gmdate( 'c' ), true );
```
chriszarate: Will anyone object to this post meta being set? Are there other VIP-specific post meta?

stats.php
chriszarate: It's not that expensive.

chriszarate:
@Automattic/vip-platform-cantina Any ideas on alternative ways to capture the original creation date for telemetry? We could look at revisions, but they are capped by some sites. Prom metrics could implement a rolling average but may not answer every product question, especially as we work to collect additional data.

stats.php
rebeccahum: > Can just check if post meta exists and set it if it doesn't.

The third parameter of `add_post_meta()` already does that 👍

stats.php
rebeccahum: Yup, better to specifically retrieve the value via key since some customers have posts that have :alot: of meta.

stats.php
rebeccahum: @chriszarate music to my ears 😂

stats.php
rebeccahum: Yeah, we have some sprinkled around in MU....as for objections, probably not since it's not like we are storing a large serialized value?

Note: For this one specifically though, we don't need it I think b/c can use the `post_date` property

rebeccahum:
Any reason why we can't just use the existing `post_date` property of the post and skip the whole `_vip_edtel_published_time` adding meta action? AFAIK, it usually doesn't get modified on save.

stats.php
```diff
@@ -15,7 +15,14 @@
 
 // Limit tracking to production
 if ( true === WPCOM_IS_VIP_ENV && false === WPCOM_SANDBOXED ) {
-	add_action( 'transition_post_status', __NAMESPACE__ . '\track_publish_post', 9999, 2 );
+	
+	// Record created time of a post
+	add_action( 'save_post', __NAMESPACE__ . '\record_created_time', 9999, 3 );
+	
+	// Record published time and send telemetry when post is published
+	add_action( 'transition_post_status', __NAMESPACE__ . '\record_published_time', 9999, 2 );
```
rebeccahum:
```suggestion
	add_action( 'transition_post_status', __NAMESPACE__ . '\record_published_time', 9999, 3 );
```
since it's taking 3 params

stats.php
rebeccahum: Yeah, I think it might be unnecessary

stats.php
```diff
@@ -27,20 +34,101 @@
...
+		if ( $created_timestamp && $published_timestamp ) {
+			$created_to_published_seconds = max( 0, $published_timestamp - $created_timestamp );
+		}
+	}
+	$properties['created_to_published_seconds'] = $created_to_published_seconds;
+
+	// Send telemetry event
+	try {
+		$tracks = new Tracks( 'vip' );
+		$tracks::record_event( 'post_published', $properties );
```
rebeccahum:
```suggestion
		$tracks->record_event( 'post_published', $properties );
```

smithjw1:
Thanks for the review @rebeccahum! I've made those changes

My next step is to test this to see what sort of latency we are adding to the publishing process.

PR #6607, opened 2025-11-06
adding integration gating to Pendo

smithjw1:
## Description

We want to restrict the inclusion of our telemetry JavaScript agent to customers using specific integrations that use front-end tracking.

This change adds that restriction to the existing logic.

### Changed
- Added additional checks before the inclusion of an admin-facing telemetry library.

## Pre-review checklist

- [N/A] This change works and has been tested locally or in Codespaces (or has an appropriate fallback).
- [N/A] This change works and has been tested on a sandbox.
- [N/A] This change has relevant unit tests (if applicable).
- [N/A] This change has relevant documentation additions / updates (if applicable).

## Pre-deploy checklist

## Steps to Test

We disable telemetry tracking on all non-production instances and sandboxes, so testing locally is complicated.

1. Check out PR.
1. Go to `wp-admin`
1. Verify nothing is broken


telemetry/pendo/class-pendo-javascript-library.php
```diff
@@ -196,6 +202,20 @@ final public static function should_enqueue_script( string $screen ): bool {
...
+		if ( ! function_exists( 'wpvip_get_enabled_integrations' ) ) {
+			return false;
+		}
+
+		/**
+		 * Filters the list of integrations that must be enabled for Pendo JavaScript library to be loaded.
+		 *
+		 * @param array $required_integrations Array of integration slugs that must be enabled.
+		 */
+		$required_integrations = apply_filters( 'vip_pendo_required_integrations', self::$required_integrations );
```
rebeccahum: If they are required, how come we are allowing it to be filtered out?

smithjw1: The goal isnt' to allow them to be filtered out (though you could do that). The goal is to allow the next plugin we want to track to be able to opt-in on their own, without an MU plugins change by using this filter.

I'm thinking this might not be the best approach, since for an integration to exist, it means an MU plugins change.

rebeccahum: Yeah, I'm not sure if there is any way around MU plugins change, but I'm not sure how I feel about having a hardcoded list as an approach. I wonder if it's worth adding a method at the integration level?

smithjw1: Or metadata? We would only want it for our integrations.

I don't love a hardcoded list, but I was basing this off our other hardcoded list of screens to track in WordPress.

I'm happy to explore another approach.

rebeccahum: I think metadata seems like overkill.. Maybe a property at the abstract integration level itself to opt in?

smithjw1: I've a new PR, more changes overall, but the long-term flexibility is nice. 

My only question is about third-party integrations. We don't really want them enabling Pendo tracking; it isn't going to start sending data to anyone but us, but we do want to only opt in our integrations. My thought is that it's a remote possibility, and probably not something to worry about, but I'm interested in what others think.

rebeccahum (approved):

smithjw1:
Thanks @rebeccahum for the 👍 

Since this is the biggest change I've ever written to MU code, I'd love a look by @chriszarate (previously implemented Pendo tracking) and @shekharnwagh (knows Integrations stuff).

integrations/integration.php
```diff
@@ -170,6 +177,15 @@ public function set_vip_config( IntegrationVipConfig $vip_config ): void {
 		$this->vip_config = $vip_config;
 	}
 
+	/**
+	 * Returns true if this integration should be tracked in Pendo.
+	 *
+	 * @return bool
+	 */
+	public function should_track_in_pendo(): bool {
```
chriszarate: We want extending classes to set the property, not override this method

```suggestion
	final public function should_track_in_pendo(): bool {
```

integrations/parsely.php
```diff
@@ -13,6 +13,16 @@
...
 class ParselyIntegration extends Integration {
+	/**
+	 * Constructor.
+	 *
+	 * @param string $slug Slug of the integration.
+	 */
+	public function __construct( string $slug ) {
+		parent::__construct( $slug );
+		$this->enable_pendo_tracking = true;
+	}
```
chriszarate: I would suggest this method of overwriting the property. It's cleaner and easier to maintain long-term in cases where the constructor args change

```suggestion
	/**
	 * Enable Pendo tracking for this integration.
	 *
	 * @var bool
	 */
	protected bool $enable_pendo_tracking = true;
```

integrations/real-time-collaboration.php
```diff
@@ -15,6 +15,16 @@
...
 
+	/**
+	 * Constructor.
+	 *
+	 * @param string $slug Slug of the integration.
+	 */
+	public function __construct( string $slug ) {
+		parent::__construct( $slug );
+		$this->enable_pendo_tracking = true;
+	}
```
chriszarate:
```suggestion
	/**
	 * Enable Pendo tracking for this integration.
	 *
	 * @var bool
	 */
	protected bool $enable_pendo_tracking = true;
```

integrations/remote-data-blocks.php
```diff
@@ -15,6 +15,15 @@
...
 
+	/**
+	 * Constructor.
+	 *
+	 * @param string $slug Slug of the integration.
+	 */
+	public function __construct( string $slug ) {
+		parent::__construct( $slug );
+		$this->enable_pendo_tracking = true;
+	}
```
chriszarate:
```suggestion
	/**
	 * Enable Pendo tracking for this integration.
	 *
	 * @var bool
	 */
	protected bool $enable_pendo_tracking = true;
```

integrations/vip-governance.php
```diff
@@ -22,6 +22,16 @@ class VipGovernanceIntegration extends Integration {
...
 
+	/**
+	 * Constructor.
+	 *
+	 * @param string $slug Slug of the integration.
+	 */
+	public function __construct( string $slug ) {
+		parent::__construct( $slug );
+		$this->enable_pendo_tracking = true;
+	}
```
chriszarate:
```suggestion
	/**
	 * Enable Pendo tracking for this integration.
	 *
	 * @var bool
	 */
	protected bool $enable_pendo_tracking = true;
```

telemetry/pendo/class-pendo-javascript-library.php
```diff
@@ -196,6 +196,21 @@ final public static function should_enqueue_script( string $screen ): bool {
 			return false;
 		}
 
+		if ( ! function_exists( '\Automattic\VIP\Integrations\wpvip_get_enabled_integrations' ) ) {
+			return false;
+		}
+
+		$enabled_integrations    = \Automattic\VIP\Integrations\wpvip_get_enabled_integrations();
```
chriszarate: Since this function is defined in this codebase and has already been deployed, you can safely add a `use function Automattic\VIP\Integrations\wpvip_get_enabled_integrations;` statement at the top of this file and call it without a check.


```suggestion
		$enabled_integrations = wpvip_get_enabled_integrations();
```

tests/integrations/fake-integration.php
```diff
@@ -19,3 +19,19 @@ public function load(): void { }
...
+	}
+
+	public function is_loaded(): bool {
+		return false;
+	}
+
+	public function load(): void { }
+
+	public function configure(): void { }
+}
```
chriszarate:
```suggestion
class FakeIntegrationWithPendoTracking extends FakeIntegration {
	protected bool $enable_pendo_tracking = true;
}
```

tests/integrations/test-integration.php
```diff
@@ -203,4 +203,16 @@ public function test__get_child_env_configs_delegates_to_vip_config_when_present
...
 		$this->assertEquals( $child_env_configs, $integration->get_child_env_configs() );
 	}
+
+	public function test__should_track_in_pendo_returns_false_by_default(): void {
+		$integration = new FakeIntegration( 'fake' );
+
+		$this->assertFalse( $integration->should_track_in_pendo() );
+	}
+
+	public function test__should_track_in_pendo_can_be_set_to_true_in_constructor(): void {
```
chriszarate:
```suggestion
	public function test__should_track_in_pendo_returns_true_when_enabled(): void {
```

chriszarate (changes requested):
Flagged some nitpicky PHP stuff. I think you unintentionally committed an update to the wp-paresely submodule.

telemetry/pendo/class-pendo-javascript-library.php
smithjw1: So I understand this, these checks are necessary so that the first time new functionality is deployed, it doesn't break, but then can be safely removed.

I thought the same situation could happen again on re-deploys where this file "lands" first and needs a check.

chriszarate: > I thought the same situation could happen again on re-deploys where this file "lands" first and needs a check.

A file can exist on the file system and cause no harm. It's the `require` call—and specifically the load order—that matters. Looking at the file that requires both of the relevant files, there's not a situation that could load this file without already having the `wpvip_get_enabled_integrations` present and defined.

smithjw1:
@chriszarate I've made all the suggested changes.

@shekharnwagh, I'd still like a look at the integrations changes to make sure I'm not introducing anything odd.

integrations/vip-governance.php
```diff
@@ -14,6 +14,21 @@
  */
 class VipGovernanceIntegration extends Integration {
 
+	/**
+	 * The version of the VIP Governance plugin to load, that's set to the latest version.
+	 * This should be higher than the lowestVersion set in "vip-governance" config (https://github.com/Automattic/vip-go-mu-plugins-ext/blob/trunk/config.json)
+	 *
+	 * @var string
+	 */
+	protected string $version = '1.0';
```
shekharnwagh: This was removed in #6689 recently, so updating the branch should fix this

shekharnwagh (approved):
Apologies for delay - this looks good 🚢

chriszarate (approved):
great work

PR #6613, opened 2025-11-12
[PLTFRM-1839] Search: Prevent parse_query hooks from corrupting CLI `wp vip-search health validate-counts`

rebeccahum:
## Description
When running `wp vip-search health validate-counts`, some third-party plugins hook into `parse_query` and mutate
the `orderby` clause, adding unsupported keys. This breaks things in `get_index_entity_count_from_elastic_search()` when we call `query_objects()` which calls the action `parse_query` in `new WP_Query()`. We should disallow those potential `query_vars` modifications by temporarily unhooking `parse_query` actions since `get_index_entity_count_from_elastic_search` is only called in a CLI context from the `validate-counts` command.

### Fixed
-  Enterprise Search: prevent parse_query hooks from corrupting CLI `wp vip-search health validate-counts`

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

1) Add the below snippet to break things in `plugin-loader.php`:
```
add_action( 'parse_query', function ( WP_Query $q ) {
	$orderby = (array) $q->get( 'orderby' );

	// Inject an unsupported key.
	$orderby['blabla'] = 'ASC';

	$q->set( 'orderby', $orderby );
}, 50 );
```
2) Run `wp vip-search health validate-counts` and see:
```
Validating post count

Warning: Error while validating count: failure querying ES. #vip-search
Warning: Error while validating count: failure querying ES. #vip-search
```

### DIFF
--- a/search/includes/classes/class-health.php
+++ b/search/includes/classes/class-health.php
@@ -176,8 +176,10 @@ class Health {
 				add_filter( 'ep_exclude_password_protected_from_search', '__return_false' );
 			}
 
-			// Avoid modifying query_vars to avoid breaking `wp vip-search health validate-counts` CLI.
+			// Ensure query_vars stay un-modified because some third-party plugins modify it 
+			// (e.g. orderby) and break `wp health validate-counts` CLI.
 			remove_all_actions( 'parse_query' );
+
 			$query = self::query_objects( $query_args, $indexable->slug );
 			if ( 'user' === $indexable->slug && isset( $query->query_vars['blog_id'] ) ) {
 				// Since the user indexable is global, we want to include ALL in count.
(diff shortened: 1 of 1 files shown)


rebeccahum:
Since this specific use case, it's `orderby`, we could potentially do an `unset( $query->query_vars['orderby'] );` but I figured that since it can break in other ways, i.e.:
```
add_action( 'parse_query', function( WP_Query $q ) {
    $meta_query = (array) $q->get( 'meta_query' );
    $meta_query[] = [
        'key'   => '_featured',
        'value' => '1',
    ];
    $q->set( 'meta_query', $meta_query );
});
```
Probably best to just strip it all out to prevent `query_vars` from being modified unexpectedly.

rinatkhaziev:
Umm I fear it's a bit overreaching, but I have no proof. 

image

rebeccahum:
If it makes you feel better, we already strip out these filters later on: https://github.com/Automattic/vip-go-mu-plugins/blob/594ddd92a9c2c1f2099752e47ec8e64083781bdd/search/includes/classes/class-health.php#L186-L187

rinatkhaziev (approved):

PR #6665, opened 2025-12-17
ci: update actions

sjinks:
Update all actions to their latest versions in bulk so we don't have to wait forever for CI jobs to complete.


sjinks:
Docker is dead... nice.

rebeccahum (approved):

PR #6689, opened 2026-01-07
Switch the RDB, RTC, Block Data API and Governance plugins to use the latest versions

ingeniumed:
## Description

Just like https://github.com/Automattic/vip-go-mu-plugins/pull/6246, this'll allow the VIP Block Data API and the VIP Governance plugins to use the latest available integration rather than hardcoding the version to use. This'd mean one less place to update should the major/minor version of the plugin change, rather than the patch version.

### Changed

- Dynamically load the latest version of the VIP Block Data API and VIP Governance plugins

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

1. Check out the PR
2. Create a dev-env with this PR, as well as with latest mu-plugins-ext
3. Activate block data api and vip governance plugins
4. Ensure the latest version of the plugins are loaded

### DIFF
--- a/tests/integrations/test-real-time-collaboration.php
+++ b/tests/integrations/test-real-time-collaboration.php
@@ -150,11 +150,11 @@ class Real_Time_Collaboration_Integration_Test extends WP_UnitTestCase {
 		$this->assertFalse( defined( 'VIP_RTC_WS_URL' ) );
 	}
 
-	public function test_get_latest_version_returns_latest_version(): void {
+	public function test_get_latest_version_returns_null_when_no_versions_are_found(): void {
 		$rtc_integration = new RealTimeCollaborationIntegration( $this->slug );
 		$latest_version  = $rtc_integration->get_latest_version();
 
-		$this->assertIsString( $latest_version );
+		$this->assertNull( $latest_version );
 	}
 
 	public function test_load_sets_inactive_when_ws_auth_secret_missing(): void {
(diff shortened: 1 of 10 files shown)


alecgeatches (approved):
Looks good! This will make minor/major updates easier to do correctly.

ingeniumed:
The duplicated code is expected due to how the generic `get_available_versions` is to be called. Those lines have already been simplified a lot over the last year, hence why I've left it as is.

ingeniumed:
The code duplication is expected given how the latest version is loaded, after confirming that it exists or not. I've tried to further reduce this duplication by pulling out the latest version code into a utility as well. I've also added some extra tests so this should help.

The other integrations have been updated to take advantage of this too.

tests/integrations/test-real-time-collaboration.php
```diff
@@ -150,11 +150,11 @@ public function test_configure_handles_missing_config_values(): void {
 		$this->assertFalse( defined( 'VIP_RTC_WS_URL' ) );
 	}
 
-	public function test_get_latest_version_returns_latest_version(): void {
+	public function test_get_latest_version_returns_null_when_no_versions_are_found(): void {
```
ingeniumed, line 153: This used to return an empty array previously, so technically no version was being loaded. This is the correct test case now

alecgeatches (approved):
Looks great! I haven't tested locally, but the tests look good as verification.

PR #6786, opened 2026-02-23
fix: update Akismet version to match the used plugin

sjinks:
In #5321, we updated Akismet but forgot to update its version in the VIP MU plugin.

This PR does that.


sjinks:
Closing in favor of #6787

PR #6973, opened 2026-05-19
Enterprise Search: Prevent site's index version leaking into cross-site ES queries

rebeccahum:
## Description
This pull request addresses a bug in multisite environments where the search index name version could be incorrectly resolved from the calling site's context rather than the target site. The update ensures that cross-site queries use the correct version data for each site, and adds a test to verify this behavior.

### Fixed
- Enterprise Search: Prevent site's index version leaking into cross-site ES queries

## Pre-review checklist

## Pre-deploy checklist

## Steps to Test

1. On a MS, create 2 more subsites
2. On site 2, create an index version 2 and activate it
3. Add:
```
$query = new WP_Query(
	array(
		's' => 'carrot',
		'sites' => array( 2, 3 ), // Search only in site IDs 2 and 3 
		'ep_integrate' => true,
	)
);
```
4. In site 2 context, you should see in the QM API call list:
```
http://elasticsearch:9200/vip-200508-post-2-v2,vip-200508-post-3/_search
```

### DIFF
--- a/search/includes/classes/class-search.php
+++ b/search/includes/classes/class-search.php
@@ -815,7 +815,18 @@ class Search {
 			$index_name .= sprintf( '-%s', $blog_id );
 		}
 
+		// get_versions() uses get_option(), which is blog-context-sensitive. Without switching,
+		// cross-site queries resolve the version from the calling site instead of the target site.
+		$switched = $blog_id && \is_multisite() && \get_current_blog_id() !== (int) $blog_id;
+		if ( $switched ) {
+			\switch_to_blog( $blog_id );
+		}
+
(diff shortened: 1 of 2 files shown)


rinatkhaziev:
Is there a way to do this without switch_to_blog? probably no bueno for very large sites.

rebeccahum:
I'm not sure how to retrieve the blog context better w/o `switch_to_blog()` :/

rinatkhaziev:
@rebeccahum how about `get_site_option`?

rebeccahum:
@rinatkhaziev We can't do `get_site_option` because that reads from the overall network-wide options table, not the actual blog....so we'd need `get_blog_option()` which basically does the blog switch...

PR #6996, opened 2026-06-01
Add Safe Publish VIP integration loader

maxschmeling:
## Summary
- register a Safe Publish VIP integration loader
- map integration config values to Safe Publish constants
- load versioned Safe Publish artifacts from vip-integrations/safe-publish-*

## Dependencies
- Depends on the Safe Publish PR that adds constant-backed configuration support.
- Depends on vip-go-mu-plugins-ext ingesting Safe Publish release artifacts.

### DIFF
--- a/integrations/safe-publish.php
+++ b/integrations/safe-publish.php
@@ -1,0 +1,114 @@
+<?php
+/**
+ * Integration: Safe Publish.
+ *
+ * @package Automattic\VIP\Integrations
+ */
+
+namespace Automattic\VIP\Integrations;
+
+/**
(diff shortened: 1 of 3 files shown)


shekharnwagh (approved):
The copilot review flags the multisite scenario with missing mocks, which causes CI test failures. Once that's fixed, this should be fine, so approving.

maxschmeling:
Updated in d148293a2.

- Replaced the two env-config-only mocks with activate([ 'config' => ... ]) so those tests behave consistently in single-site and multisite runs.
- Added a multisite-only test that verifies the current network site config is used instead of env config.

PR #7127, opened 2026-07-23
Add Safe Publish Mirror integration loader

aswasif007:
## Description

Adds the platform loader for the [safe-publish-mirror](https://github.com/automattic/safe-publish-mirror/) integration. When the integration is active for a site, it loads the plugin from the `vip-integrations` drop directory and defines a single associative-array constant, `VIP_SAFE_PUBLISH_MIRROR_CONFIG`, from the platform-provided config — matching the fields in the partner handoff manifest (`connected_site_url`, `sync_mode`, `shared_secret`).

The constant is always defined while the integration is active, so the plugin can detect and report an incomplete configuration gracefully instead of fataling on a missing constant.

The `safe-publish-mirror` integration is a mirror of the existing `safe-publish` integration, but developed differently as a PoC of the VIP Integrations SDK project. This is an experimentation on whether VIP can build VIP's end of a partner-built integration, from a manifest file that we receive from the partner.

### DIFF
--- a/integrations/safe-publish-mirror.php
+++ b/integrations/safe-publish-mirror.php
@@ -58,21 +58,19 @@ class SafePublishMirrorIntegration extends Integration {
 				return;
 			}
 
-			$versions = $this->get_versions();
-
-			if ( empty( $versions ) ) {
+			$version_folder = $this->get_selected_version_folder( $this->get_versions() );
+			if ( '' === $version_folder ) {
 				$this->is_active = false;
 				return;
 			}
 
-			$selected_version_folder = $this->get_selected_version_folder( $versions );
-			$load_path               = WPVIP_MU_PLUGIN_DIR . '/vip-integrations/' . $selected_version_folder . '/safe-publish-mirror.php';
-
-			if ( file_exists( $load_path ) ) {
-				require_once $load_path;
-			} else {
+			$load_path = WPVIP_MU_PLUGIN_DIR . '/vip-integrations/' . $version_folder . '/safe-publish-mirror.php';
+			if ( ! file_exists( $load_path ) ) {
 				$this->is_active = false;
+				return;
 			}
+
+			require_once $load_path;
 		}, 1 );
 	}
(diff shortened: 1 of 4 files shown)


integrations/safe-publish-mirror.php
```diff
@@ -0,0 +1,134 @@
...
+class SafePublishMirrorIntegration extends Integration {
+	/**
+	 * The version of Safe Publish Mirror to load, defaults to the latest version.
+	 *
+	 * @var string
+	 */
+	public string $version = 'latest';
+
+	public function is_loaded(): bool {
+		return defined( 'SAFE_PUBLISH_MIRROR_LOADED' ) || defined( 'SAFE_PUBLISH_MIRROR_PLUGIN_FILE' );
```
pandah3, line 31: I think the ones in the plugin have the `VIP` prefix: https://github.com/Automattic/safe-publish-mirror/blob/main/safe-publish-mirror.php#L23-L25

```
define( 'VIP_SAFE_PUBLISH_MIRROR_LOADED', true );
define( 'VIP_SAFE_PUBLISH_MIRROR_VERSION', '1.0.0' );
define( 'VIP_SAFE_PUBLISH_MIRROR_FILE', __FILE__ );
```

pandah3 (commented):
One thing I noticed while digging into the manifest: it tells you what constant VIP passes in, but not what constants the plugin sets when it loads. We should prob add it to the manifest. Something like:

```
runtime:
  wordpress_plugin:
    entry_file: safe-publish-mirror.php
    loaded_constants:
      - VIP_SAFE_PUBLISH_MIRROR_LOADED
      - VIP_SAFE_PUBLISH_MIRROR_FILE
```

And then the validator can check against that.

---

Also, did we leave out the Pendo tracking intentionally? Did we resolve the Telemetry/tracking stuff in the contract?

aswasif007:
Closing because we won't be merging it anyway.

PR #7093, opened 2026-07-07
[PLTFRM-1061] Respect theme.json layout widths for image resize

rebeccahum:
## Description
Fixes #5339 

### Changed
- VIP Image Resize: Respect theme.json layout contentSize and wideSize when determining resize widths for block themes without $content_width.

## Pre-review checklist

Please make sure the items below have been covered before requesting a review:

- [ ] This change works and has been tested locally or in Codespaces (or has an appropriate fallback).
- [ ] This change works and has been tested on a sandbox.
- [ ] This change has relevant unit tests (if applicable).
- [ ] This change uses a rollout method to ease with deployment (if applicable - especially for large scale actions that require writes).
- [ ] This change has relevant documentation additions / updates (if applicable).
- [ ] I've created a changelog description that aligns with the provided examples.

## Pre-deploy checklist

- [ ] VIP staff: Ensure any alerts added/updated conform to internal standards (see internal documentation).

Issue #5339, opened 2024-03-15
Max image with should respect `contentWidth` & `wideWidth` of theme.json


fabiankaegy:
## Expected/Desired Behavior

When using a block / hybrid theme that contains a `theme.json` file the `settings.layout.contentSize` & `settings.layout.wideSize` sizes should be used to determine the maximum width attribute that is added to images.

## Actual Behavior

Currently the width gets defined by the `$content_width` variable which is no longer needed / used in themes that have a `theme.json`. (See documentation: https://developer.wordpress.org/themes/basics/theme-functions/#content-width)

https://github.com/Automattic/vip-go-mu-plugins/blob/e1e7f5543722aee8eba6a13ee667cbbce8fab9f6/a8c-files.php#L288-L291

Because of this block themes that don't define the old `$content_width` variable get the fallback value of 1024px applied which leads to blurry images when the align wide / align full option is used on an image.

PR #7131, opened 2026-07-23
Default theme.json image width filter to false

philipjohn:
## Description

PR #7093 changed the meaning of the `full` image size by constraining default image sizes to theme.json layout widths platform-wide. PR #7128 added the `vip_image_resize_use_theme_json_layout_widths` filter to allow opting out, but it still defaults to `true` — meaning core WordPress behaviour for what a "full" size image is remains altered across all sites unless they explicitly opt out.

Core behaviour should not be changed platform-wide. This PR flips the filter default to `false`, restoring core's default full-size image behaviour. Sites that want the theme.json layout width constraints can opt in by returning `true` from the filter.

## Changes

- `a8c-files.php`: default `vip_image_resize_use_theme_json_layout_widths` to `false` and update the docblock to reflect the opt-in behaviour.
- `tests/files/test-a8c-files-image-resize.php`: theme.json layout width tests now opt in via the filter; added a test asserting the default (opt-out) behaviour uses the core `1024` fallback width.

## Related

- #7093 (introduced the behaviour change)
- #7128 (added the opt-out filter)

### DIFF
--- a/a8c-files.php
+++ b/a8c-files.php
@@ -269,11 +269,14 @@ class A8C_Files {
 			/**
 			 * Filters whether theme.json layout widths are used when $content_width is unavailable.
 			 *
+			 * Defaults to false to preserve core's default full-size image behaviour. Set to true
+			 * to constrain default image sizes to the theme.json layout widths.
+			 *
 			 * @param bool         $use_theme_json_layout_widths Whether to use theme.json layout widths.
 			 * @param int          $id                           Attachment ID.
 			 * @param array|string $size                         Requested image size.
 			 */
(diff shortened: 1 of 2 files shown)


rebeccahum:
Heyoooo @philipjohn !
I don’t think we should default this filter to false b/c #5339 is still valid where falling back to 1024px causes blurry wide/full images & making this opt-in effectively restores that bug for every site that does not know about this VIP-specific filter.

rinatkhaziev:
@philipjohn we're looking/thinking. So far the thinking is that we should restore "full" behavior by handling that case separately. I do agree we shouldn't break the core contracts and that was not the intent here.

rebeccahum:
Closing in favour of #7132

PR #7198, opened 2026-08-24
refactor: prepare the codebase for mitigation metrics

sjinks:
* Load Prometheus before `vip-secrets.php` is loaded;
* Ensure that the metrics registry is initialized only once;
* Add a helper to retrieve the registry to register additional metrics.


000-vip-init.php
```diff
@@ -93,6 +93,9 @@
 	}
 }
 
+// Stats collection
+require_once __DIR__ . '/prometheus.php';
+
```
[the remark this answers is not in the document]
sjinks: You are late to the party.

000-vip-init.php
```diff
@@ -93,6 +93,15 @@
 	}
 }
 
+// Enable the Potential Multi Dataset Queries Prometheus Collector
+// This is a temporary measure to help us identify sites that are using multi dataset queries
+if ( ! defined( 'VIP_POTENTIAL_MULTI_DATASET_QUERIES_COLLECTOR_ENABLED' ) && defined( 'VIP_ENV_VAR_VIP_POTENTIAL_MULTI_DATASET_QUERIES_COLLECTOR_ENABLED' ) && constant( 'VIP_ENV_VAR_VIP_POTENTIAL_MULTI_DATASET_QUERIES_COLLECTOR_ENABLED' ) === 'true' ) {
+	define( 'VIP_POTENTIAL_MULTI_DATASET_QUERIES_COLLECTOR_ENABLED', true );
+}
+
```
luismulinari: FYI: All the code related to the potential multi-dataset query collector is no longer used and can be removed.

You can do it now (if it makes sense) or we can do it in a follow-up PR.

luismulinari (approved):

rinatkhaziev:
~@sjinks should we try loading in https://github.com/Automattic/vip-go-mu-plugins/blob/develop/000-pre-vip-config/requires.php or is it too early?~

nm, it is too early, as we call add_action in the plugin.

000-vip-init.php
rinatkhaziev: Let's just do it here.

PR #7215, opened 2026-08-28
files: remove redundant Files Service round trips and fix filesystem correctness bugs

luismulinari:
## What this changes

Removes Files Service round trips that were never needed, and fixes several filesystem operations that were returning wrong answers.

Scoped deliberately: no directory listing, no connection reuse, no virtual directories.

## Fewer calls

| | Before | After |
|---|---:|---:|
| Read a file | 2 calls | **1** |
| Overwrite a file | 3 calls | **1** |
| `WP_Filesystem` write then read | 4 calls | **1** |
| Ask 4 questions about one file | 4 calls | **1** |

- **Reads no longer pre-check.** `get_file()` asked `file_exists` before every download, to spot a 404 early and to size the timeout. The download answers both by itself, so the check is gone and its answer is taken from the response (`Last-Modified` + the downloaded size) and cached.
- **Writes no longer download first.** `fopen( $path, 'w' )` fetched the whole remote file before truncating it to zero.
- **One shared answer per path.** `file_exists()`, `is_file()`, `filesize()` and `filemtime()` now share a single lookup, and a read seeds it too. Absence is deliberately **not** cached: a missing file is re-checked on every call, as on `develop`, so a file created by another process during a long WP-CLI or cron run becomes visible at once.
- **Upload responses seed the same lookup** when the Files Service returns `mtime` and `size` after a PUT. That field is added by the Files Service branch `files-service-improvements` and is not yet on `trunk`, so until it ships the client falls back to clearing the stats after an upload, exactly as `develop` does. The "write then read" row above depends on it.

## Correctness fixes

Some of this is repair, not optimisation:

- **`rename()` remote → local wrote the wrong bytes.** It treated `get_file()`'s return value as file contents; it is a path. The destination received the string `/tmp/vipXXXXXX`.
- **`rename()` local → remote could not work.** It passed an in-memory handle to a function whose first line is `file_exists()` on that argument.
- **`size()` was wrong for binary files.** It downloaded the file and measured with `mb_strlen()`, which counts characters, not bytes.
- **Cache keys did not always match.** Paths were stored without a leading slash and returned with one, so one file could occupy two entries and neither be found.
- **A thrown upload leaked its file into later requests.** `Curl_Streamer` attaches the body via the global `http_api_curl` hook; if the request threw, the hook stayed registered and the next unrelated outbound request would have carried our file as its body. Now removed in a `finally`.
- **`mtime()` / `atime()`** were unimplemented.
- **Failed downloads leaked a temp file.** `get_file()` creates its `tempnam()` target before the request and only hands it to the cache on a 200. Every 404, error response, transport error and rejected path left an orphan in the temp dir that shutdown cleanup never saw, and the transport had written the error body into it. Now removed on every non-200 exit.

## Timeouts

Without the pre-check there is no size to scale a download timeout from. Replaced with:

- **Known size:** scale from the cached size when a prior stat or read supplied one, using the same 1 s per 500 KB formula uploads use.
- **Unknown size:** scale from `wp_max_upload_size()`, capped at **300 s in web requests**. The platform limit is 4 GB, which uncapped would allow 8398 s (2.3 h) per cold read. WP-CLI and cron keep the full size-derived value because long transfers are legitimate there.
- **Stall guard:** a download averaging below **256 KiB/s for 30 s** is aborted. That is half the rate the timeout formula already assumes, so the guard is more tolerant than the budget it protects while still cutting off a crawling connection in seconds rather than minutes. Worst-case hold time for a 10 MB cold read drops from the full ceiling to about 40 s. The guard rides on `http_api_curl`, so it exists only on the cURL transport; on any other transport the ceiling above is the only limit.

## Benchmark

25 comparable scenarios, PHP 8.2.33 / WP 7.1, 20 measured runs, 180 KiB payload.

```
calls               : 73 -> 44   (-29, -40%)
fewer / same / more : 18 / 7 / 0
summed median time  : 11.02s -> 10.74s  (-2.6%)
```

Nothing makes more calls than before. Two scenarios that failed every run now pass (`local-file`, `rename-local-to-remote`).

**On timing:** per-scenario numbers from this box are not trustworthy — several scenarios moved 15-60% with identical call counts, which no code path explains. The aggregate is the only figure worth reading, and it says "no worse".


rinatkhaziev (approved):
🤞🏻

PR #7223, opened 2026-09-01
Add VIP Workflows integration loader

maxschmeling:
## Summary

- register a `vip-workflows` MU integration wrapper
- load the latest versioned bundle on `plugins_loaded` with site-copy guards
- use the plural v0.0.2 constants and `vip-workflows.php` entry file
- add focused loader, version-selection, sentinel, and tracking tests

## Validation

- PHP syntax checks for all changed PHP files
- targeted PHPCS
- `git diff --check`
- PHPUnit was not available locally because OrbStack/Docker was stopped and the host WordPress test library was not installed; GitHub CI will run the suite

Linear: https://linear.app/a8c/issue/VIPPROD-610/c3-surface-3-vip-go-mu-plugins-wrapper-safe-publish-loader-model

### DIFF
--- a/integrations/vip-workflows.php
+++ b/integrations/vip-workflows.php
@@ -32,27 +32,34 @@ class VipWorkflowsIntegration extends Integration {
 	}
 
 	public function load(): void {
-		add_action( 'plugins_loaded', function () {
-			if ( $this->is_loaded() ) {
-				return;
-			}
-
-			$versions = $this->get_versions();
-
-			if ( empty( $versions ) ) {
-				$this->is_active = false;
-				return;
-			}
-
-			$selected_version_folder = $this->get_selected_version_folder( $versions );
-			$load_path               = WPVIP_MU_PLUGIN_DIR . '/vip-integrations/' . $selected_version_folder . '/vip-workflows.php';
-
-			if ( file_exists( $load_path ) ) {
-				require_once $load_path;
-			} else {
-				$this->is_active = false;
-			}
+		add_action( 'plugins_loaded', function (): void {
+			$this->load_plugin();
 		}, 1 );
+	}
+
+	/**
+	 * Load the selected bundled plugin file when it is available.
+	 */
+	private function load_plugin(): void {
+		if ( $this->is_loaded() ) {
+			return;
+		}
+
+		$versions = $this->get_versions();
+		if ( [] === $versions ) {
+			$this->is_active = false;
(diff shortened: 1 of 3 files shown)


integrations/vip-workflows.php
```diff
@@ -0,0 +1,95 @@
...
+
+	/**
+	 * Enable Pendo tracking for this integration.
+	 *
+	 * @var bool
+	 */
+	protected bool $enable_pendo_tracking = true;
+
+	public function configure(): void {
+		$configs = is_multisite() ? $this->get_network_site_config() : $this->get_env_config();
```
shekharnwagh, line 31: If we ever needed to combine integration config at both site and network site level we would need to combine them - https://github.com/Automattic/vip-go-mu-plugins/blob/b3893a00c00eac5514d1e8e1777c5b535b388c85/integrations/safe-publish.php#L117

I don't think we have any configs as of now so should be fine.

vip-integrations.php
```diff
@@ -50,6 +50,11 @@
 	require_once __DIR__ . '/integrations/safe-publish.php';
 }
 
+$vip_workflows_wrapper = __DIR__ . '/integrations/vip-workflows.php';
+if ( file_exists( $vip_workflows_wrapper ) ) {
+	require_once $vip_workflows_wrapper;
+}
```
shekharnwagh: Minor nit for consistency: Why not use the path inline like other integrations above this?

vip-integrations.php
```diff
@@ -81,6 +86,11 @@
 	IntegrationsSingleton::instance()->register( new SafePublishIntegration( 'safe-publish' ) );
 }
 
+if ( class_exists( __NAMESPACE__ . '\\VipWorkflowsIntegration' ) ) {
+	$vip_workflows_integration = new VipWorkflowsIntegration( 'vip-workflows' );
+	IntegrationsSingleton::instance()->register( $vip_workflows_integration );
+}
```
shekharnwagh, line 90: Same as above for consistency skip the intermediate variable

shekharnwagh (approved):

vip-integrations.php
maxschmeling: 52f5a5ca6

vip-integrations.php
maxschmeling: 52f5a5ca6

Issue #1368, opened 2019-12-03
Jetpack: Mandatory "manage" module causing errors


WPprodigy:
We require the `manage` module here: https://github.com/Automattic/vip-go-mu-plugins/blob/master/vip-jetpack/jetpack-mandatory.php#L20

Jetpack then tries to load this module here: https://github.com/Automattic/jetpack/blob/master/class.jetpack.php#L1862

But there is no `modules/manage.php` file, so you end up with a PHP error:

> PHP Warning:  include_once(): Failed opening '/wp-content/mu-plugins/jetpack/modules/manage.php' for inclusion (include_path='.') in /wp-content/mu-plugins/jetpack/class.jetpack.php on line 1860

Looks like manage is no longer a module, so we should just remove this entry: https://github.com/Automattic/jetpack/pull/9089

simonwheatley:
Noted in JIRA as [GH-3344]

WPprodigy:
Just noting that the `manage.php` file was removed here: https://github.com/Automattic/jetpack/pull/13374

PR #7242, opened 2026-09-11
Add Connector Controls runtime integration

maxschmeling:
## Description

Adds the runtime integration that supplies centrally managed credentials to the OpenAI, Anthropic, and Google connectors built into WordPress 7.0.

The integration resolves the most-specific network-site, environment, or organization credential and defines its provider API-key constant during integration configuration. Existing nonempty environment variables and nonempty string customer constants retain precedence. Customer constants are never redefined. The AI Client discovers the selected value through its normal provider-registration path, so ordinary constant delivery does not replace authentication explicitly installed by another plugin.

Two compatibility cases use a runtime fallback at `wp_connectors_init`: an empty environment variable masking the selected constant, and an existing empty/non-string constant that cannot be replaced with the managed key. The empty-environment fix also honors a valid customer constant when no managed credential is assigned. Environment variables are never modified.

While active, the integration makes matching database options inert, rejects option updates, sanitizes option creation to prevent credential persistence, and keeps the three Core fields read-only. Terminal guards on both the option-specific and global read/update hooks prevent later global filters from restoring database credentials or changing stored values. Custom connectors and unrelated options are untouched. Explicit per-provider opt-outs stop managed credential inheritance; existing external credentials remain eligible.

The WordPress Connectors cards append Integration Center guidance to the existing provider descriptions through Core's supported `wp_connectors_init` metadata override. They distinguish managed credentials, customer environment variables/constants, no configured credential, unavailable providers, and unknown sources using the same observed runtime status as SDS. Other metadata and custom connectors remain unchanged; Core's credential-field source message remains visible. Descriptions observe the registry at connector initialization, while SDS reads it again when collecting the report.

The existing SDS report includes only per-provider source/status and an active flag. It inspects authentication actually held by the AI Client: constants created by the integration report `integration`, pre-existing customer constants report `php_constant`, and environment variables report `environment_variable`. Missing providers, empty credentials, and unrecognized overrides are reported separately. It guards switched-blog contexts and adds no network request or option write during collection. Existing SDS timestamps and heartbeats are reused.

## Changelog Description

Omitted while the feature remains restricted to an internal rollout.

## Pre-review checklist

- [x] Tested locally against WordPress 7.0 in single-site and multisite mode.
- [x] Added coverage for constant delivery, external precedence, empty-value fallbacks, runtime reporting, database isolation, write protection, org gating, field locking, and management descriptions with metadata preservation.
- [ ] Test on a sandbox with the matching platform and Dashboard changes.
- [x] Rollout is controlled by integration status and Dashboard feature flags.

## Pre-deploy checklist

- [x] No alerts are added or changed.

## Steps to Test

1. Run the Connector Controls, SDS collector, and smoke tests in single-site mode, then with `--multisite 1`.
2. Enable Connector Controls with an OpenAI credential in the generated integration config. Confirm `OPENAI_API_KEY` is defined before provider registration, the AI Client discovers it, and Core's credential field is read-only.
3. Repeat with a customer-defined constant and confirm it is preserved and reported as `php_constant`; the integration-created constant must report `integration`.
4. Repeat with a nonempty environment variable and confirm it takes precedence and is reported as `environment_variable`.
5. Set the environment variable to an empty string and confirm the selected constant reaches the AI Client via fallback. Cover both a managed credential and a customer constant with no managed assignment.
6. Define an empty or non-string constant and confirm it remains unchanged while the managed credential reaches the registry through fallback and reports `integration`.
7. Store a sentinel value in `connectors_ai_openai_api_key` and confirm Core cannot read or update it while the integration is active. Adding a missing credential option must not persist a submitted key. Global `pre_option` filters returning either a credential or `false`, and global `pre_update_option` filters injecting a value, must not bypass these guards. Unrelated options must remain readable and writable with their existing filters.
8. Inspect `connector_controls` in the SDS payload: no key, key fragment, or fingerprint may be present. An unrecognized registry override must report an unknown source.
9. With matching SDS/API/Dashboard changes, confirm Integration Center shows the last observed source and timestamp separately from the managed assignment.
10. Open Settings → Connectors. Confirm each provider keeps its original description and gains Integration Center guidance matching the initialized runtime source. Existing environment/constant field help remains unchanged; custom and removed connectors are untouched.

Normal constant discovery requires the constant to exist before provider registration. Exceptional runtime fallbacks require the provider to be registered by the callback. Runtime reporting reflects the actual registry state rather than assuming the selected credential was applied. The no-credential/fallback Core field-lock marker still uses Core's constant label; the card description adds management context and the Integration Center report provides the observed source/status.

Runtime status does not verify provider acceptance or exact credential identity after a recent assignment change.

Related PRs: [API #7852](https://github.com/Automattic/vip-go-api/pull/7852), [Dashboard #5551](https://github.com/Automattic/vip-dashboard/pull/5551), [SDS #1368](https://github.com/Automattic/vip-site-details/pull/1368), and [SDK types #832](https://github.com/Automattic/vip-go-node-internal/pull/832). Deploy SDS before consumers request the new runtime metadata. No deployment or package publication is included.

Validation: WordPress 7.0 Connector Controls, SDS collector, and smoke suites pass: PHP 8.2 single-site 83 tests / 438 assertions (4 multisite-only skips), PHP 8.4 multisite 83 tests / 444 assertions. The nine global read/write bypass regression cases fail against the previous implementation and pass with the guards. PHP lint, PHPCS for the changed files, and diff checks pass.


integrations/connector-controls.php
```diff
@@ -0,0 +1,145 @@
...
+		}
+	}
+
+	/**
+	 * Keep managed database credentials inert and reject option writes.
+	 *
+	 * @param string $option_name WordPress option name.
+	 */
+	private function make_database_credential_inert( string $option_name ): void {
+		add_filter( "pre_option_{$option_name}", '__return_empty_string' );
```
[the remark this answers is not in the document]
maxschmeling: Fixed in [64630abd0](https://github.com/Automattic/vip-go-mu-plugins/commit/64630abd0). Both managed option read and write filters now run at PHP_INT_MAX, with a regression test proving later ordinary-priority callbacks cannot reveal or persist the stored value.

integrations/connector-controls.php
```diff
@@ -0,0 +1,145 @@
...
+			if ( ! is_string( $credential ) || '' === trim( $credential ) ) {
+				continue;
+			}
+
+			// Core gives environment variables precedence over constants. Preserve either
+			// existing source rather than silently replacing customer configuration.
+			$environment_credential = getenv( $connector['env'] );
+			if ( ( is_string( $environment_credential ) && '' !== trim( $environment_credential ) ) || defined( $connector['constant'] ) ) {
+				continue;
+			}
```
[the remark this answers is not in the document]
maxschmeling: Fixed in [0004c62e1](https://github.com/Automattic/vip-go-mu-plugins/commit/0004c62e1) and covered further in [78dab9844](https://github.com/Automattic/vip-go-mu-plugins/commit/78dab9844). The integration now matches Core’s raw non-empty environment check and non-empty-string constant check. Because PHP cannot redefine an invalid existing constant, that edge is handled by applying the managed credential to the AI Client registry after Core registers providers.

maxschmeling:
Addressed all four Copilot findings:

- [bf79b3dd3](https://github.com/Automattic/vip-go-mu-plugins/commit/bf79b3dd3) gates Connector Controls on the WordPress Connectors API before changing constants, options, or UI state.
- [64630abd0](https://github.com/Automattic/vip-go-mu-plugins/commit/64630abd0) enforces managed option reads and writes at terminal priority.
- [0004c62e1](https://github.com/Automattic/vip-go-mu-plugins/commit/0004c62e1) matches Core credential-source semantics and handles invalid pre-defined constants without attempting to redefine them.
- [78dab9844](https://github.com/Automattic/vip-go-mu-plugins/commit/78dab9844) covers preservation of customer environment credentials, including Core’s whitespace behavior.

Verified with the focused suite on WordPress 7.0 single-site (15 tests / 30 assertions), WordPress 7.0 multisite (15 / 32), and the unsupported-version gate on WordPress 6.5. Targeted PHPCS, repository PHP lint, and smoke tests pass. Repository-wide PHPCS is currently blocked by the pre-existing deprecated finfo_close() call in files/class-api-client.php:190.

maxschmeling:
Addressed [this review](https://github.com/Automattic/vip-go-mu-plugins/pull/7242#pullrequestreview-5239425068) in 2929cb22c:

- Moved the connector field lock to `PHP_INT_MAX`, with a regression test that attempts to reopen all three managed fields at priority 101. Custom connectors remain unchanged.
- Protected option creation with a terminal `sanitize_option_{$option_name}` filter. WordPress 7.0 does not have the suggested `pre_add_option_*` filter. When a missing option is cached, `add_option()` can still create an empty row, but the submitted credential is discarded. Tests cover cached and uncached creation for all three providers, checking the database directly and reading again after removing the read filter.
- Applied the AI Client runtime fallback when an empty environment variable masks the provider constant. This uses the managed key when needed and preserves a valid existing external constant. Tests cover undefined, valid, empty, and invalid constants.

Validation on WordPress 7.0 / PHP 8.2: connector suite plus smoke test passed in single-site mode (27 tests, 80 assertions, two expected multisite-only skips) and multisite mode (27 tests, 82 assertions). `npm run phplint` and `npm run phpcs` passed. Running the new regressions against the original implementation reproduced six failing cases spanning all three findings.

maxschmeling:
Simplified credential delivery in d111235e2: every selected credential now goes directly to the AI Client registry on `wp_connectors_init`. Removed provider constant creation and the special empty-environment/invalid-constant fallback paths.

Credential selection still preserves the existing precedence: nonempty environment variable, then a valid existing PHP constant, then the resolved managed credential. Explicit managed opt-outs and the database/UI safeguards remain in place. The selected credential replaces earlier registry authentication; managed keys are no longer exposed through newly defined provider constants.

Updated the PR description and regression tests. An actual AI Client registry test verifies direct authentication, empty-environment handling, missing-provider handling, and survival through Core's later database handoff.

Validation on WordPress 7.0 / PHP 8.2: connector suite plus smoke passed in single-site mode (38 tests, 142 assertions, two expected multisite-only skips) and multisite mode (38 tests, 145 assertions). `npm run phplint` and `npm run phpcs` passed.

maxschmeling:
Implemented the runtime source reporting in bc33a05c7.

- The collector inspects the authentication currently held by the AI Client. It reports managed credentials, environment variables, PHP constants, no credential, unavailable providers, or an unknown source after an unrecognized override.
- The existing SDS payload carries only an active flag and provider source/status enums. Collection adds no HTTP request or option write, and sends no key, key fragment, or fingerprint. Existing SDS timestamps/change detection/heartbeats are reused.
- A switched blog cannot report the original blog's runtime. Inactive and unavailable observations are explicit.

PHP lint/PHPCS pass. WordPress 7.0/PHP 8.2 Connector Controls, SDS collector, and smoke coverage passes: single-site 61 tests/238 assertions (4 multisite-only skips), multisite 61 tests/243 assertions.

This connects to [SDS #1368](https://github.com/Automattic/vip-site-details/pull/1368), [API #7852](https://github.com/Automattic/vip-go-api/pull/7852), and [Dashboard #5551](https://github.com/Automattic/vip-dashboard/pull/5551); [SDK #832](https://github.com/Automattic/vip-go-node-internal/pull/832) supplies the shared types. These reports describe the last observed source, not provider acceptance or proof that a newly saved assignment has propagated.

maxschmeling:
Pushed a251ec10b to restore provider constant creation while retaining the SDS runtime reporting.

- On the normal path, Connector Controls defines the provider constant from the resolved managed credential and lets the AI Client discover it during provider registration.
- Existing nonempty environment variables and nonempty string customer constants retain precedence. Customer constants are never redefined.
- SDS distinguishes constants we created (`integration`) from pre-existing customer constants (`php_constant`) by inspecting the actual AI Client authentication. A later unrecognized override reports `unknown`.
- Direct injection remains only for compatibility fallbacks: an empty environment variable masks the chosen constant, or an existing empty/non-string constant cannot be replaced. The empty-environment repair also works for customer constants with no managed assignment.
- Database credential protections and the existing SDS/API/Dashboard contract are unchanged. Normal constant delivery no longer overwrites earlier explicit registry authentication.

Tests cover real AI Client constant discovery, constant ownership (including identical managed/customer values), environment precedence, empty/invalid values, late provider registration, custom authentication, and SDS source reporting. PHP lint and PHPCS pass.

Core's constant label is now backed by a real constant for ordinary managed credentials. The field-lock fallback for an unconfigured provider still uses Core's constant marker; Integration Center reports the actual source/status. No deployment is included.

Validation: WordPress 7.0 / PHP 8.2 Connector Controls, SDS collector, and smoke suites pass: single-site 71 tests / 322 assertions (4 multisite-only skips), multisite 71 tests / 328 assertions. PHP lint, PHPCS, and diff checks pass.

maxschmeling:
Added the WordPress Connectors card descriptions in 9dadae0ecc9fd8c96b480838e0facc287a75d1ea.

The built-in OpenAI, Anthropic, and Google cards retain their provider descriptions and append Integration Center guidance. Managed credentials show “Credentials managed in VIP Integration Center.” Customer constants and environment variables are identified as taking precedence; missing credentials, unavailable providers, and unknown sources get separate explanations.

This uses Core's supported `wp_connectors_init` unregister/modify/register pattern and the existing runtime-source detection. Authentication, logos, plugin metadata, and Core's field help are preserved. Custom connectors and connectors removed by other code are untouched. The descriptions observe the initialized registry; SDS continues to inspect it when collecting its report, including any later custom authentication changes.

Validation: WordPress 7.0 / PHP 8.2 Connector Controls, SDS collector, and smoke suites passed in single-site mode (71 tests, 399 assertions; 4 multisite-only skips) and multisite mode (71 tests, 405 assertions). PHP lint, PHPCS, and diff checks passed. The local preview now matches the implemented descriptions.

maxschmeling:
Fixed the PHP 8.5 test failures in 8007b9c631bf0a2ad0e6d00b0c9a32e5fc24e2d8.

Both failing jobs reported 20 errors from the same two `ReflectionProperty::setAccessible()` calls in the Connector Controls registry test helper. PHP 8.5 deprecates that method; it has had no effect since PHP 8.1. Removed both calls while preserving registry setup, restoration, and all assertions. Production code is unchanged.

Validation:
- Fresh CI passes for both PHP 8.5 single-site and multisite jobs: https://github.com/Automattic/vip-go-mu-plugins/actions/runs/35384323590.
- Local WordPress 7.0 Connector Controls and smoke tests pass on PHP 8.2 single-site (61 tests, 344 assertions, 4 multisite-only skips) and PHP 8.4 multisite (61 tests, 350 assertions).
- Repository PHP lint, PHPCS for the changed file, and `git diff --check` pass.

integrations/connector-controls.php
```diff
@@ -0,0 +1,236 @@
...
+		return null;
+	}
+
+	/**
+	 * Keep managed database credentials inert and prevent credential writes.
+	 *
+	 * @param string $option_name WordPress option name.
+	 */
+	private function make_database_credential_inert( string $option_name ): void {
+		add_filter( "pre_option_{$option_name}", '__return_empty_string', PHP_INT_MAX );
```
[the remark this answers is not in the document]
maxschmeling: Addressed in da1ea6a5d. `configure()` now registers terminal `pre_option` and `pre_update_option` guards after the integration's enablement/API checks. They match only the three managed credential option names: reads return an empty string, updates return the old value, and other options preserve the incoming filter result. The existing option-specific and insert-sanitization guards remain in place.

Regression tests cover global read filters injecting a credential or returning `false` to restore database fallback, global write filters injecting replacement credentials, all three providers, unrelated/custom connector options, and disabled/unavailable integration states. All nine managed-option bypass cases fail against the previous code and pass with this fix.

WordPress 7.0 Connector Controls, SDS collector, and smoke tests pass in single-site mode on PHP 8.2 (83 tests, 438 assertions, four multisite-only skips) and multisite mode on PHP 8.4 (83 tests, 444 assertions). PHP lint, scoped PHPCS, and diff checks pass.

integrations/connector-controls.php
```diff
@@ -0,0 +1,416 @@
...
+			$this->make_database_credential_inert( $connector['option'] );
+
+			$credential = $this->resolve_credential( $connector );
+
+			// Non-empty environment variables take precedence in both Core and the
+			// AI Client. Leave their existing authentication path unchanged.
+			$environment_credential = getenv( $connector['env'] );
+			if ( false !== $environment_credential && '' !== $environment_credential ) {
+				continue;
+			}
```
shekharnwagh, line 111: This check misses keys set with `vip config envvar set`. Those exist only as `VIP_ENV_VAR_OPENAI_API_KEY` until `vip-helpers/vip-connectors.php` runs at `init` 16. So `getenv()` is false here, and we define the managed constant. The helper then puts the customer key into the registry later after this.

That means customer key wins and SDS reports `environment_variable`, which is expected. But the card text is wrong. `describe_managed_connectors` runs on `wp_connectors_init`, and core fires that hook inside [_wp_connectors_init at init 15](https://github.com/WordPress/wordpress-develop/blob/1034cdd7ed2af97b7bb888d0a3e5faa3369bb31b/src/wp-includes/default-filters.php#L563). So even at PHP_INT_MAX, it runs before the helper. The card says "managed in VIP Integration Center" while the site uses the customer key.

integrations/connector-controls.php
```diff
@@ -0,0 +1,416 @@
...
+		$report = [
+			'active'    => null !== $this->configured_blog_id,
+			'providers' => array_fill_keys( array_keys( self::CONNECTORS ), [
+				'source' => 'unknown',
+				'status' => 'unknown',
+			] ),
+		];
+		if ( ! $report['active'] || ! did_action( 'wp_connectors_init' ) || ! class_exists( AiClient::class ) ) {
+			return $report;
+		}
```
shekharnwagh, line 213: Even when the integration is not enabled, we always send this to SDS, which stores this in its DB. Returning `null` means we don't add anything to the SDS DB. If the latter is acceptable, should we do that?

shekharnwagh (approved):

integrations/connector-controls.php
maxschmeling: https://github.com/Automattic/vip-go-mu-plugins/pull/7242/commits/af95c1c0c8df7626b51faffed97acbd370e429d3

integrations/connector-controls.php
maxschmeling: 563d7adcce054c212dffafa7ea2552572968ebd7

PR #7258, opened 2026-09-15
Add internal-only Abilities API invocation telemetry

smithjw1:
## Description

Adds a small, internal-only telemetry mechanism to measure how registered WordPress Abilities are invoked (name, success), following internal discussion about starting to collect some Core usage telemetry without VIP's usual all-or-nothing rollout. Two independent pieces:

1. A reusable `telemetry-internal-only` feature flag (environment-ID allowlist) on `Automattic\VIP\Feature`, so any internal-only telemetry probe can scope itself to a small set of internal test environments rather than production-wide.
2. A first probe built on it: wraps every registered WP Ability's `execute_callback` via the `wp_register_ability_args` filter (the only extension point the Abilities API fires around registration) and records a Tracks event per invocation — gated behind that flag, so it only runs on the allowlisted test environments.

Baseline overhead measured on one allowlisted test environment before this change: the ability-execution span itself (`WP_Ability::invoke_callback`) runs ~1.1ms of a ~114ms total request. The actual before/after comparison with this change deployed is still pending.

### Added
- Added a `telemetry-internal-only` feature flag environment allowlist to `Automattic\VIP\Feature`.
- Added internal-only telemetry for WP Abilities API invocations, gated to a small set of internal test environments.

## Pre-review checklist

- [x] This change works and has been tested locally or in Codespaces (or has an appropriate fallback). — unit tests pass locally against WordPress 7.1 via this repo's own Docker test runner; also verified against a live test environment
- [ ] This change works and has been tested on a sandbox. N/A
- [x] This change has relevant unit tests (if applicable). — `tests/telemetry/test-abilities.php`: covers untouched args without a callback, the wrapped callback still returning the original result, and both success and WP_Error outcomes recording the right event/properties
- [x] This change uses a rollout method to ease with deployment (if applicable - especially for large scale actions that require writes). — feature-flag-gated to a handful of environments
- [ ] This change has relevant documentation additions / updates (if applicable). N/A

## Steps to Test

1. Check out this branch.
2. On an environment where `FILES_CLIENT_SITE_ID` is one of the IDs in `Feature::$feature_ids['telemetry-internal-only']`, invoke a registered ability, e.g. `GET /wp-json/wp-abilities/v1/abilities/core/get-environment-info/run`.
3. Confirm a `vip_abilities_api_invoked` Tracks event fires with `ability_name`/`success`.
4. On an environment *not* in the allowlist, repeat and confirm no event fires.


smithjw1:
Good Co-pilot review; did most of what it asked, but I didn't  "add process-isolated coverage for allowlisted and non-allowlisted environment IDs, or refactor the registration predicate into a testable function."

The existing test file already documents why it bypasses the top-level gate, a pattern that test-stats.php uses for XML_RPC_Auth_Tracker, which also doesn't test its own top-level environment/feature-flag gate directly.

Adding gate-boundary coverage here would mean different methods and a lot of code with little value.

